PR #8369, opened 2024-10-17
ci: Enforce `dev` as base branch for development

Pwuts:
We want to keep `master` as the default branch for easy and safe cloning, but `dev` as the target for all pull requests.

### Changes 🏗️

- ci: Add a workflow that automatically changes PRs' target branch from `master` to `dev` unless the head branch is `dev` or `hotfix/*`


Torantulino (approved):

aarushik93 (approved):

Pwuts:
This won't work until it is in `master` btw @Torantulino @aarushik93

PR #8358, opened 2024-10-16
feat(platform, blocks): Webhook-triggered blocks

Pwuts:
- Resolves #8352

## Changes 🏗️

- feat(blocks): Add GitHub Pull Request Trigger block

### feat(platform): Add support for Webhook-triggered blocks
- ⚠️ Add `PLATFORM_BASE_URL` setting

- Add webhook config option and `BlockType.WEBHOOK` to `Block`
  - Add check to `Block.__init__` to enforce type and shape of webhook event filter
  - Add check to `Block.__init__` to enforce `payload` input on webhook blocks
  - Add check to `Block.__init__` to disable webhook blocks if `PLATFORM_BASE_URL` is not set

- Add `Webhook` model + CRUD functions in `backend.data.integrations` to represent webhooks created by our system
  - Add `IntegrationWebhook` to DB schema + reference `AgentGraphNode.webhook_id`
    - Add `set_node_webhook(..)` in `backend.data.graph`

- Add webhook-related endpoints:
  - `POST /integrations/{provider}/webhooks/{webhook_id}/ingress` endpoint, to receive webhook payloads, and for all associated nodes create graph executions
    - Add `Node.is_triggered_by_event_type(..)` helper method
  - `POST /integrations/{provider}/webhooks/{webhook_id}/ping` endpoint, to allow testing a webhook
  - Add `WebhookEvent` + pub/sub functions in `backend.data.integrations`

- Add `backend.integrations.webhooks` module, including:
  - `graph_lifecycle_hooks`, e.g. `on_graph_activate(..)`, to handle corresponding webhook creation etc.
    - Add calls to these hooks in the graph create/update endpoints
  - `BaseWebhooksManager` + `GithubWebhooksManager` to handle creating + registering, removing + deregistering, and retrieving existing webhooks, and validating incoming payloads

### Other improvements
- fix(blocks): Allow having an input and output pin with the same name
- fix(blocks): Add tooltip with description in places where block inputs are rendered without `NodeHandle`
- feat(blocks): Allow hiding inputs (e.g. `payload`) with `SchemaField(hidden=True)`
- fix(frontend): Fix `MultiSelector` component styling
- feat(frontend): Add `AlertDialog` UI component
- feat(frontend): Add `NodeMultiSelectInput` component
- feat(backend/data): Add `NodeModel` with `graph_id`, `graph_version`; `GraphModel` with `user_id`
  - Add `make_graph_model(..)` helper function in `backend.data.graph`
[...]

PR #8375, opened 2024-10-18
refactor(platform): Update database model to reflect domain

Swiftyos:
# Background

We are enhancing our platform’s capabilities by introducing several new features that require an updated database schema. These features include group support, user-specific agent configurations, an accounting and credit system, and a marketplace for agent listings. These changes are essential to support the domino feature, which integrates these components to provide a more robust and user-friendly experience.

## Changes 🏗️

- **Renamed AgentGraph to Agent**:
  - Updated the AgentGraph model to Agent for clarity.
  - Adjusted all references and relations to accommodate the new naming.
  - Added `@@map("AgentGraph")` to maintain backward compatibility with the existing database.

- **Group Support**:
  - Introduced a new `UserGroup` model to enable user group functionalities.
  - Fields include `id`, `name`, `description`, `groupIconUrl`, and relations to `User`, `Agent`, `Profile`, and `StoreListing`.

- **Updated the User model**:
  - Added relations `groups` and `adminGroups` to represent group memberships and admin roles.
  - Included additional relations for accounting and subscriptions.

- **User-Specific Agent Configurations**:
  - Added `ConfiguredAgent` model:
    - Stores user-specific configurations of agents.
    - Fields include `id`, `name`, `description`, and relations to `User`, `Agent`, and `AgentNodeExecutionInputOutput`.
  - Introduced `UserAgents` model:
    - Tracks agents associated with users, including favorites and published agents.
    - Fields include `id`, `userId`, `agentId`, and flags like `isFavorite`, `isPublished`, etc.

- **Accounting and Credit System**:
  - Created `UserAccounting` model to manage user balances and transactions.
    - Fields include `id`, `userId`, `usdBalance`, and relations to `UserBlockCredit`.
  - Updated `UserBlockCredit` model:
    - Added references to `UserAccounting`, `Agent`, and `StoreListing`.
    - Included new enum values in `UserBlockCreditType`: `COMMISSION`, `PURCHASE`, `SALE`.
  - Added `UserSubscription` and `SubscriptionPlan` models:
    - `UserSubscription` tracks user subscriptions, including status and plan details.
    - `SubscriptionPlan` defines available subscription plans with pricing and credit information.

- **Marketplace Functionality**:
  - Introduced `Profile` model to represent user or group profiles in the marketplace.
    - Fields include `id`, `username`, `description`, social links, and avatar URL.
  - Added `StoreListing` model to enable users to list agents for sale.
    - Fields include `id`, `slug`, `name`, `description`, and relations to `Agent`, `User`, `UserGroup`, and store-related models.
  - Created `StoreListingVersion`, `StorePricing`, and `StoreListingApproval` models to handle versioning, pricing, and approval workflows for listings.

- **Updated Triggering Mechanism**:
  - Modified `AgentGraphExecutionSchedule` to now trigger a `ConfiguredAgent` using the input data from there
  - Add `WebhookTriggers` table for tracking web hook triggers
  - Added `triggerIdentifier` so agents can identify what tigger resulted in there execution

- **Updated Existing Models**:
  - Modified models like `AgentNode`, `AgentGraphExecution`, and `AgentNodeExecutionInputOutput`:
    - Updated references from `AgentGraph` to `Agent`.
    - Adjusted relations and fields to align with the new schema structure.
    - Added indexes and comments for better query performance and readability.

- **Organized the schema** into sections with clear comments for maintainability.

These changes collectively update the database schema to support the new features associated with the domino initiative, enhancing the platform’s functionality and scalability.


autogpt_platform/backend/schema.prisma
```diff
@@ -19,45 +19,141 @@ model User {
...
-  isActive    Boolean @default(true)
-  isTemplate  Boolean @default(false)
 
   // Link to User model
-  userId String
-  user   User   @relation(fields: [userId], references: [id], onDelete: Cascade)
+  createdByUserId String
+  createdByUser   User   @relation(fields: [createdByUserId], references: [id], onDelete: Cascade)
+
+  isGroupAgent Boolean    @default(false)
```
Swiftyos: Agents ownership can be changed to a group, in which case only group members can modify them

autogpt_platform/backend/schema.prisma
```diff
@@ -19,45 +19,141 @@ model User {
...
+  @@map("AgentGraph")
+}
+
+//////////////////////////////////////////////////////////// 
+//////////////////////////////////////////////////////////// 
+////////////////   USER SPECIFIC DATA   ////////////////////
+//////////////////////////////////////////////////////////// 
+//////////////////////////////////////////////////////////// 
+
+model ConfiguredAgent {
```
Swiftyos: When an agent is configured it can be triggered by a web hook or cronjob

autogpt_platform/backend/schema.prisma
```diff
@@ -19,45 +19,141 @@ model User {
...
+  agentVersion Int
+  agent        Agent  @relation(fields: [agentId, agentVersion], references: [id, version])
+
+  Configuration AgentNodeExecutionInputOutput[] @relation("AgentNodeExecutionInput")
+  UserAgents    UserAgents[]
+
+  @@index([userId])
 }
 
+model UserAgents {
```
Swiftyos: A User managed list of agents that they can see in there library

autogpt_platform/backend/schema.prisma
```diff
@@ -257,9 +361,18 @@ model AnalyticsMetrics {
...
+
 enum UserBlockCreditType {
   TOP_UP
   USAGE
+  COMMISSION
+  PURCHASE
+  SALE
 }
 
 model UserBlockCredit {
```
Swiftyos: This probably should be renamed... Not 100% Sure about how I'm handling the accounting yet. I could do with some input @majdyz

autogpt_platform/backend/schema.prisma
```diff
@@ -272,11 +385,220 @@ model UserBlockCredit {
...
+  isActive         Boolean         @default(true)
+  metadata         Json?
+  UserAccounting   UserAccounting? @relation(fields: [userAccountingId], references: [id])
+  userAccountingId String?
 
   @@id(name: "creditTransactionIdentifier", [transactionKey, userId])
 }
+
+// This is the top level accounting table for a user
+model UserAccounting {
```
Swiftyos: This is meant to be the table that is used to get all subscription, transaction and account  information about a user.

autogpt_platform/backend/schema.prisma
```diff
@@ -272,11 +385,220 @@ model UserBlockCredit {
...
+  transactions UserBlockCredit[]
+
+  usdBalance Float @default(0.0)
+
+  subscription   UserSubscription? @relation(fields: [subscriptionId], references: [id])
+  subscriptionId String?
+}
+
+// This table tracks a user's subscription to autogpt
+model UserSubscription {
```
Swiftyos: Need to add stripe ids here

autogpt_platform/backend/schema.prisma
```diff
@@ -272,11 +385,220 @@ model UserBlockCredit {
...
+  group          UserGroup? @relation(fields: [groupId], references: [id])
+
+  username    String @unique
+  description String
+
+  website  String?
+  github   String?
+  linkedin String?
+  facebook String?
+  twitter  String?
```
Swiftyos: x seems to short a name...

autogpt_platform/backend/schema.prisma
```diff
@@ -272,11 +385,220 @@ model UserBlockCredit {
...
+model StoreListing {
+  id        String   @id @default(uuid())
+  createdAt DateTime @default(now())
+  updatedAt DateTime @updatedAt
+
+  isDeleted  Boolean @default(false)
+  // Not needed but makes lookups faster
+  isApproved Boolean @default(false)
+
+  slug        String
```
Swiftyos: I'm planning on using `username/slug` or `groupname/slug` as the url so slug does not need to be unique

autogpt_platform/backend/schema.prisma
```diff
@@ -272,11 +385,220 @@ model UserBlockCredit {
...
+  updatedAt DateTime @updatedAt
+
+  // Owner can be found from the storeListing
+  storeListingId String
+  storeListing   StoreListing @relation(fields: [storeListingId], references: [id])
+
+  canPurchase   Boolean @default(true)
+  pruchasePrice Float?
+
+  commissionPerRun Float?
```
Swiftyos: This is a bit at odds with our block transaction accounting. We need to decide how this commission is applied. Is it only applied on a successful run?

Swiftyos:
When we come to an agreement on the schema changes I will propagate all these changes throughout the database

autogpt_platform/backend/schema.prisma
```diff
@@ -196,24 +302,65 @@ model AgentGraphExecutionSchedule {
...
 
-  agentGraphId      String
-  agentGraphVersion Int        @default(1)
-  AgentGraph        AgentGraph @relation(fields: [agentGraphId, agentGraphVersion], references: [id, version], onDelete: Cascade)
+  configuredAgentId String
+  configuredAgent   ConfiguredAgent @relation(fields: [configuredAgentId], references: [id], onDelete: Cascade)
 
   schedule  String // cron expression
   isEnabled Boolean @default(true)
-  inputData String // JSON serialized object
```
Swiftyos: inputData is now taken from the Configured Agent setup

autogpt_platform/backend/schema.prisma
```diff
@@ -272,11 +428,220 @@ model UserBlockCredit {
...
+
+  hasInitiatedChargeBackProcess Boolean          @default(false)
+  chargeBackProcessInitiatedAt  DateTime?
+  UserAccounting                UserAccounting[]
+
+  @@index([isSubscribed, isCancelling])
+}
+
+// This table tracks the different subscription plans available to a user
+model SubsctiptionPlan {
```
Swiftyos: @Torantulino please take a look at the accounting tables, as there are some assumptions baked in here about how it's going to work.

Thats this table and a few above it

autogpt_platform/backend/schema.prisma
```diff
@@ -19,45 +19,148 @@ model User {
...
+  agents       Agent[]
+  Profile      Profile[]
+  StoreListing StoreListing[]
+
+  @@index([name])
+}
+
 // This model describes the Agent Graph/Flow (Multi Agent System).
-model AgentGraph {
+model Agent {
```
Pwuts: I think `AgentGraph` is a better name for what it is. It's specific and descriptive of what it represents. It's also consistent with the rest of the naming in this schema.

If we rename this to `Agent`, we are saying this is what all agents in our system will look like into the future, and I think we're in too early a stage to assert that.

autogpt_platform/backend/schema.prisma
```diff
@@ -19,45 +19,148 @@ model User {
...
 
   @@id(name: "graphVersionId", [id, version])
+  @@map("AgentGraph")
+}
+
+//////////////////////////////////////////////////////////// 
+//////////////////////////////////////////////////////////// 
+////////////////   USER SPECIFIC DATA   ////////////////////
+//////////////////////////////////////////////////////////// 
+////////////////////////////////////////////////////////////
```
Pwuts: nit: this is excessive, 3 lines would be plenty and easier to read for me

autogpt_platform/backend/schema.prisma
```diff
@@ -19,45 +19,148 @@ model User {
...
   AgentGraphExecutionSchedules AgentGraphExecutionSchedule[]
   AnalyticsDetails             AnalyticsDetails[]
   AnalyticsMetrics             AnalyticsMetrics[]
   UserBlockCredit              UserBlockCredit[]
+  ConfiguredAgents             ConfiguredAgent[]
+  UserAgents                   UserAgents[]
+
+  // User Group relations
+  groups                 UserGroup[]              @relation("GroupUsers")
+  adminGroups            UserGroup[]              @relation("GroupAdminUsers")
```
Pwuts: what would be in this list?

autogpt_platform/backend/schema.prisma
```diff
@@ -19,45 +19,148 @@ model User {
...
+  name         String
+  description  String
+  groupIconUrl String?
+
+  users      User[] @relation("GroupUsers")
+  adminUsers User[] @relation("GroupAdminUsers")
+
+  agents       Agent[]
+  Profile      Profile[]
+  StoreListing StoreListing[]
```
Pwuts: Inconsistent field case (upper vs lower camel case). To be consistent with the convention so far, all of these should be upper camel case (PascalCase).

autogpt_platform/backend/schema.prisma
```diff
@@ -136,14 +239,14 @@ model AgentGraphExecution {
...
+  AgentGraph        Agent  @relation(fields: [agentGraphId, agentGraphVersion], references: [id, version], onDelete: Cascade)
 
   AgentNodeExecutions AgentNodeExecution[]
 
-  // Link to User model
-  userId String
-  user   User   @relation(fields: [userId], references: [id], onDelete: Cascade)
+  // This is so we can track which user executed the agent graph for accounting purposes.
+  executedByUserId String
+  executedByUser   User   @relation(fields: [executedByUserId], references: [id], onDelete: Cascade)
```
Pwuts: - As far as I know, this isn't just for accounting purposes. Also essential for access control.
- Can we add `executionTriggerType` that indicates `manual`, `schedule`, `webhook` + maybe `executionTriggerId` to tie it to the user/schedule/webhook that triggered it?

autogpt_platform/backend/schema.prisma
Swiftyos: The 2 different lists are not really intentional, more an artefact of the db structure. It is to differentiate the roles the user has in a group. Maybe it is better modelled with a GroupUserRole table linking the Users table and the Group table.

What do you think?

autogpt_platform/backend/schema.prisma
```diff
@@ -186,6 +289,9 @@ model AgentNodeExecutionInputOutput {
   referencedByOutputExecId String?
   ReferencedByOutputExec   AgentNodeExecution? @relation("AgentNodeExecutionOutput", fields: [referencedByOutputExecId], references: [id], onDelete: Cascade)
 
+  configuredAgentId String?
+  ConfiguredAgent   ConfiguredAgent? @relation("AgentNodeExecutionInput", fields: [configuredAgentId], references: [id])
```
Pwuts: ~~Once an `AgentNodeExecutionInputOutput` has been created, it is time-invariant, so it should be tied to the corresponding agent *version*. Otherwise we can't interpret the inputs/outputs properly as they are tied to specific nodes in a specific graph version.~~
wrote this before I understood how the `ConfiguredAgent.InputPresets` worked

autogpt_platform/backend/schema.prisma
```diff
@@ -196,24 +302,65 @@ model AgentGraphExecutionSchedule {
...
-  AgentGraph        AgentGraph @relation(fields: [agentGraphId, agentGraphVersion], references: [id, version], onDelete: Cascade)
+  configuredAgentId String
+  configuredAgent   ConfiguredAgent @relation(fields: [configuredAgentId], references: [id], onDelete: Cascade)
 
   schedule  String // cron expression
   isEnabled Boolean @default(true)
-  inputData String // JSON serialized object
+
+  // Allows triggers to be routed down different execution paths in an agent graph
+  triggerIdentifier String
```
Pwuts: How do you imagine this working?

autogpt_platform/backend/schema.prisma
```diff
@@ -19,45 +19,148 @@ model User {
...
   // Link to User model
-  userId String
-  user   User   @relation(fields: [userId], references: [id], onDelete: Cascade)
+  createdByUserId String
+  createdByUser   User   @relation(fields: [createdByUserId], references: [id], onDelete: Cascade)
 
-  AgentNodes                  AgentNode[]
-  AgentGraphExecution         AgentGraphExecution[]
-  AgentGraphExecutionSchedule AgentGraphExecutionSchedule[]
+  isGroupAgent Boolean    @default(false)
```
Pwuts: This seems redundant with `groupId`: if `groupId` is set, it's a group-owned agent.

autogpt_platform/backend/schema.prisma
Pwuts: Nice, this is a better construction than `isActive`.

Nit: I would've named it `AgentGraphConfiguration` to stay consistent with the rest of the schema.

autogpt_platform/backend/schema.prisma
Swiftyos: This is my thinking around this field. 

Let say I have an agent that process emails. I've set it up so it is triggered on a webhook from outlook and a web hook from gmail.

Both of these services send a slightly different message format.

When creating the agent, set the triggerIdentifier to outlook or gmail for the respective web hooks.
When the web hook is called, the triggering input data would be:
```json
{
   "triggerIdentifier": "outlook",
   "data": webhook data
}
```

Then when the agent is triggered, I check the triggerIdentifier key, and if it is gmail send it to the gmail processor if it is outlook to the outlook processor.

autogpt_platform/backend/schema.prisma
```diff
@@ -19,45 +19,148 @@ model User {
...
+  isActive Boolean @default(true)
+
+  userId String
+  user   User   @relation(fields: [userId], references: [id])
+
+  agentId      String
+  agentVersion Int
+  agent        Agent  @relation(fields: [agentId, agentVersion], references: [id, version])
+
+  Configuration               AgentNodeExecutionInputOutput[] @relation("AgentNodeExecutionInput")
```
Pwuts:
```suggestion
  InputPresets                AgentNodeExecutionInputOutput[] @relation("AgentNodeExecutionInput")
```

autogpt_platform/backend/schema.prisma
Pwuts: Also, this references `@relation("AgentNodeExecutionInput"` which I don't think is correct (copy-pasted?)

autogpt_platform/backend/schema.prisma
Pwuts: Also, (same comment as below) this references `@relation("AgentNodeExecutionInput"` which I don't think is correct.

autogpt_platform/backend/schema.prisma
```diff
@@ -196,24 +302,65 @@ model AgentGraphExecutionSchedule {
...
+
+  isActive           Boolean   @default(true)
+  lastReceivedDataAt DateTime?
+  isDeleted          Boolean   @default(false)
+  Agent              Agent?    @relation(fields: [agentId, agentVersion], references: [id, version])
+  agentId            String?
+  agentVersion       Int?
+
+  @@index([configuredAgentId])
+}
```
Pwuts: This doesn't match the implementation I'm working on atm so will need updating after that merges.

autogpt_platform/backend/schema.prisma
Pwuts: also, for consistency:
```suggestion
  Groups                 UserGroup[]              @relation("GroupUsers")
  AdminGroups            UserGroup[]              @relation("GroupAdminUsers")
```

Pwuts: The group already has a list specifying which users are admins, so whether we need this distinction between `groups` and `adminGroups` here depends on when/where/how we need access to this information.

I'd replace these by a relationship table `UserGroupMembership` which also specifies the user's role in the group.

autogpt_platform/backend/schema.prisma
Pwuts: If an execution is triggered by a specific node (e.g. a webhook trigger), we can add a reference to that node rather than using an arbitrary identifier like that.

In any case, I'm going to have to implement something like this in #8358 so I'll see what works there.

autogpt_platform/backend/schema.prisma
Swiftyos: I agree with replacing these relationships to `UserGroupMembership`

---

> The group already has a list specifying which users are admins, so whether we need this distinction between groups and adminGroups here depends on when/where/how we need access to this information.

These are the other end of the many-to-many relationship so `UserGroup. adminUsers` maps to `User. adminGroups`.

autogpt_platform/backend/schema.prisma
Swiftyos: Yes it was, thanks for pointing it out. I'll check what the correct way to handle that is 👍

autogpt_platform/backend/schema.prisma
Swiftyos: I'll bring in your schema changes when they are merged. Thanks

autogpt_platform/backend/schema.prisma
Swiftyos: np, will do

autogpt_platform/backend/schema.prisma
Swiftyos: My thinking here is we want to be using the same name convention end-to-end. Atm we refer to them as agents to the users, agentFlows / graphs in the frontend, then `AgentGraph`

We should get some other options on this too. Thanks for sharing your thoughts

autogpt_platform/backend/schema.prisma
Swiftyos: Yes absolutely, great idea. I'll update the comment too

autogpt_platform/backend/schema.prisma
```diff
@@ -19,45 +19,148 @@ model User {
...
+  Profile                Profile[]
+  StoreListing           StoreListing[]
+  StoreListingSubmission StoreListingSubmission[]
 
   @@index([id])
   @@index([email])
 }
 
+model UserGroup {
+  id        String   @id @default(uuid())
```
ntindle: This feels weirdly designed at first read. Can a group contain a group? Why can a user be a member of multiple groups? Do I bill by the group? Why does a group need a description? Is this for allowing multiple people to mange the same store posting?

can you walk me through the choices for groups design?

ntindle: Should we have group and group members as separate tables so we can have a clearer coupling? Prisma I think is doing this behind the scenes with it’s linking tables but we may want it explicit so we can have more than two roles down the line

group consists of stuff like name, description etc

Group members is a simple table with
User id, group id, role - admin, member (later billing manager etc)

autogpt_platform/backend/schema.prisma
```diff
@@ -19,45 +19,148 @@ model User {
...
   name        String?
   description String?
-  isActive    Boolean @default(true)
-  isTemplate  Boolean @default(false)
 
   // Link to User model
-  userId String
-  user   User   @relation(fields: [userId], references: [id], onDelete: Cascade)
+  createdByUserId String
+  createdByUser   User   @relation(fields: [createdByUserId], references: [id], onDelete: Cascade)
```
ntindle: If a user creates the agent in a group and then they delete their account, the agent will be deleted even if the group persists, correct?

ntindle: And by extension this would delete it from the marketplace

autogpt_platform/backend/schema.prisma
ntindle: It also may make more sense to have everything owned by a group by default with only one member in it, the user. I think that could streamline a lot of the logic in the rest of the tables

autogpt_platform/backend/schema.prisma
```diff
@@ -272,11 +428,220 @@ model UserBlockCredit {
...
+}
+
+model StoreListingVersion {
+  id        String   @id @default(uuid())
+  createdAt DateTime @default(now())
+  updatedAt DateTime @updatedAt
+
+  agentId      String
+  agentVersion Int
+  agent        Agent  @relation(fields: [agentId, agentVersion], references: [id, version])
```
ntindle: Agents are immutable after saving right? One of the reasons the marketplace doesn’t reference the other schemas right now is as a dirty way to prevent users from updating agents without approval

autogpt_platform/backend/schema.prisma
Pwuts: `agentFlow` should definitely be renamed, that's legacy naming left over from my first draft of those parts of the frontend.

Referring to them as "Agents" to the users makes sense (at least right now) *and would be easy to change*. Renaming everything in the code is a lot more involved, and so I'd prefer to stick to a more technical/descriptive approach to naming things, as that's a more future proof and stable strategy than basing technical naming schemes on design/marketing decisions.

autogpt_platform/backend/schema.prisma
ntindle: Why is this different than getting the agents that match the collection of groups the user is in?

autogpt_platform/backend/schema.prisma
Pwuts: https://github.com/Significant-Gravitas/AutoGPT/pull/8375#discussion_r1806290774

autogpt_platform/backend/schema.prisma
ntindle: Maybe it’s better to have just a list of platform links and an enum (if not string and handle in Python) for keys rather than row by row if we want to to add like mastadon or threads or printable or whatever

autogpt_platform/backend/schema.prisma
ntindle: Thanks, didn’t show as I was reviewing bc outdated

autogpt_platform/backend/schema.prisma
Swiftyos: Done

autogpt_platform/backend/schema.prisma
Swiftyos: I agree with your comments @ntindle 
It is something that @Pwuts pointed out as well. 
This change has been made now

autogpt_platform/backend/schema.prisma
Swiftyos: The plan is to make it so it is not possible to delete an entry in the database, they are just marked as delete. Then we run our own cleanup logic periodically.

When the agent becomes a group agent, someone in the group with the correct permissions needs to delete it.

The user can delete the agent out of there library, which is stored in the UserAgents table

autogpt_platform/backend/schema.prisma
Swiftyos: This is to manage what the User sees in there library. It can include agents created by other people, recently executed agents. Anything we want really.

autogpt_platform/backend/schema.prisma
Swiftyos: Yeah probably. a better idea, I'll change it to a list of url strings. 

We can determine the link time from the domain name, so guess doing more than just checking its a valid domain is not needed

autogpt_platform/backend/schema.prisma
Swiftyos: Agents are immutable after saving.

autogpt_platform/backend/schema.prisma
Swiftyos: @aarushik93 @majdyz @ntindle your thoughts on this?

autogpt_platform/backend/schema.prisma
```diff
@@ -19,45 +19,164 @@ model User {
...
+
+model UserGroupMembership {
+  id        String   @id @default(uuid())
+  createdAt DateTime @default(now())
+  updatedAt DateTime @updatedAt
+
+  userId      String
+  user        User          @relation(fields: [userId], references: [id])
+  userGroupId String
+  userGroup   UserGroup     @relation(fields: [userGroupId], references: [id])
```
Pwuts: relations should be PascalCase, also in the rest of the file

autogpt_platform/backend/schema.prisma
```diff
@@ -272,11 +452,233 @@ model UserBlockCredit {
...
+  eventId     String    @unique
+  eventType   String
+  createdAt   DateTime  @default(now())
+  processedAt DateTime?
+}
+
+//////////////////////////////////////////////////////////// 
+//////////////////////////////////////////////////////////// 
+//////////////   Store TABLES    /////////////////////
+////////////////////////////////////////////////////////////
```
Pwuts: nit: inconsistent line length

autogpt_platform/backend/schema.prisma
```diff
@@ -257,9 +428,18 @@ model AnalyticsMetrics {
...
+
 enum UserBlockCreditType {
   TOP_UP
   USAGE
+  COMMISSION
+  PURCHASE
+  SALE
 }
 
 model UserBlockCredit {
```
Pwuts: can this be renamed to be more clear? 
- `UserBlockCreditMutation`
- `UserBlockCreditTransaction`
- `UserCreditMutation`
- `UserCreditTransaction`

autogpt_platform/backend/schema.prisma
```diff
@@ -196,24 +326,65 @@ model AgentGraphExecutionSchedule {
...
+  triggerIdentifier String
 
   // default and set the value on each update, lastUpdated field has no time zone.
   lastUpdated DateTime @updatedAt
 
   // Link to User model
-  userId String
-  user   User   @relation(fields: [userId], references: [id], onDelete: Cascade)
+  userId       String
+  user         User    @relation(fields: [userId], references: [id], onDelete: Cascade)
```
Pwuts: case

autogpt_platform/backend/schema.prisma
```diff
@@ -19,45 +19,164 @@ model User {
...
-  userId String
-  user   User   @relation(fields: [userId], references: [id], onDelete: Cascade)
+  createdByUserId String
+  createdByUser   User   @relation(fields: [createdByUserId], references: [id], onDelete: Cascade)
 
-  AgentNodes                  AgentNode[]
-  AgentGraphExecution         AgentGraphExecution[]
-  AgentGraphExecutionSchedule AgentGraphExecutionSchedule[]
+  groupId String?
+  group   UserGroup? @relation(fields: [groupId], references: [id])
```
Pwuts: case

autogpt_platform/backend/schema.prisma
Pwuts: how was this resolved? cases in the schema are still inconsistent

autogpt_platform/backend/schema.prisma
Torantulino: > If we rename this to `Agent`, we are saying this is what all agents in our system will look like into the future, and I think we're in too early a stage to assert that.

I suppose you're alluding to something along the lines of an Agent Block?

Personally I think being consistent with our communications around what Agents on the AutoGPT Platform are is going to make things more rather than less clear. It'll simplify technical doccumentation and align it with the other content we publish.

autogpt_platform/backend/schema.prisma
Torantulino: @Swiftyos Is this implying that people would earn platform **credits** from the marketplace? Whilst this is a possibility, that's not currently the option that's planned.

autogpt_platform/backend/schema.prisma
Torantulino: Let's chat about this one to make sure we're on the same page

autogpt_platform/backend/schema.prisma
Swiftyos: I'm waiting for @majdyz to weight in on this part before making more changes

autogpt_platform/backend/schema.prisma
Swiftyos: Could you be more specific please.

In this case you referring to just User not userId?

autogpt_platform/backend/schema.prisma
Swiftyos: I'd had looked through and changed the ones I spotted, missed a few by the looks of it.

autogpt_platform/backend/schema.prisma
aarushik93: I think Agent is okay as well, even from a technical point of view, AgentGraph needs clarification to someone who's not familiar with the codebase. I think calling it Agent is consistent, as well almost as simple as being like "In the AutoGPT system this is an Agent" even if it doesnt technically align with other current definitions of agents (even those change and have been changing since the 90s lol)

autogpt_platform/backend/schema.prisma
```diff
@@ -19,45 +19,164 @@ model User {
...
   AnalyticsMetrics             AnalyticsMetrics[]
   UserBlockCredit              UserBlockCredit[]
+  ConfiguredAgents             AgentPreset[]
+  UserAgents                   UserAgents[]
+
+  // User Group relations
+  UserGroupMemberships   UserGroupMembership[]
+  UserAccounting         UserAccounting[]
+  UserSubscription       UserSubscription[]
+  Profile                Profile[]
```
aarushik93: Missing on deletes? As well as in the rest of the schema, where appropriate

autogpt_platform/backend/schema.prisma
```diff
@@ -19,45 +19,164 @@ model User {
...
+
+  InputPresets                AgentNodeExecutionInputOutput[] @relation("AgentNodeExecutionPresets")
+  UserAgents                  UserAgents[]
+  WebhookTrigger              WebhookTrigger[]
+  AgentGraphExecutionSchedule AgentGraphExecutionSchedule[]
+
+  @@index([userId])
+}
+
+model UserAgents {
```
kcze: Better would be `UserAgent` if this type represents a single agent.

autogpt_platform/backend/schema.prisma
```diff
@@ -272,11 +452,233 @@ model UserBlockCredit {
...
+}
+
+model StoreListing {
+  id        String   @id @default(uuid())
+  createdAt DateTime @default(now())
+  updatedAt DateTime @updatedAt
+
+  isDeleted  Boolean @default(false)
+  // Not needed but makes lookups faster
+  isApproved Boolean @default(false)
```
kcze: If the information about approval is also somewhere else that creates multiple sources of truth and potentially needless bugs when that information becomes desynchronised (can it?). If that's the case is this optimization worth the risk? The same for `StoreListingVersion`.

autogpt_platform/backend/schema.prisma
```diff
@@ -19,45 +19,164 @@ model User {
   updatedAt DateTime @updatedAt
 
   // Relations
-  AgentGraphs                  AgentGraph[]
+  AgentGraphs                  Agent[]
   AgentGraphExecutions         AgentGraphExecution[]
   AgentGraphExecutionSchedules AgentGraphExecutionSchedule[]
```
kcze: Similarly to what [Pwuts mentioned](https://github.com/Significant-Gravitas/AutoGPT/pull/8375/files#r1806232014), change from `AgentGraph` to `Agent` makes this schema inconsistent; why isn't `AgentGraphExecutions*`, `AgentGraphVersion`, etc. not renamed?

autogpt_platform/backend/schema.prisma
kcze: `Agent` is more generic and thus I would argue requires even more explanation because it means more things. Now schema is inconsistent because there's one `Agent` and a lot of `AgentGraph*` fields. I prefer naming to be consistent so everything could be renamed but if I had to sacrifice consistency I'd keep `AgentGraph` in schema and user-facing `Agent` naming.

autogpt_platform/backend/schema.prisma
majdyz: The convention is relation field use capital. e.g: User.

autogpt_platform/backend/schema.prisma
majdyz: We definitely could! Is the scope of this PR DB renamings though? Can we simply start with the DB changes while retaining the name for now? Because it's going to be complicated changes already. You also need to propagate the code changes. 

I don't think others would be able to review it properly if the changes are humongous.

autogpt_platform/backend/schema.prisma
Torantulino: Yeah I agree that Ideally we make naming consistent. 
Would it be a lot of effort to replace mentions of `AgentGraph` from throughout the codebase?

autogpt_platform/backend/schema.prisma
```diff
@@ -19,45 +19,164 @@ model User {
...
+  agentId      String
+  agentVersion Int
+  agent        Agent  @relation(fields: [agentId, agentVersion], references: [id, version])
+
+  configuredAgentId String?
+  configuredAgent   AgentPreset? @relation(fields: [configuredAgentId], references: [id])
+
+  isFavorite      Boolean @default(false)
+  isCreatedByUser Boolean @default(false)
+  isPublished     Boolean @default(false)
```
Swiftyos: isPublished needs to be removed

autogpt_platform/backend/schema.prisma
```diff
@@ -19,45 +19,164 @@ model User {
...
+  isPublic        Boolean @default(false)
```
Swiftyos: isPublic is not needed either

autogpt_platform/backend/schema.prisma
```diff
@@ -19,45 +19,164 @@ model User {
...
+  @@map("AgentGraph")
 }
 
+//////////////////////////////////////////////////////////// 
+//////////////////////////////////////////////////////////// 
+////////////////   USER SPECIFIC DATA   ////////////////////
+//////////////////////////////////////////////////////////// 
+//////////////////////////////////////////////////////////// 
+
+model AgentPreset {
```
Swiftyos: Need to make sure AgentPresets are associated with there executions

autogpt_platform/backend/schema.prisma
```diff
@@ -272,11 +452,233 @@ model UserBlockCredit {
...
+  isApproved             Boolean                  @default(false)
+  StoreListing           StoreListing?            @relation(fields: [storeListingId], references: [id])
+  storeListingId         String?
+  StoreListingSubmission StoreListingSubmission[]
+
+  @@index([agentId, agentVersion, isApproved])
+}
+
+enum SubmissionStatus {
+  PENDING
```
Swiftyos: add DRAFT

autogpt_platform/backend/schema.prisma
```diff
@@ -272,11 +452,233 @@ model UserBlockCredit {
   blockId String?
   block   AgentBlock? @relation(fields: [blockId], references: [id])
 
+  // We need to be able to associate a credit transaction with an agent
```
Swiftyos: There is redundancy in having both executedAgent and storeListing here. Need to check the optimisation is needed

autogpt_platform/backend/schema.prisma
```diff
@@ -272,11 +452,233 @@ model UserBlockCredit {
...
+  // We need to be able to associate a credit transaction with a store listing
+  storeListingId String?
+  storeListing   StoreListing? @relation(fields: [storeListingId], references: [id])
+
   amount Int
   type   UserBlockCreditType
 
-  isActive Boolean @default(true)
-  metadata Json?
+  isActive         Boolean         @default(true)
```
Swiftyos: @majdyz What does isActive do here?

autogpt_platform/backend/schema.prisma
```diff
@@ -272,11 +452,233 @@ model UserBlockCredit {
   blockId String?
   block   AgentBlock? @relation(fields: [blockId], references: [id])
```
Swiftyos: We need to associate a cost to the `AgentNodeExecution`

autogpt_platform/backend/schema.prisma
Swiftyos: Updated

autogpt_platform/backend/schema.prisma
Swiftyos: Hopefully got them all now

autogpt_platform/backend/schema.prisma
Swiftyos: Removed add reference to Graph, renaming all fields

autogpt_platform/backend/schema.prisma
Swiftyos: I hope I've done it right now: I have userId and User

autogpt_platform/backend/schema.prisma
Swiftyos: I've removed all the accounting stuff in this pr

autogpt_platform/backend/schema.prisma
Swiftyos: We don't want to over normalise the database here. The key is that there is only 1 source of truth, so if there ever is a disconnect between 2 locations, then we know which is correct.

In this case isApproved is based on the `StoreListingSubmission` how as isApproved is going to be checked for the majority of  queries on this table it makes sense to be able to index the table on it, rather than needing to do a join for all these queries.

autogpt_platform/backend/schema.prisma
Swiftyos: I'm always a bit nervous about adding cascading deletes. I feel one should be very intentional about what is being deleted. To prevent accidental data loss.

aarushik93: Valid concern, I would say though that if we're representing a parent child relationship with true composition then when a parent is deleted, an orphaned child becomes meaningless

I think we can aleviate your concerns though by making it really clear in the docs, and foreign key constraints.

And if we're super worried about data loss, enabling backups or soft deletes is an option

aarushik93: Correction, I just checked and we already have back ups enabled on supabase by default 😄

Swiftyos: Okay thanks for that, I'll add in the cascading deletes

Swiftyos: I've updated the relationships that I think make sense to have a cascading delete. Let me know what you think please

autogpt_platform/backend/schema.prisma
```diff
@@ -20,55 +20,174 @@ model User {
...
+  @@id(name: "agentVersionId", [id, version])
+}
+
+//////////////////////////////////////////////////////////// 
+//////////////////////////////////////////////////////////// 
+////////////////   USER SPECIFIC DATA   ////////////////////
+//////////////////////////////////////////////////////////// 
+//////////////////////////////////////////////////////////// 
+
+model AgentPreset {
```
ntindle: I think documenting what each of these tables is supposed to be for would be helpful long term

// Agent Presets is used to store????????? about an agent and is used

autogpt_platform/backend/schema.prisma
```diff
@@ -20,55 +20,174 @@ model User {
   metadata  Json?
```
ntindle: probably should make this default to {}

autogpt_platform/backend/schema.prisma
```diff
@@ -127,34 +246,47 @@ enum AgentExecutionStatus {
...
+  AgentPreset   AgentPreset? @relation(fields: [agentPresetId], references: [id])
 
   AgentNodeExecutions AgentNodeExecution[]
 
-  // Link to User model
-  userId String
-  user   User   @relation(fields: [userId], references: [id], onDelete: Cascade)
+  // This is so we can track which user executed the agent.
+  executedByUserId String
+  ExecutedByUser   User   @relation(fields: [executedByUserId], references: [id], onDelete: Cascade)
```
ntindle: is there anything that could trigger an agent that this wouldn't make sense for? 

webhook? another agent?

Swiftyos: It is needed for to determine who it was this agent ran for, think as we know who is using what resources, for billing or even just monitoring the system. Even a webhook triggered agent or one triggered by another agent, would need an executedByUser value. In these cases, the user that setup the thing that triggered it.

ntindle (approved):

PR #8384, opened 2024-10-21
feat(platform/ci) Add workflow for deploying to production

aarushik93:
### Background

Based on our new release process, we want to deploy to production after creating a release but with a manual approval

### Changes 🏗️

Added environments to the job which is configured to require approval from a maintainer

- Upload agent to marketplace


Swiftyos (dismissed):
Nice idea linking prod deployments to GitHub tags. Makes the process very transparent

PR #8387, opened 2024-10-21
feat(backend): RedisEventQueue into Pub/Sub

majdyz:
### Background

There are two issues with the current Redis event queue implementation:
1. It's a queue, which means there can only be 1 event subscriber.
2. It's implemented in a suboptimal way on ws_api:
    - The subscription loop lives in a fast API task (which is supposed to be async).
    - The task is implemented with a continuous sync loop with sleep.

### Changes 🏗️

* Converted RedisEventQueue --> RedisEventBus, and provided two versions, the async & sync version.
* The WebSocket service / fast API will use the async version for subscribing to the event update.
* The ExecutionManager will use the sync version for publishing the events.

- Upload agent to marketplace


autogpt_platform/backend/backend/data/queue.py
```diff
@@ -16,35 +18,74 @@ def default(self, o):
         return super().default(o)
 
 
-class AbstractEventQueue(ABC):
-    @abstractmethod
-    def put(self, execution_result: ExecutionResult):
-        pass
+class BaseRedisEventBus:
```
Pwuts: could you make this `Generic[T]` or rename it to reflect what kind of events it handles?

Pwuts: I did the same here for the `RedisEventQueue`: https://github.com/Significant-Gravitas/AutoGPT/pull/8358/commits/a82acc77880962a0842a774707622d80cbe87e0f

majdyz: Updated. Are you using this event queue using different type by any chance on your PR?

ntindle (approved):

PR #8391, opened 2024-10-22
fix: show error toast on Run failure when inputs or credentials are i…

### LINKED ISSUE #8332: Show error if queueing graph execution fails
If clicking "Run" results in a failure, e.g. the hard-coded input values are no longer valid, this isn't visible in the UI. This can happen e.g. with credentials when they are deleted from the user but not from the graph. The execution request returns an HTTP 400 but the UI doesn't show anything.

Suggestion: show a toast (`variant="destructive"`) with the error message.

Abhi1992002:
## Background

The issue is that when “Run” fails due to invalid inputs (e.g., deleted credentials), no error message is shown in the UI.

### Changes 🏗️

- Added toast notification (variant="destructive") for "Run" failures.
- Displays error when the execution request returns an HTTP 400 due to invalid hard-coded input values 

https://github.com/user-attachments/assets/e5686fcf-e376-40ca-876a-8cb99aec2e2e


autogpt_platform/supabase
ntindle, line 1: if we revert the change to supabase, this is good to go!

ntindle (changes requested):
revert supabase change and this is ready after

Abhi1992002:
Done..!!

aarushik93:
@Abhi1992002 can you fix the tests please and then we can get this in

aarushik93 (approved):

ntindle (approved):

PR #8396, opened 2024-10-22
feat(platform/ci): Add prod migrations

aarushik93:
### Background

So prod migrations run automatically

### Changes 🏗️

Added migrate job

- Upload agent to marketplace


.github/workflows/platform-autgpt-deploy-prod.yml
```diff
@@ -15,6 +15,49 @@ env:
...
+        env:
+          DATABASE_URL: ${{ secrets.BACKEND_DATABASE_URL }}
+
+      - name: Run Market Migrations
+        if: steps.check_changes.outputs.market_changed == 'true'
+        working-directory: ./autogpt_platform/market
+        run: |
+          python -m prisma migrate deploy
+        env:
+          DATABASE_URL: ${{ secrets.MARKET_DATABASE_URL }}
```
ntindle: is there a reason we wouldn't always run migrations?

ntindle (dismissed):

.github/workflows/platform-autgpt-deploy-prod.yml
aarushik93: when there are no changes in the migrations?

ntindle: Doesn’t the tool determine that?

ntindle: If I'm crazy feel free to resolve btw

aarushik93: fair i was thinking we just don't do this step at all thats why. but ill update!

PR #8398, opened 2024-10-22
fix: hide content except login when not authenticated to prevent errors

### LINKED ISSUE #8112: Hide Monitor, Build, and possible Marketplace on a non-authenticated user & server auth enabled
If we are not yet logged-in, and the server enabled the user auth, we should hide everything except login box (maybe exclude marketplace)? Since clicking anything causes an error.

Abhi1992002:
### Fixes #8112 

I did this to stop errors when users aren't logged in and try to access restricted content. If they go to another page, they'll be redirected to "/login" for a smooth experience.

### Changes 🏗️

- I hid the navbar if the user is not logged in.
- If the user visits a page without being logged in, they will be redirected to "/login".

https://github.com/user-attachments/assets/ec625880-9fbf-4140-a8d0-9b47c2eed1ae


autogpt_platform/supabase
ntindle, line 1: revert this change

 cc @aarushik93 thoughts on marking this as code owned by devops?

ntindle (changes requested):
Revert the supabase change

Abhi1992002:
@ntindle I've removed the `supabase` folder as requested.

ntindle:
Sorry for the lack of clarity on my request. I meant remove your changes to it so its unaltered rather than removed

Abhi1992002:
Sorry for the mix-up—I’ve undone my changes and restored the original document.

.gitignore
```diff
@@ -165,9 +165,12 @@ package-lock.json
...
-pri* 
+pri*
 # ignore
 ig*
 .github_access_token
 LICENSE.rtf
 autogpt_platform/backend/settings.py
+
+# Ignoring the supabase folder from pushing on github
+supabase/
```
ntindle: @aarushik93 is this a good idea? we can force push over when desired after all

ntindle (dismissed):
only q for @aarushik93

.gitignore
ntindle: (merge if its having this in gitignore is okay)

aarushik93: @Abhi1992002 can we not have this change please? If you could just not add your changes that should be enough

aarushik93 (approved):
sorry for the delay in review, this looks good, let's gooo

aarushik93 (approved):

PR #8403, opened 2024-10-23
feat(backend): credential swap to api keys in cred store

ntindle:
### Background

Many of our blocks use api keys to authenticate with various services. We want those to no longer be stored in the graph on a block level, but as a reference to a credential

### Changes 🏗️

//Backend
- Swaps many blocks, but not all (LLM blocks mostly due to needing to select from many valid credentials of different providers), to the new format of credentials
- Drops discord continuous mode as a side effect, approved by Toran
- uses the store to load default set credentials for API keys injected by the system
- Uses the credit system to detect and charge for injected API keys

// Frontend
- Updates the way we handle providers in the front end to deduplicate the keys for providers we have

- Upload agent to marketplace


autogpt_platform/backend/backend/data/credit.py
```diff
@@ -49,36 +61,128 @@ def __init__(
...
+            cost_amount=10,
+            cost_filter={
+                "credentials": {
+                    "id": replicate_credentials.id,
+                    "provider": replicate_credentials.provider,
+                    "type": replicate_credentials.type,
+                }
+            },
+        )
+    ],
```
ntindle, line 185: @Torantulino we need cost #s on these

aarushik93 (approved):

lotuswerx:
You guys do amazing work I want to look to see when I started using AutoGPT but it was pretty early on and I've been donating to the @Torantulino monthly since. When this thing comes live (I did join the waitlist) but I will relearn making Agents here. This new AI Agent phase of the sector is going to lead straight to you because I started a little Marketing campaign to drive some pretty savvy to AutoGPT, but I believe coding (programming what I used to call it) will scare them off so now I am going to drive them to the Sign Up page. You have a wonderful WebDesigner as well, hat tip. Anyway, I just saw the "leave a comment" button so I clicked it. 
Thank you, 
Dana G
www.lotuswerx.com

lotuswerx:
Oops, my bad😁

lotuswerx:
Oops, my bad😁

autogpt_platform/autogpt_libs/autogpt_libs/supabase_integration_credentials_store/store.py
```diff
@@ -42,9 +110,25 @@ def add_creds(self, user_id: str, credentials: Credentials) -> None:
...
+            all_credentials.append(groq_credentials)
+        if settings.secrets.replicate_api_key:
+            all_credentials.append(replicate_credentials)
+        if settings.secrets.openai_api_key:
+            all_credentials.append(openai_credentials)
+        if settings.secrets.anthropic_api_key:
+            all_credentials.append(anthropic_credentials)
+        if settings.secrets.did_api_key:
+            all_credentials.append(did_credentials)
+        return all_credentials
```
Pwuts, line 127: I'll try to find a better place for this

autogpt_platform/backend/backend/blocks/google/gmail.py
Pwuts, line 1: Was this passing CI before?

autogpt_platform/backend/backend/data/user.py
Pwuts, line 1: Since storing the user's credentials is the only thing we're using the `metadata` field for, we should just make it a separate column `integration_credentials` imo. Using metadata to store secret stuff is fishy anyway.

autogpt_platform/backend/migrations/20241030063332_drop_all_credentials_from_constant_input/migration.sql
Pwuts, line 1: This is missing `smtp_username`, `smtp_password`, `client_id`, `client_secret`, `username`, `password`

autogpt_platform/backend/pyproject.toml
```diff
@@ -46,6 +46,7 @@ youtube-transcript-api = "^0.6.2"
 googlemaps = "^4.10.0"
 replicate = "^0.34.1"
 pinecone = "^5.3.1"
+cryptography = "^43.0.3"
```
Pwuts, line 49: nit: inserted in the wrong place

autogpt_platform/backend/migrations/20241030063332_drop_all_credentials_from_constant_input/migration.sql
Pwuts: The blocks with those inputs also need to be ported to the credentials system

autogpt_platform/backend/backend/util/settings.py
```diff
@@ -208,6 +208,8 @@ class Secrets(UpdateTrackingModel["Secrets"], BaseSettings):
         default="", description="Supabase service role key"
     )
 
+    encryption_key: str = Field(default="", description="Encryption key")
```
Pwuts, line 211: - Add to `.env.example` with advice on how to generate a secure key
- Does an empty string work as an encryption key? Should the application fail to start if this isn't set?

autogpt_platform/backend/migrations/20241030061705_encrypt_user_metadata/migration.sql
Pwuts, line 1: Make a column `integration_credentials` instead of `metadata`, since we aren't using `metadata` for anything else and putting secret stuff in `metadata` is fishy.

Then migrate `auth.user.raw_user_meta_data["integration_credentials"]` to `platform.User.integration_credentials`.

autogpt_platform/backend/migrations/20241030063332_drop_all_credentials_from_constant_input/migration.sql
Pwuts: ~~Can we migrate these to `User.integration_credentials` instead of simply deleting? I think it's possible and want to try as a stretch goal so we don't break every single user's graphs.~~ not worth it

~~We definitely don't need to delete `credentials` as that isn't a `SecretField` and doesn't contain any credentials in itself.~~ fixed

autogpt_platform/backend/migrations/20241030061705_encrypt_user_metadata/migration.sql
Pwuts: Migrating `integration_credentials` from `auth.user` metadata to `platform.User.metadata`: https://github.com/Significant-Gravitas/AutoGPT/pull/8403/commits/befcef5e3152635d7dc3a8c6f69e71f5469d13df

Pwuts: done

autogpt_platform/backend/backend/util/settings.py
Pwuts: Fernet is quite finnicky about this: `ValueError: Fernet key must be 32 url-safe base64-encoded bytes.`

aarushik93 (dismissed):

autogpt_platform/backend/backend/util/settings.py
ntindle: added a default

aarushik93 (approved):

PR #8404, opened 2024-10-23
fix(backend): Fix DatabaseManager usage by calling it on-demand

majdyz:
### Background

We unintentionally cached the `DatabaseManager` connection object and shared it across different threads in `SupabaseIntegrationCredentialsStore`

### Changes 🏗️

* Make `DatabaseManager` connection acquisition on the fly and cache it within the class.
* Remove the need to provide a port number for connecting to pyro service (inferred from the class).
* Moved utils/cache.py to autogptlib.

- Upload agent to marketplace


ntindle (approved):

autogpt_platform/backend/backend/util/service.py
```diff
@@ -99,16 +100,24 @@ def custom_dict_to_class(qualname, data: dict):
...
         self.uri = None
 
+    @classmethod
+    @abstractmethod
+    def get_port(cls) -> int:
+        pass
+
+    @classmethod
+    def get_host(cls) -> str:
+        return os.environ.get(f"{cls.service_name.upper()}_HOST", Config().pyro_host)
```
ntindle, line 119: Is there a reason settings can’t be used here

majdyz: the pyro_host uses Settings,
For the dynamic config, idk, it's dynamic after all. I think @aarushik93 introduced this for prod purpose, e.g: dynamically separating host between services.

PR #8405, opened 2024-10-23
fix(platform): Enable auth on local-backend mode by default

majdyz:
### Background

The app with ENABLE_AUTH=False without login is pretty much broken now.

### Changes 🏗️

* Makes the ENABLE_AUTH env variable default for local development to True:
* Introduced `docker compose --profile local up` to bring up dependency services for ENABLE_AUTH, except the AutoGPT apps.
* Fixed broken HTTP rerouting on failed auth.

- Upload agent to marketplace


autogpt_platform/docker-compose.yml
```diff
@@ -142,3 +142,23 @@ services:
     extends:
       file: ./supabase/docker/docker-compose.yml
       service: vector
+
+  local_dependencies:
+    <<: *supabase-services
+    profiles:
+      - local
+    image: busybox
+    command: /bin/true
```
ntindle, line 151: This is big brain

autogpt_platform/backend/.env.example
```diff
@@ -20,13 +20,13 @@ PYRO_HOST=localhost
 SENTRY_DSN=
 
 ## User auth with Supabase is required for any of the 3rd party integrations with auth to work.
-ENABLE_AUTH=false
+ENABLE_AUTH=true
```
ntindle, line 23: I ageee with this change. I think we should take it one step further and just make it the only option

majdyz: I guess the local test might still need this, I don't think the integration test should import these humongous containers just for auth

autogpt_platform/backend/test/__init__.py
```diff
@@ -0,0 +1,3 @@
+import os
+
+os.environ["ENABLE_AUTH"] = "false"
```
ntindle, line 3: Probably needs to be a bool

autogpt_platform/backend/.env.example
ntindle: We do it for a few other tests on frontend unfortunately. May as well commit to the bit since all of them will have to pass anyway  not going to block on this point but worth considering as it’s a goal of the Automated Testing project

autogpt_platform/backend/README.md
```diff
@@ -58,17 +58,18 @@ We use the Poetry to manage the dependencies. To set up the project, follow thes
...
    ```
 
 ## Running The Server
 
 ### Starting the server without Docker
 
-Run the following command to build the dockerfiles:
+Run the following command to run database in docker but the application locally:
 
 ```sh
```
ntindle, line 71: Do you need the profile local if you do deps? Just checking, as I’m not super familiar

ntindle (dismissed):

autogpt_platform/backend/test/__init__.py
majdyz: env variable is always string

autogpt_platform/backend/.env.example
majdyz: I think there is already a ticket for this somewhere. But yes I agree.

autogpt_platform/backend/README.md
majdyz: The profile is added only to prevent the normal `docker compose up` to execute this `deps` service. Although `deps` pretty much does nothing, I just don't want this affecting the prod at all.

ntindle (approved):

PR #8406, opened 2024-10-23
feat(frontend): Improve added block positioning logic to handle collisions and dynamic dimensions size

Abhi1992002:
### Background #8268 

These changes enhance the placement algorithm for new blocks, ensuring they don’t overlap with existing ones despite varying dimensions. They dynamically adjust block positions and the viewing canvas to optimize layout and visibility.

### Changes 🏗️

1. Added findNodeDimensions function to calculate the dimensions of all nodes.
2. Introduced findNewlyAddedBlockCoordinates to determine the coordinates for newly added blocks.
3. Implemented rectanglesOverlap to ensure that two nodes do not overlap.
4. Enhanced functionality to zoom into a node immediately after it is added.

https://github.com/user-attachments/assets/03c46eea-b507-4595-80a0-2b06fc3f106c


autogpt_platform/frontend/src/components/Flow.tsx
```diff
@@ -402,16 +424,37 @@ const FlowEditor: React.FC<{
         return;
       }
 
-      // Calculate the center of the viewport considering zoom
-      const viewportCenter = {
-        x: (window.innerWidth / 2 - x) / zoom,
-        y: (window.innerHeight / 2 - y) / zoom,
-      };
+      // What we are going to do??
```
majdyz: nit: I think we can remove this first line comment

autogpt_platform/frontend/src/components/Flow.tsx
```diff
@@ -402,16 +424,37 @@ const FlowEditor: React.FC<{
...
+      // Calculate a position to the right of the newly added block, allowing for some margin.
+      // If adding to the right side causes the new block to collide with an existing block, attempt to place it at the bottom or left.
+      // Why not the top? Because the height of the new block is unknown.
+      // If it still collides, run a loop to find the best position where it does not collide.
+      // Then, adjust the canvas to center on the newly added block.
+      // Note: The width is known, e.g., w = 300px for a note and w = 500px for others, but the height is dynamic.
+
+      // We could also use D3 force, Intersection for this
```
majdyz: nit: merge this comment with above. I think it's cleaner to use multiline comment for this /* */.

> We could also use D3 force, Intersection for this

Add `Future improvement: `  or `Alternative:`?
Btw, I think the current approach is really good as it is 🔥

autogpt_platform/frontend/src/components/Flow.tsx
```diff
@@ -402,16 +424,37 @@ const FlowEditor: React.FC<{
...
+          60 / zoom,
+          zoom,
+        );
+      } else {
+        // Calculate the center of the viewport considering zoom
+        viewportCoordinates = {
+          x: (window.innerWidth / 2 - x) / zoom,
+          y: (window.innerHeight / 2 - y) / zoom,
+        };
+      }
```
majdyz: small nit: we can keep `viewportCoordinates` a const (since we should not modify it) by using the ternary operator.
```suggestion
      const viewportCoordinates =
        nodeDimensions && Object.keys(nodeDimensions).length > 0
          ? // we will get all the dimension of nodes, then store
            findNewlyAddedBlockCoordinates(
              nodeDimensions,
              (nodeSchema.uiType == BlockUIType.NOTE ? 300 : 500) / zoom,
              60 / zoom,
              zoom,
            )
          : // we will get all the dimension of nodes, then store
            {
              x: (window.innerWidth / 2 - x) / zoom,
              y: (window.innerHeight / 2 - y) / zoom,
            };
```

autogpt_platform/frontend/src/components/Flow.tsx
```diff
@@ -433,6 +476,15 @@ const FlowEditor: React.FC<{
       setNodeId((prevId) => prevId + 1);
       clearNodesStatusAndOutput(); // Clear status and output when a new node is added
 
+      setViewport(
+        {
+          x: -viewportCoordinates.x * zoom + window.innerWidth / 2,
+          y: -viewportCoordinates.y * zoom + window.innerHeight / 2 - 100,
+          zoom: 0.8,
```
majdyz, line 482: Why is this 0.8 not 1.0 ?

majdyz (dismissed):
This is really good! Thank you so much for this 👍 
I left a few very minor code styling nits if you don't mind addressing it.
Otherwise, I'm happy to merge this PR.

autogpt_platform/frontend/src/components/Flow.tsx
Abhi1992002: I have tried different values for zoom, and 0.8 looks visually better, but we could also use 1.0.

majdyz (approved):
LGTM

PR #8408, opened 2024-10-23
fix(backend): Fix error pin output not being propagated into the next nodes

majdyz:
### Background

Some blocks have an error output pin, this pin should be auto-populated upon any exception in the block code.
But currently, this output is not propagated into the next node, because the execution is halted on error.

### Changes 🏗️

Add propagation logic to the block exception processing step, to allow connected error pin output to be propagated into the next node.
image

- Upload agent to marketplace


autogpt_platform/backend/backend/executor/manager.py
```diff
@@ -155,18 +155,19 @@ def update_execution(status: ExecutionStatus) -> ExecutionResult:
     # changes during execution. ⚠️ This means a set of credentials can only be used by
     # one (running) block at a time; simultaneous execution of blocks using same
     # credentials is not supported.
-    credentials = creds_lock = None
```
majdyz, line 158: unused variable

autogpt_platform/backend/backend/executor/manager.py
```diff
@@ -185,31 +186,41 @@ def update_execution(status: ExecutionStatus) -> ExecutionResult:
             ):
                 yield execution
 
-        # Release lock on credentials ASAP
```
majdyz, line 188: moving this piece to the finally block

autogpt_platform/backend/backend/executor/manager.py
```diff
@@ -185,31 +186,41 @@ def update_execution(status: ExecutionStatus) -> ExecutionResult:
...
-
     except Exception as e:
+        end_status = ExecutionStatus.FAILED
         error_msg = str(e)
         log_metadata.exception(f"Node execution failed with error {error_msg}")
         db_client.upsert_execution_output(node_exec_id, "error", error_msg)
-        update_execution(ExecutionStatus.FAILED)
 
-        raise e
+        for execution in _enqueue_next_nodes(
```
majdyz, line 195: main code for the fix

autogpt_platform/backend/backend/executor/manager.py
```diff
@@ -155,18 +155,19 @@ def update_execution(status: ExecutionStatus) -> ExecutionResult:
     # changes during execution. ⚠️ This means a set of credentials can only be used by
     # one (running) block at a time; simultaneous execution of blocks using same
     # credentials is not supported.
-    credentials = creds_lock = None
+    creds_lock = None
     if CREDENTIALS_FIELD_NAME in input_data:
         credentials_meta = CredentialsMetaInput(**input_data[CREDENTIALS_FIELD_NAME])
         credentials, creds_lock = creds_manager.acquire(user_id, credentials_meta.id)
```
ntindle, line 161: Isn’t it used here?

majdyz: it's another assignment, so it's overridden

autogpt_platform/backend/backend/executor/manager.py
ntindle: i think the goal was to release ASAP here so we'd want that put back as it will block any other block from executing

autogpt_platform/backend/backend/executor/manager.py
```diff
@@ -185,31 +186,41 @@ def update_execution(status: ExecutionStatus) -> ExecutionResult:
...
-        s = input_size + output_size
-        t = (
-            (r.end_time - r.start_time).total_seconds()
-            if r.end_time and r.start_time
-            else 0
-        )
-        db_client.spend_credits(user_id, credit, node_block.id, input_data, s, t)
-
     except Exception as e:
+        end_status = ExecutionStatus.FAILED
```
ntindle, line 190: we should also release here tbh if speed is the goal

autogpt_platform/backend/backend/executor/manager.py
majdyz: There is no code between this line and finally block where the cred_lock release is done.

autogpt_platform/backend/backend/executor/manager.py
majdyz: the main code is not releasing before `upsert_execution_output` too, I don't want to introduce any side-effects.

ntindle (approved):

ntindle (approved):

PR #8418, opened 2024-10-24
Add URL swapping for marketplace based on environment

### LINKED ISSUE #8371: [Market] Allow use of dev marketplace in dev project
Currently, we can use only the local marketplace or the production market place. We should be able to swap the URL based on dev or prod environment in marketplace

Abhi1992002:
### Fixes #8371

These changes are needed to automatically switch between local and production marketplace URLs, ensuring the app connects to the correct environment (dev or prod) without manual intervention.

### Changes 🏗️

1. Swaps marketplace URL based on APP_ENV (dev or prod).
2. Ensures correct URL is used for local or production environments.


autogpt_platform/frontend/src/components/nav/MarketPopup.tsx
```diff
@@ -7,7 +7,9 @@ interface MarketPopupProps extends ButtonHTMLAttributes<HTMLButtonElement> {
 
 export default function MarketPopup({
   className = "",
-  marketplaceUrl = "http://platform.agpt.co/marketplace",
+  marketplaceUrl = process.env.APP_ENV === "prod"
+    ? "https://production-marketplace-url.com"
```
ntindle: can we keep the existing link for prod and add  a spot for one for "dev": https://dev-builder.agpt.co/marketplace and finally have the localhost fallback for local

ntindle (changes requested):

ntindle (approved):

autogpt_platform/frontend/src/components/nav/MarketPopup.tsx
```diff
@@ -7,7 +7,15 @@ interface MarketPopupProps extends ButtonHTMLAttributes<HTMLButtonElement> {
...
-  marketplaceUrl = "http://platform.agpt.co/marketplace",
+  marketplaceUrl = (() => {
+    if (process.env.APP_ENV === "prod") {
+      return "https://production-marketplace-url.com";
+    } else if (process.env.APP_ENV === "dev") {
+      return "https://dev-builder.agpt.co/marketplace";
+    } else {
+      return "http://localhost:3000/marketplace";
+    }
+  })(),
```
ntindle: This isn't actually accessible on the frontend in testing :/ i think we will need a different way to handle this

ntindle (changes requested):

autogpt_platform/frontend/src/components/nav/MarketPopup.tsx
aarushik93: I think you want to use the build variables

autogpt_platform/frontend/src/components/nav/MarketPopup.tsx
```diff
@@ -7,7 +7,15 @@ interface MarketPopupProps extends ButtonHTMLAttributes<HTMLButtonElement> {
 
 export default function MarketPopup({
   className = "",
-  marketplaceUrl = "http://platform.agpt.co/marketplace",
+  marketplaceUrl = (() => {
+    if (process.env.NEXT_PUBLIC_APP_ENV === "prod") {
+      return "https://production-marketplace-url.com";
+    } else if (process.env.NEXT_PUBLIC_APP_ENV === "dev") {
```
aarushik93, line 13: this is a new variable, right, can you please add it to the .env example files in frontend

PR #8429, opened 2024-10-24
fix(backend): Avoid long synchronous call to block FastAPI event-loop

majdyz:
### Background

We are doing some synchronous calls (pyro calls) on our FastAPI handler, which can potentially block the event loop and render the API unavailable.

### Changes 🏗️

* Remove `async` on function handlers that call pyro & don't require async calls.
* Uses `asyncio.to_thread` to execute pyro calls that mix with async calls.

- Upload agent to marketplace


autogpt_platform/autogpt_libs/autogpt_libs/utils/cache.py
```diff
@@ -21,7 +18,3 @@ def wrapper(*args: P.args, **kwargs: P.kwargs) -> R:
         return cache[key]
 
     return wrapper
-
-
-def thread_cached_property(func: Callable[[T], R]) -> property:
```
majdyz, line 26: Removing this since it's useless and doesn't have a strong typing

autogpt_platform/backend/backend/server/integrations/router.py
```diff
@@ -29,7 +29,7 @@ class LoginResponse(BaseModel):
 
 
 @router.get("/{provider}/login")
-async def login(
```
majdyz, line 32: All these routes use db_manager inside and doesn't use prisma, so replace all of them to sync def

autogpt_platform/backend/backend/server/rest_api.py
```diff
@@ -539,7 +542,9 @@ async def stop_graph_run(
                 404, detail=f"Agent execution #{graph_exec_id} not found"
             )
 
-        self.execution_manager_client.cancel_execution(graph_exec_id)
+        await asyncio.to_thread(
+            lambda: self.execution_manager_client.cancel_execution(graph_exec_id)
+        )
```
majdyz, line 547: pyro call inside the async function need to do this, to avoid blocking the event loop in case the function call is slow.

autogpt_platform/backend/backend/usecases/block_autogen.py
```diff
@@ -252,7 +252,7 @@ async def block_autogen_agent():
         test_user = await create_test_user()
         test_graph = await create_graph(create_test_graph(), user_id=test_user.id)
         input_data = {"input": "Write me a block that writes a string into a file."}
-        response = await server.agent_server.execute_graph(
+        response = server.agent_server.execute_graph(
```
majdyz, line 255: Make execute_graph sync, these are the propagated changes

ntindle (approved):

Issue #7809, opened 2024-08-18
Add Support for Inbound Webhooks


ntindle:
Tools like Payman, compass need inbound webhooks to work. This being in the same block is probably a good idea. Figure out how we do it

PR #8430, opened 2024-10-24
docs(pr_template): Add reminder for config and infra changes

aarushik93:
### Background

I've noticed we tend to forget about making the correct infra and configuration changes when we make local changes to the way the application runs. This breaks local dev flow for some people as well as the cloud. 

### Changes 🏗️

Adding a reminder for such changes

- Upload agent to marketplace


ntindle (approved):

ntindle:
We should be able to make an agent for this when #7809 comes in

PR #8501, opened 2024-10-30
feat(frontend): Show Agent Output on Monitor page

kcze:
### Background

There's no way to see Agent Outputs in Monitor, there are only on the Build page when running an agent.

### Changes 🏗️

- Add `View Outputs` button to Agent runs in Monitor
- Reuse `Sheet` component used on the Build page to show results of Agent Output nodes

- Upload agent to marketplace


autogpt_platform/frontend/src/components/monitor/FlowRunInfo.tsx
```diff
@@ -1,79 +1,160 @@
...
+  const api = useMemo(() => new AutoGPTServerAPI(), []);
+
+  // Fetch graph and execution data
+  useEffect(() => {
+    api.getGraph(flow.id, flow.version).then((graph) => {
+      setGraph(graph);
+    });
+    api.getGraphExecutionInfo(flow.id, flowRun.id).then((executions) => {
+      setExecutionResults(executions);
+    });
```
kcze: I had to get executions info because `flowRun.nodeExecutionResults` is always `undefined`

autogpt_platform/frontend/src/components/monitor/FlowRunInfo.tsx
```diff
@@ -1,79 +1,160 @@
...
+        }
+
+        return acc;
+      }, new Map<string, NodeExecutionResult>());
+
+    // Transform output nodes to BlockOutput format
+    setBlockOutputs(
+      outputNodes.map((node) => ({
+        id: node.id,
+        outputSchema: {} as BlockIORootSchema,
```
kcze: `outputSchema` isn't used in `RunnerOutputUI` so leaving it empty here. All Blocks metadata would have to be fetched otherwise.

autogpt_platform/frontend/src/components/monitor/FlowRunInfo.tsx
majdyz: I think there are `with_runs` query params being added on /graphs ? can we utilize that?
Also, are we going to fetch literally all the executions of the graph?

autogpt_platform/frontend/src/components/monitor/FlowRunInfo.tsx
```diff
@@ -1,79 +1,160 @@
...
+  }, [api]);
+
+  useEffect(() => {
+    if (!graph || !executionResults) {
+      return;
+    }
+
+    // Filter nodes to only get output blocks
+    const outputNodes = graph.nodes.filter(
+      (node) => node.block_id === "363ae599-353e-4804-937e-b2ee3cef3da4",
```
majdyz: There uiType == BlockType.OUTPUT

autogpt_platform/frontend/src/components/monitor/FlowRunInfo.tsx
majdyz: Can we fetch the output only when it's clicked ?

autogpt_platform/frontend/src/components/monitor/FlowRunInfo.tsx
majdyz: If we load output lazily, this would not be required?
We are fetching latest completedrequsult either way

autogpt_platform/frontend/src/components/monitor/FlowRunInfo.tsx
```diff
@@ -1,79 +1,160 @@
...
-        {/* <p><strong>Total cost:</strong> €1,23</p> */}
-      </CardContent>
-    </Card>
+          <p>
+            <strong>Finished:</strong>{" "}
+            {moment(flowRun.endTime).format("YYYY-MM-DD HH:mm:ss")}
+          </p>
+          <p>
+            <strong>Duration (run time):</strong> {flowRun.duration} (
+            {flowRun.totalRunTime}) seconds
```
majdyz, line 142: It's unrelated to your PR but I don't see the point of totalRunTime here, the maintainer of the system might care but why does the user care about this number? We might just remove this info altogether.

autogpt_platform/frontend/src/components/monitor/FlowRunInfo.tsx
kcze: > I think there are with_runs query params being added on /graphs ? can we utilize that?

Why? It doesn't return actual results but only all runs metadata.

> Also, are we going to fetch literally all the executions of the graph?

No, it fetches execution info for a single run that is selected. I'm unsure about fetching all runs metadata on the Monitor but that's unrelated to this PR.

> Can we fetch the output only when it's clicked ?

Done

autogpt_platform/frontend/src/components/monitor/FlowRunInfo.tsx
kcze: I believe I did what you think now

autogpt_platform/frontend/src/components/monitor/FlowRunInfo.tsx
kcze: Where? I don't see in in ts `Node` and there's no `BlockType`
edit: I found `data.uiType` (`BlockUIType`) but that's for frontend `Node` type which I cannot access because this isn't Build page. Here I have mirror `Node` type from backend.

autogpt_platform/frontend/src/components/monitor/FlowRunInfo.tsx
kcze: I've removed `totalRunTime`

autogpt_platform/frontend/src/components/monitor/FlowRunInfo.tsx
```diff
@@ -1,79 +1,158 @@
...
     flow: GraphMeta;
     flowRun: FlowRun;
   }
 > = ({ flow, flowRun, ...props }) => {
+  const [isOutputOpen, setIsOutputOpen] = useState(false);
+  const [blockOutputs, setBlockOutputs] = useState<BlockOutput[]>([]);
+  const api = useMemo(() => new AutoGPTServerAPI(), []);
+
+  const fetchBlockResults = useCallback(async () => {
+    console.log("fetchBlockResults");
```
majdyz: nit: remove this log ?

autogpt_platform/frontend/src/components/monitor/FlowRunInfo.tsx
```diff
@@ -1,79 +1,158 @@
...
+    const graph = await api.getGraph(flow.id, flow.version);
```
majdyz: `/graphs/{graph_id}/executions/{graph_exec_id}` has block_id as part of its response. I don't think we need to do nested requests for graph here?

autogpt_platform/frontend/src/components/monitor/FlowRunInfo.tsx
```diff
@@ -1,79 +1,158 @@
...
+  }, [api, flow.id, flow.version, flowRun.id]);
+
+  // Fetch graph and execution data
+  useEffect(() => {
+    if (!isOutputOpen || blockOutputs.length > 0) {
+      return;
+    }
+
+    fetchBlockResults();
+  }, [api, isOutputOpen]);
```
majdyz: Does this pass the linting? should `blockOutputs` part of the deps list ? should we ignore the lint and add some comments ?

autogpt_platform/frontend/src/components/monitor/FlowRunInfo.tsx
```diff
@@ -1,79 +1,158 @@
...
-          <strong>Duration (run time):</strong> {flowRun.duration} (
-          {flowRun.totalRunTime}) seconds
-        </p>
-        {/* <p><strong>Total cost:</strong> €1,23</p> */}
-      </CardContent>
-    </Card>
+          <p>
+            <strong>Duration (run time):</strong> {flowRun.duration}s
+          </p>
+          {/* <p><strong>Total cost:</strong> €1,23</p> */}
```
majdyz: nit: I know this might not be your change, but can you remove this comment if you end up adding more commits on this PR ? :D

autogpt_platform/frontend/src/components/monitor/FlowRunInfo.tsx
majdyz: This is what confuses me, should we just remove this field then if it's not used? can we do that?

majdyz (approved):

autogpt_platform/frontend/src/components/monitor/FlowRunInfo.tsx
kcze: right! removed

autogpt_platform/frontend/src/components/monitor/FlowRunInfo.tsx
kcze: thanks, missed it

autogpt_platform/frontend/src/components/monitor/FlowRunInfo.tsx
kcze: You are right, ts type was missing `block_id`, added

autogpt_platform/frontend/src/components/monitor/FlowRunInfo.tsx
kcze: I've fixed deps

majdyz (approved):
I'm adding https://github.com/Significant-Gravitas/AutoGPT/pull/8501/commits/e623d1692a2759d7ac6d71541644441b4daeef06 to cleanup the hardcoded ID, since now we have two places where this is needed. Let me know if this is okay.

Otherwise looks good! Thanks for adding this!

Btw, why aren't we shoding the agent input too? I think it's equally useful.

Torantulino:
> I'm adding [e623d16](https://github.com/Significant-Gravitas/AutoGPT/commit/e623d1692a2759d7ac6d71541644441b4daeef06) to cleanup the hardcoded ID, since now we have two places where this is needed. Let me know if this is okay.
> 
> Otherwise looks good! Thanks for adding this!
> 
> By the way, why aren't we showing the agent input? It's equally useful.

The new library page does do this so I don't want to spend too much time on work that's going to be overwritten, but yeah it would also be very easy to add that button, good point!

PR #8502, opened 2024-10-30
docs: replace docker-compose with docker compose

simoneb:
### Background

`docker-compose` is deprecated as compose is now built-into docker, and the command becomes `docker compose`.

### Changes 🏗️

Updates docs to reflect the right command.

### Testing 🔍

no tests required

### Configuration Changes 📝

no configuration changes


aarushik93 (changes requested):
Please remove irrelvant changes

PR #8519, opened 2024-11-01
refactor(platform/rest): Update REST API to use standard FastAPI structure

Swiftyos:
### Background

The rest FastAPI server is currently packaged up and ran from within a separate process. This has resulted in needing to develop the FastAPI server in a unique way, preventing us from leveraging some of its features and slightly increasing the difficulty of adding new routes.

Work has been done to decouple the Rest server from the executors. This is the follow on work to refactor the API server to a canonical FastAPI setup.

### Changes 🏗️

- Moved to a standard FastAPI server, ran via uvicorn directly, in the docker container.
- Ensured openapi schemas matched exactly:

    Added python script to ensure openapi specs do not change during refactor
    ```
     % python comp.py                                     
    The two OpenAPI specifications are identical.
    ```
- Moved endpoints from `/api/x` to `/api/v1/x`

- Upload agent to marketplace

- Changing ports


autogpt_platform/backend/backend/rest.py
```diff
@@ -1,17 +1,16 @@
...
-
+import uvicorn
+import backend.util.settings
 
 def main():
     """
     Run all the processes required for the AutoGPT-server REST API.
     """
-    run_processes(
-        ExecutionScheduler(),
```
Swiftyos: The server was still coupled with the Scheduler.

As we are not using it, I've marked the endpoints as depreciated, until such time that we fix how we are handling the scheduler. 

cc: @majdyz

autogpt_platform/backend/Dockerfile
```diff
@@ -75,4 +75,4 @@ COPY autogpt_platform/backend /app/autogpt_platform/backend
 ENV DATABASE_URL=""
 ENV PORT=8000
 
-CMD ["poetry", "run", "rest"]
+CMD ["poetry", "run", "rest"]
```
majdyz: not required

autogpt_platform/backend/test/conftest.py
```diff
@@ -12,7 +12,7 @@ async def server():
 @pytest.fixture(scope="session", autouse=True)
 async def graph_cleanup(server):
     created_graph_ids = []
-    original_create_graph = server.agent_server.create_graph
+    original_create_graph = server.agent_server.test_create_graph
```
majdyz, line 28: This fixture wraps `create_graph` so that you will clean up all the created graph post-tests.
So the code below it.

If you move everything to test_create_graph, then each test execution will keep piling more graphs without any cleanups.

autogpt_platform/backend/test/executor/test_manager.py
```diff
@@ -9,9 +11,22 @@
...
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
+
+# Create console handler with formatting
+ch = logging.StreamHandler()
+ch.setLevel(logging.INFO)
+formatter = logging.Formatter("%(asctime)s - %(name)s - %(levelname)s - %(message)s")
+ch.setFormatter(formatter)
+logger.addHandler(ch)
+
```
majdyz, line 15: Shouldn't you do this on conftest?

autogpt_platform/backend/backend/server/rest_api.py
```diff
@@ -1,695 +1,117 @@
...
-    ) -> list[execution_db.ExecutionResult]:
-        if not await execution_db.get_graph_execution(graph_exec_id, user_id):
-            raise HTTPException(
-                404, detail=f"Agent execution #{graph_exec_id} not found"
-            )
-
-        await asyncio.to_thread(
-            lambda: self.execution_manager_client.cancel_execution(graph_exec_id)
+    @staticmethod
+    async def test_get_graph_run_node_execution_results(
```
majdyz, line 105: why is it renamed?

Swiftyos: To mark it clearly as a function that is only used during testing. It is not used when running the system

autogpt_platform/backend/test/conftest.py
Swiftyos: I've changed it to wrap `test_create_graph` instead.

The reason for this, is to allow these functions to still be accessable via the AppSever class, which is how the tests access this functionality.

I renamed them test to clearly denote they are not used during runtime.

majdyz (approved):
codewise LGTM,
I have no opinion on v1 routing, is this something that we want to do?
Should we not retain the default one to point to the latest version to maintain backward compatibility?

PR #8524, opened 2024-11-01
feat(platform): Simplify Credentials UX

kcze:
### Background

Current credentials system always forces users to choose correct credentials, even when they are provided by the cloud platform and should be chosen automatically (e.g. there's only one possible choice).

### Changes 🏗️

- Change `provider` of default credentials to actual provider names (e.g. `anthropic`), remove `llm` provider
- Add `discriminator` and `discriminator_mapping` to `CredentialsField` that allows to filter credentials input to only allow  providers for matching models in `useCredentials` hook (thanks @ntindle for the idea!); e.g. user chooses `GPT4_TURBO` so then only OpenAI credentials are allowed
- Choose credentials automatically and hide credentials input on the node completely if there's only one possible option
- Move `getValue` and `parseKeys` to utils
- Add `ANTHROPIC`, `GROQ` and `OLLAMA` to providers in frontend `types.ts`
- Add `hidden` field to credentials that is used for default system keys to hide them in user profile
- Now `provider` field in `CredentialsField` can accept multiple providers as a list

- Upload agent to marketplace


autogpt_platform/frontend/src/lib/utils.ts
```diff
@@ -312,3 +312,48 @@ export function findNewlyAddedBlockCoordinates(
...
+      }
+    }
+  });
+
+  if (currentKey !== null) {
+    keys.push({ key: currentKey });
+  }
+
+  return keys;
+}
```
ntindle, line 346: we should probably add security test for this lol

ntindle:
This is exactly what I was thinking! great job!

autogpt_platform/frontend/src/lib/autogpt-server-api/types.ts
```diff
@@ -96,16 +96,19 @@ export type CredentialsType = "api_key" | "oauth2";
 
 // --8<-- [start:BlockIOCredentialsSubSchema]
 export const PROVIDER_NAMES = {
+  ANTHROPIC: "anthropic",
   D_ID: "d_id",
   DISCORD: "discord",
   GITHUB: "github",
   GOOGLE: "google",
   GOOGLE_MAPS: "google_maps",
+  GROQ: "groq",
```
ntindle, line 107: probably need logos for these too

ntindle: (even if fallback)

ntindle:
I think the devx and ux of this leaves something to be desired still. I think the dev specifying the valid providers as a list would be useful and then it returns one of the credentials depending on the selected API key added. 

https://github.com/user-attachments/assets/afe1cee3-430f-48e1-94c4-d0c39bbde8f6

Also not sure how they are added to the db and filtered but I'm not seeing them after creation, and crashing anytime they're rendered. I think there's an issue somewhere. lmk if you can't replicate

ntindle (changes requested):
Tested, needs addtional work

autogpt_platform/frontend/src/lib/utils.ts
kcze: I agree but this code is already in the codebase I just moved it!

autogpt_platform/frontend/src/lib/autogpt-server-api/types.ts
kcze: done

autogpt_platform/backend/backend/blocks/llm.py
```diff
@@ -52,6 +52,10 @@ def AICredentialsField() -> AICredentials:
         description="API key for the LLM provider.",
         provider="llm",
```
ntindle: imo this needs to be a structure that shows the allowable providers, not its own provider. something like 

valid_providers = ['openai', 'grok', 'anthropic']
and be used in conjunction with discriminator always to have one of the providers be selected

autogpt_platform/frontend/src/app/profile/page.tsx
```diff
@@ -76,15 +74,15 @@ export default function PrivatePage() {
...
+      .filter((cred) => !cred.hidden)
+      .map((credentials) => ({
         ...credentials,
         provider: provider.provider,
         providerName: provider.providerName,
         ProviderIcon: providerIcons[provider.provider],
         TypeIcon: { oauth2: IconUser, api_key: IconKey }[credentials.type],
-      }),
-    ),
+      })),
```
ntindle: this works fine for me right now

autogpt_platform/frontend/src/components/integrations/credentials-input.tsx
```diff
@@ -56,6 +58,7 @@ export const providerIcons: Record<
   ideogram: fallbackIcon,
   llm: fallbackIcon,
```
ntindle: we should be able to remove this once we get this right

autogpt_platform/frontend/src/components/integrations/credentials-input.tsx
```diff
@@ -91,14 +94,10 @@ export const CredentialsInput: FC<{
...
 
-  if (!credentials) {
+  if (!credentials || credentials.isLoading) {
     return null;
   }
 
-  if (credentials.isLoading) {
-    return <div>Loading...</div>;
-  }
-
```
ntindle, line 101: why this change?

autogpt_platform/frontend/src/components/integrations/credentials-input.tsx
```diff
@@ -246,6 +256,31 @@ export const CredentialsInput: FC<{
...
+  } else if (savedApiKeys.length === 1 && savedOAuthCredentials.length === 0) {
+    if (selectedCredentials === undefined) {
+      onSelectCredentials({
+        id: savedApiKeys[0].id,
+        type: "api_key",
+        provider,
+        title: savedApiKeys[0].title,
+      });
+    }
+    return null;
```
ntindle: What's the purpose of this? This whole chain of logic seems questionably necessary.

ntindle (changes requested):
This still isn't wokring for me 
image
There's nowhere to select credentials, and we're still creating llm providers which we shouldn't need

ntindle:
```
app-index.tsx:25 
 Warning: Cannot update a component (`BatchProvider`) while rendering a different component (`CredentialsInput`). To locate the bad setState() call inside `CredentialsInput`, follow the stack trace as described in https://reactjs.org/link/setstate-in-render Error Component Stack
    at CredentialsInput (credentials-input.tsx:87:9)
    at div (<anonymous>)
    at NodeCredentialsInput (node-input-components.tsx:309:9)
    at NodeGenericInputField (
```
maybe this error helps

autogpt_platform/frontend/src/components/integrations/credentials-input.tsx
kcze: Now `llm` provider is removed

autogpt_platform/frontend/src/components/integrations/credentials-input.tsx
kcze: So it doesn't show "Loading..." and then disappear if there's one credentials (it's chosen by default and nothing is displayed)

ntindle (approved):

ntindle:
```
app-index.tsx:25 
 Warning: Cannot update a component (`BatchProvider`) while rendering a different component (`CredentialsInput`). To locate the bad setState() call inside `CredentialsInput`, follow the stack trace as described in https://reactjs.org/link/setstate-in-render Error Component Stack
    at CredentialsInput (credentials-input.tsx:86:9)
    at div (<anonymous>)
    at NodeCredentialsInput (node-input-components.tsx:309:9
    ```
    
   Open a ticket to fix this.
   
   
   Add docs in a separate PR

File: autogpt_platform/backend/backend/data/model.py [hunk 8]
```diff
@@ -164,16 +166,30 @@ def CredentialsField(
...
+    # Convert provider list to enum for schema
+    provider_list = [provider] if isinstance(provider, str) else provider
+
     json_extra = {
         k: v
         for k, v in {
-            "credentials_provider": provider,
+            "credentials_provider": (
+                provider if isinstance(provider, str) else list(provider)
+            ),
```
Pwuts: this seems to be the wrong way around

autogpt_platform/backend/backend/data/model.py
```diff
@@ -164,16 +166,30 @@ def CredentialsField(
...
         if v is not None
     }
 
+    # Force provider to always be a string
+    # Otherwise input provider will fail jsonschema validation
+    # if multiple providers are provided
+    json_extra.update(
+        {"properties": {"provider": {"type": "string", "enum": provider_list}}}
+    )
+
```
Pwuts: I think you can (and should) avoid this hack by changing the `provider` parameter to `provider: CP | list[CP]` and change the definition of `CP` back to `TypeVar("CP", bound=str)`

autogpt_platform/backend/backend/blocks/llm.py
```diff
@@ -30,11 +30,11 @@
 #     "ollama": BlockSecret(value=""),
 # }
 
-AICredentials = CredentialsMetaInput[Literal["llm"], Literal["api_key"]]
+AICredentials = CredentialsMetaInput[list[str], Literal["api_key"]]
```
Pwuts:
```suggestion
LLMProviderName = Literal["anthropic", "groq", "openai", "ollama"]
AICredentials = CredentialsMetaInput[LLMProviderName, Literal["api_key"]]
```

autogpt_platform/frontend/src/components/integrations/credentials-input.tsx
```diff
@@ -80,7 +82,7 @@ export type OAuthPopupResultMessage = { message_type: "oauth_popup_result" } & (
 export const CredentialsInput: FC<{
   className?: string;
   selectedCredentials?: CredentialsMetaInput;
-  onSelectCredentials: (newValue: CredentialsMetaInput) => void;
+  onSelectCredentials: (newValue: CredentialsMetaInput | undefined) => void;
```
Pwuts:
```suggestion
  onSelectCredentials: (newValue?: CredentialsMetaInput) => void;
```

autogpt_platform/frontend/src/lib/autogpt-server-api/types.ts
```diff
@@ -258,7 +263,7 @@ export type CredentialsMetaInput = {
   id: string;
   type: CredentialsType;
   title?: string;
-  provider: string;
+  provider: string | string[];
```
Pwuts:
```suggestion
  provider: string;
```

autogpt_platform/autogpt_libs/autogpt_libs/supabase_integration_credentials_store/types.py
```diff
@@ -8,6 +8,7 @@ class _BaseCredentials(BaseModel):
     id: str = Field(default_factory=lambda: str(uuid4()))
     provider: str
     title: Optional[str]
+    hidden: Optional[bool] = False
```
Pwuts: 1. "Hidden" isn't the right name for this, since it doesn't generally hide the item except on a specific screen.
2. Putting this here will also add and persist it to the DB on all new credentials created. This isn't necessary and should be avoided if possible.

autogpt_platform/frontend/src/components/integrations/credentials-input.tsx
```diff
@@ -246,6 +255,31 @@ export const CredentialsInput: FC<{
         )}
       </>
     );
+    // Only one saved credential
+  } else if (
+    savedApiKeys.length === 0 &&
+    savedOAuthCredentials.length === 1 &&
+    selectedCredentials === undefined
+  ) {
+    if (selectedCredentials === undefined) {
```
Pwuts: redundant check

autogpt_platform/frontend/src/components/integrations/credentials-input.tsx
Pwuts: Reading through this PR, I don't really see why this shortcut is being taken. What's the big hurdle preventing the regular credentials input from working on a multi-provider block?

autogpt_platform/autogpt_libs/autogpt_libs/supabase_integration_credentials_store/types.py
Pwuts: Putting this here will also persist it in the DB, which I think should be avoided if possible.

IMO there should be minimal involvement of the credentials system when a user wants to use credits to run a block. Right now we're injecting credentials into the credential system in a way it definitely wasn't designed for and that doesn't make sense to me as anything but a quick fix (which is what it was). This PR preferably shouldn't further entrench this solution by expanding on it.

I'd extend the `CredentialsInput` with a `credits_available: bool | list[CP]` parameter. Then the frontend can render an option "Use credits" in the credentials picker, which (when selected) sets `credentials = null`, or some magic value like `credentials = {id: 0, provider: "agpt-credits"}`.

Pwuts: Another option would be to put the `credentials` and `model` input together in a discriminated union sub-input, with `model` as the discriminator, and implement some JSON Schema based discrimination logic in the frontend.

autogpt_platform/backend/backend/blocks/llm.py
kcze: this is done! now `provider` can be a `list` (and is for AI blocks)

autogpt_platform/frontend/src/components/integrations/credentials-input.tsx
kcze: I'm not sure I understand @Pwuts. The purpose of this code is to choose the only one existing credentials and then not show dropdown at all. I simplified this code.

autogpt_platform/autogpt_libs/autogpt_libs/supabase_integration_credentials_store/types.py
kcze: I've removed `hidden`, hardcoded hidden ids on the frontend and left a `TODO` for now.

kcze: `hidden` is removed. As for the `CredentialsInput` extension and merging `credentials` and `model` (I also thought about this and it is a good solution imo) I think that'a outside the PR scope here.

autogpt_platform/backend/backend/data/model.py
Pwuts, line 1: Just a note: this really has to be considered a tempfix. Scopes are provider-specific. So is `supported_credential_types`. The `CredentialsField` was made to handle credentials from a specific provider. If a block supports multiple providers, the discrimination logic should be handled outside the `CredentialsField`.

autogpt_platform/frontend/src/app/profile/page.tsx
```diff
@@ -62,7 +60,22 @@ export default function PrivatePage() {
...
+      "fdb7f412-f519-48d1-9b5f-d2f73d0e01fe", // Revid
+      "760f84fc-b270-42de-91f6-08efe1b512d0", // Ideogram
+      "6b9fc200-4726-4973-86c9-cd526f5ce5db", // Replicate
+      "53c25cb8-e3ee-465c-a4d1-e75a4c899c2a", // OpenAI
+      "24e5d942-d9e3-4798-8151-90143ee55629", // Anthropic
+      "4ec22295-8f97-4dd1-b42b-2c6957a02545", // Groq
+      "7f7b0654-c36b-4565-8fa7-9a52575dfae2", // D-ID
+    ],
+    [],
+  );
```
Pwuts, line 76: As temporary fixes go this is nice

autogpt_platform/frontend/src/components/integrations/credentials-input.tsx
Pwuts: My question is: would it break anything to allow users the choice of adding their own API key? I understand this will be resolved in a follow-up PR, but I don't immediately see what would break right now if you still showed the credentials picker in all cases.

autogpt_platform/backend/backend/data/model.py
Pwuts: See also https://github.com/Significant-Gravitas/AutoGPT/pull/8524#discussion_r1833555961

Pwuts: (I'd like to leave this discussion here for reference in the follow-up PR)

Torantulino: Okay, I'm happy for us to merge this as a much needed temp hotfix and we can adjust that in a follow up.

Pwuts (approved):

PR #8531, opened 2024-11-03
fix (backend): Patching the SSRF vulnerability in Github/Web Search/Request related blocks

jackfromeast:
### Background

Refer to the report on huntr.com: https://huntr.com/bounties/1d91e1e1-7d45-4bda-bc27-bfe9052fd975

### Changes 🏗️

- Add the domain validation (i.e., `github.com`) to all the blocks under `/blocks/github`
- Add the URL validation to the Web Search block and Web Request block: Block all the private IP requests by default unless listed in the config file's `trust_endpoints_for_requests` field.

### Testing 🔍

Tested locally.

### Configuration Changes 📝

- Add a `trust_endpoints_for_requests` field for the developer to create whitelist domains to request.


autogpt_platform/backend/backend/blocks/github/issues.py
```diff
@@ -62,6 +68,10 @@ def __init__(self):
     def post_comment(
         credentials: GithubCredentials, issue_url: str, body_text: str
     ) -> tuple[int, str]:
+
+        if is_github_url(issue_url) is False:
+            raise ValueError("The input URL must be a valid GitHub URL.")
+
```
ntindle: What if we update input to be numeber/ string of the relevant inputs instead? @majdyz

majdyz: Any ideas in mind? what the user need to provide?

autogpt_platform/backend/backend/util/request.py
```diff
@@ -0,0 +1,128 @@
...
+BLOCKED_IP_NETWORKS = [
+    ipaddress.ip_network("0.0.0.0/8"),  # "This" Network
+    ipaddress.ip_network("10.0.0.0/8"),  # Private-Use
+    ipaddress.ip_network("127.0.0.0/8"),  # Loopback
+    ipaddress.ip_network("169.254.0.0/16"),  # Link Local
+    ipaddress.ip_network("172.16.0.0/12"),  # Private-Use
+    ipaddress.ip_network("192.168.0.0/16"),  # Private-Use
+    ipaddress.ip_network("224.0.0.0/4"),  # Multicast
+    ipaddress.ip_network("240.0.0.0/4"),  # Reserved for Future Use
+]
```
majdyz, line 20: Hi @jackfromeast 

So I ended up needing to do this to reduce the impact of blocked IP ranges.
Domains like google.com mapped to a [special purpose ip](https://datatracker.ietf.org/doc/rfc6890/) like `192.0.0.88` which your initial code blocks. 

Do you think this will have a security repercussion?

autogpt_platform/backend/backend/blocks/github/_api.py
```diff
@@ -0,0 +1,50 @@
+from urllib.parse import urlparse
+
+from backend.blocks.github._auth import GithubCredentials
+from backend.util.request import Requests
+
+
+def _validate_github_url(url: str) -> None:
+    parsed_url = urlparse(url)
+    if parsed_url.netloc != "github.com":
```
ntindle: github can also be hosted via enterprise on custom urls ex: github.agpt.co so this probably isn't a good design choice

ntindle: imo its insanity we require urls at all rather than the repo, owner, and source and then just build from there

autogpt_platform/backend/backend/util/request.py
ntindle, line 20: add tests for this with varoius things

ntindle (dismissed):

autogpt_platform/backend/backend/blocks/github/_api.py
ntindle: this is a series of nits ^

autogpt_platform/backend/backend/util/request.py
ntindle: This is serious

autogpt_platform/backend/backend/blocks/github/_api.py
majdyz: The two nits are contradictory. 

I removed the github.com check validation entirely, since we are replacing everything to api.github.com

> github can also be hosted via enterprise on custom URLs

so maybe you need to provide an option either way for changing the host because we are ignoring them.
The existing implementation doesn't support this. I don't even know if the self-hosted github would also have an API or not.

autogpt_platform/backend/backend/util/request.py
majdyz: Added

majdyz (approved):

PR #8541, opened 2024-11-04
dx: Clean up PR template

Pwuts:
image

There is stuff in the PR template that is not template stuff. It should either be commented out so that it's only visible to the author/editor, or moved to the contribution guide.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,5 +1,3 @@
-### Background
-
 <!-- Clearly explain the need for these changes: -->
 
 ### Changes 🏗️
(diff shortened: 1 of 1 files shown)


ntindle:
IMO Add a test plan requirement

autogpt_platform/README.md
Pwuts, line 1: These changes are from the base branch change and should be reverted

Pwuts:
@ntindle what do you mean exactly?

ntindle:
As one of the few reviewing PRs right now, it’s hard to know what things interact with and I need to test

Pwuts:
@ntindle I get that, but what does that mean for this PR? What text do you want to have placed where?

ntindle:
Swap those sections to writing test plans

ntindle:
In looking back at this, tbh I like how it is

Pwuts:
Reasons I don't like this stuff being in the PR template like it is now:
- It's a lot of text, but not relevant for a significant fraction of PRs.
- It doesn't provide any value to read it to anyone but the PR author.
- It isn't a checklist, which would be a much more useful for something like this.
- We're putting basic contribution instructions here, without pointing to any of our more elaborate contribution guides/instructions. This fragmentation isn't good for the contributor experience.

ntindle:
I think a better move would be converting it to a checklist and adding links to the areas you're discussing

ntindle (approved):

PR #8561, opened 2024-11-05
fix(docker): Removed all docker containers that are not strictly required by the platform

Swiftyos:
### Background

Removed all the supabase containers for services that are not being actively used by the platform atm

### Changes 🏗️

- Upload agent to marketplace

- Changing ports

### DIFF
--- a/autogpt_platform/docker-compose.yml
+++ b/autogpt_platform/docker-compose.yml
@@ -71,13 +71,14 @@ services:
 #      service: frontend
 
   # Supabase services
-  # Uncomment these 2 if you want to use the ui
+  # Uncomment theis if you want to use the ui
   # studio:
   #   <<: *supabase-services
   #   extends:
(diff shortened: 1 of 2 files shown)


autogpt_platform/docker-compose.yml
```diff
@@ -137,29 +141,31 @@ services:
...
-      - storage
-      - imgproxy
-      - meta
-      - functions
-      - analytics
-      - db
-      - vector
-      - redis
+  # deps:
+  #   <<: *supabase-services
```
majdyz, line 153: I need this for local development, this is the images need to be run except the application services, can we adjust this accordingly instead of commenting ?

Swiftyos: Currently in dev, it only requires the containers I have left uncommented. 
Which of the additional services do you need that have been removed?

This specifically doesn't really do anything other than exiting with success when the deps have started.

PR #8574, opened 2024-11-06
feat(backend): Improve pyro reliability by adding connection timeout, retry, cleanup, and dynamic connection thread size

majdyz:
### Background

To avoid the same issue happening with database we need to do some prevention on managing the lifecycle of the pyro connection.

### Changes 🏗️

Applying the recommendation here https://pyro5.readthedocs.io/en/latest/tipstricks.html#release-proxies-when-no-longer-used-avoids-after-x-simultaneous-proxy-connections-pyro-seems-to-freeze, e.g:
* Add pyro connection release timeout
* Add pyro connection retry/auto-reconnect
* Adjust the connection thread pool cap to match the maximum possible number of execution threads.
* Add manual cleanup for pyro connection on thread/process termination on the block executor.

- Upload agent to marketplace

- Changing ports


autogpt_platform/backend/backend/util/settings.py
```diff
@@ -69,6 +69,14 @@ class Config(UpdateTrackingModel["Config"], BaseSettings):
         default="localhost",
         description="The default hostname of the Pyro server.",
     )
+    pyro_client_comm_timeout: int = Field(
+        default=15,
+        description="The default timeout for Pyro client connections.",
```
ntindle: nit: add units here

ntindle (dismissed):

autogpt_platform/backend/backend/util/settings.py
```diff
@@ -69,6 +69,14 @@ class Config(UpdateTrackingModel["Config"], BaseSettings):
...
+    )
+    pyro_client_comm_retry: int = Field(
```
aarushik93, line 76: These need to be added to helm charts and to .env.examples and docker compose

majdyz: similar to the other confs, this is configurable but doesn't need to be configured right away. setting conf already has a default value.

autogpt_platform/backend/backend/util/settings.py
majdyz: Good point, added unit in description.

autogpt_platform/backend/backend/util/settings.py
aarushik93: I understand, in that case please put them in the correct places just with the default values set. Missing configurations has consistently caused various issues and I would like us to be consistent with adding configuration to our docker and k8s set ups at the very least

majdyz: My point is that settings.py default value is the right place for placing a default value and it's used in all of the environments.

Default configurations are excessively replicated and cause misconfiguration, I raised this issue so that we can use the lowest level config as the place to put the default value and place the default value in the other place only if it's strictly necessary: https://github.com/Significant-Gravitas/AutoGPT/issues/8579

aarushik93: It causes confusion when they're not being added to any of our k8s environments. You can think of the k8s part as a way to also see and know what is and has been set in our k8s environment easily without digging through code.

And if you just add them to a .env.example in the backend then they're clear for everyone as well. The de-duplication will happen when we use the .env variables in the docker compose as well. 

Having them in the docker compose  in one way also makes it a lot easier for anyone (which is generally me atm) that needs to have a comparable environment to the k8s environment. 

If you don't want to put them in .env files - at the least do the docker compose and k8s ones - it's a minor addition, and as far as I know most of the team is not on docker compose so it doesn't affect them either

And the config.default.json should go anyways, it's a remnant from local install

majdyz: Added here: https://github.com/Significant-Gravitas/AutoGPT/pull/8574/files#diff-9717c507b6d4ac9429ed967cc2f0840920661e44b5609f1e7097ea0ec2a39b03R114-R115

aarushik93 (approved):

PR #8576, opened 2024-11-06
feat(platform): Add settings page

Bentlybro:
### Background

This is to add the settings page, it will end up using the new nav bar + side bar that i updated in https://github.com/Significant-Gravitas/AutoGPT/pull/8518 once that has been added 😄 

### Changes 🏗️
Added SettingsInputForm.tsx
Added SettingsPage.tsx

image


ntindle:
It is weird to me that the page is in the agptui folder

Bentlybro:
> It is weird to me that the page is in the agptui folder

that's where we are placing all the new ui screens for now while we work on them @ntindle

PR #8593, opened 2024-11-08
feat(backend): Add API key DB table

aarushik93:
### Background

We will be adding API level access, as such need to have access to API keys for users. This PR migrates the DB

### Changes 🏗️

- Introduces `APIKey` model for managing API access credentials
- Implements granular permissions system with 4 core permissions:
  - `EXECUTE_GRAPH`: Run agent graphs
  - `READ_GRAPH`: Access graph information and versions
  - `EXECUTE_BLOCK`: Run individual blocks
  - `READ_BLOCK`: Access block information
- Adds API key status tracking (ACTIVE/REVOKED/SUSPENDED)
- Adds relates migrations

- Upload agent to marketplace

- Changing ports


autogpt_platform/backend/schema.prisma
```diff
@@ -277,3 +278,41 @@ model UserBlockCredit {
 
   @@id(name: "creditTransactionIdentifier", [transactionKey, userId])
 }
+
+enum APIKeyPermission {
```
aarushik93, line 282: Keeping scope of project to just graph and block level for now

autogpt_platform/backend/schema.prisma
```diff
@@ -277,3 +278,41 @@ model UserBlockCredit {
...
+  EXECUTE_GRAPH // Can execute agent graphs
+  READ_GRAPH // Can get graph versions and details
+  EXECUTE_BLOCK // Can execute individual blocks
+  READ_BLOCK // Can get block information
+}
+
+model APIKey {
+  id          String             @id @default(uuid())
+  name        String
+  prefix      String             @unique // First 8 chars for identification
```
Pwuts: what's `prefix`?

autogpt_platform/backend/schema.prisma
```diff
@@ -277,3 +278,41 @@ model UserBlockCredit {
...
+  key         String             @unique // Hashed key
```
Pwuts, line 293: hashed or encrypted?

Pwuts, line 294: If we're hashing the API key, do we need a column to store the salt? Or is it included in this column as a suffix?

autogpt_platform/backend/schema.prisma
Pwuts: OH I get it. Maybe it's my english but I find `prefix` a slightly confusing term here. E.g. I'd expect the prefix to be something like the `sk-proj-` in OpenAI API keys or `sk-ant-api03-` for Anthropic.

Maybe call it `preview`? Also, I don't think we can require this column to be unique as we're only using a small part of the actual key.

aarushik93: I hear your point about prefix potentially being confusing, I think preview might actually be less precise here. The term preview suggests we're showing a representative sample of the key, which could be any part of it (middle, end, etc). But we're specifically storing the beginning/leftmost portion, which is technically a prefix. 


On uniqueness, sure I can remove that

ntindle: You could also add a postfix too if you want

autogpt_platform/backend/schema.prisma
aarushik93: For API keys, typically don't need a salt because:

An API key is already:

Randomly generated
Unique
High entropy (using token_urlsafe)
Long enough to be secure

unlike passwords, where you would want a salt

ntindle: https://auth0.com/blog/adding-salt-to-hashing-a-better-way-to-store-passwords/

A salt has little benefit here imo as if the keys themselves are high entropy it's unlikely they are in hash and rainbow tables. If this is a concern, yes it would need a column but beware that you need to pick an algorithm that allows for len(salt) + len(key) to be within its range. [Bcrypt](https://security.stackexchange.com/questions/39849/does-bcrypt-have-a-maximum-password-length) for example has a max of 72 bytes (keep in mind 💩is 4 bytes when counting) so that will need checked too.

autogpt_platform/backend/schema.prisma
aarushik93: >  You could also add a postfix too if you want

what usecase are you thinking for postfix as well as prefix?

ntindle: Showing apikey preview as form agpt-dhehbrkwowj*********ahhekso

See also this when choosing format
https://docs.github.com/en/code-security/secret-scanning/secret-scanning-partnership-program/secret-scanning-partner-program

aarushik93: oh cool -thanks

aarushik93: Added postfix

ntindle (approved):

PR #8596, opened 2024-11-08
fix(backend): Fix iteration and timer block

majdyz:
### Background

Improvement on the Step Through Items & Timer block.

### Changes 🏗️

* Avoid mixing data types for items input on Step Through Items block.
* Make input pins on CountDownTimer block non-advanced.

- Upload agent to marketplace

- Changing ports


ntindle (dismissed):
I agree with the changes here, but we need to figure out migration routes for these type of changes in general. Can you check the marketplace for these being used? if not problematic, feel free to merge

majdyz:
@ntindle good point, I don't see it being used at all in the marketplace, but I can see some historical graphs using it. So I reverted the breaking change.

Torantulino:
Is this compatible with loading graphs that have the old version of the block in it now?
Since you've (just) added inputs I wonder whether it will break the graphs or not.

Nick is right that we need to not break blocks until we implement block versioning.

majdyz:
The added input is not mandatory, so it's not going to break the old block

majdyz:
The only case that it might break is if there is a graph that uses `item` input but as str not as list

majdyz:
Superseded by: https://github.com/Significant-Gravitas/AutoGPT/pull/8713

Issue #8585, opened 2024-11-08
Update DB schema for API Key support


aarushik93:
Add permissions Enum, API key table with hashed key, add FK relationship between user and api keys<br><br>Draft:

```
enum APIKeyPermission {
  EXECUTE_GRAPH      // Can execute agent graphs
  READ_GRAPH        // Can get graph versions and details
  EXECUTE_BLOCK     // Can execute individual blocks
  READ_BLOCK        // Can get block information
}

model APIKey {
  id          String             @id @default(uuid())
  name        String    
  prefix      String             @unique // First 8 chars for identification
  key         String             @unique // Hashed key
  status      APIKeyStatus       @default(ACTIVE)
  permissions APIKeyPermission[] 
  
  // Core tracking fields
  createdAt   DateTime           @default(now())
  lastUsedAt  DateTime?
  revokedAt   DateTime?
  
  
  // Basic metadata
  description String?
  
  // Relation to user
  userId      String
  user        User               @relation(fields: [userId], references: [id], onDelete: Cascade)

  @@index([key])
  @@index([prefix]) 
  @@index([userId])
  @@index([status])
  @@index([userId, status])
}

enum APIKeyStatus {
  ACTIVE
  REVOKED
  SUSPENDED
}
```

PR #8597, opened 2024-11-08
feat(platform): Add api generator functions and endpoints

### LINKED ISSUE #8589: API key generator (backend)
Create a function and related endpoint to actually generate and store API key (hashed).

Abhi1992002:
### Fix #8589 

This issue depends on #8585 

---

These changes make it easy to create and manage API keys, so we can control who gets access and what they can do. It’s like giving out special keys that we can turn on or off anytime, keeping things secure and organized.

### Changes 🏗️

- Created `/api-keys` endpoint to generate a new API key with custom permissions.

- Added `generate_api_key` function to handle API key generation and hashing.

- Built `validate_api_key` function to check if a provided API key is valid.

- Created `DELETE /api-keys/{key_id}` endpoint to revoke an API key.

- Made `POST /api-keys/{key_id}/suspend` and `POST /api-keys/{key_id}/reactivate` endpoints to suspend/reactivate keys.

- Added `list_user_api_keys` function to list all API keys for a user, ordered by creation date.

- Created `update_api_key_permissions` function to modify an existing key’s permissions.

- Defined `request/response models (CreateAPIKeyRequest, CreateAPIKeyResponse)` for API key handling.

- Added permission validation with `has_permission` function to check if a key has the required access.

---

Depends on this schema

```
enum APIKeyPermission {
	EXECUTE_GRAPH // Can execute agent graphs
	READ_GRAPH // Can get graph versions and details
	EXECUTE_BLOCK // Can execute individual blocks
	READ_BLOCK // Can get block information
}

model APIKey {
	id          String             @id @default(uuid())
	name        String
	prefix      String             @unique // First 8 chars for identification
	key         String             @unique // Hashed key
	status      APIKeyStatus       @default(ACTIVE)
	permissions APIKeyPermission[]

	// Core tracking fields
	createdAt  DateTime  @default(now())
	lastUsedAt DateTime?
	revokedAt  DateTime?

	// Basic metadata
	description String?

	// Relation to user
	userId String
	user   User   @relation(fields: [userId], references: [id], onDelete: Cascade)

	@@index([key])
	@@index([prefix])
	@@index([userId])
	@@index([status])
	@@index([userId, status])
}

enum APIKeyStatus {
	ACTIVE
	REVOKED
	SUSPENDED
}
```


autogpt_platform/backend/backend/data/api.py
```diff
@@ -0,0 +1,209 @@
...
+        where={
+            "id": key_id,
+            "userId": user_id
+        },
+        data={"status": APIKeyStatus.SUSPENDED.value}
+    )
+
+    return APIKeyWithoutHash.from_db(api_key)
+
+async def reactivate_api_key(key_id: str, user_id: str) -> APIKeyWithoutHash:
```
aarushik93: We probably don't want to reactivate a key once it's been suspended or revoked

autogpt_platform/backend/backend/util/api_key.py
```diff
@@ -0,0 +1,43 @@
+import secrets
```
aarushik93: Have you seen the autogpt_libs? I think the APIkey helper code can live there

aarushik93 (changes requested):
@Abhi1992002 thanks for the speedy work as usual

I have a few notes:

We don't need a salt or bcrypt as that's needed for passwords, not for API keys which should already be unique and high entropy strings. So we can use something like token_urlsafe e.g.

secrets.token_urlsafe(32)

for hashing: sha256

We also need a prefix pattern: agpt_<random-string> 

Lastly, in my last PR, I also added a postfix, not sure if you've seen that, would you be able to incorporate that in as well, please

autogpt_platform/autogpt_libs/autogpt_libs/api_key/key_manager.py
```diff
@@ -0,0 +1,44 @@
+import secrets
+import string
+import hmac
+
+class APIKeyManager:
```
aarushik93: thank you for making changes so quickly. 

I think maybe I didn't explain myself clearly enough on the API key generation.

We want to use token_urlsafe to generate the random key: https://docs.python.org/3/library/secrets.html#secrets.token_urlsafe

then we can use hashlib.sha256 to hash the key (and we never store or keep the raw key again). So for example, this class would become something like this: 


```
from typing import NamedTuple
import secrets
import hashlib

class APIKey(NamedTuple):
    """Container for API key parts."""
    raw: str       # Raw key to return to user once
    prefix: str    # For database lookups
    postfix: str   # For display in UI
    hash: str      # For storage and verification

class APIKeyManager:
    PREFIX: str = "agpt_"
    PREFIX_LENGTH: int = 8
    POSTFIX_LENGTH: int = 8

    def generate_api_key(self) -> APIKey:
        """Generate a new API key with all its parts."""
        raw_key = f"{self.PREFIX}{secrets.token_urlsafe(32)}"
        return APIKey(
            raw=raw_key,
            prefix=raw_key[:self.PREFIX_LENGTH],
            postfix=raw_key[-self.POSTFIX_LENGTH:],
            hash=hashlib.sha256(raw_key.encode()).hexdigest()
        )

    def verify_api_key(self, provided_key: str, stored_hash: str) -> bool:
        """Verify if a provided API key matches the stored hash."""
        if not provided_key.startswith(self.PREFIX):
            return False
        return hashlib.sha256(provided_key.encode()).hexdigest() == stored_hash

File: autogpt_platform/backend/backend/server/routers/v1.py [hunk 4]
```diff
@@ -521,3 +523,110 @@ async def update_configuration(
         }
     except Exception as e:
         raise HTTPException(status_code=400, detail=str(e))
+
+
+########################################################
+##################### API ##############################
+########################################################
+
+class CreateAPIKeyRequest(BaseModel):
```
aarushik93: These can be moved into model.py

autogpt_platform/backend/backend/data/api.py
```diff
@@ -0,0 +1,203 @@
+import uuid
```
aarushik93: Can we rename the file api_key

autogpt_platform/backend/backend/server/routers/v1.py
```diff
@@ -521,3 +523,110 @@ async def update_configuration(
         }
     except Exception as e:
         raise HTTPException(status_code=400, detail=str(e))
+
+
+########################################################
+##################### API ##############################
```
aarushik93: nit: API KEY

aarushik93 (commented):
Looks good, if you can just address a few minor comments and fix the linting then we can merge please

autogpt_platform/backend/backend/server/routers/v1.py
```diff
@@ -521,3 +523,110 @@ async def update_configuration(
...
+class UpdatePermissionsRequest(BaseModel):
+    permissions: List[APIKeyPermission]
+
+async def update_permissions(
+    key_id: str,
+    request: UpdatePermissionsRequest,
+    user_id: Annotated[str, Depends(get_user_id)]
+) -> APIKeyWithoutHash:
+    """Update API key permissions"""
+    return await update_api_key_permissions(key_id, user_id, request.permissions)
```
ntindle: I think this decorator got detached from its home

PR #8630, opened 2024-11-12
Add provision for other languages in Youtube Video Block

### LINKED ISSUE #8422: Transcribe YouTube Video Block should automatically fallback to other languages if `en` is not available
The [Transcribe YouTube Video Block ](https://github.com/Significant-Gravitas/AutoGPT/blob/e908068f22eb78f525656bd14f6846ef92e540f8/autogpt_platform/backend/backend/blocks/youtube.py#L63)currently only checks for English by default, but this causes the block to fail if another language is present but not `en` (even US English!).<br>To fix this, we need to do the following (Warning Sudo code):

```python
# Get the list of available transcripts
transcript_list = YouTubeTranscriptApi.list_transcripts(video_id)

# CHECK IF transcript_list is empty, if it is then return a clear error

# Get the first available transcript in its respective language
first_transcript = transcript_list.find_transcript([transcript_list[0].language])

# Get the default transcript for the video
transcript = YouTubeTranscriptApi.get_transcript(video_id, languages=[first_transcript.language])
```

This will fix a common error on this block.

vishesh10:
### Background

Youtube Video Block could only transcribe videos if the language was "en". This PR enhances the capability of the block for other languages as well.
- Resolves #8422 

### Changes 🏗️
* Implemented the changes in the `youtube.py` script to return one of the available transcripts. An exception is thrown If a transcript is not available.

### Test
* Happy path

Screenshot 2024-11-12 at 11 08 19 PM

* Transcript is not available

Screenshot 2024-11-12 at 11 05 34 PM


aarushik93:
Just a comment on the expected [behaviour](https://github.com/Significant-Gravitas/AutoGPT/issues/8422) @Torantulino is this the expected behaviour in this? Shouldn't it be to choose the language? Seems odd to me that it would just choose the first available language?

Torantulino:
Thanks for taking this on @vishesh10!

@aarushik93 this resolves a known issue with the transcriber block that will currently fail to get a transcription even if there is a `us-en` transcript available and not one exactly named `en`. As to whether we should defaultly pick the first available transcription, that's a good question, I don't know how the application orders them. I'd assume that the first transcription available would be the native language of the video.

To default to English you can first attempt a call to: `YouTubeTranscriptApi.get_transcript(video_id)`
> Note: By default, this will try to access the English transcript of the video. 

There is other improvements that could be made here such as preferring manually created transcriptions over Google's auto-generated ones. That can be done via:
`transcript.is_generated`

All in all though this is an undeniable improvement over the existing functionality and a step in the right direction.

Torantulino (approved):

vishesh10:
Thanks for considering the PR @Torantulino @aarushik93.

Per documentation the `find_transcript` method returns the manually created transcripts first. If none are found, generated transcripts are returned.

PR #8653, opened 2024-11-14
feat(blocks): Add Open Router integration credentials

Torantulino:
- Added support for Open Router integration credentials in the Supabase integration credentials store.
- Updated the LLM provider field to include "open_router" as a valid provider option.
- Added Open Router API key field to the backend settings.
- Updated the profile page to display the Open Router integration credentials.
- Updated the credentials input and provider components to include Open Router as a provider option.
- Updated the autogpt-server-api types to include "open_router" as a provider name.
- Updated the LLM provider schema to include "open_router" as a valid provider name.

- Added GEMINI_FLASH_1_5_8B as the first Open Router LLM


ntindle:
Code on this looks good and like it touches all the relevant places. Needs testing still on my side before approval

autogpt_platform/backend/.env.example
```diff
@@ -57,6 +57,7 @@ GOOGLE_CLIENT_SECRET=
 OPENAI_API_KEY=
 ANTHROPIC_API_KEY=
 GROQ_API_KEY=
+OPEN_ROUTER_API_KEY=
```
aarushik93, line 61: is this one we plan to offer by default?

Torantulino: Yep! I've shared the credentials with you in the usual way.

Torantulino:
Added perplexity here too :) Very useful for live info Q/A

classic/forge/forge/llm/providers/schema.py
```diff
@@ -55,6 +55,7 @@ class ModelProviderName(str, enum.Enum):
     OPENAI = "openai"
     ANTHROPIC = "anthropic"
     GROQ = "groq"
+    OPEN_ROUTER = "open_router"
```
Torantulino:
```suggestion
```

Torantulino: Edited this file by mistake so I've removed it now

autogpt_platform/backend/backend/blocks/llm.py
```diff
@@ -354,6 +377,34 @@ def llm_call(
...
             )
+        elif provider == "open_router":
+            client = openai.OpenAI(
+                base_url="https://openrouter.ai/api/v1",
+                api_key=credentials.api_key.get_secret_value(),
+            )
+
+            response = client.chat.completions.create(
+                extra_headers={
+                    "HTTP-Referer": "https://agpt.co",
```
aarushik93, line 388: Do we want these always being agpt/AutoGPT, even for self host?

aarushik93 (approved):
One small question on extra headers, other than that let's go!

autogpt_platform/backend/backend/blocks/llm.py
Torantulino: Yeah, it's just for a usage based leaderboard for our application on their website so the more "points" the better, and there's no expectation to distinguish them.

PR #8710, opened 2024-11-19
fix(backend): Re-work the connection input consumption logic for Agent Executor Block

majdyz:
Agent Executor Block puts all the data within the `data` field of the input. And the rest of the input is metadata.
The current implementation only validates the input of block input (which is the metadata), but not validating the `data` value against the schema of the `graph`. 

This makes the graph always executed once there is a trigger input, without waiting the inbound connection data and uses the default value.

Issue Repro:

* Create an agent with > 1 input, and set the default value on each.
* Re-use that agent as a block, and provide the input value from another node as opposed to the default value.
* Run the agent.
* Before this change: one of the inputs will use the default value instead of a provided node from the inbound links.

### Changes 🏗️

Add a graph input schema validation to the graph to force the execution to wait for all the connecting input fields are completed and provided.

image

#### For code changes:
- [ ] I have made a test plan
  
  - [ ] ...

<details>
  
</details>

<details>

  - Changing ports
</details>


autogpt_platform/backend/backend/executor/manager.py
```diff
@@ -376,31 +385,46 @@ def validate_exec(
...
+            exec_data = AgentExecutorBlock.Input(**node.input_default)
+        except Exception as e:
+            return None, f"Input data doesn't match {node_block.name}: {str(e)}"
+
+        # Validation input
+        input_schema = exec_data.input_schema
+        required_fields = set(input_schema["required"])
+        input_default = exec_data.data
+    else:
+        # Convert non-matching data types to the expected input schema.
```
majdyz, line 400: This is the original flow, the branch above is the special handling for AgentExecutorBlock.
I'm only decoupling the code from below.

Torantulino:
Working on my end!

autogpt_platform/backend/backend/util/json.py
```diff
@@ -30,3 +31,15 @@ def loads(data: str, *args, target_type: Type[T] | None = None, **kwargs) -> Any
     if target_type:
         return type_match(parsed, target_type)
     return parsed
+
+
+def validate_jsonschema(schema: dict[str, Any], data: dict[str, Any]) -> str | None:
```
Pwuts:
```suggestion
def validate(data: dict[str, Any], schema: dict[str, Any]) -> str | None:
```
or
```suggestion
def validate_with_schema(schema: dict[str, Any], data: dict[str, Any]) -> str | None:
```

since it doesn't validate the schema, it validates the data.

autogpt_platform/backend/backend/executor/manager.py
```diff
@@ -376,31 +385,46 @@ def validate_exec(
...
-    error_prefix = f"Input data missing for {node_block.name}:"
+    if isinstance(node_block, AgentExecutorBlock):
+        # Validate the execution metadata for the agent executor block.
+        try:
+            exec_data = AgentExecutorBlock.Input(**node.input_default)
+        except Exception as e:
+            return None, f"Input data doesn't match {node_block.name}: {str(e)}"
+
+        # Validation input
+        input_schema = exec_data.input_schema
```
Pwuts, line 396: Nit: Why is the schema being stored in the block's input data rather than deriving it at runtime? Now we need extra logic for handling untyped schemas instead of `BlockSchema`.

Pwuts (dismissed):

autogpt_platform/backend/backend/executor/manager.py
majdyz: Because the schema is not coming from the block but from the graph

majdyz: Deriving it from runtime works, but it requires loading a specific version of the graph and iterates its nodes. And it doesn't solved the issue of storing untyped schema.

autogpt_platform/backend/backend/util/json.py
majdyz: Updated

majdyz:
@Pwuts addressed your comment, pls re-review 🙏

Pwuts (approved):

PR #8736, opened 2024-11-21
fix(backend): Fix .env file read contention on pyro connection setup

majdyz:
On high concurrent pyro connections initialization, the .env file being on the high read contention causing connection establishment to fail.

Calling `Secrets()`, `Config()`, and `Settings()`, will always trigger a file read. And when this is executed too many times it will cause read contention and pass the file descriptor limit set by the OS. This will trigger retry and causes delay & thundering herd issues. In Unix this is errored out as:
* `Too Many Open Files` or
* `There are only ... file descriptors (hard limit) available`

To reproduce the error you can try to do the run locally and run hundreds of concurrent requests.

### Changes 🏗️

Move all the usage of Config as a globally initialized variable so it's only called once per process.

#### For code changes:
- [ ] I have made a test plan
  
  - [ ] ...

<details>
  
</details>

<details>

  - Changing ports
</details>


majdyz:
@aarushik93 Described in the PR like this:
```
On high concurrent pyro connections initialization, the .env file being on the high read contention causing connection establishment to fail.

Calling `Secrets()`, `Config()`, and `Settings()`, will always trigger a file read. And when this is executed too many times it will cause read contention and pass the file descriptor limit set by the OS. This will trigger retry and causes delay & thundering herd issues. In Unix this is errored out as:
* `Too Many Open Files` or
* `There are only ... file descriptors (hard limit) available`
```

Let me know if there is any part that needs to be clarified.

aarushik93:
Gotcha, thanks that clears it up for me!

aarushik93 (approved):

PR #8749, opened 2024-11-24
feat(blocks): Various block QoL improvements

ntindle:
There's a few QoL things I found when reviewing the webhooks PR for the github blocks
- ~~There's not a good way to get the patch out of the GitHub PR Trigger block~~
- The GitHub comment block doesn't work with pull requests
- the GitHub API wrapper always force redirects the API to the `api.github` endpoint
- The ai text block doesn't tell you how to use it to fill fields

### Changes 🏗️

Fixes/Changes the expected behaviors in a non breaking way for the above issues

#### For code changes:
- [x] I have made a test plan

Test plan is simply an agent using the changes
image
[pr trigger testat_v37.json](https://github.com/user-attachments/files/17892533/pr.trigger.testat_v37.json): Note that this agent is tested on edit for a pr so I wasn't constantly recreating PRs on other repos

https://github.com/ntindle/gridfinity-space-optimizer/pull/2


autogpt_platform/backend/backend/blocks/github/triggers.py
```diff
@@ -145,8 +148,18 @@ def __init__(self):
...
+    ) -> BlockOutput:
+        api = get_api(credentials, convert_urls=False)
         yield from super().run(input_data, **kwargs)
+        if input_data.payload["pull_request"]["patch_url"]:
+            response = api.get(
+                input_data.payload["pull_request"]["patch_url"], allow_redirects=True
+            )
+            yield "patch", response.text
+        else:
+            yield "patch", ""
```
Pwuts: AFAICS this is redundant with the `GitHubReadPullRequestBlock`, which has an output for this. Since diffs can be rather large (and take a while to fetch?? not sure), I'd prefer not fetching it by default.

ntindle: thoughts on a checkbox?

ntindle: but also for me they're pretty quick and getting the diff on a pr when you're notified of it seems pretty reasonable to me

Pwuts: Checkbox could be good. Still, there are multiple ways to represent a diff, so I think it deserves its own block rather than stuffing *a version* of it in this block and `GitHubReadPullRequestBlock`. Modularity FTW.

ntindle: removed. I still think this is a reasonable behavior to want but don't think its a good experience without dynamic output pins

Pwuts (approved):

PR #8752, opened 2024-11-25
feat(platform): Include all agent versions in Runs in Monitor

kcze:
The graph version is bumped on each save. While the agent version is changed, the past execution history is gone because the monitor page only shows the latest version's execution history.

### Changes 🏗️

- Add `get_executions` on the backend that returns all executions of all graphs for a user
- Display all executions (for all versions) for graphs in Monitor
- Rename ts mirror type `ExecutionMeta` to `GraphExecution` for consistency with the backend
- Remove redundant `FlowRun` type on the frontend and use `GraphExecution` instead
- Round execution duration text in Monitor to one decimal place

#### For code changes:
- [ ] I have made a test plan
  
  - [ ] ...

<details>
  
</details>

<details>

  - Changing ports
</details>


autogpt_platform/frontend/src/app/page.tsx
kcze, line 1: Combine executions on the frontend because the graph version is lost when done on the backend; there's no graph information in the `backend/data/graph.py:GraphExecution` model.

autogpt_platform/backend/backend/data/graph.py
```diff
@@ -353,38 +353,46 @@ async def get_node(node_id: str) -> Node:
...
     """
-    where_clause: AgentGraphWhereInput = {}
+    where_clause: AgentGraphWhereInput = {"userId": user_id}
 
     if filter_by == "active":
         where_clause["isActive"] = True
     elif filter_by == "template":
         where_clause["isTemplate"] = True
 
-    where_clause["userId"] = user_id
```
Pwuts, line 377: ?

autogpt_platform/backend/backend/data/graph.py
```diff
@@ -353,38 +353,46 @@ async def get_node(node_id: str) -> Node:
...
+            order={"version": "desc"},
+            include=graph_include,
+        )
+    else:
+        graphs = await AgentGraph.prisma().find_many(
+            where=where_clause,
+            distinct=["id"],
+            order={"version": "desc"},
+            include=graph_include,
+        )
```
Pwuts:
```suggestion
    graphs = await AgentGraph.prisma().find_many(
        where=where_clause,
        distinct=["id"] if filter_by else None,
        order={"version": "desc"},
        include=graph_include,
    )
```

autogpt_platform/backend/backend/server/routers/v1.py
```diff
@@ -141,9 +141,10 @@ class DeleteGraphResponse(TypedDict):
 async def get_graphs(
     user_id: Annotated[str, Depends(get_user_id)],
     with_runs: bool = False,
+    filter_by: Literal["all", "active", "template"] | None = "active",
 ) -> list[graph_db.Graph]:
     return await graph_db.get_graphs(
-        include_executions=with_runs, filter_by="active", user_id=user_id
+        include_executions=with_runs, filter_by=filter_by, user_id=user_id
```
Pwuts:
```suggestion
) -> list[graph_db.Graph]:
    return await graph_db.get_graphs(
        include_executions=with_runs, filter_by="active" if not with_runs else None, user_id=user_id

File: autogpt_platform/frontend/src/app/page.tsx
Comment (Pwuts) [reply autogpt_platform/frontend/src/app/page.tsx:1]: The `AgentGraphExecution` database model has `agentGraphId` and `agentGraphVersion` fields, so if you just propagate those to the `GraphExecution` model you can do the combination on the backend.

**Review by Pwuts — COMMENTED** [review]
Small fix but big UX improvement. Thanks for picking this up :)

File: autogpt_platform/backend/backend/data/graph.py [hunk 4]
```diff
@@ -353,38 +353,46 @@ async def get_node(node_id: str) -> Node:
 async def get_graphs(
     user_id: str,
     include_executions: bool = False,
-    filter_by: Literal["active", "template"] | None = "active",
+    filter_by: Literal["all", "active", "template"] | None = "active",
 ) -> list[Graph]:
     """
     Retrieves graph metadata objects.
     Default behaviour is to get all currently active graphs.
+    When filter_by="all", retrieves all versions of graphs.
```
majdyz: I'm pretty sure we have bajillions of graphs without an execution because of the way we save the graph. Should we have a filter to fetch all graphs with at least a single execution instead? Or even better, the fetch is done using `get_executions` (which fetches all graph executions for a user, I think we don't have it yet, but it should be easy to add) instead of `get_graph` and the graph we are fetching is still only the active ones.

kcze: Thanks! I've changed my approach and reverted some code.
1. Now Monitor fetches in two calls:
- Graphs metadata (with no executions or nodes/links)
- All executions for a user
2. We still want to fetch all graphs, even without executions, so the user can see and edit it.
3. Now `get_graphs` with runs isn't needed (can remove if you want).
4. It would be better to fetch only the executions for the graph currently selected but graph table needs execution count and there is no way to get it other than getting/listing all executions.

autogpt_platform/backend/backend/data/graph.py
kcze: No longer relevant, see https://github.com/Significant-Gravitas/AutoGPT/pull/8752#discussion_r1878297781

autogpt_platform/backend/backend/server/routers/v1.py
kcze: No longer relevant

autogpt_platform/frontend/src/app/page.tsx
kcze: I've included graph id and version in the `GraphExecution`, there's no combining needed now

autogpt_platform/backend/backend/data/graph.py
```diff
@@ -445,6 +447,17 @@ async def get_graphs(
     return [GraphModel.from_db(graph) for graph in graphs]
 
 
+async def get_executions(user_id: str) -> list[GraphExecution]:
```
majdyz, line 450: nit: Should we place this in execution.py instead?

majdyz (approved):
Thanks! this is really cool!

autogpt_platform/backend/backend/data/graph.py
```diff
@@ -445,6 +447,17 @@ async def get_graphs(
...
+    where_clause: AgentGraphExecutionWhereInput = {"userId": user_id}
+
+    executions = await AgentGraphExecution.prisma().find_many(
+        where=where_clause,
+        order={"createdAt": "desc"},
+    )
+
+    return [GraphExecution.from_db(execution) for execution in executions]
```
majdyz, line 458: do we have `from_db` here ?

autogpt_platform/backend/backend/data/graph.py
majdyz: Hmmm I just found a new rabbit hole, there are two GraphExecution models one in graph.py & execution.py 😢 
I think this calls for another follow-up cleanup unrelated to this PR.

No need to change it, feel free to resolve this and merge.

kcze: There were a lot of other things I was tempted to update but the Monitor is getting refreshed anyway so I stopped myself

majdyz: I'm doing a followup cleanup here:
https://github.com/Significant-Gravitas/AutoGPT/pull/8939

So you can add your update there too if you want.

autogpt_platform/frontend/src/lib/autogpt-server-api/types.ts
```diff
@@ -188,17 +188,19 @@ export type LinkCreatable = Omit<Link, "id" | "is_static"> & {
   id?: string;
 };
 
-/* Mirror of autogpt_server/data/graph.py:ExecutionMeta */
-export type ExecutionMeta = {
+/* Mirror of backend/data/graph.py:GraphExecution */
+export type GraphExecution = {
   execution_id: string;
   started_at: number;
   ended_at: number;
```
Pwuts, line 195: `started_at` and `ended_at` are not unix timestamps on `GraphExecution` but ISO-formatted timestamps!

Pwuts: This turns out to have broken the `FlowRunsTimeline`, maybe more

PR #8802, opened 2024-11-27
fix(frontend): advanced-toggle-default

### LINKED ISSUE #8739: Hide "advanced" toggle on blocks if there are no advanced inputs
The toggle is shown but does nothing if the block doesn't have any advanced inputs.

Abhi1992002:
- resolve #8739

I don't think so that this is a frontend issue [might be wrong] , because if we are not classifying that a particular input is `advanced = true/false`. Then we automatically get `advanced = True`.

Screenshot 2024-11-27 at 10 36 59 AM

### DIFF
--- a/autogpt_platform/backend/backend/data/model.py
+++ b/autogpt_platform/backend/backend/data/model.py
@@ -110,7 +110,7 @@ def SchemaField(
     title: Optional[str] = None,
     description: Optional[str] = None,
     placeholder: Optional[str] = None,
-    advanced: Optional[bool] = None,
+    advanced: Optional[bool] = False,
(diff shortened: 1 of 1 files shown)


Pwuts:
Could you check why the toggle isn't working on the `GithubPullRequestTriggerBlock` then?

Abhi1992002:
> If it works for other blocks, then could you check why the toggle isn't working on the `GithubPullRequestTriggerBlock`?

To make it work, you need to explicitly set `advanced = False` in both the event and the trigger.

aarushik93 (approved):

PR #8803, opened 2024-11-27
fix(frontend): Add integer type definition for node handles

### LINKED ISSUE #8740: Add `integer` to `NodeHandle` type list
The JSON schema type `integer` is not defined in the type list in `<NodeHandle>`, causing it to show up as `(any)` rather than `(integer)` on block inputs/outputs with that type.

[https://github.com/Significant-Gravitas/AutoGPT/blob/86535b5811f8d1cc0bdde2232693919c4b1115e3/autogpt_platform/frontend/src/components/NodeHandle.tsx#L22-L29](https://github.com/Significant-Gravitas/AutoGPT/blob/86535b5811f8d1cc0bdde2232693919c4b1115e3/autogpt_platform/frontend/src/components/NodeHandle.tsx#L22-L29)

Abhi1992002:
- resolve #8740

- Adding the ability to recognize integers to the handle.

Screenshot 2024-11-27 at 11 40 13 AM

### DIFF
--- a/autogpt_platform/frontend/src/components/NodeHandle.tsx
+++ b/autogpt_platform/frontend/src/components/NodeHandle.tsx
@@ -22,6 +22,7 @@ const NodeHandle: FC<HandleProps> = ({
   const typeName: Record<string, string> = {
     string: "text",
     number: "number",
+    integer: "integer",
     boolean: "true/false",
     object: "object",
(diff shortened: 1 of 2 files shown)


Pwuts (approved):
Thanks! LGTM

PR #8805, opened 2024-11-27
feat(blocks): Add Slant 3D printing via API service

ntindle:
I want to be able to have agents 3d print things and deliver them to my house!

### Changes 🏗️

- Adds slant3d as a provider
- Adds slant3d order webhook (disabled on the cloud by default due to how it notifies users)
- Adds several blocks to order from slant3d
- Diables Get Orders (for the same reason as webhook)

#### For code changes:
- [x] I have made a test plan

<details>
  <summary>Test Plan</summary>
  
  - [ ] Add filament block and fill API key
  - [ ] Run filament block
  - [ ] Add slice block and use this value: https://files.printables.com/media/prints/1081287/stls/8176524_a9edde2d-68c1-41de-a207-b584fcf42f30_f9127d5b-39ed-4ef8-b59f-d3a0bc874373/rod-holder.stl
  - [ ] Run slice block
  - [ ] Add estimate blocks (print and shipping) and use your address, and the above file
  - [ ] select petg and count 1
  - [ ] run the blocks
  - [ ] Create an order using same information
  - [ ] Run the block and note the order number
  - [ ] Delete the create order block so you don't keep ordering stuff
  - [ ] Run get orders block
  - [ ] Check your order exists
  - [ ] Run the cancel order block with the order id
  - [ ] run the get orders block and check the order no longer exists
</details>


autogpt_platform/backend/backend/blocks/slantthreed/_api.py
Pwuts, line 1: I think you can just call this module `slant3d` rather than `slantthreed`

autogpt_platform/backend/backend/blocks/slantthreed/base.py
```diff
@@ -0,0 +1,90 @@
...
+        )
+        if profile == Profile.PLA:
+            color_tag = color.value
+        else:
+            color_tag = f"{profile.value.lower()}{color.value.capitalize()}"
+        valid_tags = [filament["colorTag"] for filament in response["filaments"]]
+
+        if color_tag not in valid_tags:
+            raise ValueError(
+                f"Invalid color profile combination {color_tag}. Valid colors for {profile.value} are: {','.join([filament['colorTag'].replace(profile.value.lower(), '') for filament in response['filaments'] if filament['profile'] == profile.value])}"
```
Pwuts: Please split this into multiple lines

autogpt_platform/backend/backend/integrations/webhooks/slantthreed.py
```diff
@@ -0,0 +1,108 @@
...
+            f"{self.BASE_URL}/customer/webhookSubscribe", headers=headers, json=payload
+        )
+
+        if not response.ok:
+            error = response.json().get("error", "Unknown error")
+            raise RuntimeError(f"Failed to register webhook: {error}")
+
+        # Slant3D doesn't return a webhook ID, so we generate one
+        # The actual webhook is identified by its endpoint URL
+        webhook_id = str(__import__("uuid").uuid4())
```
Pwuts: 1. `__import__("uuid").uuid4()` 👎🏼
2. This isn't a structural solution. If there is no real ID the field should be left empty.

autogpt_platform/backend/backend/integrations/webhooks/slantthreed.py
```diff
@@ -0,0 +1,108 @@
+from typing import ClassVar, Dict, List, Tuple
```
Pwuts: Pls use `dict`, `list`, `tuple` instead of `Dict`, `List`, `Tuple`

autogpt_platform/backend/backend/integrations/webhooks/slantthreed.py
```diff
@@ -0,0 +1,108 @@
...
+        """
+        # Log warning since we can't properly deregister
+        print(f"Warning: Manual deregistration required for webhook {webhook.id}")
+        pass
+
+    async def trigger_ping(self, webhook: integrations.Webhook) -> None:
+        """
+        Note: Slant3D automatically sends a test payload during registration,
+        so we don't need a separate ping implementation.
+        """
```
Pwuts: 1. `trigger_ping` is used to check a webhook's status. This is implemented in #8838.
2. If triggering a ping event is not supported by the provider, this method should be omitted from the implementation. The base implementation raises `NotImplementedError`. With the current empty override, the status check will fail and report the webhook as broken.

autogpt_platform/backend/backend/integrations/webhooks/slantthreed.py
```diff
@@ -0,0 +1,108 @@
...
+
+    async def _deregister_webhook(
+        self, webhook: integrations.Webhook, credentials: Credentials
+    ) -> None:
+        """
+        Note: Slant3D API currently doesn't provide a deregistration endpoint.
+        This would need to be handled through support.
+        """
+        # Log warning since we can't properly deregister
+        print(f"Warning: Manual deregistration required for webhook {webhook.id}")
```
Pwuts: 1. With this implementation, the webhook just ends up deleted but not deregistered, and the webhook ingress endpoint will return a HTTP 500 (should probably be amended to do 404 if webhook not found). If Slant3D have been smart, they stop trying after X failures and delete the webhook on their end. But I don't know this.
2. If you find the time, please contact Slant3D and ask them to implement this.

autogpt_platform/backend/backend/integrations/webhooks/slantthreed.py
```diff
@@ -0,0 +1,108 @@
...
+        if missing_fields:
+            raise ValueError(f"Missing required fields: {', '.join(missing_fields)}")
+
+        # Normalize payload structure
+        normalized_payload = {
+            "orderId": payload["orderId"],
+            "status": payload["status"],
+            "trackingNumber": payload["trackingNumber"],
+            "carrierCode": payload["carrierCode"],
+        }
```
Pwuts: What else is or can be in the payload?

autogpt_platform/backend/backend/blocks/slantthreed/webhook.py
```diff
@@ -0,0 +1,125 @@
...
+
+        try:
+            # Extract and normalize values from the payload
+            yield "status", input_data.payload["status"]
+            yield "tracking_number", input_data.payload["trackingNumber"]
+            yield "carrier_code", input_data.payload["carrierCode"]
+        except KeyError as e:
+            error_msg = f"Missing required field in webhook payload: {e}"
+            yield "error", error_msg
+            raise RuntimeError(error_msg)
```
Pwuts: Redundant with check in `validate_payload`

autogpt_platform/backend/backend/blocks/slantthreed/webhook.py
```diff
@@ -0,0 +1,125 @@
...
+            """
+            Currently Slant3D only supports 'SHIPPED' status updates
+            Could be expanded in the future with more status types
+            """
+
+            shipped: bool = True
+
+        events: EventsFilter = SchemaField(
+            title="Events", description="Order status events to subscribe to"
+        )
```
Pwuts: You can give this filter a default value and hide it, so it doesn't show up on the block. I'll make a ticket to make the event filter optional as a structural solution.

autogpt_platform/backend/backend/blocks/slantthreed/webhook.py
Pwuts: Update: not redundant if `payload` has default value `{}`. And I assume it has that default value so it doesn't completely break a graph if there are multiple entrypoints. We really need a better solution for that...

autogpt_platform/backend/backend/blocks/slantthreed/_api.py
ntindle: You unfortunately cannot, it won’t load

autogpt_platform/backend/backend/integrations/webhooks/slantthreed.py
aarushik93: can we  make this a log not a print?

autogpt_platform/backend/backend/blocks/slantthreed/webhook.py
Pwuts: Update: #8853 makes `payload = {}` unnecessary.

autogpt_platform/backend/backend/blocks/slantthreed/_api.py
Pwuts: That limitation is self-imposed:
https://github.com/Significant-Gravitas/AutoGPT/blob/25912067f2a3778bc85158eb49f68bb78c7772cd/autogpt_platform/backend/backend/blocks/__init__.py#L17-L22

I'll make a PR to FTFY

Pwuts: - #8857

autogpt_platform/backend/backend/blocks/slantthreed/base.py
ntindle: done

autogpt_platform/backend/backend/integrations/webhooks/slantthreed.py
ntindle: This should be documented better, I'd expect this to be None, but the signature only allows str

autogpt_platform/backend/backend/integrations/webhooks/slantthreed.py
ntindle: nothing?
```
Request Response:
{
    "message": "Endpoint Configured",
    "endPoint": "https://example.com"
}
    
Webhook Example:
{
  "orderId": "1234567890",
  "status": "SHIPPED",
  "trackingNumber": "ABCDEF123456",
  "carrierCode": "usps"
}
```

autogpt_platform/backend/backend/integrations/webhooks/slantthreed.py
ntindle: yes

autogpt_platform/backend/backend/integrations/webhooks/slantthreed.py
ntindle: done

autogpt_platform/backend/backend/integrations/webhooks/slantthreed.py
ntindle: That is unfortunately how it works, but I've requested the change with slant3d

autogpt_platform/backend/backend/integrations/webhooks/slantthreed.py
ntindle: done

autogpt_platform/backend/backend/blocks/slantthreed/webhook.py
ntindle: yeeted the check

ntindle:
image

aarushik93 (approved):

PR #8847, opened 2024-12-01
feat(frontend/feature-flags): Add LaunchDarkly feature flagging UI

aarushik93:
This PR allows us to feature flag on the frontend, this means we can rollout features in stages, hide features, do AB testing etc. 

### Changes 🏗️

Added a LaunchDarkly Provider
Added a withFeatureFlag component
Added two env vars for: 
- enabling LD 
- specifying the _public_ client side key

Usage: 

```
'use client'

import { useFlags } from 'launchdarkly-react-client-sdk'
import { withFeatureFlag } from '@/components/feature-flag/with-feature-flag'

function TestFlagPage() {
  const flags = useFlags()

  return (
    <div className="p-4">
      <h1>If you can see this, the feature flag is ON</h1>
      <pre>Current flag value: {JSON.stringify(flags, null, 2)}</pre>
    </div>
  )
}

export default withFeatureFlag(TestFlagPage, 'test-flag')
```

#### For code changes:
- [x] I have made a test plan
  
  - [ ] ...

<details>
  <summary>Test plan</summary>
- Set LD to false
- Navigate to a test page, should not be visible
- Set LD to true
- Navigate to same test page, should be visible
</details>

- [x] I have updated infra repo

<details>

  - Changing ports
</details>


autogpt_platform/frontend/package.json
```diff
@@ -57,6 +57,8 @@
     "date-fns": "^4.1.0",
     "dotenv": "^16.4.5",
     "elliptic": "6.6.1",
+    "launchdarkly-js-client-sdk": "^3.5.0",
```
Pwuts: This seems unnecessary since it's already a dependency of `launchdarkly-react-client-sdk` :)
```suggestion
```

Pwuts:
Could you update the checklists in the description? Especially relevant since this PR has a config change.

autogpt_platform/frontend/src/components/feature-flag/with-feature-flag.tsx
```diff
@@ -0,0 +1,46 @@
+"use client";
```
Pwuts, line 1: (Why) does this only work on client side?

autogpt_platform/frontend/src/components/feature-flag/with-feature-flag.tsx
```diff
@@ -0,0 +1,46 @@
...
+    }, [hasFlagLoaded, flags, flagKey, router]);
+
+    // Show loading state until flags loaded
+    if (!hasFlagLoaded) {
+      return (
+        <div className="flex min-h-screen items-center justify-center">
+          <div className="h-8 w-8 animate-spin rounded-full border-4 border-primary border-t-transparent" />
+        </div>
+      );
+    }
```
Pwuts, line 36: Is this noticeable to users, in terms of a delay?

aarushik93:
> does this only work for enabling/disabling full pages or also individual components?
rn, just for full pages - we can expand later to individual components too, I'll add a ticket

> The PR description also mentions A/B testing. How would you do that using withFeatureFlags?
We can configure/control on the launch darkly side how many users we show what feature to. We'll probably at some point want to extend to choosing specific users, but haven't had that requirement yet. I'll add docs for the launch darkly side as well on notion so it's clear how to use it.

autogpt_platform/frontend/src/components/feature-flag/with-feature-flag.tsx
aarushik93: This component needs React hooks e.g [use effect](https://github.com/Significant-Gravitas/AutoGPT/pull/8847/files/a42e563aa662b8be17dc535cc05ff76c3f816a42#diff-33a8a6aff2e5117038be99b173a6490aa4a91caa78d8fc98b4177ac5cef48c7aR16).  In particular because we're responding to real time updates from LaunchDarkly (see useFlags). So can't run this server side.

https://nextjs.org/docs/app/building-your-application/rendering/client-components

autogpt_platform/frontend/src/components/feature-flag/with-feature-flag.tsx
aarushik93: no it's miniscule

Issue #8731, opened 2024-11-20
Add Input Dependencies and Exclusive Options to Prevent Conflicting Inputs


Abhi1992002:
We need to implement functionality on both the frontend and backend to handle input dependencies and mutual exclusivity in blocks. This includes disabling dependent inputs when prerequisites are not met and ensuring only one input from a group of mutually exclusive options can be selected.

### Examples 

Suppose we have a block that helps you to create a tweet.

Possible inputs in this could be:

-`text`

-`media_ids`

-`place_ids`

-`quote_ids`

-`dm_deep_link`

-`poll_options`

-`poll_minutes` (must be used with poll_options)

With text, we could only provide one option from the above inputs. If the user provides more than one option, they might get an error. Suppose they have given `poll_options` as well as `media_ids`; they get a 400 bad request.

Now, what do you need to do?

You need to add functionality on both frontend and backend so that:

1. If one input depends upon another input and you fill only one, then you can't run the block; you get a frontend error.

2. You are able to make a way so that if you have 5 inputs and you need to fill only one, once you fill it, all the other inputs are disabled. Or if some user tries to send data (I know he can't after disabled but as an extra security layer), if he sends it then he gets a frontend error.

**My tips that you can consider:**

1. We could add a `depends` field in the input.

```python
poll_options: list = SchemaField(
    description="List of poll options",
    placeholder="Enter poll options",
)

poll_duration_minutes: int = SchemaField(
    description="Duration of the poll in minutes",
    default=1440,  # 24 hours
    depends=[poll_options],
    required=False,
)
```

If you haven't written anything in `poll_options`, then `poll_duration_minutes` is disabled.

2. You could give a field `network` similar to `Docker.` If 3 inputs are on the same network, then you can only select one.

```python
place_id: str | None = SchemaField(
    description="Adds optional location information to a tweet if geo settings are enabled in your profile.",
    default=None,
    required=False,
    network='dumbledore'
)

quote_tweet_id: str | None = SchemaField(
    description="Link to the Tweet being quoted",
    default=None,
    placeholder="e.g., 1455953449422516226",
    required=False,
    network='dumbledore'
)

direct_message_deep_link: str | None = SchemaField(
    description="Link directly to a Direct Message conversation with an account",
    default=None,
    placeholder="e.g., https://twitter.com/messages/compose?recipient_id={your_id}",
    required=False,
    network='dumbledore'
)

```
If multiple inputs are on the same network, then you can only fill one.

[...]

PR #8852, opened 2024-12-02
feat(blocks): Add `depends_on` support for input fields

Abhi1992002:
- Resolves part of #8731 

### Changes
- Added `depends_on` parameter to SchemaField in `model.py` to specify field dependencies.
- Updated `useAgentGraph` hook to validate input fields based on their dependencies, ensuring required fields are set when dependent fields are filled.
- Modified `BlockIOSubSchemaMeta` to include `depends_on` as an optional property.

https://github.com/user-attachments/assets/64fd47b3-34dc-48fa-ad90-1c9c5cd4c4a3


autogpt_platform/backend/backend/blocks/checking_input_validation.py
Pwuts, line 1: pls remove this block before marking the PR as ready for review to make sure we don't merge it

autogpt_platform/frontend/src/hooks/useAgentGraph.ts
```diff
@@ -405,7 +405,34 @@ export default function useAgentGraph(
             setNestedProperty(errors, key, "This field is required");
           }
         });
-      }
+      } 
+      
+      Object.entries(node.data.inputSchema.properties || {}).forEach(([key, schema]) => {
+        if ('depends_on' in schema) {
+          const dependencies = Array.isArray(schema.depends_on) ? schema.depends_on : [schema.depends_on];
```
Pwuts: This check and conversion seem unnecessary because `depends_on` is only allowed to be `string[]` in the data model.

autogpt_platform/frontend/src/hooks/useAgentGraph.ts
```diff
@@ -405,7 +405,34 @@ export default function useAgentGraph(
...
+          
+          // Check if required fields are set but dependent field is empty
+          const hasAllDependencies = dependencies.every(dep => 
+            inputData[dep as keyof typeof inputData] && 
+            String(inputData[dep as keyof typeof inputData]).length > 0
+          );
+          if (hasAllDependencies && (!inputData[key as keyof typeof inputData] || String(inputData[key as keyof typeof inputData]).length === 0)) {
+            errors[key] = `This field is required when ${dependencies.join(', ')} are set`;
+            errorMessage = `${key} is required when ${dependencies.join(', ')} are set`;
+          }
```
Pwuts: `key as keyof typeof inputData`

Is this necessary?

autogpt_platform/frontend/src/hooks/useAgentGraph.ts
```diff
@@ -405,7 +405,34 @@ export default function useAgentGraph(
...
+      Object.entries(node.data.inputSchema.properties || {}).forEach(([key, schema]) => {
+        if ('depends_on' in schema) {
+          const dependencies = Array.isArray(schema.depends_on) ? schema.depends_on : [schema.depends_on];
+          
+          // Check if dependent field is set while required fields are empty
+          const hasValue = inputData[key as keyof typeof inputData] && String(inputData[key as keyof typeof inputData]).length > 0;
+          const missingDependencies = dependencies.filter(dep => !inputData[dep as keyof typeof inputData] || String(inputData[dep as keyof typeof inputData]).length === 0);
+          
+          if (hasValue && missingDependencies.length > 0) {
+            errors[key] = `Requires ${missingDependencies.join(', ')} to be set`;
```
Pwuts: gotta use `setNestedProperty` instead of `errors[key]`

autogpt_platform/frontend/src/hooks/useAgentGraph.ts
```diff
@@ -405,7 +405,34 @@ export default function useAgentGraph(
...
         });
-      }
+      } 
+      
+      Object.entries(node.data.inputSchema.properties || {}).forEach(([key, schema]) => {
+        if ('depends_on' in schema) {
+          const dependencies = Array.isArray(schema.depends_on) ? schema.depends_on : [schema.depends_on];
+          
+          // Check if dependent field is set while required fields are empty
+          const hasValue = inputData[key as keyof typeof inputData] && String(inputData[key as keyof typeof inputData]).length > 0;
```
Pwuts: - Does this take defaults into account?
- This check breaks when the value is `false`. 

```suggestion
          const mustHaveValue = inputSchema.required && key in inputSchema.required;
          const hasValue = schema.default != null || inputData[key] != null;
```

Pwuts (commented):
This is a good feature. Needs some refinement though, see comments. :)

autogpt_platform/frontend/src/hooks/useAgentGraph.ts
```diff
@@ -405,7 +405,34 @@ export default function useAgentGraph(
             setNestedProperty(errors, key, "This field is required");
           }
         });
-      }
+      } 
+      
+      Object.entries(node.data.inputSchema.properties || {}).forEach(([key, schema]) => {
```
Pwuts: Since you're doing this validation on save, it should be done in the backend. Frontend-only validation = vulnerable.

autogpt_platform/frontend/src/hooks/useAgentGraph.ts
Abhi1992002: It’s giving me a TS error, which is why I added it.

autogpt_platform/backend/backend/executor/manager.py
```diff
@@ -421,6 +421,35 @@ def validate_exec(
...
+            if (has_value or must_have_value) and missing_deps:
+                return None, f"{error_prefix} Field {field_name} requires {', '.join(missing_deps)} to be set"
+
+            # Check if field is required when dependencies are present
+            has_all_deps = all(
+                data.get(dep) and str(data[dep]).strip() != ""
+                for dep in dependencies
+            )
+            if has_all_deps and not has_value:
+                return None, f"{error_prefix} {field_name} is required when {', '.join(dependencies)} are set"
```
Pwuts: Could you move this to `GraphModel.validate_graph`?

autogpt_platform/frontend/src/hooks/useAgentGraph.ts
```diff
@@ -406,6 +406,55 @@ export default function useAgentGraph(
...
+      Object.entries(node.data.inputSchema.properties || {}).forEach(
+        ([key, schema]) => {
+          if (schema.depends_on) {
+            const dependencies = schema.depends_on;
+
+            // Check if dependent field has value
+            const hasValue =
+              (inputData[key] != null &&
+                String(inputData[key]).trim() !== "") ||
+              (schema.default != null && String(schema.default).trim() !== "");
```
Pwuts: Please remove this string conversion. The not-null check is enough, right?

Pwuts: You also need to check `"default" in schema` before using `schema.default`:
```suggestion
              inputData[key] != null ||
              ("default" in schema && schema.default != null);
```

Abhi1992002: Sometimes, we set the default value as “”. If we don’t check “”, it will be considered as the value and passes the checks.
If we haven’t given any dependent field a default value, it will be considered as required. (I want my dependent field to be required only if the dependent field is filled.)

Abhi1992002: It’s just an additional check (it poses no harm), and if we modify it, it could potentially break some future block that requires a field to be a string (not an optional string) and non-required with empty value.

Pwuts: > Sometimes, we set the default value as “”. If we don’t check “”, it will be considered as the value and passes the checks.

Can you give an example? If a default value is set, the block must work with that value as input.

autogpt_platform/backend/backend/data/graph.py
```diff
@@ -275,6 +275,54 @@ def sanitize(name):
                         f"Node {block.name} #{node.id} required input missing: `{name}`"
                     )
 
+            # Get input schema properties and check dependencies
+            input_schema = block.input_schema.model_json_schema()
+            properties = input_schema.get("properties", {})
+            required_fields = set(input_schema.get("required", []))
```
majdyz: There is already block.input_schema.get_required_fields()

autogpt_platform/backend/backend/data/graph.py
```diff
@@ -275,6 +275,54 @@ def sanitize(name):
...
+                        or not node.input_default[dep]
+                        or str(node.input_default[dep]).strip() == ""
+                    ]
+
+                    if (has_value or must_have_value) and missing_deps:
+                        raise ValueError(
+                            f"Node {block.name} #{node.id}: Field {field_name} requires {', '.join(missing_deps)} to be set"
+                        )
+
+                        # Check if field is required when dependencies are present
```
majdyz: nit: unexpected space

majdyz: Is this check needed? 

If the field is not required, is it wrong to have all deps defined but not the field depending on those ?
If the field is required, do we need the check in the first place, because if it has no value and the field is required, it will fail the validation check.

autogpt_platform/backend/backend/data/graph.py
```diff
@@ -275,6 +275,54 @@ def sanitize(name):
...
+                    )
+                    must_have_value = field_name in required_fields
+
+                    # Check for missing dependencies when dependent field is present
+                    missing_deps = [
+                        dep
+                        for dep in dependencies
+                        if dep not in node.input_default
+                        or not node.input_default[dep]
+                        or str(node.input_default[dep]).strip() == ""
```
majdyz: This check seems replicated in 3 places, and some of them are missing the default ("default" field) value check.
We can add a lambda / inner function like:

```
            def has_value(name):
                return (
                    name in node.input_default and
                    node.input_default[name] is not None and
                    str(node.input_default[name]).strip() != ""
                ) or (
                    name in input_schema and
                    input_schema[name].default is not None
                )           
```

majdyz:
Do we have a sample of the use case when depends_on is used? Should we add one to this PR for testing?

majdyz:
@Abhi1992002 I went ahead and updated the Python code for this:
https://github.com/Significant-Gravitas/AutoGPT/pull/8852/commits/2ff033aeb25634df6a394a86b2d7e32113b1b4d0

Which basically:
* Avoid the use of raw JSON schema model (and avoid dumping the schema into dict) and used the structured field.
* Addressing comments I described above
* Removed the last check that I thought it wasn't needed (let me know if this is wrong).

Feel free to update/revert if this doesn't match your needs, or let me know if this is fine.

ntindle (approved):

ntindle (approved):

PR #8856, opened 2024-12-02
Feat/single input validation lock

### LINKED ISSUE #8731: Add Input Dependencies and Exclusive Options to Prevent Conflicting Inputs

Abhi1992002:
- resolve #8731 

### Changes
- Introduced `mutually_exclusive` parameter in `SchemaField` to manage input exclusivity.
- Implemented logic in `NodeGenericInputField` to disable inputs based on mutual exclusivity.
- Updated related components to support the new `disabled` state for inputs.
- Enhanced `BlockIOSubSchemaMeta` to include `mutually_exclusive` property.

> Currently, I’m disabling the input from the same group (I haven’t added any frontend validation to prevent users from bypassing it).

https://github.com/user-attachments/assets/71fb9fe4-943b-4724-8acb-6aed2232ed6b

### DIFF
--- a/autogpt_platform/backend/backend/blocks/test_mutual_exclusive.py
+++ b/autogpt_platform/backend/backend/blocks/test_mutual_exclusive.py
@@ -1,48 +1,49 @@
-from typing import Optional, Literal
-from pydantic import BaseModel
+from typing import Optional
 
 from backend.data.block import Block, BlockCategory, BlockOutput, BlockSchema
 from backend.data.model import SchemaField
+
 
 class TestMutualExclusiveBlock(Block):
     class Input(BlockSchema):
         text_input_1: Optional[str] = SchemaField(
             title="Text Input 1",
             description="First mutually exclusive input",
+            mutually_exclusive="group_1",
         )
 
         text_input_2: Optional[str] = SchemaField(
             title="Text Input 2",
             description="Second mutually exclusive input",
+            mutually_exclusive="group_1",
         )
 
         text_input_3: Optional[str] = SchemaField(
             title="Text Input 3",
             description="Third mutually exclusive input",
+            mutually_exclusive="group_1",
         )
 
         number_input_1: Optional[int] = SchemaField(
             title="Number Input 1",
             description="First number input (mutually exclusive)",
-            mutually_exclusive="group2"
+            mutually_exclusive="group2",
         )
 
         number_input_2: Optional[int] = SchemaField(
             title="Number Input 2",
             description="Second number input (mutually exclusive)",
-            mutually_exclusive="group2"
+            mutually_exclusive="group2",
         )
 
         independent_input: str = SchemaField(
             title="Independent Input",
             description="This input is not mutually exclusive with others",
-            default="This can be filled anytime"
+            default="This can be filled anytime",
         )
 
     class Output(BlockSchema):
-        result: str = SchemaField(
-            description="Shows which inputs were filled"
-        )
+        result: str = SchemaField(description="Shows which inputs were filled")
 
     def __init__(self):
         super().__init__(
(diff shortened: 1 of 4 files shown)


autogpt_platform/backend/backend/blocks/test_mutual_exclusive.py
```diff
@@ -0,0 +1,74 @@
+from typing import Optional
+
+from backend.data.block import Block, BlockCategory, BlockOutput, BlockSchema
+from backend.data.model import SchemaField
+
+
+class TestMutualExclusiveBlock(Block):
```
aarushik93: This PR is a good addition - for the tests however, I wonder if you could put the tests in one of the real blocks to show it working?

Pwuts (commented):
This is a good feature, but I'm not sure about the implementation. JSON schema has `oneOf` keyword which seems perfect for this purpose. What do you think?

I always prefer using an existing mechanism or standard rather than inventing our own, unless that makes implementation unreasonably complex.

ntindle:
> This is a good feature, but I'm not sure about the implementation. JSON schema has `oneOf` keyword which seems perfect for this purpose. What do you think?
> 
> I always prefer using an existing mechanism or standard rather than inventing our own, unless that makes implementation unreasonably complex.

How would that look? I figure if we did that we'd need to implement the [rest](https://json-schema.org/draft/2020-12/json-schema-core#name-keywords-for-applying-subsch): allOf, anyOf, oneOf, not, if, then, etc

Abhi1992002:
@Pwuts and @ntindle I have changed the implementation from mutually exclusive to oneOf. Now you can check it.

Demo 

https://github.com/user-attachments/assets/c52ebb44-e7f7-49ac-9270-7b255ee362a0

ntindle (dismissed):

PR #8872, opened 2024-12-02
docs: huntr no longer is offering a security bounty so remove it

ntindle:
Huntr isn't offering a security bounty for autogpt at the moment so remove it in favor of github security adviosories

### Changes 🏗️

comments out huntr line in case they decide to offer it again in the future

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -8,7 +8,7 @@ We take the security of our project seriously. If you believe you have found a s
 
 Instead, please report them via:
 - [GitHub Security Advisory](https://github.com/Significant-Gravitas/AutoGPT/security/advisories/new)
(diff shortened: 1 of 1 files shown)


Pwuts (approved):
Why did they stop? Out of money?

ntindle:
no real answer was given but 

> Hey folks, we're rejigging our open source vulnerability programme to a tiered structure where the new top prize will exceed the previous one ($1500). The changes you see now are just a temporary stop gap until we roll out the tiered system. Since we're trying to break ground in AI/ML supply chain security we constantly have to make tweaks until we find a model that works for everybody. This specific change aims to better reward enterprise critical vulns while still maintaining our usual service for the other great vulns you find. We are working hard to roll out the changes before the holiday break. As always; all past bounties and promises will be honoured.

PR #8908, opened 2024-12-08
fix(backend): Fix Github PR blocks

### LINKED ISSUE #8667: GitHub Read PR block with "Include PR Changes" fails
The URL conversion logic that the `read_pr_changes` method relies on doesn't work correctly. It ends up trying to fetch `api.github.com/repos/{repo}/pull/{number}/files`, while the REST API only defines `.../{repo}/pulls/...`

vishesh10:
### Background

The Github PR blocks are not able to function properly because the correct endpoint is not getting called.
- Resolves #8667

### Changes 🏗️
* Added logic to derive correct endpoint from the given PR url.
* This logic is implemented in multiple blocks viz. `GithubReadPullRequestBlock`, `GithubAssignPRReviewerBlock`, `GithubUnassignPRReviewerBlock`, `GithubListPRReviewersBlock`.

### Test
* Github List PR Reviewers
Screenshot 2024-12-03 at 11 03 59 PM

* Github Read Pull Request (Include Pr Changes checked)
Screenshot 2024-12-06 at 12 01 41 PM


autogpt_platform/backend/backend/blocks/github/pull_requests.py
```diff
@@ -499,3 +501,11 @@ def run(
...
         )
         yield from (("reviewer", reviewer) for reviewer in reviewers)
+
+
+def prepare_pr_api_url(pr_url: str, path: str) -> str:
+    if (match := re.match(r"(.+)(/pull/)+?(.+)", pr_url)) is not None and len(
+        match.groups()
+    ) == 3:
+        return f"{match.groups()[0]}/pulls/{match.groups()[2]}/{path}"
+    raise ValueError("There is some issue with the PR Url")
```
aarushik93: nit: can we do the error more clearly such as: 

unable to parse ur: pr_url

majdyz (approved):
Comment addressed and approved

PR #8838, opened 2024-11-28
feat(blocks): Add webhook block status indicator

Pwuts:
- Resolves #8743
- Follow-up to #8358

### Demo

https://github.com/user-attachments/assets/f983dfa2-2dc2-4ab0-8373-e768ba17e6f7

### Changes 🏗️

- feat(frontend): Add webhook status indicator on `CustomNode`
   - Add `webhookId` to frontend node data model

- fix(backend): Fix webhook ping endpoint
   - Remove `provider` path parameter
   - Fix return values and error handling
   - Fix `WebhooksManager.trigger_ping(..)`
      - Add `credentials` parameter
      - Fix usage of credentials
   - Fix `.data.integrations.wait_for_webhook_event(..)`
      - Add `AsyncRedisEventBus.wait_for_event(..)`

- feat(frontend): Add `BackendAPIProvider` + `useBackendAPI`

- feat(frontend): Improve layout of node header

    Before:
    image
    After:
    image

- refactor(backend): Clean up `.data.integrations`
- refactor(backend): Fix naming in `.data.queue` for understandability

#### For code changes:
- [x] I have clearly listed my changes in the PR description
- [x] I have made a test plan
- [x] I have tested my changes according to the test plan:
  
  - [x] Add webhook block, save -> gray indicator
  - [x] Add necessary info to webhook block, save -> green indicator
  - [x] Remove necessary info, save -> gray indicator

PR #8916, opened 2024-12-09
fix(frontend): Automatically choose client/server-side API client

### LINKED ISSUE #8912: Automatic server-side/client-side backend API client usage
The `frontend` has `@/lib/autogpt-server-api/client` and `@/lib/autogpt-server-api/clientServer`, for client-side and server-side use respectively. This choice should be handled automatically in the module, rather than leaving it up to other modules to import the right one.

The `useBackendAPI` hook introduced in #8838 always uses `./client`, which is basically the same issue.

Pwuts:
- Resolves #8912
- Fixes server-side issue with `useBackendAPI` hook introduced in #8838

We have separate client-side and server-side API clients. Choosing the right one should be done in the exporting module, not in all import locations.

### Changes 🏗️

- Rename `AutoGPTServerAPI` to `AutoGPTServerAPIClientSide`
- Merge `autogpt-server-api/clientServer` into `autogpt-server-api/client`
- Add environment-dependent `AutoGPTServerAPI` export in `autogpt-server-api/client`

#### For code changes:
- [ ] I have made a test plan
  
  - [ ] ...

<details>
  
</details>


autogpt_platform/frontend/src/lib/autogpt-server-api/index.ts
```diff
@@ -1,5 +1,6 @@
 import { AutoGPTServerAPI } from "./client";
 
+// TODO: rename to BackendAPI
```
kcze, line 3: Why not do this in this PR?

autogpt_platform/frontend/src/lib/autogpt-server-api/clientServer.ts
kcze, line 1: Are there no imports of this anywhere?

Pwuts: nope

autogpt_platform/frontend/src/lib/autogpt-server-api/index.ts
Pwuts: Out of scope, would cause a huge amount of changes and also conflicts on ongoing PRs

Abhi1992002:
@Pwuts getting this error after making these changes?

Screenshot 2024-12-09 at 6 49 02 PM

ntindle:
I spent a lot of time playing this game, and couldn’t get it working which is why I had it the other way. I think making test components for each would probably be productive as you try various methods

PR #8940, opened 2024-12-11
fix(frontend) : Optional number input

### LINKED ISSUE #8928: Frontend schema validation always breaks on optional 'integer' input field
This makes the `max_tokens` field on AI blocks can't be defined

Abhi1992002:
- Resolve #8928 

Currently, the frontend renders a string input for an optional integer. I have now corrected it.

Screenshot 2024-12-11 at 10 23 48 AM


Abhi1992002:
I think we also need to render other optional types also.

Bentlybro (approved):
Tested and works, Thank you :)

PR #8982, opened 2024-12-13
feat(frontend) : Add optional input support for object, array, multi-select, and select as well.

### LINKED ISSUE #8976: Render the correct input for optional fields.
Currently, we are not rendering the correct input for Optionals. By default, we provide them with a string, which is not good for the UI.

Abhi1992002:
- Resolve #8976 

> Once you have checked whether this is working or not, then I will remove the optional field block.

### Changes 
- Updated `NodeGenericInputField` to handle additional input types:
  - Added support for `array` and `object` optional types.
  - Enhanced schema definitions for `string` optional  type to include enumerations

### Testing 🔍
- Verified that the new input types function correctly within the frontend component.

Screenshot 2024-12-13 at 7 08 22 PM


autogpt_platform/backend/migrations/20241213141341_/migration.sql
```diff
@@ -0,0 +1,11 @@
+-- AlterTable
+ALTER TABLE "CreditTransaction" RENAME CONSTRAINT "UserBlockCredit_pkey" TO "CreditTransaction_pkey";
```
Swiftyos: Please can you remove this migration and check if it is needed again

Swiftyos (changes requested):

Swiftyos (dismissed):

ntindle (approved):

PR #9043, opened 2024-12-18
feat(platform): Add basic library functionality

Swiftyos:
Add functionality to allow users to add agents to their library from the store page.


Bentlybro (approved):
Tested and this works

PR #9050, opened 2024-12-18
feat(platform): Implement top-up flow for PAYG System

kcze:
This PR adds Stripe integration and payment processing for topping-up user accounts with credits.

### Changes 🏗️

Includes:
- https://github.com/Significant-Gravitas/AutoGPT/pull/9176

#### Top-up flow

1. To top-up a user visits their settings and clicks `Credits` button (it's unavailable if `NEXT_PUBLIC_STRIPE_PUBLISHABLE_KEY` isn't present)
2. User inputs top-up amount (min 5$ in 1$ increments) and click the button to confirm.
3. Backend receives top-up request, creates database entry and requests stripe to provide url for this specific checkout.
4. User gets redirected to externally hosted Stripe checkout page, after payment (or cancelling) they get redirected back to Credits page.
5. In the meantime Stripe processes payment and sends webhook confirmation to the backend, backend updates database to activate bought credits.
6. Credits page shows success (or failure) information (by using url param `topup=success|cancel`). Credit counter won't update without refreshing the page unless payment was confirmed before user was back on Credits page which is the case when testing checkout locally.

Screenshot 2025-01-01 at 2 55 35 PM

#### Backend
- Add `stripe` package
- Add environment variables:
  - `STRIPE_API_KEY`
  - `STRIPE_WEBHOOK_SECRET`
- Add routes:
  - `POST /credits`: top-up request, returns Stripe checkout url.
  - `POST /credits/stripe_webhook`: Stripe webhook endpoint to notify of successful payment.
  - `PATCH /credits`: prompts beckend to check payment status. It's an additional failsafe in case webhook fails.
- Update `credit.py` and related files to handle top-up request and payment confirmation

#### Frontend
- Add `stripe-js` package
- Add `NEXT_PUBLIC_STRIPE_PUBLISHABLE_KEY` environment variable
- Modify user settings sidebar to show `Credits` if `NEXT_PUBLIC_STRIPE_PUBLISHABLE_KEY` is available
- Add `store/credits` page where user can top-up their account, it shows confirmation (or failure) after completing checkout.
- Add `useCredits` hook that returns user credits and allows to request top-up.

#### For code changes:
- [ ] I have made a test plan
  
  - [ ] ...

<details>
  
</details>

<details>

  - Changing ports
</details>


autogpt_platform/backend/backend/data/credit.py
```diff
@@ -211,6 +229,62 @@ async def top_up_credits(self, user_id: str, amount: int):
...
+        if not user:
+            raise ValueError(f"User not found: {user_id}")
+
+        # Create customer if not exists
+        if not user.stripeCustomerId:
+            customer = stripe.Customer.create(name=user.name or "", email=user.email)
+            await User.prisma().update(
+                where={"id": user_id}, data={"stripeCustomerId": customer.id}
+            )
+            user.stripeCustomerId = customer.id
```
majdyz:
```suggestion
            user = await User.prisma().update(
                where={"id": user_id}, data={"stripeCustomerId": customer.id}
            )
```

majdyz: Or even better: avoid calling Prisma here and add set_stripe_customer_id in model class on user.py

autogpt_platform/backend/backend/data/credit.py
```diff
@@ -211,6 +229,62 @@ async def top_up_credits(self, user_id: str, amount: int):
...
+                        "product_data": {
+                            "name": "AutoGPT Platform Credits",
+                        },
+                        "unit_amount": amount * 100,
+                    },
+                    "quantity": 1,
+                }
+            ],
+            mode="payment",
+            success_url="",# TODO kcze
```
majdyz: So these will be webhook URLs ?

autogpt_platform/frontend/src/hooks/useCredits.ts
```diff
@@ -0,0 +1,44 @@
...
+  useEffect(() => {
+    fetchCredits();
+  }, [fetchCredits]);
+
+  const requestTopUp = useCallback(async (amount: number) => {
+    const stripe = await stripePromise;
+
+    if (!stripe) {
+      console.error("Stripe failed to load");
+      return;
```
majdyz: Should we throw an error instead ?

autogpt_platform/backend/backend/data/credit.py
kcze: These are redirect urls after checkout

autogpt_platform/backend/backend/data/credit.py
```diff
@@ -211,6 +239,90 @@ async def top_up_credits(self, user_id: str, amount: int):
...
+        # This can be called multiple times for one id, so ignore if already fulfilled
+        if credit_transaction.isActive:
+            return
+
+        # Retrieve the Checkout Session from the API
+        checkout_session = stripe.checkout.Session.retrieve(session_id)
+
+        # Check the Checkout Session's payment_status property
+        # to determine if fulfillment should be peformed
+        if checkout_session.payment_status != "unpaid":
```
majdyz: Isn't this riskier than just match 'paid' & 'no_payment_required' ?

autogpt_platform/backend/backend/server/routers/v1.py
```diff
@@ -137,6 +140,43 @@ async def get_user_credits(
...
 
+@v1_router.post(path="/credits", tags=["credits"], dependencies=[Depends(auth_middleware)])
+async def request_top_up(
+    request: RequestTopUp, user_id: Annotated[str, Depends(get_user_id)]
+):
+    checkout_url = await _user_credit_model.top_up_intent(user_id, request.amount)
+    return {"checkout_url": checkout_url}
+
+
+@v1_router.post(path="/credits/stripe_webhook", tags=["credits"])
```
majdyz, line 160: Curious, Is this a non-authenticated request ?

autogpt_platform/backend/backend/server/routers/v1.py
```diff
@@ -137,6 +140,43 @@ async def get_user_credits(
...
+            payload, sig_header, settings.secrets.stripe_webhook_secret
+        )
+    except ValueError:
+        # Invalid payload
+        raise HTTPException(status_code=400)
+    except stripe.SignatureVerificationError:
+        # Invalid signature
+        raise HTTPException(status_code=400)
+    
+    print(event)
```
majdyz: nit: make this logging ?

autogpt_platform/backend/backend/server/routers/v1.py
```diff
@@ -137,6 +140,43 @@ async def get_user_credits(
...
+
+    if (
+        event['type'] == 'checkout.session.completed'
+        or event['type'] == 'checkout.session.async_payment_succeeded'
+    ):
+        await _user_credit_model.fulfill_checkout(event['data']['object']['id'])
```
majdyz: the nested check: uses `get()` with fallback for safety?

majdyz: Also can we log/warn the non matching events ?

autogpt_platform/backend/backend/data/credit.py
```diff
@@ -211,6 +239,90 @@ async def top_up_credits(self, user_id: str, amount: int):
...
+            print("Payment status is not unpaid!")
```
majdyz: Nit: make this and the one above logging with more information ?

autogpt_platform/frontend/src/hooks/useCredits.ts
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    const response = await api.requestTopUp(amount);
+
+    router.push(response.checkout_url);
+  }, [api]);
```
majdyz: Won' this warn about router not part of the deplist ?

autogpt_platform/frontend/src/lib/autogpt-server-api/client.ts
```diff
@@ -442,7 +446,7 @@ export default class BackendAPI {
         await new Promise((resolve) => setTimeout(resolve, 100 * retryCount));
       }
     }
-    console.log("Request: ", method, path, "from: ", page);
+    console.log("Request: ", method, this.baseUrl, path, "from: ", page);
```
majdyz: nit: console.debug ?

autogpt_platform/backend/backend/data/credit.py
kcze: I left the original code - trying to change it or move to the function cascades into multiple errors and requires additional checks.

autogpt_platform/frontend/src/hooks/useCredits.ts
kcze: I removed `console.error` altogether, this should never happen; `stripePromise` should never be `null` and `loadStripe` throws when we don't provide API key.

autogpt_platform/backend/backend/data/credit.py
kcze: Not now, there are only three options and this is from Stripe example but just to be sure I updated it anyway!

autogpt_platform/backend/backend/server/routers/v1.py
kcze: I guess no but it requires HTTPS on live data and webhook secret to read correctly.

autogpt_platform/backend/backend/server/routers/v1.py
kcze: removed

autogpt_platform/backend/backend/server/routers/v1.py
kcze: > the nested check: uses get() with fallback for safety?

What do you mean? I don't think using `event.get(..)` would change anything.

> Also can we log/warn the non matching events ?

Do you think this is necessary on our side? Stripe logs everything and each event has tons of data and only 4 out of hundreds are enabled for this webhook. I could disabled 2, so only the 2 matching will be sent.

autogpt_platform/backend/backend/data/credit.py
kcze: Removed, that was only for testing.

autogpt_platform/frontend/src/hooks/useCredits.ts
kcze: Added router

autogpt_platform/frontend/src/lib/autogpt-server-api/client.ts
kcze: Changed but this needs a refactor imo

autogpt_platform/backend/backend/data/credit.py
```diff
@@ -202,18 +232,124 @@ async def spend_credits(
...
+        # Retrieve CreditTransaction
+        credit_transaction = await CreditTransaction.prisma().find_first_or_raise(
+            where={
+                "OR": [
+                    (
+                        {"transactionKey": session_id}
+                        if session_id is not None
+                        else {"transactionKey": ""}
+                    ),
+                    {"userId": user_id} if user_id is not None else {"userId": ""},
```
majdyz: 1. Shouldn't we have isActive == False filter here? I see that you have a `fulfill_checkout` flow that only uses `user_id`, this won't work if there is already an active transaction after it.
2. Can't we make the userId mandatory filter? is the webhook from stripe won't provide the user id ?
3. We can make the clause "AND" instead of "OR": 
```
"AND": [
    ...[transaction_filter if session_id else []],
    ...[userid_filter if user_id else []],
]
```

autogpt_platform/backend/backend/data/credit.py
```diff
@@ -222,12 +358,19 @@ async def spend_credits(self, *args, **kwargs) -> int:
...
+        return ""
+
+    async def fulfill_checkout(self, *args, **kwargs):
+        pass
+
 
 def get_user_credit_model() -> UserCreditBase:
-    if config.enable_credit.lower() == "true":
-        return UserCredit(config.num_user_credits_refill)
+    # return UserCredit()
```
majdyz: nit: revert the comment ?

slightly unrelated: also I wonder why we are using string config instead of boolean in the first place :D

autogpt_platform/backend/backend/server/model.py
```diff
@@ -56,3 +56,8 @@ class SetGraphActiveVersion(pydantic.BaseModel):
 
 class UpdatePermissionsRequest(pydantic.BaseModel):
     permissions: List[APIKeyPermission]
+
+
+class RequestTopUp(pydantic.BaseModel):
+    amount: int
+    """Amount of credits to top up."""
```
majdyz, line 63: The comment seems unnecessary?

majdyz (dismissed):

autogpt_platform/backend/backend/server/model.py
kcze: It makes it clear that it's about *credits* and not *USD*

autogpt_platform/backend/backend/data/credit.py
kcze: 1. Done
2. No, webhook won't provide `userId` and it was tricky enough to pass `checkoutId` to the frontend after checkout that I decided to just use `userId` and the most recent transaction. More than one checkout shouldn't be happening for a single user at the same time and even if it should still work assuming webhook worked. We need a different way to retrieve/confirm old transactions.
3. I might be misunderstanding your example and I don't think it'd make a difference? Every `CreditTransaction` has both `transactionKey` and `userId` and this query searches for either, so you can't `AND` it because it'd try to search for some `transactionId` without `userId` (or the other way around).

autogpt_platform/backend/backend/data/credit.py
kcze: I've changed it to bool and removed the comment.

autogpt_platform/backend/backend/server/model.py
kcze: Now when I think about it again it'd make more sense to call it `credit_amount` and not use comment.

majdyz (dismissed):

autogpt_platform/backend/.env.example
```diff
@@ -15,6 +15,9 @@ REDIS_PORT=6379
 REDIS_PASSWORD=password
 
 ENABLE_CREDIT=false
+STRIPE_API_KEY=
+STRIPE_WEBHOOK_SECRET=
+
```
majdyz, line 20: Should we set `PLATFORM_BASE_URL` https://github.com/Significant-Gravitas/AutoGPT/pull/9050/files#diff-62020caf1b9a15e0e3b9b3b1b69d5f6464bf7643f62354cbbaabf755d57b6064L39 using local default value ? e.g: `http://localhost:3000` ?

autogpt_platform/frontend/src/app/store/(user)/credits/page.tsx
```diff
@@ -0,0 +1,73 @@
...
+          />
+        </div>
+      </div>
+      <Button
+        type="submit"
+        variant="default"
+        className="font-circular mt-4 h-[50px] rounded-[35px] bg-neutral-800 px-6 py-3 text-base font-medium text-white transition-colors hover:bg-neutral-900 dark:bg-neutral-200 dark:text-neutral-900 dark:hover:bg-neutral-100"
+        onClick={() => requestTopUp(amount)}
+      >
+        {"Top-up"}
```
majdyz, line 69: Can we add `add/edit payment method` here for configuring future payment & auto top-up ?

autogpt_platform/backend/backend/server/model.py
majdyz: Feel free to do this on a followup PR

autogpt_platform/frontend/src/app/store/(user)/credits/page.tsx
majdyz: Let's do it on a separate PR

majdyz (dismissed):

majdyz (approved):

autogpt_platform/frontend/src/app/store/(user)/credits/page.tsx
majdyz: cc @kcze

majdyz (approved):

autogpt_platform/backend/backend/data/credit.py
```diff
@@ -178,42 +332,209 @@ async def spend_credits(
...
+
+    async def _get_missing_grant_months(self, user_id: str) -> list[date]:
+        """
+        Get list of missing grant months after the latest grant.
+        If no grant is found, return only the current month.
+        """
+        cur_date = self.time_now().date().replace(day=1)
+
+        # Find latest grant using transaction key pattern
+        latest_grant = await CreditTransaction.prisma().find_first(
```
majdyz: this will cause the full transaction lookup for the user through the history, and this happens on each get_balance command

autogpt_platform/backend/backend/data/credit.py
```diff
@@ -178,42 +332,209 @@ async def spend_credits(
...
+                user_id=credit_transaction.userId,
+                metadata=Json({"checkout_session": checkout_session}),
+            )
+
+    async def get_credits(self, user_id: str) -> int:
+        balance = await self._get_credits(user_id)
+        grant_amount = await self.get_credit_grant(user_id)
+
+        # Apply grants for each missing month
+        for grant_date in await self._get_missing_grant_months(user_id):
```
majdyz: Do we even need to grant a credit to a user who hasn't logged in to our platform for the whole month?
This is the only scenario where this is needed, isn't it?

autogpt_platform/backend/backend/data/credit.py
```diff
@@ -178,42 +332,209 @@ async def spend_credits(
...
+            try:
+                new_balance = balance + grant_amount
+                await CreditTransaction.prisma().create(
```
majdyz: we need to use _add_transaction function for this or add the lock

autogpt_platform/backend/backend/data/credit.py
```diff
@@ -178,42 +332,209 @@ async def spend_credits(
...
+        # Check the Checkout Session's payment_status property
+        # to determine if fulfillment should be performed
+        if checkout_session.payment_status in ["paid", "no_payment_required"]:
+            await self._enable_transaction(
+                transaction_key=credit_transaction.transactionKey,
+                user_id=credit_transaction.userId,
+                metadata=Json({"checkout_session": checkout_session}),
+            )
+
+    async def get_credits(self, user_id: str) -> int:
```
majdyz, line 389: My main concern on this is the get_credits is becoming too heavy of a process with all the lookups, and fetching the whole missing grant month is kind of defeating the whole point of snapshotting.

autogpt_platform/backend/backend/data/credit.py
```diff
@@ -178,42 +332,209 @@ async def spend_credits(
...
+                "transactionKey": {
+                    "startsWith": "MONTHLY-CREDIT-TOP-UP-"  # type: ignore
+                },
+            },
+            order={"createdAt": "desc"},
         )
 
+        if not latest_grant:
+            # No previous grants, just return current month
+            return [cur_date]
```
majdyz: Why is this the case? e.g: no latest_grant == no missing month, instead of no latest_grant == no grant at all.

majdyz (approved):

majdyz (approved):

Issue #8919, opened 2024-12-09
Credential ID Exports into Agent JSON


Torantulino:
Steps to reproduce:

1\. On account/machine 1, export an agent using your own local credentials (not the shared cloud credentials) 2. (Try to) Import this agent on account/machine

2\. It won't import due to the credentials in the Agent JSON being missing from your machine.

Basic Example Agents to reproduce with attached.

```json
        "credentials": {
          "id": "6767232a-3407-4c34-85a3-6887d4969f0c",
          "title": "Anthropic Toran",
          "provider": "anthropic",
          "type": "api_key"
        },
```

[Credentials Export test_v2.json](https://uploads.linear.app/a47946b5-12cd-4b3d-8822-df04c855879f/0b4d780c-c9fb-49ac-bfda-308a2c3b9a66/4e12660f-ddc1-4823-bef3-4ba8937b0275?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiL2E0Nzk0NmI1LTEyY2QtNGIzZC04ODIyLWRmMDRjODU1ODc5Zi8wYjRkNzgwYy1jOWZiLTQ5YWMtYmZkYS0zMDhhMmMzYjlhNjYvNGUxMjY2MGYtZGRjMS00ODIzLWJlZjMtNGJhODkzN2IwMjc1IiwiaWF0IjoxNzMzNzQ5MjQxLCJleHAiOjMzMzA0MzA5MjQxfQ.knekDmbPPX6jtYOsWeqUjOaK7xM3kyzdmRUHVunZmJY)

[...]

PR #9081, opened 2024-12-19
feat(frontend) Remove "credentials" from export & import of agents

Bentlybro:
### Changes 🏗️

This is for [Credential ID Exports into Agent JSON #8919 ](https://github.com/Significant-Gravitas/AutoGPT/issues/8919)

I have added a new function ``removeCredentials`` into [``utils.ts``](https://github.com/Significant-Gravitas/AutoGPT/compare/dev...bently/open-2153-credential-id-exports-into-agent-json?expand=1#diff-db26a69e6fb7546dc621634f3c8ee6efa3639e72e02837f753af18b2fdddf7be) which will go through and look for any "credentials" that are in the JSON during a agent export, this will then remove them and let the user download the file.

I have also added the same function to the importing of agents for old agents that where exported that still contain the credentials, this means that old agents can be imported with out breaking/causing issues. 

When I say it looks for credentials I dont mean actual credentials like api keys them self, in the JSON that is exported it contains the following, this needs removing 
```
"credentials": {
  "id": "6767232a-3407-4c34-85a3-6887d4969f0c",
  "title": "Anthropic Toran",
  "provider": "anthropic",
  "type": "api_key"
},
```

If there is a better way to go about this let me know!

### DIFF
--- a/autogpt_platform/frontend/src/components/agent-import-form.tsx
+++ b/autogpt_platform/frontend/src/components/agent-import-form.tsx
@@ -15,7 +15,7 @@ import { Button } from "@/components/ui/button";
 import { Switch } from "@/components/ui/switch";
 import { Textarea } from "@/components/ui/textarea";
 import { Graph, GraphCreatable } from "@/lib/autogpt-server-api";
-import { cn } from "@/lib/utils";
+import { cn, removeCredentials } from "@/lib/utils";
 import { EnterIcon } from "@radix-ui/react-icons";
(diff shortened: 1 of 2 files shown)


Swiftyos (approved):
This is a good idea

aarushik93 (approved):

Pwuts (commented):
The sanitization of agent objects for export was already implemented in the backend. Is that broken? If so, it should be debugged rather than duplicating it in the frontend.

PR #9098, opened 2024-12-20
fix(store): Fixing add agent to library

Swiftyos:
Do a deep copy of the store agent so the new agent is under the current users id

⚠️  Hacky fix!!


autogpt_platform/backend/backend/data/graph.py
```diff
@@ -521,13 +521,13 @@ async def get_graph(
     """
     where_clause: AgentGraphWhereInput = {
         "id": graph_id,
-        "isTemplate": template,
```
Snapp949, line 524: Hmm

PR #9128, opened 2024-12-25
fix(platform): Fields with default value are not set to advanced by default

majdyz:
https://github.com/Significant-Gravitas/AutoGPT/issues/8739 causes input fields that are supposed to be an advanced field end up being a mandatory field:

image

*See the retry count field here.

### Changes 🏗️

Set the `advanced` field on each input field, and set the default value using this logic:
* If it has a default value, set it to True.
* otherwise, False.

#### For code changes:
- [ ] I have made a test plan
  
  - [ ] ...

<details>
  
</details>

<details>

  - Changing ports
</details>

### DIFF
--- a/autogpt_platform/backend/backend/data/block.py
+++ b/autogpt_platform/backend/backend/data/block.py
@@ -97,11 +97,6 @@ class BlockSchema(BaseModel):
 
         cls.cached_jsonschema = cast(dict[str, Any], ref_to_dict(model))
 
-        # Set default properties values
-        for field in cls.cached_jsonschema.get("properties", {}).values():
-            if isinstance(field, dict) and "advanced" not in field:
-                field["advanced"] = True
-
         return cls.cached_jsonschema
 
     @classmethod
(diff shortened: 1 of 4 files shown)


Swiftyos:
We made a conscious choice to allow fields with default values that aren’t marked as advanced. There are times when we don’t want to automatically mark a field as advanced while still providing a default value. 

Does this updated version still allow inputs with default values that aren’t advanced?

majdyz:
> We made a conscious choice to allow fields with default values that aren’t marked as advanced. There are times when we don’t want to automatically mark a field as advanced while still providing a default value.
Does this updated version still allow inputs with default values that aren’t advanced?

Yes, This is the only reason I added this `advanced` field feature.

But https://github.com/Significant-Gravitas/AutoGPT/issues/8739 makes all the values with default non-advanced unless it's explicitly defined to do so, which is not the expectation.

Swiftyos (approved):

PR #9143, opened 2024-12-29
feat(blocks): Add Image to video block with Kling Pro 1.5

Torantulino:


majdyz (commented):
@Torantulino is this still relevant? 
Feel free to re-tag me if it is

PR #9158, opened 2024-12-30
CodeExecutionBlock split into InstantiationBlock & StepExecutionBlock

mariosacaj:
Especially in case of heavy sandbox customization we don't want to spawn a new VM every time we need to execute code, especially in circumstances  when time of inactivity is negligible.

### Changes 🏗️

Added two now blocks inside `autogpt_platform/backend/backend/blocks/code_executor.py` which split the logic of the CodeExecutionBlock in two parts: InstantiationBlock and StepExecutionBlock. 

The overall setup shall be the done in the InstantiationBlock which then passes as output the `sandbox_id` which identifies the sandbox instance. 

At a later stage any line of code can be executed through StepExecutionBlock inside the same instance using the e2b library function `sandbox = Sandbox.connect(sandbox_id=sandbox_id, api_key=api_key)`. Variables and memory form earlier executions or instantiation are persisted.

Beware, this approach does not make use of the beta apis to pause and resume an instance. By using the approach above the VM will likely be killed after the timeout inactivity set at instantiation time.

### New Features:

* **InstantiationBlock Class**:
  - Added a new class to instantiate an isolated sandbox environment with internet access for code execution. This class includes input and output schemas, methods for running and executing code, and handles sandbox setup commands and code execution.

* **StepExecutionBlock Class**:
  - Introduced a new class to execute code in a previously instantiated sandbox environment. This class also includes input and output schemas, methods for running and executing step code, and handles connecting to an existing sandbox for code execution.

### Minor Changes:

* **ProgrammingLanguage Enum**:
  - Added a blank line for better readability within the `ProgrammingLanguage` enum definition.


ntindle:
I think there’s a healthy case for all three blocks here with the Code Execution block calling the other two to handle its business. What are your thoughts on that?

aarushik93 (approved):

PR #9172, opened 2025-01-02
feat(blocks): Add github list stargazers block

Bentlybro:
This adds a list stargazers block, its using https://docs.github.com/en/rest/activity/starring?apiVersion=2022-11-28#list-stargazers 

image


Swiftyos (approved):

autogpt_platform/backend/backend/blocks/github/pull_requests.py
Pwuts, line 1: Also a nice addition but also in the wrong file: should be in `github/repo.py`

PR #9179, opened 2025-01-03
feature(platform): add ability to execute store agents without agent ownership

Swiftyos:
#### For code changes:
- [ ] I have made a test plan
  
  - [ ] ...

<details>
  
</details>

<details>

  - Changing ports
</details>


autogpt_platform/backend/backend/data/graph.py
```diff
@@ -543,20 +548,38 @@ async def get_graph(
...
+    store_listing_where: StoreListingWhereInput = {
+        "agentId": graph_id,
+        "agentVersion": version,
+    }
+
+    store_listing = await StoreListing.prisma().find_first(where=store_listing_where)
+
+    # If it is not a store listing, return None
+    if not store_listing:
+        return None
```
Pwuts, line 580: What's this for? Doesn't this prevent users from viewing their own private agents?

autogpt_platform/backend/backend/data/graph.py
```diff
@@ -543,20 +548,38 @@ async def get_graph(
...
+        version = graph.version
+
+    store_listing_where: StoreListingWhereInput = {
+        "agentId": graph_id,
+        "agentVersion": version,
+    }
+
+    store_listing = await StoreListing.prisma().find_first(where=store_listing_where)
+
+    # If it is not a store listing, return None
```
Pwuts: just to clarify:
```suggestion
    # If it does not belong to the user nor is not a store listing, return None
```
or even:
```suggestion
    # If it does not belong to the user nor is not a store listing, deny access
```

autogpt_platform/backend/backend/executor/manager.py
```diff
@@ -780,7 +780,7 @@ def add_execution(
         graph_id: str,
         data: BlockInput,
         user_id: str,
-        graph_version: int | None = None,
+        graph_version: int,
```
Pwuts, line 783: Would it be difficult to keep the default behavior of executing the active version?

autogpt_platform/backend/backend/data/graph.py
```diff
@@ -543,20 +548,38 @@ async def get_graph(
     where_clause: AgentGraphWhereInput = {
         "id": graph_id,
     }
+
     if version is not None:
         where_clause["version"] = version
```
Pwuts, line 553: I understand removing the template check, but removing the `isActive` filter seems like an unintended behavior change?
```suggestion
    if version is not None:
        where_clause["version"] = version
    else:
        where_clause["isActive"] = True
```

autogpt_platform/backend/backend/server/routers/v1.py
```diff
@@ -230,8 +228,6 @@ async def do_create_graph(
             status_code=400, detail="Either graph or template_id must be provided."
         )
 
-    graph.is_template = is_template
-    graph.is_active = not is_template
```
Pwuts, line 234: not sure this should be removed
```suggestion
    graph.is_active = True
```

autogpt_platform/backend/backend/data/graph.py
Swiftyos: `isActive` was meant for agents that are ‘running’, like webhooked agents, since we didn’t have a better place to store that idea.

Now, we can move this concept to an `AgentPreset` and remove it from the `AgentGraph`. This way, the `AgentGraph` table can focus solely on defining an agent.

autogpt_platform/backend/backend/executor/manager.py
Swiftyos: Before we reach the executor, we should know exactly what we want to execute. If we force the version number to be present, any calls to this function without specifying a version will fail linting.

But just to be clear, technically, it would still work if we pass graph_version = None to the get_graph function. It would still get the latest graph if we don’t specify a version.

autogpt_platform/backend/backend/server/routers/v1.py
Swiftyos: Defaults are:

```
class Graph(BaseDbModel):
    version: int = 1
    is_active: bool = True
    is_template: bool = False
 ```
 
 So no need to do this.

File: autogpt_platform/backend/backend/data/graph.py (hunk 4 above)
Comment (Pwuts) [reply autogpt_platform/backend/backend/data/graph.py:553]: > `isActive` was meant for agents that are ‘running’, like webhooked agents

Not sure if I understand what you mean, but that seems incorrect to me. `isActive` indicates the active version of a graph: the version that is fetched and executed if no version number is specified in a fetch/execute call. This isn't necessarily the latest version, as users can technically revert to an older version. So `isActive` is part of the version management system, and can't just be removed without causing behavior changes.

> Now, we can move this concept to an `AgentPreset` and remove it from the `AgentGraph`. This way, the `AgentGraph` table can focus solely on defining an agent.

That would mean we need to adjust all code paths that rely on `isActive` to rely on `AgentPresets` instead. I don't see that in this PR. Also, if `isActive` ends up not being used, it should be removed from the DB schema right?

Comment (majdyz) [reply autogpt_platform/backend/backend/data/graph.py:553]: > Now, we can move this concept to an AgentPreset and remove it from the AgentGraph. This way, the AgentGraph table can focus solely on defining an agent.

What happens when we save new changes on an agent? won't it still mark the old version `isActive` to False, and set the new one to True ?

Comment (Swiftyos) [reply autogpt_platform/backend/backend/data/graph.py:553]: Okay looks like I misinterpreted what isActive meant, probably should rename that to isActiveVersion, but out of scope for this PR.

I have added the filter back in

File: autogpt_platform/backend/backend/data/graph.py [hunk 6]
```diff
@@ -543,20 +548,40 @@ async def get_graph(
...
-        where_clause["userId"] = user_id
-
     graph = await AgentGraph.prisma().find_first(
         where=where_clause,
         include=AGENT_GRAPH_INCLUDE,
         order={"version": "desc"},
     )
+
+    if not graph:
+        return None
```
majdyz: If we already have this, we can remove `if graph else None` in both returns of this function.

autogpt_platform/backend/backend/data/graph.py
```diff
@@ -543,20 +548,40 @@ async def get_graph(
...
+        "agentVersion": version,
+    }
+
+    store_listing = await StoreListing.prisma().find_first(where=store_listing_where)
+
+    # If it does not belong to the user nor is not a store listing, return None
+    if not store_listing:
+        return None
+
+    # If it is a store listing, return the graph model
```
majdyz:
```suggestion
    # The Graph has to be owned by the user or a store listing.
    if graph.userId != user_id and not (await StoreListing.prisma().find_first(where={
        "agentId": graph_id,
        "agentVersion": version or graph.version,
    })):
        return None
```
I think a concise form is easier to understand.

majdyz: Btw don't we put model queries for StoreListing in store/db.py? can we put this function there?

majdyz (dismissed):

autogpt_platform/backend/backend/data/graph.py
majdyz: Also ["agentId", "agentVersion"] has to be indexed.

We can replace the index from:
```
  @@index([agentId])
```

to 

```
  @@index([agentId, agentVersion])
```

Or even further, why do we need ID for StoreListing, StoreListingVersion ?
Can we make `[agentId, agentVersion]` `StoreListing` id and `[agentId, agentVersion, version]` StoreListingVersion id ?

majdyz: Also does that mean the query has to filter the `isApproved` & `isDeleted` too?
I think the query needs to be placed on store/db.py with these filters ?

autogpt_platform/backend/backend/data/graph.py
```diff
@@ -529,7 +535,6 @@ async def get_execution(user_id: str, execution_id: str) -> GraphExecution | Non
 async def get_graph(
     graph_id: str,
     version: int | None = None,
-    template: bool = False,
     user_id: str | None = None,
```
majdyz, line 538: The `user_id` meaning is no longer a filter here but a security requirement, I think we need to make this mandatory.
The template is deprecated, right?

autogpt_platform/backend/backend/data/graph.py
Swiftyos: graph could be None at this point as the return of `await AgentGraph.prisma().find_first(...)` is `AgentGraph.prisma | None`

I have this line in to prevent it crashing on the next line when I try an access a property of `graph`

autogpt_platform/backend/backend/data/graph.py
Swiftyos: I've changed it the following, keeping the not null check in there:

```python
    # The Graph has to be owned by the user or a store listing.
    if graph and graph.userId != user_id and not (await StoreListingVersion.prisma().find_first(where={
        "agentId": graph_id,
        "agentVersion": version or graph.version,
    })):
        return None
```


> Or even further, why do we need ID for StoreListing, StoreListingVersion ?

I see your point on the `StoreListing` as we want it to be unique by agent we could just use the `agentId` as the PK

We don't force agents-version to be unique in the StoreListingVersion  table. So the creator can "modify" the store listing, by resubmitting the same agent-version with new store listing data i.e. new description etc.

I agree adding isApproved and isDelete is a good idea

autogpt_platform/backend/backend/data/graph.py
Swiftyos: UserId can still be None and it be okay as:

It would fail this part of the check `graph.userId != user_id` meaning the graph would only be returned if it's a store listing

autogpt_platform/backend/backend/data/graph.py
majdyz: I meant this last one:
https://github.com/Significant-Gravitas/AutoGPT/pull/9179/commits/714913a7097e0b7e2267b0f542a36d049b3eec5e

majdyz (approved):

autogpt_platform/backend/backend/data/graph.py
majdyz: @Swiftyos if you don't add mark this as a PK you definitely need to add an index (I mean for `StoreListing`).

> We don't force agents-version to be unique in the StoreListingVersion table. So the creator can "modify" the store listing, by resubmitting the same agent-version with new store listing data i.e. new description etc.

[agentId, agentVersion] StoreListing id and [agentId, agentVersion, version] StoreListingVersion.

While talking about index:

```
model StoreListingVersion {
  @@unique([agentId, agentVersion])
  @@index([agentId, agentVersion, isApproved])
}
```
* Index is not used if unique already present
* As you said the listing version should not be unique, so the unique should be [agentId, agentVersion, version] ? This can be problematic.
* Also if we did above, might as well remove the id and make it a PK.


Same with Store agent:
```
view StoreAgent {
  listing_id            String   @id
  storeListingVersionId String
  updated_at            DateTime

  slug        String
  agent_name  String
  agent_video String?
  agent_image String[]

  featured         Boolean  @default(false)
  creator_username String
  creator_avatar   String
  sub_heading      String
  description      String
  categories       String[]
  runs             Int
  rating           Float
  versions         String[]

  @@unique([creator_username, slug])
  @@index([creator_username])
  @@index([featured])
  @@index([categories])
  @@index([storeListingVersionId])
}
```
* creator_username index is not needed since we already have unique.
* index on only 1 boolean (featured) is not a good idea, this will not be used most of the time.
* Also index on categories which is a list, I can't imagine how it will be used.
* storeListingVersionId, shouldn't this What the query is look like? I think what you want is [`listing_id`, `storeListingVersionId`] ?

Swiftyos: I've gone through all the queries including the views and made sure the indexing matches what we are currently using.

autogpt_platform/backend/schema.prisma
```diff
@@ -553,7 +555,7 @@ model StoreListingVersion {
   StoreListingReview StoreListingReview[]
 
   @@unique([agentId, agentVersion])
-  @@index([agentId, agentVersion, isApproved])
+  @@index([agentId, agentVersion, isDeleted])
```
majdyz, line 558: AgentId and agent version is already unique

Adding another index on this will be useless and slowing down the write

autogpt_platform/backend/schema.prisma
```diff
@@ -510,9 +508,13 @@ model StoreListing {
   StoreListingVersions   StoreListingVersion[]
   StoreListingSubmission StoreListingSubmission[]
 
-  @@index([isApproved])
-  @@index([agentId])
+  // Unique index on agentId to ensure only one listing per agent, regardless of number of versions the agent has.
+  @@unique([agentId])
+  @@index([agentId, owningUserId])
```
majdyz, line 513: Same as below,

autogpt_platform/backend/schema.prisma
```diff
@@ -510,9 +508,13 @@ model StoreListing {
...
   @@index([owningUserId])
+  // Used in the view query
+  @@index([isDeleted, isApproved])
```
majdyz, line 516: These two indices with Boolean values will only be used on a query like this

`Select ... From StoreListing where is deleted = ...`

If you have another column in the filter, it will never be used. So I don't think this will be useful.

majdyz: If `[isDeleted, isApproved]` is indeed the filter column, the index with two columns will override the one below.
So we can safely remove it.

PR #9211, opened 2025-01-07
feat(platform): Add Support for Managing Agent Presets with Pagination and Soft Delete

Swiftyos:
#### Summary
- **New Models**: Added `LibraryAgentPreset`, `LibraryAgentPresetResponse`, `Pagination`, and `CreateLibraryAgentPresetRequest`.
- **Database Changes**:
  - Added `isDeleted` column in `AgentPreset` for soft delete.
  - CRUD operations for `AgentPreset`:
    - `get_presets` with pagination.
    - `get_preset` by ID.
    - `create_or_update_preset` for upsert.
    - `delete_preset` to soft delete.
- **API Routes**:
  - `GET /presets`: Fetch paginated presets.
  - `GET /presets/{preset_id}`: Fetch a single preset.
  - `POST /presets`: Create a preset.
  - `PUT /presets/{preset_id}`: Update a preset.
  - `DELETE /presets/{preset_id}`: Soft delete a preset.
- **Tests**:
  - Coverage for models, CRUD operations, and pagination.
- **Migration**:
  - Added `isDeleted` field to support soft delete.

#### Review Notes
- Validate migration scripts and test coverage.
- Ensure API aligns with project standards.

### DIFF
--- a/autogpt_platform/backend/backend/server/model.py
+++ b/autogpt_platform/backend/backend/server/model.py
@@ -56,3 +56,18 @@ class SetGraphActiveVersion(pydantic.BaseModel):
 
 class UpdatePermissionsRequest(pydantic.BaseModel):
     permissions: List[APIKeyPermission]
+
+
+class Pagination(pydantic.BaseModel):
+    total_items: int = pydantic.Field(
+        description="Total number of items.", examples=[42]
+    )
+    total_pages: int = pydantic.Field(
+        description="Total number of pages.", examples=[97]
+    )
+    current_page: int = pydantic.Field(
+        description="Current_page page number.", examples=[1]
+    )
+    page_size: int = pydantic.Field(
+        description="Number of items per page.", examples=[25]
+    )
--- a/autogpt_platform/backend/backend/server/v2/library/db.py
+++ b/autogpt_platform/backend/backend/server/v2/library/db.py
@@ -1,3 +1,4 @@
+import json
 import logging
 from typing import List
 
@@ -7,6 +8,7 @@ import prisma.types
 
 import backend.data.graph
 import backend.data.includes
+import backend.server.model
 import backend.server.v2.library.model
 import backend.server.v2.store.exceptions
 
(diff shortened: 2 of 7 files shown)


autogpt_platform/backend/backend/server/model.py
```diff
@@ -56,3 +56,18 @@ class SetGraphActiveVersion(pydantic.BaseModel):
...
+    )
+    total_pages: int = pydantic.Field(
+        description="Total number of pages.", examples=[97]
+    )
+    current_page: int = pydantic.Field(
+        description="Current_page page number.", examples=[1]
+    )
+    page_size: int = pydantic.Field(
+        description="Number of items per page.", examples=[25]
+    )
```
Pwuts: Nit: these numbers don't go together to make a realistic example.

42 items / 25 items per page = 2 pages total, or 97 pages * 25 items per page = 2400-2425 items total.

```suggestion
    total_items: int = pydantic.Field(
        description="Total number of items.", examples=[42]
    )
    total_pages: int = pydantic.Field(
        description="Total number of pages.", examples=[2]
    )
    current_page: int = pydantic.Field(
        description="Current_page page number.", examples=[1]
    )
    page_size: int = pydantic.Field(
        description="Number of items per page.", examples=[25]
    )
```

autogpt_platform/backend/backend/server/v2/library/db.py
```diff
@@ -163,3 +165,122 @@ async def add_agent_to_library(store_listing_version_id: str, user_id: str) -> N
...
+            where={"userId": user_id},
+        )
+        total_pages = (total_items + page_size - 1) // page_size
+
+        presets = [
+            backend.server.v2.library.model.LibraryAgentPreset.from_db(preset)
+            for preset in presets
+        ]
+
+        return backend.server.v2.library.model.LibraryAgentPresetResponse(
```
Pwuts, line 257: I'd personally very much prefer if you just did `from .model import LibraryAgentPreset, LibraryAgentPresetResponse` instead of this. These super long references aren't great for readability.

autogpt_platform/backend/backend/server/v2/library/routes.py
```diff
@@ -121,3 +121,89 @@ async def add_agent_to_library(
...
+    user_id: typing.Annotated[
+        str, fastapi.Depends(autogpt_libs.auth.depends.get_user_id)
+    ],
+) -> backend.server.v2.library.model.LibraryAgentPreset:
+    try:
+        return await backend.server.v2.library.db.create_or_update_preset(
+            user_id, preset, preset_id
+        )
+    except Exception:
+        logger.exception("Exception occurred whilst updating preset")
```
Pwuts: to make debugging possible:
```suggestion
    except Exception as e:
        logger.exception("Exception occurred whilst updating preset: {e}")
```

same for the rest of the endpoints

autogpt_platform/backend/backend/server/v2/library/model_test.py
Pwuts, line 1: Just curious: what's the reasoning behind these tests? It seems they just assert that Pydantic still works, which I don't completely understand (yet) scope-wise.

autogpt_platform/backend/backend/server/v2/library/db.py
Swiftyos: There are a few key advantages to this approach:
1. Clear Origin
When you see the full path, it’s immediately obvious where the function or class comes from—whether it’s part of our codebase or a third-party library. If we used from x import y, the code might read simply LibraryAgentPresetResponse, which doesn’t indicate if it’s defined in the current file, provided by us, or sourced from a third-party module.

2. Name Collision Prevention
Using fully qualified imports can help avoid naming clashes, which can occasionally occur and introduce subtle bugs. 

Overall, while you may find the fully qualified style more verbose, I believe the clarity and reduced risk of name collisions make it a better long-term choice. It's value defo becomes more apparent as code bases grows in size too.

It does also have a tangental benfit of encouraging short names for modules, which I feel  is best practice too.

autogpt_platform/backend/backend/server/model.py
Swiftyos: I really value your input, but I’d prefer if we could avoid minor nitpicks in my PRs. They can sometimes overshadow the main goal of the changes and slow our progress. Instead, it might be more efficient to tackle those small tweaks in a separate discussion or when you’re actively working on that area of the code. This way, we can keep each PR focused on its primary objectives while still ensuring all important feedback gets addressed.

autogpt_platform/backend/backend/server/v2/library/model_test.py
Swiftyos: These tests ensure alignment between database objects and Pydantic models, acting as a contract check to catch breaking changes early. When they fail in CI, it signals that something significant has changed, prompting a careful review to maintain data integrity and compatibility across the system.

autogpt_platform/backend/backend/server/v2/library/routes.py
Swiftyos: Thank you, I'm updating all the endpoints

autogpt_platform/backend/backend/server/v2/library/db.py
Pwuts: How much do these two issues affect your personal developer experience in practice? I personally don't recognize them to the extent that it justifies switching to inline fully qualified module references.

These extremely long references inflate the amount of information you need to process while reading through code. They also cause a lot of extra line breaks. Both of these are detrimental to code readability.

1. That can still be achieved with shorter imports. E.g. we could leave the relevant part of the fully qualified module name, like `model.LibraryAgentPreset` or `library.model.LibraryAgentPreset` which already improves readability and keeps the relevant part of the origin inline.

2. In practice I have only experienced name collisions a handful of times in the past year and a half, so imo that's not too much of an issue.

autogpt_platform/backend/backend/server/v2/library/model_test.py
Pwuts: How do they ensure alignment between database objects and Pydantic models? As far as I can see they just create Pydantic models and check their properties, without any involvement of the DB or its models or functions.

autogpt_platform/backend/backend/server/v2/library/db.py
Swiftyos: I feel strongly enough that this is a much more readable and better approach that I will always code in this style.

autogpt_platform/backend/backend/server/v2/library/model_test.py
Swiftyos: There are two parts to this. First, we need to make sure that important models are flagged when they change. The first two tests do this.

The third test checks if the database schema translation works.

Some people might say that linting should do the first two tests. But I prefer to have that extra security and make sure we write tests for everything.

Pwuts: > we need to make sure that important models are flagged when they change

I think it makes a lot of sense to do this for API models, so for the direct interface of the application. For other models it just seems like extra maintenance without added benefit since we have static type checking.

autogpt_platform/backend/backend/server/model.py
Pwuts: I understand your request, but don't think that'll work. Because of our current situation, only important feature requests and bugfixes are picked up, so if feedback on small things isn't raised on the initial PR, it often isn't picked up or addressed at all.

autogpt_platform/backend/backend/server/v2/library/db.py
Pwuts: Your opinion is valid. Still, it is a "deviation" from the code style that we've all been using so far, and I personally think it's not an improvement nor a trivial change in terms of DX. Also, code readability affects everyone working on this project, so using different patterns like this in *some* modules doesn't seem like a great idea either.

If you feel this is how we should be doing imports from now on, I suggest you make a proposal to the team.

Pwuts: I'll approve this PR for now pending discussion with the team, because we need the functionality

Pwuts (approved):

PR #9212, opened 2025-01-07
feat(platform) : Add api key generation frontend

Abhi1992002:
Allow users to create API keys for the AutoGPT platform. The backend is already set up, and here I’ve added the frontend for it.

### Changes
1. Fix the `response-model` of the API keys endpoints.  
2. Add a new page `/store/api_keys`.  
3. Add an `APIKeySection` component to create, delete, and view all your API keys.

Screenshot 2025-01-07 at 3 59 25 PM


Shadow062309:
GOOD IDEA

Swiftyos (approved):

PR #9217, opened 2025-01-08
feat(backend): Add graph/node id & execution id on CreditTransaction table

majdyz:
We need to be able to determine the cost of graph/node execution.

### Changes 🏗️

* Add these columns into CreditTransaction `metadata` column:
    - graph_id
    - node_id
    - graph_exec_id
    - node_exec_id
    - block_id
* Drop the `blockId` column and backfill the dropped value into metadata->>block_id.
* Frequent queries on these values will require an index created on demand through a migration, depending on the use case.


aarushik93 (dismissed):

autogpt_platform/backend/migrations/20250108003202_reshape_transaction_metadata_column/migration.sql
```diff
@@ -0,0 +1,18 @@
...
+  - You are about to drop the column `blockId` on the `CreditTransaction` table. All the data in the column will be moved to metadata->block_id.
+
+*/
+BEGIN;
+
+-- DropForeignKey blockId
+ALTER TABLE "CreditTransaction" DROP CONSTRAINT "CreditTransaction_blockId_fkey";
+
+-- Update migrate blockId into metadata->"block_id"
+UPDATE "CreditTransaction" SET "metadata" = jsonb_set("metadata"::jsonb, '{block_id}', to_jsonb("blockId")) WHERE "blockId" IS NOT NULL;
```
Pwuts:
```prisma
model CreditTransaction {
    // ...
    metadata Json?
}
```
Does `jsonb_set("metadata"::jsonb, ...)` work if `metadata` is currently `NULL`?

Pwuts: Robot says to use `COALESCE("metadata"::jsonb, '{}')`:
```suggestion
UPDATE "CreditTransaction"
SET "metadata" = jsonb_set(
    COALESCE("metadata"::jsonb, '{}'),
    '{block_id}',
    to_jsonb("blockId")
)
WHERE "blockId" IS NOT NULL;
```

Pwuts (commented):
I understand we can't make data out of thin air, but can we still use actual columns for `node_id`, `node_execution_id`? These columns can be optional for now, and I think we should be able to make them non-optional (and archive old `CreditTransactions` to a different place, e.g. a different table or some other system) after closing off the current fiscal year.

Pwuts:
Ack on this table not being for just node execution credit transactions, I was missing that bit of context.

My concern is that "informal" relationships which aren't stored in actual columns means we can't use Prisma to query them together afaik. If we can make that work with one of the solutions you proposed I'm happy.

kcze (approved):

Pwuts (approved):

ntindle (approved):

PR #9285, opened 2025-01-16
feat: mem0 ai memory block

ntindle:
We want to have some more intelligent and less user managed memory methods, so we add mem0

### Changes 🏗️

- Adds user_id to kwargs for blocks
- Add mem0 blocks

### Checklist 📋

- [x] document adding user_id to kwargs for blocks
- [x] Add run and agent Id as optional checkboxes that will be passed down to mem0

#### For code changes:
- [x] I have made a test plan
  
  - [ ] Build and submit an agent to @Torantulino and the marketplace for a personal AI tutor based on recommendations from the mem0 team


autogpt_platform/backend/backend/blocks/mem0.py
```diff
@@ -0,0 +1,296 @@
...
+}
+
+
+class Mem0Base:
+    """Base class with shared utilities for Mem0 blocks"""
+
+    @staticmethod
+    def _get_client(credentials: APIKeyCredentials) -> MemoryClient:
+        """Get initialized Mem0 client"""
+        return MemoryClient(api_key=credentials.api_key.get_secret_value())
```
Dev-Khant, line 37: Hey, here you can also have the option of passing Org/Project, [doc](https://docs.mem0.ai/api-reference/overview#organizations-and-projects-optional).

Although it's not necessary, if not passed then it will pick Org/Proj attached to the given API key.

autogpt_platform/backend/backend/blocks/mem0.py
```diff
@@ -0,0 +1,296 @@
...
+
+
+class AddMemoryBlock(Block, Mem0Base):
+    """Block for adding memories to Mem0
+
+    Always limited by user_id and optional graph_id, run_id, agent_id"""
```
Dev-Khant: We don't have `graph_id`, instead we have `app_id`.

autogpt_platform/backend/backend/blocks/mem0.py
```diff
@@ -0,0 +1,296 @@
...
+            test_mock={"_get_client": lambda credentials: MockMemoryClient()},
+        )
+
+    def run(
+        self,
+        input_data: Input,
+        *,
+        credentials: APIKeyCredentials,
+        user_id: str,
+        graph_id: str,
```
Dev-Khant, line 118: Same in context of Mem0, there is no graph_id.

autogpt_platform/backend/backend/blocks/mem0.py
```diff
@@ -0,0 +1,296 @@
...
+
+            filters: dict[str, list[dict[str, str | dict[str, list[str]]]]] = {
+                # This works with only one filter, so we can allow others to add on later
+                "AND": [
+                    {"user_id": user_id},
+                ]
+            }
+            if input_data.categories_filter:
+                filters["AND"].append(
+                    {"categories": {"in": input_data.categories_filter}}
```
Dev-Khant: Here instead of `in` it would be `contains`.
Please refer to the doc: https://docs.mem0.ai/platform/quickstart#search-using-custom-filters

autogpt_platform/backend/backend/blocks/mem0.py
```diff
@@ -0,0 +1,296 @@
...
+                "AND": [
+                    {"user_id": user_id},
+                ]
+            }
+            if input_data.limit_memory_to_run:
+                filters["AND"].append({"run_id": run_id})
+            if input_data.limit_memory_to_agent:
+                filters["AND"].append({"agent_id": graph_id})
+            if input_data.categories:
+                filters["AND"].append({"categories": {"in": input_data.categories}})
```
Dev-Khant: Same instead of `in` it would be `contains`

autogpt_platform/backend/pyproject.toml
```diff
@@ -55,6 +55,7 @@ sqlalchemy = "^2.0.36"
 psycopg2-binary = "^2.9.10"
 google-cloud-storage = "^2.18.2"
 launchdarkly-server-sdk = "^9.8.0"
+mem0ai = "^0.1.44"
```
Dev-Khant, line 58: This can be upgraded to latest version - `0.1.46`

docs/content/platform/new_blocks.md
```diff
@@ -102,6 +102,11 @@ Follow these steps to create and test a new block:
...
+   - **kwargs**: The `kwargs` parameter is used to pass additional arguments to the block. It is not used in the example above, but it is available to the block. You can also have args as inline signatures in the run method ala `def run(self, input_data: Input, *, user_id: str, **kwargs) -> BlockOutput:`.
+       Available kwargs are:
+       - `user_id`: The ID of the user running the block.
+       - `run_id`: The ID of the run that is executing the block. This changes every time the agent has a new "run"
+       - `graph_id`: The ID of the agent that is executing the block. This is the same for every version of the agent
```
Dev-Khant, line 108: In context of Mem0, it's better to rename it as `agent_id` as we have that param

autogpt_platform/backend/backend/blocks/mem0.py
ntindle: graph id is internal to us in this case. I'm mapping to agent id in your system

Our graph is an instance of an agent that can contain multiple versions so this is saying for any agent you can segment across version

Dev-Khant: Alright, got it. Thanks.

docs/content/platform/new_blocks.md
ntindle: this is a tbd when we fix it internally later -- mem0 will receive the same name

autogpt_platform/backend/backend/blocks/mem0.py
ntindle: this is a tbd when we fix it internally later -- mem0 will receive the same name

autogpt_platform/backend/backend/executor/manager.py
```diff
@@ -181,6 +181,11 @@ def update_execution(status: ExecutionStatus) -> ExecutionResult:
         credentials, creds_lock = creds_manager.acquire(user_id, credentials_meta.id)
         extra_exec_kwargs[field_name] = credentials
 
+    # Inject extra execution arguments for the blocks via kwargs
+    extra_exec_kwargs["user_id"] = user_id
+    extra_exec_kwargs["run_id"] = graph_exec_id
+    extra_exec_kwargs["graph_id"] = graph_id
```
majdyz: It's much clearer to have it as an initial value, also let's keep the graph_exec_id as graph_exec_id, since we also have node_exec_id:
https://github.com/Significant-Gravitas/AutoGPT/pull/9320/files#diff-1b278ebf10a9da0fb5030010222b3a6df2b05a5463cad428cd6c38a1541b0f73R173-R178

autogpt_platform/backend/backend/blocks/basic.py
```diff
@@ -471,6 +471,48 @@ def run(self, input_data: Input, **kwargs) -> BlockOutput:
         yield "updated_list", updated_list
 
 
+class FindInListBlock(Block):
+    class Input(BlockSchema):
+        list: List[Any] = SchemaField(description="The list to search in.")
```
majdyz, line 513: Does list[Any] actually rendered as a list ?

autogpt_platform/backend/backend/blocks/basic.py
```diff
@@ -636,3 +678,43 @@ def run(self, input_data: Input, **kwargs) -> BlockOutput:
             yield "value", converted_value
         except Exception as e:
             yield "error", f"Failed to convert value: {str(e)}"
+
+
+class TextSplitBlock(Block):
```
majdyz: This should live in text.py

autogpt_platform/backend/backend/blocks/mem0.py
```diff
@@ -0,0 +1,310 @@
...
+class AddMemoryBlock(Block, Mem0Base):
+    """Block for adding memories to Mem0
+
+    Always limited by user_id and optional graph_id, run_id, agent_id"""
+
+    class Input(BlockSchema):
+        credentials: CredentialsMetaInput[
+            Literal[ProviderName.MEM0], Literal["api_key"]
+        ] = CredentialsField(description="Mem0 API key credentials")
+        content: Union[str, list[dict[str, str]]] = SchemaField(
```
majdyz: How is this going to be rendered? in what format we should provide for this value

autogpt_platform/backend/backend/blocks/mem0.py
```diff
@@ -0,0 +1,310 @@
...
+        query: str = SchemaField(
+            description="Search query",
+            advanced=False,
+        )
+        trigger: bool = SchemaField(
+            description="An unused field that is used to (re-)trigger the block when you have no other inputs",
+            default=False,
+            advanced=False,
+        )
+        categories_filter: list[str] | None = SchemaField(
```
majdyz: Does it need to be none or we can just set [] as default.
If this is only for the same of making it non advance, we can set advanced=False

autogpt_platform/backend/backend/blocks/mem0.py
```diff
@@ -0,0 +1,310 @@
...
+        credentials: APIKeyCredentials,
+        user_id: str,
+        graph_id: str,
+        run_id: str,
+        **kwargs
+    ) -> BlockOutput:
+        try:
+            client = self._get_client(credentials)
+
+            filters: dict[str, list[dict[str, str | dict[str, list[str]]]]] = {
```
majdyz: type alias ?

autogpt_platform/backend/backend/executor/manager.py
ntindle: done

autogpt_platform/backend/backend/blocks/basic.py
ntindle: moved

autogpt_platform/backend/backend/blocks/basic.py
ntindle: image
yes

autogpt_platform/backend/backend/blocks/mem0.py
ntindle: image
it renders as this, but it should be a string or an ai block llm output. I think string is okay, but not sure how I'd improve this

autogpt_platform/backend/backend/blocks/mem0.py
ntindle: fixed by making advanced

ntindle: fixed

autogpt_platform/backend/backend/blocks/mem0.py
ntindle: done

autogpt_platform/backend/backend/blocks/mem0.py
majdyz: Then the input should be just str?

Or we can do list[str] and it will be rendereded as list input pin

majdyz (dismissed):

majdyz (approved):
Re-approved without re-review

PR #9295, opened 2025-01-19
fix(backend): Make monthly top-up adjust the target balance instead of top-up the target amount

majdyz:
To maintain the transaction running-balance integrity, the monthly top-up balance has to adjust the top-up amount to maintain the balance of the user to be at least X amount, instead of top-up the user balance with a fixed X amount.

### Changes 🏗️

* Add an additional query to sum the remaining un-snapshotted balance (if any).
* Monthly top-up transaction set the amount target_amount - current_balance instead of fixed target_amount.

#### For code changes:
- [ ] I have made a test plan
  
  - [ ] ...

<details>
  
</details>

<details>

  - Changing ports
</details>

### DIFF
--- a/autogpt_platform/backend/backend/data/credit.py
+++ b/autogpt_platform/backend/backend/data/credit.py
@@ -129,16 +129,16 @@ class UserCreditBase(ABC):
             },
         )
         transaction_balance = (
-            transactions[0].get("_sum", {}).get("amount", 0) if transactions else 0
+            transactions[0].get("_sum", {}).get("amount", 0) + snapshot_balance
+            if transactions
+            else snapshot_balance
         )
         transaction_time = (
             transactions[0].get("_max", {}).get("createdAt", datetime_min)
             if transactions
-            else datetime_min
+            else snapshot_time
         )
-        return snapshot_balance + transaction_balance, max(
-            snapshot_time, transaction_time
-        )
+        return transaction_balance, transaction_time
 
     async def _enable_transaction(
         self, transaction_key: str, user_id: str, metadata: Json
--- a/autogpt_platform/backend/backend/data/credit.py
+++ b/autogpt_platform/backend/backend/data/credit.py
@@ -399,19 +399,16 @@ class BetaUserCredit(UserCredit):
         if (snapshot_time.year, snapshot_time.month) == (cur_time.year, cur_time.month):
             return balance
 
-        top_up_amount = max(self.num_user_credits_refill - balance, 0)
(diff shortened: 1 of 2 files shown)


autogpt_platform/backend/backend/data/credit.py
```diff
@@ -389,22 +400,15 @@ async def get_credits(self, user_id: str) -> int:
...
-                    "metadata": Json({}),
-                    "isActive": True,
-                    "createdAt": self.time_now(),
-                }
+            return await self._add_transaction(
+                user_id=user_id,
+                amount=max(self.num_user_credits_refill - balance, 0),
+                transaction_type=CreditTransactionType.TOP_UP,
+                transaction_key=f"MONTHLY-CREDIT-TOP-UP-{cur_time}",
             )
```
kcze, line 408: Unfortunately, this won't allow fair top-ups for beta users, because top-ups will decrease grant amount.
If you want; it's possible to `sum` up negative transactions from last month and give a grant of `min(abs(sum), max_grant_amount)`. This will top up up to a max grant amount the money you spent last month and still allow for top-ups.
```py
async def calculate_replenishment_amount(self, user_id: str) -> int:
    """Calculate how much to replenish based on last month's negative transactions."""
    cur_time = self.time_now()
    last_month_start = (cur_time.replace(day=1) - timedelta(days=1)).replace(day=1, hour=0, minute=0, second=0, microsecond=0)
    last_month_end = cur_time.replace(day=1, hour=0, minute=0, second=0, microsecond=0) - timedelta(microseconds=1)
    
    # Sum all negative transactions (spends) from last month
    spent_transactions = await CreditTransaction.prisma().group_by(
        by=["userId"],
        sum={"amount": True},
        where={
            "userId": user_id,
            "isActive": True,
            "amount": {"lt": 0},  # Only negative transactions
            "createdAt": {
                "gte": last_month_start,
                "lte": last_month_end
            }
        }
    )
    
    # Get absolute value of spent amount (or 0 if no spends)
    spent_amount = abs(spent_transactions[0].get("_sum", {}).get("amount", 0)) if spent_transactions else 0
    
    # Return minimum of spent amount and grant value
    return min(spent_amount, self.num_user_credits_refill)
```

majdyz: I don't think this matters, we will stop the beta not more than 1 month after release. And even if it is, we are not guaranteeing anything about giving them constant 1500points for beta, we only want them have at least 1500 for testing.

kcze (approved):

PR #9308, opened 2025-01-21
feat: add screenshotone block

ntindle:
I want to be able to screenshot things

### Changes 🏗️

Adds ScreenshotOne blocks with the ability to screenshot stuff

#### For code changes:
- [x] I have made a test plan
  
  - [x] Built agent with it


autogpt_platform/backend/backend/blocks/screenshotone.py
```diff
@@ -0,0 +1,161 @@
...
+    APIKeyCredentials,
+    CredentialsField,
+    CredentialsMetaInput,
+    SchemaField,
+)
+from backend.integrations.providers import ProviderName
+from backend.util.request import Requests
+
+
+class ScreenshotOneBlock(Block):
```
majdyz: Should we name it ScreenshotWebPageBlock or something ?

autogpt_platform/backend/backend/blocks/screenshotone.py
```diff
@@ -0,0 +1,161 @@
...
+        block_chats: bool = SchemaField(
+            description="Whether to block chat widgets", default=True
+        )
+        cache: bool = SchemaField(
+            description="Whether to enable caching", default=False
+        )
+
+    class Output(BlockSchema):
+        image_data: bytes = SchemaField(description="The screenshot image data")
+        content_type: str = SchemaField(description="The MIME type of the image")
```
majdyz: This would go really well with https://github.com/Significant-Gravitas/AutoGPT/pull/9320
The content_type & image_data will be represented as files using the `store_temp_file` function. It's also good to avoid persisting base64 content in the DB.

Can we add a to do? and linking the PR to update it in case this PR goes first?

autogpt_platform/backend/backend/blocks/screenshotone.py
```diff
@@ -0,0 +1,161 @@
...
+            "block_cookie_banners": str(block_cookie_banners).lower(),
+            "block_chats": str(block_chats).lower(),
+            "cache": str(cache).lower(),
+        }
+
+        response = api.get("https://api.screenshotone.com/take", params=params)
+
+        return {
+            "image_data": b64encode(response.content).decode("utf-8"),
+            "content_type": f"image/{format}",
```
majdyz: e.g: replacing this to 
return {
    "image": store_temp_file(f"data:{image/{format}};base64,{b64encode(response.content).decode("utf-8")})
}

majdyz (approved):

autogpt_platform/backend/backend/blocks/screenshotone.py
ntindle: yes, we should

autogpt_platform/backend/backend/blocks/screenshotone.py
ntindle: Done

autogpt_platform/backend/backend/blocks/screenshotone.py
```diff
@@ -0,0 +1,166 @@
...
+            description="Whether to capture the full page length", default=False
+        )
+        format: str = SchemaField(
+            description="Output format (png, jpeg, webp)", default="png"
+        )
+        block_ads: bool = SchemaField(description="Whether to block ads", default=True)
+        block_cookie_banners: bool = SchemaField(
+            description="Whether to block cookie banners", default=True
+        )
+        block_chats: bool = SchemaField(
```
majdyz, line 52: Should we add a Boolean option whether to return the image (base64) or return the Image file (relative path) to be passed around? We can make it non advanced default to return path. And user can toggle it if they want to see it on the builder.

majdyz (approved):

autogpt_platform/backend/backend/blocks/screenshotone.py
ntindle: We should probably show on ui under both conditions to be honest

autogpt_platform/backend/backend/util/test.py
```diff
@@ -104,7 +105,12 @@ def execute_block_test(block: Block):
             log.info(f"{prefix} mock {mock_name} not found in block")
 
     # Populate credentials argument(s)
-    extra_exec_kwargs = {}
+    extra_exec_kwargs: dict = {
```
aarushik93, line 108: Could you explain this change pls?

autogpt_platform/backend/backend/blocks/screenshotone.py
```diff
@@ -0,0 +1,166 @@
...
+            description="Whether to block chat widgets", default=True
+        )
+        cache: bool = SchemaField(
```
aarushik93, line 55: general q, what does caching do in this block?

aarushik93:
will this display the screenshot on the UI too?

autogpt_platform/backend/backend/blocks/screenshotone.py
ntindle: caching happens on the sc side and will only regrab the photo if cache is set to false. Otherwise they will use the existing to save $$

ntindle: cache happens on their side so they will pull from the cache if they've already grabbed an image saving you $$

autogpt_platform/backend/backend/util/test.py
ntindle: Basically we need user details in the block as arguements so we can slice memory automatically into parts. This injects those variables into the block

aarushik93 (approved):

PR #9331, opened 2025-01-24
Use .env vars in docker-compose.platform.yml

elabbarw:
The compose file was not using the environment variables . This update changes all the hardcoded passwords, etc to properly use the .env file that should reside in the same folder.

### Changes 🏗️

updated autogpt_platform/docker-compose.platform.yml to use ${ENV VARS} instead of the hardcoded postgres password ,etc.. 

- [x ] `.env.example` is updated or already compatible with my changes
- [ x] `docker-compose.yml` is updated or already compatible with my changes
- [ x] I have included a list of my configuration changes in the PR description (under **Changes**)

<details>

  - Modifying compose file to make use of .env vars
  - 
</details>

### DIFF
--- a/autogpt_platform/frontend/src/components/agptui/composite/PublishAgentPopout.tsx
+++ b/autogpt_platform/frontend/src/components/agptui/composite/PublishAgentPopout.tsx
@@ -263,7 +263,7 @@ export const PublishAgentPopout: React.FC<PublishAgentPopoutProps> = ({
                   onClose={handleClose}
                   onDone={handleClose}
                   onViewProgress={() => {
-                    router.push("/marketplace/dashboard");
+                    router.push("/profile/dashboard");
                     handleClose();
                   }}
                 />


ntindle:
Any specific reason for closing this? Was in the list to review

elabbarw:
@ntindle reopened and updated PR.

autogpt_platform/docker-compose.platform.yml
```diff
@@ -26,7 +26,7 @@ services:
 
   redis:
     image: redis:latest
-    command: redis-server --requirepass password
+    command: redis-server --requirepass ${REDIS_PASSWORD}
```
aarushik93: Coudl you please add this to the .env.example

elabbarw: The .env.example comes from Supabase which is why i added the comment.

aarushik93: yeah we a have few diff ones unfortunately - https://github.com/Significant-Gravitas/AutoGPT/blob/master/autogpt_platform/backend/.env.example this is the one I meant

aarushik93: so if this only reads from the supabase one then I think we need to put the hardcoded one back in here

elabbarw: Ah, I see! Sure, I'll get this sorted.

elabbarw:
Hey @aarushik93 , can you resolve this?

PR #9338, opened 2025-01-26
fix: check if linear is enabled before showing blocks

ntindle:
Linear blocks show up even if oauth isn't configured

### Changes 🏗️
disables the linear blocks if oauth isn't configured

#### For code changes:
- [x] I have made a test plan
  
  - [x] checked block list with and without oauth configured


majdyz (approved):
LGTM 
So we are not allowing user manually configuring the linear creds?

PR #9350, opened 2025-01-28
fix(backend): Strip credential and user IDs from store agent download

Abhi1992002:
### Changes 🏗️

- Remove FastAPI exception raises from the `get_agent` database function.
- Enable the `get_agent` function to include the `user_id` in the response.
- Implement optional graph_data handling for the `/download/agents/{store_listing_version_id}` endpoint.
- Remove the `user_id` from the graph data response for the `/download/agents/{store_listing_version_id}` endpoint.

#### For code changes:
- [x] I have made a test plan
- [x] I have manually checked the downloading agent functionality


autogpt_platform/backend/backend/server/v2/store/routes.py
```diff
@@ -620,6 +625,8 @@ def remove_credentials(obj):
                 del obj["credentials"]
             if "creds" in obj:
```
ntindle, line 631: This should probably be a greater utility function that gets called :) 

we can probably also strip anything that has secret = true

kcze (changes requested):
Please check and update frontend types if applicable.

Swiftyos (approved):
Generally a good idea to push raises out the route layer.

Fix the merge conflicts and I'll reapprove

autogpt_platform/backend/backend/server/v2/store/db.py
```diff
@@ -765,28 +764,20 @@ async def get_agent(
         )
 
         if not store_listing_version or not store_listing_version.Agent:
-            raise fastapi.HTTPException(
-                status_code=404,
-                detail=f"Store listing version {store_listing_version_id} not found",
-            )
+            logger.error(f"Store listing version {store_listing_version_id} not found")
```
Pwuts, line 767: whether this is an error condition is up to the caller of `get_agent`, no need to treat/log it as one here

autogpt_platform/backend/backend/server/v2/store/db.py
```diff
@@ -765,28 +764,20 @@ async def get_agent(
...
+            return None
 
         graph_id = store_listing_version.agentId
         graph_version = store_listing_version.agentVersion
         graph = await backend.data.graph.get_graph(graph_id, graph_version)
```
Pwuts, line 772: If you change the include statement on line 762 to `include={"Agent": AGENT_GRAPH_INCLUDE}` then you can omit this second query and the check below.

autogpt_platform/backend/backend/server/v2/store/db.py
```diff
@@ -765,28 +764,20 @@ async def get_agent(
...
 
         if not graph:
-            raise fastapi.HTTPException(
-                status_code=404,
-                detail=(
-                    f"Agent #{graph_id} not found "
-                    f"for store listing version #{store_listing_version_id}"
-                ),
-            )
+            logger.error(f"Agent {agent.id} not found")
```
Pwuts, line 775: If you leave this check it should be a `raise DatabaseError`, since a `StoreListingVersion` can't not have an associated `AgentGraph` afaik

autogpt_platform/backend/backend/server/v2/store/routes.py
Pwuts: Even better: there already is a utility function for this: `GraphModel.from_db(..., for_export=True)`
https://github.com/Significant-Gravitas/AutoGPT/blob/f9d88d2981002e9c0a4831d038fe8732629902aa/autogpt_platform/backend/backend/data/graph.py#L459-L511

Pwuts:
I'm working on something similar in #9627

PR #9380, opened 2025-01-30
fix(backend): fix missing agent object requirement

ntindle:
We oops'd and didn't return the agent as a downloadable

### Changes 🏗️
- attaches agent object

### DIFF
--- a/autogpt_platform/backend/backend/server/v2/store/db.py
+++ b/autogpt_platform/backend/backend/server/v2/store/db.py
@@ -809,12 +809,11 @@ async def get_agent(
     try:
         store_listing_version = (
             await prisma.models.StoreListingVersion.prisma().find_unique(
-                where={"id": store_listing_version_id},
-                include={"Agent": True},
+                where={"id": store_listing_version_id}
             )
         )
 
-        if not store_listing_version or not store_listing_version.Agent:
+        if not store_listing_version:
             raise fastapi.HTTPException(
                 status_code=404,
                 detail=f"Store listing version {store_listing_version_id} not found",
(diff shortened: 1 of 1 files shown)


autogpt_platform/backend/backend/server/v2/store/db.py
```diff
@@ -809,7 +809,8 @@ async def get_agent(
     try:
         store_listing_version = (
             await prisma.models.StoreListingVersion.prisma().find_unique(
-                where={"id": store_listing_version_id}
+                where={"id": store_listing_version_id},
+                include={"Agent": True},
```
Swiftyos, line 813: By not including it, we’re intentionally keeping the store listings implementation under wraps. All the necessary details are already stored in the store tables.

Swiftyos: By not including it, we’re intentionally keeping the store listings implementation under wraps. All the necessary details are already stored in the store tables.

Swiftyos (changes requested):

autogpt_platform/backend/backend/server/v2/store/db.py
Swiftyos: Can you elaborate on why you think it’s important to include this?

ntindle:
```py
        if not store_listing_version or not store_listing_version.Agent:
            raise fastapi.HTTPException(
                status_code=404,
                detail=f"Store listing version {store_listing_version_id} not found",
            )
```

this segment checks if it exists explicitly and raises if not.  How do we want to handle this? Issue here is visible when 
- logged out
- Navigate to store agent
- Hit download

Swiftyos (approved):

PR #9386, opened 2025-01-31
Docs migration from Mintlify

ruhanponnada:
This is Mintlify's Docs Migration!
I've deleted the previous files in this folder and added the content of the migration.

To run this locally install the Mintlify CLI
`npm i -g mintlify`
then 
`mintlify dev`


docs/platform/new_blocks.mdx
```diff
@@ -0,0 +1,1442 @@
...
+response = requests.post(
+    url,
+    headers={"Authorization": credentials.bearer()},
+)
+```
+
+The `ProviderName` enum is the single source of truth for which providers exist in our system.
+Naturally, to add an authenticated block for a new provider, you'll have to add it here too.
+
+<Accordion
```
ntindle, line 297: Is this still reading from the codebase to dynamically import it?

kcze:
What is the purpose of this migration? Do we need this?

PR #9424, opened 2025-02-05
platform(fix): Fix missing Profiles

Swiftyos:
### What's This PR About?

This PR makes a few simple improvements to how user profiles are handled in the app:

- **Always Have a Profile:**  
  If a user doesn't already have a profile, the system now automatically creates one with some default info (including a fun, randomly generated username). This way, you never end up with a missing profile.

- **Better Profile Updates:**  
  Removes the creation of profiles on failed get requests


Swiftyos:
> ## PR Reviewer Guide 🔍
> Here are some key observations to aid the review process:
> 
> ⏱️ **Estimated effort to review**: 3 🔵🔵🔵⚪⚪
> 🧪 **PR contains tests**
> 🔒 **Security concerns**
> 
> **SQL injection:** The migrations and trigger functions directly concatenate user input (email) into SQL statements when creating profiles. While the risk is low since the email comes from auth system, it would be safer to use parameterized queries or proper escaping.
> 
> ⚡ **Recommended focus areas for review**
> 
> [**Error Handling**](https://github.com/Significant-Gravitas/AutoGPT/pull/9424/files#diff-0ae74821d3efda630442bf8fed250da69c0c414481a8dfb8ba5b8c6c90ba6277R639-R641)
> The update_profile function raises ProfileNotFoundError with an incorrect string format that could cause errors. The error message uses %s but doesn't provide the user_id argument correctly.
> 
> [**Unique Constraint**](https://github.com/Significant-Gravitas/AutoGPT/pull/9424/files#diff-c4b461751c2805fb2d883e0c7bd5f3470dfb2526dbab669d7a1a0188229cfffdR1-R73)
> The trigger function generates random usernames but doesn't verify uniqueness before insertion. This could lead to username collisions and failed inserts as the system scales.

I have addressed these concerns

autogpt_platform/backend/backend/server/v2/store/db.py
```diff
@@ -623,115 +601,90 @@ async def get_user_profile(
...
-            if profile.links is not None:
-                update_data["links"] = profile.links
-            if profile.avatar_url is not None:
-                update_data["avatarUrl"] = profile.avatar_url
-
-            # Update the existing profile
-            updated_profile = await prisma.models.Profile.prisma().update(
-                where={"id": existing_profile.id},
-                data=prisma.types.ProfileUpdateInput(**update_data),
+            raise fastapi.HTTPException(
```
aarushik93: I don't think we should be raising HTTPExceptions here, in the DB layer. Can you do another error here?

autogpt_platform/frontend/src/app/signup/actions.ts
```diff
@@ -38,7 +38,7 @@ export async function signup(values: z.infer<typeof signupFormSchema>) {
       }
       console.log("Signed up");
```
majdyz: I know it's unrelated to the PR but can we remove it ?

autogpt_platform/backend/backend/server/v2/store/db.py
```diff
@@ -17,6 +16,25 @@
 logger = logging.getLogger(__name__)
 
 
+def sanitize_query(query: str | None) -> str | None:
```
aarushik93, line 19: what's the usecase for this one? Are you trying to avoid sql injection or is there another reason?

autogpt_platform/frontend/src/app/signup/actions.ts
```diff
@@ -38,7 +38,7 @@ export async function signup(values: z.infer<typeof signupFormSchema>) {
...
       revalidatePath("/", "layout");
-      redirect("/marketplace/profile");
+      redirect("/");
```
majdyz, line 40: > After signing up, users are now redirected directly to their profile page instead of the homepage. This helps new users immediately see and edit their profile if needed.

Wait, Isn't this the reverse?

autogpt_platform/backend/backend/server/v2/store/db.py
Swiftyos: I don’t think it’s strictly necessary, but the Deepsource code quality checks flag it, so I added it to avoid SQL injection.

The change in this PR is to move it into a new function to make the get_agent function simpler. Again, for the code quality checks.

autogpt_platform/backend/backend/server/v2/store/db.py
Swiftyos: Sure, done

aarushik93 (approved):

autogpt_platform/backend/backend/server/v2/store/db.py
```diff
@@ -29,29 +47,7 @@ async def get_store_agents(
...
-            .replace("[", "\\[")
-            .replace("]", "\\]")
-            .replace("'", "\\'")
-            .replace('"', '\\"')
-            .replace(";", "\\;")
-            .replace("--", "\\--")
-            .replace("/*", "\\/*")
-            .replace("*/", "\\*/")
-        )
+    sanitized_query = sanitize_query(search_query)
```
majdyz, line 50: Why do we need to sanitize this, isn't Prisma already does this?

autogpt_platform/backend/migrations/20250205100104_add_profile_trigger/migration.sql
```diff
@@ -0,0 +1,108 @@
...
+        IF NOT EXISTS (
+            SELECT 1 FROM platform."Profile" WHERE username = generated_username
+        ) THEN
+            -- Username is unique, exit the loop
+            EXIT;
+        END IF;
+
+        -- If we've tried 10 times and still haven't found a unique username
+        IF i = 10 THEN
+            RAISE EXCEPTION 'Unable to generate unique username after 10 attempts';
```
majdyz, line 38: Do we really need to have a clever name here? can we randomize it and make this query easier to understand?
Is this for all users or only for cornercases ? won't the user need to be asked to fill these detail during the signup ?

autogpt_platform/frontend/src/app/signup/actions.ts
Swiftyos: yes no problem

autogpt_platform/frontend/src/app/signup/actions.ts
Swiftyos: So, the reason for redirecting to the profile was to automatically create a profile entry when the get_profile function was called on page load.

But I’ve got a new trigger now that automatically generates their profile. 

They can still edit it whenever they want, but I thought it would be a bit more intuitive if they went straight to a product page instead.

autogpt_platform/backend/backend/server/v2/store/db.py
Swiftyos: Please see my response to the [earlier question](https://github.com/Significant-Gravitas/AutoGPT/pull/9424#discussion_r1942983598) about it:

I don’t think it’s strictly necessary, but the Deepsource code quality checks flag it, so I added it to avoid SQL injection.

The change in this PR is to move it into a new function to make the get_agent function simpler. Again, for the code quality checks.

autogpt_platform/backend/migrations/20250205100104_add_profile_trigger/migration.sql
Swiftyos: No we don't. However, I felt it looked nicer than displaying a hash, currently we don't have a flow where they are asked to enter their username during signup.

autogpt_platform/backend/backend/server/v2/store/db.py
```diff
@@ -623,115 +601,91 @@ async def get_user_profile(
...
+            update_data["description"] = profile.description
+        if profile.links is not None:
+            update_data["links"] = profile.links
+        if profile.avatar_url is not None:
+            update_data["avatarUrl"] = profile.avatar_url
+
+        # Update the existing profile
+        updated_profile = await prisma.models.Profile.prisma().update(
+            where={"id": existing_profile.id},
+            data=prisma.types.ProfileUpdateInput(**update_data),
```
majdyz, line 667: I this code pass the linter?
You can set `prisma.types.ProfileUpdateInput` as the type of update_data, so you will get the type hint for the code above. e.g: you can avoid setting `update_data` with an invalid key.

autogpt_platform/backend/backend/server/v2/store/db.py
```diff
@@ -623,115 +601,91 @@ async def get_user_profile(
...
+            links=updated_profile.links,
+            avatar_url=updated_profile.avatarUrl or "",
+            agent_rating=0.0,
+            agent_runs=0,
+            top_categories=[],
+        )
+
     except prisma.errors.PrismaError as e:
-        logger.error(f"Database error updating profile: {str(e)}")
+        logger.error("Database error updating profile: %s", str(e))
```
majdyz: str(e) can just be e

autogpt_platform/backend/backend/server/v2/store/db.py
```diff
@@ -623,115 +601,91 @@ async def get_user_profile(
...
-            avatar_url="",
-        )
+        logger.error("Error getting user profile: %s", str(e))
+        raise backend.server.v2.store.exceptions.DatabaseError(
+            "Failed to get user profile"
+        ) from e
 
 
-async def update_or_create_profile(
+async def update_profile(
```
majdyz, line 610: If this function is indeed create a profile, why change the name ?

autogpt_platform/frontend/src/app/signup/actions.ts
majdyz: I see, then we need to update the PR description because it's stating the reverse.

autogpt_platform/backend/migrations/20250205100104_add_profile_trigger/migration.sql
majdyz: I see, not super important, but it would be cleaner to create function for this name generation, since now this trigger and the migration below use different logic for generating the name, this has more name while others only 5x5

autogpt_platform/backend/backend/server/v2/store/db.py
```diff
@@ -623,115 +601,91 @@ async def get_user_profile(
...
         profile: Updated profile details
-
     Returns:
-        CreatorDetails: The updated profile
-
+        CreatorDetails: The updated or created profile details
     Raises:
-        HTTPException: If user is not authorized to update this profile
-        DatabaseError: If profile cannot be updated due to database issues
+        HTTPException: If the user is not authorized to update the profile
```
majdyz: I don't see HTTPException is used at all, can we remove this ?

majdyz (dismissed):

autogpt_platform/backend/backend/server/v2/store/db.py
Swiftyos: I removed creating as it is no longer needed

autogpt_platform/backend/backend/server/v2/store/db.py
Swiftyos: It does pass the linter. I did this to allow dynamically creating the update_data.

autogpt_platform/backend/migrations/20250205100104_add_profile_trigger/migration.sql
Swiftyos: I've moved it too a new function

aarushik93 (approved):

autogpt_platform/backend/backend/server/v2/store/db.py
majdyz: You can set the type hint of ProfileUpdateInput to dict too

majdyz (approved):

PR #9435, opened 2025-02-06
platform(fix): Improve performance of builder

Swiftyos:
1. Remove isHovered / onMouseEnter / onMouseLeave state updates
2. Wrap Custom Node in React.memo
3. Avoid re-renders for context menus

### DIFF
--- a/autogpt_platform/frontend/src/components/CustomEdge.tsx
+++ b/autogpt_platform/frontend/src/components/CustomEdge.tsx
@@ -40,7 +40,6 @@ export function CustomEdge({
   targetY,
   markerEnd,
 }: EdgeProps<CustomEdge>) {
-  const [isHovered, setIsHovered] = useState(false);
   const [beads, setBeads] = useState<{
     beads: Bead[];
     created: number;
@@ -182,13 +181,7 @@ export function CustomEdge({
       <BaseEdge
         path={svgPath}
         markerEnd={markerEnd}
-        style={{
-          strokeWidth: (isHovered ? 3 : 2) + (data?.isStatic ? 0.5 : 0),
-          stroke:
-            (data?.edgeColor ?? "#555555") +
-            (selected || isHovered ? "" : "80"),
-          strokeDasharray: data?.isStatic ? "5 3" : "0",
-        }}
+        className={`transition-all duration-200 ${data?.isStatic ? "[stroke-dasharray:5_3]" : "[stroke-dasharray:0]"} [stroke-width:${data?.isStatic ? 2.5 : 2}px] hover:[stroke-width:${data?.isStatic ? 3.5 : 3}px] ${selected ? `[stroke:${data?.edgeColor ?? "#555555"}]` : `[stroke:${data?.edgeColor ?? "#555555"}80] hover:[stroke:${data?.edgeColor ?? "#555555"}]`}`}
       />
       <path
         d={svgPath}
@@ -196,8 +189,6 @@ export function CustomEdge({
         strokeOpacity={0}
         strokeWidth={20}
         className="react-flow__edge-interaction"
-        onMouseEnter={() => setIsHovered(true)}
-        onMouseLeave={() => setIsHovered(false)}
       />
       <EdgeLabelRenderer>
         <div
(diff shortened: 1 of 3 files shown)


autogpt_platform/frontend/src/components/CustomNode.tsx
```diff
@@ -92,812 +92,826 @@ export type CustomNodeData = {
...
-    throw new Error("FlowContext consumer must be inside FlowEditor component");
-  }
-
-  const { setIsAnyModalOpen, getNextNodeId } = flowContext;
-
-  useEffect(() => {
-    if (data.executionResults || data.status) {
-      setIsOutputOpen(true);
+export const CustomNode = React.memo(
+  function CustomNode({
```
kcze, line 96: This wouldn't cause such a massive diff and likely merge conflicts:
```suggestion
export const CustomNode = React.memo(CustomNodeInternal);

function CustomNodeInternal({
```

autogpt_platform/frontend/src/components/CustomEdge.tsx
```diff
@@ -182,22 +181,14 @@ export function CustomEdge({
...
         path={svgPath}
         markerEnd={markerEnd}
-        style={{
-          strokeWidth: (isHovered ? 3 : 2) + (data?.isStatic ? 0.5 : 0),
-          stroke:
-            (data?.edgeColor ?? "#555555") +
-            (selected || isHovered ? "" : "80"),
-          strokeDasharray: data?.isStatic ? "5 3" : "0",
-        }}
+        className={`transition-all duration-200 ${data?.isStatic ? "[stroke-dasharray:5_3]" : "[stroke-dasharray:0]"} [stroke-width:${data?.isStatic ? 2.5 : 2}px] hover:[stroke-width:${data?.isStatic ? 3.5 : 3}px] ${selected ? `[stroke:${data?.edgeColor ?? "#555555"}]` : `[stroke:${data?.edgeColor ?? "#555555"}80] hover:[stroke:${data?.edgeColor ?? "#555555"}]`}`}
```
kcze, line 184: imo this is worse to process than the previous version but I appreciate that you want to use tailwind consistently and remove hover state, maybe at least make this multiline

kcze (changes requested):

autogpt_platform/frontend/src/components/CustomEdge.tsx
Swiftyos: The key difference here is that we don't need to register a listener and use javascript to make style changes based on the hover state. 

My understanding is that using css hover is considerably faster - this was the number 1 performance issue, that was suggested. In terms of human readability, what you had done is considerable nicer.

However, I agree that this change has an small impact on performance, but I suspect our performance issues are caused by many small things adding up to a big thing.

autogpt_platform/frontend/src/components/CustomNode.tsx
Swiftyos: Probably not a massive difference but every little helps. Are there other changes to the graph going on atm that you expect this will conflict with?

kcze: I don't know, I know my version looks worse but this is unnecessary diff and nesting.

autogpt_platform/frontend/src/components/CustomEdge.tsx
kcze: I agree that your way is likely faster. If linter doesn't complain about such as long line then that's fine but I'd still make multiline if possible.

autogpt_platform/frontend/src/components/CustomNode.tsx
Swiftyos: > Even after removing hover states, CustomNode can still be expensive if it re-renders frequently. You can wrap it with React.memo (or useMemo around large sub-sections) so that it only re-renders when its props.data really changes

This is the advice from o1, we can remove it if you know its not useful

autogpt_platform/frontend/src/components/CustomEdge.tsx
Swiftyos: Okay, happy for you to take this PR over and make the changes as @Torantulino has told me to focus on the library. It doesn't look like the linter has any line length checks

ntindle (approved):
honestly better arch overall with dot too

PR #9444, opened 2025-02-07
feat(platform): Schedule specific agent version

kcze:
Scheduling always takes the newest version of an agent.

### Changes 🏗️

This PR allows to schedule any graph version by adding number input to schedule popup. Number is automatically set to the newest version when agent is chosen.

Screenshot 2025-02-07 at 5 05 56 PM

- Update API, so it accepts graph version
- Update schedule pop up, so it lets user input version number
- Open and schedule correct agent
- Add `Version` column to the schedules table

#### For code changes:
- [x] I have made a test plan
  - [x] Can schedule version between 1 and max version
  - [x] Reject incorrect version
  - [x] Table shows proper version
  - [x] Removing schedule works

### DIFF
--- a/autogpt_platform/frontend/src/components/monitor/scheduleTable.tsx
+++ b/autogpt_platform/frontend/src/components/monitor/scheduleTable.tsx
@@ -31,6 +31,8 @@ import {
   DialogTitle,
 } from "@/components/ui/dialog";
 import { TextRenderer } from "../ui/render";
+import { Input } from "../ui/input";
+import { Label } from "../ui/label";
 
 interface SchedulesTableProps {
   schedules: Schedule[];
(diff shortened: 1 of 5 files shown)


autogpt_platform/frontend/src/components/monitor/scheduleTable.tsx
```diff
@@ -198,6 +240,7 @@ export const SchedulesTable = ({
                     {agents.find((a) => a.id === schedule.graph_id)?.name ||
```
majdyz, line 240: Should we just append the version here? e.g: {graph_id}#V{graph_version} instead of a new column ?

majdyz (approved):

autogpt_platform/frontend/src/components/monitor/scheduleTable.tsx
kcze: I'll leave it as is - the Library/Home page is getting refresh anyway.

PR #9445, opened 2025-02-07
feat(backend): schema updates, migration, queries for Email Notification Service

ntindle:
The email service has requirements to
- Email users when some activity has happened on their account on some scheduled basis -> We need a way to get active users and the executions that happened while they were active
- Allow users to configure what emails they get -> Need a user preference
- Get User email by Id so that we can email them -> Pretty self-explanatory

We need to add a few new backend queries + db models for the notification to start handling these details. This is the first set of those changes based on experience building the app service

### Changes 🏗️

- Adds a new DB Model, `UserNotificationPreferences,` with related migration
- Adds queries to get users and executions by `datetime` ranges as `ISO` strings
- Adds new queries to the `DatabaseManager` and exposes them to the other `AppService`s
- Exposes all new queries plus an existing one `get_user_by_id`

#### For code changes:
- [x] I have made a test plan
  
  - [x] I extracted these changes from a working implementation of the service, and tested they don't bring down the service by being uncalled

### DIFF
--- a/autogpt_platform/backend/backend/data/execution.py
+++ b/autogpt_platform/backend/backend/data/execution.py
@@ -364,6 +364,26 @@ async def get_execution_results(graph_exec_id: str) -> list[ExecutionResult]:
     return res
 
 
+async def get_executions_in_timerange(
+    user_id: str, start_time: str, end_time: str
+) -> list[ExecutionResult]:
+    executions = await AgentGraphExecution.prisma().find_many(
+        where={
+            "AND": [
+                {
+                    "startedAt": {
+                        "gte": datetime.fromisoformat(start_time),
+                        "lte": datetime.fromisoformat(end_time),
+                    }
+                },
+                {"userId": user_id},
+            ]
+        },
+        include=GRAPH_EXECUTION_INCLUDE,
+    )
+    return [ExecutionResult.from_graph(execution) for execution in executions]
+
+
 LIST_SPLIT = "_$_"
 DICT_SPLIT = "_#_"
 OBJC_SPLIT = "_@_"
--- a/autogpt_platform/backend/backend/data/user.py
+++ b/autogpt_platform/backend/backend/data/user.py
@@ -1,4 +1,5 @@
 import logging
+from datetime import datetime
 from typing import Optional, cast
 
 from autogpt_libs.auth.models import DEFAULT_USER_ID
@@ -7,7 +8,12 @@ from prisma import Json
 from prisma.models import User
 
 from backend.data.db import prisma
-from backend.data.model import UserIntegrations, UserMetadata, UserMetadataRaw
+from backend.data.model import (
+    NotificationPreference,
+    UserIntegrations,
+    UserMetadata,
+    UserMetadataRaw,
+)
 from backend.util.encryption import JSONCryptor
 
 logger = logging.getLogger(__name__)
@@ -130,3 +136,38 @@ async def migrate_and_encrypt_user_integrations():
             where={"id": user.id},
             data={"metadata": Json(raw_metadata)},
         )
+
+
+async def get_active_user_ids_in_timerange(start_time: str, end_time: str) -> list[str]:
+    users = await User.prisma().find_many(
+        where={
+            "AgentGraphExecutions": {
+                "some": {
+                    "createdAt": {
+                        "gte": datetime.fromisoformat(start_time),
+                        "lte": datetime.fromisoformat(end_time),
+                    }
+                }
+            }
+        },
+    )
+    return [user.id for user in users]
+
+
+async def get_user_notification_preference(user_id: str) -> NotificationPreference:
+    user = await User.prisma().find_unique_or_raise(
+        where={"id": user_id},
+        include={
+            "UserNotificationPreference": True,
+        },
+    )
+    notification_preference = NotificationPreference(
+        user_id=user.id,
+        email=user.email,
+        # TODO with the UI when it comes in
+        preferences={},
+        daily_limit=3,
+        emails_sent_today=0,
+        last_reset_date=datetime.now(),
+    )
+    return NotificationPreference.model_validate(notification_preference)
--- a/autogpt_platform/backend/migrations/20250207203559_add_user_notification_preferences/migration.sql
+++ b/autogpt_platform/backend/migrations/20250207203559_add_user_notification_preferences/migration.sql
@@ -1,0 +1,13 @@
+-- CreateTable
+CREATE TABLE "UserNotificationPreference" (
+    "id" TEXT NOT NULL,
+    "createdAt" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    "updatedAt" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    "userId" TEXT NOT NULL,
+    "preferences" JSONB NOT NULL DEFAULT '{}',
+
+    CONSTRAINT "UserNotificationPreference_pkey" PRIMARY KEY ("id")
+);
+
+-- AddForeignKey
+ALTER TABLE "UserNotificationPreference" ADD CONSTRAINT "UserNotificationPreference_userId_fkey" FOREIGN KEY ("userId") REFERENCES "User"("id") ON DELETE CASCADE ON UPDATE CASCADE;
--- a/autogpt_platform/backend/schema.prisma
+++ b/autogpt_platform/backend/schema.prisma
@@ -33,12 +33,13 @@ model User {
   AgentPreset AgentPreset[]
   UserAgent   UserAgent[]
 
-  Profile                Profile[]
-  StoreListing           StoreListing[]
-  StoreListingReview     StoreListingReview[]
-  StoreListingSubmission StoreListingSubmission[]
-  APIKeys                APIKey[]
-  IntegrationWebhooks    IntegrationWebhook[]
+  Profile                    Profile[]
+  StoreListing               StoreListing[]
+  StoreListingReview         StoreListingReview[]
+  StoreListingSubmission     StoreListingSubmission[]
+  APIKeys                    APIKey[]
+  IntegrationWebhooks        IntegrationWebhook[]
+  UserNotificationPreference UserNotificationPreference[]
 
   @@index([id])
   @@index([email])
@@ -108,6 +109,17 @@ model AgentPreset {
   AgentExecution AgentGraphExecution[]
 
   @@index([userId])
+}
+
+model UserNotificationPreference {
+  id        String   @id @default(uuid())
+  createdAt DateTime @default(now())
+  updatedAt DateTime @default(now()) @updatedAt
+
+  userId String
+  user   User   @relation(fields: [userId], references: [id], onDelete: Cascade)
+
+  preferences Json @default("{}")
 }
 
 // For the library page
--- a/autogpt_platform/backend/backend/data/model.py
+++ b/autogpt_platform/backend/backend/data/model.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import base64
 import logging
 from datetime import datetime
+from enum import Enum
 from typing import (
     TYPE_CHECKING,
     Annotated,
@@ -22,6 +23,7 @@ from prisma.enums import CreditTransactionType
 from pydantic import (
     BaseModel,
     ConfigDict,
+    EmailStr,
     Field,
     GetCoreSchemaHandler,
     SecretStr,
@@ -389,3 +391,23 @@ class UserTransaction(BaseModel):
 class TransactionHistory(BaseModel):
     transactions: list[UserTransaction]
     next_transaction_time: datetime | None
+
+
+class NotificationPreference(BaseModel):
+    user_id: str
+    email: EmailStr
+    preferences: dict[NotificationType, bool] = {}  # Which notifications they want
+    daily_limit: int = 10  # Max emails per day
+    emails_sent_today: int = 0
+    last_reset_date: datetime = datetime.now()
+
+
+class NotificationType(str, Enum):
+    AGENT_RUN = "agent_run"
+    ZERO_BALANCE = "zero_balance"
+    LOW_BALANCE = "low_balance"
+    BLOCK_EXECUTION_FAILED = "block_execution_failed"
+    CONTINUOUS_AGENT_ERROR = "continuous_agent_error"
+    DAILY_SUMMARY = "daily_summary"
+    WEEKLY_SUMMARY = "weekly_summary"
+    MONTHLY_SUMMARY = "monthly_summary"
--- a/autogpt_platform/backend/backend/executor/database.py
+++ b/autogpt_platform/backend/backend/executor/database.py
@@ -8,6 +8,7 @@ from backend.data.execution import (
     RedisExecutionEventBus,
     create_graph_execution,
     get_execution_results,
+    get_executions_in_timerange,
     get_incomplete_executions,
     get_latest_execution,
     update_execution_status,
@@ -18,8 +19,11 @@ from backend.data.execution import (
 )
 from backend.data.graph import get_graph, get_node
 from backend.data.user import (
+    get_active_user_ids_in_timerange,
+    get_user_by_id,
     get_user_integrations,
     get_user_metadata,
+    get_user_notification_preference,
     update_user_integrations,
     update_user_metadata,
 )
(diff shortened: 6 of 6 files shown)


autogpt_platform/backend/backend/data/notifications.py
```diff
@@ -0,0 +1,262 @@
...
+    pass
+
+
+class AgentRunData(BaseNotificationData):
+    agent_name: str
+    credits_used: float
+    # remaining_balance: float
+    execution_time: float
+    graph_id: str
+    node_count: int = Field(..., description="Number of nodes executed")
```
ntindle, line 41: these are all flexible and subject to change later as we define what they actually mean :_)

autogpt_platform/backend/backend/data/user.py
```diff
@@ -130,3 +132,46 @@ async def migrate_and_encrypt_user_integrations():
...
+        },
+    )
+    notification_preference = NotificationPreference(
+        user_id=user.id,
+        email=user.email,
+        # TODO with the UI when it comes in
+        preferences={},
+        daily_limit=3,
+        emails_sent_today=0,
+        last_reset_date=datetime.now(),
```
ntindle: this will come a bit later when I have the systems all built out to test/count it

autogpt_platform/backend/schema.prisma
```diff
@@ -110,6 +112,57 @@ model AgentPreset {
...
+model NotificationEvent {
+  id        String   @id @default(uuid())
+  createdAt DateTime @default(now())
+  updatedAt DateTime @default(now()) @updatedAt
+
+  UserNotificationBatch   UserNotificationBatch? @relation(fields: [userNotificationBatchId], references: [id])
+  userNotificationBatchId String?
+
+  type NotificationType
+  data Json
```
Swiftyos, line 145: Are we expecting the data field to be dynamic with an unknown set of keys used?

autogpt_platform/backend/schema.prisma
```diff
@@ -110,6 +112,57 @@ model AgentPreset {
...
+model UserNotificationPreference {
+  id        String   @id @default(uuid())
+  createdAt DateTime @default(now())
+  updatedAt DateTime @default(now()) @updatedAt
+
+  userId String
+  user   User   @relation(fields: [userId], references: [id], onDelete: Cascade)
+
+  // Dict[NotificationType, bool]
+  preferences Json @default("{}")
```
Swiftyos: Using json here seems unnecessary. Doing something like this seems more efficient for a database, given the storage size for a bool, vs a json blob.
```
  notify_on_AGENT_RUN
  notify_on_ZERO_BALANCE
  notify_on_LOW_BALANCE
  notify_on_BLOCK_EXECUTION_FAILED
  notify_on_CONTINUOUS_AGENT_ERROR
  notify_on_DAILY_SUMMARY
  notify_on_WEEKLY_SUMMARY
  notify_on_MONTHLY_SUMMARY
  ```

**Review by Swiftyos — CHANGES_REQUESTED** [review]

File: autogpt_platform/backend/schema.prisma (hunk 4 above)
Comment (ntindle) [reply autogpt_platform/backend/schema.prisma]: Happy to make this change. Are we okay with migrations being needed for every new email type?

File: autogpt_platform/backend/schema.prisma (hunk 3 above)
Comment (ntindle) [reply autogpt_platform/backend/schema.prisma:145]: Yes, similar pattern used throughout the system

File: autogpt_platform/backend/schema.prisma (hunk 4 above)
Comment (ntindle) [reply autogpt_platform/backend/schema.prisma]: Also should we auto sign everyone up for emails or no

File: autogpt_platform/backend/migrations/20250207203559_add_user_notification_preferences/migration.sql [hunk 5]
```diff
@@ -0,0 +1,13 @@
+-- CreateTable
```
Swiftyos: Why had you decided to go with 3 separate migrations?

I'd suggest minimizing the number of migrations the system needs by collapsing these 3 down to a single migration.

autogpt_platform/backend/backend/data/execution.py
```diff
@@ -364,6 +364,26 @@ async def get_execution_results(graph_exec_id: str) -> list[ExecutionResult]:
     return res
 
 
+async def get_executions_in_timerange(
+    user_id: str, start_time: str, end_time: str
+) -> list[ExecutionResult]:
+    executions = await AgentGraphExecution.prisma().find_many(
```
Swiftyos: Can we wrap this in a try except block to catch database connection issues and log them correctly

autogpt_platform/backend/migrations/20250207203559_add_user_notification_preferences/migration.sql
ntindle: Its part of the stacking process and isn't productive to redo -- migrations are one time short costs

autogpt_platform/backend/backend/data/execution.py
Swiftyos: This comment applies to other database queries in this pr

ntindle: No, not the existing pattern, and not consistent in how the system expects queries to work based on those patterns

autogpt_platform/backend/schema.prisma
ntindle: going with yes

autogpt_platform/backend/backend/data/execution.py
Swiftyos: What is the error handling strategy then?

ntindle: let it break

autogpt_platform/backend/migrations/20250207203559_add_user_notification_preferences/migration.sql
Swiftyos: They are all in this pr, it's easy to delete them all and run prisma migrate dev to consulate them to a single migration per pr at least

autogpt_platform/backend/schema.prisma
Swiftyos: Can you elaborate for me on how the data is expected to be dynamic. What do we believe will be stored here?

autogpt_platform/backend/backend/data/execution.py
Swiftyos: I think we can do better than that. Catching the error, logging what's going on and the rasing it. 

Whilst not a complete solution at least will help the debugging process.

What would you suggest as a best practice?

autogpt_platform/backend/schema.prisma
Swiftyos: Yeah I think required a migration whenever we add a new notification type is acceptable

autogpt_platform/backend/backend/data/execution.py
ntindle: Best practice-> have a complete db modeling system that wraps all calls transparently and handles all of this for you magically with a base class so that any call to a db goes through a repository that manages all of this and we can swap DBs and ORMs lickety-split if we decide we hate prisma

Moving fast practice -> prisma will raise an error if its a problem and I'll handle it (and all other errors) by requeuing into the failed queue

Moving fast and breaking things -> I won't even enable the failed queue and let it break

autogpt_platform/backend/migrations/20250207203559_add_user_notification_preferences/migration.sql
ntindle: 🖖

autogpt_platform/backend/schema.prisma
ntindle: okay cool, will make the change and include the migration smash

autogpt_platform/backend/backend/data/execution.py
ntindle: Called with Craig, we're going to raise a custom error here so that when we get a notification late at night we are able to handle it

autogpt_platform/backend/backend/data/notifications.py
```diff
@@ -0,0 +1,333 @@
...
+
+# from backend.notifications.models import NotificationEvent
+from pydantic import BaseModel, EmailStr, Field, field_validator
+
+from .db import transaction
+
+logger = logging.getLogger(__name__)
+
+
+T_co = TypeVar("T_co", bound="BaseNotificationData", covariant=True)
```
Pwuts, line 21: - `T` is usually unbound; bound type vars can get their own acronyms for clarity
- `BaseNotificationData` == `BaseModel`, so why not use `BaseModel`?

autogpt_platform/backend/backend/data/notifications.py
```diff
@@ -0,0 +1,333 @@
...
+
+
+class BatchingStrategy(str, Enum):
```
Pwuts: I'm going to sort of contradict myself from earlier here, but: I think this should be:
```suggestion
class BatchingStrategy(Enum):
```
Otherwise instances of passing this enum into string parameters are not flagged, and you get stuff like `<BatchingStrategy.IMMEDIATE: "immediate">` inserted in your string, where you really just want `immediate`.

autogpt_platform/backend/backend/data/notifications.py
```diff
@@ -0,0 +1,333 @@
...
+        return v
+
+    @property
+    def template(self) -> str:
+        return NotificationTypeOverride(self.type).template
+
+
+def get_data_type(
+    notification_type: NotificationType,
+) -> Type[BaseNotificationData]:
```
Pwuts: nit:
```suggestion
) -> type[BaseNotificationData]:
```

autogpt_platform/backend/backend/data/notifications.py
```diff
@@ -0,0 +1,333 @@
...
+            NotificationType.DAILY_SUMMARY: "daily_summary.html",
+            NotificationType.WEEKLY_SUMMARY: "weekly_summary.html",
+            NotificationType.MONTHLY_SUMMARY: "monthly_summary.html",
+        }[self.notification_type]
+
+
+class NotificationPreference(BaseModel):
+    user_id: str
+    email: EmailStr
+    preferences: dict[NotificationType, bool] = {}  # Which notifications they want
```
Pwuts:
```suggestion
    preferences: dict[NotificationType, bool] = Field(default_factory=dict, description="Which notifications the user wants")
```

autogpt_platform/backend/backend/data/notifications.py
```diff
@@ -0,0 +1,333 @@
...
+    daily_limit: int = 10  # Max emails per day
+    emails_sent_today: int = 0
+    last_reset_date: datetime = datetime.now()
```
Pwuts:
```suggestion
    last_reset_date: datetime = Field(default_factory=datetime.now)
```

Pwuts: same for the other instances

autogpt_platform/backend/backend/data/notifications.py
```diff
@@ -0,0 +1,333 @@
...
+        NotificationType.LOW_BALANCE: timedelta(minutes=60),
+        NotificationType.BLOCK_EXECUTION_FAILED: timedelta(minutes=60),
+        NotificationType.CONTINUOUS_AGENT_ERROR: timedelta(minutes=60),
+    }[notification_type]
+
+
+async def create_or_add_to_user_notification_batch(
+    user_id: str,
+    notification_type: NotificationType,
+    data: str,  # type: 'NotificationEventModel'
```
Pwuts, line 231: nit:
```suggestion
    data_json: str,  # stringified NotificationEventModel
```

autogpt_platform/backend/backend/data/notifications.py
```diff
@@ -0,0 +1,333 @@
...
+    logger.info(
+        f"Creating or adding to notification batch for {user_id} with type {notification_type} and data {data}"
+    )
+
+    notification_data = NotificationEventModel[
+        get_data_type(notification_type)
+    ].model_validate_json(data)
+
+    # Serialize the data
+    # serialized_data = json.dumps(notification_data.data.model_dump())
```
Pwuts: delete?

autogpt_platform/backend/backend/data/notifications.py
```diff
@@ -0,0 +1,333 @@
...
+        async with transaction() as tx:
+            notification_event = await tx.notificationevent.create(
+                data={
+                    "type": notification_type,
+                    "data": json_data,
+                }
+            )
+
+            # Create new batch
+            resp = await tx.usernotificationbatch.create(
```
Pwuts: nit: you can do:
```suggestion
        async with transaction() as tx:
            notification_event = await NotificationEvent.prisma(tx).create(
                data={
                    "type": notification_type,
                    "data": json_data,
                }
            )

            # Create new batch
            resp = await NotificationBatch.prisma(tx).create(
```
which looks more like other DB code

autogpt_platform/backend/backend/data/user.py
```diff
@@ -130,3 +132,46 @@ async def migrate_and_encrypt_user_integrations():
             where={"id": user.id},
             data={"metadata": Json(raw_metadata)},
         )
+
+
+async def get_active_user_ids_in_timerange(start_time: str, end_time: str) -> list[str]:
```
Pwuts, line 143: I suggest passing in the `datetime` instead of the string repr

autogpt_platform/backend/migrations/20250207203559_add_user_notification_preferences/migration.sql
```diff
@@ -0,0 +1,13 @@
...
+    "createdAt" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    "updatedAt" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    "userId" TEXT NOT NULL,
+    "preferences" JSONB NOT NULL DEFAULT '{}',
+
+    CONSTRAINT "UserNotificationPreference_pkey" PRIMARY KEY ("id")
+);
+
+-- AddForeignKey
+ALTER TABLE "UserNotificationPreference" ADD CONSTRAINT "UserNotificationPreference_userId_fkey" FOREIGN KEY ("userId") REFERENCES "User"("id") ON DELETE CASCADE ON UPDATE CASCADE;
```
Pwuts: nit: missing newline

autogpt_platform/backend/schema.prisma
Pwuts: > ```
>   notify_on_AGENT_RUN
>   notify_on_ZERO_BALANCE
>   notify_on_LOW_BALANCE
>   notify_on_BLOCK_EXECUTION_FAILED
>   notify_on_CONTINUOUS_AGENT_ERROR
>   notify_on_DAILY_SUMMARY
>   notify_on_WEEKLY_SUMMARY
>   notify_on_MONTHLY_SUMMARY
> ```

just a note: the casing on this snippet is incorrect for column names

autogpt_platform/backend/backend/data/notifications.py
ntindle: if you hadn't changed the variable name i'd have added this to batch -- for feedback on writing suggestions in the future

autogpt_platform/backend/backend/data/notifications.py
ntindle: This pattern was copied from other places transactions are used throughout the system lol

autogpt_platform/backend/backend/data/user.py
ntindle: does not seralize correctly over pyro

autogpt_platform/backend/backend/data/notifications.py
ntindle:
```suggestion
 ```

autogpt_platform/backend/backend/data/notifications.py
ntindle: I want only things that inherit from `BaseNotificationData`

I can fix this in a follow up since I'll probs need a contravariant too

autogpt_platform/backend/schema.prisma
Swiftyos: I'd really like to understand this and why it's happening in the code base. 

Can we open a discussion about it on discord?

Swiftyos (approved):

autogpt_platform/backend/backend/data/user.py
Pwuts: that's fixable, like we did for Pydantic models

Pwuts: - #9472

Not tested, take or leave at your discretion

itsababseh (approved):
If this breaks anything I John am not liable cc @ntindle :)

PR #9452, opened 2025-02-10
feat(backend): Migrate old models in existing agents

kcze:
Some existing nodes use models that no longer exist as values on `LlmModel` enum.

### Changes 🏗️

- Update models for all blocks with `LlmModel` fields that do not exist in `LlmModel` enum to `gpt-4o`, directly in `AgentNode->constantInput` db column, on server startup

#### For code changes:
- [x] I have made a test plan
  - [x] Updates wrong models to `gpt-4o` for all affected `AgentNode`s
  - [x] Doesn't update correct models
  - [x] Doesn't insert model when unnecessary
  - [x] Doesn't break other values in jsonb


autogpt_platform/backend/backend/data/graph.py
```diff
@@ -71,6 +72,9 @@ class NodeModel(Node):
     def from_db(node: AgentNode):
         if not node.AgentBlock:
             raise ValueError(f"Invalid node {node.id}, invalid AgentBlock.")
+
+        NodeModel.migrate_llm_models(node)
```
ntindle: What if we do that here autogpt_platform/backend/backend/server/rest_api.py
```python
@contextlib.asynccontextmanager
async def lifespan_context(app: fastapi.FastAPI):
    await backend.data.db.connect()
    await backend.data.block.initialize_blocks()
    await backend.data.user.migrate_and_encrypt_user_integrations()
    await backend.data.graph.fix_llm_provider_credentials()
    with launch_darkly_context():
        yield
    await backend.data.db.disconnect()
```

Swiftyos (dismissed):
In general happy with this approach. However, I don't feel we should be migrating ai models with clearly communicating it to the user. 

Can we add a ticket to push a notification to the user when a model has to be migrated and we can address this at an appropriate time

autogpt_platform/backend/backend/data/graph.py
kcze: Thanks, I did this

AutoGPT-Agent:
While the PR has a descriptive title with proper scope and the changes seem focused on migrating LLM models (staying in scope), the PR template is inadequately filled out. The checklist items are unchecked and no actual test plan is provided for what appears to be a significant database migration operation. For database migrations especially, having a clear test plan is crucial.

autogpt_platform/backend/backend/data/graph.py
```diff
@@ -932,3 +934,40 @@ async def fix_llm_provider_credentials():
...
+        for field_name, field in fields.items():
+            if field.annotation == LlmModel:
+                llm_model_fields[block.id] = field_name
+
+    # Update each block
+    for id, path in llm_model_fields.items():
+        # Convert enum values to a list of strings for the SQL query
+        enum_values = [v.value for v in LlmModel.__members__.values()]
+
+        await db.execute_raw(
```
kcze: I had to do a raw query because Prisma doesn't handle:
1. Checking if a key exists in jsonb column (and throws when checks value for non-existing key)
2. Updating individual value for a key in jsonb

-> 1. would force me to get all AgentNodes (*a lot* of nodes if we remove popular model in future) and 2. to update entire json for each AgentNode individually.
I think this query is safe, none of the variables come from the user.

kcze:
@Swiftyos  Added ticket for notifications https://github.com/Significant-Gravitas/AutoGPT/issues/9614

Swiftyos (approved):

majdyz (approved):
Won't this get slower on every run? we will keep increasing the number of nodes we are checking

Though we've been doing it on other startup query so I guess it's fine 🤷

PR #9469, opened 2025-02-12
feat(backend): add ability to send emails to notification service

ntindle:
We need a way to send emails for the email service to function. We will depend on Postmark to do that.

This PR adds a simple email-sending service with the required settings to make it work. It also builds on the previous agent run by sending the emails that are in the immediate queue. Keep in mind that the email template leaves a bit to be desired. 

### Changes 🏗️

- Add `email.py` with the minimum required to send an email (plus type handling)
- Add settings configs for the token and the send address to the `settings.py` and `.env.example`
- Add a db call to get user email by ID since the `metadata` field of `prisma.models.User` isn't serializable over our message bus tool `Pyro` that the `DatabaseManager` uses
- Add a horrible `AgentRun` email template using `jinja2`
- Add `postmarker` to `pyproject.toml`

#### For code changes:
- [x] I have made a test plan
  
  - [x] Build and run an agent and make sure it emails me (must be signed into same domain as receiving address for now)

    - [x] I have added a check that disables email if config is not set correctly


autogpt_platform/backend/backend/notifications/email.py
```diff
@@ -0,0 +1,69 @@
...
+        self,
+        notification: NotificationType,
+        user_email: str,
+        data: NotificationEventModel[T_co] | list[NotificationEventModel[T_co]],
+    ):
+        body = self._get_template(notification)
+        # use the jinja2 library to render the template
+        body = self.formatter.format_string(body, data)
+        logger.info(
+            f"Sending email to {user_email} with subject {"subject"} and body {body}"
```
ntindle: yeah we need to pull the subject from **somewhere** but not in this pr

autogpt_platform/backend/backend/notifications/email.py
```diff
@@ -0,0 +1,69 @@
...
+        )
+        self._send_email(user_email, "subject", body)
+
+    def _get_template(self, notification: NotificationType):
+        # convert the notification type to a notification type override
+        notification_type_override = NotificationTypeOverride(notification)
```
ntindle, line 75: this override is a superset of the Notification Type that has useful functions built in -- like looking up what template to load

Pwuts (dismissed):

itsababseh (approved):
Passed CI and Pwuts reviewed this one, so approving.

Pwuts (approved):

PR #9478, opened 2025-02-13
[Edited] Refactor code to optimize performance

MayureshMore:
Changes made in branch: **MayureshMore:Fix-for-id-constraint-violation**
[Edited] Refactor code to optimize performance


ntindle:
Thanks for the contribution, closing until clarity about why this is better

JulianStiebler:
His PR's are all a little weird or have some comment about "Automated". 


 The literal only changes is again a comment about automated changes and hes not doing find_unique but find_first because he now is also `block.name` which should make it even slower. 
`find_unique` directly queries a unique field which is indexed and optimized for fast lookups. 

Since a name will always have a assigned id theres no point in his changes at all.

Issue #9252, opened 2025-01-13
Editing notes causes cursor to always reset to bottom


Bentlybro:
when you try to edit text at the top of a note, the cursor moves to the bottom after any change.

(Video to better show it)

[2025-01-13 11-43-39.mp4](https://uploads.linear.app/a47946b5-12cd-4b3d-8822-df04c855879f/352c6c9b-889d-4766-99b5-b59860e7d761/d7700743-9e19-4d37-a549-9cf1156a2b43?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiL2E0Nzk0NmI1LTEyY2QtNGIzZC04ODIyLWRmMDRjODU1ODc5Zi8zNTJjNmM5Yi04ODlkLTQ3NjYtOTliNS1iNTk4NjBlN2Q3NjEvZDc3MDA3NDMtOWUxOS00ZDM3LWE1NDktOWNmMTE1NmEyYjQzIiwiaWF0IjoxNzM2NzY5OTg5LCJleHAiOjMzMzA3MzI5OTg5fQ._COT8ZKB4-Ra2dglEaz7wezYtK_neFLOO02NW-dg5MI)

mdimado:
I can try. please assign

jlopezmarti20:
Hey guys I'm super new to open source contributions and I would like to see if I can be assigned and contribute to this issue if still open. This is a great opportunity for me to get involve.

ntindle:
@jlopezmarti20 assigned! Good luck 🍀

jlopezmarti20:
Can you help me identify where the issue might be? Also, as a new contributor, any guidance or resources would be greatly appreciated.

Bentlybro:
[...]

PR #9486, opened 2025-02-16
Fix the cursor jumping issue in sticky notes

jlopezmarti20:
Fix the cursor jumping issue in sticky notes issue #9252


ntindle:
Can't merge without CLA :( feel free to reopen when its signed

PR #9487, opened 2025-02-16
Fix the cursor jumping issue in sticky notes

jlopezmarti20:
Fix the cursor jumping issue in sticky notes


ntindle:
Cant merge without cla sorry :( feel free to reopened if signed

PR #9490, opened 2025-02-17
Swiftyos/open 2373 tool discovery input mapping for smart decision maker block

Swiftyos:
## Task

Implement a mechanism for the Smart Decision Maker (SDM) block to dynamically retrieve all available tools (blocks and agents) along with their required inputs.
This allows the SDM block to make informed decisions on which tool to call and how to structure the input data for each call.

## Changes 🏗️
	•	WIP: Add the Smart Decision Maker (SDM) block.
	•	Break circular imports in integration code.

### Current Status

I’m currently focused on the smart decision logic. The main additions in the ongoing PR include:
	•	Defining function signatures for OpenAI function-calling schemas based on node links and the linked blocks.
	•	Adding tests for function signature generation.

### What’s Next

After discussing with Toran, we plan to:
	1.	Force all tool calls to be made via an agent.
	2.	Restrict each tool call entry to a single node.

Currently, we store outputs in a dictionary with links in the format {tool_name: parameter}. However, if multiple parameters need to be passed to the same tool—which we’d like to do in order to specify the parameters the SDM block should generate—we need a better approach. Specifically, we must automatically generate links in the format tool_name_parameter_name under the hood.

I will address this linking issue next.

### WIP Reviewers

Please take a look at the general approach that I'm taking and let me know if there are any specifics you think should be considered.

If you have any knowledge around my next steps please also comment giving my the context you have.

Out of scope:  code style and organization at this stage


autogpt_platform/backend/backend/blocks/smart_decision_maker.py
```diff
@@ -0,0 +1,184 @@
...
+                - "description": A description of the parameter.
+
+        Raises:
+        - ValueError: If no tool links are found for the given node ID or if no tool sink nodes are identified.
+        """
+        # Filter the graph links to find those that are tools and are linked to the specified node_id
+        tool_links = [
+            link
+            for link in graph.links
+            if link.source_name == "tool" and link.source_id == node_id
```
majdyz: instead of pulling the whole graph you can use `get_node` function and find the output_links of the "tool" source.

Swiftyos: Would that approach not result in requiring many more database calls as I need to get the node details for all the sink blocks linked via the tool output

majdyz: I see, if you need some detail I think you need to make the function for it, sth like:

```
async def get_connected_output_nodes(source_node_id: str) -> list[Node]:
    links = await AgentNodeLink.prisma().find_many(
        where={"agentNodeSourceId": source_node_id},
        include={"AgentNode": {"include": AGENT_NODE_INCLUDE}},
    )
    return [NodeModel.from_db(link.AgentNodeSink) for link in links]
```

You can still do one query, and avoid pulling the whole graph in case it's a large graph.

majdyz: Anyway the idea looks good though :+1:

autogpt_platform/backend/backend/blocks/smart_decision_maker.py
```diff
@@ -0,0 +1,197 @@
...
+    ) -> BlockOutput:
+        db_client = get_database_manager_client()
+
+        # Retrieve the current graph and node details
+        graph = db_client.get_graph(graph_id)
+
+        if not graph:
+            raise ValueError(f"Graph not found {graph_id}")
+
+        tool_functions = self._create_function_signature(node_id, graph)
```
ntindle: Test this works with mem0 https://github.com/Significant-Gravitas/AutoGPT/blob/016ec0ff6b38341873735a11d2585a8ad8c623e8/autogpt_platform/backend/backend/blocks/mem0.py#L121 since it uses the kwargs based on its place in the graph

Swiftyos: @ntindle It seems that high level decision is that tools must be an AgentExecutionBlock. 

If you feel strongly that this is a bad idea, let me know and we can discuss it with Toran

ntindle: Why should it require that? I can see plenty of use for this not needing a whole agent and the overhead that would come with it.

Swiftyos: The main reason why is to make a smart decision, a certain amount of information is required about a tool. We can backprogate this information from an agent - name, description, required params, etc.

But with a normal block, we can't really do this unless we assume that block can't be connected to another block...

There are a fair bit of subtleties with this system

Swiftyos:
@Pwuts tests are failing due to an issue with the credentials system that does not occur locally. Can you advise on how to fix this please

autogpt_platform/backend/backend/blocks/llm.py
```diff
@@ -238,6 +242,190 @@ class Message(BlockSchema):
...
+    completion_tokens: int
+
+
+def llm_call(
+    credentials: APIKeyCredentials,
+    llm_model: LlmModel,
+    prompt: list[dict],
+    json_format: bool,
+    max_tokens: int | None,
+    tools: list[dict] | None = None,
```
majdyz, line 300: What is `tools` here how do I use it? can we add it to the docstring too?

autogpt_platform/backend/backend/blocks/llm.py
```diff
@@ -238,6 +242,190 @@ class Message(BlockSchema):
...
 
+class LLMResponse(BaseModel):
+    prompt: str
+    response: str
+    tool_calls: Optional[List[ChatCompletionMessageToolCall]] | None
+    prompt_tokens: int
+    completion_tokens: int
+
+
+def llm_call(
```
majdyz, line 294: This is copied function, can we remove the old one and make the function caller use this?

autogpt_platform/backend/backend/blocks/smart_decision_maker.py
```diff
@@ -0,0 +1,307 @@
...
+        Returns:
+            List[Any]: A list of metadata for the tool graphs linked to the specified node.
+        """
+        db_client = get_database_manager_client()
+        graph_meta = []
+
+        tool_links = {
+            link.sink_id
+            for link in graph.links
+            if link.source_name.startswith("tools_") and link.source_id == node_id
```
majdyz: Can we make `tools_` constant, and this is so prone to conflict, can we use a special character?

autogpt_platform/backend/backend/data/graph.py
```diff
@@ -312,12 +312,49 @@ def reassign_ids(self, user_id: str, reassign_graph_id: bool = False):
...
+            return sanitized_name
+
+        # Validate smart decision maker nodes
+        smart_decision_maker_nodes = set()
+        smd_unique_tool_links = {}
+        agent_nodes = set()
+
+        for node in self.nodes:
+            # Smart decision maker nodes
+            if node.block_id == "3b191d9f-356f-482d-8238-ba04b6d18381":
```
majdyz: We intentionally match stuff based on the block type like below to avoid this hardcoded ID check to happen:
https://github.com/Significant-Gravitas/AutoGPT/pull/9490/files#diff-fe699074de9055342cbf5c2aef8eba0ae08ef41429c2f012789c20eb9dec532cR381

We can get the block object early here.

autogpt_platform/backend/backend/data/graph.py
```diff
@@ -312,12 +312,49 @@ def reassign_ids(self, user_id: str, reassign_graph_id: bool = False):
...
+            elif node.block_id == "e189baac-8c20-45a1-94a7-55177ea42565":
+                agent_nodes.add(node.id)
 
         input_links = defaultdict(list)
+        tool_name_to_node = {}
+
         for link in self.links:
             input_links[link.sink_id].append(link)
 
+            if (
```
majdyz, line 346: Add comments here?

majdyz (dismissed):
Are these true:
* SMD can't be connected to non-AgentExecutorBlock, if so, why not? Would it calling non-agent block would be simpler, e.g: We have a block definition, type, description, etc, plus it's a more common pattern.

* the `tools` output pin will connected to N other input pins, and then on each yield, N input will be produced, but N-1 of it will raise an exception due to the validation failure since it doesn't match the tool that the SMD is using. Is this the flow we are expecting ?

autogpt_platform/backend/backend/blocks/llm.py
Swiftyos: I really want too, but I know how much people hate me making bigger PRs. So I was planning on refactoring llm.py in a separate PR if thats okay.

autogpt_platform/backend/backend/blocks/smart_decision_maker.py
Swiftyos: Sure good idea

autogpt_platform/backend/backend/data/graph.py
Swiftyos: When I did that I was hitting circular import issues. I've hit this problem several times during adding this block...

majdyz: I tweaked this a bit, feel free to check and update

autogpt_platform/backend/backend/blocks/llm.py
majdyz: Okay sounds good, though if the change is just removing `self.` in `self.llm_call` and removing the llm_call function I definitely don't mind doing it in a single PR. Both are fine for me :+1:

autogpt_platform/backend/backend/blocks/smart_decision_maker.py
Swiftyos: Yes, done - I went for  _^_

Swiftyos:
> * SMD can't be connected to non-AgentExecutorBlock, if so, why not? Would it calling non-agent block would be simpler, e.g: We have a block definition, type, description, etc, plus it's a more common pattern.
> * the `tools` output pin will connected to N other input pins, and then on each yield, N input will be produced, but N-1 of it will raise an exception due to the validation failure since it doesn't match the tool that the SMD is using. Is this the flow we are expecting ?

We have discussed this on discord. Here are the answers for anyone reading GitHub.

1. The SDM can only run AgentExecutorBlock, this decision was made as there a situations where calling a block just doesnt transmit enough information for smart decisions to be made. We are going to add block execution in another PR and accept this limitation
2. The SDM block yields specific outputs required for a tool so whilst the image suggests this will be an issue, its actually is not.

autogpt_platform/backend/backend/data/integrations.py
```diff
@@ -8,16 +8,25 @@
...
+app_config = Config()
+
 logger = logging.getLogger(__name__)
 
 
+def webhook_ingress_url(provider_name: ProviderName, webhook_id: str) -> str:
+    return (
+        f"{app_config.platform_base_url}/api/integrations/{provider_name.value}"
+        f"/webhooks/{webhook_id}/ingress"
+    )
```
Pwuts: code duplication?

Swiftyos: Yes correct, as the description said, its to avoid a circular import issue with the integration system.

I've tired out reverting this change and it looks to be working again, something must have been fixed in dev 🤷‍♂️

autogpt_platform/backend/backend/blocks/llm.py
```diff
@@ -238,6 +242,190 @@ class Message(BlockSchema):
...
+        return LLMResponse(
+            prompt=json.dumps(prompt),
+            response=response.choices[0].message.content or "",
+            tool_calls=response.choices[0].message.tool_calls,
+            prompt_tokens=response.usage.prompt_tokens if response.usage else 0,
+            completion_tokens=response.usage.completion_tokens if response.usage else 0,
+        )
+    elif provider == "anthropic":
+        if tools:
+            raise ValueError("Anthropic does not support tools.")
```
Pwuts: https://docs.anthropic.com/en/docs/build-with-claude/tool-use

autogpt_platform/backend/backend/blocks/llm.py
```diff
@@ -238,6 +242,190 @@ class Message(BlockSchema):
     content: str
 
 
+class LLMResponse(BaseModel):
+    prompt: str
+    response: str
+    tool_calls: Optional[List[ChatCompletionMessageToolCall]] | None
```
Pwuts: imo don't use provider-specific library types to define an internal type like this

autogpt_platform/backend/backend/blocks/llm.py
```diff
@@ -238,6 +242,190 @@ class Message(BlockSchema):
...
+            ]
+        elif json_format:
+            response_format = {"type": "json_object"}
+
+        response = oai_client.chat.completions.create(
+            model=llm_model.value,
+            messages=prompt,  # type: ignore
+            response_format=response_format,  # type: ignore
+            max_completion_tokens=max_tokens,
+            tools=tools_param,  # type: ignore
```
Pwuts, line 346: slapping `# type: ignore` on everything is probably a bad idea

autogpt_platform/backend/test/executor/test_tool_use.py
```diff
@@ -0,0 +1,246 @@
...
+        ),
+    ]
+
+    test_graph = graph.Graph(
+        name="TestGraph",
+        description="Test graph",
+        nodes=nodes,
+        links=links,
+    )
+    with pytest.raises(Exception):
```
Pwuts: non-specific, will give false-negative if something else fails

autogpt_platform/backend/backend/blocks/llm.py
Swiftyos: message and response_format already had a type ignore

autogpt_platform/backend/backend/data/graph.py
Swiftyos: This bit of code was deleted

autogpt_platform/backend/backend/blocks/llm.py
Swiftyos: Added

autogpt_platform/backend/backend/blocks/llm.py
Swiftyos: I've had to keep a wrapped class function to allow the block testing framework to mock the call. However, I have made this change

Swiftyos:
Screenshot 2025-02-25 at 11 32 28
Screenshot 2025-02-25 at 11 33 33
Tested with both openai and anthropic

autogpt_platform/backend/backend/blocks/smart_decision_maker.py
```diff
@@ -0,0 +1,307 @@
...
+            output_schema=SmartDecisionMakerBlock.Output,
+            test_input={
+                "prompt": "Hello, World!",
+                "credentials": llm.TEST_CREDENTIALS_INPUT,
+            },
+            test_output=[],
+            test_credentials=llm.TEST_CREDENTIALS,
+        )
+
+    # If I import Graph here, it will break with a circular import.
```
ntindle, line 98: ? can you clarify this

if for typing, you can add the 
```
from typing import TYPE_CHECKING
if TYPE_CHECKING:
    from fastapi import FastAPI # whatever you want to import to break the cricle
```
```

    def _get_tool_graph_metadata(self, node_id: str, graph: `Graph`) -> List[Any]:
```

autogpt_platform/backend/backend/data/block.py
```diff
@@ -351,6 +352,14 @@ def run(self, input_data: BlockSchemaInputType, **kwargs) -> BlockOutput:
         Run the block with the given input data.
         Args:
             input_data: The input data with the structure of input_schema.
+
+        Kwargs: Currently 14/02/2025 these include
```
ntindle, line 356: aside: this should be updated in the new_blocks docs every time something is added to this list

ntindle (approved):

autogpt_platform/backend/backend/blocks/smart_decision_maker.py
Swiftyos: Okay thanks Nick, I'll merge this now and make a new PR that updates the type checking

PR #9497, opened 2025-02-18
fix(backend): correctly check if email service is set up

ntindle:
I made a mistake in how we check if postmark exists

### Changes 🏗️
- adds a more explicit setting of postmark to none and extra checking to prevent its use if it isn’t set

#### For code changes:
- [x] I have made a test plan
  
  - [x] I have no plan, it’s a simple logic bug so if it passes CI it’s good


Pwuts (approved):
looks good to me

PR #9501, opened 2025-02-19
feat(backend): move failed messages into a dead letter queue

ntindle:
When we fail to process something, we don't want to keep retrying forever. We should store those and process them later

### Changes 🏗️

- Fix the type of the failed exchange from Direct to Topic to allow filtering based on name (allows us later to do more advanced handling of queue types)
- abstract processing the messages in a queue a bit to reduce repeated code
- abstract how we check if a user wants a notification so that its a bit easier to process
- Handle errors better
- Abstract model parsing

#### For code changes:
- [ ] I have made a test plan
  
  - [ ] ...

<details>
  
</details>


Bentlybro (approved):
Tested and works :)

PR #9508, opened 2025-02-23
feat(backend): Pyro to FastAPI migration for micro service

majdyz:
Due to legacy reasons, we've been using Pyro for our inter-process communication channel. While it fulfilled our initial needs, there were a few limitations that have been encountered:
* Each connection will reserve 1 thread, when the thread is running out there will be no connection being accepted by the service.
* Lack of asynchronous execution mode, we are locked in the sync execution which ended up wasting the I/O bound workload. Moving away from this will unlock async execution support for agent blocks.
* Low throughput, while the database is still the main bottleneck, we've started seeing instances where the service is being denied due to the high traffic of the Pyro service.

### Changes 🏗️

Replace the usage of Pyro with the FastAPI Rest HTTP server and make the code work.

Introduced the new config:
`use_http_based_rpc`: Whether to use HTTP-based RPC for communication between services.

If it's enabled FastAPI will be used, if it's disabled existing Pyro will be used.

#### For code changes:
- [x] I have made a test plan
  
  - [x] Create from scratch and execute an agent with at least 3 blocks with cost (AI blocks).
  - [x] Import an agent from file upload, and confirm it executes correctly

<details>
  
</details>

<details>

  - Changing ports
</details>


autogpt_platform/backend/backend/executor/database.py
```diff
@@ -50,17 +54,11 @@ def send_execution_update(self, execution_result: ExecutionResult):
...
-
-        # Register serializers for annotations on bare function
-        register_pydantic_serializers(f)
-
-        return wrapper
+        # TODO:
+        #  This function lies about its return type to make the DynamicClient
+        #  call the function synchronously, fix this when DynamicClient can choose
+        #  to call a function synchronously or asynchronously.
+        return expose(cast(Callable[Concatenate[object, P], R], f))
```
ntindle: You can use the inspect module from Python to check this

ntindle:
I'm not sure I love the idea of making a endpoint for all these -- this seems like a massive security challenge

majdyz:
@ntindle this has been a network accessible API endpoints all along. We are just now using https friendly json format

ntindle:
My b I thought this was adding them to our current one -- why this over rabbitmq?

majdyz:
This is orthogonal to message queue, we will move away from Redis to RabbitMQ. This is for the synchronous RPC. We want to eventually unlock async execution for our block executor hence we can't stick with Pyro.

I was considering gRPC but it requires a lot more changes, like defining proto and its signature. Somehow I managed to auto generate endpoint routes using fast API to behave as close as the current Pyro implementation. I'll improve this overtime:
* Support async client
* Support async run method on block
* Support parallel execution on blocks, etc

autogpt_platform/backend/backend/executor/database.py
majdyz: The function lying on its signature is unfortunately intentional.

To make the fast API call behaves like a function call, we need to make the dynamic client (for now) thinks it's calling a sync function, where in fact the server can now write the endpoints in async mode.

autogpt_platform/backend/backend/util/service.py
```diff
@@ -151,27 +70,85 @@
...
+        if asyncio.iscoroutinefunction(f):
+
+            async def async_endpoint(body: RequestBodyModel):  # type: ignore
+                try:
+                    return await f(
+                        **{name: getattr(body, name) for name in body.model_fields}
+                    )
+                except Exception as e:
+                    logger.exception(f"Error in {func.__name__}: {e}")
+                    return responses.JSONResponse(status_code=500, content=e)
```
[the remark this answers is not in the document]
majdyz: I can't accept this, this is an internal service and we intentionally want to propagate the exception message.

autogpt_platform/backend/backend/util/service.py
```diff
@@ -151,27 +70,85 @@
...
+        else:
+
+            def sync_endpoint(body: RequestBodyModel):  # type: ignore
+                try:
+                    return f(
+                        **{name: getattr(body, name) for name in body.model_fields}
+                    )
+                except Exception as e:
+                    logger.exception(f"Error in {func.__name__}: {e}")
+                    return responses.JSONResponse(status_code=500, content=e)
```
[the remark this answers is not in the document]
majdyz: I can't accept this, this is an internal service and we intentionally want to propagate the exception message.

ntindle:
https://www.rabbitmq.com/tutorials/tutorial-six-python

https://docs.aio-pika.com/rabbitmq-tutorial/6-rpc.html

majdyz:
@ntindle so your recommendation is to replace all of our synchronous RPCs with async ones through message queues?

Imagine this existing scenario:

a service `X` needs to do a process that requires:
* get an info from service `Y`
* get an info from service `Z`
* process the info on the existing service
* Publish the result to service `Z`

On the sync system, this will be done like this:
```
info_1 = serviceY.get_info()
info_2 = serviceZ.get_info()
processed_info = do_local_stuff(info_1, info_2)
serviceZ.publish_info(processed_info)
```

On a purely asynchronous RPC system, how do you imagine this going to look like? I don't see this to be possible unless we somehow overhaul the system to a purely event-driven system. While this is interesting, I still can't figure out whether this is feasible at all at the moment.

Alternatively, assuming we managed to create a wrapper using pika that allows us to make the async call behave synchronously through the message queue, how is it going to be? 

Let's say we are doing the `info_1 = serviceY.get_info()` part.  You will need two queues, the server one and the callback one. The server part will just consume the request and publish the result with the correlation ID. What should the client do after publishing the request message? should it wait and spawn another task that listens to the callback queue, filters out the matching correlation ID, and notifies the waiting caller to continue the process? If so, how does the calling task & subscriber task communicate?

ntindle:
To be clear, not my recommendation , but asking why not. This isn't something I'm super familiar with so trying to learn

majdyz:
Acknowledged. Yes, the reply is a discussion.

The approach above will push the responsibility of the event listening to be the same event loop & thread as the caller, which implies:
* Each caller will create its own callback queue
* Each caller will hold and manage two continuous connections to RabbitMQ (callback & server).
* Each caller that calls X different services, on its event loop will subscribe to 2*X different queues while holding 2*X connections.

One reason we are moving away from Pyro to a stateless REST API as I stated in the description is the limited connection we can provision for a service. Considering we will do this on each of our graph/node executor instances, those executors/processes will hold their own 2*X connections & queues for their service. Is this within the limitation of the MQ?

majdyz:
Also, this is code implementation specific, but one challenge I see is this can only be achieved when the code is fully async.  The existing sync code needs to be converted into async code directly otherwise it will require threads since we are listening to multiple events while doing the client stuff at the same time.

ntindle:
https://www.rabbitmq.com/docs/connections

This makes it sound like you're hitting OS exhaustion limits before rabbitmq limit

ntindle:
Okay yeah I see what you're saying. Here's the sync docs btw

```python

#!/usr/bin/env python
import pika
import uuid


class FibonacciRpcClient(object):

    def __init__(self):
        self.connection = pika.BlockingConnection(
            pika.ConnectionParameters(host='localhost'))

        self.channel = self.connection.channel()

        result = self.channel.queue_declare(queue='', exclusive=True)
        self.callback_queue = result.method.queue

        self.channel.basic_consume(
            queue=self.callback_queue,
            on_message_callback=self.on_response,
            auto_ack=True)

        self.response = None
        self.corr_id = None

    def on_response(self, ch, method, props, body):
        if self.corr_id == props.correlation_id:
            self.response = body

    def call(self, n):
        self.response = None
        self.corr_id = str(uuid.uuid4())
        self.channel.basic_publish(
            exchange='',
            routing_key='rpc_queue',
            properties=pika.BasicProperties(
                reply_to=self.callback_queue,
                correlation_id=self.corr_id,
            ),
            body=str(n))
        while self.response is None:
            self.connection.process_data_events(time_limit=None)
        return int(self.response)


fibonacci_rpc = FibonacciRpcClient()

print(" [x] Requesting fib(30)")
response = fibonacci_rpc.call(30)
print(f" [.] Got {response}")

```

majdyz:
> Hitting OS exhaustion
Which OS and what part is exhausted?
We are talking about the centralized message queue vs peer-to-peer HTTP calls between services.

ntindle:
Number of open connections

>  Operating systems have a [limit around how many TCP connections (sockets) a single process can have open simultaneously](https://www.rabbitmq.com/docs/networking#open-file-handle-limit). The limit is often sufficient for development and some QA environments. [Production](https://www.rabbitmq.com/docs/production-checklist) environments must be configured to use a higher limit in order to support a larger number of concurrent client connections.

majdyz:
Right as you can see from the code above, you either need to poll the system to wait for the matching correlation ID messages to match or spawn another thread to do the consuming. Now imagine you need to call two services as the scenario above. How are you going to handle it without blocking each other on synchronous code.

ntindle:
Why would you not use an async client in that case and await both at the same time so they can both process?

majdyz:
That's the whole point, our code needs to be fully async, you can't easily introduce this and expect the code will just work.

ntindle:
And our code isn't fully async so this would be a bad thing to try?

majdyz:
Definitely not a bad thing, it's just not yet possible without making our code do an async-first system. Hopefully by doing this incrementally we can move toward a fully async system.

It's an interesting option to consider indeed, I still need to consider a few logistics though:
* Does a sync RPC using MQ causes more latency, and even if it is, is it a deal breaker difference.
* Is it still trivial to handle timeout & retry using this pattern.
* Can a single point of failure on a message queue cause a scalability risk or is it easy (or even needed) to shard the MQ.
* Does this pattern give more benefit rather than dreict API calls.

One benefit I see is auditing since we are logging all interaction placed in a single location, [[event sourcing](https://microservices.io/patterns/data/event-sourcing.html)](https://microservices.io/patterns/data/event-sourcing.html) pattern used to be cool.

autogpt_platform/backend/backend/util/service.py
```diff
@@ -151,27 +71,88 @@ def run_service(self) -> None:
...
+        if asyncio.iscoroutinefunction(f):
+
+            async def async_endpoint(body: RequestBodyModel):  # type: ignore
+                try:
+                    return await f(
+                        **{name: getattr(body, name) for name in body.model_fields}
+                    )
+                except Exception as e:
+                    logger.exception(f"Error in {func.__name__}: {e}")
+                    return responses.JSONResponse(status_code=500, content=e)
```
Pwuts:
```suggestion
                    raise HTTPException(status_code=500, detail=e)
```

autogpt_platform/backend/backend/util/service.py
```diff
@@ -151,27 +71,88 @@ def run_service(self) -> None:
...
+
+            return sync_endpoint
 
     def run(self):
+        self.fastapi_app = FastAPI()
+        # Register the exposed API routes.
+        for attr_name, attr in type(self).__dict__.items():
+            if getattr(attr, "__exposed__", False):
```
Pwuts:
```suggestion
        for attr_name, attr in vars(type(self)).items():
            if getattr(attr, "__exposed__", False):
                # attr is an exposed method
```

autogpt_platform/backend/backend/executor/database.py
Pwuts: nit: if the return type is known to be incorrect, don't use `cast` but `# type: ignore`:
```suggestion
        # TODO:
        #  This function lies about its return type to make the DynamicClient
        #  call the function synchronously, fix this when DynamicClient can choose
        #  to call a function synchronously or asynchronously.
        return expose(f)  # type: ignore
```

Imo type casts should only be used in cases where the type checker can't correctly infer the type but we know what it will be at runtime.

autogpt_platform/backend/backend/util/process.py
```diff
@@ -54,11 +54,11 @@ def cleanup(self):
         """
         pass
 
-    def health_check(self):
+    def health_check(self) -> str:
         """
         A method to check the health of the process.
         """
-        pass
+        return "OK"
```
Pwuts, line 61: Magic value? Are there more magic values?

autogpt_platform/backend/backend/util/service.py
```diff
@@ -44,85 +23,26 @@
 C = TypeVar("C", bound=Callable)
 
 config = Config()
-pyro_host = config.pyro_host
-pyro_config.MAX_RETRIES = config.pyro_client_comm_retry  # type: ignore
-pyro_config.COMMTIMEOUT = config.pyro_client_comm_timeout  # type: ignore
+api_host = config.pyro_host
```
Pwuts, line 58: Is this still something to do with Pyro?

autogpt_platform/backend/backend/util/service.py
```diff
@@ -151,27 +71,88 @@ def run_service(self) -> None:
...
+                        **{name: getattr(body, name) for name in body.model_fields}
+                    )
+                except Exception as e:
+                    logger.exception(f"Error in {func.__name__}: {e}")
+                    return responses.JSONResponse(status_code=500, content=e)
+
+            return async_endpoint
+        else:
+
+            def sync_endpoint(body: RequestBodyModel):  # type: ignore
```
Pwuts: `# type: ignore` for what?

autogpt_platform/backend/backend/util/service.py
```diff
@@ -203,104 +178,92 @@ def cleanup(self):
...
-    def __start_pyro(self):
-        maximum_connection_thread_count = max(
-            Pyro5.config.THREADPOOL_SIZE,
-            config.num_node_workers * config.num_graph_workers,
+    @conn_retry("FastAPI server", "Starting FastAPI server")
+    def __start_fastapi(self):
+        port = self.get_port()
+        host = self.get_host()
+        logger.info(
+            f"[{self.service_name}] Starting FastAPI server at http://{host}:{port}"
```
Pwuts:
```suggestion
            f"[{self.service_name}] Starting RPC server at http://{host}:{port}"
```

autogpt_platform/backend/backend/util/service.py
```diff
@@ -151,27 +71,88 @@ def run_service(self) -> None:
...
+        :param func: The original function (sync/async, bound or unbound)
+        :return: A FastAPI endpoint function.
+        """
+        sig = inspect.signature(func)
+        fields = {}
+
+        is_bounded_method = False
+        for name, param in sig.parameters.items():
+            if name in ("self", "cls"):
+                is_bounded_method = True
```
Pwuts:
```suggestion
        is_bound_method = False
        for name, param in sig.parameters.items():
            if name in ("self", "cls"):
                is_bound_method = True
```

autogpt_platform/backend/backend/util/service.py
```diff
@@ -151,27 +71,88 @@ def run_service(self) -> None:
...
+            )
+
+            # If a default value is provided, use it; otherwise, mark the field as required with '...'
+            default = param.default if param.default != inspect.Parameter.empty else ...
+
+            fields[name] = (annotation, default)
+
+        # Dynamically create a Pydantic model for the request body
+        RequestBodyModel = create_model("RequestBodyModel", **fields)
+        f = func.__get__(self) if is_bounded_method else func
```
Pwuts:
```suggestion
        f = func.__get__(self) if is_bound_method else func
```

autogpt_platform/backend/backend/util/service.py
```diff
@@ -203,104 +178,92 @@ def cleanup(self):
...
-        client.proxy._pyroRelease()
+    def call_method(self, method_name: str, **kwargs) -> Any:
+        try:
+            url = f"{self.base_url}/{method_name}"
+            response = self.client.post(url, json=to_dict(kwargs))
+            response.raise_for_status()
+            return response.json()
+        except httpx.HTTPStatusError as e:
+            logger.error(f"HTTP error in {method_name}: {e.response.text}")
+            raise Exception(e.response.text) from e
```
Pwuts: curious: why wrap the exception rather than just `raise`?

autogpt_platform/backend/backend/util/service.py
```diff
@@ -203,104 +178,92 @@ def cleanup(self):
...
-            annotype is not None
-            and not hasattr(typing, getattr(annotype, "__name__", ""))
-            and isinstance(annotype, type)
-        ):
-            if issubclass(annotype, BaseModel):
-                yield annotype
-            elif annotype not in builtin_types and not issubclass(annotype, Enum):
-                raise TypeError(f"Unsupported type encountered: {annotype}")
+    client = cast(AS, DynamicClient())
+    if client.health_check() != "OK":
```
Pwuts: I guess this works, but `"OK"` is a magic value that can't be differentiated by type from an unhealthy response. I'd suggest letting `health_check` raise an error on failure and treating an error-less call to `health_check` as healthy.

autogpt_platform/backend/backend/executor/database.py
```diff
@@ -46,22 +46,6 @@ def get_port(cls) -> int:
     def send_execution_update(self, execution_result: ExecutionResult):
         self.event_queue.publish(execution_result)
 
-    @staticmethod
```
majdyz, line 49: Moved to service.py

autogpt_platform/backend/backend/util/process.py
majdyz: Nothing magic here, we don't care about its value, we just need an implementation.

autogpt_platform/backend/backend/util/service.py
majdyz: Nope, just retaining the existing config without breaking stuff, we can migrate this in the followup PR when the pyro is completely killed

Pwuts (commented):
Looks good, comments are mostly minor technical.

One question though: Uvicorn's default timeout is 60 seconds. AFAIK Pyro doesn't have that. Do we have any slow RPC's that are at risk of exceeding the 60 second timeout imposed by Uvicorn?

autogpt_platform/backend/backend/util/service.py
majdyz: Added comment

autogpt_platform/backend/backend/util/service.py
majdyz: The exception does not include the content of the response. We need the response of the error.

autogpt_platform/backend/backend/util/service.py
majdyz: health_check has already raised an error when it's called by the client if the server is not healthy, we just need the endpoint to return literally anything.

majdyz:
> One question though: Uvicorn's default timeout is 60 seconds. AFAIK Pyro doesn't have that. Do we have any slow RPC's that are at risk of exceeding the 60 second timeout imposed by Uvicorn?

Not so far, something good that unintentionally introduced by this change :D 

> Also, please fill out the PR checklist as there are major system and config changes.

Filled.

ntindle (approved):

PR #9485, opened 2025-02-15
feat(frontend): Onboarding flow UI

kcze:
### Changes 🏗️
This PR adds interactive UI for the onboarding flow, without any connection to the backend.
Visit `/onboarding` to see it!

- Add Onboarding pages to `app/onboarding/`
- Add Onboarding components to `components/onboarding`

Note:
- Backend isn't connected, so the agents won't run and state isn't preserved
- Onboarding state is lost on refresh

#### For code changes:
- [ ] I have clearly listed my changes in the PR description
- [ ] I have made a test plan
- [ ] I have tested my changes according to the test plan:
  
  - [ ] ...

PR #9511, opened 2025-02-24
feat(platform): Onboarding flow backend

kcze:
This PR adds backend to make Onboarding UI added in https://github.com/Significant-Gravitas/AutoGPT/pull/9485 functional and adds missing confetti screen at the end of Onboarding.

*Make sure to have at least any 2 agents in store when testing locally.* Otherwise the onboarding will finish without showing agents.
Visit `/onboarding/reset` to reset onboarding state, otherwise it'll always redirect to `/library` once finished.

### Changes 🏗️

- Onboarding opens automatically on sign up and login (if unfinished) for all users
- Update db schema to add `UserOnboarding` and add migration
- Add GET and PATCH `/onboarding` endpoints and logic to retrieve and update data Onboarding for a user
- Update `POST /library/agents` endpoint (`addMarketplaceAgentToLibrary`), so it includes input and output nodes; these are needed to know input schema for the Onboarding
- Use new endpoints during onboarding to fetch and update data
- Use agents from the marketplace and their input schema for the onboarding
- Add algorithm to choose store agents based on user answers and related endpoint `GET /onboarding/agents`
- Redirect outside onboarding if finished and resume on proper page
- Add `canvas-confetti` and `@types/canvas-confetti` frontend packages
- Add and show congrats confetti screen when user runs and agent and before opening library
- Minor design updates and onboarding fixes

#### For code changes:
- [x] I have made a test plan
  - [x] Redirect to `/onboarding/*` on sign up and sign in (if onboarding unfinished)
  - [x] Onboarding works and can be finished
  - [x] Agent runs on finish
  - [x] Onboarding state is saved and logging out or refreshing page restores correct state (user choices)
  - [x] When onboarding finished, trying to go into `/onboarding` redirects to `/library`

### DIFF
--- a/autogpt_platform/backend/backend/data/onboarding.py
+++ b/autogpt_platform/backend/backend/data/onboarding.py
@@ -1,5 +1,5 @@
 import re
-from typing import Any, Dict, List, Optional
+from typing import Any, Optional
 
 import prisma
 import pydantic
@@ -17,7 +17,7 @@ from backend.server.v2.store.db import get_store_agent_details
 from backend.server.v2.store.model import StoreAgentDetails
 
 # Mapping from user reason id to categories to search for when choosing agent to show
-REASON_MAPPING: Dict[str, List[str]] = {
+REASON_MAPPING: dict[str, list[str]] = {
     "content_marketing": ["writing", "marketing", "creative"],
     "business_workflow_automation": ["business", "productivity"],
     "data_research": ["data", "research"],
@@ -29,7 +29,7 @@ REASON_MAPPING: Dict[str, List[str]] = {
 class UserOnboardingUpdate(pydantic.BaseModel):
     step: int
     usageReason: Optional[str] = None
-    integrations: List[str] = pydantic.Field(default_factory=list)
+    integrations: list[str] = pydantic.Field(default_factory=list)
     otherIntegrations: Optional[str] = None
     selectedAgentCreator: Optional[str] = None
     selectedAgentSlug: Optional[str] = None


autogpt_platform/backend/backend/server/v2/library/db.py
```diff
@@ -371,51 +372,62 @@ async def add_store_agent_to_library(
...
-            logger.warning(
-                f"User #{user_id} attempted to add their own agent to their library"
+            if existing_library_agent:
+                logger.debug(
+                    f"User #{user_id} already has agent #{store_agent.id} in their library"
+                )
+                # Even if agent exists it needs to be marked as not deleted
+                await set_is_deleted_for_library_agent(
+                    user_id, store_agent.id, store_agent.version, False
+                )
```
Pwuts:
```suggestion
                if existing_library_agent.isDeleted:
                    # Even if agent exists it needs to be marked as not deleted
                    await set_is_deleted_for_library_agent(
                        user_id, store_agent.id, store_agent.version, False
                    )
                else:
                    logger.debug(
                        f"User #{user_id} already has agent #{store_agent.id} in their library"
                    )
```

autogpt_platform/backend/backend/server/v2/library/db.py
```diff
@@ -371,51 +372,62 @@ async def add_store_agent_to_library(
...
-            where={
-                "userId": user_id,
-                "agentId": store_agent.id,
-                "agentVersion": store_agent.version,
-            }
-        )
-        if existing_library_agent:
             logger.debug(
-                f"User #{user_id} already has agent #{store_agent.id} in their library"
+                f"Added agent #{store_agent.id} to library for user #{user_id}"
```
Pwuts:
```suggestion
                f"Added store agent #{store_agent.id} to library for user #{user_id}"
```

autogpt_platform/backend/backend/data/onboarding.py
```diff
@@ -0,0 +1,246 @@
...
+        data={
+            "create": {"userId": user_id, **update},  # type: ignore
+            "update": update,
+        },
+    )
+
+
+def clean_and_split(text: str) -> list[str]:
+    """
+    Removes all special characters from a string and splits it by whitespace and commas.
```
Pwuts:
```suggestion
    Removes all special characters from a string, truncates it to 100 characters,
    and splits it by whitespace and commas.
```

autogpt_platform/backend/backend/server/v2/library/db.py
```diff
@@ -371,51 +372,62 @@ async def add_store_agent_to_library(
     )
 
     try:
-        store_listing_version = (
-            await prisma.models.StoreListingVersion.prisma().find_unique(
-                where={"id": store_listing_version_id}, include={"Agent": True}
+        async with db.locked_transaction(f"user_trx_{user_id}"):
```
Pwuts: curious: what does this need to be a locked transaction for (as opposed to a regular transaction)?

autogpt_platform/backend/schema.prisma
```diff
@@ -57,6 +58,25 @@ model User {
...
+  usageReason          String?
+  integrations         String[] @default([])
+  otherIntegrations    String?
+  selectedAgentCreator String?
+  selectedAgentSlug    String?
+  agentInput           Json?
+  isCompleted          Boolean  @default(false)
+
+  userId String @unique
+  user   User   @relation(fields: [userId], references: [id])
```
Pwuts: when the `User` object is deleted we can just delete the associated `UserOnboarding` right?
```suggestion
  user   User   @relation(fields: [userId], references: [id], onDelete: Cascade)
```

autogpt_platform/backend/backend/server/v2/library/db.py
kcze: It was adding two library agents in dev mode

autogpt_platform/backend/schema.prisma
kcze: yes it can, updated

autogpt_platform/backend/backend/server/v2/library/db.py
Pwuts: 🤔 that's weird, wouldn't that violate a `UNIQUE` constraint in the DB?

autogpt_platform/backend/schema.prisma
```diff
@@ -72,7 +72,7 @@ model UserOnboarding {
   isCompleted          Boolean  @default(false)
 
   userId String @unique
-  user   User   @relation(fields: [userId], references: [id])
+  user   User   @relation(fields: [userId], references: [id], onDelete: Cascade)
```
Pwuts: This also needs to be updated in the corresponding migration

autogpt_platform/backend/backend/server/v2/library/db.py
kcze: I was seeing duplicated agents; `id` isn't provided in the create query, so no

ntindle:
image
Credentials need a flow

No inputs needs a flow (for followup)

Requiring two agents in store breaks local installs

autogpt_platform/backend/schema.prisma
```diff
@@ -57,6 +58,25 @@ model User {
...
+  createdAt DateTime @default(now())
+
+  step                 Int      @default(0)
+  usageReason          String?
+  integrations         String[] @default([])
+  otherIntegrations    String?
+  selectedAgentCreator String?
+  selectedAgentSlug    String?
+  agentInput           Json?
+  isCompleted          Boolean  @default(false)
```
ntindle, line 72: Please STRONGLY consider enums for step and isCompleted

if we add a new stage, the both will be in weird places

step will need all previous bumped, not bumped or some other behavior
is completed will be both true and false

```prisma
enum EducationCompletionStatus {
  COMPLETED
  UPDATED_AFTER_COMPLETION
  IN_PROGRESS
  NOT_STARTED
}

enum OnboardingCompletedPages {
 SELECT_GOAL_TRAITS // productivity vs ai research page
 AGENT_RUN // page with inputs
 ROLE_AT_COMPANY // imaginary new field we want 
}
```

```
isCompleted EducationCompletionStatus @default(EducationCompletionStatus.NOT_STARTED)
completedSteps OnboardingCompletedPages[] @default([])
```

For both of these examples, we can reasonably decide how to resolve all states.

if user.EducationCompletionStatus is UPDATED_AFTER_COMPLETION and the page stands alone like ROLE_AT_COMPANY we can reasonably show it to existing users, but if its something like AGENT_RUN we probably should just not

kcze: I'll do the enum and array for completed steps but I don't understand how completion status would work. Users will either finish onboarding or not. `UPDATED_AFTER_COMPLETION` sounds like onboarding updated after user completed it? So if we need to traverse db and switch completion status on some condition that's the same as switching from completed true to false.

autogpt_platform/backend/backend/server/v2/library/db.py
```diff
@@ -362,56 +363,70 @@ async def add_store_agent_to_library(
...
-            )
-            raise store_exceptions.AgentNotFoundError(
-                f"Store listing version {store_listing_version_id} not found or invalid"
-            )
-
-        graph = store_listing_version.Agent
-        if graph.userId == user_id:
-            logger.warning(
-                f"User #{user_id} attempted to add their own agent to their library"
+        async with locked_transaction(f"user_trx_{user_id}"):
```
Pwuts, line 366: this probably shouldn't use the same key as credit transactions

autogpt_platform/backend/backend/data/onboarding.py
```diff
@@ -0,0 +1,247 @@
...
+    UserOnboarding,
+)
+from prisma.types import UserOnboardingUpdateInput
+
+from backend.server.v2.library.db import set_is_deleted_for_library_agent
+from backend.server.v2.store.db import get_store_agent_details
+from backend.server.v2.store.model import StoreAgentDetails
+
+# Mapping from user reason id to categories to search for when choosing agent to show
+REASON_MAPPING: Dict[str, List[str]] = {
```
Pwuts: nit:
```suggestion
REASON_MAPPING: dict[str, dist[str]] = {
```

Pwuts (approved):
Let's merge now and fix the remaining comments in a follow-up PR

PR #9514, opened 2025-02-25
feat(blocks): Enabled block Usage for Smart Decision Maker Block

Swiftyos:
Originally we did not allow Blocks to be used as tools due to the limitations of communicating the correct tool function signatures.

It has however, been decided to allow them to be used knowing that there are limitations with them.

### Changes 🏗️

- Added ability to execute blocks as tools

### Checklist 📋

Screenshot 2025-02-25 at 12 49 26
 - create an agent with an SDM block and a block as a tool
 - run agent and make sure the block can be called as a tool

#### For code changes:
- [x] I have made a test plan


autogpt_platform/frontend/src/hooks/useAgentGraph.tsx
```diff
@@ -814,7 +814,7 @@ export default function useAgentGraph(
                   ?.name?.toLowerCase()
                   .replace(/ /g, "_") || "agentexecutorblock"
               : "agentexecutorblock"
-            : sinkNode.data.title.toLowerCase().replace(/ /g, "_")
+            : sinkNode.data.title.toLowerCase().replace(/ /g, "_").split("_")[0]
```
majdyz, line 823: What the split is for ?
Will this cause any risk of non smart decision block code to be broken due to the split ?

autogpt_platform/backend/backend/blocks/smart_decision_maker.py
```diff
@@ -6,9 +6,18 @@
...
+from backend.blocks.agent import AgentExecutorBlock
 from backend.data.block import Block, BlockCategory, BlockOutput, BlockSchema, BlockType
 from backend.data.model import SchemaField
 
+
+# Import AVAILABLE_BLOCKS only when needed to prevent circular imports
+def get_available_blocks():
+    from backend.blocks import AVAILABLE_BLOCKS
+
+    return AVAILABLE_BLOCKS
```
majdyz: Use `def get_blocks` instead of this.

autogpt_platform/backend/backend/blocks/smart_decision_maker.py
```diff
@@ -130,6 +139,124 @@ def _get_tool_graph_metadata(self, node_id: str, graph: Any) -> List[Any]:
...
+            "strict": True,
+        }
+
+        return {"type": "function", "function": tool_function}
+
+    @staticmethod
+    def _create_agent_function_signature(
+        sink_node: Any, links: list[Any], tool_graph_metadata: list[Any]
+    ) -> dict[str, Any]:
+        """s∫
```
majdyz: Random character added

autogpt_platform/backend/backend/blocks/smart_decision_maker.py
```diff
@@ -130,6 +139,124 @@ def _get_tool_graph_metadata(self, node_id: str, graph: Any) -> List[Any]:
...
+        properties = {}
+        required = []
+
+        for link in links:
+            sink_block_input_schema = block().input_schema
+            description = (
+                sink_block_input_schema.model_fields[link.sink_name].description
+                if link.sink_name in sink_block_input_schema.model_fields
+                and sink_block_input_schema.model_fields[link.sink_name].description
+                else f"The {link.sink_name} of the tool"
```
majdyz, line 179: What is "The {link.sink_name} of the tool" supposed to represents ?

autogpt_platform/backend/backend/blocks/smart_decision_maker.py
```diff
@@ -130,6 +139,124 @@ def _get_tool_graph_metadata(self, node_id: str, graph: Any) -> List[Any]:
...
+            None,
+        )
+
+        if not sink_graph_meta:
+            raise ValueError(
+                f"Sink graph metadata not found: {graph_id} {graph_version}"
+            )
+
+        tool_function: dict[str, Any] = {
+            "name": re.sub(r"[^a-zA-Z0-9_-]", "_", sink_graph_meta.name).lower(),
```
majdyz, line 232: I see this being used in two places, is it actually fine to replace the name here ? why is it needed ?

autogpt_platform/backend/backend/blocks/smart_decision_maker.py
```diff
@@ -130,6 +139,124 @@ def _get_tool_graph_metadata(self, node_id: str, graph: Any) -> List[Any]:
...
+
+        tool_function["parameters"] = {
+            "type": "object",
+            "properties": properties,
+            "required": required,
+            "additionalProperties": False,
+            "strict": True,
+        }
+
+        return {"type": "function", "function": tool_function}
```
majdyz, line 260: nit: Can we ask LLM to unify these two functions to avoid replication ?

majdyz (dismissed):

autogpt_platform/backend/backend/blocks/smart_decision_maker.py
Swiftyos: lol, I was splitting it out so that we didn't end up failing the code quality checks due to having too complex a function.

autogpt_platform/backend/backend/blocks/smart_decision_maker.py
Swiftyos: This is so the system doesn't break when someone gives the agent graph a complex name

autogpt_platform/backend/backend/blocks/smart_decision_maker.py
Swiftyos: This is what is added in the case no description is present. Without a description there is very little that can be done. This just uses the name of the variable in a sentence as a description.

autogpt_platform/backend/backend/blocks/smart_decision_maker.py
Swiftyos: Okay thanks

autogpt_platform/frontend/src/hooks/useAgentGraph.tsx
Swiftyos: This split is because when adding blocks it adds an identifier as part of the sinkNode.data.title, either a hash or a number.

This is problematic for the sdm block, so I removed it using the split.

This will not cause risk in non smart decision block code as its within an if statement that limits the logic to only apply for the smart decision node

PR #9520, opened 2025-02-25
feat(backend): Add cost on node & graph execution stats

majdyz:
We want the agent run data to be accurate, which means we need to collect it in the right place and with the correct data

### Changes 🏗️

- Updates email templates
- Updates how we collect the data for the agent run event

#### For code changes:
- [x] I have made a test plan
  
  - [x] Run agents and read the email we get

### DIFF
--- a/autogpt_platform/backend/backend/data/notifications.py
+++ b/autogpt_platform/backend/backend/data/notifications.py
@@ -39,7 +39,7 @@ class AgentRunData(BaseNotificationData):
     execution_time: float
     node_count: int = Field(..., description="Number of nodes executed")
     graph_id: str
-    outputs: dict[str, Any] = Field(..., description="Outputs of the agent")
+    outputs: list[dict[str, Any]] = Field(..., description="Outputs of the agent")
 
 
 class ZeroBalanceData(BaseNotificationData):
--- a/autogpt_platform/backend/backend/executor/manager.py
+++ b/autogpt_platform/backend/backend/executor/manager.py
@@ -201,6 +201,7 @@ def execute_node(
         extra_exec_kwargs[field_name] = credentials
 
     output_size = 0
+    cost = 0
     try:
         # Charge the user for the execution before running the block.
         cost = db_client.spend_credits(data)
@@ -273,6 +274,7 @@ def execute_node(
             execution_stats.update(node_block.execution_stats)
             execution_stats["input_size"] = input_size
             execution_stats["output_size"] = output_size
+            execution_stats["cost"] = cost
 
 
 def _enqueue_next_nodes(
@@ -676,6 +678,7 @@ class Executor:
             "nodes_walltime": 0,
             "nodes_cputime": 0,
             "node_count": 0,
+            "cost": 0,
         }
         error = None
         finished = False
@@ -713,6 +716,7 @@ class Executor:
                         exec_stats["node_count"] += 1
                         exec_stats["nodes_cputime"] += result.get("cputime", 0)
                         exec_stats["nodes_walltime"] += result.get("walltime", 0)
+                        exec_stats["cost"] += result.get("cost", 0)
 
                 return callback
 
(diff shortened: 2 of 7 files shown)


autogpt_platform/backend/backend/util/text.py
```diff
@@ -1,21 +1,68 @@
 import logging
 
 import bleach
+
+# Import the CSS sanitizer
```
majdyz: I think most of us will know without the comment

autogpt_platform/backend/backend/notifications/templates/agent_run.html.jinja2
```diff
@@ -15,61 +15,80 @@ data.outputs: the dict[str, Any] of outputs of the agent
...
+        {% endif %}
+      {% else %}
+        {{ value }}
+      {% endif %}
+      </div>
+      {% endif %}
+    {% endfor %}
+  </div>
+  {% endfor %}
+</div>
```
majdyz: Nit new line

autogpt_platform/backend/backend/util/text.py
```diff
@@ -1,21 +1,68 @@
...
+from bleach.css_sanitizer import CSSSanitizer
 from jinja2 import BaseLoader
 from jinja2.sandbox import SandboxedEnvironment
 from markupsafe import Markup
 
 logger = logging.getLogger(__name__)
 
 
+def format_filter(value, format_string=None):
```
majdyz: More specific naming for this ?

autogpt_platform/backend/backend/data/notifications.py
```diff
@@ -39,7 +39,7 @@ class AgentRunData(BaseNotificationData):
     execution_time: float
     node_count: int = Field(..., description="Number of nodes executed")
     graph_id: str
-    outputs: dict[str, Any] = Field(..., description="Outputs of the agent")
+    outputs: list[dict[str, Any]] = Field(..., description="Outputs of the agent")
```
majdyz, line 42: You can use `CompletedBlockOutput` type

autogpt_platform/backend/backend/executor/manager.py
```diff
@@ -657,6 +642,52 @@ def on_graph_execution(
         )
         cls.db_client.send_execution_update(result)
 
+        metadata = cls.db_client.get_graph_metadata(
+            graph_exec.graph_id, graph_exec.graph_version
+        )
+        assert metadata is not None
+        outputs = cls.db_client.get_execution_results(graph_exec.graph_exec_id)
+        logger.info(f"{outputs=}")
```
majdyz: use log_metadata.info

majdyz: But the log seems unnecessary, remove it instead ?

autogpt_platform/backend/backend/executor/manager.py
```diff
@@ -657,6 +642,52 @@ def on_graph_execution(
...
+
+        # Collect named outputs as a list of dictionaries
+        named_outputs = []
+        for output in outputs:
+            if output.output_data and "name" in output.output_data:
```
majdyz: what's with the output pin named "name" ? 
Is this `AgentOutputBlock` specific can we just filter it using the block_id?

named_outputs = {
    output.output_data
    for output in outputs if output.block_id == AgentOutputBlock().id
}

autogpt_platform/backend/backend/executor/manager.py
ntindle: The output is unneeded. Good spot

autogpt_platform/backend/backend/executor/manager.py
ntindle: Good idea

autogpt_platform/backend/backend/data/notifications.py
ntindle: This does not work so good because that type is not hashable which means dict comprehension does not work

autogpt_platform/backend/backend/executor/manager.py
ntindle: didn't quite end up doing it this way but did swap the check out to check the id

ntindle (approved):

ntindle (approved):

PR #9527, opened 2025-02-26
feat(platform): Cost indication for agent runs

### LINKED ISSUE #9181: Add agent run cost indication
* Part of [#8780](https://linear.app/autogpt/issue/OPEN-2088/implement-new-agent-runs-page)

### Changes needed

- [X] Add DB relationship between `CreditTransaction` and `AgentNodeExecution`
- [ ] ~~Implement method to fetch/query/calculate the total cost of a graph execution~~

Pwuts:
- Resolves #9181

### Changes 🏗️

- Add agent run cost indication to `/monitoring` and `/library/agents/[id]`

Backend:
- Add `GraphExecutionMeta.cost` property - value sourced from `AgentGraphExecution.stats["cost"]`

#### For code changes:
- [x] I have made a test plan
  - Run an agent
  - [ ] Check out the agent run on `/library/agents/[id]` -> should show cost
  - [ ] Check out the agent run on `/monitoring` -> should show cost

### DIFF
--- a/autogpt_platform/backend/backend/data/credit.py
+++ b/autogpt_platform/backend/backend/data/credit.py
@@ -249,7 +249,6 @@ class UserCreditBase(ABC):
         metadata: Json,
         new_transaction_key: str | None = None,
     ):
-
         transaction = await CreditTransaction.prisma().find_first_or_raise(
             where={"transactionKey": transaction_key, "userId": user_id}
         )
@@ -346,7 +345,6 @@ class UsageTransactionMetadata(BaseModel):
 
 
 class UserCredit(UserCreditBase):
-
     @thread_cached
     def notification_client(self) -> NotificationManager:
         return get_service_client(NotificationManager)
@@ -840,7 +838,6 @@ class UserCredit(UserCreditBase):
         transaction_time_ceiling: datetime | None = None,
         transaction_type: str | None = None,
     ) -> TransactionHistory:
-
         transactions_filter: CreditTransactionWhereInput = {
             "userId": user_id,
             "isActive": True,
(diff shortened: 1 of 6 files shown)


autogpt_platform/backend/backend/data/graph.py
```diff
@@ -596,15 +599,25 @@ async def get_graphs_executions(user_id: str) -> list[GraphExecutionMeta]:
         where={"userId": user_id},
         order={"createdAt": "desc"},
     )
-    return [GraphExecutionMeta.from_db(execution) for execution in executions]
+    return [
+        GraphExecutionMeta.from_db(
+            execution, await get_graph_execution_cost(execution.id)
```
Pwuts: I'm not sure how much extra % of time/resources it costs to do this on every `GraphExecutionMeta` fetch. It would probably be a lot more efficient to do this through a `JOIN`, but we chose against that earlier. cc @majdyz

ntindle (dismissed):

autogpt_platform/backend/backend/data/credit.py
```diff
@@ -1023,3 +1020,26 @@ async def get_auto_top_up(user_id: str) -> AutoTopUpConfig:
         return AutoTopUpConfig(threshold=0, amount=0)
 
     return AutoTopUpConfig.model_validate(user.topUpConfig)
+
+
+async def get_graph_execution_cost(graph_exec_id: str) -> int | None:
```
majdyz: Cost is now already available in the `cost` field inside the `stats` column of the execution, we should just use that to avoid another nested db lookup https://github.com/Significant-Gravitas/AutoGPT/blob/52ee7d150e6bac2db028b4fcbe0aa94cd7362bfe/autogpt_platform/backend/backend/executor/manager.py#L747

Pwuts: Thanks, that's a lot easier :)

majdyz (approved):

PR #9528, opened 2025-02-26
fix(backend): Use Pyro for RPC by default

Pwuts:
- Follow-up to #9508 

HTTP-based RPC has not been fully tested and should be disabled by default.

### Changes 🏗️

- Disable HTTP-based RPC and use Pyro by default


Swiftyos (approved):
Agreed

PR #9534, opened 2025-02-27
feat(backend): low balance notiifcation

ntindle:
For emailing, we want the user to know when an agent stopped because their balance was too low. This is the first step of that. 

### Changes 🏗️
- Raise InsufficientBalanceError from credit system rather than value error when user runs out of money
- Handle when an agent output isn't hooked up well
- Fix the contents of the email for low balance to be a bit more aligned with the PRD
- expose the topup intent from the db manager
- objectify the execution stats so we can pass it around a bit more type safe
- extract the notification stuff in manager into a function

#### For code changes:
- [x] I have made a test plan
  
  - [x] Set balance to $0.01
  - [x] Run an agent that costs something more than $0.01
  - [x] Check you get an email
  - [x] Check your top up link works


autogpt_platform/backend/backend/executor/manager.py
```diff
@@ -658,6 +653,8 @@ def on_graph_execution(
         )
         cls.db_client.send_execution_update(result)
 
+        cls._handle_agent_run_notif(graph_exec, timing_info.wall_time, exec_stats)
```
ntindle, line 656: this never triggers??

autogpt_platform/backend/backend/executor/manager.py
```diff
@@ -594,6 +581,14 @@ def _on_node_execution(
                 log_metadata.info(
                     f"Failed node execution {node_exec.node_exec_id}: {e}"
                 )
+                if (
```
ntindle: this will trigger multiple times

autogpt_platform/backend/backend/executor/manager.py
```diff
@@ -763,6 +762,14 @@ def callback(result: object):
                 f"Failed graph execution {graph_exec.graph_exec_id}: {e}"
             )
             error = e
+            if (
```
ntindle: this one never triggers :(

autogpt_platform/backend/backend/util/exceptions.py
```diff
@@ -4,3 +4,18 @@ class MissingConfigError(Exception):
...
+    balance: float
+    amount: float
+
+    def __init__(self, message: str, user_id: str, balance: float, amount: float):
+        super().__init__(message)
+        self.args = (message, user_id, balance, amount)
+        self.message = message
+        self.user_id = user_id
+        self.balance = balance
+        self.amount = amount
```
majdyz, line 21: What does it look like on the block error output?
Does it only show the message only?
What about when printed as an exception?

autogpt_platform/backend/backend/data/execution.py
```diff
@@ -21,6 +21,35 @@
...
+    walltime: float = 0
+    cputime: float = 0
+    nodes_walltime: float = 0
+    nodes_cputime: float = 0
+    node_count: int = 0
+    node_error_count: int = 0
+    cost: float = 0
+
+
+class ExecutionStatsStrErr(BaseModel):
```
majdyz: Why can't I extend the above class and override the error field

majdyz: Also, do we really need two versions of this? Can we just stick to this class only ?

autogpt_platform/backend/backend/data/execution.py
```diff
@@ -21,6 +21,35 @@
 from backend.util.settings import Config
 
 
+class ExecutionStats(BaseModel):
```
majdyz: GraphExecutionStats ?

autogpt_platform/backend/backend/executor/database.py
```diff
@@ -74,6 +78,7 @@ def send_execution_update(self, execution_result: ExecutionResult):
 
     # Credits
     spend_credits = exposed_run_and_wait(_spend_credits)
+    top_up_intent = exposed_run_and_wait(_top_up_intent)
```
majdyz: I don't see where this function is used.

autogpt_platform/backend/backend/executor/manager.py
```diff
@@ -569,6 +570,9 @@ def _on_node_execution(
                     f"Failed node execution {node_exec.node_exec_id}: {e}"
                 )
 
+            if stats is not None:
+                stats["error"] = e
```
majdyz: Request: can we make sure the error being persisted here in the db (`AgentNodeExecution` table) is just a message, not a full stacktrace ?

majdyz: if it's a full stacktrace we should exclude it from the stats dict when persisting.

autogpt_platform/backend/backend/executor/manager.py
```diff
@@ -791,6 +772,72 @@ def callback(result: object):
...
+                outputs=named_outputs,
+                agent_name=metadata.name,
+                credits_used=exec_stats.cost,
+                execution_time=exec_stats.walltime,
+                graph_id=graph_exec.graph_id,
+                node_count=exec_stats.node_count,
+            ).model_dump(),
+        )
+
+        logger.info(
```
majdyz: we've been logging at `queue_notification`

autogpt_platform/backend/backend/executor/manager.py
```diff
@@ -773,6 +744,16 @@ def callback(result: object):
                         execution.wait(3)
 
             log_metadata.info(f"Finished graph execution {graph_exec.graph_exec_id}")
+
+            if isinstance(low_balance_error, InsufficientBalanceError):
```
majdyz, line 751: Shoudn't this is not None check only ?

autogpt_platform/backend/backend/notifications/templates/low_balance.html.jinja2
```diff
@@ -0,0 +1,114 @@
...
+    font-family: 'Poppins', sans-serif;
+    color: #070629;
+    font-size: 16px;
+    line-height: 150%;
+    margin-top: 30px;
+    margin-bottom: 10px;
+    font-style: italic;
+  ">
+  This is an automated notification. Your agent is stopped and will need manually restarted unless set to trigger automatically.
+</p>
```
majdyz: new line

autogpt_platform/backend/backend/data/execution.py
ntindle: Exception isn't serializable and pydantic will complain about it, figured this was the easiest way without building and registering searlizers while still keeping the data I need for it for handling the errors

autogpt_platform/backend/backend/executor/database.py
ntindle: Twas removed because the button was swapped to billing. Good spot

autogpt_platform/backend/backend/executor/manager.py
ntindle: Why not the whole thing sorry?

autogpt_platform/backend/backend/executor/manager.py
```diff
@@ -618,58 +623,21 @@ def on_graph_execution(
...
             graph_exec, cancel, log_metadata
         )
-        exec_stats["walltime"] = timing_info.wall_time
-        exec_stats["cputime"] = timing_info.cpu_time
-        exec_stats["error"] = str(error) if error else None
+        exec_stats.walltime = timing_info.wall_time
+        exec_stats.cputime = timing_info.cpu_time
+        exec_stats.error = error
+        str_stats = exec_stats.model_dump()
+        str_stats["error"] = str(error) if error else None
```
ntindle: So like error.message here?

autogpt_platform/backend/backend/notifications/templates/low_balance.html.jinja2
ntindle: This will have the file inserted into the email with an empty line just so you're aware

autogpt_platform/backend/backend/util/exceptions.py
ntindle: Looks like this ('Insufficient balance of $0.01, where this will cost $0.04', '8c18fc64-d4d3-42d9-9f10-c4d2f4657f9a', 1, -4)
image

autogpt_platform/backend/backend/data/execution.py
ntindle:
```python
/Users/ntindle/code/agpt/AutoGPT/autogpt_platform/backend/backend/data/execution.py
  /Users/ntindle/code/agpt/AutoGPT/autogpt_platform/backend/backend/data/execution.py:43:5 - error: "error" overrides symbol of same name in class "ExecutionStats"
    Variable is mutable so its type is invariant
      Override type "str | None" is not the same as base type "Exception | None" (reportIncompatibleVariableOverride)
```
    
We have a rule against it, and pydantic will serialize a shared object (was also tested) given the opportunity so we'd be doing fucky stuff with `del attr` or set some fields to `None` again

autogpt_platform/backend/backend/util/exceptions.py
ntindle: image
fixed it to this

autogpt_platform/backend/backend/data/execution.py
ntindle: can do optional[whatever] but then you just end up with a bunch of null fields in the db

autogpt_platform/backend/backend/executor/manager.py
ntindle: imo no harm in this

autogpt_platform/backend/backend/data/model.py
```diff
@@ -402,3 +402,96 @@ class RefundRequest(BaseModel):
...
+    cost: float = 0
+    input_size: int = 0
+    output_size: int = 0
+    llm_call_count: int = 0
+    llm_retry_count: int = 0
+    input_token_count: int = 0
+    output_token_count: int = 0
+
+
+def convert_node_execution_stats(
```
majdyz: nit: instead of creating this generic function out of context, why don't you put this as a class method and use self, instead of stats. e.g: convert_to_str_err or something

autogpt_platform/backend/backend/data/model.py
```diff
@@ -402,3 +402,96 @@ class RefundRequest(BaseModel):
...
+    walltime: float = 0
+    cputime: float = 0
+    nodes_walltime: float = 0
+    nodes_cputime: float = 0
+    node_count: int = 0
+    node_error_count: int = 0
+    cost: float = 0
+
+
+def convert_graph_execution_stats(
```
majdyz: ditto

autogpt_platform/backend/backend/executor/manager.py
```diff
@@ -256,10 +263,12 @@ def update_execution(status: ExecutionStatus) -> ExecutionResult:
 
         # Update execution stats
         if execution_stats is not None:
-            execution_stats.update(node_block.execution_stats)
-            execution_stats["input_size"] = input_size
-            execution_stats["output_size"] = output_size
-            execution_stats["cost"] = cost
+            execution_stats = execution_stats.model_copy(
```
majdyz, line 261: why do we need to copy?

autogpt_platform/backend/backend/util/text.py
```diff
@@ -34,20 +34,49 @@ def __init__(self):
...
+            "border-color",
+            "border-radius",
+            "display",
+            "float",
+            "clear",
+            "position",
+            "top",
+            "right",
+            "bottom",
+            "left",
```
majdyz: nit: you might want to ask llm to sort this, otherwise duplicates will keep being added into the list.

autogpt_platform/backend/backend/data/execution.py
majdyz: IMO, having one none field seems fine compared to complicating the code with more class duplications.

TBF you have introduced more null fields in the db by introducing Pydantic class for the dynamic dict stats, now we are forcing all blocks persisting `llm_token_count` and other non-specific metrics with null/zero values for node execution.

autogpt_platform/backend/backend/executor/manager.py
majdyz: We should avoid persisting long text in our row-based db at all cost, plus our ORM will keep loading this multiple times throughout the node execution.

majdyz (dismissed):

autogpt_platform/backend/backend/executor/manager.py
ntindle: I'm copying over the existing behavior of updating the previous with the new data. This is how you do it in pydantic as far as I know

autogpt_platform/backend/backend/util/text.py
ntindle: done

autogpt_platform/backend/backend/executor/manager.py
ntindle: okay its chill i belive

majdyz (approved):
Thank you!

PR #9543, opened 2025-02-28
docs(platform): Example block

Swiftyos:
This PR introduces `ExampleBlock` and `ExampleTriggerBlock` to showcase best practices for creating blocks, managing credentials, and integrating webhooks. Key changes:

- Added example blocks for standard and trigger-based workflows.
- Included mock credentials and cost configuration.
- Implemented webhook manager for event handling.

**Purpose**: This is an example PR, not intended for merging. We’re requesting reviews to ensure it captures the best practices for block creation and serves as an optimal reference for contributors.

### Checklist
- [ ] Changes clearly listed
- [ ] Reviews requested to confirm best practices


autogpt_platform/backend/backend/blocks/example.py
```diff
@@ -0,0 +1,160 @@
...
+            # The set of categories that the block belongs to.
+            # Each category is an instance of BlockCategory Enum.
+            categories={BlockCategory.BASIC},
+            # The schema, defined as a Pydantic model, for the input data.
+            input_schema=self.ExampleBlockInput,
+            # The schema, defined as a Pydantic model, for the output data.
+            output_schema=self.ExampleBlockOutput,
+            # The list or single sample input data for the block, for testing.
+            # This is an instance of the Input schema with sample values.
+            test_input=self.ExampleBlockInput(
```
ntindle: I think this input needs the credentials too

autogpt_platform/backend/backend/blocks/example.py
```diff
@@ -0,0 +1,160 @@
...
+
+    def my_function_that_can_be_mocked(self, input: str) -> str:
+        logger.info("my_function_that_can_be_mocked called with input: %s", input)
+        return "Hello, world!"
+
+    def run(self, input_data: ExampleBlockInput, **kwargs) -> BlockOutput:
+        try:
+            message = self.my_function_that_can_be_mocked(input_data.greeting)
+            yield "response", {"message": message}
+        except Exception as e:
```
ntindle: This should be handled already I believe by the wrapper on the block, unless you want to handle different types of errors differently in which case you should re-raise cc @majdyz

autogpt_platform/backend/backend/blocks/example.py
```diff
@@ -0,0 +1,160 @@
...
+            logger.error("Error in run method: %s", e)
+            yield "error", str(e)
+
+
+class ExampleTriggerBlock(Block):
```
ntindle: For vague reasons these go in a trigger.py file — mostly consistency

autogpt_platform/backend/backend/blocks/example.py
```diff
@@ -0,0 +1,160 @@
...
+                name="Craig", greeting="Hello", is_funny=True
+            ).model_dump(),
+            # The list or single expected output if the test_input is run.
+            # Each output is a tuple of (output_name, output_data).
+            test_output=[
+                ("response", {"message": "Hello, world!"}),
+            ],
+            # Function names on the block implementation to mock on test run.
+            # Each mock is a dictionary with function names as keys and mock implementations as values.
+            test_mock={
```
ntindle: It’s afaik not possible to mock some calls so a lot of the time static functions are made as part of the block for easy testing to mock everything up at once

```
@staticmethod
def _request_horse(client, name)-> Whatever
    pass
```

autogpt_platform/backend/backend/data/block_cost_config.py
```diff
@@ -267,4 +269,16 @@
         )
     ],
     SmartDecisionMakerBlock: LLM_COST,
+    ExampleBlock: [
+        BlockCost(
```
ntindle, line 273: Keep in mind there’s other types of costs — not saying you have to use them in your example just that they exist

autogpt_platform/backend/backend/integrations/credentials_store.py
```diff
@@ -169,6 +169,14 @@
...
 )
 
+example_credentials = APIKeyCredentials(
+    id="a2b7f68f-aa6a-4995-99ec-b45b40d33498",
+    provider="example-provider",
+    api_key=SecretStr("mock-example-api-key"),
+    title="Use Credits for Example",
+    expires_at=None,
+)
+
```
ntindle, line 179: Also needs to go in array below this and around line 230ish here if we want it to be available as a system credential we fill for users

ntindle (commented):
This is missing 
- _api.py and _auth.py file formats for when you have more than one related block
- How to split blocks up into different files that are related
- best practices/examples of inputs and output formatting such as yielding the list and the items in the list as two separate things

im sure I’ll think of more but this is a good start

autogpt_platform/backend/backend/blocks/example.py
ntindle: Also these typically aren’t done as the object, but it’s not a bad idea

autogpt_platform/backend/backend/data/block_cost_config.py
Swiftyos: I'm not sure how bytes or second type block cost would look yet.

@majdyz can you give example of these 2 types of billing

Swiftyos:
I've made all the changes you've suggested. Though still need to make examples of these:

> best practices/examples of inputs and output formatting such as yielding the list and the items in the list as two separate things

autogpt_platform/backend/.env.example
```diff
@@ -62,6 +62,9 @@ PLATFORM_BASE_URL=http://localhost:3000
 # For the OAuth callback URL, use <your_frontend_url>/auth/integrations/oauth_callback,
 # e.g. http://localhost:3000/auth/integrations/oauth_callback
 
+# Example API Key
+EXAMPLE_API_KEY=
```
ntindle: Move this down past the OAuth stuff

autogpt_platform/backend/backend/blocks/example/_api.py
```diff
@@ -0,0 +1,144 @@
+from __future__ import annotations
+
+import json
+from typing import Any, Dict, Optional
+
+from backend.data.model import APIKeyCredentials
+from backend.util.request import Requests
```
ntindle, line 16: Comment this here that we have our own wrapper

autogpt_platform/backend/backend/blocks/example/_api.py
```diff
@@ -0,0 +1,144 @@
+from __future__ import annotations
+
+import json
```
ntindle: 20% sure we have a json wrapper

autogpt_platform/backend/backend/blocks/example/example.py
```diff
@@ -0,0 +1,115 @@
...
+                "my_function_that_can_be_mocked": lambda *args, **kwargs: "Hello, world!"
+            },
+            # The credentials required for testing the block.
+            # This is an instance of APIKeyCredentials with sample values.
+            test_credentials=TEST_CREDENTIALS,
+            # The type of the block, which is an instance of BlockType Enum.
+            block_type=BlockType.STANDARD,
+            # The webhook configuration for the block, if any.
+            # This can be an instance of BlockWebhookConfig or BlockManualWebhookConfig.
+            webhook_config=None,
```
ntindle: I think we just leave this off for now

autogpt_platform/backend/backend/blocks/example/example.py
```diff
@@ -0,0 +1,115 @@
...
+        )
+
+    @staticmethod
+    def my_static_method(input: str) -> str:
+        logger.info("my_static_method called with input: %s", input)
+        return f"Hello, {input}!"
+
+    def my_function_that_can_be_mocked(
```
ntindle, line 116: Why credentials optional? Imo not good pattern

autogpt_platform/backend/backend/integrations/credentials_store.py
```diff
@@ -225,6 +234,8 @@ def get_all_creds(self, user_id: str) -> list[Credentials]:
         all_credentials.append(ollama_credentials)
 
         # These will only be added if the API key is set
+        if settings.secrets.example_api_key:
```
ntindle, line 237: Forgot this but these need added to the frontend too to hide from the integrations page

ntindle: Actually the provider needs added to the frontend too

Swiftyos: The number of files that need to be modified for adding a new block / integration is concerning me.

Do you think we could reduce that requirement?

autogpt_platform/backend/backend/blocks/example/_api.py
Swiftyos: We do, we are only using the json module for the error that is caught.

I've change the import to specifically import the error

autogpt_platform/backend/backend/integrations/credentials_store.py
ntindle: This is actually something pwuts and I have talked about. Big one would be generating the provider list on frontend and backend but somehow having it still work with typing. I'm pretty sure that's possible with a generated package which we should really consider already for our client and types layer

AutoGPT-Agent:
The PR has several issues that need addressing before it can be approved:
1. The PR template is not properly filled out - the checkboxes for code changes and configuration changes are not marked
2. The test plan section is empty and not filled out despite this being a code change
3. Configuration changes are present (.env.example modified) but the corresponding checklist items are not checked off
4. While the PR description explains the purpose, it doesn't provide a clear list of changes as required

AutoGPT-Agent:
The PR is well structured and follows most requirements but has some notable issues:
1. The PR template is filled out but missing some key checkmarks - particularly around the test plan which seems important for an example implementation
2. While the PR title has a scope 'docs(platform)', this may not be fully accurate since it's adding actual implementation code, not just documentation
3. The template appears to be partially filled but some sections like the test plan details are left unchecked

However, the implementation itself appears solid with appropriate documentation and the author explicitly states this is meant as an example PR for best practices, acknowledging it's not intended for merging. The changes are well documented and the implementation appears complete.

autogpt_platform/backend/backend/blocks/example/_api.py
majdyz: Why don't we use `requests` instead of the class itself.

autogpt_platform/backend/backend/blocks/example/_api.py
```diff
@@ -0,0 +1,128 @@
...
+from json import JSONDecodeError
+from typing import Any, Dict, Optional
+
+from backend.data.model import APIKeyCredentials
+
+# This is a wrapper around the requests library that is used to make API requests.
+from backend.util.request import Requests
+
+
+class ExampleAPIException(Exception):
```
majdyz, line 19: Why do we need to do this ?

autogpt_platform/backend/backend/blocks/example/_api.py
```diff
@@ -0,0 +1,128 @@
...
+    def __init__(self, message: str, status_code: int):
+        super().__init__(message)
+        self.status_code = status_code
+
+
+class ExampleClient:
```
majdyz, line 35: Why is this client boilerplate needed?

autogpt_platform/backend/backend/blocks/example/_api.py
```diff
@@ -0,0 +1,128 @@
...
+        Returns:
+            The parsed JSON response data.
+
+        Raises:
+            ExampleAPIException: If the API request fails.
+        """
+        if not response.ok:
+            try:
+                error_data = response.json()
+                error_message = error_data.get("error", {}).get("message", "")
```
majdyz, line 76: why are we expecting the error message to be within {error: message: ... } format?

autogpt_platform/backend/backend/blocks/example/example.py
```diff
@@ -0,0 +1,143 @@
...
+logger = logging.getLogger(__name__)
+
+
+class ExampleBlock(Block):
+
+    class Input(BlockSchema):
+        name: str = SchemaField(
+            description="The name of the example block", placeholder="Enter a name"
+        )
+        greetings: list[str] = SchemaField(
```
majdyz, line 25: It's good to show features around the field like `advanced` or `hidden` toggle

autogpt_platform/backend/backend/blocks/example/example.py
```diff
@@ -0,0 +1,143 @@
...
+        is_funny: bool = SchemaField(
+            description="Whether the block is funny", placeholder="True", default=True
+        )
+        # Only if the block needs credentials
+        credentials: ExampleCredentialsInput = CredentialsField(
+            description="The credentials for the example block"
+        )
+
+    class Output(BlockSchema):
+        response: dict[str, Any] = SchemaField(
```
majdyz: Since this is a best practice, I think we should avoid using dict and use a proper object return type

autogpt_platform/backend/backend/blocks/example/example.py
```diff
@@ -0,0 +1,143 @@
...
+            description="The response object generated by the example block."
+        )
+        all_responses: list[dict[str, Any]] = SchemaField(
```
majdyz: all_response can confuse the user especially when described as `All the responses from the example block`.
I think this field only represents responses that are grouped within a single block execution.

autogpt_platform/backend/backend/blocks/example/example.py
```diff
@@ -0,0 +1,143 @@
...
+        super().__init__(
+            # The unique identifier for the block, this value will be persisted in the DB.
+            # It should be unique and constant across the application run.
+            # Use the UUID format for the ID.
+            id="380694d5-3b2e-4130-bced-b43752b70de9",
+            # The description of the block, explaining what the block does.
+            description="The example block",
+            # The list of contributors who contributed to the block.
+            # Each contributor is represented by a ContributorDetails object.
+            contributors=[ContributorDetails(name="Craig Swift")],
```
majdyz: Seems like a contributor needs to be in a centralized place.

autogpt_platform/backend/backend/blocks/example/example.py
```diff
@@ -0,0 +1,143 @@
...
+            # Function names on the block implementation to mock on test run.
+            # Each mock is a dictionary with function names as keys and mock implementations as values.
+            test_mock={
+                "my_function_that_can_be_mocked": lambda *args, **kwargs: "Hello, world!"
+            },
+            # The credentials required for testing the block.
+            # This is an instance of APIKeyCredentials with sample values.
+            test_credentials=TEST_CREDENTIALS,
+            # The type of the block, which is an instance of BlockType Enum.
+            block_type=BlockType.STANDARD,
```
majdyz: Won't this block be marked as disabled? If not it will show up on the block list.

autogpt_platform/backend/backend/blocks/example/example.py
```diff
@@ -0,0 +1,143 @@
...
+        )
+
+    @staticmethod
+    def my_static_method(post_greeting: str) -> str:
```
majdyz: Why static, it can't be mocked

autogpt_platform/backend/backend/blocks/example/example.py
```diff
@@ -0,0 +1,143 @@
...
+            graph_exec_id: The ID of the current graph execution.
+            node_exec_id: The ID of the current node execution.
+            user_id: The ID of the user executing the block.
+        """
+        rtn_all_responses: list[dict[str, Any]] = []
+        # Here we deomonstrate best practice for blocks that need to yield multiple items.
+        # We yield each item from the list to allow for operations on each element.
+        # We also yield the complete list for situations when the full list is needed.
+        for greeting in input_data.greetings:
+            full_greeting = ExampleBlock.my_static_method(greeting)
```
majdyz: I don't think this is a good pattern

autogpt_platform/backend/backend/blocks/example/example.py
```diff
@@ -0,0 +1,143 @@
...
+            message = self.my_function_that_can_be_mocked(full_greeting, credentials)
+            rtn_all_responses.append({"message": message})
+            yield "response", {"message": message}
+        yield "all_responses", rtn_all_responses
```
majdyz, line 153: I don't know if this is a good pattern to share with others. It will pollute the output & replicate data unnecessarily. 

@ntindle @Torantulino I know you've been using this approach for batching the responses, is there a way to get around this to be handled natively?

majdyz: Each stream of output will end up having 3 pins, singular data (stream), the batch (list on each block execution, which ends up needing to be batched again across different block executions), and the count.

autogpt_platform/backend/backend/blocks/example/triggers.py
```diff
@@ -0,0 +1,65 @@
...
+    activated when a webhook event is received from the specified provider.
+    """
+
+    class Input(BlockSchema):
+        # The payload field is hidden because it's automatically populated by the webhook
+        # system rather than being manually entered by the user
+        payload: dict = SchemaField(hidden=True)
+
+    class Output(BlockSchema):
+        event_data: dict = SchemaField(
```
majdyz, line 30: Does this have to be a single event_data and be parsed manually later by another block?
The best practice is to avoid returning complex object as much as possible.

autogpt_platform/backend/backend/data/block_cost_config.py
majdyz: It was an over-engineering on my end. Probably we won't end up using it ever.

autogpt_platform/backend/backend/blocks/example/_api.py
Swiftyos: The purpose of this PR is capture the best practices of adding a new integration block. Creating a client seems to be how we have implemented this with other bocks. 

Do you feel this is not best practice? 
If so what would you suggest?

autogpt_platform/backend/backend/blocks/example/_api.py
Swiftyos: This is just an example of an api. It could be anything really.

autogpt_platform/backend/backend/blocks/example/example.py
Swiftyos: Like this?

```python
class Output(BlockSchema):
    response: SomePydanticClass = SchemaField(...
```

autogpt_platform/backend/backend/blocks/example/example.py
Swiftyos: This was added based upon a suggestion by Nick that the best practices was to yield each element of a list and the entire list as well. 

`all_reponses` is the yielding of the complete list, whilst `response` is yielding of each element within the list

autogpt_platform/backend/backend/blocks/example/example.py
Swiftyos: I don't really see the value of the contributors field. 
I feel it maybe better managed based on the git history. 

I'm happy to remove it if its not best practice.

What do you think?

autogpt_platform/backend/backend/blocks/example/example.py
Swiftyos: I do not follow - are you saying this block should be marked as disabled? or that using `BlockType.STANDARD` marks the block as disabled?

autogpt_platform/backend/backend/blocks/example/example.py
Swiftyos: Static has the advantage of being easy to test in general python coding. 

However, in this scenario I can remove static functions for ease of mocking,.

autogpt_platform/backend/backend/blocks/example/example.py
Swiftyos: removing

autogpt_platform/backend/backend/blocks/example/triggers.py
Swiftyos: @Pwuts can you help with best practice here please

autogpt_platform/backend/backend/blocks/example/_api.py
Swiftyos: Following along with other examples I've seen of the _api.py

Happy to remove it if you feel adding custom exceptions is bad practice

autogpt_platform/backend/backend/blocks/example/_api.py
majdyz: Yes I think it makes sense when the SDK support is missing.

But stuff here getting copied over and over without any refactoring: these ended up being duplicated in the HTTP client part:
* Setting up Request class
* Set content type, passing header to Authorization
* Adding to trusted origin (which is not actually needed)

Other issues:
* Return the value of Dict which should be properly typed. (also the seems like our convention is using dict/list | None as opposed to Dict List Optional)

* Creating a custom exception, then catching it within the same class, shouldn't this be the other way, catching an HTTP error and wrapping it to the custom exception (if it's even needed)?

autogpt_platform/backend/backend/blocks/example/_api.py
majdyz: The question is more on whether this work

```
        try:
            response = self._requests.get(
                f"{self.API_BASE_URL}/resources/{resource_id}"
            )
            return self._handle_response(response)
        except ExampleAPIException:
            raise
        except Exception as e:
            raise ExampleAPIException(f"Failed to get resource: {str(e)}", 500)
```

ExampleAPIException is something we are introducing by ourselves, but then we are catching it right after calling the HTTP request, is this a valid flow ?

autogpt_platform/backend/backend/blocks/example/example.py
majdyz: yes

autogpt_platform/backend/backend/blocks/example/example.py
majdyz: @ntindle What can we do to avoid this? 
What happens when there are multiple executions of the blocks, do you end up batching the batch?

autogpt_platform/backend/backend/blocks/example/example.py
majdyz: I agree, but I don't know the historical context of it either, @ntindle  ?

autogpt_platform/backend/backend/blocks/example/example.py
majdyz: There is `disabled=True` param on this __init__

autogpt_platform/backend/backend/blocks/example/example.py
majdyz: yeah better avoid static as an example so everything can be easily mocked for the block test

autogpt_platform/backend/backend/blocks/example/_api.py
ntindle: Depends on the block, some use this to try something or know they've exhausted some resource and try again later

autogpt_platform/backend/backend/blocks/example/_api.py
ntindle: we can use pydantic objects and that should be preferred except when there's a fuck off big response from the API. That case should be model dumpted to a "dict" and attached as a raw object. Github Apollo/smartlead do this

autogpt_platform/backend/backend/blocks/example/example.py
ntindle: likewise the dependent fields with descriminators

autogpt_platform/backend/backend/blocks/example/example.py
ntindle: that can work yes

autogpt_platform/backend/backend/blocks/example/example.py
ntindle: i think moving to pydantic for this would make it clearer 
```
class Horse(BaseModel):
  blah: int

horse: Horse = SchemaField(...)
horses: list[Horse] = SchemaField(...)
```

autogpt_platform/backend/backend/blocks/example/example.py
ntindle: its not used anywhere but was intended to be shown on blocks made by the organization. IMO remove it as a separate PR

autogpt_platform/backend/backend/blocks/example/example.py
ntindle: block_type Standard is normal and should not need specified

autogpt_platform/backend/backend/blocks/example/example.py
ntindle: they won't mock right if not static

autogpt_platform/backend/backend/blocks/example/example.py
ntindle: I don't care about the count but yielding the group and individual is super useful for QoL

autogpt_platform/backend/backend/blocks/example/triggers.py
ntindle: see generic webhook block pr

autogpt_platform/backend/backend/blocks/example/_api.py
majdyz: I'm referring to this:
```
        except ExampleAPIException:
            raise
```

majdyz (commented):
Sorry I've been adding comment and forgot to submit it

autogpt_platform/backend/backend/blocks/example/example.py
majdyz: wdym ?

autogpt_platform/backend/backend/blocks/example/example.py
Swiftyos: What exactly do advanced and hidden do?

I think they just hide if it is shown in the UI - is this correct?

I've added examples of them

autogpt_platform/backend/backend/blocks/example/example.py
Swiftyos: Done

autogpt_platform/backend/backend/blocks/example/example.py
Swiftyos: I've changed it to use pydantic objects, however this is an open discussion in discord regarding best practice for yielding a each element of the list and the list or just each element of the list

autogpt_platform/backend/backend/blocks/example/example.py
Swiftyos: I've removed the static function

autogpt_platform/backend/backend/blocks/example/example.py
Swiftyos: > There is disabled=True param on this init
I can't find that param on this init

Swiftyos: I've removed the block_type Standard

autogpt_platform/backend/backend/blocks/example/example.py
```diff
@@ -0,0 +1,154 @@
...
+            placeholder="True",
+            default=True,
+            # Advanced fields are not shown in the UI by default
+            advanced=True,
+        )
+        greeting_context: str = SchemaField(
+            description="The context of the greeting",
+            placeholder="Enter a context",
+            default="The user is looking for an inspirational greeting",
+            # Hidden fields are not shown in the UI by default
```
Pwuts: so what's the difference between `advanced` and `hidden`?

autogpt_platform/backend/backend/blocks/example/triggers.py
```diff
@@ -0,0 +1,65 @@
...
+    A trigger block that is activated by an external webhook event.
+
+    Unlike standard blocks that are manually executed, trigger blocks are automatically
+    activated when a webhook event is received from the specified provider.
+    """
+
+    class Input(BlockSchema):
+        # The payload field is hidden because it's automatically populated by the webhook
+        # system rather than being manually entered by the user
+        payload: dict = SchemaField(hidden=True)
```
Pwuts, line 27: this should be more specifically typed if possible

autogpt_platform/backend/backend/integrations/webhooks/example.py
```diff
@@ -0,0 +1,145 @@
...
+        Args:
+            credentials (Credentials): The credentials required for authentication.
+            webhook_type (str): The type of webhook to register.
+            resource (str): The resource associated with the webhook.
+            events (list[str]): The list of events to subscribe to.
+            ingress_url (str): The URL where the webhook will send data.
+            secret (str): A secret for securing the webhook.
+
+        Returns:
+            tuple: A tuple containing an empty string and the webhook configuration as a dictionary.
```
Pwuts:
```suggestion
            tuple: A tuple containing the provider's webhook ID, if any, and the webhook configuration as a dictionary.
```

autogpt_platform/backend/backend/integrations/webhooks/example.py
```diff
@@ -0,0 +1,145 @@
...
+
+    async def _register_webhook(
+        self,
+        credentials: Credentials,
+        webhook_type: str,
+        resource: str,
+        events: list[str],
+        ingress_url: str,
+        secret: str,
+    ) -> tuple[str, dict]:
```
kcze, line 62: 1. We return weakly typed `dict` here; if function returns a specific structure it'd be better to return `TypedDict`/`class`/`namedtuple` or even pydantic model.
2. Why return empty string or id as separate item? Can't we just return it inside the `dict`?

autogpt_platform/frontend/src/components/integrations/credentials-input.tsx
kcze, line 1: nit and a bit unrelated: I feel like we could merge this and the two files below, so the frontend integration data is in one file.

autogpt_platform/backend/backend/blocks/example/example.py
kcze: Advanced are shown when user clicks `Advanced` toggle, I guess hidden are permanently hidden.
But then, does it need to be `SchemaField`?
That means it's still sent to the frontend unless this is explicitly avoided and possibly can be force-provided with data from the user, is this ok?

autogpt_platform/backend/backend/blocks/example/_auth.py
```diff
@@ -0,0 +1,37 @@
...
+from backend.integrations.providers import ProviderName
+
+# Define the type of credentials input expected for Example API
+ExampleCredentialsInput = CredentialsMetaInput[
+    Literal[ProviderName.EXAMPLE_PROVIDER], Literal["api_key"]
+]
+
+
+# Mock credentials for testing Example API integration
+TEST_CREDENTIALS = APIKeyCredentials(
```
kcze, line 23: I'd explicitly mention not to include real keys here to not leak them.

autogpt_platform/backend/backend/blocks/example/_api.py
```diff
@@ -0,0 +1,137 @@
...
+
+    def __init__(
+        self,
+        credentials: Optional[APIKeyCredentials] = None,
+        custom_requests: Optional[Requests] = None,
+    ):
+        if custom_requests:
+            self._requests = custom_requests
+        else:
+            headers: Dict[str, str] = {
```
kcze:
```suggestion
            headers: dict[str, str] = {
```

autogpt_platform/backend/backend/blocks/example/triggers.py
kcze: I second this

autogpt_platform/frontend/src/components/integrations/credentials-input.tsx
Swiftyos: I agree, there is already a discussion thread about simplifying the adding of new blocks

AutoGPT-Agent:
The PR is well structured and follows most guidelines, but has some issues: 1) While the PR title has a scope in parentheses, it doesn't follow the exact conventional commit format which should be 'feat(scope): message' rather than docs(scope). 2) There is a partial checklist where some sections are filled but others are left incomplete (like the test plan section which has checkboxes but no indication if they were completed). 3) Changes in data/*.py files correctly handle user_id access.

AutoGPT-Agent:
The PR is well structured and follows the established patterns, but has a few issues that need to be addressed:
1. The PR title should contain a more specific scope than just 'docs' since this includes actual code changes - something like 'feat(blocks)' would be more appropriate
2. The PR template needs to be properly filled out - while they've included some items, the test plan and code changes checklist sections are not fully completed as required
3. The env.example update is documented but not explicitly checked off in the configuration changes section

However, these are relatively minor issues and the actual code changes look solid - proper user_id handling in data files, good documentation, and follows established patterns. The author also explicitly asks for feedback to improve it further.

Swiftyos:
This will be closed when #10074 is merged

PR #9565, opened 2025-03-04
Enhance process management with error handling and lifecycle improvements

rahmanlibi47:
Enhance process management with error handling, lifecycle improvements, and better resource management

- Added robust error handling for process startup failures.
- Implemented proper cleanup of processes during failure scenarios.
- Added detailed error logging for process failures and cleanup operations.
- Enhanced process lifecycle with health checks.
- Added success logging for process terminations.

These updates enhance the reliability and maintainability of the system.


majdyz (approved):
Thank you for the contribution:

> Added robust error handling for process startup failures.
When an error happens, it will break the flow and propagated to the main caller, so there is no need to wrap it. And the finally block already captures its cleanup.

> Implemented proper cleanup of processes during failure scenarios.
The try-catch on stop makes sense, thanks.

> Added detailed error logging for process failures and cleanup operations.
The error log is already propagated and breaks the flow, so it's already logged.

> Enhanced process lifecycle with health checks.
The health check has already been called inside the `process.start` function.

> Added success logging for process terminations.
I've Moved the start & stop logging inside the start & stop function.

PR #9573, opened 2025-03-05
feat(frontend): Automatic loading indicator for button click handlers

Pwuts:
Show that something is happening when a click handler doesn't return immediately.

### Changes 🏗️

- Show a loading indicator on buttons while their click handler is running

#### For code changes:
- [x] I have made a test plan
  - Go to `/library/agents/[id]`, click "Run again" on an existing run
    - [ ] -> loading indicator should show
    - [ ] -> button state should reset once the API request completes
  - [ ] Check that existing `Button` elements aren't adversely affected by the styling changes, especially `overflow-hidden`


Bentlybro (approved):
Tested and this works 😄

PR #9581, opened 2025-03-05
docs(security): Update disclosure timeline

ntindle:
Update the security.md based on some advice we got :)

### Changes 🏗️
- Adds an update time window and clarifies time spans


AutoGPT-Agent:
The PR falls short of requirements in multiple ways: 1) The PR title doesn't follow conventional commit format with scope (should be something like 'docs(security): update disclosure timeline'). 2) The description is very minimal and casual. 3) The checklist sections from the template are completely missing (not just removed as N/A). However, since this is a very simple documentation change to SECURITY.md and we're supposed to be lenient during adoption, these issues aren't severe enough to block the change entirely.

Bentlybro (approved):

PR #9586, opened 2025-03-06
feat(backend): batching operations for queues + agent run become batched

ntindle:
We don't want to spam the user with similar notification types so we want to group them up over a timespan and handle that as a group of notifications.
### Changes 🏗️
- Adds a batch queue
- Moves the ExecutionScheduleur to a generic Scheduler
- Makes the Agent run a batch operation
- Fixes various bugs in how we originally made the batch db models and queries

#### For code changes:
- [x] I have made a test plan
  
  - [x] Run 10 agents back to back
  - [x] Notice how many emails you get
  - [x] Wait a bit and you should after an hour (change the cron rule to speed up testing this) you'll get an email and see all the batches in your db are empty


autogpt_platform/backend/backend/data/notifications.py
```diff
@@ -339,27 +335,28 @@ async def create_or_add_to_user_notification_batch(
...
 
-async def get_user_notification_last_message_in_batch(
+async def get_user_notification_oldest_message_in_batch(
     user_id: str,
     notification_type: NotificationType,
 ) -> NotificationEvent | None:
     try:
         batch = await UserNotificationBatch.prisma().find_first(
             where={"userId": user_id, "type": notification_type},
-            order={"createdAt": "desc"},
```
majdyz, line 356: Don't you still need the order but asc ?

autogpt_platform/backend/backend/data/notifications.py
```diff
@@ -404,3 +401,22 @@ async def get_user_notification_batch(
...
+
+
+async def get_all_batches_by_type(
+    notification_type: NotificationType,
+) -> list[UserNotificationBatch]:
+    try:
+        return await UserNotificationBatch.prisma().find_many(
+            where={
+                "type": notification_type,
+                "notifications": {
```
majdyz, line 413: Will the entry on this table be removed or ever increased?

autogpt_platform/backend/backend/notifications/notifications.py
```diff
@@ -248,12 +494,25 @@ def _run_queue(
     def run_service(self):
         logger.info(f"[{self.service_name}] Started notification service")
 
+        # Set up scheduler for batch processing of all notification types
+        # this can be changed later to spawn differnt cleanups on different schedules
+        try:
+            get_scheduler().add_batched_notification_schedule(
+                notification_types=list(NotificationType),
+                data={},
+                cron="0 * * * *",
```
majdyz, line 494: And why is it hourly ?

majdyz (commented):
Which one goes first ? this one or https://github.com/Significant-Gravitas/AutoGPT/pull/9596 ?

ntindle:
This one goes first

autogpt_platform/backend/backend/data/notifications.py
ntindle: yes they are removed and increased by other stuff on a schedule hourly

autogpt_platform/backend/backend/notifications/notifications.py
ntindle: longest current is an hour so max 1 59/60 hours delay between every batch

autogpt_platform/backend/backend/data/notifications.py
ntindle: not really, it doesn't particularly matter for this use as they're manually sorted by notifications not the batches

autogpt_platform/backend/backend/notifications/notifications.py
```diff
@@ -60,6 +71,16 @@ def create_notification_config() -> RabbitMQConfig:
...
             },
         ),
+        # Batch Queue
+        Queue(
+            name="batch_notifications",
+            exchange=notification_exchange,
+            routing_key="notification.batch.#",
+            arguments={
+                "x-dead-letter-exchange": dead_letter_exchange.name,
+                "x-dead-letter-routing-key": "failed.batch",
```
majdyz, line 78: Shouldn't this become an actual parameter ?

autogpt_platform/backend/backend/notifications/notifications.py
```diff
@@ -91,6 +119,12 @@ def __init__(self):
     def get_port(cls) -> int:
         return settings.config.notification_service_port
 
+    @thread_cached
```
majdyz: Why do we have the exact copy with different name above

autogpt_platform/backend/backend/notifications/templates/agent_run.html.jinja2
```diff
@@ -1,96 +1,142 @@
...
+          {{ value }}
+        {% endif %}
+        </div>
+        {% endif %}
+      {% endfor %}
+    </div>
+    {% endfor %}
+  </div>
+  {% endif %}
+{% endif %}
```
majdyz: nit: newline

majdyz (dismissed):

autogpt_platform/backend/backend/notifications/notifications.py
ntindle: wdym? like on the Queue object?

autogpt_platform/backend/backend/notifications/notifications.py
ntindle: copied the pattern from elsewhere previously and forgot to remove one

autogpt_platform/backend/backend/data/notifications.py
ntindle: no i need the notifiaciots inside ordered

PR #9616, opened 2025-03-11
fix: backend admin page logic was broken

ntindle:
We're building out admin utilities so we need to bring back the `/admin` route with RBAC. This PR goes through re-enabling that to work with the latest changes

### Changes 🏗️
- Adds back removed logic
- Refactors the role checks to fix minor bug for admin page and more importantly clarify
- Updates routes to the latest 

#### For code changes:
- [x] I have made a test plan
  
  - [x] Test with admin and authenticated user roles
  - [x] Test with logged out user role
  - [x] For the above check the all the existing routes + new ones in the `middleware.ts`


autogpt_platform/frontend/src/lib/supabase/middleware.ts
```diff
@@ -62,34 +63,40 @@ export async function updateSession(request: NextRequest) {
...
+    // 2. Check if user is authenticated but lacks admin role when accessing admin pages
+    if (user && userRole !== "admin") {
+      const isAttemptingAdminPage = ADMIN_PAGES.some((page) =>
+        request.nextUrl.pathname.startsWith(page),
+      );
+
+      // If a non-admin user is trying to access admin pages,
+      // redirect to marketplace
+      if (isAttemptingAdminPage) {
+        console.log("Non-admin user redirected to marketplace");
```
majdyz: Remove console.log or make it debug/warning ?

autogpt_platform/frontend/src/lib/supabase/middleware.ts
```diff
@@ -62,34 +63,40 @@ export async function updateSession(request: NextRequest) {
...
+      );
+
+      const isAttemptingAdminPage = ADMIN_PAGES.some((page) =>
+        request.nextUrl.pathname.startsWith(page),
+      );
+
+      // If trying to access any protected content without being logged in,
+      // redirect to login page
+      if (isAttemptingProtectedPage || isAttemptingAdminPage) {
+        console.log("Unauthenticated user redirected to login");
```
majdyz: ditto

autogpt_platform/frontend/src/lib/supabase/middleware.ts
```diff
@@ -62,34 +63,40 @@ export async function updateSession(request: NextRequest) {
...
-      })
-    ) {
-      // no user, potentially respond by redirecting the user to the login page
-      url.pathname = `/login`;
-      return NextResponse.redirect(url);
+    // 1. Check if user is not authenticated but trying to access protected content
+    if (!user) {
+      // Check if the user is trying to access either a protected page or an admin page
+      const isAttemptingProtectedPage = PROTECTED_PAGES.some((page) =>
+        request.nextUrl.pathname.startsWith(page),
```
majdyz, line 70: should we trim the initial slashes ?

majdyz (dismissed):

autogpt_platform/frontend/src/lib/supabase/middleware.ts
ntindle: i don't think so? why would we vs not?

majdyz (approved):

PR #9620, opened 2025-03-12
feat(platform)!: Lock Supabase docker-compose code

majdyz:
We have been submoduling Supabase for provisioning local Supabase instances using docker-compose. Aside from the huge size of unrelated code being pulled, there is also the risk of pulling unintentional breaking change from the upstream to the platform.

The latest Supabase changes hide the 5432 port from the supabase-db container and shift it to the supavisor, the instance that we are currently not using. This causes an error in the existing setup.

## BREAKING CHANGES

This change will introduce different volume locations for the database content, pulling this change will make the data content fresh from the start. To keep your old data with this change, execute this command:
```
cp -r supabase/docker/volumes/db/data db/docker/volumes/db/data
```

### Changes 🏗️

The scope of this PR is snapshotting the current docker-compose code obtained from the Supabase repository and embedding it into our repository. This will eliminate the need for submodule / recursive cloning and bringing the entire Supabase repository into the platform.

#### For code changes:
- [x] I have made a test plan
  
  - [x] Existing CI


Swiftyos (approved):
Good idea, approved. Though all these tests are failing

majdyz:
@Swiftyos Thanks, updated, I've also added the `breaking changes` note on the description.

ntindle:
Marked as breaking via title

ntindle (approved):

PR #9630, opened 2025-03-13
feat(backend): Fix failed RPC on Notification Service

majdyz:
Although returning a Prisma object on an RPC is a bad practice, we have instances where we do so and the type contains a `prisma.Json` field. This Json field can't be seamlessly serialized and then converted back into the Prisma object.

### Changes 🏗️

Replacing prisma object as return type on notification service with a plain pydantic object as DTO.

#### For code changes:
- [x] I have made a test plan
  - [x] Calling notification APIs through the RPC client.


AutoGPT-Agent:
The PR is well-structured but missing some critical elements. While the description explains the technical need and solution, and includes tested changes, the checklist is incomplete (the example test plan section is still visible and not removed as allowed by rules). Additionally, the PR title follows the conventional commit format correctly with feat(backend). The changes don't appear to involve data/*.py files or frontend routes, so those rules aren't applicable. The changes seem reasonable and well-tested for their scope.

autogpt_platform/backend/backend/util/typeshift.py
```diff
@@ -0,0 +1,147 @@
+from __future__ import annotations
```
ntindle: nit: this seems like tests would be a good idea

ntindle (approved):

AutoGPT-Agent:
The PR is missing key information and has potential issues: 1) The title has a scope but it doesn't match what's actually being changed (notifications, not just backend). 2) The test plan is very minimal given the complexity of the notification system changes. 3) There are data/*.py changes where we need to verify user_id handling which isn't addressed. However, looking at the code it appears the user_id is being handled properly in the notification DTOs and the changes appear to be maintaining proper access controls.

autogpt_platform/backend/backend/util/typeshift.py
majdyz: I gave up on this idea, I'll just stick to DTO as return types for all RPC

ntindle (approved):

PR #9632, opened 2025-03-13
fix(frontend): Fill defaults from schema to `hardcodedValues` at node creation

kcze:
Defaults need to be handled as special cases every time there's no `hardcodedValues` in node data. This causes multiple issues such as `useCredentials` not taking into account default model and requiring user to manually switch model back and forth.

### Changes 🏗️

- Set default values from `inputSchema` to `hardcodedValues` when new node is placed.

#### For code changes:
- [x] I have made a test plan
  - [x] Newly placed node has defaults set as `hardcodedValues`
  - [x] AI Blocks: Model is recognised, node shows price and credentials work correctly without the need to switch

### DIFF
--- a/autogpt_platform/frontend/src/components/Flow.tsx
+++ b/autogpt_platform/frontend/src/components/Flow.tsx
@@ -461,6 +461,37 @@ const FlowEditor: React.FC<{
     });
   }, [nodes, setViewport, x, y]);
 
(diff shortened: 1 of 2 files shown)


majdyz (dismissed):
Is this all the required change to fix the bug ?

majdyz (approved):

PR #9657, opened 2025-03-19
feat(block): Add agent input block subtypes

majdyz:
### Changes 🏗️

Added these types of input blocks:
* TextShort
* TextLong
* Number
* Date
* Time
* FileUpload
* Dropdown
* Toggle

#### For code changes:
- [x] I have made a test plan
  - [x] Test in respective block codes.


AutoGPT-Agent:
The PR has several issues:
1. The author missed checking/filling out some important information in the PR template - specifically details about their testing plan beyond just saying 'Test in respective block codes'
2. The title follows conventional commit format correctly
3. The changes are well documented in the description
4. The commits seem focused on the scope mentioned
5. No out of scope changes detected
6. Since these are new input components, no user_id check issues
7. No new frontend base routes added

AutoGPT-Agent:
While the PR has some positive aspects like moving input/output blocks to a dedicated io.py file and adding new input block types, the PR fails to meet the requirements in a few ways: 1) The scope in the title (block) is too vague and should be more specific like 'input-blocks' 2) The changes checklist only marks the code section items but doesn't remove the configuration section which is not applicable 3) The test plan is very minimal just saying 'Test in respective block codes' without specifics. However, these are relatively minor issues that can be addressed with updates.

autogpt_platform/backend/backend/data/model.py
```diff
@@ -144,6 +144,7 @@ def SchemaField(
     depends_on: list[str] | None = None,
     image_upload: Optional[bool] = None,
     image_output: Optional[bool] = None,
+    json_extra: dict[str, Any] | None = None,
```
Pwuts:
```suggestion
    json_schema_extra: dict[str, Any] | None = None,
```

autogpt_platform/backend/backend/blocks/io.py
```diff
@@ -0,0 +1,515 @@
...
+                ("result", "Another multiline text input."),
+            ],
+        )
+
+
+class NumberAgentInputBlock(AgentInputBlock):
+    class Input(AgentInputBlock.Input):
+        value: int = SchemaField(
+            description="Number input.",
+            default=42,
```
Pwuts: I don't think this should have a default value

autogpt_platform/backend/backend/blocks/io.py
```diff
@@ -0,0 +1,515 @@
...
+                },
+            ],
+            test_output=[
+                ("result", 42),
+                ("result", 314),
+            ],
+        )
+
+
+class DateAgentInputBlock(AgentInputBlock):
```
Pwuts: This will show up as "Date Agent Input", right? I suggest changing this to either of
- Agent Date Input
- Date Input

same for the other input blocks with this naming pattern

autogpt_platform/backend/backend/blocks/io.py
```diff
@@ -0,0 +1,515 @@
...
+class FileUploadAgentInputBlock(AgentInputBlock):
+    """
+    A simplified file-upload block. In real usage, you might have a custom
+    file type or handle binary data. Here, we'll store a string path as the example.
+    """
+
+    class Input(AgentInputBlock.Input):
+        value: MediaFile = SchemaField(
+            description="Path or reference to an uploaded file.",
+            default=MediaFile("path/to/file.txt"),
```
Pwuts: This probably also shouldn't have a default value.
```suggestion
```

autogpt_platform/backend/backend/blocks/io.py
Pwuts, line 1: These are agent input blocks: can we let the graph author specify a default value rather than hard-coding it?

autogpt_platform/backend/backend/blocks/io.py
```diff
@@ -0,0 +1,515 @@
+from datetime import date, time
+from typing import Any, List
```
Pwuts: nit: use `list` instead of `List` and avoid the unnecessary import
```suggestion
from typing import Any
```

Also, `list[Any]` is equivalent to just `list` afaik

autogpt_platform/backend/backend/blocks/io.py
Pwuts, line 1: I see a date input and a time input, but no datetime input?

autogpt_platform/frontend/src/components/node-input-components.tsx
```diff
@@ -232,34 +248,131 @@ const NodeDateTimeInput: FC<{
...
+            </Button>
+          )}
+        </div>
+        <input
+          id={`${selfKey}-upload`}
+          type="file"
+          accept="*/*"
+          onChange={handleFileChange}
+          className="hidden"
+        />
```
Pwuts, line 369: https://ui.shadcn.com/docs/components/input#file

autogpt_platform/frontend/src/lib/autogpt-server-api/types.ts
```diff
@@ -500,11 +500,11 @@ export enum BlockUIType {
   WEBHOOK = "Webhook",
   WEBHOOK_MANUAL = "Webhook (manual)",
   AGENT = "Agent",
+  AI = "AI",
```
Pwuts, line 503: I don't see this being used anywhere, was this included by accident?

majdyz: It's just a new type block that is introduced in the backend but not introduced here.

autogpt_platform/frontend/src/components/node-input-components.tsx
majdyz: The whole UI for the file input will be eventually redone, note that the actual input is hidden.

autogpt_platform/backend/backend/blocks/io.py
majdyz: I guess we can combine date and time input?
cc @Torantulino do we want datetime input ?

autogpt_platform/backend/backend/blocks/io.py
majdyz: updated

autogpt_platform/backend/backend/blocks/io.py
majdyz: updated except the dropdown input

autogpt_platform/backend/backend/blocks/io.py
majdyz: Removed

autogpt_platform/backend/backend/blocks/io.py
majdyz: Removed

autogpt_platform/backend/backend/data/model.py
majdyz: Updated

autogpt_platform/backend/backend/blocks/io.py
majdyz: Updated

autogpt_platform/backend/backend/blocks/io.py
Torantulino: We'll add some more going forward for sure but let's proceed with these for the first version and iterate as these are the one's Olivia is expecting.

ntindle (approved):

PR #9660, opened 2025-03-20
fix(backend/ws): Add `user_id` to websocket event subscription key

Pwuts:
- Add `user_id` to WS subscription key
- Add error catching to WS message handler

### DIFF
--- a/autogpt_platform/backend/backend/server/ws_api.py
+++ b/autogpt_platform/backend/backend/server/ws_api.py
@@ -4,6 +4,7 @@ from contextlib import asynccontextmanager
 
 import uvicorn
 from autogpt_libs.auth import parse_jwt_token
+from autogpt_libs.utils.cache import thread_cached
 from fastapi import Depends, FastAPI, WebSocket, WebSocketDisconnect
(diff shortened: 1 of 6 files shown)


autogpt_platform/backend/backend/server/ws_api.py
```diff
@@ -85,20 +96,47 @@ async def handle_subscribe(
...
+        #     user_id=user_id,
+        # ):
+        #     await websocket.send_text(
+        #         WsMessage(
+        #             method=Methods.ERROR,
+        #             success=False,
+        #             error="Access denied",
+        #         ).model_dump_json()
+        #     )
+        #     return
```
Bentlybro, line 114: is this to stay for now?

Bentlybro (approved):

autogpt_platform/backend/backend/server/ws_api.py
Pwuts: yep, this can be re-enabled once we get the `DatabaseManager` deployed into the same pod or otherwise accessible to the `WebsocketServer` service

PR #9667, opened 2025-03-21
feat(backend): Avoid loading all node executions on large continuos agent

majdyz:
A graph that processes tens of thousands of nodes will cripple the system since the API tries to load all of them and dump them into the browser. The scope of this change is to avoid such an issue by only returning the last 1000 node executions.

### Changes 🏗️

* Return only 1000 node executions from `AgentNodeExecutions` reference.
* Unify the include clause for fetching `AgentNodeExecutions` in one place and its format.
* Fix & optimize `cancel_execution` logic to always set both the graph & node execution status in batch.

#### For code changes:
- [x] I have made a test plan
  - [x] Execute a graph in a loop that executes 10000 nodes, it should only display the last 1000 nodes when refreshed. Cancelling the graph should also not cripple the server.

### DIFF
--- a/autogpt_platform/backend/backend/data/execution.py
+++ b/autogpt_platform/backend/backend/data/execution.py
@@ -363,15 +363,21 @@ def _get_update_status_data(
     stats: dict[str, Any] | None = None,
 ) -> AgentNodeExecutionUpdateInput:
     now = datetime.now(tz=timezone.utc)
-    return {
-        **({"executionStatus": status}),
-        **({"queuedTime": now} if status == ExecutionStatus.QUEUED else {}),
-        **({"startedTime": now} if status == ExecutionStatus.RUNNING else {}),
-        **({"endedTime": now} if status == ExecutionStatus.FAILED else {}),
-        **({"endedTime": now} if status == ExecutionStatus.COMPLETED else {}),
-        **({"executionData": Json(execution_data)} if execution_data else {}),
-        **({"stats": Json(stats)} if stats else {}),
-    }  # type: ignore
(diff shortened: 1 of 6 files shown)


aarushik93 (approved):

autogpt_platform/backend/backend/data/execution.py
```diff
@@ -350,6 +357,23 @@ async def update_execution_status(
...
+    now = datetime.now(tz=timezone.utc)
+    return {
+        **({"executionStatus": status}),
+        **({"queuedTime": now} if status == ExecutionStatus.QUEUED else {}),
+        **({"startedTime": now} if status == ExecutionStatus.RUNNING else {}),
+        **({"endedTime": now} if status == ExecutionStatus.FAILED else {}),
+        **({"endedTime": now} if status == ExecutionStatus.COMPLETED else {}),
+        **({"executionData": Json(execution_data)} if execution_data else {}),
+        **({"stats": Json(stats)} if stats else {}),
+    }  # type: ignore
```
Pwuts: unchecked code
```suggestion
    update_data: AgentNodeExecutionUpdateInput = {"executionStatus": status}
    if status == ExecutionStatus.QUEUED:
        update_data["queuedTime"] = now
    elif status == ExecutionStatus.RUNNING:
        update_data["startedTime"] = now
    elif status in (ExecutionStatus.FAILED, ExecutionStatus.COMPLETED):
        update_data["endedTime"] = now
    if execution_data:
        update_data["executionData"] = Json(execution_data)
    if stats:
    	update_data["stats"] = Json(stats)

    return update_data
```

Pwuts (approved):

PR #9680, opened 2025-03-24
fix(platform/library): Fix UX for webhook-triggered runs

### LINKED ISSUE #9679: Fix crash of agent run details view for webhook-triggered runs
Webhook-triggered runs have a `payload` input, but the `payload` field isn't included in the input schema so it fails to render and crashes the page.

Pwuts:
- Resolves #9679

### Changes 🏗️

Frontend:
- Fix crash on `payload` graph input
- Fix crash on object type agent I/O values
- Hide "+ New run" if `graph.webhook_id` is set

Backend:
- Add computed field `webhook_id` to `GraphModel`
  - Add computed property `webhook_input_node` to `GraphModel`
- Refactor:
  - Move `Node.webhook_id` -> `NodeModel.webhook_id`
  - Move `NodeModel.block` -> `Node.block` (computed property)
  - Replace `get_block(node.block_id)` with `node.block` where sensible

#### For code changes:
- [x] I have made a test plan
  - [x] Create and run a simple graph
  - [x] Create a graph with a webhook trigger and ensure it works
  - [x] Check out the runs of a webhook-triggered graph and ensure the page works


AutoGPT-Agent:
This PR fails to meet requirements because: 1) The checklist is incomplete and not filled out at all 2) The test plan section is empty when it should be filled out for code changes 3) The changes description could be more detailed although it's somewhat understandable what's being changed. However, since this appears to be a minor bugfix with a clear scope and purpose, these issues are relatively minor.

autogpt_platform/backend/backend/data/graph.py
```diff
@@ -164,15 +164,30 @@ class BaseGraph(BaseDbModel):
...
+                [
+                    node.input_default
+                    for node in self.nodes
+                    if (b := get_block(node.block_id))
+                    and b.block_type == BlockType.INPUT
+                    and "name" in node.input_default
+                ]
+                if not webhook_input_block
+                else [
+                    {"name": "payload", "title": f"{webhook_input_block.name} payload"}
```
a-holm: The current logic completely replaces the schema derived from `AgentInputBlock` nodes if *any* webhook block (`WEBHOOK` or `WEBHOOK_MANUAL`) exists in the graph. Is this the intended behavior? It seems this could prevent defining graphs that can be triggered *both* manually (using inputs defined by `AgentInputBlock`s) *and* via a webhook (using the `payload`). If a webhook block is present, the schema will *only* show the `payload`, potentially hiding the necessary inputs for manual runs. Could we clarify the expected behavior for such mixed-input graphs?

Also, the PR checklist notes that testing hasn't been performed yet. Given the potential impact on the graph's input contract, adding tests covering different scenarios (no webhook, webhook only, webhook + input blocks) would be crucial before merging.

Pwuts: > It seems this could prevent defining graphs that can be triggered *both* manually (using inputs defined by `AgentInputBlock`s) *and* via a webhook (using the `payload`).

We currently explicitly disallow that, so yes, this is intended behavior.

a-holm: Alright :)

autogpt_platform/backend/backend/blocks/smart_decision_maker.py
```diff
@@ -264,9 +263,7 @@ def _create_block_function_signature(
         Raises:
```
ntindle, line 263: This needs updated i think

autogpt_platform/backend/backend/data/graph.py
```diff
@@ -88,8 +89,7 @@ def from_db(node: AgentNode, for_export: bool = False) -> "NodeModel":
         return obj
 
     def is_triggered_by_event_type(self, event_type: str) -> bool:
-        if not (block := get_block(self.block_id)):
-            raise ValueError(f"Block #{self.block_id} not found for node #{self.id}")
+        block = self.block
```
ntindle, line 92: this is unhinged syntax

ntindle (approved):

ntindle:
have not tested yet but code looks good *other than the unhinged as heck syntax*

autogpt_platform/backend/backend/blocks/smart_decision_maker.py
Pwuts: nope, technically still valid as that's what `Node.block` does

autogpt_platform/backend/backend/data/graph.py
Pwuts: please be specific

ntindle (dismissed):

ntindle:
image
we should consider disablign the run again or storing the payload somehow

autogpt_platform/backend/backend/data/graph.py
ntindle: block = self.block running code in the background subverts my expectations heavily that this could raise

Pwuts:
@ntindle yes, you're right. Lemme see if I can fix that real quick.

autogpt_platform/backend/backend/data/graph.py
Pwuts: If you think about the conditions in which this will raise, I think it's not that much of an issue. If this raises we have bigger problems.

Anyway, do you want me to turn this into a `@property`?

ntindle (approved):

Torantulino:
@Pwuts does this resolve #9752?

Pwuts:
@Torantulino no

PR #9695, opened 2025-03-25
feat(platform/library): Real-time execution updates

### LINKED ISSUE #8782: Implement (real-time) datasource for agent run statuses
* Part of #8780

Currently, the front end subscribes to all graph execution events and does aggregation to derive an execution status. We are now going to use this status in more places, so we should do the aggregation in the back end. The aggregation logic also needs updating because of some new added factors that contribute to the overall "agent status".

### Requirements

* Emits agent execution updates in real time
* Part of WebSockets API
* The same info needs to be available through the `GET /graphs/{graph_id}/runs` REST endpoint
* Events:
  * New agent run
  * Agent run status updated
    * Queued (can this be combined with "New agent run"?)
    * Running
    * Completed
    * Stopped
    * Failed
  * New agent output item
  * New error (or is this the same as "Failed"?)

Pwuts:
- Resolves #8782

### Changes 🏗️

- feat(frontend/library): Use WS subscription to get real-time execution updates
  - feat(backend/ws_api): Send `GraphExecutionUpdate` on all new agent I/O
    - Include agent I/O in `GraphExecutionUpdate` (by subclassing `GraphExecution`)
    - Add `IO_BLOCK_IDs` to `.blocks.io`
  - feat(backend/ws_api): Add `subscribe_graph_executions` method to WebSocket API

- feat(backend): Withhold `GraphExecution.node_executions` from requests by non-graph-owners
  - Split `GraphExecutionWithNodes` off of `GraphExecution`
    - Use `GraphExecution` as much as possible, as it's a much cheaper query than `GraphExecutionWithNodes`
  - refactor(frontend): Make `GraphExecution.node_executions` optional

- fix(frontend): Parse dates in responses of `/executions` and `/graphs/{graph_id}/executions`

- refactor(frontend/library): Move sorting logic for agent runs list from `AgentRunsPage` to `AgentRunsSelectorList`

- refactor(backend/ws_api): Clean up message handler implementations

- refactor(backend/tests): Use `.data.execution.get_graph_execution(..)` directly instead of `AgentServer.test_get_graph_run_results(..)`

Out-of-scope changes:
- refactor(backend): Remove unnecessary query include from `.data.graph.get_graph_metadata(..)`

Demo:

https://github.com/user-attachments/assets/8ea6225d-7334-49cb-a522-83f153d840da

#### For code changes:
- [x] I have made a test plan
  - Go to `/library/agents/[id]` for an agent with inputs and outputs
    - Draft and run a new run
      - [x] -> should appear in the list of runs at the top
      - [x] -> should be selected as soon as the request finishes
      - [x] -> new I/O should appear as it is generated
      - [x] -> status should be updated in real-time (both in list and in adjacent details view)
    - Click "Run again"
      - [x] -> should appear in the list of runs at the top
      - [x] -> should be selected as soon as the request finishes
      - [x] -> new I/O should appear as it is generated
      - [x] -> status should be updated in real-time (both in list and in adjacent details view)
    - Click "Open in builder" under "Agent actions"; run the agent from the builder
      - [x] -> should work the same as before
        - [x] -> node I/O should appear in real-time
        - [x] -> node execution statuses should update in real-time

### DIFF
--- a/autogpt_platform/backend/backend/blocks/io.py
+++ b/autogpt_platform/backend/backend/blocks/io.py
@@ -550,3 +550,17 @@ class AgentToggleInputBlock(AgentInputBlock):
                 ("result", False),
             ],
         )
+
+
+IO_BLOCK_IDs = [
+    AgentInputBlock().id,
+    AgentOutputBlock().id,
+    AgentShortTextInputBlock().id,
+    AgentLongTextInputBlock().id,
+    AgentNumberInputBlock().id,
+    AgentDateInputBlock().id,
+    AgentTimeInputBlock().id,
+    AgentFileInputBlock().id,
+    AgentDropdownInputBlock().id,
+    AgentToggleInputBlock().id,
+]
--- a/autogpt_platform/backend/backend/data/execution.py
+++ b/autogpt_platform/backend/backend/data/execution.py
@@ -292,8 +292,16 @@ async def get_graph_execution(
 async def get_graph_execution(
     user_id: str,
     execution_id: str,
-    include_node_executions: bool = False,
+    include_node_executions: Literal[False] = False,
 ) -> GraphExecution | None: ...
+
+
+@overload
+async def get_graph_execution(
+    user_id: str,
+    execution_id: str,
+    include_node_executions: bool = False,
+) -> GraphExecution | GraphExecutionWithNodes | None: ...
 
 
 async def get_graph_execution(
--- a/autogpt_platform/backend/backend/data/execution.py
+++ b/autogpt_platform/backend/backend/data/execution.py
@@ -38,8 +38,8 @@ from .block import BlockData, BlockInput, BlockType, CompletedBlockOutput, get_b
 from .db import BaseDbModel
 from .includes import (
     EXECUTION_RESULT_INCLUDE,
-    GRAPH_EXECUTION_INCLUDE_ALL,
-    GRAPH_EXECUTION_INCLUDE_IO_ONLY,
+    GRAPH_EXECUTION_INCLUDE,
+    GRAPH_EXECUTION_INCLUDE_WITH_NODES,
 )
 from .model import GraphExecutionStats, NodeExecutionStats
 from .queue import AsyncRedisEventBus, RedisEventBus
@@ -287,9 +287,9 @@ async def get_graph_execution(
     execution = await AgentGraphExecution.prisma().find_first(
         where={"id": execution_id, "isDeleted": False, "userId": user_id},
         include=(
-            GRAPH_EXECUTION_INCLUDE_ALL
+            GRAPH_EXECUTION_INCLUDE_WITH_NODES
             if include_node_executions
-            else GRAPH_EXECUTION_INCLUDE_IO_ONLY
+            else GRAPH_EXECUTION_INCLUDE
         ),
     )
     return (
@@ -337,7 +337,7 @@ async def create_graph_execution(
             "userId": user_id,
             "agentPresetId": preset_id,
         },
-        include=GRAPH_EXECUTION_INCLUDE_ALL,
+        include=GRAPH_EXECUTION_INCLUDE_WITH_NODES,
     )
 
     return GraphExecutionWithNodes.from_db(result)
@@ -438,7 +438,7 @@ async def update_graph_execution_start_time(graph_exec_id: str) -> GraphExecutio
             "executionStatus": ExecutionStatus.RUNNING,
             "startedAt": datetime.now(tz=timezone.utc),
         },
-        include=GRAPH_EXECUTION_INCLUDE_IO_ONLY,
+        include=GRAPH_EXECUTION_INCLUDE,
     )
     if not res:
         raise ValueError(f"Graph execution #{graph_exec_id} not found")
@@ -466,7 +466,7 @@ async def update_graph_execution_stats(
             "executionStatus": status,
             "stats": Json(data),
         },
-        include=GRAPH_EXECUTION_INCLUDE_IO_ONLY,
+        include=GRAPH_EXECUTION_INCLUDE,
     )
 
     return GraphExecution.from_db(res) if res else None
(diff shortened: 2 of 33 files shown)


autogpt_platform/backend/backend/blocks/io.py
```diff
@@ -550,3 +550,17 @@ def __init__(self):
...
         )
+
+
+IO_BLOCK_IDs = [
+    AgentInputBlock().id,
+    AgentOutputBlock().id,
+    AgentShortTextInputBlock().id,
+    AgentLongTextInputBlock().id,
+    AgentNumberInputBlock().id,
+    AgentDateInputBlock().id,
```
majdyz, line 561: You can get all blocks and filter by the block type

autogpt_platform/backend/backend/data/execution.py
```diff
@@ -250,12 +279,26 @@ async def get_graph_execution_meta(
...
+            if include_node_executions
+            else GRAPH_EXECUTION_INCLUDE_IO_ONLY
+        ),
+    )
+    return (
+        (
+            GraphExecutionWithNodes if include_node_executions else GraphExecution
+        ).from_db(execution)
+        if execution
+        else None
```
majdyz: Nit: can we just so normal if else here

autogpt_platform/backend/backend/data/execution.py
```diff
@@ -250,12 +279,26 @@ async def get_graph_execution_meta(
     return GraphExecutionMeta.from_db(execution) if execution else None
 
 
-async def get_graph_execution(user_id: str, execution_id: str) -> GraphExecution | None:
+async def get_graph_execution(
+    user_id: str,
+    execution_id: str,
+    include_node_executions: bool = False,
+) -> GraphExecution | GraphExecutionWithNodes | None:
```
majdyz, line 311: Isn't the the default behavior is including the node?

autogpt_platform/backend/backend/util/test.py
```diff
@@ -81,7 +81,7 @@ async def is_execution_completed():
             graph_exec = await AgentServer().test_get_graph_run_results(
                 graph_id, graph_exec_id, user_id
             )
-            return graph_exec.node_executions
+            return graph_exec.node_executions  # type: ignore
```
majdyz: What happened here?

autogpt_platform/backend/backend/data/execution.py
Pwuts: not sure what you mean?

autogpt_platform/backend/backend/data/execution.py
Pwuts: That is the current behavior, yes. But it doesn't have to be for the current usages of this function.

We'll want to exclude the node execution results once we close off access to closed-graph agent graphs, so this is also a step in that direction.

autogpt_platform/backend/backend/util/test.py
Pwuts: `graph_exec.node_executions` isn't set if `graph_exec.user_id != graph.user_id`. This isn't relevant for any (current) test case though so we can ignore.
(I've added a comment.)

I could rewrite `test_get_graph_run_results` so that it always returns the node executions regardless of ownership, but it didn't seem worth the effort at the time.

autogpt_platform/backend/backend/blocks/io.py
Pwuts: Downside of that is we have to load all block modules and initialize all 250 blocks instead of just these 10. This is much more efficient.

autogpt_platform/backend/backend/util/test.py
Pwuts: as it turns out, this did cause issues, so I rewrote the call chain

autogpt_platform/backend/backend/blocks/io.py
Pwuts: Let me know what you think. I have your proposed change in a stash locally so I can apply it if you want me to.

autogpt_platform/backend/backend/data/execution.py
majdyz: I see got it

autogpt_platform/backend/backend/data/execution.py
majdyz:
```
if not execution:
    return None
elif include_node_execution:
    return GraphExecutionWithNodes.from_db(execution)
else:
    return GraphExecution.from_db(execution)
```

It's just more obvious to read than a nested for a reviewer, lol

autogpt_platform/backend/backend/executor/manager.py
```diff
@@ -175,7 +175,15 @@ def update_execution(status: ExecutionStatus) -> NodeExecutionResult:
     input_data, error = validate_exec(node, data.data, resolve_input=False)
     if input_data is None:
         log_metadata.error(f"Skip execution, input validation error: {error}")
-        db_client.upsert_execution_output(node_exec_id, "error", error)
+        _register_node_execution_output(
```
majdyz: You can create a local function like `update_execution` above to avoid passing a large number of variables into your provide method trice. Only output_name and output_data changed; even the status for update_execution can be inferred. And the `update_execution` can be just inlined in this new local function.

autogpt_platform/backend/backend/executor/manager.py
```diff
@@ -238,7 +254,15 @@ def update_execution(status: ExecutionStatus) -> NodeExecutionResult:
 
     except Exception as e:
         error_msg = str(e)
-        db_client.upsert_execution_output(node_exec_id, "error", error_msg)
```
majdyz, line 241: I know it's not your change, but would you mind removing `# Update execution status and spend credits` comment above in this PR? :)))

autogpt_platform/backend/backend/blocks/io.py
majdyz: I think loading them all once per startup is fine.

Ironically, all the usage of `get_blocks` initiates the block immediately right after the call, which basically we can just cache the initialized block instead of the class, where we can do later if it actually matters.

majdyz (dismissed):

autogpt_platform/backend/backend/blocks/io.py
Pwuts: I've [tried](https://github.com/Significant-Gravitas/AutoGPT/pull/9695/commits/71f0cf9f0000b5c714e48e3a912ac1ec8ad79939) but it gives a lot of circular import issues

PR #9700, opened 2025-03-26
feat(frontend): Implement UI for Agent Input subtypes

majdyz:
- Follow-up to #9657

image

image

### Changes 🏗️

* Implement the input UI for Agent Input subtypes.
* Refactor node-input-component, extra out data type decision logic, share it with runner/library input.
* Add `format` field for short-text, long-text, and mediafile type.
* Unify UI data type enum.

Out of scope:
- Styling for these inputs.

#### For code changes:
- [x] I have made a test plan
  - [x] Use all the available agent input subtypes in an agent and run it


autogpt_platform/backend/backend/blocks/io.py
```diff
@@ -56,6 +53,12 @@ class Input(BlockSchema):
             advanced=True,
         )
 
+        def generate_schema(self):
+            schema = self.get_field_schema("value")
+            if possible_values := self.placeholder_values:
+                schema["enum"] = possible_values
+            return schema
```
Pwuts, line 60: This looks like it belongs on `AgentDropdownInputBlock`?

autogpt_platform/frontend/src/app/globals.css
```diff
@@ -74,7 +74,7 @@
 
 @layer components {
   .agpt-border-input {
-    @apply border border-input focus-visible:border-gray-400 focus-visible:outline-none;
+    @apply m-0.5 border border-input focus-visible:border-gray-400 focus-visible:outline-none focus-visible:ring-1 focus-visible:ring-gray-400;
```
Pwuts, line 77: Does this really work in the layout? This utility is for *border styling*, margin is a layout thing.

autogpt_platform/frontend/src/components/NodeInputComponent.tsx
Pwuts, line 1: nit: undo this rename

autogpt_platform/frontend/src/lib/autogpt-server-api/types.ts
```diff
@@ -65,6 +65,167 @@ export type BlockIOSimpleTypeSubSchema =
...
+  SHORT_TEXT = "short-text",
+  LONG_TEXT = "long-text",
+  NUMBER = "number",
+  DATE = "date",
+  TIME = "time",
+  DATE_TIME = "date-time",
+  FILE = "file",
+  SELECT = "select",
+  MULTI_SELECT = "multi-select",
+  TOGGLE = "toggle",
```
Pwuts: boolean?

autogpt_platform/frontend/src/lib/autogpt-server-api/types.ts
```diff
@@ -65,6 +65,167 @@ export type BlockIOSimpleTypeSubSchema =
...
+          return DataType.DATE_TIME;
+        case "file":
+          return DataType.FILE;
+        case "short-text":
+          return DataType.SHORT_TEXT;
+        case "long-text":
+          return DataType.LONG_TEXT;
+        default:
+          break;
+      }
```
Pwuts: I'm pretty sure TS has a better way to do this 😅

autogpt_platform/frontend/src/lib/autogpt-server-api/types.ts
```diff
@@ -65,6 +65,167 @@ export type BlockIOSimpleTypeSubSchema =
...
+  MULTI_SELECT = "multi-select",
+  TOGGLE = "toggle",
+  CREDENTIALS = "credentials",
+  OBJECT = "object",
+  KEY_VALUE = "key-value",
+  ARRAY = "array",
+}
+
+export function determineDataType(schema: BlockIOSubSchema): DataType {
+  function handleStringSchema(strSchema: BlockIOStringSubSchema): DataType {
```
Pwuts: this doesn't need to be a nested function

autogpt_platform/frontend/src/lib/autogpt-server-api/types.ts
```diff
@@ -65,6 +65,167 @@ export type BlockIOSimpleTypeSubSchema =
...
+          break;
+      }
+    }
+    if (strSchema.enum) return DataType.SELECT;
+    if (strSchema.maxLength && strSchema.maxLength > 200)
+      return DataType.LONG_TEXT;
+    return DataType.SHORT_TEXT;
+  }
+
+  function handleSingleTypeSchema(subSchema: BlockIOSubSchema): DataType {
```
Pwuts: this doesn't need to be a nested function

autogpt_platform/frontend/src/lib/autogpt-server-api/types.ts
```diff
@@ -65,6 +65,167 @@ export type BlockIOSimpleTypeSubSchema =
...
+  SELECT = "select",
+  MULTI_SELECT = "multi-select",
+  TOGGLE = "toggle",
+  CREDENTIALS = "credentials",
+  OBJECT = "object",
+  KEY_VALUE = "key-value",
+  ARRAY = "array",
+}
+
+export function determineDataType(schema: BlockIOSubSchema): DataType {
```
Pwuts: could you put these helper functions at the bottom of the file or in a different file?

autogpt_platform/frontend/src/hooks/useAgentGraph.tsx
```diff
@@ -411,7 +413,9 @@ export default function useAgentGraph(
           ) {
             return;
           }
-          console.warn("Error", error);
+          console.warn(
+            `Error in ${node.data.blockType}: ${error}. Data: ${inputData}, Schema: ${node.data.inputSchema}`,
```
Pwuts:
```suggestion
            `Error in ${node.data.blockType}: ${error}`, {data: inputData, schema: node.data.inputSchema}
```

autogpt_platform/frontend/src/components/TypeBasedInput.tsx
Pwuts, line 1: nit: `type-based-input.tsx`

autogpt_platform/frontend/src/components/TypeBasedInput.tsx
```diff
@@ -0,0 +1,428 @@
...
+ * the correct UI component.
+ */
+export const TypeBasedInput: FC<TypeBasedInputProps> = ({
+  schema,
+  value,
+  placeholder,
+  onChange,
+}) => (
+  <div className="no-drag relative flex">
+    {_TypeBasedInput({ schema, value, onChange, placeholder })}
```
Pwuts:
```suggestion
    <_TypeBasedInput schema={schema} value={value} onChange={onChange} placeholder={placeholder} />
```

autogpt_platform/frontend/src/components/TypeBasedInput.tsx
```diff
@@ -0,0 +1,428 @@
...
+}: React.InputHTMLAttributes<HTMLInputElement>) {
+  return <DefaultInput {...props} className={cn(inputClasses, className)} />;
+};
+
+/**
+ * A generic, data-type-based input component that uses Shadcn UI.
+ * It inspects the schema via `determineDataType` and renders
+ * the correct UI component.
+ */
+export const TypeBasedInput: FC<TypeBasedInputProps> = ({
```
Pwuts: Why this separate wrapper component? This can be merged with `_TypeBasedInput`

autogpt_platform/frontend/src/components/TypeBasedInput.tsx
```diff
@@ -0,0 +1,428 @@
+import React, { FC } from "react";
+import { cn } from "@/lib/utils";
+import { format } from "date-fns";
+import { CalendarIcon } from "lucide-react";
+import { Cross2Icon, FileTextIcon } from "@radix-ui/react-icons";
+
+import { Input as DefaultInput } from "@/components/ui/input";
```
Pwuts:
```suggestion
import { Input as BaseInput } from "@/components/ui/input";
```

autogpt_platform/frontend/src/components/TypeBasedInput.tsx
```diff
@@ -0,0 +1,428 @@
...
+ * can do something like setInputValues(e.target.value).
+ */
+export interface TypeBasedInputProps {
+  schema: BlockIOSubSchema;
+  value?: any;
+  placeholder?: string;
+  onChange: (value: any) => void;
+}
+
+const inputClasses = "min-h-11 rounded-[55px] border px-4 py-2.5";
```
Pwuts: `rounded-[55px]`? Shouldn't this just be `rounded-full`?

Pwuts: Every single custom pixel value is tech debt.

Sidenote: defining custom dimensions in `rem` instead of `px` would be slightly better.

Pwuts (commented):
I'm really excited for this! It's a big UX improvement :)

autogpt_platform/backend/backend/blocks/io.py
majdyz: `placeholder_values` is part of the generic AgentInputBlock too

autogpt_platform/frontend/src/app/globals.css
majdyz: yes, the margin is indeed required for border styling;  the border highlight is thick enough that the border will overflow without the thin margin.

autogpt_platform/frontend/src/components/NodeInputComponent.tsx
majdyz: The other files here use this same convention, which one is correct ?

autogpt_platform/frontend/src/lib/autogpt-server-api/types.ts
majdyz: updated

autogpt_platform/frontend/src/lib/autogpt-server-api/types.ts
majdyz: like by using an object ? Do you think it's a better way? Seems like a switch case is built literally for this purpose.

autogpt_platform/frontend/src/lib/autogpt-server-api/types.ts
majdyz: moved outside

autogpt_platform/frontend/src/lib/autogpt-server-api/types.ts
majdyz: moved

autogpt_platform/frontend/src/lib/autogpt-server-api/types.ts
majdyz: moved to the bottom

autogpt_platform/frontend/src/components/TypeBasedInput.tsx
majdyz: what is the the right naming convention here, kebab case seems to be minority on the file

autogpt_platform/frontend/src/components/TypeBasedInput.tsx
majdyz: how ?

autogpt_platform/frontend/src/components/TypeBasedInput.tsx
majdyz: rounded full is a circle, while rounded 55px is 3.4375rem

Pwuts: > rounded full is a circle

Not exactly: `rounded-full` means the border radius is set to 50% the short dimension of the rectangle, so the short sides get fully rounded while the long sides don't get collapsed to a circle (square dimensions).

Is this `inputClasses` used for any components/elements which are rectangular but of which the ends are not fully rounded? All the buttons, inputs etc. seem to be, so then `rounded-full` is fine.

Also, `3.5rem` = `56px`. In case the dimension really has to be specified explicitly: how important is that one pixel?

autogpt_platform/frontend/src/app/globals.css
Pwuts: Alright, fair point. But adding a margin to these elements by default changes their alignment and positioning wherever they are used. I think having consistent and predictable alignment/positioning behavior across all elements is important, and a highlight slightly overflowing the element is fine.

autogpt_platform/frontend/src/components/NodeInputComponent.tsx
Pwuts: That's a fair question. This PascalCasing of some component files is a legacy pattern that we should get rid of imo. React component files are TS(X) modules for which kebab-case is the standard.

autogpt_platform/frontend/src/lib/autogpt-server-api/types.ts
Pwuts: Yeah I'd either use the two-way map properties of `DataType` itself:
```ts
if (strSchema.format in DataType) return strSchema.format as DataType
```
or make an explicit map of string value to `DataType`.
A `switch` like this is *suitable* but since all it does is map a set of values to a different set of values, an object `value2 = { ...mapping }[value1]` seems like the better (for DX) choice here.

autogpt_platform/frontend/src/components/TypeBasedInput.tsx
Pwuts: https://github.com/Significant-Gravitas/AutoGPT/pull/9700#discussion_r2020916545

autogpt_platform/frontend/src/components/TypeBasedInput.tsx
Pwuts:
```tsx
let innerInputElement: React.ReactElement;

switch (dataType) {
  case DataType.NUMBER:
    innerInputElement = (
      <Input
        type="number"
        value={value ?? ""}
        placeholder={placeholder}
        onChange={(e) => onChange(Number(e.target.value))}
      />
    );
    break;

// ...

return <div className="no-drag relative flex">
  {innerInputElement}
</div>
```

autogpt_platform/frontend/src/components/TypeBasedInput.tsx
majdyz: Okay updated to rounded-full

autogpt_platform/frontend/src/components/TypeBasedInput.tsx
majdyz: I feel like we are moving back to the squire one where we render the content using a non-component function, the only difference is we are inlining it as a local variable. Updated though,

autogpt_platform/frontend/src/components/TypeBasedInput.tsx
majdyz: Updated

autogpt_platform/frontend/src/lib/autogpt-server-api/types.ts
majdyz: > if (strSchema.format in DataType) return strSchema.format as DataType

Unfortunately, we can't do this because the pedantic format uses the kebab case, while the enum uses the snake case.

> A switch like this is suitable but since all it does is map a set of values to a different set of values, an object value2 = { ...mapping }[value1] seems like the better (for DX) choice here.

Have you checked the rest of the logic below? also this is just a helper of another function, so this matching is just a small part of the bigger logic. Though I updated it using a map now.

autogpt_platform/frontend/src/components/NodeInputComponent.tsx
majdyz: Updated

autogpt_platform/frontend/src/app/globals.css
majdyz: The added margin should not affect a component where a margin is already defined.
And on no margin, the difference should not be too noticeable, plus a slightly overflowing margin is more problematic than a slight positioning. It makes the input look broken.

ntindle (approved):

PR #9704, opened 2025-03-27
feat: different signup error message

ntindle:
We keep showing local users error messages that are just not relevant

### Changes 🏗️
Swaps the error messaging logic to be dependent on the behavior of the specific platform they are on

#### For code changes:
- [x] I have made a test plan
  
  - [x] Test with auth container down (simulates incorrect setup) and validate that error message is shown
  - [x] Try normal path


autogpt_platform/frontend/src/components/auth/AuthFeedback.tsx
```diff
@@ -1,15 +1,117 @@
+import { AlertCircle, CheckCircle } from "lucide-react";
+import { Card, CardContent } from "@/components/ui/card";
+import { HelpItem } from "@/components/auth/help-item";
+import Link from "next/link";
+import { BehaveAs } from "@/lib/utils";
+
 interface Props {
   message?: string | null;
   isError?: boolean;
+  showErrorPrompt?: boolean;
```
majdyz, line 10: is there a case where we set `showErrorPrompt` = true when `isError` = False or vice-versa ?

majdyz (approved):

autogpt_platform/frontend/src/components/auth/AuthFeedback.tsx
ntindle: maybe? they were seperate before i belive

PR #9711, opened 2025-03-28
feat(frontend): Add advanced block search with relevance ranking

### LINKED ISSUE #9425: Upgrade block search functionality
Currently search only checks block name and it must contain search phrase exactly; for example searching for "AI Sum" won't include "AI Text Summarizer" but it should.

Search should be more forgiving, ideas:

1. Use string similarity algorithm instead of checking exactly (e.g. The Levenshtein Distance algorithm)
2. Disregard word search order
3. Search in block description (as a lower priority)

Abhi1992002:
- fix #9425 

- Enhancing the functionality of searching blocks on the build page

Currently, it only performs exact matching on the block name and description. I added a scoring mechanism for searching.

- The scoring algorithm works as follows:
     - Returns 1 if no query (all blocks match equally)
     - Normalized query for case-insensitive matching
     - Returns 3 for exact substring matches in block name (highest priority)
     - Returns 2 when all query words appear in the block name (regardless of order)
     - Returns 1.X for blocks with names similar to query using Jaro-Winkler distance (X is similarity score)
     - Returns 0.5 when all query words appear in the block description (lowest priority)
     - Returns 0 for no match

Higher scores will appear first in search results.

> I have used an external library for Jaro-Winkler distance - [link](https://www.npmjs.com/package/jaro-winkler)

Before
Screenshot 2025-03-28 at 12 09 24 PM

After
Screenshot 2025-03-28 at 12 09 17 PM


autogpt_platform/frontend/src/components/edit/control/BlocksControl.tsx
```diff
@@ -89,21 +89,102 @@ export const BlocksControl: React.FC<BlocksControlProps> = ({
         }) satisfies Block,
     );
 
+    // Levenshtein distance Algorithm
```
majdyz: We should move this to a helper file outside of this file.

Also, have you considered other drop-in libraries rather than hand-rolling an edit distance yourself?
There are a few other better algorithms you can use for this.

Abhi1992002: @majdyz I used the Levenshtein distance algorithm because it was mentioned in the issue. Could you recommend some alternatives that I could use?

majdyz: Got it, we should not put this logic in this file though, let's place it on the other helper files (or create a new one if it doesn't exist).

The user won't likely type the full text (where Levenshtein score makes a difference), meanwhile the user will likely just type its prefix, where Jaro-Winkler would be a better and much faster considering we will keep increasing more blocks. You can use a library instead of writing your own.

majdyz (approved):
Thank you! This is a really cool feature.

PR #9729, opened 2025-03-31
fix(ci/backend): Use Poetry version from lockfile

Pwuts:
Currently, our CI always uses the latest version of Poetry. This causes issues with the lockfile check whenever a new Poetry version is released, especially if that new version has different lockfile generation behavior.

This new mechanism determines the Poetry version to use as follows:
- Get Poetry version from backend/poetry.lock in the current branch
- Get Poetry version from backend/poetry.lock on the base branch
- Use the newest version out of the two found versions

This way, we don't automatically update to new Poetry versions, but it is still possible to update to newer versions through pull requests.


ntindle (approved):

ntindle:
```
0s
Run # Extract Poetry version from backend/poetry.lock
Found Poetry version 2.[1](https://github.com/Significant-Gravitas/AutoGPT/actions/runs/14176691607/job/39713188050#step:8:1).2 in backend/poetry.lock
fatal: invalid object name 'origin/'.
```

Bentlybro (approved):

Bentlybro (dismissed):

Bentlybro (approved):

Bentlybro (approved):

Bentlybro (approved):

PR #9734, opened 2025-04-01
fix(backend): handle notification service errors more elegently

ntindle:
We have logged 272k timeout errors in the past week from the event loop. Don't raise those as errors. 

Also along the way for diagnosing this we found that some items were inserted into batches with incomplete datasets so handle that too.

### Changes 🏗️
- Handle timeout errors explicitly 
- Add better messaging for other error types
- Add filtering for queueing bad mezsaging
- add filtering for reading bad batches

#### For code changes:
- [x] I have made a test plan
  
  - [x] Pull dev db
  - [x] Test new code to check stability + error reduction


autogpt_platform/backend/backend/data/notifications.py
```diff
@@ -398,10 +398,15 @@ async def create_or_add_to_user_notification_batch(
...
         )
+        if not notification_data.data:
+            raise ValueError("Notification data must be provided")
 
         # Serialize the data
         json_data: Json = Json(notification_data.data.model_dump())
 
+        if not json_data.data:
+            raise ValueError("Notification data must be provided")
+
```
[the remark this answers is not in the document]
majdyz: This seems to be a valid comment, Why do we do it twice and is the second check even possible ?

majdyz (approved):

autogpt_platform/backend/backend/notifications/notifications.py
```diff
@@ -245,20 +245,26 @@ def process_existing_batches(self, notification_types: list[NotificationType]):
...
+                                            "user_id": batch.user_id,
+                                            "type": db_event.type,
+                                            "data": db_event.data,
+                                            "created_at": db_event.created_at,
+                                        }
+                                    )
+                                )
+                            except Exception as e:
+                                logger.exception(
+                                    f"Error parsing notification event: {e=}, {db_event=}"
```
majdyz, line 265: Exception will print the stacktrace, maybe you are looking for logger.error?
Also should we group the log into a single entry?
Imaging 200k messages being faulty, idk

majdyz (dismissed):

autogpt_platform/backend/backend/notifications/notifications.py
ntindle: ~~Exception is goal here, I would end the event loop if it made sense but instead we stack trace it as if it's the outer loop~~

ntindle: ~~Can you clarify grouping comment?~~

ntindle: Oops wrong segment. Error yes good

autogpt_platform/backend/backend/notifications/notifications.py
```diff
@@ -676,7 +684,7 @@ def _run_queue(
                 self.run_and_wait(message.reject(requeue=False))
             else:
                 logger.error(
```
ntindle: Make this exception

autogpt_platform/backend/backend/notifications/notifications.py
ntindle: We shouldn't have this error in prod as of yet, it's only in dev

majdyz (approved):

PR #9742, opened 2025-04-02
feat(backend, libs): Tell uvicorn to use our logger + always log to stdout+stderr

ntindle:
Uvicorn and our logs were ending up in different places, this pr enures uvicorn using our logging config, not their own.

### Changes 🏗️
- Clears uvicorn's loggers for rest, ws
- always log to stdout,stderr and additionally log to gcp is appropriate

#### For code changes:
- [x] I have made a test plan
  
  - [x] Test all possible variants of the log cloud vs not and ensure that uvicorn logs show up in the same place that rest of the system logs do for all


autogpt_platform/backend/backend/server/rest_api.py
```diff
@@ -137,10 +138,15 @@ def run(self):
             allow_methods=["*"],  # Allows all methods
             allow_headers=["*"],  # Allows all headers
         )
+        log_config = dict(uvicorn.config.LOGGING_CONFIG)
```
majdyz: Can't we move this replicated logic to configure_logging or other shared helper?
This kind of hinting that we also need this on all services/processes. we have notification service, scheduler, database manager, execution manager, but I guess this might have been somehow covered by AppService?

autogpt_platform/autogpt_libs/autogpt_libs/logging/config.py
```diff
@@ -97,26 +114,7 @@ def configure_logging(force_cloud_logging: bool = False) -> None:
             transport=SyncTransport,
         )
         cloud_handler.setLevel(config.level)
-        cloud_handler.setFormatter(StructuredLoggingFormatter())
```
Pwuts, line 100: ????????

autogpt_platform/autogpt_libs/autogpt_libs/logging/formatters.py
Pwuts, line 1: ???????

autogpt_platform/backend/backend/server/rest_api.py
```diff
@@ -137,10 +138,15 @@ def run(self):
...
+        log_config["loggers"]["uvicorn"] = {"handlers": []}
+        log_config["loggers"]["uvicorn.error"] = {"handlers": []}
+        log_config["loggers"]["uvicorn.access"] = {"handlers": []}
         uvicorn.run(
             server_app,
             host=backend.util.settings.Config().agent_api_host,
             port=backend.util.settings.Config().agent_api_port,
+            log_config=log_config,
```
Pwuts: IIRC you can achieve all of this with just `log_config=None`. I did that a while ago and it worked; I'm not sure when it was unfixed / where that fixed code went.
```suggestion
        uvicorn.run(
            server_app,
            host=backend.util.settings.Config().agent_api_host,
            port=backend.util.settings.Config().agent_api_port,
            log_config=None,
```

autogpt_platform/backend/backend/server/ws_api.py
```diff
@@ -3,7 +3,9 @@
 from contextlib import asynccontextmanager
 from typing import Protocol
 
+from backend.util.logging import configure_logging
```
Pwuts: unused import?

autogpt_platform/backend/backend/server/rest_api.py
ntindle: uvicorn uses its own logging config by default, the rest of ours don't suggest one and inherit the parent logger

autogpt_platform/backend/backend/server/rest_api.py
ntindle: it was removed in this pr 
https://github.com/Significant-Gravitas/AutoGPT/pull/8519/files

ntindle: those aren't the same behavior 

Setting to None removes the logging config and depends on yours entirely

I want uvicorn's configs but my handlers for its sinks. Below is the default, I'm setting each of the sinks to use our handlers

```py
LOGGING_CONFIG: dict[str, Any] = {
    "version": 1,
    "disable_existing_loggers": False,
    "formatters": {
        "default": {
            "()": "uvicorn.logging.DefaultFormatter",
            "fmt": "%(levelprefix)s %(message)s",
            "use_colors": None,
        },
        "access": {
            "()": "uvicorn.logging.AccessFormatter",
            "fmt": '%(levelprefix)s %(client_addr)s - "%(request_line)s" %(status_code)s',  # noqa: E501
        },
    },
    "handlers": {
        "default": {
            "formatter": "default",
            "class": "logging.StreamHandler",
            "stream": "ext://sys.stderr",
        },
        "access": {
            "formatter": "access",
            "class": "logging.StreamHandler",
            "stream": "ext://sys.stdout",
        },
    },
    "loggers": {
        "uvicorn": {"handlers": ["default"], "level": "INFO", "propagate": False},
        "uvicorn.error": {"level": "INFO"},
        "uvicorn.access": {"handlers": ["access"], "level": "INFO", "propagate": False},
    },
}
```

autogpt_platform/autogpt_libs/autogpt_libs/logging/config.py
ntindle: this was double wrapping messages. GCP logger has this included and our formatter was breaking the messages so they were unreadable in gcp

autogpt_platform/autogpt_libs/autogpt_libs/logging/formatters.py
ntindle: See the structured log handler import here? 

what we are doing is rebuilding internal structures in that existing handler

```py

class CloudLoggingHandler(logging.StreamHandler):
    """Handler that directly makes Cloud Logging API calls.

    This is a Python standard ``logging`` handler using that can be used to
    route Python standard logging messages directly to the Stackdriver
    Logging API.

    This handler is used when not in GAE or GKE environment.

    This handler supports both an asynchronous and synchronous transport.

    Example:

    .. code-block:: python

        import logging
        import google.cloud.logging
        from google.cloud.logging_v2.handlers import CloudLoggingHandler

        client = google.cloud.logging.Client()
        handler = CloudLoggingHandler(client)

        cloud_logger = logging.getLogger('cloudLogger')
        cloud_logger.setLevel(logging.INFO)
        cloud_logger.addHandler(handler)

        cloud_logger.error('bad news')  # API call
    """

    def __init__(
        self,
        client,
        *,
        name: str = DEFAULT_LOGGER_NAME,
        transport: Type[Transport] = BackgroundThreadTransport,
        resource: Resource = None,
        labels: Optional[dict] = None,
        stream: Optional[IO] = None,
        **kwargs,
    ):
        """
        Args:
            client (~logging_v2.client.Client):
                The authenticated Google Cloud Logging client for this
                handler to use.
            name (str): the name of the custom log in Cloud Logging.
                Defaults to 'python'. The name of the Python logger will be represented
                in the ``python_logger`` field.
            transport (~logging_v2.transports.Transport):
                Class for creating new transport objects. It should
                extend from the base :class:`.Transport` type and
                implement :meth`.Transport.send`. Defaults to
                :class:`.BackgroundThreadTransport`. The other
                option is :class:`.SyncTransport`.
            resource (~logging_v2.resource.Resource):
                Resource for this Handler. If not given, will be inferred from the environment.
            labels (Optional[dict]): Additional labels to attach to logs.
            stream (Optional[IO]): Stream to be used by the handler.
        """
        super(CloudLoggingHandler, self).__init__(stream)
        if not resource:
            # infer the correct monitored resource from the local environment
            resource = detect_resource(client.project)
        self.name = name
        self.client = client
        self.transport = transport(client, name, resource=resource)
        self.project_id = client.project
        self.resource = resource
        self.labels = labels
        # add extra keys to log record
        log_filter = CloudLoggingFilter(project=self.project_id, default_labels=labels)
        self.addFilter(log_filter)
```

autogpt_platform/backend/backend/server/rest_api.py
ntindle: Replicated logic moved

autogpt_platform/backend/backend/server/rest_api.py
```diff
@@ -137,10 +139,12 @@ def run(self):
             allow_methods=["*"],  # Allows all methods
             allow_headers=["*"],  # Allows all headers
         )
+        log_config = generate_uvicorn_config()
         uvicorn.run(
             server_app,
             host=backend.util.settings.Config().agent_api_host,
             port=backend.util.settings.Config().agent_api_port,
+            log_config=log_config,
```
majdyz: nit: why the need of local variable, we can assign it directly to the log_config param

majdyz (approved):

autogpt_platform/backend/backend/server/rest_api.py
Pwuts: Don't we also want uvicorn to use the same logging format as the rest of the app?

Pwuts (dismissed):

PR #9750, opened 2025-04-03
refactor(backend): Un-share resource initializations from AppService + Remove Pyro

majdyz:
This is a prerequisite infra change for https://github.com/Significant-Gravitas/AutoGPT/issues/9714.

We will need a service where we can maintain our own client (db, redis, rabbitmq, be it async/sync) and configure our own cadence of initialization and cleanup. 

While refactoring the service.py, an option to use Pyro as an RPC protocol is also removed.

### Changes 🏗️

* Decouple resource initialization and cleanup from the parent AppService logic.
* Removed Pyro.

#### For code changes:
- [x] I have made a test plan
  
  - [x] CI


ntindle:
I'm not sure of this approach. My worry is that people may not realize they are responsible for cleaning this all up so maybe we should have a finalizer that kills everything registered for the service?

ntindle:
Can we raise if cleanup isn’t implemented?

majdyz:
@ntindle done

ntindle (approved):

PR #9751, opened 2025-04-03
feat(frontend,backend): user spending admin dashboard

ntindle:
We need a way to refund people who spend money on agents wihout making manual db actions

### Changes 🏗️
- Adds a bunch for refunding users
- Adds reasons and admin id for actions
- Add admin to db manager
- Add UI for this for the admin panel
- Clean up pagination controls

#### For code changes:
- [x] I have made a test plan
  
  - [x] Test by importing dev db as baseline
  - [x] Add transactions on top for "refund", and make sure all existing transactions work


autogpt_platform/backend/backend/data/credit.py
```diff
@@ -590,6 +593,7 @@ async def _top_up_credits(
         amount: int,
         key: str | None = None,
         ceiling_balance: int | None = None,
+        is_auto_top_up: bool = False,
```
majdyz: https://read.engineerscodex.com/p/the-boolean-trap

For this case, please just pass the metadata directly so others can add more information.

autogpt_platform/backend/backend/data/credit.py
```diff
@@ -956,3 +966,81 @@ async def get_auto_top_up(user_id: str) -> AutoTopUpConfig:
...
+    total_pages = (total + page_size - 1) // page_size
+
+    history = []
+    for tx in transactions:
+        admin_id = ""
+        admin_email = ""
+        reason = ""
+
+        # this is actually a dict at runtime??????????
+        metadata: dict = tx.metadata or {}  # type: ignore
```
majdyz: It probably is your code-editor issue? pretty sure you don't need `type: ignore`, or if we really can't do it, we can just use cast.

autogpt_platform/backend/backend/data/credit.py
```diff
@@ -956,3 +966,81 @@ async def get_auto_top_up(user_id: str) -> AutoTopUpConfig:
...
+
+    if page < 1 or page_size < 1:
+        raise ValueError("Invalid pagination input")
+
+    where_clause: CreditTransactionWhereInput = {}
+    if transaction_filter:
+        where_clause["type"] = transaction_filter
+    if search:
+        where_clause["OR"] = [
+            {"userId": {"contains": search, "mode": "insensitive"}},
```
majdyz, line 1043: This is not going to work, `CreditTransaction` is huuge table. We need to find another way to do it.
This will scan an entire table on each query.

We should just re-use get_transaction_history, get the userId before hand by searching different table.

autogpt_platform/frontend/src/lib/autogpt-server-api/types.ts
```diff
@@ -879,6 +879,36 @@ export type AdminPendingSubmissionsRequest = {
...
 
+export enum CreditTransactionType {
+  TOP_UP = "TOP_UP",
+  USAGE = "USAGE",
+  GRANT = "GRANT",
+  REFUND = "REFUND",
+  CARD_CHECK = "CARD_CHECK",
+}
+
+export type UserBalanceHistory = {
```
majdyz: Can we unify TransactionHistory instead? they should be care about the same information.

majdyz: Can we unify TransactionHistory instead? they should be care about the same information.

autogpt_platform/frontend/src/components/admin/spending/search-filter-form.tsx
```diff
@@ -0,0 +1,105 @@
...
+          if (value === "ALL") {
+            params.delete("status");
+          } else {
+            params.set("status", value);
+          }
+          params.set("page", "1");
+          router.push(`${pathname}?${params.toString()}`);
+        }}
+      >
+        <SelectTrigger className="w-[180px]">
```
majdyz: Is this fixed width what we actually want? Can we at least use rem ?

autogpt_platform/frontend/src/components/admin/spending/add-credit-button.tsx
```diff
@@ -0,0 +1,146 @@
...
+                    min="0"
+                    className="rounded-l-none"
+                    value={dollarAmount}
+                    onChange={(e) => setDollarAmount(e.target.value)}
+                    placeholder="0.00"
+                  />
+                </div>
+                <p className="text-xs text-gray-500">
+                  This will add {calculateCredits(dollarAmount)} credits (1
+                  credit = 1¢)
```
majdyz: There is `formatCredits` in useCredits

majdyz: Credits is no longer a thing, we don't show number of credits anywhere, and there is only dollar amount.

majdyz (changes requested):

autogpt_platform/backend/backend/data/credit.py
ntindle: DAMN IT I FELL FOR IT. I JUST CAUGHT ONE OF THESE FOR KZCE

autogpt_platform/backend/backend/data/credit.py
ntindle: To be honest, i messed with this for like 20 mins the first time trying to get the type checker to like it, but here's the problem 
```
# inherit from _PydanticJson so that pydantic will automatically
# transform the json string into python objects.
class Json(BaseJson):
    data: Serializable

    def __init__(self, data: Serializable) -> None:
        self.data = data
        super().__init__()

    @classmethod
    def keys(cls, **data: Serializable) -> Json:
        return cls(data)

    if TYPE_CHECKING:
        # Fields that are of the `Json` type are automatically
        # de-serialized from json to the corresponding python type
        # when the model is created, e.g.
        #
        # User(json_obj='{"foo": null}') -> User(json_obj={'foo': None})
        #
        # As we don't know what the type will actually be at runtime
        # we add methods here for convenience so that naive access
        # to the field is still allowed, e.g.
        #
        # user.json_obj['foo']
        # user.json_obj[1]
        # user.json_obj[1:5]
        #
        # It should be noted that users will still have
        # to validate / cast fields to the type they are expecting
        # for any strict type binding or nested index calls to work, e.g.
        #
        # isinstance(user.json_obj, dict)
        # cast(Dict[str, Any], user.json_obj)
        # prisma.validate(ExpectedType, user.json_obj)  # NOTE: not implemented yet
        @overload  # type: ignore
        def __getitem__(self, i: slice) -> List[Serializable]: ...

        @overload
        def __getitem__(self, i: _JsonKeys) -> Serializable: ...

        @override
        def __getitem__(self, i: Union[_JsonKeys, slice]) -> Serializable:  # pyright: ignore[reportIncompatibleMethodOverride]
            ...
```

ntindle: BaseJson is string if typechecking otherwise Pydantic's Json

autogpt_platform/frontend/src/lib/autogpt-server-api/types.ts
ntindle: this feels like it seems true but isn't quite. I've unified on it for now

image

autogpt_platform/backend/backend/data/credit.py
ntindle: that fundamentally changes how the feature works or takes one large query and turns it into hundreds of smaller queries. IMO we should keep it as is until it becomes problematic since its an admin page

cc @itsababseh

autogpt_platform/frontend/src/components/admin/spending/search-filter-form.tsx
ntindle: moved to 1/4 didn't do any design or anything on this so kinda just left it

autogpt_platform/backend/backend/data/credit.py
majdyz: >  IMO we should keep it as is until it becomes problematic since its an admin page

Try to search for something transaction on prod using userId contains in the table on prod, like:
`select * from CreditTransaction where userId ilike '%some_partial_id%'` and see if it works and how long.

If it's running fine in less than a second then I't's fine to proceed.

> turns it into hundreds of smaller queries

I don't get it, why can't we search userIds first from user table ? then use it to filter using `get_transaction_history`? we can modify it to take the list of user_ids instead too.

majdyz (commented):
Some searches feature can unintentionally strain our the db where replica is not existent atm.

autogpt_platform/backend/backend/data/credit.py
ntindle: unmatchable
image
matchable many times
image
toran
image

ntindle: the ui for this shows all users at the same time until you search, depending on the fact that getting all users can happen at once

Issue #7750, opened 2024-08-08
Resume dropped runs


Torantulino:
<img src="https://uploads.linear.app/a47946b5-12cd-4b3d-8822-df04c855879f/bbf5c341-dc8e-46dd-85dc-e30e83545202/9f7cbb1d-4a18-4fed-921e-db57418826f3?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiL2E0Nzk0NmI1LTEyY2QtNGIzZC04ODIyLWRmMDRjODU1ODc5Zi9iYmY1YzM0MS1kYzhlLTQ2ZGQtODVkYy1lMzBlODM1NDUyMDIvOWY3Y2JiMWQtNGExOC00ZmVkLTkyMWUtZGI1NzQxODgyNmYzIiwiaWF0IjoxNzIzMTA4Mzc4LCJleHAiOjMzMjkzNjY4Mzc4fQ.laJrqGhTu7cSwEp7iGT1uoMtvgZHP7cagj-VrcYtVI8 " alt="image.png" width="751" />

The agent execution status should be true at all times.

### 2-stage solution

1. #9732
2. (this issue): **Use system-wide locks to ensure ongoing executions are always held by one "responsible" executor node.** We can then detect dropped executions by checking the lock status of unfinished executions.

Pwuts:
After discussing possible solutions, here is a draft solution:

* Assign a (new) UUID to a node executor when it starts. The UUID corresponds to that specific executor session and is never reused after the executor goes down (planned or unplanned).
* Assign a node execution to a specific executor
* Store `executorID` in the DB on the `NodeExecution` entity etc
* The execution manager asserts that the executor for an in-progress `NodeExecution` is online. If not, it re-assigns the execution to a different executor.

Pwuts:
still relevant

Pwuts:
definitely still relevant

Pwuts:
@majdyz do you have an idea how to implement this? I just looked into it again but it seems we don't have a centralized source of information about which executor is busy with which execution, which I think we would need to implement a resumption mechanism like proposed by this issue.

Pwuts:
I think so

PR #9759, opened 2025-04-04
feat(backend/executor): Move execution queue + cancel mechanism to RabbitMQ

### LINKED ISSUE #9714: Move graph execution queue to RabbitMQ
The graph execution queue is not disk-persisted; when the executor dies, the executions are lost.

The scope of this issue is migrating the execution queue from an inter-process queue to a RabbitMQ message queue. A sync client should be used fo this.

### LINKED ISSUE #9746: Convert cancel execution request API into a fan-out event
There is already an API to cancel a running execution, however, the current implementation won't be reliable when working on a multi-executor setup. This is caused by the API call that needs to identify which executor is running the agent, and send it to the right executor to make it work. Instead of doing this, we can fan out the cancelation event that each executor subscribes to. 

The change of execution status can be done on the Rest API layer.

majdyz:
The graph execution queue is not disk-persisted; when the executor dies, the executions are lost.

The scope of this issue is migrating the execution queue from an inter-process queue to a RabbitMQ message queue. A sync client should be used for this.

- Resolves #9746
- Resolves #9714

### Changes 🏗️

Move the execution manager from multiprocess.Queue into persisted Rabbit-MQ.

#### For code changes:
- [x] I have made a test plan
  - [x] Execute agents.

<details>
  
</details>

<details>

  - Changing ports
</details>


autogpt_platform/backend/backend/executor/manager.py
```diff
@@ -952,31 +1011,110 @@ def run_service(self):
...
-            future.add_done_callback(
-                lambda _: self.active_graph_runs.pop(graph_exec_id, None)
+            if method_frame:
+                self._handle_run_message(channel, method_frame, body)
+            else:
+                time.sleep(0.1)
+
+    def _handle_cancel_message(self, body: bytes):
+        try:
+            graph_exec_id = json.loads(body).get("graph_exec_id")
```
Pwuts: Please define and use models for the message bodies

autogpt_platform/backend/backend/executor/manager.py
```diff
@@ -952,31 +1011,110 @@ def run_service(self):
...
+                cancel_event.set()
+            else:
+                logger.debug(
+                    f"[{self.service_name}] Cancel already set for {graph_exec_id}"
+                )
+
+        except Exception as e:
+            logger.exception(f"Error handling cancel message: {e}")
+
+    def _handle_run_message(self, channel, method_frame, body: bytes):
```
Pwuts: untyped parameters

Pwuts (commented):
Does this PR fix either/both of these issues? :)
- #7750
- #9746

autogpt_platform/backend/backend/executor/manager.py
majdyz: Updated

autogpt_platform/backend/backend/executor/manager.py
majdyz: Updated

Pwuts:
Does the persistence of the queue mean existing unfinished graph executions will be picked back up when the system is restarted?

autogpt_platform/backend/backend/executor/manager.py
```diff
@@ -1064,8 +1211,11 @@ def add_execution(
                 for node_exec in graph_exec.node_executions
             ],
         )
-        self.queue.add(graph_exec_entry)
-
+        self.rabbitmq_service.publish_message(
+            routing_key=GRAPH_EXECUTION_ROUTING_KEY,
+            message=json.dumps(graph_exec_entry),
```
Pwuts:
```suggestion
            message=graph_exec_entry.model_dump_json(),
```

Pwuts (dismissed):

autogpt_platform/backend/backend/executor/manager.py
```diff
@@ -952,31 +1018,112 @@ def run_service(self):
...
+        logger.info(f"[{self.service_name}] Ready to consume messages...")
         while True:
-            graph_exec_data = self.queue.get()
-            graph_exec_id = graph_exec_data.graph_exec_id
-            logger.debug(
-                f"[ExecutionManager] Dispatching graph execution {graph_exec_id}"
+            # cancel graph execution requests
+            method_frame, _, body = channel.basic_get(
+                queue=GRAPH_EXECUTION_CANCEL_QUEUE_NAME,
+                auto_ack=True,
```
ntindle, line 1026: nit: document that auto_ack is a shittly named arg

autogpt_platform/backend/backend/executor/manager.py
```diff
@@ -952,31 +1018,112 @@ def run_service(self):
...
-            future = self.executor.submit(
-                Executor.on_graph_execution, graph_exec_data, cancel_event
+            if method_frame:
+                self._handle_cancel_message(body)
+
+            # start graph execution requests
+            method_frame, _, body = channel.basic_get(
+                queue=GRAPH_EXECUTION_QUEUE_NAME,
+                auto_ack=False,
             )
```
ntindle, line 1035: are both channels using the saem frame?

autogpt_platform/backend/backend/executor/manager.py
```diff
@@ -952,31 +1018,112 @@ def run_service(self):
...
+            if not info:
+                return
+            _, _, delivery_tag = info
+            if future.exception():
+                logger.error(
+                    f"[{self.service_name}] Execution for {graph_exec_id} failed: {future.exception()}"
+                )
+                channel.basic_nack(delivery_tag, requeue=False)
+            else:
+                channel.basic_ack(delivery_tag)
```
ntindle, line 1108: the queue could be dead here and your nack or ack could fail

autogpt_platform/backend/backend/executor/manager.py
```diff
@@ -454,6 +465,51 @@ def validate_exec(
...
+)
+GRAPH_EXECUTION_QUEUE_NAME = "graph_execution_queue"
+GRAPH_EXECUTION_ROUTING_KEY = "graph_execution.run"
+
+GRAPH_EXECUTION_CANCEL_EXCHANGE = Exchange(
+    name="graph_execution_cancel",
+    type=ExchangeType.FANOUT,
+    durable=True,
+    auto_delete=True,
+)
```
ntindle, line 482: Heavily consider topic or header rather than direct. Changing exchange types later sucks

 fanout is good if you want a broadcast notification style. Not sure goal with that

autogpt_platform/backend/backend/executor/manager.py
```diff
@@ -1077,16 +1227,11 @@ def cancel_execution(self, graph_exec_id: str) -> None:
...
-                f"Graph execution #{graph_exec_id} not active/running: "
-                "possibly already completed/cancelled."
-            )
-        else:
-            future, cancel_event = self.active_graph_runs[graph_exec_id]
-            if not cancel_event.is_set():
-                cancel_event.set()
-                future.result()
+        self.rabbitmq_service.publish_message(
+            routing_key="",
```
ntindle, line 1237: is this a required param?

autogpt_platform/backend/backend/executor/manager.py
majdyz: What's so shit about it? 🥺 
The name sounds self-explanatory to me.

autogpt_platform/backend/backend/executor/manager.py
majdyz: no, it's just a variable used to read the data

autogpt_platform/backend/backend/executor/manager.py
majdyz: Why could it be dead and how can it be avoided? If a queue can be dead, then the executor breaks and it's the end of the world.

autogpt_platform/backend/backend/executor/manager.py
majdyz: yes

autogpt_platform/backend/backend/executor/manager.py
majdyz: I think for this use case we should avoid complicated filtering of the routing and should stick to having a queue of entries that is load balanced by executors. 

> Changing exchange types later sucks

Why is that so? Why can't we just introduce a new Exchange with different name & type then push ?

> fanout is good if you want a broadcast notification style. Not sure goal with that

It's obviously not the right use for execution queue, but intended for cancel execution, since all executors need to check if the execution is present on their side and cancel it accordingly.

majdyz:
> Does the persistence of the queue mean existing unfinished graph executions will be picked back up when the system is restarted?

@Pwuts yes, it should be this way.

Pwuts:
So does it resume or try to start over? We should probably test this scenario.

majdyz:
> So does it resume or try to start over? We should probably test this scenario.

Start over. Do you think the testing would be a blocker to getting this in?
I can followup with the test with my another change on removing RPC from the execution manager service.

Pwuts:
> Do you think the testing would be a blocker to getting this in?

No, I just wanted to know the implications of this change for that specific scenario.

Once this PR is merged I think it'll be fairly easy to fix this issue, which is good:
- #7750

autogpt_platform/backend/backend/executor/manager.py
ntindle: The queue can be configured to be durable so it will come back up with the same content, just worth thinking about this could throw an exception 

the service could go down between when you pop you message and when you decide to keep or reject it and it’ll leave your message hanging

autogpt_platform/backend/backend/executor/manager.py
ntindle: It’s behavior is weird compared to the name acto_ack being true means you have to manually ack or nack which is crazy

autogpt_platform/backend/backend/executor/manager.py
ntindle: Sounds good just want to make sure we like what we have

ntindle (approved):

ntindle:
Make sure durable modes are enabled if desired

PR #9773, opened 2025-04-07
feat(frontend, backend): track sentry environment on frontend + sentry init in app services

ntindle:
We want to be able to filter errors according to where they occur in sentry so we need to track and include that data. We also are not logging everything from app services correctly so fix that up

### Changes 🏗️

- Adds env tracking for frontend
- adds sentry init in app service spawn

#### For code changes:
- [x] I have made a test plan
  
  - [x] Tested by running and making sure all events + logs are inserted into sentry correctly


autogpt_platform/backend/backend/util/service.py
```diff
@@ -354,6 +355,7 @@ def __start_fastapi(self):
 
     def run(self):
         super().run()
+        sentry_init()
         self.fastapi_app = FastAPI()
```
majdyz, line 202: You might want to place it even before super().run

majdyz (approved):

autogpt_platform/frontend/src/lib/utils.ts
```diff
@@ -245,6 +245,19 @@ export function getBehaveAs(): BehaveAs {
     : BehaveAs.LOCAL;
 }
 
+export enum AppEnv {
+  DEV = "dev",
```
ntindle, line 250: should we have local?

autogpt_platform/backend/backend/util/service.py
ntindle: okay cool thanks!

autogpt_platform/frontend/src/lib/utils.ts
majdyz: if we don't have local, then local data will go to dev??

autogpt_platform/frontend/instrumentation-client.ts
```diff
@@ -2,12 +2,14 @@
 // The config you add here will be used whenever a users loads a page in their browser.
 // https://docs.sentry.io/platforms/javascript/guides/nextjs/
 
+import { getEnvironmentStr } from "@/lib/utils";
 import * as Sentry from "@sentry/nextjs";
 
 Sentry.init({
   dsn: "https://fe4e4aa4a283391808a5da396da20159@o4505260022104064.ingest.us.sentry.io/4507946746380288",
```
majdyz, line 9: can this move somewhere else ?

ntindle: I can put it here or in source but the others need to be here I think  so I figured they should stay together

autogpt_platform/frontend/src/lib/utils.ts
ntindle: The frontend doesn't currently have a concept of local, that's the point I'm questioning -- is this PR the right place to add it

ntindle:
Add

  _experiments: {
    // Enable logs to be sent to Sentry.
    enableLogs: true,
  },

autogpt_platform/frontend/instrumentation-client.ts
ntindle: the dsn is going to stay here for now :)

aarushik93 (approved):

PR #9541, opened 2025-02-28
feat(platform): Agent Presets

Pwuts:
- Resolves #9307

### Changes 🏗️

- WIP: feat(frontend): Agent Presets UI (+ needed changes to backend endpoints)

- fix(backend): Update DB schema
  - Fix relationships between `LibraryAgent`, `AgentGraph`, and `AgentPreset`
  - Impose uniqueness constraint on `LibraryAgent`
  - Update relationship names for DX (clarity and correct plurality)

- refactor(frontend): Brand `GraphExecution.execution_id` and `Schedule.id`

#### For code changes:
- [ ] I have clearly listed my changes in the PR description
- [ ] I have made a test plan
- [ ] I have tested my changes according to the test plan:
  
  - [ ] ...

<details>
  <summary>Example test plan</summary>

  - [ ] Create from scratch and execute an agent with at least 3 blocks
  - [ ] Import an agent from file upload, and confirm it executes correctly
  - [ ] Upload agent to marketplace
  - [ ] Import an agent from marketplace and confirm it executes correctly
  - [ ] Edit an agent from monitor, and confirm it executes correctly
</details>

Issue #9307, opened 2025-01-21
Implement "Pin into a template" menu option on agent runs


Pwuts:
* Part of #8780

<img src="https://uploads.linear.app/a47946b5-12cd-4b3d-8822-df04c855879f/ff5e52b8-936e-4294-a6bf-670c3fa247b6/71f0b319-57c5-401a-9689-7452a83c8bec?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiL2E0Nzk0NmI1LTEyY2QtNGIzZC04ODIyLWRmMDRjODU1ODc5Zi9mZjVlNTJiOC05MzZlLTQyOTQtYTZiZi02NzBjM2ZhMjQ3YjYvNzFmMGIzMTktNTdjNS00MDFhLTk2ODktNzQ1MmE4M2M4YmVjIiwiaWF0IjoxNzM3NDU5OTQ0LCJleHAiOjMzMzA4MDE5OTQ0fQ.OHsdTI41-n_awRsgoeIdiCpLTIEuVcRjJLHNXBAi9Hc " alt="image.png" width="451" data-linear-height="269" />

Pwuts:
Most of this has already been implented in #10167. Making it feature-complete would involve commenting out some code and writing a little bit more.

PR #9774, opened 2025-04-07
refactor(backend): Clean up Library & Store DB schema

Pwuts:
Distilled from #9541 to reduce the scope of that PR.

- Part of #9307

- ❗ Blocks #9786
  - ❗ Blocks #9541

### Changes 🏗️

- Fix `LibraryAgent` schema (for #9786)
  - Fix relationships between `LibraryAgent`, `AgentGraph`, and `AgentPreset`
  - Add uniqueness constraint on `LibraryAgent`

- Rename things that are called `agent` that actually refer to a `graph`/`agentGraph`
- Fix singular/plural forms in DB schema
- Simplify reference names of closely related objects (e.g. `AgentGraph.AgentGraphExecutions` -> `AgentGraph.Executions`)

- Eliminate use of `# type: ignore` in DB statements
  - Add `typed` and `typed_cast` utilities to `backend.util.type`

#### For code changes:
- [x] I have made a test plan
  - [x] CI static type checking (with all risky `# type: ignore` removed)
  - [x] Check that column references in views are updated

### DIFF
--- a/autogpt_platform/backend/schema.prisma
+++ b/autogpt_platform/backend/schema.prisma
@@ -141,7 +141,7 @@ model AgentPreset {
 
   agentGraphId      String
   agentGraphVersion Int
-  AgentGraph        AgentGraph @relation(fields: [agentGraphId, agentGraphVersion], references: [id, version], onDelete: Cascade)
+  AgentGraph        AgentGraph @relation(fields: [agentGraphId, agentGraphVersion], references: [id, version], onDelete: Restrict)
 
   InputPresets AgentNodeExecutionInputOutput[] @relation("AgentPresetsInputData")
   Executions   AgentGraphExecution[]
@@ -206,7 +206,7 @@ model LibraryAgent {
 
   agentGraphId      String
   agentGraphVersion Int
-  AgentGraph        AgentGraph @relation(fields: [agentGraphId, agentGraphVersion], references: [id, version])
+  AgentGraph        AgentGraph @relation(fields: [agentGraphId, agentGraphVersion], references: [id, version], onDelete: Restrict)
 
   creatorId String?
   Creator   Profile? @relation(fields: [creatorId], references: [id])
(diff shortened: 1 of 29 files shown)


autogpt_platform/backend/schema.prisma
```diff
@@ -39,19 +39,19 @@ model User {
   AgentGraphExecutions AgentGraphExecution[]
   AnalyticsDetails     AnalyticsDetails[]
   AnalyticsMetrics     AnalyticsMetrics[]
-  CreditTransaction    CreditTransaction[]
```
majdyz, line 42: The changes are rather optimistic, what's your testing strategy?

Have we checked all the type ignored include & where clause? what about raw queries?

majdyz (dismissed):
This is a good change in general but sounds a bit risky too, can we write down on what we should do to make sure things not breaking? I think this is the more important things to review.

autogpt_platform/backend/schema.prisma
ntindle: Or views in past migrations?

Pwuts: > The changes are rather optimistic, what's your testing strategy?
> 
> Have we checked all the type ignored include & where clause? what about raw queries?

Good point.
Fixed: https://github.com/Significant-Gravitas/AutoGPT/pull/9774/commits/54f58da2f34637f40d290901e10163afafb3a6a6

There are no affected raw queries.

> Or views in past migrations?

`ALTER TABLE [table] RENAME COLUMN` automatically updates references such as views and foreign keys. The migration runs successfully and the view doesn't break (I tested this locally).

I'm explicitly recreating the foreign keys because otherwise the name doesn't match up with what Prisma expects.

majdyz: Nice, last question: I suppose there would be no risk of migration failing due to data? since this is just a column renames ?

majdyz (approved):

autogpt_platform/backend/schema.prisma
Pwuts: That's what this query is for:
https://github.com/Significant-Gravitas/AutoGPT/blob/ee27ead5412884a3470fb967a77e004ed15659c5/autogpt_platform/backend/migrations/20250227140210_fix_library_presets_relations/migration.sql#L9-L18

Neither dev nor prod have entries in the DB that violate this new unique constraint.

PR #9777, opened 2025-04-07
fix(ci): Fix linter exit code on failure

Pwuts:
The linter currently exits with exit code 0 even if linting fails. This makes the CI linter permissive which isn't good.

Changes:
- Make linter exit with an error code if a linting step fails
- Fix existing formatting issues


ntindle (approved):
Should we run them all and wait then show each error side by side?

Pwuts:
> Should we run them all and wait then show each error side by side?

That's what it does now :)

autogpt_platform/backend/linter.py
```diff
@@ -21,23 +21,31 @@ def run(*command: str) -> None:
...
+        print("Lint failed, try running `poetry run format` to fix the issues")
+        raise lint_error
 
 
 def format():
     run("ruff", "check", "--fix", *TARGET_DIRS)
     run("ruff", "format", LIBS_DIR)
     run("isort", "--profile", "black", BACKEND_DIR)
     run("black", BACKEND_DIR)
-    run("pyright", *TARGET_DIRS)
```
majdyz, line 43: Please don't remove this, it will be massive annoyance to keep running stuff separately for devs :'(

majdyz: Oh we have lint nvm haha

majdyz (dismissed):

autogpt_platform/backend/linter.py
```diff
@@ -21,23 +21,31 @@ def run(*command: str) -> None:
...
+        print("Lint failed, try running `poetry run format` to fix the issues")
+        sys.exit(1)
 
 
 def format():
     run("ruff", "check", "--fix", *TARGET_DIRS)
     run("ruff", "format", LIBS_DIR)
     run("isort", "--profile", "black", BACKEND_DIR)
     run("black", BACKEND_DIR)
-    run("pyright", *TARGET_DIRS)
```
ntindle: why remove :(
I use this all the time

Pwuts: Lint = check, format = fix. This is a check that is under the `format` command. 

How do you use it?

I'd personally recommend setting up the `pre-commit` hook instead of manually running the linters/formatters, although ofc to each their own preference.

ntindle: poetry run format, yarn run format. I don't want to commit every time I run it

ntindle: from a dev perspective, running pyright is a fixer, because it tells me what I need to fix, if you remove it, please add a command called fix to both the yarn and poetry

Pwuts: ok I see, I'll put it back

ntindle (approved):

autogpt_platform/backend/linter.py
```diff
@@ -21,18 +21,27 @@ def run(*command: str) -> None:
...
+    ]
+    lint_error = None
+    for args in lint_step_args:
+        try:
+            run(*args)
+        except subprocess.CalledProcessError as e:
+            lint_error = e
+
+    if lint_error:
+        print("Lint failed, try running `poetry run format` to fix the issues")
```
majdyz, line 43: Why don't we print the error?

Pwuts: It's printed on line 23

PR #9789, opened 2025-04-09
feat(platform/library): Add credentials UX on `/library/agents/[id]`

### LINKED ISSUE #9771: Add credentials input UX on `/library/agents/[id]`
Currently, adding credentials to agents happens in the builder on the individual nodes. This only works for the graph owner though: **there is no way to do this for an agent acquired through the marketplace.**

We urgently need a way to set all the credentials for a graph from the Library.

~~I think the logical way to implement this is using presets:~~

* ~~Blocked by~~ #9307

Pwuts:
- Resolves #9771
  - ... in a non-persistent way, so it won't work for webhook-triggered agents
    For webhooks: #9541

### Changes 🏗️

Frontend:
- Add credentials inputs in Library "New run" screen (based on `graph.credentials_input_schema`)
  - Refactor `CredentialsInput` and `useCredentials` to not rely on XYFlow context

- Unsplit lists of saved credentials in `CredentialsProvider` state

- Move logic that was being executed at component render to `useEffect` hooks in `CredentialsInput`

Backend:
- Implement logic to aggregate credentials input requirements to one per provider per graph
  - Add `BaseGraph.credentials_input_schema` (JSON schema) computed field
    Underlying added logic:
    - `BaseGraph._credentials_input_schema` - makes a `BlockSchema` from a graph's aggregated credentials inputs
    - `BaseGraph.aggregate_credentials_inputs()` - aggregates a graph's nodes' credentials inputs using `CredentialsFieldInfo.combine(..)`
    - `BlockSchema.get_credentials_fields_info() -> dict[str, CredentialsFieldInfo]`
      - `CredentialsFieldInfo` model (created from `_CredentialsFieldSchemaExtra`)

- Implement logic to inject explicitly passed credentials into graph execution
  - Add `credentials_inputs` parameter to `execute_graph` endpoint
  - Add `graph_credentials_input` parameter to `.executor.utils.add_graph_execution(..)`
  - Implement `.executor.utils.make_node_credentials_input_map(..)`
  - Amend `.executor.utils.construct_node_execution_input`
  - Add `GraphExecutionEntry.node_credentials_input_map` attribute
  - Amend validation to allow injecting credentials
    - Amend `GraphModel._validate_graph(..)`
    - Amend `.executor.utils._validate_node_input_credentials`
  - Add `node_credentials_map` parameter to `ExecutionManager.add_execution(..)`
    - Amend execution validation to handle side-loaded credentials
    - Add `GraphExecutionEntry.node_execution_map` attribute
    - Add mechanism to inject passed credentials into node execution data
  - Add credentials injection mechanism to node execution queueing logic in `Executor._on_graph_execution(..)`

- Replace boilerplate logic in `v1.execute_graph` endpoint with call to existing `.executor.utils.add_graph_execution(..)`
  - Replace calls to `.server.routers.v1.execute_graph` with `add_graph_execution`

Also:
- Address tech debt in `GraphModel._validate_gaph(..)`
- Fix type checking in `BaseGraph._generate_schema(..)`

#### TODO
- [ ] ~~Make "Run again" work with credentials in `AgentRunDetailsView`~~
- [ ] Prohibit saving a graph if it has nodes with missing discriminator value for discriminated credentials inputs

#### For code changes:
- [ ] I have made a test plan
  
  - [ ] ...


autogpt_platform/backend/backend/data/graph.py
```diff
@@ -332,13 +426,6 @@ def sanitize(name):
             if (block := nodes_block.get(node.id)) is None:
                 raise ValueError(f"Invalid block {node.block_id} for node #{node.id}")
 
-            # Smart decision maker nodes
-            if block.block_type == BlockType.AI:
-                smart_decision_maker_nodes.add(node.id)
-            # Agent nodes
-            elif block.block_type == BlockType.AGENT:
-                agent_nodes.add(node.id)
```
Pwuts, line 340: These variables aren't used anywhere so this is dead code

autogpt_platform/backend/backend/server/external/routes/v1.py
```diff
@@ -97,13 +97,13 @@ async def execute_graph(
     api_key: APIKey = Depends(require_permission(APIKeyPermission.EXECUTE_GRAPH)),
 ) -> dict[str, Any]:
     try:
-        graph_exec = await internal_api_routes.execute_graph(
+        graph_exec = add_graph_execution(
```
Pwuts: I've done this in a couple of places @majdyz, everywhere (including the `execute_graph` endpoint itself) now uses `.executor.utils.add_graph_execution(..)`.

autogpt_platform/backend/backend/server/routers/v1.py
```diff
@@ -592,35 +593,21 @@ def get_credentials(credentials_id: str) -> "Credentials | None":
...
-    await bus.publish(graph_exec)
-
-    queue = await execution_queue_client()
-    await queue.publish_message(
-        routing_key=execution_utils.GRAPH_EXECUTION_ROUTING_KEY,
-        message=graph_exec.to_graph_execution_entry().model_dump_json(),
-        exchange=execution_utils.GRAPH_EXECUTION_EXCHANGE,
+        graph_version=graph_version,
+        graph_credentials_inputs=credentials_inputs,
     )
```
Pwuts: Because I've replaced this boilerplate logic with `add_graph_execution(..)`, this is now a fully sync call. Any issues with that @majdyz?

autogpt_platform/backend/backend/data/block.py
```diff
@@ -203,6 +204,16 @@ def get_credentials_fields(cls) -> dict[str, type[CredentialsMetaInput]]:
             )
         }
 
+    @classmethod
+    def get_credentials_fields_info(cls) -> dict[str, CredentialsFieldInfo]:
+        return {
+            field_name: CredentialsFieldInfo.model_validate(
+                cls.jsonschema()["properties"][field_name],
```
majdyz: There is `get_field_schema` for this

autogpt_platform/backend/backend/data/graph.py
```diff
@@ -190,14 +197,19 @@ def output_schema(self) -> dict[str, Any]:
             )
         )
 
+    @computed_field
+    @property
+    def credentials_input_schema(self) -> dict[str, Any]:
```
majdyz, line 202: Seeing the usage of this, why don't we just return the BlockSchema and use (or add more) fields to access it instead of a raw dictionary?

autogpt_platform/backend/backend/data/graph.py
```diff
@@ -217,9 +229,93 @@ def _generate_schema(
...
+        """
+        Returns:
+            dict[aggregated_field_key, tuple(
+                CredentialsFieldInfo: A spec for one aggregated credentials field
+                set[(node_id, field_name)]: Node credentials fields that are
+                    compatible with this aggregated field spec
+            )]
+        """
+        return {
+            "_".join(sorted(agg_field_info.provider))
```
majdyz, line 295: can these list fields empty ?

autogpt_platform/backend/backend/server/external/routes/v1.py
majdyz: We can't, you can move the code for execute_graph endpoint somewhere else as a fully async shrrable code and use it if you want. But we should never use blocking I/O call in async code.

majdyz: If you literally have no other option, then make the route sync, or use to_thread or spin a new thread for handling the request. But I'd like to avoid this as much as possible.

autogpt_platform/backend/backend/server/routers/v1.py
majdyz: Discussed https://github.com/Significant-Gravitas/AutoGPT/pull/9789/files#r2048731698

autogpt_platform/backend/backend/data/graph.py
Pwuts: nope, always at least 1 entry

autogpt_platform/backend/backend/data/graph.py
Pwuts: What do you mean? I use `Graph.credentials_input_schema` in the frontend to render the credentials inputs. Returning and serializing a `BlockSchema` doesn't give the same result.

autogpt_platform/backend/backend/executor/utils.py
```diff
@@ -569,7 +587,72 @@ def create_execution_queue_config() -> RabbitMQConfig:
...
+    await graph_execution_queue.publish_message(
+        routing_key=GRAPH_EXECUTION_ROUTING_KEY,
+        message=graph_exec_entry.model_dump_json(),
+        exchange=GRAPH_EXECUTION_EXCHANGE,
+    )
+
+    return graph_exec
+
+
+def add_graph_execution_sync(
```
majdyz: This looks good, but let's be consistent on the naming. We've been adding async to the function names for the async version of it, like for the event bus and event queue. So let's just do that for this and not invert it.

autogpt_platform/backend/backend/data/graph.py
majdyz: I did not know this is a property field for the autogenerating field for the api.

majdyz (approved):

autogpt_platform/backend/backend/executor/utils.py
Pwuts: This seemed the sensible way around to me because the async version is used the most. But fair enough, I'll change it.

majdyz (approved):

PR #9805, opened 2025-04-10
fix(backend): Fix unchecked Prisma statements

Pwuts:
- Follow-up to #9774

After merging #9774 it became clear that I didn't fix *all* DB statements, and that some of them were still unchecked due to limitations in the Prisma client's generated type definitions.

### Changes 🏗️

- Configure Prisma client to generate recursive types
- Fix type issues in DB statements
- Fix type issues in `test_data_creator.py`
- Remove now-unused type utilities introduced in #9774

#### For code changes:
- [x] I have made a test plan
  
  - [x] This time CI should actually suffice since there are no longer any unchecked DB statements

### DIFF
--- a/autogpt_platform/backend/backend/data/execution.py
+++ b/autogpt_platform/backend/backend/data/execution.py
@@ -39,7 +39,7 @@ from backend.util import type as type_utils
 from backend.util.settings import Config
 
 from .block import BlockData, BlockInput, BlockType, CompletedBlockOutput, get_block
-from .db import BaseDbModel
+from .db import BaseDbModel, locked_transaction
 from .includes import (
     EXECUTION_RESULT_INCLUDE,
     GRAPH_EXECUTION_INCLUDE,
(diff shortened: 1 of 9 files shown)


autogpt_platform/backend/backend/data/execution.py
```diff
@@ -492,20 +492,25 @@ async def update_graph_execution_stats(
     data = stats.model_dump() if stats else {}
     if isinstance(data.get("error"), Exception):
         data["error"] = str(data["error"])
-    res = await AgentGraphExecution.prisma().update(
-        where={
-            "id": graph_exec_id,
-            "OR": [
```
ntindle, line 498: Could we do execution status in [enums]?

ntindle (approved):

autogpt_platform/backend/backend/data/execution.py
Pwuts: Nope, `update` takes in a `...WhereUniqueInput` which only includes unique keys/indices. `update_many` would allow it but doesn't return the updated objects.

autogpt_platform/backend/backend/data/graph.py
```diff
@@ -930,12 +919,19 @@ async def migrate_llm_models(migrate_to: LlmModel):
...
+            AND "constantInput"->>$4 NOT IN {escaped_enum_values}
+            """  # type: ignore
+
+        await db.execute_raw(
+            query,  # type: ignore - is supposed to be LiteralString
+            "{" + path + "}",
+            f'"{migrate_to.value}"',
+            id,
+            path,
+        )
```
Pwuts: *not tested*

autogpt_platform/backend/backend/data/execution.py
majdyz: If the additional clause is indeed ignored, let's just use update_many return None if the result is < 1 else return the fund_unique result.

autogpt_platform/backend/test/test_data_creator.py
```diff
@@ -362,26 +361,26 @@ async def main():
...
-                subHeading=faker.sentence(),
-                videoUrl=faker.url(),
-                imageUrls=[get_image() for _ in range(3)],
-                description=faker.text(),
-                categories=[faker.word() for _ in range(3)],
-                isFeatured=random.choice([True, False]),
-                isAvailable=True,
-                storeListingId=listing.id,
-                submissionStatus=random.choice(
+            data={
```
majdyz, line 364: have you tried poetry install ? isn't this breaking with the new pydantic ?

ntindle (dismissed):

autogpt_platform/backend/test/test_data_creator.py
Pwuts: How would it break? It doesn't give any type errors.

autogpt_platform/backend/backend/data/execution.py
Pwuts: fixed

ntindle (approved):

PR #9821, opened 2025-04-15
fix(frontend): Fix Input value mixup on Library page 

majdyz:
### Changes 🏗️

Fix this broken behaviors:
Input data mix-up caused by running two different executions of the same agent with the same input.

#### For code changes:
- [x] I have made a test plan
  - [x] Run agent with old user
  - [x] Running two different executions of the same agent with the same input.


Pwuts (dismissed):

kcze (changes requested):
I've removed this from queue because both prisma schema and onboarding provider give empty array and should at no point give null/undefined, need to investigate.

PR #9828, opened 2025-04-16
fix(backend): Fix array types in database

kcze:
Array fields in `schema.prisma` are non-nullable, but generated migrations don’t add `NOT NULL` constraints. This causes existing rows to get `NULL` values when new array columns are added, breaking schema expectations and leading to bugs.

### Changes 🏗️

- Backfill all `NULL` rows on non-nullable array columns to empty arrays
- Set `NOT NULL` constraint on all array columns

#### For code changes:
- [x] I have made a test plan
  - [x] Existing `NULL` rows are properly backfilled
  - [x] Existing arrays are not set to default empty arrays
  - [x] Affected columns became non-nullable in the db


majdyz (approved):
Thank you!

PR #9838, opened 2025-04-18
fix(backend): Handle add execution API request failure

majdyz:
There are cases where the publishing agent execution is failing, making the agent execution appear to be stuck in a queue, but the execution has never been in a queue in the first place.

### Changes 🏗️

On publishing failure, we set the graph & starting node execution status to FAILED and let the UI bubble up the error so the user can try again.

#### For code changes:
- [x] I have made a test plan
  - [x] Normal add execution flow


ntindle (approved):

ntindle (approved):

majdyz:
Rebasing since it seems stuck

PR #9843, opened 2025-04-18
feat(platform): Update Marketplace Agent listing buttons

kcze:
Currently agent listing on Marketplace have bad UX.

### Changes 🏗️

- Add function and endpoint to check if user has `LibraryAgent` by given `storeListingVersionId`
- Redesign listing buttons
  - `Add to library` shown when user is logged in and doesn't have an agent in library
  - `See runs` shown when user logged in as has the agent in the library
  - `Download agent` always shown
  - Disabled buttons during processing (adding/downloading)
- Stop raising when owner is trying to add own agent. Now it'll simply redirect to Library.
- Remove button appearing/flickering after a delay on listing page - logged in status is now checked in server component.
- Show error toast on adding/redirecting to library and downloading error
- Update breadcrumbs and page title to say `Marketplace` instead of `Store`
- `font-geist` -> `font-sans` (`font-geist` var doesn't exist)

#### For code changes:
- [x] I have made a test plan
  - [x] Button on a listing is `Add to library` (no library agent)
  - [x] Agent can be added and user is redirected
  - [x] Button on the listing is `See runs` and clicking it redirects to the library agent
  - [x] Remove agent from library
  - [x] Buttons shows `Add to library` again
  - [x] Agent can be re-added
  - [x] Agent can be downloaded
  - [x] `Add to library` Button is hidden when user is logged out


autogpt_platform/backend/backend/server/v2/library/db.py
```diff
@@ -390,11 +390,6 @@ async def add_store_agent_to_library(
                 )
 
             graph = store_listing_version.AgentGraph
-            if graph.userId == user_id:
-                logger.warning(
-                    f"User #{user_id} attempted to add their own agent to their library"
-                )
-                raise store_exceptions.DatabaseError("Cannot add own agent to library")
```
majdyz, line 404: Why do we remove this ?

majdyz (dismissed):

autogpt_platform/backend/backend/server/v2/library/db.py
kcze: 1. To not show error on the frontend unnecessarily or have to handle it explicitly.
2. Previously if you already had an agent it wouldn't raise but if you have it *and* your're the owner it would which is inconsistent.
3. This allows you to re-add your agent from the marketplace if you happen to delete library agent linked to it.

autogpt_platform/backend/backend/server/v2/library/db.py
```diff
@@ -175,6 +175,52 @@ async def get_library_agent(id: str, user_id: str) -> library_model.LibraryAgent
...
+                "agentGraphVersion": store_listing_version.agentGraphVersion,
+                "isDeleted": False,
+            },
+            include={"AgentGraph": True},
+        )
+        if agent:
+            return library_model.LibraryAgent.from_db(agent)
+        else:
+            return None
+    except store_exceptions.AgentNotFoundError:
```
majdyz: what's the point of this

autogpt_platform/backend/backend/server/v2/library/db.py
```diff
@@ -175,6 +175,52 @@ async def get_library_agent(id: str, user_id: str) -> library_model.LibraryAgent
...
+        raise
+    except Exception as e:
+        logger.error(f"Error getting library agent: {e}")
+        raise store_exceptions.DatabaseError("Failed to fetch agent details") from e
```
majdyz: Do we really need to catch everything, can we just remove it?
there is also catch-all pattern (which not actually needed too, but seems like a pattern in the routes on that file unfortunately).

majdyz (approved):

autogpt_platform/backend/backend/server/v2/library/db.py
kcze: removed

autogpt_platform/backend/backend/server/v2/library/db.py
kcze: removed as well

autogpt_platform/frontend/src/components/agptui/AgentInfo.tsx
```diff
@@ -35,28 +39,48 @@ export const AgentInfo: React.FC<AgentInfoProps> = ({
...
-
-  const handleAddToLibrary = async () => {
+  const [adding, setAdding] = useState(false);
+  const [downloading, setDownloading] = useState(false);
+
+  const libraryAction = useCallback(async () => {
+    setAdding(true);
+    if (libraryAgent) {
+      // Redirect to the library agent page
+      router.push(`/library/agents/${libraryAgent.id}`);
```
majdyz: Will the router.push interrupt the toast? 
Should we at least call toast before the push?

majdyz (approved):

autogpt_platform/frontend/src/components/agptui/AgentInfo.tsx
majdyz: Also the CI seems to break

kcze: It doesn't interrupt but I moved it anyway

PR #9845, opened 2025-04-19
fix(library/run): Replace credits to cents

Abhi1992002:
Replacing credits with cents (100 credits = 1$).

I haven’t touched anything internally, just changed the UI.

Everything is working great.

On the frontend, there’s no other place where we use credits instead of dollars. 

Screenshot 2025-04-19 at 11 36 00 AM
Screenshot 2025-04-19 at 11 33 24 AM

### DIFF
--- a/autogpt_platform/frontend/src/components/agents/agent-run-details-view.tsx
+++ b/autogpt_platform/frontend/src/components/agents/agent-run-details-view.tsx
@@ -23,6 +23,7 @@ import {
   AgentRunStatus,
   agentRunStatusMap,
 } from "@/components/agents/agent-run-status-chip";
+import useCredits from "@/hooks/useCredits";
 
 export default function AgentRunDetailsView({
   agent,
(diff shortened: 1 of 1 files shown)


autogpt_platform/frontend/src/components/agents/agent-run-details-view.tsx
```diff
@@ -66,7 +66,7 @@ export default function AgentRunDetailsView({
               value: moment.duration(run.stats.duration, "seconds").humanize(),
             },
             { label: "Steps", value: run.stats.node_exec_count },
-            { label: "Cost", value: `${run.stats.cost} credits` },
+            { label: "Cost", value: `$ ${run.stats.cost / 100}` },
```
majdyz: There is formatCredit In useCredit, you should use that instead

Abhi1992002: @majdyz I’m sorry for this mistake. I should have checked how we’re rendering the cost at other places before doing this.

majdyz (approved):

PR #9858, opened 2025-04-22
feat(backend): Require discriminator value on graph save

Pwuts:
If a node has a multi-credentials input (e.g. AI Text Generator block) but the discriminator value (e.g. model choice) is missing, the input can't be discriminated into a single-provider input. Discrimination into a single-provider input is necessary to make a graph-level credentials input for use in the Library.

### Changes 🏗️

- feat(backend): Require discriminator fields to always have a value

- dx(frontend): Improve typing of discriminator stuff
- dx(frontend): Fix typing in `NodeOneOfDiscriminatorField` component

#### For code changes:
- [x] I have made a test plan
  - [x] Saving & running graphs with and without credentials works normally
  - Note: We don't have any blocks with a discriminator that doesn't have a default value, so currently I don't think it's possible to produce a case where this mechanism would be triggered.

### DIFF
--- a/autogpt_platform/backend/backend/data/graph.py
+++ b/autogpt_platform/backend/backend/data/graph.py
@@ -491,7 +491,6 @@ class GraphModel(Graph):
             # Validate dependencies between fields
             for field_name in input_fields.keys():
                 field_json_schema = InputSchema.get_field_schema(field_name)
-                logger.info(f"Field {field_name} schema: {field_json_schema}")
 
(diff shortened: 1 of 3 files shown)


majdyz (approved):

Pwuts:
Frontend checks are failing on the "add all blocks" test. 🤔

kcze (approved):
LGTM

PR #9870, opened 2025-04-23
feat(platform): Forking agent in Library

kcze:
This PR introduces copying agents feature in the Library. Users can copy and download their library agents but they can edit only the ones they own (included copied ones).

### Changes 🏗️

- DB migration: add relation in `AgentGraph`: `forked_from_id` and `forked_from_version`
- Add `fork_graph` function that makes a hardcopy of agent graph and its nodes (all with new ids)
- Add `fork_library_agent` that copies library agent and its graph for a user
- Add endpoint `/library/agents/{libraryAgentId}/fork`
- Add UI to `library/agents/[id]/page.tsx`: `Edit a copy` button with dialog confirmation

#### For code changes:
- [x] I have made a test plan
  - [x] Agent can be copied, edited and runs


autogpt_platform/backend/schema.prisma
```diff
@@ -118,6 +118,11 @@ model AgentGraph {
   // This allows us to delete user data with deleting the agent which maybe in use by other users
   User   User   @relation(fields: [userId], references: [id], onDelete: Cascade)
 
+  forkedFromId      String?
+  forkedFromVersion Int?
+  forkedFrom        AgentGraph? @relation("AgentForks", fields: [forkedFromId, forkedFromVersion], references: [id, version])
+  forks             AgentGraph[] @relation("AgentForks")
```
Pwuts:
```suggestion
  forkedFrom        AgentGraph? @relation("AgentGraphForks", fields: [forkedFromId, forkedFromVersion], references: [id, version])
  forks             AgentGraph[] @relation("AgentGraphForks")
```

autogpt_platform/backend/backend/server/v2/library/db.py
```diff
@@ -238,7 +240,8 @@ async def create_library_agent(
                         "graphVersionId": {"id": graph.id, "version": graph.version}
                     }
                 },
-            )
+            ),
+            include={"AgentGraph": True},
```
Pwuts, line 247: I'd advise changing the return type of this function to `library_model.LibraryAgent` instead of the prisma model

autogpt_platform/backend/backend/server/v2/library/routes/agents.py
```diff
@@ -190,3 +190,14 @@ async def update_library_agent(
             status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
             detail="Failed to update library agent",
         ) from e
+
+
+@router.post("/{library_agent_id}/clone")
```
Pwuts: what's the idea behind the mixed use of `clone` and `fork`?

autogpt_platform/frontend/src/lib/autogpt-server-api/client.ts
```diff
@@ -600,6 +600,14 @@ export default class BackendAPI {
     await this._request("PUT", `/library/agents/${libraryAgentId}`, params);
   }
 
+  /**
+   * Clone Library Agent and its underyling graph&nodes.
+   * User must own
+   */
+  deepCloneLibraryAgent(libraryAgentId: LibraryAgentID): Promise<LibraryAgent> {
```
Pwuts: nit: `deepClone` is a technical detail that doesn't matter to the client, right? Could be `clone` or `fork` instead.

autogpt_platform/frontend/src/lib/autogpt-server-api/client.ts
```diff
@@ -600,6 +600,14 @@ export default class BackendAPI {
     await this._request("PUT", `/library/agents/${libraryAgentId}`, params);
   }
 
+  /**
+   * Clone Library Agent and its underyling graph&nodes.
+   * User must own
```
Pwuts:
```suggestion
   * User must own the underlying graph for this to succeed.
```

autogpt_platform/frontend/src/app/(platform)/library/agents/[id]/page.tsx
```diff
@@ -246,6 +275,10 @@ export default function AgentRunsPage(): React.ReactElement {
               href: `/build?flowID=${agent.graph_id}&flowVersion=${agent.graph_version}`,
             },
             { label: "Export agent to file", callback: downloadGraph },
+            {
+              label: "Edit a copy",
```
Comment (Pwuts) [inline autogpt_platform/frontend/src/app/(platform)/library/agents/[id]/page.tsx]:
```suggestion
              label: "Create & edit a copy",
```
or

```suggestion
              label: "Copy & Edit",
```

cc @ograce1421

autogpt_platform/backend/backend/data/graph.py
```diff
@@ -26,6 +26,7 @@
     CredentialsMetaInput,
     is_credentials_field_name,
 )
+from backend.integrations.creds_manager import IntegrationCredentialsManager
```
Pwuts: I kinda wish we could avoid importing this here.. feels like importing high level stuff into a low level module. Wdyt? Could just be me.

Pwuts (commented):
Looks pretty neat, nice work! Comments are mostly technical/naming.

autogpt_platform/frontend/src/app/(platform)/library/agents/[id]/page.tsx
Comment (kcze) [reply autogpt_platform/frontend/src/app/(platform)/library/agents/[id]/page.tsx]: imo none of these, including original, are accurate - we're copying (cloning) and viewing, not editing. How about `Make a copy` or `Copy agent` or `Make editable copy` (because marketplace ones are not editable)?

autogpt_platform/backend/backend/server/v2/library/routes/agents.py
kcze: I made it consistently use `fork*`; I think it's more accurate because it's copying & changing ids + without `deep` clone might suggest we're only copying library agent (and not the graph as well).

autogpt_platform/frontend/src/lib/autogpt-server-api/client.ts
kcze: I've removed comment completely now

autogpt_platform/backend/backend/data/graph.py
kcze: I've moved it to `fork_library_agent` which is one step higher in chain.
But I had to move this as well:
```py
graph = await on_graph_activate(
    graph,
    get_credentials=lambda id: integration_creds_manager.get(user_id, id),
)
```

autogpt_platform/frontend/src/app/(platform)/library/agents/[id]/page.tsx
Comment (ograce1421) [reply autogpt_platform/frontend/src/app/(platform)/library/agents/[id]/page.tsx]: Let's keep it as how it is in the Figma File: "Edit a copy"

The reason why we're using "edit a copy" is:
1. To keep it as simple as possible for our non-technical users.
2. And i'm trying to get us into the convention of using verbs in the beginning of our call to action (CTA) buttons. 
This is to help users to quickly understand what action they can do / what they get if they press on the button

Comment (kcze) [reply autogpt_platform/frontend/src/app/(platform)/library/agents/[id]/page.tsx]: We're keeping it as is - according to the design

Comment (Pwuts) [reply autogpt_platform/frontend/src/app/(platform)/library/agents/[id]/page.tsx]: > we're copying (cloning) and viewing, not editing

> marketplace [agents] are not editable

vs

> Let's keep it as how it is in the Figma File: "Edit a copy"

> This is to help users to quickly understand what action they can do / what they get if they press on the button

This doesn't seem to add up to me. @ograce1421 I assume that the button caption being accurate trumps the need for simplicity, right? We either need to change the behavior or the caption. Otherwise it gives the false impression that this button will take them to a place where they can make edits - which it doesn't. 

Also, @kcze am I understanding correctly that marketplace agents *can* be forked, but the resulting forks *aren't* editable? If that's the case, what's the point of letting users fork them?

Comment (kcze) [reply autogpt_platform/frontend/src/app/(platform)/library/agents/[id]/page.tsx]: It is possible to edit forked agents. what I meant is we aren't opening them in Builder (i.e. editing immediately after copying)

Comment (Pwuts) [reply autogpt_platform/frontend/src/app/(platform)/library/agents/[id]/page.tsx]: In that case "Edit a copy" is misleading imo. Waiting on Olivia to reply whether to change the caption or the action.

PR #9875, opened 2025-04-24
fix(block): Fix Smart Decision Block missing input beads & incompability with input in special characters

majdyz:
Smart Decision Block was not able to work with sub agent with custom name input & the bead were not properly propagated in the execution UI. The scope of this PR is fixing it.

### Changes 🏗️

* Introduce an easy to parse format of tool edge: `{tool}_^_{func}_~_{arg}`. Graph using SmartDecisionBlock needs to be re-saved before execution to work.
* Reduce cluttering on a smart decision block logic.
* Fix beads not being shown for a smart decision block tool calling.

#### For code changes:
- [x] I have made a test plan
  - [x] Execute an SDM with some special character input as a tool

image

### DIFF
--- a/autogpt_platform/backend/backend/blocks/smart_decision_maker.py
+++ b/autogpt_platform/backend/backend/blocks/smart_decision_maker.py
@@ -247,6 +247,10 @@ class SmartDecisionMakerBlock(Block):
         )
 
     @staticmethod
+    def cleanup(s: str):
+        return re.sub(r"[^a-zA-Z0-9_-]", "_", s).lower()
+
+    @staticmethod
     def _create_block_function_signature(
         sink_node: "Node", links: list["Link"]
     ) -> dict[str, Any]:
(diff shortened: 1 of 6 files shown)


autogpt_platform/backend/backend/blocks/smart_decision_maker.py
```diff
@@ -503,7 +507,7 @@ def run(
             tool_args = json.loads(tool_call.function.arguments)
 
             for arg_name, arg_value in tool_args.items():
-                yield f"tools_^_{tool_name}_{arg_name}".lower(), arg_value
+                yield f"tools_^_{tool_name}_~_{arg_name}", arg_value
```
kcze, line 510: Why doesn't this yield a regular object instead?
Not a fan, same with `#@$` on other outputs, this is becoming a weird DSL and imo a bad idea but too late I guess.

majdyz: yeah not trying to rewrite anything at this point 😢

majdyz: but no you can't just yield regular object because a single pin is connected to multiple blocks and each block needs to filter what they are looking for.

Torantulino (approved):

PR #9876, opened 2025-04-24
feat(frontend): Update "Edit a copy" modal and buttons

kcze:
Update "Edit a copy" modal text when copying marketplace agent in Library. Update agent action buttons to reflect the design accurately.

### Changes 🏗️

- Update modal text
- Disable copying owned agents (only marketplace allowed)
- `Open in Builder` -> `Customize agent`
- Disabled `Customize agent` instead of hiding
- Change `Delete agent` to non-destructive design

#### For code changes:
- [ ] I have made a test plan
  - [ ] ...

### DIFF
--- a/autogpt_platform/frontend/src/app/(platform)/library/agents/[id]/page.tsx
+++ b/autogpt_platform/frontend/src/app/(platform)/library/agents/[id]/page.tsx
@@ -268,22 +268,22 @@ export default function AgentRunsPage(): React.ReactElement {
 
   const agentActions: ButtonAction[] = useMemo(
     () => [
-      ...(agent?.can_access_graph
+      {
+        label: "Customize agent",
+        href: `/build?flowID=${agent?.graph_id}&flowVersion=${agent?.graph_version}`,
+        disabled: !agent?.can_access_graph,
+      },
+      { label: "Export agent to file", callback: downloadGraph },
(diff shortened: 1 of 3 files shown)


autogpt_platform/frontend/src/app/(platform)/library/agents/[id]/page.tsx
```diff
@@ -268,22 +268,22 @@ export default function AgentRunsPage(): React.ReactElement {
 
   const agentActions: ButtonAction[] = useMemo(
     () => [
-      ...(agent?.can_access_graph
+      {
+        label: "Customize agent",
+        href: `/build?flowID=${agent?.graph_id}&flowVersion=${agent?.graph_version}`,
+        disabled: !agent?.can_access_graph,
```
Comment (majdyz) [inline autogpt_platform/frontend/src/app/(platform)/library/agents/[id]/page.tsx:274]: Why don't we hide this instead of disabling?

majdyz (approved):

autogpt_platform/frontend/src/app/(platform)/library/agents/[id]/page.tsx
Comment (kcze) [reply autogpt_platform/frontend/src/app/(platform)/library/agents/[id]/page.tsx:274]: Design decision but it makes sense because user can see the `Customize agent` button and then in the copy modal there's "click on the Customize agent button".

Issue #9883, opened 2025-04-26
Send Web Request block errors on 204 'No Content' response (Discord Webhook)


Torantulino:
**Summary:**
The *Send Web Request* block in the AutoGPT Platform throws an exception when it receives an HTTP 204 No Content response (for example, when posting to a Discord webhook). The exception is:

```
Expecting value: line 1 column 1 (char 0)
```

The webhook POST *does* succeed, but the AutoGPT flow halts because the block tries to parse the empty response body as JSON.

**How to Reproduce:**

1. Drag a *Send Web Request* block into a flow.
2. Configure with a Discord webhook POST:

   ```json
   {
     "url": "https://discord.com/api/webhooks/<id>/<token>",
     "method": "POST",
     "json_format": true,
     "body": { "content": "Hello from AutoGPT Platform" }
   }
   ```
3. Run the flow. The message posts to Discord, but the block crashes with the error above.

**Expected Behavior:**

* If the server returns 204 or any empty response, the block should handle this gracefully and return a success status, allowing the flow to continue. Ideally, the output pin should fire with a success/empty payload rather than halting execution.
* (Workaround: Turning `json_format` off avoids the crash, but is not ideal.)

**Impact:**

* Interoperability with webhooks and APIs returning 204 is broken.

**Request:**

* Update the *Send Web Request* block so that empty-body 204 responses do not throw an exception when parsing as JSON.

---

This bug report was submitted by a user and written up by an Autonomous Agent running on the AutoGPT Platform.

PR #9887, opened 2025-04-28
fix: handle empty 204 responses in SendWebRequestBlock

Lohith-11:
This PR fixes [Issue #9883](https://github.com/Significant-Gravitas/AutoGPT/issues/9883), where the SendWebRequestBlock crashes when receiving a 204 No Content response, such as when posting to a Discord webhook. The fix ensures that empty responses are handled gracefully, and the block does not crash.

### Changes 🏗️
- Added a check to handle empty HTTP responses (like 204 status) in SendWebRequestBlock
- Fallback to empty string or None if there is no response content
- Prevents server errors when parsing non-existent response bodies

#### For code changes:
- [x] I have made a test plan
  
  - [x] Send a POST request to an endpoint that returns 204 No Content
  - [x]  Confirm that SendWebRequestBlock handles it without crashing
  - [x] Confirm that regular 200 OK JSON responses still work

### DIFF
--- a/autogpt_platform/autogpt_libs/autogpt_libs/utils/synchronize.py
+++ b/autogpt_platform/autogpt_libs/autogpt_libs/utils/synchronize.py
@@ -31,7 +31,7 @@ class RedisKeyedMutex:
         try:
             yield
         finally:
-            if lock.locked():
+            if lock.locked() and lock.owned():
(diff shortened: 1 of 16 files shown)


Torantulino:
Resolves #9883

PR #9889, opened 2025-04-28
feat(backend): Improve cancel execution reliability

majdyz:
When an executor dies, an ongoing execution will not be retried and will just stuck in the running status.
This change avoids such a scenario by allowing an execution of an entry that is not in QUEUED status with the low-probability risk of double execution.

### Changes 🏗️

* Allow non-QUEUED status to be re-executed.
* Improve cleanup of node & graph executor.
* Make a cancellation request consumption a separate thread to avoid being blocked by other messages.
* Remove unused retry loop on the execution manager.

#### For code changes:
- [x] I have made a test plan
  
  - [x] Run agent, kill the server, re-run it, agent restarted.

### DIFF
--- a/autogpt_platform/backend/backend/data/execution.py
+++ b/autogpt_platform/backend/backend/data/execution.py
@@ -492,21 +492,12 @@ async def upsert_execution_output(
 async def update_graph_execution_start_time(
     graph_exec_id: str,
 ) -> GraphExecution | None:
-    count = await AgentGraphExecution.prisma().update_many(
-        where={
-            "id": graph_exec_id,
-            "executionStatus": ExecutionStatus.QUEUED,
-        },
+    res = await AgentGraphExecution.prisma().update(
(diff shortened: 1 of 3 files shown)


ntindle:
Does this mean if someone manages to bring down an executor via a dos, it will bring down other executors next?

autogpt_platform/backend/backend/data/execution.py
```diff
@@ -492,21 +492,12 @@ async def upsert_execution_output(
 async def update_graph_execution_start_time(
     graph_exec_id: str,
 ) -> GraphExecution | None:
-    count = await AgentGraphExecution.prisma().update_many(
```
ntindle, line 495: pausing auto merge until q is answered

ntindle (approved):

majdyz:
> Does this mean if someone manages to bring down an executor via a dos, it will bring down other executors next?
This is correct

PR #9904, opened 2025-05-01
feat(backend): Move Scheduler

ntindle:
We want the scheduler shouldn't scale with the rest API lol

### Changes 🏗️
pulls out the scheduler into its own service

#### For code changes:
- [x] I have made a test plan
  
  - [x] test it


autogpt_platform/docker-compose.platform.yml
```diff
@@ -187,6 +188,61 @@ services:
...
+    depends_on:
+      db:
+        condition: service_healthy
+      redis:
+        condition: service_healthy
+      rabbitmq:
+        condition: service_healthy
+      migrate:
+        condition: service_completed_successfully
+    # healthcheck:
```
majdyz, line 211: health_check doesn't work ?

majdyz (approved):

autogpt_platform/docker-compose.platform.yml
ntindle: No it kept it from starting in the docker compose but not sure why

ntindle: Ah the db manager is needed but they are dependent on eachother's containers. Notifications requires this, this requires db manager

PR #9922, opened 2025-05-08
fix(frontend/library): Quick Patch for Rendering Agent Outputs

Torantulino:
The goal of this change is a quick and temporary tweak to improve the displaying of output text in the Agent Runs screen.

This change is made anticipating that these outputs will be properly improved in the near future, and is thus just a temporary change in order to display text in a human readable format.  

### Changes 🏗️
There is one change in this PR:
- The class of the Agent Output textbox is changed to properly display text without impacting the design.
Below is a before and after of this change:

**Before**
image

**After**
image

#### For code changes:
- [x] I have made a test plan
  
  - [ ] ...


Bentlybro (approved):
Tested and works

PR #9925, opened 2025-05-09
feat(backend): Immediate alert on late execution check job failure

majdyz:
Currently, there is no guarantee that an error will be reported right away. And late execution is a serious issue that needs to be addressed quickly

### Changes 🏗️

Provided a direct alert when late execution occurs.

#### For code changes:
- [x] I have made a test plan
  - [x] Manual run on the late executions job on artificially created late executions.


Bentlybro (approved):
Looks good to me

PR #9933, opened 2025-05-13
refactor(frontend): Move from remote loaded GTM to local

ntindle:
As part of a small security review, we found that google won't let you load with integrity. That seems insane but the general workaround is to load a static copy of your own

### Changes 🏗️
Moves all the google analytics scripts in house instead of loading via cdn

#### For code changes:
- [x] I have made a test plan
  
  - [x] tested it in a new isolated environment to confirm events still work as expected everywhere they are used


autogpt_platform/frontend/src/components/analytics/google-analytics.tsx
```diff
@@ -0,0 +1,62 @@
+"use client";
+
+import { useEffect } from "react";
+import Script from "next/script";
+import type { GAParams } from "../../types/google";
+
+let currDataLayerName: string | undefined = undefined;
```
Pwuts, line 12: Please make this not a global variable, or add a comment explaining why it has to be

autogpt_platform/frontend/src/components/analytics/google-analytics.tsx
```diff
@@ -0,0 +1,62 @@
+"use client";
+
+import { useEffect } from "react";
+import Script from "next/script";
+import type { GAParams } from "../../types/google";
```
Pwuts:
```suggestion
import type { GAParams } from "@/types/google";
```

autogpt_platform/frontend/src/components/tutorial.ts
```diff
@@ -1,6 +1,6 @@
-import { sendGAEvent } from "@next/third-parties/google";
 import Shepherd from "shepherd.js";
 import "shepherd.js/dist/css/shepherd.css";
+import { sendGAEvent } from "./analytics/google-analytics";
```
Pwuts:
```suggestion
import { sendGAEvent } from "@/components/analytics/google-analytics";
```

autogpt_platform/frontend/src/types/google.ts
```diff
@@ -0,0 +1,27 @@
+type JSONValue =
+  | string
+  | number
+  | boolean
+  | JSONValue[]
+  | {
+      [key: string]: JSONValue;
+    };
+
+export type GTMParams = {
```
Pwuts, line 15: "GTM" isn't self-explanatory to me, could you expand the acronym or add a comment? Same for "GA" below.

autogpt_platform/frontend/src/components/analytics/google-analytics.tsx
ntindle: This is all pulled direct from next JS except loading the file location which is now local

autogpt_platform/frontend/src/types/google.ts
ntindle: No we are keeping the exact same api so it can be copy pasted

autogpt_platform/frontend/src/components/tutorial.ts
ntindle: We are keeping the same API so it can be copy pasted

autogpt_platform/frontend/src/components/analytics/google-analytics.tsx
ntindle: We are keeping the same api so it can be copy pasted

Pwuts (approved):

PR #9935, opened 2025-05-13
fix(frontend): Re-subscribe on WebSocket re-connect

### LINKED ISSUE #9929: Resubscribe through websocket after reconnecting
The frontend can sometimes lose its websocket connection. It *will* automatically re-connect, but not re-subscribe, causing the UI to no longer update with e.g. execution updates from the backend.

## TODO

1. When the frontend's websocket connection is re-connected, its subscriptions have to be re-subscribed.
2. Additionally, the state that would otherwise have been updated through WS updates has to be refreshed through the REST API, because we have to assume we may have missed WS updates while the connection was broken.

Pwuts:
- Resolves #9929

### Changes 🏗️

- Implement `BackendAPI.onWebSocketConnect(..)`

- Improve reliability and reactivity of `/library/agents/[id]`:
  - Refresh page data and (re)subscribe and on WebSocket (re)connect
  - Break up multi-action hooks into smaller parts to reduce unnecessary re-renders and requests
  - Reduce duplicate requests

- Use `onWebSocketConnect` in `useAgentGraph` as well

- Tidy up `autogpt-server-api/client.ts` a bit

#### For code changes:
- [x] I have made a test plan
  - Go to `/library/agents/[id]`
  - Run the agent
    - [x] -> UI should update normally with execution updates
  - Suspend your computer, or restart the backend (or WS server) to break the connection
    - DO NOT REFRESH THE TAB ITSELF
    - [x] -> On reconnect, page data should be refreshed (check in network tab of dev tools)
  - Run the agent again
    - [x] -> UI should update normally with execution updates

### DIFF
--- a/autogpt_platform/frontend/src/app/(platform)/library/agents/[id]/page.tsx
+++ b/autogpt_platform/frontend/src/app/(platform)/library/agents/[id]/page.tsx
@@ -140,6 +140,7 @@ export default function AgentRunsPage(): React.ReactElement {
     }
   }, [api, agentID, getGraphVersion, graph, selectedView, isFirstLoad, agent]);
 
+  // Initial load
   useEffect(() => {
     fetchAgents();
   }, []);
@@ -148,9 +149,13 @@ export default function AgentRunsPage(): React.ReactElement {
   useEffect(() => {
     if (!agent) return;
 
-    // Subscribe to all executions for this agent
-    api.subscribeToGraphExecutions(agent.graph_id);
-  }, [api, agent]);
+    return api.onWebSocketConnect(() => {
+      fetchAgents(); // Sync up on (re)connect
+
+      // Subscribe to all executions for this agent
+      api.subscribeToGraphExecutions(agent.graph_id);
+    });
+  }, [api, agent, fetchAgents]);
 
   // Handle execution updates
   useEffect(() => {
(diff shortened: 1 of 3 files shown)


autogpt_platform/frontend/src/app/(platform)/library/agents/[id]/page.tsx
```diff
@@ -179,23 +197,24 @@ export default function AgentRunsPage(): React.ReactElement {
...
     }
+
+    // Load/refresh the view, even if selected view doesn't change
+    api
+      .getGraphExecutionInfo(agent.graph_id, selectedView.id)
+      .then(async (run) => {
+        // Ensure corresponding graph version is available before rendering I/O
+        await getGraphVersion(run.graph_id, run.graph_version);
+        setSelectedRun(run);
+      });
```
Comment (Pwuts) [inline autogpt_platform/frontend/src/app/(platform)/library/agents/[id]/page.tsx]: Gotta check that this doesn't cause the page to spazz when receiving execution updates

autogpt_platform/frontend/src/app/(platform)/library/agents/[id]/page.tsx
```diff
@@ -77,34 +84,43 @@ export default function AgentRunsPage(): React.ReactElement {
     setSelectedSchedule(schedule);
   }, []);
 
-  const [graphVersions, setGraphVersions] = useState<Record<number, Graph>>({});
+  const graphVersions = useRef<Record<number, Graph>>({});
```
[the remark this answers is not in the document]
Comment (Pwuts) [reply autogpt_platform/frontend/src/app/(platform)/library/agents/[id]/page.tsx:87]: It's because `useState` only propagates *at the next render*, whereas with `useRef` any changes are instantaneous which is what I need for the caching mechanism to work properly.

autogpt_platform/frontend/src/app/(platform)/library/agents/[id]/page.tsx
Comment (Pwuts) [reply autogpt_platform/frontend/src/app/(platform)/library/agents/[id]/page.tsx]: it was; fixed

ntindle (approved):

PR #9940, opened 2025-05-14
fix(backend): Make URL pinning work with `extra_url_validator`

ntindle:
Github Blocks use an URL transformer passed to `Requests` to convert web URLs to the API URLs. This doesn't always work with the anti-SSRF URL pinning mechanism that was implemented in #8531.

### Changes 🏗️
In `Requests.request(..)`:
- Apply `validate_url` *after* `extra_url_validator`, to prevent mismatch between `pinned_url` and `original_hostname`
- Simplify logic & add clarifying comments

#### For code changes:
- [x] I have made a test plan
  - [x] Tested the github blocks that had the issue

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -32,9 +32,9 @@
       "type": "debugpy",
       "request": "launch",
       "module": "backend.app",
-      // "env": {
-      //   "ENV": "dev"
-      // },
+      "env": {
+        "OBJC_DISABLE_INITIALIZE_FORK_SAFETY": "YES"
+      },
       "envFile": "${workspaceFolder}/backend/.env",
       "justMyCode": false,
       "cwd": "${workspaceFolder}/autogpt_platform/backend"
--- a/autogpt_platform/backend/backend/util/request.py
+++ b/autogpt_platform/backend/backend/util/request.py
@@ -228,8 +228,8 @@ class Requests:
         if self.extra_url_validator is not None:
             url = self.extra_url_validator(url)
 
-        # Validate URL and get pinned URL + original hostname
-        pinned_url, original_hostname = validate_url(url, self.trusted_origins)
+        # Validate URL and get pinned URL + hostname
+        pinned_url, hostname = validate_url(url, self.trusted_origins)
 
         # Merge any extra headers
         headers = dict(headers) if headers else {}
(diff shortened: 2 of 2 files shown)


.vscode/launch.json
```diff
@@ -32,9 +32,9 @@
       "type": "debugpy",
       "request": "launch",
       "module": "backend.app",
-      // "env": {
-      //   "ENV": "dev"
-      // },
+      "env": {
+        "OBJC_DISABLE_INITIALIZE_FORK_SAFETY": "YES"
+      },
```
Bentlybro, line 37: is this actually needed and what does it do?

ntindle: TLDR on Mac lets you spy on other processes forked from the main one, which the debugger needs. We changed how we fork a while back, so this is now required to use the debugger

autogpt_platform/backend/backend/util/request.py
```diff
@@ -237,7 +237,10 @@ def request(
             headers.update(self.extra_headers)
 
         # Force the Host header to the original hostname
-        headers["Host"] = original_hostname
+        if urlparse(pinned_url).hostname not in self.trusted_origins:
+            headers["Host"] = original_hostname
+        else:
+            headers["Host"] = urlparse(pinned_url).hostname
```
Pwuts: nit:
```suggestion
        if (pinned_hostname := urlparse(pinned_url).hostname) not in self.trusted_origins:
            headers["Host"] = original_hostname
        else:
            headers["Host"] = pinned_hostname
```

autogpt_platform/backend/backend/util/request.py
```diff
@@ -237,7 +237,10 @@ def request(
             headers.update(self.extra_headers)
 
         # Force the Host header to the original hostname
-        headers["Host"] = original_hostname
+        if urlparse(pinned_url).hostname not in self.trusted_origins:
```
Pwuts: could you update this comment, to clarify what this bit of logic does now and why?

ntindle: Sorry I thought it was clear

Is 
```suggestion
# force the header host to be in trusted hostnames or overwrite to the original
```
better

Pwuts (approved):
This needs to be included in the release, so I guess it's time for one of these:
obama gives himself a medal (meme)

PR #9952, opened 2025-05-16
fix(backend): Graph execution update on terminate

### LINKED ISSUE #9947: Execution status not updated in UI after terminating a run
After clicking "Stop run", the run is stopped in the backend, but its status in the UI is not updated. Refreshing the page shows it is "Stopped".

The run termination mechanism probably isn't pushing execution updates.

Pwuts:
- Resolves #9947

### Changes 🏗️

Backend:
- Send a graph execution update after terminating a run
- Don't wipe the graph execution stats when not passed in to `update_graph_execution_stats`

Frontend:
- Don't hide the output of stopped runs

#### For code changes:
- [x] I have made a test plan
  - Go to `/library/agents/[id]`
  - Run an agent that takes a while (long enough to click stop and see the effect)
  - Hit stop after it has executed a few nodes
  - [x] -> run status should change to "Stopped"
  - [x] -> run stats (steps, duration, cost) should stay the same or increase only one last time
  - [x] -> output so far should be visible
  - [x] -> shown information should stay the same after refreshing the page


autogpt_platform/backend/backend/data/execution.py
```diff
@@ -572,9 +573,15 @@ async def update_graph_execution_stats(
...
-        data["error"] = str(data["error"])
+    update_data: AgentGraphExecutionUpdateManyMutationInput = {
+        "executionStatus": status
+    }
+
+    if stats:
+        _stats = stats.model_dump()
+        if isinstance(_stats.get("error"), Exception):
+            _stats["error"] = str(_stats["error"])
+        update_data["stats"] = Json(_stats)
```
kcze: nit, `_` prefix has special meaning and doesn't really fit here, so if not `data` like before maybe:
```suggestion
        stats_dict = stats.model_dump()
        if isinstance(stats_dict.get("error"), Exception):
            stats_dict["error"] = str(stats_dict["error"])
        update_data["stats"] = Json(stats_dict)
```

kcze (dismissed):
LGTM

autogpt_platform/backend/backend/data/execution.py
Pwuts: what special meaning does it have?

PR #9954, opened 2025-05-16
fix(blocks): disable Twitter and Todoist blocks if their OAuth is not configured

Abhi1992002:
I am disabling all the Twitter and Todoist blocks whose OAuth is not configured.

> I have already checked it locally. When OAuth is not set, the blocks do not appear in the block menu

### DIFF
--- a/autogpt_platform/backend/backend/blocks/todoist/tasks.py
+++ b/autogpt_platform/backend/backend/blocks/todoist/tasks.py
@@ -7,6 +7,7 @@ from typing_extensions import Optional
 from backend.blocks.todoist._auth import (
     TEST_CREDENTIALS,
     TEST_CREDENTIALS_INPUT,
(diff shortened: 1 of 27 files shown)


autogpt_platform/backend/backend/blocks/todoist/tasks.py
```diff
@@ -467,32 +472,24 @@ def run(
...
+                "priority": input_data.priority,
+                "due_date": due_date,
+                "deadline_date": deadline_date,
+                "assignee_id": input_data.assignee_id,
+                "duration": input_data.duration,
+                "duration_unit": input_data.duration_unit,
+            }
+
+            # Filter out None values
+            task_updates = {k: v for k, v in update_fields.items() if v is not None}
```
Pwuts, line 492: also nice :)

Pwuts (approved):

PR #9959, opened 2025-05-17
Add Gmail thread blocks and docs

Torantulino:
## Summary
- extend Gmail Read output with `threadId`
- add Gmail Get Thread block
- add Gmail Reply block with attachment handling
- document new Gmail blocks and quickstart flow

## Testing
- `poetry run python run_tests.py` *(no output)*


docs/content/platform/quickstarts/gmail-detect-reply.md
```diff
@@ -0,0 +1,10 @@
+# Detect replies to a scheduling email
```
ntindle: Is this actually linked to where it’s needed to be

PR #10003, opened 2025-05-22
dx(backend): Disable pre-commit `pytest` hooks

Pwuts:
Running the tests locally takes a lot of time and leaves test data behind in the DB, making it impractical to actually run locally.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -241,38 +241,38 @@ repos:
         language: system
         pass_filenames: false
 
-  - repo: local
-    hooks:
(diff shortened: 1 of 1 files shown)


AutoGPT-Agent:
The PR fails to meet the basic requirements. While the change itself is simple (disabling pytest hooks), the PR template is not properly filled out. The PR's purpose is explained, but the template sections are completely missing rather than being properly marked as not applicable. The PR title follows the conventional commit format with a valid scope (backend), but the description should still maintain the structure of the template even if sections are not relevant.

PR #10010, opened 2025-05-22
feat(frontend): Redirect from www. to root domain

Swiftyos:
## Summary
- redirect `www` subdomains to the root domain in middleware

## Testing
- `yarn lint`
- `yarn test` *(fails: Process from config.webServer exited early)*


majdyz (approved):
Why are we doing this ?

Swiftyos:
It was a linear ticket, but I dont think it makes sense

PR #10066, opened 2025-05-30
feat(executor;frontend): Move output processing step from node executor to graph executor & simplify input beads calculation

majdyz:
**Goal**
Allow parallel runs within a single node. Currently, we prevent this to avoid unexpected ordering of the execution.

#### Executor changes

We decoupled the node execution output processing, which is responsible for deciding the next executions from the node executor code.

Currently, `execute_node` does two big things:
* Runs the block’s execute(...) (which yields outputs).
* immediately enqueues the next nodes based on those outputs.

This PR makes:
* execute_node(node_exec) -> stream of (output_name, data). That purely runs the block and yields each output as soon as it’s available.
* Move _enqueue_next_nodes into the graph executor. So the next execution is handled serially by the graph executor to avoid concurrency issues.

#### UI changes

The change on the executor also fixes the behavior of the execution update to the UI We will report the execution output to the UI as soon as it is available, not when the node execution is fully completed. This, however, broke the bread calculation logic that assumes each execution update will never overlap. So the change in this PR makes the bead calculation take the overlap / duplicated execution update into account, and simplify the overall calculation logic.

#### For code changes:
- [x] I have made a test plan
  
  - [x] Execute this agent and observe its concurrency ordering
  
image

### DIFF
--- a/autogpt_platform/frontend/src/components/CustomEdge.tsx
+++ b/autogpt_platform/frontend/src/components/CustomEdge.tsx
@@ -12,6 +12,7 @@ import "./customedge.css";
 import { X } from "lucide-react";
 import { useBezierPath } from "@/hooks/useBezierPath";
 import { FlowContext } from "./Flow";
+import { NodeExecutionResult } from "@/lib/autogpt-server-api";
 
 export type CustomEdgeData = {
   edgeColor: string;
@@ -19,7 +20,7 @@ export type CustomEdgeData = {
   isStatic?: boolean;
   beadUp?: number;
   beadDown?: number;
-  beadData?: any[];
+  beadData?: Map<string, NodeExecutionResult["status"]>;
 };
 
 type Bead = {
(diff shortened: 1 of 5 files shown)


autogpt_platform/frontend/src/components/CustomEdge.tsx
```diff
@@ -12,14 +12,15 @@ import "./customedge.css";
...
+import { NodeExecutionResult } from "@/lib/autogpt-server-api";
 
 export type CustomEdgeData = {
   edgeColor: string;
   sourcePos?: XYPosition;
   isStatic?: boolean;
   beadUp?: number;
   beadDown?: number;
-  beadData?: any[];
+  beadData?: Map<string, NodeExecutionResult["status"]>;
```
ntindle, line 23: it would be cool to hover a bead and get its details

majdyz: I agree, @kcze do you think it's feasible?
I'll log an issue anyway.

PR #10079, opened 2025-06-02
chore(backend): Downgrade poetry to 2.1.1 for dependabot

Bentlybro:
Downgrades poetry to 2.1.1 for dependabot by re-locking the poetry.lock as 2.1.1 and setting

```
poetry = "^2.1.1"
```
in [autogpt_platform/backend/pyproject.toml](https://github.com/Significant-Gravitas/AutoGPT/compare/dev...downgrade-poetry-to-2.1.1?expand=1#diff-26ebebd91da791c6484f07d9d91484a66f52836708f5294b24365603438b880c)

Checklist 📋
For code changes:

- [x] I have made a test plan
- [ ] Not sure what to test


AutoGPT-Agent:
Thanks for your PR to downgrade Poetry to 2.1.1 for Dependabot compatibility!

Your PR description clearly explains the changes, but I noticed your checklist is incomplete. You've marked "Not sure what to test" instead of completing a test plan.

Even for dependency version changes like this, we need a complete test plan. In this case, the test plan could be as simple as:

- [ ] Verified the backend builds successfully with Poetry 2.1.1
- [ ] Ran basic backend operations to ensure functionality is maintained

Please update your PR description with a completed test plan and check it off to meet our requirements for merging.

.github/workflows/platform-backend-ci.yml
```diff
@@ -32,7 +32,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.10"]
+        python-version: ["3.11"]
```
Pwuts, line 35: need this to use `tomllib` in the script

Bentlybro: and that only supports python 3.11?

Pwuts:
CI doesn't allow *downgrading* Poetry, which for this one time we're going to do anyway and ignore CI's complaints about it

PR #10081, opened 2025-06-03
fix(storybook): make font load in Stories

0ubbe:
#### Before

Screenshot 2025-06-03 at 16 54 36

#### After

Screenshot 2025-06-03 at 16 58 38

#### For code changes:
- [x] I have made a test plan
  - [ ] checkout this branch ( _we should have PR previews for the app and Storybook_ )
  - [ ] `cd autogpt_platform/frontend`
  - [ ] `yarn storybook`
  - [ ] the stories road with the right font ( Poppins ) not a serif one 😄  

- [ ] ~~`.env.example` is updated or already compatible with my changes~~
- [ ] ~~`docker-compose.yml` is updated or already compatible with my changes~~
- [ ] ~~I have included a list of my configuration changes in the PR description (under **Changes**)~~


0ubbe:
@kcze when you have a moment to review 🙏🏽

autogpt_platform/frontend/.storybook/manager-head.html
```diff
@@ -0,0 +1,31 @@
...
+  body,
+  #storybook-root {
+    font-family:
+      var(--font-poppins),
+      system-ui,
+      -apple-system,
+      sans-serif;
+  }
+
+  .font-sans {
```
Abhi1992002: Here we are using the Poppins font as font-sans, but we need to use the Geist font instead.

More information about the current font we are using - https://github.com/Significant-Gravitas/AutoGPT/blob/master/autogpt_platform/frontend/tailwind.config.ts

autogpt_platform/frontend/.storybook/preview.tsx
```diff
@@ -0,0 +1,41 @@
...
+        date: /Date$/i,
+      },
+    },
+  },
+  loaders: [mswLoader],
+  decorators: [
+    (Story) => (
+      <div
+        className="font-sans"
+        style={
```
Abhi1992002: Maybe you could use Next.js for font loading instead of inline styles — something like this :
```tsx
import React from "react";
import type { Preview } from "@storybook/react";
import { initialize, mswLoader } from "msw-storybook-addon";
import { Inter, Poppins } from "next/font/google";
import localFont from "next/font/local";
import "../src/app/globals.css";
import { Providers } from "../src/app/providers";

const poppins = Poppins({
  subsets: ["latin"],
  weight: ["400", "500", "600", "700"],
  variable: "--font-poppins",
});

const inter = Inter({ subsets: ["latin"], variable: "--font-inter" });

const GeistSans = localFont({
  src: "../fonts/geist-sans/Geist-Variable.woff2",
  variable: "--font-geist-sans",
});

const GeistMono = localFont({
  src: "../fonts/geist-mono/GeistMono-Variable.woff2",
  variable: "--font-geist-mono",
});

// Initialize MSW
initialize();

const preview: Preview = {
  parameters: {
    nextjs: {
      appDirectory: true,
    },
    controls: {
      matchers: {
        color: /(background|color)$/i,
        date: /Date$/i,
      },
    },
    layout: "fullscreen",
  },
  decorators: [
    (Story, context) => {
      const mockOptions = context.parameters.mockBackend || {};

      return (
        <div
          className={`${poppins.variable} ${inter.variable} ${GeistMono.variable} ${GeistSans.variable}`}
        >
          <Providers mockClientProps={mockOptions}>
            <Story />
          </Providers>
        </div>
      );
    },
  ],
  loaders: [mswLoader],
};

export default preview;
```

Abhi1992002: Are there any benefits to using inline CSS for fonts instead of Next.js font loading?

Abhi1992002: If you use Next.js font loading, don’t forget to add the Geist fonts to the static directory.

```tsx
  staticDirs: [
    "../public",
    {
      from: "../node_modules/geist/dist/fonts/geist-sans",
      to: "/fonts/geist-sans",
    },
    {
      from: "../node_modules/geist/dist/fonts/geist-mono",
      to: "/fonts/geist-mono",
    },
  ],
  ```

**Review by Abhi1992002 — CHANGES_REQUESTED** [review]

**Comment by Abhi1992002** [comment]
While I was fixing this error, I noticed another issue.

When I open an extra component inside Storybook — for example, clicking on a popover — the Popover content doesn’t seem to have any font applied.

Could you please check if this happens only with Next.js font loading, or is it also happening with inline CSS fonts?

You can test font loading in an external modal by checking the BecomeACreator component.

File: autogpt_platform/frontend/.storybook/preview.tsx (hunk 2 above)
Comment (Abhi1992002) [reply autogpt_platform/frontend/.storybook/preview.tsx]: my previous pr on this issue - https://github.com/Significant-Gravitas/AutoGPT/pull/9859

Comment (0ubbe) [reply autogpt_platform/frontend/.storybook/preview.tsx]: I don’t believe `next/font` will work in Storybook out of the box, since Storybook 8 uses its own Vite config:

Screenshot 2025-06-04 at 15 34 57

Because `next/font` relies on Next.js’s build system, it’s better to keep Storybook as framework‐agnostic as possible. That way we avoid extra Webpack tweaks and ensure components render consistently in both environments 💆🏽 💭 

Before introducing mocking, providers, or any other complexities, we should first clean up Storybook with a minimal setup that reflects our component library/design system 🙏🏽 .

Ideally, Storybook is used initially to document our design system ( _components that are agnostic of APIs and services_ ). Start small and build slowly. 

For component testing, we should be using Vitest in browser mode (and add MSW there), rather than over-complicating Storybook’s setup.

Comment (Abhi1992002) [reply autogpt_platform/frontend/.storybook/preview.tsx]: Yeah, makes sense! Sounds good to me.

File: autogpt_platform/frontend/src/components/styles/fonts.css [hunk 3]
```diff
@@ -0,0 +1,30 @@
...
+}
+
+/* CSS Variables matching config from fonts.ts */
+:root {
+  --font-poppins: "Poppins", sans-serif;
+  --font-geist-sans: "Geist", ui-sans-serif, system-ui, sans-serif;
+  --font-geist-mono:
+    "GeistMono", ui-monospace, "Cascadia Code", "Source Code Pro", Menlo,
+    Consolas, monospace;
+}
```
0ubbe, line 30: ### Simplified the setup 💭

I intend to make `components/` folder hold exclusively our design system/UI elements. 

For now, font configuration lives here as well: `fonts.ts` for Next.js and a separate `fonts.css` for Storybook. This small duplication avoids any need for build scripts or complex Storybook config tweaks.

Next, I’ll suggest moving page-specific components into the `app/` folder (per the [Next.js docs](https://nextjs.org/docs/app/getting-started/project-structure#split-project-files-by-feature-or-route)). That will give us a clear separation between design-system components and page-specific components (the later can then import services, do data fetching, etc... ).

0ubbe:
> When I open an extra component inside Storybook — for example, clicking on a popover — the Popover content doesn’t seem to have any font applied.

@Abhi1992002 should be working on popover etc...

Screenshot 2025-06-04 at 16 45 47
Screenshot 2025-06-04 at 16 45 58

Simplified the config quite a bit 👍🏽 , could you have another look ? 

> You can test font loading in an external modal by checking the BecomeACreator component.

This one does not at all in the story:

Screenshot 2025-06-04 at 17 34 09

Because it needs the Back-end API context. However, as I mentioned in the earlier comment, for now, we should primarily use Storybook to document our design system components cleanly, and wrangle the more complex ones later, making API calls (we need to evaluate the cost versus benefits of doing that).

PR #10101, opened 2025-06-04
chore(backend): Tell Dependabot not to update Poetry

### LINKED ISSUE #10098: Update dependabot for ignoring poetry
https://github.com/Significant-Gravitas/AutoGPT/blob/master/.github/dependabot.yml

https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/controlling-dependencies-updated#ignoring-specific-dependencies

we need to ignore poetry so that dependabot doesn't try and update it and break the world

ntindle:
- Fixes #10098

### Changes 🏗️
Tells dependabot to ignore poetry

### Checklist 📋
N/A

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,6 +10,8 @@ updates:
     commit-message:
       prefix: "chore(libs/deps)"
       prefix-development: "chore(libs/deps-dev)"
+    ignore:
+      - dependency-name: "poetry"
     groups:
       production-dependencies:
         dependency-type: "production"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -10,6 +10,8 @@ updates:
     commit-message:
       prefix: "chore(libs/deps)"
       prefix-development: "chore(libs/deps-dev)"
+    ignore:
+      - dependency-name: "poetry"
```
Pwuts, line 14: I don't think we need this one

ntindle: better safe tan sorry

Pwuts:
Example of the problem: #10096

PR #10110, opened 2025-06-05
fix(backend): Fix graph fetching of non-owned marketplace agent

majdyz:
Currently, the get_graph function, with no graph version specifier will try to fetch the latest version, and when the graph is not owned and the latest version is not available for listing, it will return `None` instead of picking the latest graph version available on the store.

### Changes 🏗️

Instead of using the latest graph.version to fetch the store listing, don't provide any version filter at all and pick up whatever available version in the store.

#### For code changes:
- [x] I have made a test plan
  
  - [x] CI, existing tests

### DIFF
--- a/autogpt_platform/backend/backend/data/graph.py
+++ b/autogpt_platform/backend/backend/data/graph.py
@@ -12,6 +12,7 @@ from prisma.types import (
     AgentGraphWhereInput,
     AgentNodeCreateInput,
     AgentNodeLinkCreateInput,
+    StoreListingVersionWhereInput,
 )
 from pydantic import create_model
 from pydantic.fields import computed_field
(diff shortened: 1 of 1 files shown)


AutoGPT-Agent:
Thanks for fixing this issue with graph fetching for non-owned marketplace agents. The approach of removing the version filter when none is specified makes sense and should solve the described problem.

However, I noticed something that looks like a potential issue in your implementation:

On line 734, there's a condition `and not ()` which seems to be a leftover from the previous code. This empty condition will always evaluate to `True` (since `not ()` is `True`), making the second part of the condition redundant since the first part `graph is None` would already handle that case. 

I think the intention was to remove this second condition entirely since you've already handled the store listing check above in the new implementation.

Could you please review this and update the code accordingly before we merge? Otherwise, the implementation looks good and should fix the issue you described.

AutoGPT-Agent:
Thanks for the PR! The fix for non-owned marketplace agent graph fetching looks good overall. 

The approach of not filtering by version when checking store listings makes sense and should solve the problem you described.

However, I noticed something that appears to be an oversight:

On line 733, there's a condition `(graph is None or (graph.userId != user_id and not ()))` that looks like leftover code from your refactoring. The `and not ()` part doesn't do anything and should be removed, since empty tuples are truthy in Python. 

Also, since you've already checked for `graph is None` on lines 716-717 and returned early, that part of the condition on line 733 is redundant too.

Once this is cleaned up, the PR should be good to go!

Abhi1992002 (approved):

PR #10120, opened 2025-06-05
fix(frontend): hydration warning

0ubbe:
We're encountering a hydration warning on the frontend due to a mismatch in CSS module hashes. This happens because auto-generated classnames from `next/font` and `geist` differ between server-side and client-side rendering. The inconsistency triggers a warning when the client rehydrates the server-rendered HTML.

Screenshot 2025-06-05 at 17 20 06

### Changes 🏗️

Since the mismatch only affects the `<html>` tag and has no visible impact on the UI, the most straightforward workaround is to suppress the warning and still take advantage of `next/font` optimisations.

#### For code changes:
- [x] I have made a test plan
  - [x] Run the frontend locally
  - [x] Page loads without hydration warnings


autogpt_platform/frontend/.npmrc
```diff
@@ -0,0 +1,5 @@
+# Configure pnpm to save exact versions
+save-exact=true
+
+# Disable hoisting for better dependency isolation
+hoist=false
```
0ubbe, line 5: We need to pin the dependencies so we avoid mismatch also in minor versions ...  This should help with it.

autogpt_platform/frontend/.npmrc
```diff
@@ -0,0 +1,2 @@
+# Configure pnpm to save exact versions
+save-exact=true
```
0ubbe: - `save-exact=true` ensures that all dependencies are pinned to exact versions (including patch and minor). This helps prevent unexpected version mismatches across environments and CI

0ubbe: Also addressed in a separate PR https://github.com/Significant-Gravitas/AutoGPT/pull/10121

0ubbe:
@Abhi1992002 this should be good now 💅🏽

autogpt_platform/frontend/src/hooks/useTurnstile.ts
```diff
@@ -47,10 +47,13 @@ export function useTurnstile({
...
     const hasTurnstileKey = !!TURNSTILE_SITE_KEY;
-    const turnstileDisabled = process.env.NEXT_PUBLIC_DISABLE_TURNSTILE === "true";
+    const turnstileDisabled =
+      process.env.NEXT_PUBLIC_DISABLE_TURNSTILE === "true";
 
     // Only render Turnstile in cloud environment if not explicitly disabled
-    setShouldRender(behaveAs === BehaveAs.CLOUD && hasTurnstileKey && !turnstileDisabled);
+    setShouldRender(
+      behaveAs === BehaveAs.CLOUD && hasTurnstileKey && !turnstileDisabled,
+    );
```
0ubbe: Formatting issue. See https://github.com/Significant-Gravitas/AutoGPT/pull/10121#discussion_r2129065449

Abhi1992002:
@0ubbe, there are some conflicts now. Can you fix them?

Pwuts (commented):
The PR contains a lot of changes that are not covered by the description or title

0ubbe:
> The PR contains a lot of changes that are not covered by the description or title

@Pwuts rebased, is only 3 files changed which in my view relate to what is stated in the PR title/description 💭 
@Abhi1992002 rebased 👍🏽

PR #10125, opened 2025-06-06
fix(frontend): avoid Sentry initialisation and warnings on local dev

0ubbe:
## Changes 🏗️

We were getting the following warnings in the console when running the local server:

```
 ⚠ ./node_modules/.pnpm/@sentry+node@9.26.0/node_modules/@sentry/node/build/cjs/sdk
Package import-in-the-middle can't be external
The request import-in-the-middle matches serverExternalPackages (or the default list).
The request could not be resolved by Node.js from the project directory.
Packages that should be external need to be installed in the project directory, so they can be resolved from the output files.
Try to install it into the project directory by running npm install import-in-the-middle from the project directory.
```

### Why were the warnings happening?

The Sentry SDK for Next.js tries to hook into Node.js internals using packages like `import-in-the-middle` and `require-in-the-middle`. When Sentry is imported at the top level on a file (even if not enabled), it tries to load these dependencies... If they are not installed, then we get these warnings...

### Why does installing the packages fix it?

By installing `import-in-the-middle` and `require-in-the-middle` as dev dependencies, Sentry finds them and the warnings disappear. This is a safe workaround for local/dev, and does not affect production.

### Loading Sentry conditionally

One way to avoid these warnings ⚠️  is by loading Sentry conditionally. That is the approach I took in an earlier PR. However I realised that it would have to apply to any file importing Sentry:
```ts
```
which would end quite messy and affecting a lot of files. I realised installing the packages is just simpler ( _they won't in the bundle or affect page load_ ) and using `enabled` in the Sentry initialisation is also cleaner.  

### For code changes:
- [x] I have made a test plan
  - [x] There are not Sentry warnings when running the local dev server ( with or without `--turbo` )
  - [x] Sentry still reports issues in production or staging ( _but not locally_ )

### DIFF
--- a/autogpt_platform/frontend/instrumentation-client.ts
+++ b/autogpt_platform/frontend/instrumentation-client.ts
@@ -2,61 +2,64 @@
 // The config you add here will be used whenever a users loads a page in their browser.
 // https://docs.sentry.io/platforms/javascript/guides/nextjs/
 
-import { getEnvironmentStr } from "@/lib/utils";
+import { BehaveAs, getBehaveAs, getEnvironmentStr } from "@/lib/utils";
 import * as Sentry from "@sentry/nextjs";
 
-if (process.env.NODE_ENV === "production") {
-  Sentry.init({
-    dsn: "https://fe4e4aa4a283391808a5da396da20159@o4505260022104064.ingest.us.sentry.io/4507946746380288",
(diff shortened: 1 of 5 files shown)


0ubbe:
@Pwuts Sentry changes separate as we spoke 👍🏽 💯

PR #10138, opened 2025-06-10
feat(backend): Make execution engine async-first

majdyz:
This change introduced async execution for blocks and the execution engine. Paralellism will be achieved through a single process asynchronous execution instead of process concurrency.

### Changes 🏗️

* Support async execution for the graph executor
* Removed process creation for node execution
* Update all blocks to support async executions

#### For code changes:
- [x] I have made a test plan
  
  - [x] Manual graph executions, tested many of the impacted blocks/


Swiftyos (approved):
I got Opus to go through the change and I checked all run functions have been converted. They have ✅ 

However, Opus did find this -  which I'm adding here so it's documented as we should probably check all blocks use none blocking methods at some point. 

I don't think this should stop us merging this in though


  🚨 Blocking Operations in Async Methods:

  1. Google Sheets/Gmail/Calendar blocks - Using synchronous Google API calls:
  async def run(self, input_data: Input, **kwargs):
      result = service.spreadsheets().values().get(...).execute()  # Blocks!

  2. Code Executor blocks - Synchronous E2B Sandbox API calls:
  async def run(self, input_data: Input, **kwargs):
      sandbox = Sandbox(template="...")  # Synchronous initialization

  3. HTTP block - Synchronous file reading:
  async def run(self, input_data: Input, **kwargs):
      with open(file_path, 'rb') as f:  # Blocks!
          files = {'file': f}

  Recommended Fixes:

  1. For Google APIs: Use aiogoogle or wrap in asyncio.to_thread():
  result = await asyncio.to_thread(
      service.spreadsheets().values().get(...).execute
  )

  2. For file operations: Use aiofiles:
  async with aiofiles.open(file_path, 'rb') as f:
      content = await f.read()

  3. For E2B Sandbox: Check if they have async client, otherwise use thread pool.

  These blocking operations will significantly impact the performance benefits of the async executor, as they'll block the event loop and prevent concurrent
  execution of other blocks.

majdyz:
@Swiftyos Thanks for this! I have addressed those points.

Swiftyos (approved):

autogpt_platform/backend/backend/blocks/github/pull_requests.py
```diff
@@ -242,39 +245,39 @@ def __init__(self):
...
         for file in files:
-            filename = file.get("filename")
-            patch = file.get("patch")
-            if filename and patch:
-                changes.append(f"File: {filename}\n{patch}")
-        return "\n\n".join(changes)
+            filename = file.get("filename", "")
+            status = file.get("status", "")
+            changes.append(f"{filename}: {status}")
+        return "\n".join(changes)
```
Pwuts, line 271: sneaky and breaking :(

PR #10017, opened 2025-05-22
Fix credential refresh on user switch

Pwuts:
## Summary
- update Supabase hook to listen for auth changes
- refresh credentials whenever the logged in user changes

## Testing
- `npm run lint` *(fails: `next` not found)*
- `npm run type-check` *(fails: cannot find modules)*

PR #10143, opened 2025-06-10
fix(frontend): account menu still showing after logout

0ubbe:
### Changes 🏗️

Screenshot 2025-06-10 at 14 21 48

When logging out, I'm redirected to the `/login` page but the account menu is still visible (however I'm not longer logged out). Refreshing the page fixes it.

The problem was that `supabase.logOut()` is a client side action, and `<Navbar />` is a RSC who fetchs the session on the server to display the state and does not know on the client it was invalidated. `router.refresh()` solves the issue by forcing RSC on the page to refetch their state and update server side. Further reading [here](https://nextjs.org/docs/app/deep-dive/caching#invalidation-1).

I also improved the UX awaiting the promise and displaying a spinner while the log out action is happening. If logout fails ( _should be very rare_ ) I'm displaying a toast to not let the user be hanging wondering what happened.

#### For code changes:
- [x] I have made a test plan
  - [x] Login
  - [x] Open account menu
  - [x] Click `Logout`
  - [x] I see a spinner while the action is happening
  - [x] I'm redirected to `/login` and I no longer see the account menu

### DIFF
--- a/autogpt_platform/frontend/.storybook/main.ts
+++ b/autogpt_platform/frontend/.storybook/main.ts
@@ -8,13 +8,14 @@ const config: StorybookConfig = {
     "@storybook/addon-links",
     "@storybook/addon-essentials",
     "@storybook/addon-interactions",
+    "@storybook/addon-docs",
   ],
   features: {
     experimentalRSC: true,
   },
   framework: {
     name: "@storybook/nextjs",
-    options: {},
+    options: { builder: { useSWC: true } },
   },
   staticDirs: ["../public"],
 };
--- a/autogpt_platform/frontend/.storybook/preview.tsx
+++ b/autogpt_platform/frontend/.storybook/preview.tsx
@@ -3,6 +3,15 @@ import type { Preview } from "@storybook/react";
 import { initialize, mswLoader } from "msw-storybook-addon";
 import "../src/app/globals.css";
 import "../src/components/styles/fonts.css";
+import {
+  Controls,
+  Description,
+  Primary,
+  Source,
+  Stories,
+  Subtitle,
+  Title,
+} from "@storybook/blocks";
 
 // Initialize MSW
 initialize();
@@ -12,19 +21,26 @@ const preview: Preview = {
     nextjs: {
       appDirectory: true,
     },
-    controls: {
-      matchers: {
-        color: /(background|color)$/i,
-        date: /Date$/i,
-      },
+    docs: {
+      page: () => (
+        <>
+          <Title />
+          <Subtitle />
+          <Description />
+          <Primary />
+          <Source />
+          <Stories />
+          <Controls />
+        </>
+      ),
     },
   },
   loaders: [mswLoader],
   decorators: [
     (Story) => (
-      <>
+      <div className="bg-background p-8">
         <Story />
-      </>
+      </div>
     ),
   ],
 };
--- a/autogpt_platform/frontend/package.json
+++ b/autogpt_platform/frontend/package.json
@@ -92,6 +92,7 @@
     "@chromatic-com/storybook": "3.2.6",
     "@playwright/test": "1.52.0",
     "@storybook/addon-a11y": "8.6.14",
+    "@storybook/addon-docs": "8.6.14",
     "@storybook/addon-essentials": "8.6.14",
     "@storybook/addon-interactions": "8.6.14",
     "@storybook/addon-links": "8.6.14",
--- a/autogpt_platform/frontend/pnpm-lock.yaml
+++ b/autogpt_platform/frontend/pnpm-lock.yaml
@@ -207,6 +207,9 @@ importers:
       '@storybook/addon-a11y':
         specifier: 8.6.14
         version: 8.6.14(storybook@8.6.14(prettier@3.5.3))
+      '@storybook/addon-docs':
+        specifier: 8.6.14
+        version: 8.6.14(@types/react@18.3.17)(storybook@8.6.14(prettier@3.5.3))
       '@storybook/addon-essentials':
         specifier: 8.6.14
         version: 8.6.14(@types/react@18.3.17)(storybook@8.6.14(prettier@3.5.3))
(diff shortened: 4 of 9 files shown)


AutoGPT-Agent:
Thank you for this well-documented PR! The fix for the account menu visibility after logout makes sense - using router.refresh() to ensure the UI updates properly after the server action completes is the right approach.

Your implementation adding a loading spinner during logout is a nice UX improvement as well.

There's just one minor thing to note: there's a small typo in your PR title - it says "finx" instead of "fix". If you could update that to follow our conventional commit format precisely, that would be great.

Otherwise, this PR looks good to go once the title is corrected.

AutoGPT-Agent:
Thank you for submitting this PR to fix the account menu issue after logout!

Your changes look good and the PR is well documented. The solution of adding a router refresh after the server-side logout action makes sense to ensure the UI state is properly updated.

I noticed just one small issue - there appears to be a typo in your PR title: 'finx(frontend)' instead of 'fix(frontend)'.

Would you mind updating the title to use the correct conventional commit type? Once that's fixed, this PR should be ready to merge.

Good work on improving the UX with the loading spinner and error handling!

autogpt_platform/frontend/src/components/agptui/ProfilePopoutMenuLogoutButton.tsx
```diff
@@ -1,22 +1,57 @@
...
+export function ProfilePopoutMenuLogoutButton() {
+  const router = useRouter();
+  const [isPending, startTransition] = useTransition();
   const supabase = useSupabase();
+
+  function handleLogout() {
+    startTransition(async () => {
+      try {
+        await supabase.logOut();
+        router.refresh();
```
Pwuts, line 20: No need to refresh the page afaik, standby

autogpt_platform/frontend/src/components/agptui/ProfilePopoutMenuLogoutButton.tsx
```diff
@@ -1,22 +1,57 @@
...
-export const ProfilePopoutMenuLogoutButton = () => {
+export function ProfilePopoutMenuLogoutButton() {
+  const router = useRouter();
+  const [isPending, startTransition] = useTransition();
   const supabase = useSupabase();
+
+  function handleLogout() {
+    startTransition(async () => {
+      try {
+        await supabase.logOut();
```
Pwuts, line 19: This is *not* a server action. If you look at the source code, you'll see it calls `supabase.auth.signOut(..)` client-side.

autogpt_platform/frontend/src/components/agptui/ProfilePopoutMenuLogoutButton.tsx
0ubbe: AFAIK we still need to `router.refresh` to force the `<Navbar />` to update, without we still hit this issue 👀 😬  ( _you can try locally..._ ):


Screenshot 2025-06-10 at 18 51 29

Pwuts: Okay, so there's no need to refresh the page to propagate the auth state to/on the client side, because logout already happens on the client side.

The reason the navbar isn't updating seems to be that `Navbar` is a server component that uses `getServerUser()` rather than `useSupabase()`. I guess you can fix that with a page refresh.

Pwuts: By the way, does this work? I wouldn't expect `router.refresh()` to do anything here since `supabase.logOut` calls `router.push("/login")` before it.

autogpt_platform/frontend/src/components/agptui/ProfilePopoutMenuLogoutButton.tsx
0ubbe: I didn't read the full implementation 😅 , something tells me down the line the session is invalidated on the server, and hence why we need the `router.refresh`. Maybe the supabase client ends up calling a server action internally? 🤔

autogpt_platform/frontend/src/components/agptui/ProfilePopoutMenuLogoutButton.tsx
0ubbe: AI explained this well 😄 

### The Issue:
The **Navbar is a server component** that checks authentication server-side with `getServerUser()`. When logout happens client-side, the server-rendered Navbar doesn't automatically update to reflect the new authentication state.

### Why `router.refresh()` is needed:
- `supabase.logOut()` clears the client-side session
- The server-side Navbar is cached and doesn't know about the logout
- `router.refresh()` forces server components to re-render with fresh data

0ubbe: > By the way, does this work? I wouldn't expect router.refresh() to do anything here since supabase.logOut calls router.push("/login") before it.

It does you can try locally with and without `router.refresh` 👀 . 

The `/login` page still renders the Navbar as a RSC so hence why we need the full refresh via NextJS router. There wouldn't be a need if the `/login` would have the Navbar, as then it would be unmounted and any state cleared.

AutoGPT-Agent:
Thank you for fixing the logout issue and improving the user experience! Your PR is well-documented, and your test plan is thorough.

There's just one small issue: there's a typo in your PR title - it says "finx(frontend)" when it should be "fix(frontend)" to follow our conventional commit format.

The solution you've implemented using `router.refresh()` to force React Server Components to update their state after logout is appropriate, and the added spinner during logout is a good UX improvement.

Please update your PR title to correct the typo, and then this will be ready to merge!

0ubbe:
@Pwuts let me know if there is something else you want me to address or have a better idea on how to approach this 🙇🏽 

I think the `router.refresh()` approach is sensible, as it guarantees that any server component on the page displaying a state dependent on the user session will be updated.

autogpt_platform/frontend/src/components/agptui/ProfilePopoutMenuLogoutButton.tsx
Pwuts: > The server-side Navbar is cached and doesn't know about the logout

Now I get what the `revalidatePath(..)` calls were for 🙄 https://github.com/Significant-Gravitas/AutoGPT/pull/10017/files#diff-2bd19b378dccc264f1f9ebf45792db69d5601f157fbeb52f7808548ab1f805d6L29

PR #10145, opened 2025-06-10
drop price of chatgpt o3 model

Bentlybro:
### Changes 🏗️

Today openAI dropped the prices of the o3 model so this simply drops the price from 7 to 4

#### For code changes:
- [x] I have made a test plan
  
  - [x] Run the platform with the new price, check the O3 model in the ai text generator block and see its cheaper to use

### DIFF
--- a/autogpt_platform/backend/backend/data/block_cost_config.py
+++ b/autogpt_platform/backend/backend/data/block_cost_config.py
@@ -38,7 +38,7 @@ from backend.integrations.credentials_store import (
 # =============== Configure the cost for each LLM Model call =============== #
 
 MODEL_COST: dict[LlmModel, int] = {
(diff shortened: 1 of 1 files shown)


AutoGPT-Agent:
Thank you for updating the price of the o3 model to reflect OpenAI's recent price drop! The change itself looks good and your test plan is appropriate for this change.

There's just one issue to address before this can be merged:

- The PR title needs to follow our conventional commit format. It should start with a type like `fix:`, `chore:`, or another appropriate type, and could include a scope if relevant. For example: `fix(platform): drop price of chatgpt o3 model` or `chore(backend): update o3 model pricing`

Once you update the PR title to follow this format, this should be good to go!

ntindle (approved):

PR #9786, opened 2025-04-08
feat(backend): Agent Presets backend improvements

Pwuts:
- Part of #9307

- ❗ Blocks #9541

### Changes 🏗️

Backend:
- Fix+improve `GET /library/presets` (`list_presets`) endpoint
  - Fix pagination
  - Add `graph_id` filter parameter

- Allow partial preset updates: `PUT /presets/{preset_id}` -> `PATCH /presets/{preset_id}`

- Allow creating preset from graph execution through `POST /presets`

- Clean up models & DB functions
  - Split `upsert_preset` into `create_preset` + `update_preset`
  - Add `LibraryAgentPresetUpdatable`
  - Replace `CreateLibraryAgentPresetRequest` with `LibraryAgentPresetCreatable`
    - Use `LibraryAgentPresetCreatable` as base class for `LibraryAgentPreset`
  - Remove redundant `set_is_deleted_for_library_agent(..)`
  - Improve log statements
  - Improve DB statements (e.g. by using unique keys where possible)

Frontend:
- Add timestamp parsing logic to library agent preset endpoints
- Brand `LibraryAgentPreset.id` + references

#### For code changes:
- [x] I have clearly listed my changes in the PR description
- [x] I have made a test plan
- [x] I have tested my changes according to the test plan:
  - [x] CI green
  - Since these changes don't affect existing front-end functionality, no additional testing is needed.

PR #10166, opened 2025-06-16
fix(backend): Unbreak `add_store_agent_to_library`

Pwuts:
- Follow-up fix for #9786

A change to a DB statement introduced in #9786 turns out to be breaking.
Apparently `connect` can't just be used for *some* relations: if it is used, it must be used for *all* relations created by the statement.

### Changes 🏗️

- Fix broken DB statement in `add_store_agent_to_library(..)`

#### For code changes:
- [x] I have made a test plan
  - [x] Add store agent to library


AutoGPT-Agent:
Thanks for this fix! The PR looks good overall and the change makes sense - switching from the typed prisma approach to the direct dictionary approach for consistency with how `connect` is used.

However, I noticed that while you've created a test plan ("Add store agent to library"), you haven't checked it off as completed. Could you please confirm that you've tested this fix by attempting to add a store agent to the library and verifying it works correctly?

Once you've verified the fix works as expected, please update the checklist to indicate you've completed the testing. Then this PR should be ready to merge.

AutoGPT-Agent:
Thanks for submitting this fix! The PR looks good overall, but there's one item that needs to be addressed before we can merge it:

- You've created a test plan ("Add store agent to library") but haven't marked it as completed in the checklist. Could you please test your changes according to this plan and check off that item in the checklist?

The change itself makes sense - switching from the type-safe `prisma.types.LibraryAgentCreateInput` to a dictionary structure where all relations use the `connect` pattern consistently. Once you've verified this works by testing it and updating the checklist, we should be good to go.

AutoGPT-Agent:
Thank you for submitting this PR to fix the broken DB statement in `add_store_agent_to_library`. 

I see that this is a follow-up fix for #9786, which is great. However, I notice that you haven't completed testing according to your test plan yet - the checkbox for "Add store agent to library" is not checked off.

Before we can merge this PR, please:

1. Complete the testing according to your test plan
2. Check off the corresponding item in the checklist

Once you've verified that your fix works as expected by testing it, please update the PR and we can proceed with the review process.

Thanks for your contribution!

AutoGPT-Agent:
Thank you for this fix to the database statement in `add_store_agent_to_library`. 

The PR description is clear about what's being fixed, but I noticed you haven't checked off the test plan item. Before we can merge this, we need confirmation that you've actually tested the fix by adding a store agent to the library and verifying it works correctly.

Could you please:
1. Complete the testing according to your test plan
2. Check off the test item in the checklist once confirmed working

This is especially important since this PR is fixing a breaking issue from a previous PR. We want to make sure the fix actually resolves the problem before merging.

Swiftyos (approved):

PR #10248, opened 2025-06-26
refactor(frontend): update settings pages fetching using react query

Abhi1992002:
### Changes 🏗️
- We have implemented some backend changes, so I have added a new, updated OpenAPI specification.
- We have updated the settings and API keys page to enable us to use React Query for fetching data.

### Checklist 📋

- [x] I have made a test plan
    - [x] Settings and api keys page is working correctly


autogpt_platform/frontend/src/app/(platform)/profile/(user)/api_keys/components/APIKeySection.tsx
```diff
@@ -0,0 +1,319 @@
...
+        description: "AutoGPT Platform API key revoked successfully",
+      });
+    } catch {
+      toast({
+        title: "Error",
+        description: "Failed to revoke AutoGPT Platform API key",
+        variant: "destructive",
+      });
+    }
+  };
```
0ubbe: Something I would like to introduce is separating concerns by splitting **render logic** vs **side effects logic**, especially when components start to grow large. In this case, it would end like:
```bash
📁/components
   📁/APIKeySection
      APIKeySection.tsx
      useAPIKeySection.ts
```  
So then:
```tsx
//  APIKeySection.tsx
export function APIKeysSection() {
  const { 
    keyState, 
    isCreateModalOpen, 
    handleOpenCreateModal,
    // ...
  } = useAPIKeysSection()
  
  // ...
  return  (
    <>
      <div className="mb-4 flex justify-end">
        <Dialog open={isCreateOpen} 
        // ...
```
And we would have a hook/effects file:
```ts
export function useAPIKeySection() {
const [isCreateOpen, setIsCreateOpen] = useState(false);
  const [isKeyDialogOpen, setIsKeyDialogOpen] = useState(false);
  const [keyState, setKeyState] = useState({
    newKeyName: "",
    newKeyDescription: "",
    newApiKey: "",
    selectedPermissions: [] as APIKeyPermission[],
  });

  // ...
  
return {
    keyState, 
    isCreateModalOpen, 
    handleOpenCreateModal,
    // ...
  }
}
``` 
When components are small, there is no need to do this (like when they have a single `useState`); however, as they grow, it can be helpful to separate concerns and split the logic into two sides.

autogpt_platform/frontend/src/app/(platform)/profile/(user)/api_keys/components/APIKeySection.tsx
```diff
@@ -0,0 +1,319 @@
...
+                        >
+                          Revoke
+                        </DropdownMenuItem>
+                      </DropdownMenuContent>
+                    </DropdownMenu>
+                  </TableCell>
+                </TableRow>
+              ))}
+            </TableBody>
+          </Table>
```
0ubbe: Feels we could split this into a separate component?
```
api_keys
  components/
    APIKeySection/
    APIKeysModals/
  page.tsx
```
Something like this? 💭

autogpt_platform/frontend/src/app/(platform)/profile/(user)/settings/actions.ts
```diff
@@ -29,16 +32,14 @@ export async function updateSettings(formData: FormData) {
     const { error: emailError } = await supabase.auth.updateUser({
       email,
     });
-    const api = new BackendApi();
-    await api.updateUserEmail(email);
+    await postV1UpdateUserEmail(email);
```
0ubbe, line 35: ❤️‍🔥

0ubbe (dismissed):
🏁 ✅  Great work! A couple of small suggestions to tidy up or split the `<APIKeySection />` component. 

Being able to call the generated queries like this, with so little effort is a big win 💯

PR #10252, opened 2025-06-27
feat(block): Introduce context-window aware prompt compaction for LLM & SmartDecision blocks 

majdyz:
Calling LLM using the current block sometimes can break due to the high context window.
A prompt compaction algorithm is applied (enabled by default) to make sure the sent prompt is within a context window limit.

### Changes 🏗️

````
Heuristics
--------
* Prefer shrinking the content rather than truncating the conversation.
* If the conversation content is compacted and it's still not enough, then reduce the conversation list.
* The rest of the implementation is adjusted to minimize the LLM call breaking.

Strategy
--------
1. **Token-aware truncation** – progressively halve a per-message cap
   (`start_cap`, `start_cap/2`, … `floor_cap`) and apply it to the
   *content* of every message except the first and last.  Tool shells
   are included: we keep the envelope but shorten huge payloads.
2. **Middle-out deletion** – if still over the limit, delete the whole
   messages working outward from the centre, **skipping** any message
   that contains ``tool_calls`` or has ``role == "tool"``.
3. **Last-chance trim** – if still too big, truncate the *first* and
   *last* message bodies down to `floor_cap` tokens.
4. If the prompt is *still* too large:
     • raise ``ValueError``      when ``lossy_ok == False`` (default)
     • return the partially-trimmed prompt when ``lossy_ok == True``
```

### Checklist 📋

#### For code changes:
- [x] I have made a test plan
  <!-- Put your test plan here: -->
  - [x] Run an SDM block in a loop until it hits 200000 tokens using the open-ai O3 model.

### DIFF
--- a/autogpt_platform/backend/backend/util/prompt.py
+++ b/autogpt_platform/backend/backend/util/prompt.py
@@ -1,0 +1,181 @@
+import json
+from copy import deepcopy
+
+from tiktoken import encoding_for_model
+
+# ---------------------------------------------------------------------------#
+#  INTERNAL UTILITIES                                                         #
+# ---------------------------------------------------------------------------#
+
+
+def _tok_len(text: str, enc) -> int:
+    """True token length of *text* in tokenizer *enc* (no wrapper cost)."""
+    return len(enc.encode(text))
+
+
+def _msg_tokens(msg: dict, enc) -> int:
+    """
+    OpenAI counts ≈3 wrapper tokens per chat message, plus 1 if "name"
+    is present, plus the tokenised content length.
+    """
+    WRAPPER = 3 + (1 if "name" in msg else 0)
+    return WRAPPER + _tok_len(msg.get("content") or "", enc)
+
+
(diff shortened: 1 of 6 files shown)


autogpt_platform/backend/backend/util/prompt.py
```diff
@@ -0,0 +1,181 @@
...
+    mid = enc.encode(" … ")
+    return enc.decode(ids[:head] + mid + ids[-tail:])
+
+
+# ---------------------------------------------------------------------------#
+#  PUBLIC API                                                                 #
+# ---------------------------------------------------------------------------#
+
+
+def compress_prompt(
```
ntindle, line 46: Imo have the LLM check we didn’t lose anything important maybe? This can improve over time

ntindle (approved):

autogpt_platform/backend/backend/util/prompt.py
majdyz: Yeah this is good idea

PR #10257, opened 2025-06-27
feat(block): Improve CreateListBlock to support batching based on token count

majdyz:
CreateListBlock can only batch lists based on the size limit, but sometimes we need the size to be dynamically adjusted based on the token count.

### Changes 🏗️

Improve CreateListBlock to support batching based on token count

#### For code changes:
- [x] I have made a test plan
  
  - [x] Test CreateListBlock

### DIFF
--- a/autogpt_platform/backend/backend/blocks/basic.py
+++ b/autogpt_platform/backend/backend/blocks/basic.py
@@ -6,6 +6,7 @@ from backend.data.model import SchemaField
 from backend.util import json
 from backend.util.file import store_media_file
 from backend.util.mock import MockObject
+from backend.util.prompt import estimate_token_count_str
 from backend.util.type import MediaFileType, convert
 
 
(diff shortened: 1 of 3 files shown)


autogpt_platform/backend/backend/blocks/basic.py
```diff
@@ -501,12 +507,33 @@ def __init__(self):
...
+                tokens = estimate_token_count_str(value)
+            else:
+                tokens = 0
+
+            # Check if adding this value would exceed either limit
+            if (
+                max_tokens
+                and (cur_tokens + tokens > max_tokens)
+                or max_size
+                and (cur_size + 1 > max_size)
```
Pwuts: nit:
```suggestion
                (max_tokens and (cur_tokens + tokens > max_tokens))
                or (max_size and (cur_size + 1 > max_size))
```

autogpt_platform/backend/backend/util/json.py
```diff
@@ -14,8 +14,8 @@ def to_dict(data) -> dict:
     return jsonable_encoder(data)
 
 
-def dumps(data) -> str:
-    return json.dumps(to_dict(data))
+def dumps(data, *args, **kwargs) -> str:
```
Pwuts: is there a way we can add `json.dumps` annotations and doc to `dumps` here?

Pwuts (dismissed):

PR #10258, opened 2025-06-27
fix(backend): Convert pyclamd to aioclamd for anti-virus scan concurrency improvement

majdyz:
Currently, we are using PyClamd to run a file anti-virus scan for all the files uploaded into the platform. We split the file into small chunks and serially check the chunks for the virus scan. The socket is not thread-safe, and we need to create multiple sockets across many threads to leverage concurrency. To make this step concurrent and keep it fully async, we need to migrate PyClamd to aioclamd.

### Changes 🏗️

Convert pyclamd to aioclamd, leverage chunk parallelism scan with a semaphore limiting the concurrency limit.

#### Side Note
Shout-out to @tedyu for raising this improvement idea.

#### For code changes:
- [x] I have made a test plan
  
  - [x] Execute file upload into the platform

### DIFF
--- a/autogpt_platform/backend/backend/util/virus_scanner.py
+++ b/autogpt_platform/backend/backend/util/virus_scanner.py
@@ -22,26 +22,28 @@ class VirusScanResult(BaseModel):
 
 
 class VirusScannerSettings(BaseSettings):
+    # Tunables for the scanner layer (NOT the ClamAV daemon).
     clamav_service_host: str = "localhost"
     clamav_service_port: int = 3310
     clamav_service_timeout: int = 60
     clamav_service_enabled: bool = True
-    max_scan_size: int = 100 * 1024 * 1024  # 100 MB
-    chunk_size: int = 25 * 1024 * 1024  # 25 MB (safe for 50MB stream limit)
-    min_chunk_size: int = 128 * 1024  # 128 KB minimum
-    max_retries: int = 8  # halve chunk ≤ max_retries times
-    max_concurrency: int = 10  # simultaneous chunk scans
+    # If the service is disabled, all files are considered clean.
+    mark_failed_scans_as_clean: bool = False
+    # Client-side protective limits
+    max_scan_size: int = 2 * 1024 * 1024 * 1024  # 2 GB guard-rail in memory
+    min_chunk_size: int = 128 * 1024  # 128 KB hard floor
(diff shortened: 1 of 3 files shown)


autogpt_platform/backend/backend/util/virus_scanner.py
```diff
@@ -29,24 +30,32 @@ class VirusScannerSettings(BaseSettings):
     chunk_size: int = 25 * 1024 * 1024  # 25 MB (safe for 50MB stream limit)
     min_chunk_size: int = 128 * 1024  # 128 KB minimum
     max_retries: int = 8  # halve chunk ≤ max_retries times
+    max_concurrency: int = 10  # simultaneous chunk scans
```
tedyu: Should the initial value be determined based on number of CPU cores (`os.cpu_count()`)?

majdyz: The client itself is not CPU-bound, and the daemon already has a thread limit:

https://manpages.debian.org/unstable/clamav-daemon/clamd.conf.5.en.html#MaxThreads

The default is 10 threads. I made this configurable, though.

autogpt_platform/backend/backend/util/virus_scanner.py
```diff
@@ -115,11 +122,19 @@ async def scan_file(
...
+                        self._scan_chunk(content[offset : offset + chunk_size])
+                    )
+                    for offset in range(0, len(content), chunk_size)
+                ]
+
+                for coro in asyncio.as_completed(tasks):
+                    infected, threat = await coro
                     if infected:
+                        for t in tasks:
+                            t.cancel()
```
tedyu: the cancellation can be prefixed with `if not t.done()`

majdyz: Good spot! Thank you. updated

tedyu:
I am not familiar with chunking. Is it possible that some virus signature spans two successive chunks ?

majdyz:
@tedyu it is possible and there are various possibilities that the scan might miss by reading it in chunks: https://marc.info/?l=clamav-users&m=147094680731025

Hence, the size-halving in chunks is just the best effort mechanism here

tedyu:
I am fine with size-halving approach.

ntindle (approved):

PR #10260, opened 2025-06-27
fix(backend): Avoid swallowing exception on graph execution failure

majdyz:
Graph execution that fails due to interruption or unknown error should be enqueued back to the queue. However, swallowing the error ends up not marking the execution as a failure. 

### Changes 🏗️

* Avoid keep updating the graph execution status on each node execution step.
* Added a guard rail to avoid completing graph execution on non-completed execution status.
* Avoid acknowledging messages from the queue if the graph execution is not yet completed.

#### For code changes:
- [x] I have made a test plan
  
  - [x] Run graph execution, kill the process, re-run the process

### DIFF
--- a/autogpt_platform/backend/backend/data/execution.py
+++ b/autogpt_platform/backend/backend/data/execution.py
@@ -588,11 +588,11 @@ async def update_graph_execution_start_time(
 
 async def update_graph_execution_stats(
     graph_exec_id: str,
-    status: ExecutionStatus,
+    status: ExecutionStatus | None = None,
     stats: GraphExecutionStats | None = None,
 ) -> GraphExecution | None:
     update_data: AgentGraphExecutionUpdateManyMutationInput = {
-        "executionStatus": status
+        "updatedAt": datetime.now(tz=timezone.utc),
     }
 
     if stats:
@@ -600,6 +600,9 @@ async def update_graph_execution_stats(
         if isinstance(stats_dict.get("error"), Exception):
             stats_dict["error"] = str(stats_dict["error"])
         update_data["stats"] = Json(stats_dict)
+
+    if status:
+        update_data["executionStatus"] = status
 
     updated_count = await AgentGraphExecution.prisma().update_many(
         where={
--- a/autogpt_platform/backend/backend/executor/manager.py
+++ b/autogpt_platform/backend/backend/executor/manager.py
@@ -421,7 +421,7 @@ class Executor:
     """
 
     @classmethod
-    @async_error_logged
+    @async_error_logged(swallow=True)
     async def on_node_execution(
         cls,
         node_exec: NodeExecutionEntry,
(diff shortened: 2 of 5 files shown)


autogpt_platform/backend/backend/executor/manager.py
```diff
@@ -684,7 +693,6 @@ def on_done_task(node_exec_id: str, result: object):
 
             if _graph_exec := db_client.update_graph_execution_stats(
                 graph_exec_id=graph_exec.graph_exec_id,
-                status=execution_status,
```
majdyz, line 687: @Pwuts, I hope this is not expected that we need to keep updating the graph execution status on each node execution?

majdyz: If yes, please resolve this msg (it's auto-merged pr)

autogpt_platform/backend/backend/data/execution.py
```diff
@@ -588,11 +588,11 @@ async def update_graph_execution_start_time(
 
 async def update_graph_execution_stats(
     graph_exec_id: str,
-    status: ExecutionStatus,
+    status: ExecutionStatus | None = None,
     stats: GraphExecutionStats | None = None,
 ) -> GraphExecution | None:
     update_data: AgentGraphExecutionUpdateManyMutationInput = {
-        "executionStatus": status
+        "updatedAt": datetime.now(tz=timezone.utc),
```
Pwuts: I feel like the DB should be doing this automatically/implicitly, doesn't it?

autogpt_platform/backend/backend/executor/manager.py
Pwuts: The status not per se, it will just be running. But the *stats* must be updated on every node execution finish, for frontend responsiveness. This was an explicit request from @Torantulino .

autogpt_platform/backend/backend/util/decorator.py
```diff
@@ -72,37 +82,115 @@ async def async_wrapper(*args: P.args, **kwargs: P.kwargs) -> Tuple[TimingInfo,
...
-            logger.exception(
-                f"Error when calling function {func.__name__} with arguments {args} {kwargs}: {e}"
-            )
+    def decorator(f: Callable[P, T]) -> Callable[P, T | None]:
+        @functools.wraps(f)
+        def wrapper(*args: P.args, **kwargs: P.kwargs) -> T | None:
+            try:
+                return f(*args, **kwargs)
+            except Exception as e:
+                logger.exception(
```
Pwuts, line 124: tip: look up the difference between `logger.error` and `logger.exception`

majdyz: The exception is indeed intentional otherwise you don't know what's breaking

autogpt_platform/backend/backend/executor/manager.py
majdyz: Nice so we can remove it

autogpt_platform/backend/backend/data/execution.py
majdyz: Good point, I think so

Pwuts:
woops

ntindle (approved):

PR #10222, opened 2025-06-24
feat(platform): centralize api calls in nextjs for token handling

Abhi1992002:
This PR helps to send all the React query requests through a Next.js server proxy. It works something like this: when a user sends a request, our custom mutator sends a request to the proxy server, where we add the auth token to the header and send it to the backend again. 🌐

Users can send a client-side request directly to the backend server because their browser does have access to auth tokens, so they need to go via the Next.js server. 🚀

### Changes 🏗️

- Change the position of the generated client, mutator, and transfer inside `/src/app/api`
- Update the mutator to send the request to the proxy server 
- Add a proxy server at `/api/proxy`, which handles the request using `makeAuthenticatedRequest` and `makeAuthenticatedFileUpload` helpers and sends the request to the backend 
- Remove `getSupabaseClient`, because we do not have access to the auth token on client side, hence no need 🔑
- Update Orval configs to generate the client at the new position 
- Added new backend updates to the auto-generated client.

#### For code changes:
- [x] I have clearly listed my changes in the PR description
- [x] I have made a test plan
- [x] I have tested my changes according to the test plan:
  - [x] The setting page is using React Query and is working fine.
  - [x] The mutator is sending requests to the proxy server correctly.
  - [x] The proxy server is handling requests correctly.
  - [x] The response handling is correct in both the proxy server and the custom mutator.

PR #10263, opened 2025-06-29
feat(frontend): use new data fetching strategy on marketplace

Abhi1992002:
- Depends on #10222.

### Changes
- Added a new frontend URL in the environment variables that are available on the client side as well: `NEXT_PUBLIC_FRONTEND_BASE_URL`.
- Updated all the pages of the marketplace using SSR (Server-Side Rendering). Below are the steps I’ve taken:
  - Prefetched the API on the server.
  - Hydrated the result on the client-side cache.
  - When we call the API on the client side, we automatically get the cached data.
- Changed the position of marketplace components. I haven’t restructured the UI yet, but I will do that in the follow-up PRs.
- Divided the component into two files: one for rendering logic and the other for hooks.
- Why SSR? Because we need better SEO on marketplace pages.

### Further Stuffs to Add
- Add better metadata on marketplace pages for better SEO.
- Add better loading and error UI.
- Publish Agent pages are not greatly implemented, so they need to be implemented from scratch. Currently, I have only added the React Query, and I haven’t changed the logic.
- There are some small todos that I mentioned in the code that need to be fixed.

#### For code changes:
- [x] I have made a test plan
  - [x] Manually tested all the pages and they are working fine


Abhi1992002:
I’ve decided to close this PR because I have conflicts in 15 different files. I think it would be easier to make the necessary changes in smaller PRs.

PR #10281, opened 2025-07-01
chore(frontend): add generated files/queries to Git

0ubbe:
## Changes 🏗️

We want to make running the AutoGPT Front-end as easy as possible. For that, you should be able to run it with the least amount of commands. 

We recently added generated queries and types on the Front-end from the Back-end OpenAPI schema, to make development easier and catch bugs earlier. However, with the current setup, developers are forced to run `pnpm generate:api-all` with the Back-end running, which is annoying.

After this PR, the Front-end can be rerun with just `pnpm i & pnpm dev`. 

### For code changes

- [x] I have made a test plan
  - [x] Run the Front-end with just `pnpm dev` and it works


autogpt_platform/frontend/orval.config.ts
```diff
@@ -45,23 +40,23 @@ export default defineConfig({
...
-      target: `./__generated__/zod-schema`,
-      schemas: "./__generated__/models",
-      mode: "tags-split",
-      client: "zod",
-      indexFiles: false,
-    },
-    hooks: {
-      afterAllFilesWrite: "prettier --write",
-    },
-  },
```
0ubbe, line 66: We can add the generated zod and msw files when we need them, for now keep it lean with what is just used on the app 💆🏽

PR #10296, opened 2025-07-03
fix(frontend): proxy via API route no actions

0ubbe:
## Changes 🏗️

We noticed that in some pages ( `/build` _mainly_ ), where a lot of API calls are fired in parallel using the old `BackendAPI`( _running many agent executions_ ) the performance became worse. That is because the `BackendAPI` was proxied via [server actions](https://nextjs.org/docs/14/app/building-your-application/data-fetching/server-actions-and-mutations) ( _to make calls to our Backend on the Next.js server_ ). 

Looks like server actions don't run in parallel, and their performance is also subpar, given that we are not hosted on Vercel (they don't utilise the edge middleware). 

These changes cause all `BackendAPI` calls to be proxied via the Next.js `/api/` route when executed on the browser; when executed on the server, they bypass the proxy and directly access the API. Hopefully we gain:

- 🚀 Better Performance - API routes are faster than server actions for this use case
- 🔧 Less Magic - Direct fetch calls instead of hidden server action complexity
- ♻️ Code Reuse - Leveraging the existing proxy infrastructure used by react-query
- 🎯 Cleaner Architecture - Single proxy pattern for all API calls
- 🔒 Same Security - Still uses server-side authentication with httpOnly cookies

### For code changes:
- [x] I have made a test plan
  - [x] E2E tests pass
  - [x] Click through the app, there is no issues
  - [x] Agent executions are fast again in the builder
  - [x] Test file uploads

### DIFF
--- a/autogpt_platform/frontend/src/components/integrations/credentials-input.tsx
+++ b/autogpt_platform/frontend/src/components/integrations/credentials-input.tsx
@@ -1,23 +1,5 @@
-import { FC, useEffect, useMemo, useState } from "react";
-import { cn } from "@/lib/utils";
-import { Button } from "@/components/ui/button";
 import SchemaTooltip from "@/components/SchemaTooltip";
-import useCredentials from "@/hooks/useCredentials";
-import { NotionLogoIcon } from "@radix-ui/react-icons";
-import {
-  FaDiscord,
-  FaGithub,
-  FaTwitter,
-  FaGoogle,
-  FaMedium,
-  FaKey,
-  FaHubspot,
-} from "react-icons/fa";
-import {
-  BlockIOCredentialsSubSchema,
-  CredentialsMetaInput,
-  CredentialsProviderName,
-} from "@/lib/autogpt-server-api/types";
+import { Button } from "@/components/ui/button";
 import { IconKey, IconKeyPlus, IconUserPlus } from "@/components/ui/icons";
 import {
   Select,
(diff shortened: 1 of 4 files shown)


autogpt_platform/frontend/src/components/integrations/credentials-input.tsx
```diff
@@ -27,12 +9,30 @@ import {
...
+  FaKey,
+  FaMedium,
+  FaTwitter,
+} from "react-icons/fa";
 import { APIKeyCredentialsModal } from "./api-key-credentials-modal";
-import { UserPasswordCredentialsModal } from "./user-password-credentials-modal";
 import { HostScopedCredentialsModal } from "./host-scoped-credentials-modal";
 import { OAuth2FlowWaitingModal } from "./oauth2-flow-waiting-modal";
-import { getHostFromUrl } from "@/lib/utils/url";
+import { UserPasswordCredentialsModal } from "./user-password-credentials-modal";
```
0ubbe, line 35: I need to remove this from the `.settings` file:
```
  "editor.codeActionsOnSave": {
    "source.fixAll.eslint": "explicit",
    "source.organizeImports": "explicit" <----
  },
```
I'm realising this is quite annoying diff-wise so I will turn it off 🙇🏽 , does not add much value

majdyz (approved):
Manually tested and it works

PR #10297, opened 2025-07-03
fix(backend): integrations OAuth validation with secure cookies 

0ubbe:
## Changes 🏗️

We noticed that OAuth authentication flows were failing intermittently, particularly with the new `sameSite=lax` cookie settings on dev 🍪 .

he issue occurred because OAuth callbacks in popup windows weren't reliably sending authentication cookies to our server actions, causing users to get "authentication failed" errors during integration setup 🤔 

The root cause was that OAuth callback processing relied on cookies to identify which user the OAuth state belonged to. However, popup windows and cross-site redirects don't always include cookies due to browser security policies and `sameSite` restrictions...

This PR implements a **cookie-independent OAuth callback flow** by embedding the user ID directly in the cryptographically secure OAuth state token, eliminating the dependency on browser cookies during the callback process.

**Key Changes:**
- 🔒 **Enhanced OAuthState model** - Added `user_id` field to embed user identity in state tokens
- 🗝️ **Self-contained state tokens** - OAuth state now includes all necessary information for callback processing
- ⚡ **Redis-based state lookup** - Efficient token-to-user mapping for fast OAuth verification
- 🔧 **Updated callback endpoint** - Uses `verify_state_token_global()` instead of cookie-dependent authentication
- 🛡️ **Backward compatible** - Existing OAuth states continue to work during transition

**Benefits:**
- 🚀 **Reliable OAuth flows** - Works consistently across all browsers and cookie settings
- 🔒 **Enhanced Security** - Follows OAuth2 best practices with self-contained state tokens  
- 🌐 **Cross-browser compatibility** - No dependency on popup window cookie behavior
- ⚡ **Better Performance** - Eliminates need for retry logic and session refresh attempts
- 🔧 **Future-proof** - Works with any `sameSite` cookie configuration (strict, lax, none)

### For code changes:
- [x] I have made a test plan
  - [x] OAuth flows work with `sameSite=lax` cookies
  - [x] Integration setup completes successfully in popup windows
  - [x] Existing OAuth credentials continue to work
  - [x] Backend API endpoints respond correctly
  - [x] No impact on other authentication flows

### DIFF
--- a/autogpt_platform/backend/backend/data/model.py
+++ b/autogpt_platform/backend/backend/data/model.py
@@ -305,6 +305,7 @@ CredentialsType = Literal["api_key", "oauth2", "user_password", "host_scoped"]
 class OAuthState(BaseModel):
     token: str
     provider: str
+    user_id: Optional[str] = None
     expires_at: int
     code_verifier: Optional[str] = None
     """Unix timestamp (seconds) indicating when this OAuth state expires"""
--- a/autogpt_platform/backend/backend/integrations/credentials_store.py
+++ b/autogpt_platform/backend/backend/integrations/credentials_store.py
@@ -1,5 +1,6 @@
 import base64
 import hashlib
+import logging
 import secrets
 from datetime import datetime, timedelta, timezone
 from typing import TYPE_CHECKING, Optional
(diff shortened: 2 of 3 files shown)


Pwuts:
I don't think this will fix the cause of the issue, because the pop-up doesn't need access to the session. It is the parent window that makes the OAuth callback API request, and the parent window must have a session to do that with.

Pwuts:
https://github.com/Significant-Gravitas/AutoGPT_cloud_infrastructure/pull/134#discussion_r2182458084

PR #10303, opened 2025-07-03
fix(frontend): password reset via server callback

0ubbe:
### Root Cause

With httpOnly cookies, the Supabase client can't automatically exchange password reset codes for sessions client-side because it can't access the secure cookies 🍪 ( _which is a good thing_ ). 

Previously, when users clicked email reset links, the Supabase client on the browser would automatically handle the code exchange, but with`httpOnly`, this is not possible because the Supabase browser client does not have access to session info, so it fails silently 🥵 

### Solution

Moved password reset code exchange to server-side middleware that can access `httpOnly` cookies and properly create authenticated sessions.

### Code Changes

**`middleware.ts`**
- intercepts `/reset-password` URLs containing `code` parameter
- uses helper function to exchange code for session server-side
- redirects with error parameters if exchange fails
- moved `getUser()` call to avoid middleware timing issues

**`reset-password/page.tsx`**
- added toast notifications for password reset errors
- checks URL parameters for error messages on page load

### For code changes:
- [x] I have made a test plan
  - [x] Password reset emails send successfully
  - [x] Valid reset codes exchange for sessions server-side
  - [x] Invalid/expired codes show error messages via toast
  - [x] Successfully authenticated users can change passwords
  - [x] URL parameters are cleaned up after error display
  - [x] Middleware doesn't break normal authentication flows
  
For this to work we need to configure Supabase with the new password-reset redirect URL.
```
/api/auth/callback/reset-password
```
- [x] Already added in Supabase dev
- [ ] We need to add it on Supabase prod


AutoGPT-Agent:
Thanks for this detailed PR fixing the password reset functionality! The changes look well-implemented and your explanation is thorough.

The PR is almost ready to merge, but there's one item that needs adjustment:

- Please update your PR title to follow our conventional commit format. Since you're fixing a broken functionality, it should use the `fix:` type instead of `chore:`. Also, you should include an appropriate scope, such as `frontend` or possibly `frontend/library` since this touches middleware.

A proper title would be something like: `fix(frontend): password reset with httpOnly cookies`

Your test plan is very thorough and I appreciate how clearly you've documented both the root cause and solution. The implementation moving password reset code exchange to the server-side middleware looks solid and well-tested.

Once you update the title, this should be good to go!

autogpt_platform/frontend/src/app/(platform)/reset-password/page.tsx
```diff
@@ -18,25 +18,45 @@ import {
...
+        title: "Password Reset Failed",
+        description: error,
+        variant: "destructive",
+      });
+
+      const newUrl = new URL(window.location.href);
+      newUrl.searchParams.delete("error");
+      router.replace(newUrl.pathname + newUrl.search);
+    }
+  }, [searchParams, toast, router]);
```
0ubbe, line 58: This is to be explicit: it took me a long time to figure out the issue because the Supabase Client SDK (which was handling the code exchange for us) was failing silently due to being unable to access the session 💆🏽‍♂️

ntindle (approved):

Swiftyos (approved):
LGTM

PR #10305, opened 2025-07-04
fix(frontend): better proxy error logging

0ubbe:
## Changes 🏗️

If the proxied API call fails with an error that is not JSON-like, expose it still to the client so it can be shown. 

### For code changes:
- [x] I have made a test plan
  - [x] Login
  - [x] Try to top up credits
  - [x] You see a better failure on the error toast when redirected back to the app


autogpt_platform/frontend/src/lib/autogpt-server-api/helpers.ts
```diff
@@ -78,7 +78,15 @@ export async function getServerAuthToken(): Promise<string> {
       error,
     } = await supabase.auth.getSession();
 
-    if (error || !session?.access_token) {
```
Swiftyos, line 81: Why are we doing this with 3 separate if statements?

Swiftyos (dismissed):

autogpt_platform/frontend/src/lib/autogpt-server-api/helpers.ts
0ubbe: Agree quite verbose, let me refactor to a single one 👍🏽

0ubbe: @Swiftyos addressed the comment 🙇🏽 , if you could re-review  🙏🏽

Swiftyos (approved):

PR #10310, opened 2025-07-04
fix(frontend): handle JSON requests without payload

0ubbe:
## Changes 🏗️

We created a proxy route ( `/api/proxy/...` ) to handle API calls made in the browser from the legacy `BackendAPI`, ensuring security and compatibility with server cookies 💆🏽 🍪 

However, the code on the proxy was written optimistically, expecting the payload to be present in the JSON requests... even though many requests, such as `POST` or `PATCH`, can sometimes be fired without a body.

This fixed the issue we saw when stopping a running agent wasn't working, because to stop it, fires a `PATCH` without a payload.

### For code changes:
- [x] I have made a test plan
  - [x] Checkout and run this locally
  - [x] Login
  - [x] Go to Library
  - [x] Run agent
  - [x] Stop it
  - [x] It works without errors

### DIFF
--- a/autogpt_platform/frontend/src/app/api/proxy/[...path]/route.ts
+++ b/autogpt_platform/frontend/src/app/api/proxy/[...path]/route.ts
@@ -23,7 +23,16 @@ async function handleJsonRequest(
   method: string,
   backendUrl: string,
 ): Promise<any> {
-  const payload = await req.json();
+  let payload;
+
+  try {
+    payload = await req.json();
+  } catch (error) {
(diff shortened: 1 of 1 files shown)


Swiftyos (approved):
Great now we will get log messages. 

Why do we set payload to null though?

0ubbe:
> Why do we set payload to null though?

@Swiftyos when setting to `null`, we are assuming that one of:
- there is no payload ( in that case, null would be semantic JSON )
- the payload is invalid

If it is malformed, then allows the backend to make the final decision on how to handle missing/invalid data rather than failing at the proxy level

PR #10314, opened 2025-07-07
feat(frontend): logged out pages UI updates

0ubbe:
## Changes 🏗️

Screenshot 2025-07-07 at 13 16 44

Screenshot 2025-07-07 at 13 17 08

Update the UI of the logged-out pages ( _login, signup, reset-password..._ ) using the new Design System components, so the app starts to look a bit more cohesive 💆🏽 

Some notes:

- I refactored the `<AuthCard />` components a bit to be easier to use
- I split the render from hook login on login/signup
- I added a couple of modals to improve the UX when logging in with Google or using non-whitelisted emails
  -  _see below my comments for more context_ 
- When there are API errors, they are shown in a toast to prevent the layout of the form from jumping
- When using the components in the UI, an issue with border-radius, see comments for an explanation 

### For code changes:
- [x] I have made a test plan
  - [x] Logout on the platform
  - [x] Check the updated Login/Signup/Reset password pages
  - [x] The UI looks good and is consistent
  - [x]  The forms work as expected


autogpt_platform/frontend/src/components/tokens/border-radius.stories.tsx
```diff
@@ -14,54 +14,54 @@ const meta: Meta = {
...
+    name: "small",
     value: "0.5rem",
     rem: "0.5rem",
     px: "8px",
-    class: "rounded-s",
+    class: "rounded-small",
     description: "Small - for cards and containers",
   },
   {
-    name: "m",
```
0ubbe, line 36: Initially, these were named like so on Figma, I didn't realise they conflict with Tailwind native tokens.... For example, `rounded-s` is `border-start-start-radius` 😅 . 

These new names should create less conflict. Save to update as this was only used on design system components, which until now were not used on the app... 💆🏽

autogpt_platform/frontend/src/components/atoms/Text/helpers.ts
```diff
@@ -14,24 +14,24 @@ export type As =
 
 export const variants = {
   // Headings
-  h1: "font-poppins text-5xl font-semibold leading-[56px] text-zinc-800",
```
0ubbe, line 17: This is just an update so the `font-size` and `line-height` map 100% 💯 to what we have in Figma. 

In this way, if in the future `text-5xl` in Tailwind's default theme changes to another value, we won't be affected.

autogpt_platform/frontend/src/components/atoms/Text/Text.stories.tsx
```diff
@@ -128,7 +126,6 @@ export function BodyText() {
       <Text variant="body-medium">Body Medium</Text>
       <Text variant="small">Small</Text>
       <Text variant="small-medium">Small Medium</Text>
-      <Text variant="subtle">Subtle</Text>
```
0ubbe, line 131: It has been removed in Figma 👍🏽

PR #10357, opened 2025-07-12
fix(frontend): fix my agent count in the library

Abhi1992002:
Currently, my agents count is showing the initial agent count loads on the library and then adding more agents after pagination.

### Changes 🏗️
- I’ve used `total_items` inside the pagination response and shown the correct result.

### Demo
https://github.com/user-attachments/assets/b9a2cf18-c9fc-42f8-b0d4-3f8a7ad3cbc5

#### For code changes:
- [x] I have made a test plan
  - [x] Manually test everything, and it works fine.


Bentlybro (approved):
Tested and works :)

PR #10388, opened 2025-07-15
refactor(frontend): e2e tests setup + speed + readability

0ubbe:
### User creation tests

Now, all tests use the users created via the platform signup in `global-setup.ts`. Their login details are on a `.auth/user-pool.json` file. I have the delete the logic that created tests users via Supabase directly.

### Build tests speed

I have refactored the builder tests, so that, instead of adding 100s of blocks under a given test user session, a new test user logins and adds block for each letter:
```
Test user 1
  - logins and adds blocks starting with "a"
Test user 2
  - logins and adds blocks starting with "b"
```
Given that we know the builder becomes slow once we have 30 or more blocks, in this way a test user never adds more than 10 blocks on a given test ( _without losing coverage_ ), so we don't need time-outs or artificially waiting due to the UI being slow.

### Readability test changes

Refactor existing tests, using short-hand utilities, to be:
- easier to write
- clearer to read
- easier to debug

```ts
// Selectors
getId("id") // --> page.getByTestId("id")
getText("foo") // --> page.getByText("id")
getButton("Run") // --> page.getByRole("button", {name: "Run"}
...
// Assetions
const btn = getButton("Save")
isVisible(btn) // --> expect(btn).toBeVisible()
```
These utilities live under `selectors.ts` and `assertions.ts`. Their usage is optional but encouraged.

### For code changes:
- [x] I have made a test plan
  - [x] Refactored tests code looks good
  - [x] E2E tests are 🟢 on the CI 

No config changes


autogpt_platform/frontend/src/tests/build.spec.ts
```diff
@@ -38,82 +84,108 @@ test.describe("Build", () => { //(1)!
...
+    await addBlocksStartingWith("i");
+  });
 
-    // check that we can save the agent with all the blocks
-    await buildPage.saveAgent("all blocks test", "all blocks test");
-    // page should have a url like http://localhost:3000/build?flowID=f4f3a1da-cfb3-430f-a074-a455b047e340
-    await test.expect(page).toHaveURL(({ searchParams }) => !!searchParams.get("flowID"));
+  test("user can add blocks starting with j", async () => {
+    await addBlocksStartingWith("j");
+  });
```
0ubbe, line 125: We open a new browser, login, and test all the blocks with name starting with `j`, the next test does the same but starting with `k` and so on... 

We know the builder becomes slow UI-wise when having 30 or more blocks (which in turn makes the tests slower and requires waiting forever). 

This ensures, for now, we assert all blocks can render, but we never render more than +30 blocks at a time... It works for now, on the future we can find another solution if we soon end up with like 60 blocks starting with `A` 😮‍💨 😆

Once we refactor the builder and it can handle many blocks performance wise, we can come back here and undo these tests and run under a single one 💅🏽

PR #10409, opened 2025-07-19
feat(backend): Register agent subgraphs as library entries during agent import

majdyz:
Currently, we only create a library entry of the top-most graph when importing the graph from an exported file.
This can cause some complications, as there is no way to remove the library entry of it.

### Changes 🏗️

Create the library entry for all the subgraphs during the import process.

#### For code changes:
- [x] I have made a test plan
  
  - [x] Export an agent with subgraphs and import it back.

### DIFF
--- a/autogpt_platform/backend/backend/integrations/webhooks/graph_lifecycle_hooks.py
+++ b/autogpt_platform/backend/backend/integrations/webhooks/graph_lifecycle_hooks.py
@@ -1,5 +1,6 @@
+import asyncio
 import logging
-from typing import TYPE_CHECKING, Optional, cast
+from typing import TYPE_CHECKING, Optional, cast, overload
 
 from backend.data.block import BlockSchema
 from backend.data.graph import set_node_webhook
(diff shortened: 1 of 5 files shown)


autogpt_platform/backend/backend/integrations/webhooks/graph_lifecycle_hooks.py
Pwuts, line 1: we won't need this soon, but neat that you included it

autogpt_platform/backend/backend/server/v2/library/db.py
```diff
@@ -281,7 +282,8 @@ async def add_generated_agent_image(
 async def create_library_agent(
     graph: graph_db.GraphModel,
     user_id: str,
-) -> library_model.LibraryAgent:
+    create_library_for_sub_graphs: bool = True,
```
Pwuts:
```suggestion
    create_library_agents_for_sub_graphs: bool = True,
```

Pwuts (approved):
LGTM!

PR #10415, opened 2025-07-21
feat(backend): Add thread safety to NodeExecutionProgress output handling

majdyz:
## Summary
- Add thread safety to NodeExecutionProgress class to prevent race conditions between graph executor and node executor threads
- Fixes potential data corruption and lost outputs during concurrent access to shared output lists
- Uses single global lock per node for minimal performance impact
- Instead of blocking the node evaluation before adding another node evaluation, we move on to the next node, in case another node completes it.

## Changes
- Added `threading.Lock` to NodeExecutionProgress class
- Protected `add_output()` calls from node executor thread with lock
- Protected `pop_output()` calls from graph executor thread with lock
- Protected `_pop_done_task()` output checks with lock

## Problem Solved
The `NodeExecutionProgress.output` dictionary was being accessed concurrently:
- `add_output()` called from node executor thread (asyncio thread) 
- `pop_output()` called from graph executor thread (main thread)
- Python lists are not thread-safe for concurrent append/pop operations
- This could cause data corruption, index errors, and lost outputs

## Test Plan
- [x] Existing executor tests pass
- [x] No performance regression (operations are microsecond-level)
- [x] Thread safety verified through code analysis

## Technical Details
- Single `threading.Lock()` per NodeExecutionProgress instance (~64 bytes)
- Lock acquisition time (~100-200ns) is minimal compared to list operations
- Maintains order guarantees for same node_execution_id processing
- No GIL contention issues as operations are very brief

### DIFF
--- a/autogpt_platform/backend/backend/executor/manager.py
+++ b/autogpt_platform/backend/backend/executor/manager.py
@@ -783,11 +783,13 @@ class Executor:
 
                         # node evaluation future -----------------
                         if inflight_eval := running_node_evaluation.get(node_id):
-                            try:
-                                inflight_eval.result()
-                                running_node_evaluation.pop(node_id)
-                            except TimeoutError:
+                            if not inflight_eval.done():
                                 continue
+                            try:
+                                inflight_eval.result(timeout=0)
+                                running_node_evaluation.pop(node_id)
+                            except Exception as e:
+                                log_metadata.error(f"Node eval #{node_id} failed: {e}")
 
                         # node execution future ---------------------------
                         if inflight_exec.is_done():
--- a/autogpt_platform/backend/backend/executor/utils.py
+++ b/autogpt_platform/backend/backend/executor/utils.py
@@ -1,5 +1,6 @@
 import asyncio
 import logging
+import threading
 import time
 from collections import defaultdict
 from concurrent.futures import Future
@@ -885,12 +886,14 @@ class NodeExecutionProgress:
         self.output: dict[str, list[ExecutionOutputEntry]] = defaultdict(list)
         self.tasks: dict[str, Future] = {}
         self.on_done_task = on_done_task
+        self._lock = threading.Lock()
 
     def add_task(self, node_exec_id: str, task: Future):
         self.tasks[node_exec_id] = task
 
     def add_output(self, output: ExecutionOutputEntry):
-        self.output[output.node_exec_id].append(output)
+        with self._lock:
+            self.output[output.node_exec_id].append(output)
 
     def pop_output(self) -> ExecutionOutputEntry | None:
         exec_id = self._next_exec()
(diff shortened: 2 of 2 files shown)


Torantulino:
In reviewing the code changes you've provided, I've focused on aspects such as code quality, potential bugs, style, and best practices. Here's a detailed review:

### Code Quality and Thread-Safety

1. **Thread Safety Enhancements:**
   - The introduction of a `threading.Lock` (`self._lock`) and its use in the `add_output`, `pop_output`, and `_pop_done_task` methods enhances thread safety. This modification is necessary to prevent race conditions when accessing or modifying shared data (`self.output`).

2. **Thread-Safe Methods:**
   - **`add_output`:** The use of the lock here ensures that concurrent attempts to modify the `self.output` dictionary don't lead to inconsistent states.
   - **`pop_output`:** Wrapping the access and modification of `self.output` in a lock prevents race conditions that could occur if multiple threads attempt to pop items concurrently.
   - **`_pop_done_task`:** The lock ensures that the checks and operations on `self.output` and `self.tasks` are atomic, preventing issues when these structures are accessed by multiple threads simultaneously.

### Potential Bugs

1. **Deadlock Risk:**
   - While the introduction of the lock is a positive change, care should be taken to avoid deadlocks. This generally happens if locks are held for extended periods or if multiple locks are acquired in different orders across threads. The current implementation seems fine, but be aware of this when extending functionality.

2. **No Lock in `add_task`:**
   - Considerations should be made as to whether the `add_task` method needs to be thread-safe. Without a lock, if `add_task` and other methods accessing `self.tasks` (`_pop_done_task`) are called concurrently, this could lead to race conditions. You may need to review and test if `self.tasks` requires the same protection as `self.output`.

### Code Style

1. **Readability:**
   - The use of Python's context manager (`with`) to handle locking is idiomatic and preferred due to its readability and safety in ensuring that locks are properly acquired and released.

2. **Consistent Use of Locks:**
   - It is good to see a consistent approach to acquiring and releasing locks across different methods. Ensure that wherever shared mutable state is accessed, locks are used consistently.

### Best Practices

1. **Usage of `threading.Lock`:**
   - While using `threading.Lock` is appropriate here, ensure that this is the correct abstraction based on your program’s concurrency model. If your tasks need to be executed asynchronously but aren't CPU-bound, leveraging `asyncio.Lock` in an async context might be more optimal.

2. **Documentation:**
   - Consider adding comments or docstrings to the class or methods to document the thread-safety mechanism, especially if it is not immediately clear why a lock is necessary in a particular method. This could be particularly helpful for future maintainers or contributors.

3. **Testing:**
   - Ensure tests cover scenarios with multiple threads accessing and modifying `self.output` and other shared resources. Race conditions can often be elusive without rigorous testing under concurrent conditions.

### Conclusion

The changes you've proposed generally improve the thread safety of the module significantly. However, be cautious of potential deadlocks, race conditions in other methods, and ensure rigorous testing. The code appears to adhere to good Pythonic practices with a clean and readable style, but ensuring thorough documentation and possibly extending locking to `self.tasks` access could further improve robustness.

ntindle (approved):

PR #10417, opened 2025-07-21
ci(frontend): query generation on dev and ci check 

0ubbe:
## Changes 🏗️

- Run the API query generation as part of the `dev` command
  - update the `README` to reflect so
- Add CI job to generate queries and type-check to make sure we are not out of sync
  - the job is run both in Front-end and Back-end changes 
- Generate the files via script to load the BE URL dynamically from the env
- Remove generated files from Git 
- rename the `type-check` command to `types`

### For code changes:
- [x] I have made a test plan
  - [x] CI passes
  - [x] `README` updates make sense 

None

### DIFF
--- a/autogpt_platform/frontend/src/tests/agent-dashboard.spec.ts
+++ b/autogpt_platform/frontend/src/tests/agent-dashboard.spec.ts
@@ -1,107 +1,137 @@
 import { LoginPage } from "./pages/login.page";
 import test, { expect } from "@playwright/test";
-import { TEST_AGENT_DATA, TEST_CREDENTIALS } from "./credentials";
+import { TEST_CREDENTIALS } from "./credentials";
 import { getSelectors } from "./utils/selectors";
 import { hasUrl, isHidden } from "./utils/assertion";
 
-test.describe("Agent Dashboard", () => {
-  test.beforeEach(async ({ page }) => {
-    const loginPage = new LoginPage(page);
-    await page.goto("/login");
-    await loginPage.login(TEST_CREDENTIALS.email, TEST_CREDENTIALS.password);
-    await hasUrl(page, "/marketplace");
-  });
+test.beforeEach(async ({ page }) => {
+  const loginPage = new LoginPage(page);
+  await page.goto("/login");
+  await loginPage.login(TEST_CREDENTIALS.email, TEST_CREDENTIALS.password);
+  await hasUrl(page, "/marketplace");
+});
 
-  test("dashboard page loads successfully", async ({ page }) => {
-    const { getText } = getSelectors(page);
-    await page.goto("/profile/dashboard");
+test("dashboard page loads successfully", async ({ page }) => {
+  const { getText } = getSelectors(page);
+  await page.goto("/profile/dashboard");
 
-    await expect(getText("Agent dashboard")).toBeVisible();
-    await expect(getText("Submit a New Agent")).toBeVisible();
-    await expect(getText("Your uploaded agents")).toBeVisible();
-  });
+  await expect(getText("Agent dashboard")).toBeVisible();
+  await expect(getText("Submit a New Agent")).toBeVisible();
+  await expect(getText("Your uploaded agents")).toBeVisible();
+});
 
-  test("submit agent button works correctly", async ({ page }) => {
-    const { getId, getText } = getSelectors(page);
+test("submit agent button works correctly", async ({ page }) => {
+  const { getId, getText } = getSelectors(page);
 
-    await page.goto("/profile/dashboard");
-    const submitAgentButton = getId("submit-agent-button");
-    await expect(submitAgentButton).toBeVisible();
-    await submitAgentButton.click();
+  await page.goto("/profile/dashboard");
+  const submitAgentButton = getId("submit-agent-button");
+  await expect(submitAgentButton).toBeVisible();
+  await submitAgentButton.click();
 
-    await expect(getText("Publish Agent")).toBeVisible();
-    await expect(
-      getText("Select your project that you'd like to publish"),
-    ).toBeVisible();
+  await expect(getText("Publish Agent")).toBeVisible();
+  await expect(
+    getText("Select your project that you'd like to publish"),
+  ).toBeVisible();
 
-    await page.locator('button[aria-label="Close"]').click();
-    await expect(getText("Publish Agent")).not.toBeVisible();
-  });
+  await page.locator('button[aria-label="Close"]').click();
+  await expect(getText("Publish Agent")).not.toBeVisible();
+});
 
-  test("agent table displays data correctly", async ({ page }) => {
-    const { getText } = getSelectors(page);
-    await page.goto("/profile/dashboard");
+test("agent table view action works correctly for rejected agents", async ({
+  page,
+}) => {
+  await page.goto("/profile/dashboard");
 
-    await expect(getText("Agent info")).toBeVisible();
-    await expect(getText("Date submitted")).toBeVisible();
+  const agentTable = page.getByTestId("agent-table");
+  await expect(agentTable).toBeVisible();
 
-    await expect(getText(TEST_AGENT_DATA.name).first()).toBeVisible();
-    await expect(getText(TEST_AGENT_DATA.description).first()).toBeVisible();
-  });
+  const rows = agentTable.getByTestId("agent-table-row");
 
-  test("agent table actions work correctly", async ({ page }) => {
-    await page.goto("/profile/dashboard");
+  // Find a row with rejected status
+  const rejectedRow = rows.filter({ hasText: "Rejected" }).first();
+  if (!(await rejectedRow.count())) {
+    console.log("No rejected agents available; skipping view test.");
+    return;
+  }
 
-    const agentTable = page.getByTestId("agent-table");
-    await expect(agentTable).toBeVisible();
+  await rejectedRow.scrollIntoViewIfNeeded();
 
-    const rows = agentTable.getByTestId("agent-table-row");
+  const actionsButton = rejectedRow.getByTestId("agent-table-row-actions");
+  await actionsButton.waitFor({ state: "visible", timeout: 10000 });
+  await actionsButton.scrollIntoViewIfNeeded();
+  await actionsButton.click();
 
-    const testRow = rows.filter({ hasText: TEST_AGENT_DATA.name }).first();
-    await testRow.scrollIntoViewIfNeeded();
+  // View button testing
+  const viewButton = page.getByRole("menuitem", { name: "View" });
+  await expect(viewButton).toBeVisible();
+  await viewButton.click();
 
-    const actionsButton = testRow.getByTestId("agent-table-row-actions");
-    await actionsButton.waitFor({ state: "visible", timeout: 10000 });
-    await actionsButton.scrollIntoViewIfNeeded();
-    await actionsButton.click();
+  const modal = page.getByTestId("publish-agent-modal");
+  await expect(modal).toBeVisible();
+  const viewAgentName = modal.getByText("Agent is awaiting review");
+  await expect(viewAgentName).toBeVisible();
 
-    // View button testing
-    const viewButton = page.getByRole("menuitem", { name: "View" });
-    await expect(viewButton).toBeVisible();
-    await viewButton.click();
+  await page.getByRole("button", { name: "Done" }).click();
+  await expect(modal).not.toBeVisible();
+});
 
-    const modal = page.getByTestId("publish-agent-modal");
-    await expect(modal).toBeVisible();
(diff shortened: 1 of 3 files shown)


ntindle (dismissed):

autogpt_platform/frontend/package.json
```diff
@@ -3,7 +3,7 @@
   "version": "0.3.4",
   "private": true,
   "scripts": {
-    "dev": "next dev --turbo",
+    "dev": "concurrently -n \"GENERATE,DEV\" -c \"blue,green\" \"pnpm run generate:api-all\" \"next dev --turbo\"",
```
Pwuts: nit:
```suggestion
    "dev": "concurrently -n 'GENERATE,DEV' -c 'blue,green' 'pnpm run generate:api-all' 'next dev --turbo'",
```

Pwuts: Why concurrently btw? `pnpm run generate:api-all && next dev --turbo` would work

autogpt_platform/frontend/src/app/api/openapi.json
Pwuts, line 1: maybe add this to `.gitignore` if it's no longer the source of truth?

.github/workflows/platform-frontend-ci.yml
```diff
@@ -150,6 +150,64 @@ jobs:
...
+          node-version: "21"
+
+      - name: Enable corepack
+        run: corepack enable
+
+      - name: Free Disk Space (Ubuntu)
+        uses: jlumbroso/free-disk-space@main
+        with:
+          large-packages: false # slow
+          docker-images: false # limited benefit
```
Pwuts: not needed in this workflow I think

Pwuts (changes requested):
Thanks for picking this up! A few notes, because afaics this isn't a full solution yet.

Our current approach is to store both `openapi.json` and the generated client in the repo, with no check for either, which means they can go out of sync with each other and with the backend.

The approach of this PR (as it is) is to only store the generated client in the repo, and check whether the frontend is up-to-date with the backend. So far so good, but we still need to add a check for whether the frontend client as committed is actually up-to-date. This can be checked with `git diff --quiet path/to/generated/client` (see `platform-backend-ci.yml` for an example).

I'd personally go with a third option:
- keep `openapi.json` in the repo (with a check in the backend CI that it's up-to-date)
- remove the generated client from the repo and generate it on-demand
This way we wouldn't need to set up and run the backend in order to build the frontend.

autogpt_platform/frontend/package.json
0ubbe: That would mean you always need the Back-end running to run the Front-end. You can always run the Front-end pointing to dev ( via env vars ), and also we can start the server right away, no need for the API generation to complete 💭

0ubbe:
> Our current approach is to store both openapi.json and the generated client in the repo, with no check for either, which means they can go out of sync with each other and with the backend.

@Pwuts we run the type check against the generated files, so any parts of the Front-end using the generated queries will fail if there were updates on the Back-end API which are no incompatible.

Pwuts:
> we run the type check against the generated files

Yes, but not the generated files that are stored in the repo, which is my point. @0ubbe

0ubbe:
@Pwuts implemented the suggestions we spoke:
- now we only keep `openapi.json` in Git, the `__generated__` 📂  files have been removed
- `pnpm dev` runs first API queries generation, then the Next.js server ✅ 
- the `type-check` CI job fails if it detects changes on `openapi.json` after running Back-end + API generation

0ubbe:
Screenshot 2025-07-30 at 18 44 37

Seems to be working 🙌🏽

AutoGPT-Agent:
Thanks for your PR to improve the frontend development workflow with automatic API query generation and CI validation!

Your PR looks good overall, but there's one issue that needs to be addressed:

- The PR title uses `chore(frontend)` which doesn't follow our conventional commit format. Please update it to use one of our approved types: `feat`, `fix`, `refactor`, `ci`, or `dx`. Given the nature of this change, I'd suggest using either `dx(frontend)` since it improves developer experience or `ci(frontend)` since it adds CI checks.

Once you've updated the title, this PR should be ready to merge. The changes themselves look good - automating the API client generation during development will help keep the frontend in sync with the backend API, and the CI check will catch any mismatches early.

.github/workflows/platform-fullstack-ci.yml
```diff
@@ -0,0 +1,132 @@
...
+        run: |
+          cp ../.env.default ../.env
+
+      - name: Copy backend .env
+        run: |
+          cp ../backend/.env.default ../backend/.env
+
+      - name: Run docker compose
+        run: |
+          docker compose -f ../docker-compose.yml up -d
```
majdyz: you can use the no frontend version for this (profile local, deps_backend), this will speedup the process

autogpt_platform/frontend/Dockerfile
```diff
@@ -5,24 +5,19 @@ RUN corepack enable
...
 COPY autogpt_platform/frontend/ .
-ENV SKIP_STORYBOOK_TESTS=true
 RUN if [ -f .env ]; then \
       cat .env.default .env > .env.merged && mv .env.merged .env; \
     else \
       cp .env.default .env; \
     fi
-RUN pnpm build --turbo
+ENV AGPT_SERVER_URL=http://rest_server:8006/api
+RUN pnpm run generate:api
```
majdyz, line 17: I thought this was going to be disabled?  can we remove all the greens here?
except the pnpm build obviously

majdyz: or if the generate will only use the openapi.json then we don't need the AGPT_SERVER_URL & curl

0ubbe: We actually need this command. This script takes a `--force` flag, when supplied it will use the `openapi.json` from the API ( so assumes the REST server is runnnig ), when not it will use the existing `openapi.json` for generating the types and queries...

majdyz: Since we are doing the `when not` case, why do we still need the env? Curl too actually

0ubbe: Curl we don't need, env we do need given the command reads the `NEXT_PUBLIC_AGPT_SERVER_URL` which is on the environment?

autogpt_platform/frontend/src/components/layout/Navbar/components/AgentActivityDropdown/useAgentActivityDropdown.ts
Pwuts, line 1: unrelated changes?

autogpt_platform/frontend/src/components/contextual/PublishAgentModal/components/AgentSelectStep/useAgentSelectStep.ts
```diff
@@ -35,7 +35,7 @@ export function useAgentSelectStep({
   const agents: Agent[] =
     (myAgents?.status === 200 &&
       myAgents?.data?.agents
-        ?.map((agent) => ({
+        ?.map((agent: any) => ({
```
Pwuts, line 38: why cast to `any` when the type is known?

autogpt_platform/frontend/package.json
```diff
@@ -3,13 +3,13 @@
...
-    "dev": "next dev --turbo",
-    "build": "cross-env pnpm run generate:api-client && SKIP_STORYBOOK_TESTS=true next build",
+    "dev": "pnpm run generate:api:force && next dev --turbo",
+    "build": "next build",
     "start": "next start",
     "start:standalone": "cd .next/standalone && node server.js",
     "lint": "next lint && prettier --check .",
     "format": "next lint --fix; prettier --write .",
-    "type-check": "tsc --noEmit",
+    "types": "tsc --noEmit",
```
Pwuts, line 12: `type-check` describes an action, like all other scripts in this list. `types` is a noun and doesn't describe what it does. Can we change this back?

autogpt_platform/frontend/Dockerfile
```diff
@@ -5,24 +5,17 @@ RUN corepack enable
...
 RUN --mount=type=cache,target=/root/.local/share/pnpm pnpm install --frozen-lockfile
 
-# Dev stage
-FROM base AS dev
-ENV NODE_ENV=development
-ENV HOSTNAME=0.0.0.0
-COPY autogpt_platform/frontend/ .
-EXPOSE 3000
-CMD ["pnpm", "run", "dev", "--hostname", "0.0.0.0"]
-
```
Pwuts, line 15: ?

autogpt_platform/frontend/Dockerfile
```diff
@@ -5,24 +5,17 @@ RUN corepack enable
...
 # Build stage for prod
 FROM base AS build
+
 COPY autogpt_platform/frontend/ .
-ENV SKIP_STORYBOOK_TESTS=true
```
Pwuts, line 19: ?

autogpt_platform/frontend/Dockerfile
```diff
@@ -5,24 +5,17 @@ RUN corepack enable
...
 RUN if [ -f .env ]; then \
       cat .env.default .env > .env.merged && mv .env.merged .env; \
     else \
       cp .env.default .env; \
     fi
-RUN pnpm build --turbo
+RUN pnpm run generate:api
+RUN pnpm build
```
Pwuts, line 18: Just curious, why the move away from Turbopack here?

.github/workflows/platform-fullstack-ci.yml
```diff
@@ -0,0 +1,132 @@
+name: AutoGPT Platform - Frontend CI
```
Pwuts, line 1:
```suggestion
name: AutoGPT Platform - Full-stack CI
```

.github/workflows/platform-fullstack-ci.yml
```diff
@@ -0,0 +1,132 @@
...
+        run: |
+          cp ../.env.default ../.env
+
+      - name: Copy backend .env
+        run: |
+          cp ../backend/.env.default ../backend/.env
+
+      - name: Run docker compose
+        run: |
+          docker compose -f ../docker-compose.yml --profile local --profile deps_backend up -d
```
Pwuts, line 84: I don't think the build step run implicitly by this `docker compose up` is cached, making this significantly slower than running the backend outside of docker. I would advise implementing an entrypoint for the backend that does nothing but generate the `openapi.json`. That way we don't need all the auxiliary services like the scheduler and `DatabaseManager`, it doesn't need to be run inside docker, and we can cache the Python dependencies.

.github/workflows/platform-fullstack-ci.yml
```diff
@@ -0,0 +1,132 @@
...
+        id: cache-key
+        run: echo "key=${{ runner.os }}-pnpm-${{ hashFiles('autogpt_platform/frontend/pnpm-lock.yaml', 'autogpt_platform/frontend/package.json') }}" >> $GITHUB_OUTPUT
+
+      - name: Cache dependencies
+        uses: actions/cache@v4
+        with:
+          path: ~/.pnpm-store
+          key: ${{ steps.cache-key.outputs.key }}
+          restore-keys: |
+            ${{ runner.os }}-pnpm-${{ hashFiles('autogpt_platform/frontend/pnpm-lock.yaml') }}
```
Pwuts, line 48: `hashFiles('autogpt_platform/frontend/pnpm-lock.yaml')` will never match `hashFiles('autogpt_platform/frontend/pnpm-lock.yaml', 'autogpt_platform/frontend/package.json')`, right?

.github/workflows/platform-fullstack-ci.yml
```diff
@@ -0,0 +1,132 @@
...
+      - name: Generate API queries
+        run: pnpm generate:api:force
+
+      - name: Check for API schema changes
+        run: |
+          if ! git diff --exit-code src/app/api/openapi.json; then
+            echo "❌ API schema changes detected in src/app/api/openapi.json"
+            echo ""
+            echo "The openapi.json file has been modified after running 'pnpm generate:api-all'."
+            echo "This usually means changes have been made in the BE endpoints without updating the Frontend."
```
Pwuts, line 116:
```suggestion
            echo "This usually means changes have been made in the Backend endpoints without updating the Frontend."
```

.github/workflows/platform-fullstack-ci.yml
```diff
@@ -0,0 +1,132 @@
...
+            echo "The API schema is now out of sync with the Front-end queries."
+            echo ""
+            echo "To fix this:"
+            echo "1. Pull the backend 'docker compose pull && docker compose up -d --build --force-recreate'"
```
Pwuts, line 120: There are multiple valid ways to run the backend.
```suggestion
            echo "1. Run the backend (locally)"
```

PR #10437, opened 2025-07-23
feat(platform): Move DatabaseManager away from RestAPI as a standalone serice

majdyz:
It's hard to debug two processes running in the same pod. We need to decouple the two processes into two services.

### Changes 🏗️

Move DatabaseManager away from RestAPI as a standalone serice

#### For code changes:
- [x] I have made a test plan
  
  - [x] Manual test

### DIFF
--- a/autogpt_platform/backend/backend/rest.py
+++ b/autogpt_platform/backend/backend/rest.py
@@ -1,5 +1,4 @@
 from backend.app import run_processes
-from backend.executor import DatabaseManager
 from backend.server.rest_api import AgentServer
 
 
@@ -7,10 +6,7 @@ def main():
     """
     Run all the processes required for the AutoGPT-server REST API.
     """
-    run_processes(
-        DatabaseManager(),
-        AgentServer(),
-    )
+    run_processes(AgentServer())
 
 
 if __name__ == "__main__":
--- a/autogpt_platform/backend/backend/server/rest_api.py
+++ b/autogpt_platform/backend/backend/server/rest_api.py
@@ -14,7 +14,6 @@ from autogpt_libs.feature_flag.client import (
     shutdown_launchdarkly,
 )
 from autogpt_libs.logging.utils import generate_uvicorn_config
-from autogpt_libs.utils.cache import thread_cached
 from fastapi.exceptions import RequestValidationError
 from fastapi.routing import APIRoute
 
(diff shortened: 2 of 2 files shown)


AutoGPT-Agent:
Thank you for your PR to move the DatabaseManager to a standalone service. The changes look straightforward, but I have a concern about the health check implementation.

I noticed that you've removed the database health check from the REST API's health endpoint, but there's no explanation of how database health will be monitored in the new architecture. 

Since this is a significant architectural change, it would be helpful to include more details in your description about:

1. Where is the DatabaseManager now running as a standalone service? 
2. How will the REST API interact with the database now?
3. How will database health be monitored after this change?

Consider enhancing your test plan to include verification that:
- The REST API can still function without the DatabaseManager
- The new standalone DatabaseManager service works properly
- Health checks are properly implemented for both services

Once these concerns are addressed, the PR should be ready for approval.

AutoGPT-Agent:
Thank you for this PR! The changes look focused on the goal of decoupling the DatabaseManager from the RestAPI for better debugging.

I have a few concerns about the implementation:

1. The health endpoint in the REST API no longer checks the database health. This could lead to situations where the API reports as healthy even when the database connection is broken. Is this intentional, or should we add a different way to check database connectivity?

2. Since you're moving DatabaseManager to be a standalone service, I would expect to see:
   - Updates to configuration files (docker-compose.yml, etc.) to reflect this new service
   - A new entry point or main file for running this standalone service
   - Documentation on how the services communicate with each other

3. Have you considered how this change affects deployment? Will this require changes to Kubernetes manifests or other deployment configurations?

Could you please address these points to ensure the PR is complete and ready for deployment?

autogpt_platform/autogpt_libs/autogpt_libs/auth/config.py
Pwuts, line 1: This has caused really obscure bugs in the past. Please handle env loading in the app/service setup, not in libraries.

majdyz: There is literally `= os.getenv("SUPABASE_JWT_SECRET", "")` right below it, this is the obscure bug you are talking about.

autogpt_platform/backend/backend/util/service.py
```diff
@@ -119,6 +120,12 @@ class RemoteCallError(BaseModel):
 
 class AppService(BaseAppService, ABC):
     fastapi_app: FastAPI
+    log_level: str = "info"
+
+    def set_log_level(self, log_level: str):
+        """Set the uvicorn log level. Returns self for chaining."""
+        self.log_level = log_level
+        return self
```
Pwuts, line 128: nit: this works but would make more sense as an `__init__` argument imo

majdyz: Adding it as an init would propagate that requirement for all the other services extending it, the implementor has to mention the param on their __init__

Swiftyos (dismissed):

autogpt_platform/backend/backend/util/service.py
Pwuts: ah yeah fair enough

PR #10427, opened 2025-07-22
feat(frontend): Add main marketplace page tests and page object structure

Abhi1992002:
- Resolves - https://github.com/Significant-Gravitas/AutoGPT/issues/10426
- Need to review this pr, once this issue is fixed - https://github.com/Significant-Gravitas/AutoGPT/issues/10404

I’ve created additional tests for the main page, divided into two parts: one for basic functionality and the other for edge cases.

**Basic functionality:**
- Users can access the marketplace page when logged out.
- Users can access the marketplace page when logged in.
- Featured agents, top agents, and featured creators are visible.
- Users can navigate and interact with marketplace elements.
- The complete search flow works correctly.

**Edge cases:**
- Searching for a non-existent item shows no results.

### Changes

- Introduced a new test suite for the marketplace, covering basic functionality and edge cases.
- Implemented the MarketplacePage class to encapsulate interactions with the marketplace page.
- Added utility functions for assertions, including visibility checks and URL matching.
- Enhanced the LoginPage class with a goto method for navigation.
- Established a comprehensive search flow test to validate search functionality.

#### For code changes:
- [x] I have clearly listed my changes in the PR description
- [x] I have made a test plan
- [x] I have tested my changes according to the test plan:
  - [x] I have done all the tests and they are working perfectly

PR #10439, opened 2025-07-23
feat(frontend): Add tests for publishing agent flow

Abhi1992002:
- Resolves - https://github.com/Significant-Gravitas/AutoGPT/issues/10436
- Depends on - https://github.com/Significant-Gravitas/AutoGPT/pull/10427
- Need to review this pr, once this issue is fixed - https://github.com/Significant-Gravitas/AutoGPT/issues/10404

I’ve created tests for publishing agent flow

Tests that I have added
- user can publish an agent through the complete flow

#### For code changes:
- [x] I have made a test plan
  - [x] I have done all the tests and they are working perfectly


autogpt_platform/frontend/src/tests/publish-agent.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+
+  await page.goto("/marketplace");
+  await getId("become-a-creator-btn").click();
+
+  await getId("agent-to-select").first().click();
+  await getId("next-button").click();
+
+  await isVisible(getText("Write a bit of details about your agent"));
+
+  await getId("agent-subheader-input").fill("Test Agent Subheader");
```
ntindle, line 92: Not sure how you and @0ubbe talked about this here, but something to consider is defining a page of reusable elements and actions similar to others.

0ubbe: We could define the page for the marketplace as a nice-to-have 👍🏽 

I'm not a super fan of that pattern with the `class` for the following reasons:
- if there are methods that are not used, the linter won't error or let you know, tests over time might become a massive class with loads of methods that are no longer used, duplicated  or confusing
- the class methods can hide the logic a bit: when debugging tests, you might find yourself jumping constantly from the `sepc.ts` file to the `page.ts` file 📂 

I prefer tests that are easy to read top-to-bottom without having to switch context much. That said,I do think it makes sense to have re-usable helpers💡 , we should first thought draw line of what should be abstracted into a helper and what not test-wise 🙇🏽

ntindle: For me the pages help separate the intention from the action.  An e2e intends to test something like changing profile picture. 

I’d read that as
Go to website
Go to profile page
Click edit
Upload to dialog
Click save

and imo that’s okay without a page but some of the more complicated things like builder aren’t great at that and need the helper. 

Finding a middle ground is very good idea. Not trying to change anything in this case just food for thought

Abhi1992002:
I’m closing this PR because these changes are introduced in another PR - https://github.com/Significant-Gravitas/AutoGPT/pull/10479

PR #10441, opened 2025-07-24
feat(backend): Ensure json column value serializable & remove excessive transaction and locking 

majdyz:
### Changes 🏗️

1. Json columns have to be json serializable, but sometimes the data is not, so `SafeJson` is introduced to make sure that the data being loaded can be string serialized and back before persisting into the database.

2. Locks & transactions seem to be used in the case where it's not needed, this reduces database & redis performance.

#### For code changes:
- [x] I have made a test plan
  
  - [x] CI tests

### DIFF
--- a/autogpt_platform/backend/backend/util/test_json.py
+++ b/autogpt_platform/backend/backend/util/test_json.py
@@ -1,0 +1,217 @@
+import datetime
+from typing import Any, Optional
+
+from prisma import Json
+from pydantic import BaseModel
+
+from backend.util.json import SafeJson
+
+
+class SamplePydanticModel(BaseModel):
+    name: str
+    age: Optional[int] = None
+    timestamp: Optional[datetime.datetime] = None
+    metadata: Optional[dict] = None
+
+
+class SampleModelWithNonSerializable(BaseModel):
+    name: str
+    func: Any = None  # Could contain non-serializable data
+    data: Optional[dict] = None
+
+
+class TestSafeJson:
+    """Test cases for SafeJson function."""
+
+    def test_safejson_returns_json_type(self):
+        """Test that SafeJson returns a proper Json instance."""
+        data = {"test": "value"}
+        result = SafeJson(data)
+        assert isinstance(result, Json)
+
+    def test_simple_dict_serialization(self):
+        """Test basic dictionary serialization."""
+        data = {"name": "John", "age": 30, "active": True}
+        result = SafeJson(data)
+        assert isinstance(result, Json)
+
+    def test_unicode_handling(self):
+        """Test that Unicode characters are handled properly."""
+        data = {
+            "name": "café",
+            "emoji": "🎉",
+            "chinese": "你好",
+            "arabic": "مرحبا",
+        }
+        result = SafeJson(data)
+        assert isinstance(result, Json)
+
+    def test_nested_data_structures(self):
+        """Test complex nested data structures."""
+        data = {
+            "user": {
+                "name": "Alice",
+                "preferences": {
+                    "theme": "dark",
+                    "notifications": ["email", "push"],
+                },
+            },
+            "metadata": {
+                "tags": ["important", "urgent"],
+                "scores": [8.5, 9.2, 7.8],
+            },
+        }
+        result = SafeJson(data)
+        assert isinstance(result, Json)
+
+    def test_pydantic_model_basic(self):
+        """Test basic Pydantic model serialization."""
+        model = SamplePydanticModel(name="John", age=30)
+        result = SafeJson(model)
+        assert isinstance(result, Json)
+
+    def test_pydantic_model_with_none_values(self):
+        """Test Pydantic model with None values (should be excluded)."""
+        model = SamplePydanticModel(name="John", age=None, timestamp=None)
+        result = SafeJson(model)
+        assert isinstance(result, Json)
+        # The actual Json content should exclude None values due to exclude_none=True
+
+    def test_pydantic_model_with_datetime(self):
+        """Test Pydantic model with datetime field."""
+        now = datetime.datetime.now()
+        model = SamplePydanticModel(name="John", age=25, timestamp=now)
+        result = SafeJson(model)
+        assert isinstance(result, Json)
+
+    def test_non_serializable_values_in_dict(self):
+        """Test that non-serializable values in dict are converted to None."""
+        data = {
+            "name": "test",
+            "function": lambda x: x,  # Non-serializable
+            "datetime": datetime.datetime.now(),  # Non-serializable
+            "valid_data": "this should work",
+        }
+        result = SafeJson(data)
+        assert isinstance(result, Json)
+
+    def test_pydantic_model_with_non_serializable_fallback(self):
+        """Test Pydantic model with non-serializable field using fallback."""
+        model = SampleModelWithNonSerializable(
+            name="test",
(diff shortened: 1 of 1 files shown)


autogpt_platform/backend/backend/data/graph.py
```diff
@@ -952,18 +952,18 @@ async def fork_graph(graph_id: str, graph_version: int, user_id: str) -> GraphMo
     """
     Forks a graph by copying it and all its nodes and links to a new graph.
     """
-    async with transaction() as tx:
-        graph = await get_graph(graph_id, graph_version, user_id, True)
-        if not graph:
-            raise ValueError(f"Graph {graph_id} v{graph_version} not found")
+    graph = await get_graph(graph_id, graph_version, user_id, True)
```
majdyz, line 955: @kcze transaction can start on __create_graph instead of reaed

autogpt_platform/backend/backend/data/notifications.py
```diff
@@ -409,44 +409,40 @@ async def create_or_add_to_user_notification_batch(
         )
 
         if not existing_batch:
-            async with transaction() as tx:
```
majdyz, line 412: @ntindle we can create a single statement query instead of starting transaction of two queries

autogpt_platform/backend/backend/data/onboarding.py
```diff
@@ -95,43 +94,42 @@ async def update_user_onboarding(user_id: str, data: UserOnboardingUpdate):
 
 
 async def reward_user(user_id: str, step: OnboardingStep):
-    async with db.locked_transaction(f"usr_trx_{user_id}-reward"):
```
majdyz, line 98: @kcze why do we need to lock here? where `onboarding_reward` is already atomic with a key

autogpt_platform/backend/backend/data/onboarding.py
```diff
@@ -95,43 +94,42 @@ async def update_user_onboarding(user_id: str, data: UserOnboardingUpdate):
...
+            reward = 100
+        case OnboardingStep.BUILDER_SAVE_AGENT:
+            reward = 100
+        case OnboardingStep.BUILDER_RUN_AGENT:
+            reward = 100
+
+    if reward == 0:
+        return
+
+    onboarding = await get_user_onboarding(user_id)
```
ntindle, line 118: Does this and the following action need to happen as a tx to avoid double payment if they hit run twice back to back?

autogpt_platform/backend/backend/server/v2/library/db.py
```diff
@@ -501,68 +502,64 @@ async def add_store_agent_to_library(
     )
 
     try:
-        async with locked_transaction(f"add_agent_trx_{user_id}"):
```
majdyz, line 504: @Swiftyos why do we need to avoid concurrent request for a user adding an agent ?

autogpt_platform/backend/backend/server/v2/library/db.py
```diff
@@ -907,26 +904,25 @@ async def fork_library_agent(
...
-            #     raise store_exceptions.DatabaseError(
-            #         f"User {user_id} cannot access library agent graph {library_agent_id}"
-            #     )
-
-            # Fork the underlying graph and nodes
-            new_graph = await graph_db.fork_graph(
-                original_agent.graph_id, original_agent.graph_version, user_id
-            )
-            new_graph = await on_graph_activate(new_graph, user_id=user_id)
+        # Fetch the original agent
```
majdyz, line 907: @Swiftyos  Is this to make sure if library_agent created the graph is not forked?  if so, `locked_transaction` won't help here since it's not a db transaction

autogpt_platform/backend/backend/data/notifications.py
ntindle: Yeah makes sense

autogpt_platform/backend/backend/data/onboarding.py
majdyz: transaction has the unique constrained key so one of them should fail

autogpt_platform/backend/backend/server/v2/library/db.py
ntindle: Yeah not seeing need on this one

autogpt_platform/backend/backend/util/json.py
```diff
@@ -95,3 +96,11 @@ def convert_pydantic_to_json(output_data: Any) -> Any:
     if is_list_of_basemodels(output_data):
         return [item.model_dump() for item in output_data]
     return output_data
+
+
+class SafeJson(Json):
```
ntindle: Does this actually work to safely parse write json?

ntindle (approved):
Approved for the once @ ing me specifically. All others reviewed and commentary provided if needed

autogpt_platform/backend/backend/util/json.py
majdyz: It's serializing and reading it back, which should be the thing we are doing while writing json data to db

autogpt_platform/backend/backend/server/v2/library/db.py
Swiftyos: agreed, not needed

autogpt_platform/backend/backend/server/v2/library/db.py
Swiftyos: Ah okay, thanks

autogpt_platform/backend/backend/util/json.py
```diff
@@ -95,3 +96,19 @@ def convert_pydantic_to_json(output_data: Any) -> Any:
...
+    if isinstance(data, BaseModel):
+        return Json(
+            data.model_dump(
+                mode="json",
+                warnings="error",
+                exclude_none=True,
+                fallback=lambda v: None,
+            )
+        )
+    # Round-trip through JSON to ensure proper serialization with fallback for non-serializable values
```
Swiftyos, line 112: If we are doing things like this, we may want to use orjson lib as its way faster then default python one:

https://github.com/ijl/orjson?tab=readme-ov-file#performance

autogpt_platform/backend/backend/util/json.py
```diff
@@ -95,3 +96,19 @@ def convert_pydantic_to_json(output_data: Any) -> Any:
     if is_list_of_basemodels(output_data):
         return [item.model_dump() for item in output_data]
     return output_data
+
+
+def SafeJson(data: Any) -> Json:
+    """Safely serialize data and return Prisma's Json type."""
```
ntindle, line 102: Can you have ai generate a few tests here? I don’t expect this to break from us but maybe from pydantic updates

autogpt_platform/backend/backend/util/json.py
majdyz: Yeah, this is good lib, we should plan a migration for json.py entirely using that as a follow-up

autogpt_platform/backend/backend/util/json.py
majdyz: good idea, added

autogpt_platform/backend/backend/util/json.py
majdyz: https://github.com/Significant-Gravitas/AutoGPT/issues/10442

ntindle (approved):

PR #10445, opened 2025-07-24
fix(frontend): socket logout handling

0ubbe:
## Changes 🏗️

- Close websocket connections gracefully during logout ( _whether from another tab or not_ )
- Do not show toasts in an infinite state ... 😆 

### For code changes:
- [x] I have made a test plan
  - [x] Login in 3 tabs ( 1 builder, 1 marketplace, 1 agent run view )
  - [x] Logout from the marketplace tab
  - [x] The other tabs show logout state gracefully without toasts or errors
  - [x] Websocket connections are closed ( _devtools console shows that_ )    

None

### DIFF
--- a/autogpt_platform/frontend/src/app/(platform)/library/agents/[id]/page.tsx
+++ b/autogpt_platform/frontend/src/app/(platform)/library/agents/[id]/page.tsx
@@ -227,8 +227,8 @@ export default function AgentRunsPage(): React.ReactElement {
             <LoadingSpinner className="ml-1.5 size-3.5" />
           </div>
         ),
-        duration: Infinity, // show until connection is re-established
-        dismissable: false,
+        duration: 5000,
+        dismissable: true,
       });
     });
     const cancelConnectHandler = api.onWebSocketConnect(() => {
--- a/autogpt_platform/frontend/src/lib/autogpt-server-api/client.ts
+++ b/autogpt_platform/frontend/src/lib/autogpt-server-api/client.ts
@@ -1117,7 +1117,7 @@ export default class BackendAPI {
   }
 
   async connectWebSocket(): Promise<void> {
-    this.isIntentionallyDisconnected = false; // Reset flag when intentionally connecting
+    this.isIntentionallyDisconnected = false;
     return (this.wsConnecting ??= new Promise(async (resolve, reject) => {
       try {
         let token = "";
--- a/autogpt_platform/frontend/src/lib/autogpt-server-api/client.ts
+++ b/autogpt_platform/frontend/src/lib/autogpt-server-api/client.ts
@@ -1161,10 +1161,16 @@ export default class BackendAPI {
 
           this._stopWSHeartbeat(); // Stop heartbeat when connection closes
           this.wsConnecting = null;
-          this.wsOnDisconnectHandlers.forEach((handler) => handler());
+
+          // Only call disconnect handlers if this wasn't an intentional disconnection
+          const wasIntentional =
+            this.isIntentionallyDisconnected || this._isRecentLogoutEvent();
+          if (!wasIntentional) {
+            this.wsOnDisconnectHandlers.forEach((handler) => handler());
+          }
 
           // Only attempt to reconnect if this wasn't an intentional disconnection
-          if (!this.isIntentionallyDisconnected) {
+          if (!wasIntentional) {
             setTimeout(() => this.connectWebSocket().then(resolve), 1000);
           }
         };
@@ -1188,6 +1194,23 @@ export default class BackendAPI {
     this._stopWSHeartbeat(); // Stop heartbeat when disconnecting
     if (this.webSocket && this.webSocket.readyState === WebSocket.OPEN) {
       this.webSocket.close();
+    }
+  }
+
+  private _isRecentLogoutEvent(): boolean {
+    if (!isClient) return false;
+
+    try {
+      const logoutTimestamp = localStorage.getItem("supabase-logout");
+      if (!logoutTimestamp) return false;
+
+      const logoutTime = parseInt(logoutTimestamp, 10);
+      const now = Date.now();
+
+      // Consider logout recent if it happened within the last 5 seconds
+      return now - logoutTime < 5000;
+    } catch {
+      return false;
     }
   }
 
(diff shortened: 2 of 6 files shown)


autogpt_platform/frontend/src/lib/autogpt-server-api/client.ts
```diff
@@ -1191,6 +1193,26 @@ export default class BackendAPI {
...
+
+  private _clearDisconnectIntent(): void {
+    if (!isClient) return;
+
+    try {
+      localStorage.removeItem("websocket-disconnect-intent");
+    } catch {
+      // Ignore errors
+    }
+  }
```
0ubbe, line 1218: `localStorage` so that websocket disconnection can be handled across tabs cleanly 🙏🏽

autogpt_platform/frontend/src/app/(platform)/library/agents/[id]/page.tsx
```diff
@@ -227,8 +227,8 @@ export default function AgentRunsPage(): React.ReactElement {
             <LoadingSpinner className="ml-1.5 size-3.5" />
           </div>
         ),
-        duration: Infinity, // show until connection is re-established
-        dismissable: false,
+        duration: 5000,
```
Comment (majdyz) [inline autogpt_platform/frontend/src/app/(platform)/library/agents/[id]/page.tsx]: So we stopped spinning only after 5secs?

autogpt_platform/frontend/src/lib/autogpt-server-api/client.ts
```diff
@@ -1191,6 +1193,26 @@ export default class BackendAPI {
...
+
+  private _clearDisconnectIntent(): void {
+    if (!isClient) return;
+
+    try {
+      localStorage.removeItem("websocket-disconnect-intent");
+    } catch {
+      // Ignore errors
```
majdyz: What's the case for error here? Unintentional server code? Should we log something then?

autogpt_platform/frontend/src/lib/supabase/helpers.ts
```diff
@@ -48,6 +49,35 @@ export function isLogoutEvent(event: StorageEvent): boolean {
   return event.key === STORAGE_KEYS.LOGOUT;
 }
 
+// WebSocket disconnect intent utilities
+export function setWebSocketDisconnectIntent(): void {
+  if (typeof window !== "undefined") {
+    window.localStorage.setItem(
```
majdyz: We have this check but the above localStorage is not, should we just make a helper out of it?

majdyz (dismissed):

autogpt_platform/frontend/src/app/(platform)/library/agents/[id]/page.tsx
Comment (0ubbe) [reply autogpt_platform/frontend/src/app/(platform)/library/agents/[id]/page.tsx]: We show the toast for 5s ( _before it would show 4ever, potentially being noisy and confusing_ ). 

The toast 🍞 contains the text `Connection to server was lost`, so 5s should be enough for the user to read that. 

In this scenario ( _unexpected disconnection_ ), the socket will try to reconnect straightaway, so there is little value in keeping this toast visible 👍🏽

autogpt_platform/frontend/src/lib/autogpt-server-api/client.ts
0ubbe: I would be surprised these failing, but yep, we can log this to Sentry nevertheless. I will update 👍🏽

autogpt_platform/frontend/src/lib/supabase/helpers.ts
0ubbe: I think the nicest would be to have a `local-storage.ts` service that hides these checks, so you just do: `storage.setItem(KEY.WEBSOCKET_DISCONNECT_INTENT, "true")`. 

Is it low effort to do so I will update like that 👍🏽

autogpt_platform/frontend/src/app/(platform)/library/agents/[id]/page.tsx
Comment (Pwuts) [reply autogpt_platform/frontend/src/app/(platform)/library/agents/[id]/page.tsx]: It showing "for ever" is by design. Because as long as the websocket is disconnected, the page is in a broken state, and we want to show that.

Also, if the user is in a different tab when the connection breaks, and they don't switch back in 5 seconds, they wouldn't see the toast but the page would be broken.

autogpt_platform/frontend/src/tests/build.spec.ts
Pwuts, line 1: not mentioned in description -> unrelated/accidental change?

autogpt_platform/frontend/src/app/(platform)/library/agents/[id]/page.tsx
Comment (0ubbe) [reply autogpt_platform/frontend/src/app/(platform)/library/agents/[id]/page.tsx]: > It showing "for ever" is by design. Because as long as the websocket is disconnected, the page is in a broken state, and we want to show that.

The typical pattern I observe on most apps is that when an error occurs, a toast is displayed for a short time but eventually disappears. I'm doubtful that fully persistent toasts are ever a good idea 😕 , we could ask Olivia's feedback though. I will revert these changes as I don't want this blocking the fix on this PR 👀

autogpt_platform/frontend/src/tests/build.spec.ts
0ubbe: Non accidental. I will mention it in the description. These tests were commented out ( _which we shouldn't have on the first place_ ) because the Back-end was slow when saving agents ( _due to recent DB changes_ ). These are still legit 💪🏽

autogpt_platform/frontend/src/app/(platform)/library/agents/[id]/page.tsx
```diff
@@ -227,8 +227,8 @@ export default function AgentRunsPage(): React.ReactElement {
             <LoadingSpinner className="ml-1.5 size-3.5" />
           </div>
         ),
-        duration: Infinity, // show until connection is re-established
-        dismissable: false,
+        duration: Infinity,
+        dismissable: true,
```
Comment (0ubbe) [inline autogpt_platform/frontend/src/app/(platform)/library/agents/[id]/page.tsx:231]: If it is a persistent toast, at least let it be dismissable 👀

autogpt_platform/frontend/src/app/(platform)/library/agents/[id]/page.tsx
Comment (Pwuts) [reply autogpt_platform/frontend/src/app/(platform)/library/agents/[id]/page.tsx]: @0ubbe for the usual case of error toasts I would agree with you. The difference here imo is that this isn't an instantaneous error but a *continuous error state*. Like I said, as long as the websocket connection isn't working, the page is in a broken state with various functions not working properly. If we don't actively show this, users will file bug reports because they can't see any indication that we know something is broken.

PR #10454, opened 2025-07-25
fix(backend): Ensure we only present working auth options on blocks

Swiftyos:
To allow for a simpler dev experience, the new SDK now auto discovers providers and registers them. However, the OAuth system was still requiring these credentials to be hardcoded in the settings object.

This PR changes that to verify the env var is present during registration and then allows the OAuth system to load them from the env.

### Changes 🏗️

- **OAuth Registration**: Modified `ProviderBuilder.with_oauth(..)` to check OAuth env vars exist during registration
- **OAuth Loading**: Updated OAuth system to load credentials from env vars if not using secrets
- **Block Filtering**: Added `is_block_auth_configured()` function to check if a block has valid authorization options configured at runtime
- **Test Updates**: Fixed failing SDK registry tests to properly mock environment variables for OAuth registration

#### For code changes:
- [x] I have made a test plan
  - [x] Verified OAuth system checks that env vars exist during provider registration
  - [x] Confirmed OAuth system can use env vars directly without requiring hardcoded secrets
  - [x] Tested that blocks with unconfigured OAuth providers are filtered out
  - [x] All SDK registry tests pass with proper env var mocking

  - OAuth providers now require their client ID and secret env vars to be set for registration
  - No changes required to `.env.example` or `docker-compose.yml`


autogpt_platform/backend/backend/blocks/__init__.py
```diff
@@ -99,7 +103,22 @@ def load_all_blocks() -> dict[str, type["Block"]]:
...
         available_blocks[block.id] = block_cls
 
-    return available_blocks
+    # Filter out blocks with OAuth-only providers when OAuth is not configured
+    from backend.data.block import is_block_provider_configured
+
+    filtered_blocks = {}
+    for block_id, block_cls in available_blocks.items():
+        if not is_block_provider_configured(block_cls):
+            # Skip this block as it requires OAuth which is not configured
```
majdyz: nit: this comment does exactly like the error log says, we can remove it.

autogpt_platform/backend/backend/blocks/__init__.py
```diff
@@ -99,7 +103,22 @@ def load_all_blocks() -> dict[str, type["Block"]]:
 
         available_blocks[block.id] = block_cls
 
-    return available_blocks
+    # Filter out blocks with OAuth-only providers when OAuth is not configured
+    from backend.data.block import is_block_provider_configured
+
+    filtered_blocks = {}
+    for block_id, block_cls in available_blocks.items():
+        if not is_block_provider_configured(block_cls):
```
majdyz: nit: inverted logic is a bit confusing, also it's simpler

```
if is_....:
   filtered_blocks[block_id] = block_cls
else:
   logger.error(...
```

the continue is not even needed

majdyz (commented):
LGTM, but the Copilot comments are valid. @Pwuts review would be needed here

Swiftyos:
> LGTM, but the Copilot comments are valid. @Pwuts review would be needed here

Thank you, all pr comments have been actioned and pushed

autogpt_platform/backend/backend/sdk/registry.py
```diff
@@ -63,27 +64,35 @@ def register_provider(cls, provider: "Provider") -> None:
...
-                # Dynamically set PROVIDER_NAME if not already set
-                if (
-                    not hasattr(provider.oauth_config.oauth_handler, "PROVIDER_NAME")
-                    or provider.oauth_config.oauth_handler.PROVIDER_NAME is None
-                ):
-                    # Import ProviderName to create dynamic enum value
-                    from backend.integrations.providers import ProviderName
+                oauth_env_vars_set = os.getenv(
+                    provider.oauth_config.client_id_env_var
+                ) and os.getenv(provider.oauth_config.client_secret_env_var)
```
Pwuts, line 69: unnecessary nested `if` can be replaced with one (and avoid the indent change):
```suggestion
            if (
                provider.oauth_config
                and os.getenv(provider.oauth_config.client_id_env_var)
                and os.getenv(provider.oauth_config.client_secret_env_var)
            ):
```

autogpt_platform/backend/backend/sdk/registry.py
```diff
@@ -63,27 +64,35 @@ def register_provider(cls, provider: "Provider") -> None:
...
+
+                if oauth_env_vars_set:
```
Pwuts: unnecessary nested `if` can be replaced with one (and avoid the indent change):
```suggestion
            if (
                provider.oauth_config
                and os.getenv(provider.oauth_config.client_id_env_var)
                and os.getenv(provider.oauth_config.client_secret_env_var)
            ):
```

autogpt_platform/backend/backend/data/block.py
```diff
@@ -492,6 +510,127 @@ def get_blocks() -> dict[str, Type[Block]]:
...
+    required_fields = block.input_schema.get_required_fields()
+    required_credentials = [f for f in credentials_fields if f in required_fields]
+
+    if not required_credentials:
+        logger.debug(
+            f"Block {block_cls.__name__} has no required credentials fields - Treating as valid"
+        )
+        return True
+
+    # Check if ANY required credentials field has authentication configured
```
Pwuts: I think this check needs to be:
```suggestion
    # Check if ALL required credentials fields have ≥1 supported auth type configured

File: autogpt_platform/backend/backend/data/block.py [hunk 6]
```diff
@@ -492,6 +510,127 @@ def get_blocks() -> dict[str, Type[Block]]:
...
+    configured_providers = []
+
+    for field_name in required_credentials:
+        provider_name = block.input_schema.get_provider_from_credentials_field(
+            field_name
+        )
+        if not provider_name or provider_name in ProviderName.__members__.values():
```
Pwuts: In what case can `provider_name` be `None` or `""` here?

autogpt_platform/backend/backend/data/block.py
```diff
@@ -492,6 +510,127 @@ def get_blocks() -> dict[str, Type[Block]]:
...
+            )
+            continue
+
+        # Check the provider's supported auth types
+        supported_types = provider.supported_auth_types
+
+        # If provider has no supported auth types, it means no auth method is configured
+        if not supported_types:
+            # This provider has no authentication methods configured at all
+            configured_providers.append(False)
```
Pwuts: If any required credentials field doesn't have any supported auth types configured, the block is unusable, so we can go straight to:
```suggestion
            return False
```

autogpt_platform/backend/backend/data/block.py
```diff
@@ -492,6 +510,127 @@ def get_blocks() -> dict[str, Type[Block]]:
...
+            logger.debug(
+                f"Block {block_cls.__name__} has a required credentials field {field_name} but no authentication methods are configured - Treating as valid"
```
Pwuts: What does "Treating as valid" mean here? It seems to be appended regardless of whether the block is usable or not

autogpt_platform/backend/backend/integrations/creds_manager.py
```diff
@@ -196,8 +197,25 @@ async def _get_provider_oauth_handler(provider_name_str: str) -> "BaseOAuthHandl
     if provider_name not in HANDLERS_BY_NAME:
         raise KeyError(f"Unknown provider '{provider_name}'")
 
-    client_id = getattr(settings.secrets, f"{provider_name.value}_client_id")
-    client_secret = getattr(settings.secrets, f"{provider_name.value}_client_secret")
+    provider_creds = CREDENTIALS_BY_PROVIDER[provider_name]
```
Pwuts, line 200: I see it wasn't introduced in this PR but this naming is mildly confusing because now we have two things called "credentials" in the system

autogpt_platform/backend/backend/sdk/provider.py
```diff
@@ -83,7 +83,7 @@ def credentials_field(self, **kwargs) -> CredentialsMetaInput:
             "credentials_types": (
                 list(self.supported_auth_types)
                 if self.supported_auth_types
-                else ["api_key"]
+                else ["not_configured"]
```
Pwuts: Why not just:
```suggestion
                else []
```
?

Then we don't need a separate check and also don't have to treat `"not_configured"` as if it's an auth type

autogpt_platform/backend/backend/data/block.py
```diff
@@ -166,6 +170,20 @@ def get_required_fields(cls) -> set[str]:
...
+        field_info = cls.model_fields.get(field_name)
+        if not field_info:
+            return None
+
+        json_schema_extra = field_info.json_schema_extra
+        if isinstance(json_schema_extra, dict):
+            providers = json_schema_extra.get("credentials_provider", [])
+            if providers and isinstance(providers, list):
+                return str(providers[0])
+        return None
```
Pwuts: Could be:
```suggestion
        return cls.get_credentials_fields_info()[field_name].provider
```
although this isn't as forgiving to invalid `field_name` inputs

autogpt_platform/backend/backend/data/block.py
```diff
@@ -166,6 +170,20 @@ def get_required_fields(cls) -> set[str]:
...
+        """Extract provider name from a credentials field's schema."""
+        field_info = cls.model_fields.get(field_name)
+        if not field_info:
+            return None
+
+        json_schema_extra = field_info.json_schema_extra
+        if isinstance(json_schema_extra, dict):
+            providers = json_schema_extra.get("credentials_provider", [])
+            if providers and isinstance(providers, list):
+                return str(providers[0])
```
Pwuts: This isn't valid in all cases, because `providers` can contain more than one entry where all should be considered. E.g. check out the blocks and `AICredentialsField` in `blocks/llm.py`.

autogpt_platform/backend/backend/data/model.py
```diff
@@ -302,7 +302,9 @@ def matches_url(self, url: str) -> bool:
 ]
 
 
-CredentialsType = Literal["api_key", "oauth2", "user_password", "host_scoped"]
+CredentialsType = Literal[
+    "api_key", "oauth2", "user_password", "host_scoped", "not_configured"
```
Pwuts: I'd like to avoid adding none-values like this if possible, see also [this comment](https://github.com/Significant-Gravitas/AutoGPT/pull/10454/files#r2239140297)

autogpt_platform/backend/backend/data/block.py
Swiftyos: When the filed name does not exist and if the field is not a credential_provider field

autogpt_platform/backend/backend/data/block.py
Swiftyos: Neither of our attempts here is particularly clear. I've updated it to something that is clear

autogpt_platform/backend/backend/data/block.py
Swiftyos: For this field... there maybe other fields where it does have a correctly configured credentials provider.

autogpt_platform/backend/backend/data/block.py
Swiftyos: error in the message ive updated it

autogpt_platform/backend/backend/integrations/creds_manager.py
Swiftyos: none actionable

autogpt_platform/backend/backend/sdk/provider.py
Swiftyos: Good point, gonna test this

autogpt_platform/backend/backend/data/block.py
Swiftyos: This change completely breaks the system, its needs to be very forgiving so that the default is essentially enabling blocks

autogpt_platform/backend/backend/data/block.py
Swiftyos: I've changed this to return a list of providers and update the rest of the logic to support that

autogpt_platform/backend/backend/sdk/registry.py
Swiftyos: Your suggested edit isn't quite right. However, I have collapsed the  provider.oauth_config and oauth_env_vars_set into a single if statement

Pwuts: Yes the generated diff looks weird, I tried to fix it by changing the selected lines but that didn't work

autogpt_platform/backend/backend/data/block.py
```diff
@@ -492,6 +510,135 @@ def get_blocks() -> dict[str, Type[Block]]:
...
+
+    For example if a block is an OAuth-only block and there env vars are not set,
+    do not show it in the UI.
+
+    """
+    from backend.sdk.registry import AutoRegistry
+
+    # Create an instance to access input_schema
+    try:
+        block = block_cls.create()
```
Pwuts: `create()` is just a wrapper for `__init__`, but with a less well-defined type, causing type checking issues (`block.input_schema` currently resolves to `Unknown` in the code below)
```suggestion
        block = block_cls()
```
I have checked that `create()` isn't overridden by any block implementations

autogpt_platform/backend/backend/data/block.py
```diff
@@ -492,6 +510,135 @@ def get_blocks() -> dict[str, Type[Block]]:
     return load_all_blocks()
 
 
+def is_block_provider_configured(block_cls: type["Block"]) -> bool:
```
Pwuts: Using `Block` without type arguments leaves `Block.input_schema` and `Block.output_schema` undefined
```suggestion
def is_block_provider_configured(
    block_cls: type["Block[BlockSchema, BlockSchema]"]
) -> bool:
```

autogpt_platform/backend/backend/data/block.py
```diff
@@ -492,6 +510,135 @@ def get_blocks() -> dict[str, Type[Block]]:
...
+    if not credentials_fields:
+        logger.debug(
+            f"Block {block_cls.__name__} has no credentials fields - Treating as valid"
+        )
+        return True
+
+    # Check required credentials fields
+    required_fields: dict[str, CredentialsMetaInput] = (
+        block.input_schema.get_required_fields()
+    )
```
Pwuts: Fixing the above mentioned type-checking issues revealed this issue:
image
```suggestion
    required_fields: set[str] = block.input_schema.get_required_fields()
```

autogpt_platform/backend/backend/data/block.py
```diff
@@ -492,6 +510,135 @@ def get_blocks() -> dict[str, Type[Block]]:
...
+    required_credentials = [f for f in credentials_fields if f in required_fields]
+
+    if len(required_credentials) == 0:
+        logger.warn(
+            f"Block {block_cls.__name__} has no required credentials fields - therefore will worth without credentials configured"
+        )
```
Pwuts: 1. not a problem
2. typo
```suggestion
        logger.debug(
            f"Block {block_cls.__name__} has no required credentials fields - therefore will work without credentials configured"
        )
```

although: if a credentials input is not required and also not configured, should it be hidden? do we have a mechanism for this?

autogpt_platform/backend/backend/data/block.py
```diff
@@ -492,6 +510,135 @@ def get_blocks() -> dict[str, Type[Block]]:
...
+        return False
+    if len(required_credentials) > 1:
+        logger.warn(
+            f"Block {block_cls.__name__} has multiple required credentials fields"
+        )
```
Pwuts: not a problem, and imo not worth logging
```suggestion
```

autogpt_platform/backend/backend/data/block.py
```diff
@@ -492,6 +510,135 @@ def get_blocks() -> dict[str, Type[Block]]:
...
+                    f"Block {block_cls.__name__} is part of the legacy provider system - Treating as valid"
+                )
+                configured_providers[field_name] = [True]
+                break
+
+            provider = AutoRegistry.get_provider(provider_name)
+            if not provider:
+                logger.debug(
+                    f"Block {block_cls.__name__} has a required credentials field {field_name} that has no provider - Treating as valid"
+                )
```
Pwuts: In what case would this happen? Wouldn't this be a breaking issue, if we have an active block with no backing registered provider?

Swiftyos: Not exactly as it maybe configured via the ProviderEnum based credentials system. This really is focused on the new autogregistered sdk system

autogpt_platform/backend/backend/data/block.py
Swiftyos: Its not but I wanted to know when it occurs, so when it does we can test carefully

autogpt_platform/backend/backend/data/block.py
Swiftyos: we are only checking credentials  if they are required

autogpt_platform/backend/backend/data/block.py
Pwuts: In such cases this check isn't accessible, right? Because of the check above `provider_name in ProviderName.__members__.values()`?

autogpt_platform/backend/backend/sdk/registry.py
```diff
@@ -62,22 +64,22 @@ def register_provider(cls, provider: "Provider") -> None:
             cls._providers[provider.name] = provider
 
             # Register OAuth handler if provided
-            if provider.oauth_config:
+            if (
+                provider.oauth_config
+                and os.getenv(provider.oauth_config.client_id_env_var)
+                and os.getenv(provider.oauth_config.client_secret_env_var)
+            ):
```
Pwuts: I guess just `if provider.oauth_config:` suffices here now :)

Swiftyos: yeah true that, fixing

autogpt_platform/backend/backend/data/block.py
```diff
@@ -492,6 +510,135 @@ def get_blocks() -> dict[str, Type[Block]]:
...
+
+    # Check required credentials fields
+    required_fields: set[str] = block.input_schema.get_required_fields()
+    required_credentials = [f for f in credentials_fields if f in required_fields]
+
+    if len(required_credentials) == 0:
+        logger.debug(
+            f"Block {block_cls.__name__} has no required credentials fields - therefore will work without credentials configured"
+        )
+        return False
```
Pwuts: I think this is something we want to log but still go through with the other checks, right? In any case, not a reason to disable a block?
```suggestion
```

autogpt_platform/backend/backend/data/block.py
Swiftyos: It should not be, however, the types checking complains if I remove it as provider could be none

autogpt_platform/backend/backend/data/block.py
Swiftyos: Actually if none of the credentials are required, we should be returning true as it does not matter to the system if they are configured or not by virtue of them not being required

autogpt_platform/backend/backend/data/block.py
```diff
@@ -492,6 +510,136 @@ def get_blocks() -> dict[str, Type[Block]]:
...
+
+    # Check required credentials fields
+    required_fields: set[str] = block.input_schema.get_required_fields()
+    required_credentials = [f for f in credentials_fields if f in required_fields]
+
+    if len(required_credentials) == 0:
+        logger.debug(
+            f"Block {block_cls.__name__} has no required credentials fields - therefore will work without credentials configured"
+        )
+        return True
```
Pwuts: Skipping the checks seems risky: it allows misconfigured credential inputs to still show up in the UI if optional.

autogpt_platform/backend/backend/data/block.py
```diff
@@ -492,6 +510,136 @@ def get_blocks() -> dict[str, Type[Block]]:
...
+                logger.debug(
+                    f"Block {block_cls.__name__} has a required credentials field {field_name} that has a provider - Treating as valid"
+                )
+                configured_providers[field_name].append(True)
+
+    is_provider_configured = {
+        field_name: any(configured_providers[field_name])
+        for field_name in required_credentials
+    }
+    return all(is_provider_configured.values())
```
Pwuts: nit: can be simplified quite a bit with early exit:
```suggestion
    # Check if the credential inputs for this block are correctly configured
    for field_name in required_credentials:
        provider_names = block.input_schema.get_provider_from_credentials_field(
            field_name
        )
        if not provider_names:
            logger.warning(
                f"Block {block_cls.__name__} "
                f"has required credentials field {field_name} with no providers"
            )
            return False

        # If a field has multiple possible providers, each one needs to be usable to
        # prevent breaking the UX
        for provider_name in provider_names:
            if provider_name in ProviderName.__members__.values():
                logger.debug(
                    f"Block {block_cls.__name__} input {field_name} provider {provider_name} is part of the legacy provider system - Treating as valid"
                )
                break

            provider = AutoRegistry.get_provider(provider_name)
            if not provider:
                logger.warning(
                    f"Block {block_cls.__name__} has a required credentials field {field_name} with no provider options - Disabling"
                )
                return False

            # Check the provider's supported auth types
            supported_types = provider.supported_auth_types

            # If provider has no supported auth types, it means no auth method is configured
            if not supported_types:
                logger.warning(
                    f"Block {block_cls.__name__} has a required credentials field {field_name} but provider {provider_name} has no authentication methods configured - Disabling"
                )
                return False

            # Check if provider supports OAuth
            if "oauth2" in supported_types:
                # Check if OAuth environment variables are set
                if provider.oauth_config:
                    oauth_configured = bool(
                        os.getenv(provider.oauth_config.client_id_env_var)
                        and os.getenv(provider.oauth_config.client_secret_env_var)
                    )
                    if not oauth_configured:
                        logger.warning(
                            f"Block {block_cls.__name__} has a required credentials field {field_name} but provider {provider_name} is missing OAuth credentials (client ID or client secret) - Disabling"
                        )
                        return False

        logger.debug(
            f"Block {block_cls.__name__} credentials field {field_name} is valid; "
            f"supported credential types: {supported_types}"
        )

    return True
```

autogpt_platform/backend/backend/data/block.py
Swiftyos: However, they are marked as not required. If we do the checks we are implicitly stating they are required going against the block creators design

autogpt_platform/backend/backend/data/block.py
Swiftyos: This change would prevent the showing the block where both api_key and Oauth credentials are allowed and OAuth is not configured. 

We explicitly want to allow this case

autogpt_platform/backend/backend/data/block.py
```diff
@@ -492,6 +510,136 @@ def get_blocks() -> dict[str, Type[Block]]:
...
+    # Check if theres is any correctly configured credential provider for this block
+    configured_providers = {field_name: [] for field_name in required_credentials}
+
+    for field_name in required_credentials:
+        provider_names = block.input_schema.get_provider_from_credentials_field(
+            field_name
+        )
+        if not provider_names:
+            configured_providers[field_name].append(True)
+            continue
```
Pwuts: If this happens, the block has a misconfigured credentials input, which we probably don't want to show in the UI:
```suggestion
        provider_names = block.input_schema.get_provider_from_credentials_field(
            field_name
        )
        if not provider_names:
            logger.warning(
                f"Block {block_cls.__name__} "
                f"has required credentials field {field_name} with no providers"
            )
            return False
```

autogpt_platform/backend/backend/data/block.py
Pwuts: Afaik this is taken care of by `ProviderBuilder.with_oauth(..)` now, right?

ntindle (approved):

PR #10490, opened 2025-07-30
feat(backend/AM): Integrate AutoMod content moderation

Bentlybro:
Adds AutoMod input and output moderation to the execution flow. Introduces a new AutoMod manager and models, updates settings for moderation configuration, and modifies execution result handling to support moderation-cleared data. Moderation failures now clear sensitive data and mark executions as failed.

### Changes 🏗️

I have made some small changes to ``autogpt_platform\backend\backend\executor\manager.py`` to send the needed into to the AutoMod system which collects the data, combines and makes the api call to AM and based on its reply lets it run or not!

I also had to make small changes to ``autogpt_platform\backend\backend\data\execution.py`` to add checks that allow me to clear the content from the blocks if it was flagged

I am working on finalizing the AM repo then that will be public

To note: we will want to set this up behind launch darkly first for testing on the team before we roll it out any more

#### For code changes:
- [x] I have made a test plan
  
  - [x] Setup and run the platform with ``automod_enabled`` set to False and it works normally
  - [x] Setup and run the platform with ``automod_enabled`` set to True, set the AM URL and API Key and test it runs safe blocks normally
  - [x] Test AM with content that would trigger it to flag and watch it stop and clear all the blocks outputs

Message @Bentlybro for the URL and an API key to AM for local testing!

## Changes made to Settings.py 

I have added a few new options to the settings.py for AutoMod Config!

```
    # AutoMod configuration
    automod_enabled: bool = Field(
        default=False,
        description="Whether AutoMod content moderation is enabled",
    )
    automod_api_url: str = Field(
        default="",
        description="AutoMod API base URL - Make sure it ends in /api",
    )
    automod_timeout: int = Field(
        default=30,
        description="Timeout in seconds for AutoMod API requests",
    )
    automod_retry_attempts: int = Field(
        default=3,
        description="Number of retry attempts for AutoMod API requests",
    )
    automod_retry_delay: float = Field(
        default=1.0,
        description="Delay between retries for AutoMod API requests in seconds",
    )
    automod_fail_open: bool = Field(
        default=False,
        description="If True, allow execution to continue if AutoMod fails",
    )
    automod_moderate_inputs: bool = Field(
        default=True,
        description="Whether to moderate block inputs",
    )
    automod_moderate_outputs: bool = Field(
        default=True,
        description="Whether to moderate block outputs",
    )
```
and
```
automod_api_key: str = Field(default="", description="AutoMod API key")
```

### DIFF
--- a/autogpt_platform/frontend/src/app/(platform)/auth/auth-code-error/page.tsx
+++ b/autogpt_platform/frontend/src/app/(platform)/auth/auth-code-error/page.tsx
@@ -1,5 +1,6 @@
 "use client";
 
+import { isServerSide } from "@/lib/utils/is-server-side";
 import { useEffect, useState } from "react";
 
 export default function AuthErrorPage() {
@@ -9,7 +10,7 @@ export default function AuthErrorPage() {
 
   useEffect(() => {
     // This code only runs on the client side
-    if (typeof window !== "undefined") {
+    if (!isServerSide()) {
       const hash = window.location.hash.substring(1); // Remove the leading '#'
       const params = new URLSearchParams(hash);
 
--- a/autogpt_platform/frontend/src/app/api/mutators/custom-mutator.ts
+++ b/autogpt_platform/frontend/src/app/api/mutators/custom-mutator.ts
@@ -1,6 +1,22 @@
+import {
+  createRequestHeaders,
+  getServerAuthToken,
+} from "@/lib/autogpt-server-api/helpers";
+import { isServerSide } from "@/lib/utils/is-server-side";
+
 const FRONTEND_BASE_URL =
   process.env.NEXT_PUBLIC_FRONTEND_BASE_URL || "http://localhost:3000";
 const API_PROXY_BASE_URL = `${FRONTEND_BASE_URL}/api/proxy`; // Sending request via nextjs Server
+
+const getBaseUrl = (): string => {
+  if (!isServerSide()) {
+    return API_PROXY_BASE_URL;
+  } else {
+    return (
+      process.env.NEXT_PUBLIC_AGPT_SERVER_BASE_URL || "http://localhost:8006"
+    );
+  }
+};
 
 const getBody = <T>(c: Response | Request): Promise<T> => {
   const contentType = c.headers.get("content-type");
@@ -30,11 +46,12 @@ export const customMutator = async <T = any>(
     | "DELETE"
     | "PATCH";
   const data = requestOptions.body;
-  const headers: Record<string, string> = {
+  let headers: Record<string, string> = {
     ...((requestOptions.headers as Record<string, string>) || {}),
   };
 
   const isFormData = data instanceof FormData;
+  const contentType = isFormData ? "multipart/form-data" : "application/json";
 
   // Currently, only two content types are handled here: application/json and multipart/form-data
   if (!isFormData && data && !headers["Content-Type"]) {
@@ -45,12 +62,40 @@ export const customMutator = async <T = any>(
     ? "?" + new URLSearchParams(params).toString()
     : "";
 
-  const response = await fetch(`${API_PROXY_BASE_URL}${url}${queryString}`, {
+  const baseUrl = getBaseUrl();
+
+  // The caching in React Query in our system depends on the url, so the base_url could be different for the server and client sides.
+  const fullUrl = `${baseUrl}${url}${queryString}`;
+
+  if (isServerSide()) {
+    try {
+      const token = await getServerAuthToken();
+      const authHeaders = createRequestHeaders(token, !!data, contentType);
+      headers = { ...headers, ...authHeaders };
+    } catch (error) {
+      console.warn("Failed to get server auth token:", error);
+    }
+  }
+
+  const response = await fetch(fullUrl, {
     ...requestOptions,
     method,
     headers,
     body: data,
   });
+
+  // Error handling for server-side requests
+  // We do not need robust error handling for server-side requests; we only need to log the error message and throw the error.
+  // What happens if the server-side request fails?
+  // 1. The error will be logged in the terminal, then.
+  // 2. The error will be thrown, so the cached data for this particular queryKey will be empty, then.
+  // 3. The client-side will send the request again via the proxy. If it fails again, the error will be handled on the client side.
+  // 4. If the request succeeds on the server side, the data will be cached, and the client will use it instead of sending a request to the proxy.
+
+  if (!response.ok && isServerSide()) {
+    console.error("Request failed on server side", response);
+    throw new Error(`Request failed with status ${response.status}`);
+  }
 
   const response_data = await getBody<T>(response);
 
--- a/autogpt_platform/frontend/src/components/auth/Turnstile.tsx
+++ b/autogpt_platform/frontend/src/components/auth/Turnstile.tsx
@@ -1,5 +1,6 @@
 "use client";
 import { cn } from "@/lib/utils";
+import { isServerSide } from "@/lib/utils/is-server-side";
 import { useEffect, useRef, useState } from "react";
 
 export interface TurnstileProps {
@@ -31,7 +32,7 @@ export function Turnstile({
 
   // Load the Turnstile script
   useEffect(() => {
-    if (typeof window === "undefined" || !shouldRender) return;
+    if (isServerSide() || !shouldRender) return;
 
     // Skip if already loaded
     if (window.turnstile) {
--- a/autogpt_platform/frontend/src/components/contextual/InfiniteScroll/useInfiniteScroll.ts
+++ b/autogpt_platform/frontend/src/components/contextual/InfiniteScroll/useInfiniteScroll.ts
@@ -1,3 +1,4 @@
+import { isServerSide } from "@/lib/utils/is-server-side";
 import { debounce } from "lodash";
 import { useCallback, useEffect, useRef, useState } from "react";
 
@@ -24,7 +25,7 @@ export const useInfiniteScroll = ({
   const [isLoading, setIsLoading] = useState(false);
 
   const handleScroll = useCallback(() => {
-    if (containerRef.current && typeof window !== "undefined") {
+    if (containerRef.current && !isServerSide()) {
       const container = containerRef.current;
       const { bottom } = container.getBoundingClientRect();
       const { innerHeight } = window;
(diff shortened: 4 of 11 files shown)


AutoGPT-Agent:
Thanks for implementing the AutoMod content moderation feature! This is a valuable addition for content safety. Here's my feedback:

### Positive Aspects
- The code is well-structured with clear separation of concerns (manager, models)
- Comprehensive test plan that covers various scenarios
- Good error handling with retry logic
- Clear documentation in settings fields
- Proper handling of user_id throughout the code

### Issues to Address

1. **Merge Conflicts**: There are conflicts that need to be resolved before this can be merged.

2. **Documentation**: Consider adding a brief comment at the top of the AutoMod/manager.py file explaining the purpose and functionality of the AutoMod system.

3. **Edge Cases**: Have you considered how this would behave in these scenarios:
   - What happens if AutoMod API is temporarily unavailable?
   - Are there timeouts that could affect user experience?
   - How does this impact performance for normal executions?

4. **Testing**: Once ready, we should test this behind Launch Darkly as you mentioned.

Great work overall! Let me know if you need any clarification on the feedback.

autogpt_platform/backend/backend/data/execution.py
```diff
@@ -317,6 +317,9 @@ class NodeExecutionResult(BaseModel):
 
     @staticmethod
     def from_db(_node_exec: AgentNodeExecution, user_id: Optional[str] = None):
+        # Check if this execution was cleared due to moderation
+        stats = type_utils.convert(_node_exec.stats, dict) if _node_exec.stats else {}
```
majdyz: Are we clearing stats too?  why?

majdyz: or the comment is just confusing

autogpt_platform/backend/backend/data/execution.py
```diff
@@ -326,9 +329,22 @@ def from_db(_node_exec: AgentNodeExecution, user_id: Optional[str] = None):
             for data in _node_exec.Input or []:
                 input_data[data.name] = type_utils.convert(data.data, type[Any])
 
+        # Check if inputs should be cleared due to moderation
+        if stats.get("moderation_cleared") and stats.get("cleared_inputs"):
```
majdyz: Why do we have two fields? Can we just use cleared_inputs?
Also, please add the field to the NodeExecutionStats / GraphExecutionStats. We should avoid accessing the dict by string and use Pydantic.

Use `NodeExecutionStats.model_validate(_node_exec.stats or {})` then add cleared_inputs or whaterver field we want to introduce.

autogpt_platform/backend/backend/data/execution.py
```diff
@@ -326,9 +329,22 @@ def from_db(_node_exec: AgentNodeExecution, user_id: Optional[str] = None):
...
+            # Replace input data with moderation messages
+            for name, message in stats["cleared_inputs"].items():
+                input_data[name] = message
+
         output_data: CompletedBlockOutput = defaultdict(list)
-        for data in _node_exec.Output or []:
-            output_data[data.name].append(type_utils.convert(data.data, type[Any]))
+
+        if stats.get("moderation_cleared") and stats.get("cleared_outputs"):
```
majdyz: Same, is `moderation_cleared` needed?

autogpt_platform/backend/backend/executor/manager.py
```diff
@@ -736,6 +737,20 @@ def on_done_task(node_exec_id: str, result: object):
...
+            # Input moderation
+            moderation_success, moderation_error = (
+                automod_manager.moderate_graph_execution_inputs(
+                    db_client=db_client,
+                    graph_exec=graph_exec,
+                    event_loop=cls.node_evaluation_loop,
+                    send_update_func=send_execution_update,
+                )
+            )
+            if not moderation_success:
```
majdyz: Let's do the same as above, introduce a new error similar to InsufficientBalanceError, and raise ModerationError.
now there is this line:
```
known_errors = (InsufficientBalanceError,)
```
then you can add the moderation error there, so we know which error is expected and which one are not without breaking the cleanup flow.

autogpt_platform/backend/backend/executor/manager.py
```diff
@@ -864,6 +879,23 @@ def on_done_task(node_exec_id: str, result: object):
...
+                    user_id=graph_exec.user_id,
+                    graph_id=graph_exec.graph_id,
+                    event_loop=cls.node_evaluation_loop,
+                    send_update_func=send_execution_update,
+                )
+            )
+            if not moderation_success:
+                execution_status = ExecutionStatus.FAILED
+                error = moderation_error
+                return execution_stats, execution_status, error
```
majdyz: Same, also we can make the`moderate_graph_execution_outputs` returning content that we want to update on the stats, and just let `execution_stats` be updated before raising the error. 

This way you don't need to do db update inside your moderation function.

autogpt_platform/backend/.env.example
```diff
@@ -121,6 +121,18 @@ LINEAR_CLIENT_SECRET=
 TODOIST_CLIENT_ID=
 TODOIST_CLIENT_SECRET=
 
+
+# AutoMods Configuration
+AUTOMOD_ENABLED=false
```
majdyz: Are these defaults set in settings.py? If so, maybe we don't need to list it here

Because we also have a Docker-Compose env we need to track and putting in two places is kinda tedious.

autogpt_platform/backend/backend/server/v2/AutoMod/manager.py
```diff
@@ -0,0 +1,296 @@
...
+            retry_attempts=settings.config.automod_retry_attempts,
+            retry_delay=settings.config.automod_retry_delay,
+            fail_open=settings.config.automod_fail_open,
+            moderate_inputs=settings.config.automod_moderate_inputs,
+            moderate_outputs=settings.config.automod_moderate_outputs,
+        )
+
+    def moderate_graph_execution_inputs(
+        self, db_client, graph_exec, event_loop, send_update_func, timeout: int = 10
+    ) -> Tuple[bool, Exception | None]:
```
majdyz: Can we just return `Exception | None` without the bool? I think is kinda inferred if the exception is none then the moderation passed.

autogpt_platform/backend/backend/server/v2/AutoMod/manager.py
```diff
@@ -0,0 +1,296 @@
...
+        """
+        Complete input moderation flow for graph execution
+        Returns: (success, error_if_failed)
+        """
+        if not self.config.moderate_inputs or not self.config.enabled:
```
majdyz: Why do we need to have two flags? If we want to toggle off the input & output moderation why do we still have another knob `self.config.enabled` ?

autogpt_platform/backend/backend/server/v2/AutoMod/manager.py
```diff
@@ -0,0 +1,296 @@
...
+                )
+                return False, Exception("Execution failed due to content moderation")
+
+            return True, None
+
+        except Exception as e:
+            logger.error(f"Output moderation execution failed: {e}")
+            return False, Exception("Execution failed due to content moderation")
+
+    def handle_moderation_failure(self, db_client, executions: List, send_update_func):
```
majdyz: Why do we need to do this? If we clear **all** input & output anyway, why don't we just mark the error and graph level, and not return any input & output when the graph is marked as failing moderation?

This way, we can't avoid updating the db entirely. Plus, no need to call `send_update_func` since raising an Error will do this. (just make sure to update execution_stats before raising the error so it will get updated.

autogpt_platform/backend/backend/server/v2/AutoMod/manager.py
```diff
@@ -0,0 +1,296 @@
...
+                "cleared_inputs": cleared_inputs,
+            },
+        )
+
+        # Send websocket updates for each failed node to update the UI
+        for exec_entry in executions:
+            if updated_exec := db_client.get_node_execution(exec_entry.node_exec_id):
+                send_update_func(updated_exec)
+
+    def _extract_text_simple(self, data: Any) -> str:
```
majdyz: where is this function used ?

autogpt_platform/backend/backend/server/v2/AutoMod/manager.py
```diff
@@ -0,0 +1,296 @@
...
+
+        if not all_inputs:
+            return True, None
+
+        # Combine all content and moderate directly
+        content = " ".join(all_inputs)
+
+        # Run moderation
+        try:
+            moderation_passed = asyncio.run_coroutine_threadsafe(
```
majdyz: We end up mixing db_client that is sync and async functions here. We will lose the benefit of async but will get its complexity. Let's just make the function async all the way, change db_client to async db_client (there is get_db_async_client()). and do this `asyncio.run_coroutine_threadsafe` on `moderate_graph_execution_inputs` function call.

autogpt_platform/backend/backend/server/v2/AutoMod/manager.py
```diff
@@ -0,0 +1,296 @@
...
+                )
+                return False, Exception("Execution failed due to content moderation")
+
+            return True, None
+
+        except Exception as e:
+            logger.error(f"Input moderation execution failed: {e}")
+            return False, Exception("Execution failed due to content moderation")
+
+    def moderate_graph_execution_outputs(
```
majdyz: Similar comment here as the input counterparts

autogpt_platform/backend/backend/server/v2/AutoMod/manager.py
```diff
@@ -0,0 +1,296 @@
...
+
+    async def _make_request(self, request_data: AutoModRequest) -> AutoModResponse:
+        """Make HTTP request to AutoMod API"""
+        url = f"{self.config.api_url}/moderate"
+        headers = {
+            "Content-Type": "application/json",
+            "X-API-Key": self.config.api_key,
+        }
+
+        timeout = aiohttp.ClientTimeout(total=self.config.timeout)
```
majdyz: You can use request.py for this, it has all of this handling

majdyz (changes requested):
Nice work, but we need to align with the current state of our execution engine so some adjustments are needed.

autogpt_platform/backend/backend/data/execution.py
Bentlybro: Yes its to clear the input and output of every block in a run if any are flagged by the moderation, do we have a better way to do it? if so please let me know, comment is probs confusing 

here we are just extracting the execution stats and moderation metadata, ill try clean it up some more

autogpt_platform/backend/.env.example
Bentlybro: yeah they are in settings.py, ill remove from the env

Bentlybro:
Moved to [feat(backend/AM): Integrate AutoMod content moderation - By Bentlybro - PR #10539](https://github.com/Significant-Gravitas/AutoGPT/pull/10539) as i broke this PR

Issue #9145, opened 2024-12-30
AI Music Generator Block Fails, but Replicate shows Successful Generation


Torantulino:
> The AI Music Generator block showed that it failed after 3 attempts, but Replicate shows that it created 3 audio files

This is a user reported issue. It is unknown whether this happens every time or intermittently.<br><br>**Steps to Reproduce:**

1. Add the AI Music Generator Block to an Agent
2. Run the Agent
3. Observe output being Error as pictured below
4. Check the generation logs on the replicate platform

<img src="https://uploads.linear.app/a47946b5-12cd-4b3d-8822-df04c855879f/2e989dd1-479b-40b6-bcbc-55872d9ed1e6/01e97413-6422-431a-9e86-5ee59f0e1390?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiL2E0Nzk0NmI1LTEyY2QtNGIzZC04ODIyLWRmMDRjODU1ODc5Zi8yZTk4OWRkMS00NzliLTQwYjYtYmNiYy01NTg3MmQ5ZWQxZTYvMDFlOTc0MTMtNjQyMi00MzFhLTllODYtNWVlNTlmMGUxMzkwIiwiaWF0IjoxNzM1NTUzOTk3LCJleHAiOjMzMzA2MTEzOTk3fQ.HtOvHZd5j3Ij2Zuq6rICFUJU5x6WV36AJLSiFG3hEBs " alt="Screenshot_2024-12-29_at_7.40.16_PM.png" width="637" data-linear-height="308" />

[...]

PR #10533, opened 2025-08-03
fix(blocks): handle invalid or empty response from MusicGen model

mitanshhh:
Handle invalid or empty response from MusicGen model

Fixes: #9145
> ⚠️ Note: This PR does not directly fix issue #9145 (failed run marked as success), but improves the validation of the URL to reduce the chances of invalid states entering the system. This is a related improvement, but not the root cause fix.

### Description
During execution of the meta/musicgen model via Replicate API, the application failed
with an error indicating the model returned an empty or invalid response.
Although some API calls succeeded, this error showed the logic was not checking the
structure and content of the result properly before processing it.

PROBLEM:
CONTEXT:
API: Replicate
MODEL: meta/musicgen:671ac645
STATUS: Failed after 3 attempts
ERROR_MESSAGE: "Unexpected error: Model returned empty or invalid response"
CAUSE:
- The original logic did not validate result structure.
- It assumed any non-null output was valid, including strings like "No output received".
- This led to invalid/malformed results being passed to the frontend.

### Changes 🏗️

- Added `AIMusicGeneratorBlock` to support music generation using Meta’s MusicGen models via Replicate API.
- Supports configurable inputs like prompt, model version, duration, temperature, top_k/p, and normalization.
- Uses robust retry logic for reliability.
- Output returns audio URL; errors return user-friendly message.

BEFORE_CODE: |
```
if result and result != "No output received":
     yield "result", result
     return
```

AFTER_CODE: |

```
if result and isinstance(result, str) and result.startswith("http"):
      yield "result", result
      return
```

#### For code changes:
- [x] Clearly listed changes in the PR description
- [x] Added test plan and mock outputs
- [x] Tested with various prompts and confirmed working output

### Test Plan

- [x] Ran locally with valid Replicate API key
- [x] Generated audio with different prompts
- [x] Simulated failure to verify retry and error message


mitanshhh:
@majdyz @Swiftyos  Please review when you get a chance. Let me know if any changes are needed!
The error was due to the original logic which did not validate result structure.

majdyz:
@mitanshhh thank you for the contribution! A question about the PR:
Is this addressing https://github.com/Significant-Gravitas/AutoGPT/issues/9145 (success run marked as failed run)? If so, how come a stricter validation will help solve the issue? I think it makes sense to address the issue where a failed run is marked as a successful run. 

It is still a good change that we should keep it, but would you mind either:
* Change the description to tell that this doesn't fix the issue mentioned, or
* Complement the change with a fix that could solve the initial issue?

majdyz (approved):

mitanshhh:
Thanks for pointing that out! You're right — this change improves validation, but doesn't directly resolve #9145. I've updated the PR description to reflect that, and removed the issue reference to avoid confusion. Let me know if you'd prefer I keep it linked as a related improvement.

PR #10537, opened 2025-08-04
feat(backend): Enable Ayrshare TikTok support

Swiftyos:
## Summary
- Enabled the TikTok posting block that was previously disabled
- The block provides comprehensive TikTok-specific posting options

## Changes 🏗️
- Removed `disabled=True` from TikTok posting block to enable functionality  
- Added full TikTok API integration with all supported options:

#### For code changes:
- [x] I have made a test plan
  - [x] Verified YouTube block is now available in the block list


AutoGPT-Agent:
Thank you for your contribution to enable TikTok support in the Ayrshare integration! The implementation looks good with improved typing for the visibility options and better descriptions for the various TikTok-specific options.

I noticed one small issue in your description - you mentioned removing `disabled=True` from the YouTube posting block, but the changes actually remove this flag from the TikTok block. Could you please update your description to reflect this?

Your test plan verifies that the block is now available, which is appropriate for this change. The enum implementation for TikTok visibility options is a nice improvement that helps with type safety.

Once you've fixed the description, this PR should be good to go!

ntindle:
We can go ahead and enable all of these behind ld if we want?

Swiftyos:
We can but we are doing them 1 at a time as each one's needs to be carefully checked that it's custom options have been set correctly

majdyz (approved):

PR #10539, opened 2025-08-04
feat(backend/AM): Integrate AutoMod content moderation

Bentlybro:
Adds AutoMod input and output moderation to the execution flow. Introduces a new AutoMod manager and models, updates settings for moderation configuration, and modifies execution result handling to support moderation-cleared data. Moderation failures now clear sensitive data and mark executions as failed.

### Changes 🏗️

I have made some small changes to ``autogpt_platform\backend\backend\executor\manager.py`` to send the needed into to the AutoMod system which collects the data, combines and makes the api call to AM and based on its reply lets it run or not!

I also had to make small changes to ``autogpt_platform\backend\backend\data\execution.py`` to add checks that allow me to clear the content from the blocks if it was flagged

I am working on finalizing the AM repo then that will be public

To note: we will want to set this up behind launch darkly first for testing on the team before we roll it out any more

#### For code changes:
- [x] I have made a test plan
  
  - [x] Setup and run the platform with ``automod_enabled`` set to False and it works normally
  - [x] Setup and run the platform with ``automod_enabled`` set to True, set the AM URL and API Key and test it runs safe blocks normally
  - [x] Test AM with content that would trigger it to flag and watch it stop and clear all the blocks outputs

Message @Bentlybro for the URL and an API key to AM for local testing!

## Changes made to Settings.py 

I have added a few new options to the settings.py for AutoMod Config!

```
    # AutoMod configuration
    automod_enabled: bool = Field(
        default=False,
        description="Whether AutoMod content moderation is enabled",
    )
    automod_api_url: str = Field(
        default="",
        description="AutoMod API base URL - Make sure it ends in /api",
    )
    automod_timeout: int = Field(
        default=30,
        description="Timeout in seconds for AutoMod API requests",
    )
    automod_retry_attempts: int = Field(
        default=3,
        description="Number of retry attempts for AutoMod API requests",
    )
    automod_retry_delay: float = Field(
        default=1.0,
        description="Delay between retries for AutoMod API requests in seconds",
    )
    automod_fail_open: bool = Field(
        default=False,
        description="If True, allow execution to continue if AutoMod fails",
    )
    automod_moderate_inputs: bool = Field(
        default=True,
        description="Whether to moderate block inputs",
    )
    automod_moderate_outputs: bool = Field(
        default=True,
        description="Whether to moderate block outputs",
    )
```
and
```
automod_api_key: str = Field(default="", description="AutoMod API key")
```


ntindle (dismissed):

autogpt_platform/backend/backend/data/execution.py
```diff
@@ -317,6 +317,13 @@ class NodeExecutionResult(BaseModel):
 
     @staticmethod
     def from_db(_node_exec: AgentNodeExecution, user_id: Optional[str] = None):
+        # Parse execution stats using proper Pydantic model instead of dict access
+        try:
+            stats = NodeExecutionStats.model_validate(_node_exec.stats or {})
+        except (ValueError, ValidationError):
+            # Fallback to empty stats if parsing fails
```
majdyz: nit: I don't think the two comments above are needed.

autogpt_platform/backend/backend/data/execution.py
```diff
@@ -326,9 +333,23 @@ def from_db(_node_exec: AgentNodeExecution, user_id: Optional[str] = None):
...
+
+        # Apply moderation output clearing if present
+        if stats.cleared_outputs:
+            # Use the cleared outputs instead of actual output data
+            for name, message in stats.cleared_outputs.items():
+                output_data[name].append(message)
+        else:
+            # Normal case - use actual output data
+            for data in _node_exec.Output or []:
+                output_data[data.name].append(type_utils.convert(data.data, type[Any]))
```
majdyz, line 344: So, input data is only replacing specific keys and leaving the other remain uncleared?
And for output data, it's entirely overriding the content?

There seems to be a slight inconsistency.

autogpt_platform/backend/backend/data/model.py
```diff
@@ -655,6 +655,9 @@ class NodeExecutionStats(BaseModel):
     output_token_count: int = 0
     extra_cost: int = 0
     extra_steps: int = 0
+    # Moderation fields
+    cleared_inputs: Optional[dict[str, str]] = None
+    cleared_outputs: Optional[dict[str, str]] = None
```
majdyz: output is a stream of values (a block can yield multiple times) for the same name. Is this okay to have the normal dict?

autogpt_platform/backend/backend/executor/manager.py
```diff
@@ -899,14 +924,21 @@ def _on_graph_execution(
...
-            known_errors = (InsufficientBalanceError,)
+            known_errors = (InsufficientBalanceError, ModerationError)
             if isinstance(error, known_errors):
                 execution_stats.error = str(error)
                 return ExecutionStatus.FAILED
 
-            log_metadata.exception(
-                f"Failed graph execution {graph_exec.graph_exec_id}: {error}"
-            )
+            if isinstance(exc, CancelledError):
```
majdyz: Have you pulled the latest dev? I think this change is not expected.
Or maybe just a resolving conflict issue.

autogpt_platform/backend/backend/server/v2/AutoMod/manager.py
```diff
@@ -0,0 +1,305 @@
...
+    ) -> Exception | None:
+        """
+        Complete input moderation flow for graph execution
+        Returns: error_if_failed (None means success)
+        """
+        if not self.config.enabled:
+            return None
+
+        # Check if AutoMod feature is enabled for this user
+        if not is_feature_enabled("AutoMod", graph_exec.user_id, default=False):
```
majdyz, line 54: Nice, let's keep it this way for now. You can target using user_id, but unfortunately, the segment/email target is still not working. 

Unrelated to this PR:
cc @ntindle  do you think we should just fetch email and all the user context for the feature toggle whenever we need it?

autogpt_platform/backend/backend/server/v2/AutoMod/manager.py
```diff
@@ -0,0 +1,305 @@
...
+            logger.warning(f"Output moderation execution failed: {e}")
+            return ModerationError(
+                message="Execution failed due to output content moderation error",
+                user_id=user_id,
+                graph_exec_id=graph_exec_id,
+                moderation_type="output",
+            )
+
+    async def _update_failed_nodes_for_moderation(
+        self, db_client, graph_exec_id: str, moderation_type: str
```
majdyz: nit moderation_type can be Literal["input", "output"]

majdyz (changes requested):
Looks good, there non-intentional merge conflict issue that you need to resolve though. Let me know if it's been fixed so I can re-approve

autogpt_platform/backend/backend/server/v2/AutoMod/manager.py
```diff
@@ -0,0 +1,305 @@
...
+            return ModerationError(
+                message="Execution failed due to input content moderation error",
+                user_id=graph_exec.user_id,
+                graph_exec_id=graph_exec.graph_exec_id,
+                moderation_type="input",
+            )
+
+    async def moderate_graph_execution_outputs(
+        self,
+        db_client,
```
majdyz: nit: add type here so type hinting can work

majdyz: from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from backend.executor import DatabaseManagerAsyncClient

db_client: "DatabaseManagerAsyncClient"

autogpt_platform/backend/backend/server/v2/AutoMod/manager.py
ntindle: IMO we should build the full context stuff into a function with caching on the userID lookup then pass to LD to eval without caching LD

majdyz: [[feat(feature-flag): add LaunchDarkly user context and metadata support by majdyz · Pull Request #10595 · Significant-Gravitas/AutoGPT](https://github.com/Significant-Gravitas/AutoGPT/pull/10595)](https://github.com/Significant-Gravitas/AutoGPT/pull/10595)

autogpt_platform/backend/backend/executor/manager.py
```diff
@@ -759,6 +760,16 @@ def _on_graph_execution(
...
                 )
 
+            # Input moderation
+            if moderation_error := asyncio.run_coroutine_threadsafe(
+                automod_manager.moderate_graph_execution_inputs(
+                    db_client=get_db_async_client(),
+                    graph_exec=graph_exec,
+                ),
+                self.node_evaluation_loop,
+            ).result():
```
majdyz: Let's add some timeout just to be safe :)

autogpt_platform/backend/backend/executor/manager.py
```diff
@@ -897,6 +908,19 @@ def _on_graph_execution(
...
+            # Output moderation
+            if moderation_error := asyncio.run_coroutine_threadsafe(
+                automod_manager.moderate_graph_execution_outputs(
+                    db_client=get_db_async_client(),
+                    graph_exec_id=graph_exec.graph_exec_id,
+                    user_id=graph_exec.user_id,
+                    graph_id=graph_exec.graph_id,
+                ),
+                self.node_evaluation_loop,
+            ).result():
```
majdyz: Here too

majdyz (approved):
One more minor request, thanks for working on this!

autogpt_platform/backend/backend/server/v2/AutoMod/manager.py
```diff
@@ -0,0 +1,311 @@
...
+            if exec_entry.input_data:
+                for name in exec_entry.input_data.keys():
+                    cleared_inputs[name] = ["Failed due to content moderation"]
+
+            if exec_entry.output_data:
+                for name in exec_entry.output_data.keys():
+                    cleared_outputs[name] = ["Failed due to content moderation"]
+
+            # Update the node execution status
+            updated_exec = await db_client.update_node_execution_status(
```
majdyz: We can append all the tasks into a list, and at the end, we do `updated_execs = asyncio.gather(*exec_updates)`
So we don't need to wait for each of them:

Then do the same thing for the output

`asynctio.gather(*[send_async_execution_update(v) for v in updated_execs])`

autogpt_platform/backend/backend/server/v2/AutoMod/manager.py
```diff
@@ -0,0 +1,311 @@
...
+                    "error": "Failed due to content moderation",
+                    "cleared_inputs": cleared_inputs,
+                    "cleared_outputs": cleared_outputs,
+                },
+            )
+
+            # Send websocket update to notify frontend
+            await send_async_execution_update(updated_exec)
+
+    async def _moderate_content(self, content: str, metadata: Dict[str, Any]) -> bool:
```
majdyz: nit in case you are changing this file anyway: dict instead of Dict

PR #10587, opened 2025-08-08
fix(backend): Support dynamic values_#_* fields in CreateDictionaryBlock

ntindle:
## Summary

Fixed Smart Decision Maker's function signature generation to properly handle dynamic fields (e.g., `values_#_*`, `items_$_*`) when connecting to any block as a tool.

### Context

When Smart Decision Maker calls other blocks as tools, it needs to generate OpenAI-compatible function signatures. Previously, when connected to blocks via dynamic fields (which get merged by the executor at runtime), the signature generation would fail because blocks don't inherently know about these dynamic field patterns.

### Changes 🏗️

- **Modified `SmartDecisionMakerBlock._create_block_function_signature()`** to detect and handle dynamic fields:
  - Detects fields containing `_#_` (dict merge), `_$_` (list merge), or `_@_` (object merge)
  - Provides generic string schema for dynamic fields (OpenAI API compatible)
  - Falls back gracefully for unknown fields
- **Added comprehensive tests** for dynamic field handling with both dictionary and list patterns
- **No changes needed to individual blocks** - this solution works universally

### Why This Approach

Instead of modifying every block to handle dynamic fields (original PR approach), we handle it centrally in Smart Decision Maker where the function signatures are generated. This is cleaner and more maintainable.

### Test Plan 📋

- [x] Created test cases for Smart Decision Maker generating function signatures with dynamic dict fields (`_#_`)
- [x] Created test cases for Smart Decision Maker generating function signatures with dynamic list fields (`_$_`)
- [x] Verified Smart Decision Maker can successfully call blocks like CreateDictionaryBlock via dynamic connections
- [x] All existing Smart Decision Maker tests pass
- [x] Linting and formatting pass

### DIFF
--- a/autogpt_platform/backend/backend/blocks/data_manipulation.py
+++ b/autogpt_platform/backend/backend/blocks/data_manipulation.py
@@ -18,19 +18,6 @@ class CreateDictionaryBlock(Block):
             placeholder="e.g., {'name': 'Alice', 'age': 25}",
         )
 
-        @classmethod
-        def get_field_schema(cls, field_name: str) -> dict[str, Any]:
-            # Handle dynamic values_#_* fields (original format)
-            if field_name.startswith("values_#_"):
-                # Return a simple string type for compatibility with OpenAI
-                # The actual value can be any JSON-serializable type
-                return {
-                    "type": "string",
-                    "description": f"Dynamic value for key '{field_name[9:]}'",
-                }
-            # For regular fields, use the parent implementation
-            return super().get_field_schema(field_name)
-
     class Output(BlockSchema):
         dictionary: dict[str, Any] = SchemaField(
             description="The created dictionary containing the specified key-value pairs"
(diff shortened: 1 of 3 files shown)


autogpt_platform/backend/backend/blocks/data_manipulation.py
```diff
@@ -18,6 +18,19 @@ class Input(BlockSchema):
             placeholder="e.g., {'name': 'Alice', 'age': 25}",
         )
 
+        @classmethod
+        def get_field_schema(cls, field_name: str) -> dict[str, Any]:
```
majdyz: Where is this used ?

majdyz (changes requested):
image

The dynamic field seems to be working fine on CreateDictionaryBlock, how to reproduce this issue?

ntindle:
Repro with Smart Dceiosn maker going into a pre structured dictionary with keys set

autogpt_platform/backend/backend/blocks/data_manipulation.py
ntindle: its called by the smart decsion maker to build the schema

PR #10596, opened 2025-08-11
test(frontend): add e2e test for profile form page

Abhi1992002:
This PR has added end-to-end tests for the profile form page. These tests include:

- Redirects to the login page when the user is not authenticated.
- Can save profile changes successfully.
- Can cancel profile changes (skipped because we need to fix the form for this test).

### Changes 🏗️
- Added test-id's inside the ProfileInfoForm.
- Created a page object for the profile form page.
- Added a test for this page in `profile-form.spec.ts`.

#### For code changes:
- [x] I have made a test plan
  - [x] All test are working perfectly locally

### DIFF
--- a/autogpt_platform/frontend/src/components/agptui/ProfileInfoForm.tsx
+++ b/autogpt_platform/frontend/src/components/agptui/ProfileInfoForm.tsx
@@ -56,7 +56,10 @@ export function ProfileInfoForm({ profile }: { profile: ProfileDetails }) {
 
   return (
     <div className="w-full min-w-[800px] px-4 sm:px-8">
-      <h1 className="font-circular mb-6 text-[28px] font-normal text-neutral-900 dark:text-neutral-100 sm:mb-8 sm:text-[35px]">
+      <h1
+        data-testid="profile-info-form-title"
+        className="font-circular mb-6 text-[28px] font-normal text-neutral-900 dark:text-neutral-100 sm:mb-8 sm:text-[35px]"
+      >
         Profile
       </h1>
 
@@ -67,6 +70,7 @@ export function ProfileInfoForm({ profile }: { profile: ProfileDetails }) {
               <Image
                 src={profileData.avatar_url}
                 alt="Profile"
+                data-testid="profile-info-form-profile-photo"
                 fill
                 className="rounded-full"
               />
@@ -79,6 +83,7 @@ export function ProfileInfoForm({ profile }: { profile: ProfileDetails }) {
               type="file"
               accept="image/*"
               className="hidden"
+              data-testid="profile-info-form-edit-photo"
               onChange={async (e) => {
                 const file = e.target.files?.[0];
                 if (file) {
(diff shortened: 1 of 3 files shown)


autogpt_platform/frontend/src/components/agptui/ProfileInfoForm.tsx
```diff
@@ -99,6 +104,7 @@ export function ProfileInfoForm({ profile }: { profile: ProfileDetails }) {
               <input
                 type="text"
                 name="displayName"
+                data-testid="profile-info-form-display-name"
```
0ubbe, line 109: Generally, it is better to query this by field name:
```ts
page.getByLabel("Display name") // The text within the associated `<label>`
```
This ensures the form is accessible with corresponding labels associated with the fields. Same for the rest of the fields queried 👍🏽

autogpt_platform/frontend/src/components/agptui/ProfileInfoForm.tsx
```diff
@@ -200,6 +209,7 @@ export function ProfileInfoForm({ profile }: { profile: ProfileDetails }) {
 
           <div className="flex h-[50px] items-center justify-end gap-3 py-8">
             <Button
+              data-testid="profile-info-form-cancel"
```
0ubbe: Better to grab by role:
```ts
page.getByRole("button", {name: "Cancel"})
```
or with shorthand the helpers/selectors we have:
```ts
const { getButton } = getSelectors(page)

getButton("Cancel") // just a shorthand for the code above
```
Same for the `Save` button below. This ensures the buttons on the UI are either using the `<button>` HTML tag or have the `role="button"` attribute. So we are not only querying elements but ensuring accessibility at the same time 🙇🏽 

A bit more context [here](https://playwright.dev/docs/locators#locating-elements).

autogpt_platform/frontend/src/tests/profile-form.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+    ];
+
+    await profileFormPage.setDisplayName(newDisplayName);
+    await profileFormPage.setHandle(newHandle);
+    await profileFormPage.setBio(newBio);
+    await profileFormPage.setLinks(newLinks);
+
+    try {
+      await page.keyboard.press("Escape");
+    } catch {}
```
0ubbe: What is the purpose of this one? To remove focus from the form fields? If so, it is not necessary to have it, a user should be able to submit the form a field is focused or not 🤞🏽

autogpt_platform/frontend/src/tests/profile-form.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+    await expect(profileFormPage.isLoaded()).resolves.toBeTruthy();
+
+    expect(await profileFormPage.getDisplayName()).toBe(newDisplayName);
+    expect(await profileFormPage.getHandle()).toBe(newHandle);
+    expect(await profileFormPage.getBio()).toBe(newBio);
+    for (let i = 1; i <= 5; i++) {
+      expect(await profileFormPage.getLink(i)).toBe(newLinks[i - 1]);
+    }
+  });
+
```
0ubbe, line 75: If `Cancel` button is not working, we might just hide it for now until we refactor the form to use `react-hook-form` ?

Issue #8964, opened 2024-12-13
Marketplace - Fix margin underneath "Top agents" increase it to 37px


ograce1421:
### Describe your issue.

Fix this margin, increase it to 37px

![image](https://github.com/user-attachments/assets/21aec9ee-70b4-4f13-9c3e-10a022289c2f)

ograce1421:
Reopening this issue because it's still not solved.. 
The margins are still wrong

<img width="681" alt="Image" src="https://github.com/user-attachments/assets/3ed9347f-67ae-49e5-a347-3edc85e85b91" />

Pwuts:
As far as I can see it's fixed now
![](https://github.com/user-attachments/assets/9863fe38-ac91-47f8-a5fd-24bf54d0059e)

PR #10630, opened 2025-08-12
margin edit

AvinashYerra:
### Changes 🏗️
Just added margin as required

I haven't tested manually because couldn't run the project in my local. But this would suffice (I think so)
This is for the bug #8964


ntindle:
@AvinashYerra thanks for the contribution! In order to review it, we'll need you to agree to the CLA. Poke @0ubbe when you do and we can get this reviewed!

Pwuts:
The underlying issue has been solved, so this PR is no longer needed. Thank you for submitting @AvinashYerra!

PR #10650, opened 2025-08-15
fix(frontend): remove console logs + prevent via lint rule

0ubbe:
## Changes 🏗️

- Remove all the `console.log` / `console.debug` and `console.error` we have on the Front-end source code
- Instead of spamming `console.error`, log the error to Sentry instead where it makes sense
- Remove noisy logs on E2E tests
- Setup **ESLint** to throw an error if there is a `console` statement on the code
  - to prevent any future logs on the app 

#### For code changes:
- [x] I have made a test plan
  - [x] The lint job passes without error with the `no-console: "error"` rule enabled

None


autogpt_platform/frontend/src/tests/marketplace.spec.ts
```diff
@@ -32,10 +28,6 @@ test.describe("Marketplace – Basic Functionality", () => {
 
     const marketplaceTitle = await marketplacePage.getMarketplaceTitle(page);
     await isVisible(marketplaceTitle);
-
-    console.log(
```
ntindle, line 36: Can we allow in tests :)

PR #10665, opened 2025-08-18
fix(backend/executor): reduce scheduler max_workers to match database pool size

majdyz:
## Summary
- **Experimental fix** for scheduler pod crashes during peak scheduling periods (e.g., 03:00:00)
- Reduces APScheduler ThreadPoolExecutor max_workers from 10 to 3 (matching scheduler_db_pool_size)
- Attempts to prevent resource contention that may be causing health check timeouts

## Problem Statement
During peak scheduling periods, multiple jobs execute simultaneously and the scheduler pod crashes with:
- Exit code 137 (SIGKILL from liveness probe failure)
- Health check timeouts during high concurrency
- Executions created in DB but never published to RabbitMQ queue

## Hypothesis (Not 100% Certain)
Previous investigation showed threads were blocked on logging steps during high concurrency. The theory is that having 10 workers competing for only 3 database connections creates resource contention affecting:

1. **Database connection pool exhaustion** - workers waiting for available connections
2. **Thread contention on shared resources** - logging, synchronous operations
3. **System resource pressure** - too many concurrent operations

**Important**: We are NOT 100% sure this will fix the issue. This is an experimental approach based on matching worker count to available database connections.

## What We Know vs Don't Know
✅ **Confirmed facts:**
- Scheduler and FastAPI use separate event loops (not shared)
- Pod crashes occur during peak scheduling (7+ jobs at 03:00:00)
- Health check is simple (just checks scheduler.running, returns "OK")
- Previous investigation found thread blocking on logging

❓ **Uncertain:**
- Exact root cause of health check timeouts
- Whether DB connection limits are the bottleneck
- If this worker reduction will actually resolve the crashes

## Test Plan
- [ ] Monitor scheduler pod stability during peak scheduling periods
- [ ] Verify no executions remain QUEUED without being published to RabbitMQ
- [ ] Confirm health checks remain responsive under load
- [ ] Check that job execution still works correctly with reduced concurrency
- [ ] **If this doesn't work, we'll need to investigate other causes**

## Rollback Plan
If this experimental fix doesn't resolve the issue or causes problems, we can easily revert by changing `max_workers` back to 10.

### DIFF
--- a/autogpt_platform/backend/backend/executor/scheduler.py
+++ b/autogpt_platform/backend/backend/executor/scheduler.py
@@ -269,7 +269,9 @@ class Scheduler(AppService):
 
         self.scheduler = BackgroundScheduler(
             executors={
-                "default": ThreadPoolExecutor(max_workers=self.db_pool_size()),  # Match DB pool size to prevent resource contention
+                "default": ThreadPoolExecutor(
+                    max_workers=self.db_pool_size()
+                ),  # Match DB pool size to prevent resource contention
             },
             job_defaults={
                 "coalesce": True,  # Skip redundant missed jobs - just run the latest


majdyz:
Closing PR - analysis was incorrect. Scheduler and FastAPI use separate event loops, so max_workers reduction doesn't fix the health check blocking issue. Need to investigate the actual root cause of pod crashes during high concurrency.

majdyz:
Reopening PR with corrected analysis. While scheduler and FastAPI use separate event loops, reducing max_workers still addresses the root cause: thread contention on shared resources (DB connections, logging) during high concurrency. Previous investigation showed threads blocked on logging steps.

AutoGPT-Agent:
Thanks for this well-documented PR addressing the scheduler pod crashes! The change to match thread pool workers with the DB connection pool size is a sensible approach to test.

A couple of things to address before we can merge:

1. Please update the PR description to include the standard checklist format. Even for small changes like this, we need the basic checklist items checked off (you can omit configuration-related items since they don't apply).

2. The PR title should follow our conventional commit format more precisely. Instead of "hotfix:", please use "fix:" and include the appropriate scope: "fix(backend/executor): reduce scheduler max_workers to match database pool size"

The technical change itself looks good, and your detailed explanation of the problem and approach is excellent. Once you've addressed these process items, we can proceed with the merge.

PR #10670, opened 2025-08-18
fix(frontend/builder): Prevent unnecessary saves before run

### LINKED ISSUE #10444: Fix: "Run" in Builder Triggers Save Even When No Changes Are Made
Description:<br>Clicking "Run" in the Builder currently triggers a full save of the agent, even when there have been no edits or changes made.

Reported by: Pwuts<br>Context from Zamil: Our algorithm to detect changes may now be stale.

Pwuts:
- Resolves #10444

Sometimes, the order of nodes and/or links isn't consistent between frontend and backend, which currently can result in unnecessary re-saving of the graph when the user tries to run it.
Also, `sub_graphs` was not included in the frontend `Graph` type, which can cause unchecked code issues when the object is propragated using spread operators.

### Changes 🏗️

- fix(frontend/builder): Make `graphsEquivalent` insensitive to link and node order
- dx(frontend): Fix typing of `Graph.sub_graphs` (and its variants)

#### For code changes:
- [x] I have made a test plan
  - Import an agent and open it in the builder
  - Run it without making any changes to the graph itself
    - [x] -> graph shouldn't re-save

### DIFF
--- a/autogpt_platform/frontend/src/hooks/useAgentGraph.tsx
+++ b/autogpt_platform/frontend/src/hooks/useAgentGraph.tsx
@@ -936,6 +936,8 @@ function graphsEquivalent(saved: Graph, current: GraphCreatable): boolean {
   const sortLinks = (links: LinkCreatable[]) =>
     links.toSorted(
       (a, b) =>
+        8 * a.source_id.localeCompare(b.source_id) +
+        4 * a.sink_id.localeCompare(b.sink_id) +
         2 * a.source_name.localeCompare(b.source_name) +
         a.sink_name.localeCompare(b.sink_name),
     );


AutoGPT-Agent:
Thank you for this PR to fix the unnecessary saves issue! The code changes look good - adding sorting to make the graph comparison order-insensitive and fixing the type for `sub_graphs` are sensible improvements.

I noticed that you haven't checked off the testing item in your checklist. Before we can merge this PR, could you please:

1. Complete the testing according to your test plan (importing an agent, opening it in the builder, and running it without making changes to verify it doesn't re-save)
2. Check off the testing item in the checklist once verified

The changes themselves look straightforward and appropriate for the issue you're solving. Let us know once you've completed the testing so we can proceed with the merge.

autogpt_platform/frontend/src/hooks/useAgentGraph.tsx
```diff
@@ -930,24 +930,36 @@ export default function useAgentGraph(
 }
 
 function graphsEquivalent(saved: Graph, current: GraphCreatable): boolean {
+  const sortNodes = (nodes: NodeCreatable[]) =>
+    nodes.sort((a, b) => a.id.localeCompare(b.id));
+
+  const sortLinks = (links: LinkCreatable[]) =>
+    links.sort(
+      (a, b) =>
+        2 * a.source_name.localeCompare(b.source_name) +
```
majdyz, line 941: Why do we do this? Is there a more normal way to compare?

AutoGPT-Agent:
Thanks for your PR addressing the unnecessary saves issue when running agents! 

The changes look good and well-targeted to the problem - making the graph comparison order-insensitive should prevent those unnecessary saves when node/link order differs between frontend and backend.

A couple of things to address before we can merge this:

1. Your checklist indicates you haven't completed testing according to your test plan yet. Could you please test your changes and check off that item?

2. I notice you've added proper typing for `sub_graphs` which is a good improvement. Have you verified that this doesn't cause any issues in other parts of the code that might be relying on the previous typing?

Once you've completed testing and can confirm these changes work as expected, this PR should be ready to merge. Thanks for improving the developer and user experience!

autogpt_platform/frontend/src/hooks/useAgentGraph.tsx
Pwuts: it may look a bit weird but it's just a prioritized sort: sort by `source_id` first, then `sink_id`, then `source_name`, then `sink_name`

Pwuts: Can't sort by link ID because it's only guaranteed to be available for one of the two graphs that we are comparing: one of them is `GraphCreatable`, and `LinkCreatable` doesn't have an ID.

0ubbe: Can we abstract this into a helper naming what it does?
```ts
function sortLinksByPiority(a, b) {
// ...
```
Or if not, add a comment explaining? The numbers looks a bit magic and might not be obvious for everyone 👀

majdyz: We can do a normal 
if x not zero return x
Instead of doing bitwise shift which is much easier to understand and extend

PR #10716, opened 2025-08-22
fix(frontend): Propagate API auth errors to original requestor

### LINKED ISSUE #10713: Fix auth failure handling in API proxy
The frontend's React Query API client generates types and logic based on the OpenAPI spec of the backend API. These types also state what kinds of responses are possible, e.g. when `query.status === 200` then `query.data` must be `MyAgentsResponse`.

The API proxy however returns `null` if it can't authenticate an outgoing request, causing practical cases that diverge from what's possible according to the generated types.

Two fundamental issues here:

* Authentication failure isn't treated as a real failure (error), and `null` is returned with status 200 which may not match the expected response type
* We have authenticated API calls in unauthenticated sections of the application. E.g. users can visit `/marketplace` without logging in, but that page contains the `PublishAgentDialog` which tries to fetch the user's library agents.

Pwuts:
- Resolves #10713

### Changes 🏗️

- Remove early exit in API proxy that suppresses auth errors
- Remove unused `proxy-action.ts`

#### For code changes:
- [x] I have made a test plan
  - [x] Publish Agent dialog works when logged out
  - [x] Publish Agent dialog works when logged in

### DIFF
--- a/autogpt_platform/backend/backend/blocks/ai_music_generator.py
+++ b/autogpt_platform/backend/backend/blocks/ai_music_generator.py
@@ -166,7 +166,7 @@ class AIMusicGeneratorBlock(Block):
                     output_format=input_data.output_format,
                     normalization_strategy=input_data.normalization_strategy,
                 )
-                if result and result != "No output received":
+                if result and isinstance(result, str) and result.startswith("http"):
                     yield "result", result
                     return
                 else:
--- a/autogpt_platform/backend/backend/server/v2/AutoMod/manager.py
+++ b/autogpt_platform/backend/backend/server/v2/AutoMod/manager.py
@@ -99,7 +99,7 @@ class AutoModManager:
                 )
                 # Update node statuses for frontend display before raising error
                 await self._update_failed_nodes_for_moderation(
-                    db_client, graph_exec.graph_exec_id, "input"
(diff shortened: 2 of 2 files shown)


autogpt_platform/frontend/src/app/api/proxy/[...path]/route.ts
```diff
@@ -98,8 +98,27 @@ function createResponse(
...
+    console.warn(
+      `Authentication error in API proxy for ${method} ${path}:`,
+      "message" in error ? error.message : error,
+    );
+  } else {
+    console.error(
+      `API proxy received error response from ${method} ${path}:`,
+      error,
+    );
+  }
```
Comment (0ubbe) [inline autogpt_platform/frontend/src/app/api/proxy/[...path]/route.ts]: Why is this helpful? You have the network tab on the browser devtools to see this 👀

0ubbe (approved):

autogpt_platform/frontend/src/app/api/proxy/[...path]/route.ts
Comment (Pwuts) [reply autogpt_platform/frontend/src/app/api/proxy/[...path]/route.ts]: The `console.error` statement was already there. But point taken, I'll remove this one.

Should we leave the other warning (`Authentication error in API proxy for ${method} ${path}:`) as a possible bug indicator?

(answer received over discord: yes)

0ubbe (approved):

PR #10779, opened 2025-08-29
feat(frontend): Add editable node titles with hover pencil icon

Swiftyos:
## Summary
- Adds ability to edit custom node titles by clicking a pencil icon that appears on hover
- Custom titles are saved in node metadata and persist across saves
- Original node type is shown in tooltip when hovering over custom titles

## Changes
- **CustomNode.tsx**: 
  - Added inline title editing with pencil icon on hover
  - Implemented state management for title editing mode
  - Added tooltip to show original node type for custom titles
  - Prevents custom names from being copied when duplicating nodes

- **useAgentGraph.tsx**:
  - Updated graph save/load logic to preserve metadata including custom titles
  - Ensures metadata persistence through all node operations

## Technical Details
- Uses existing `metadata` JSON field in AgentNode model (no database changes needed)
- Stores custom title in `metadata.customized_name`
- Backward compatible - nodes without custom titles display normally

## Test Plan
- [x] Hover over node title shows pencil icon
- [x] Click pencil icon to edit title
- [x] Press Enter or blur to save, Escape to cancel
- [x] Custom title persists after saving graph
- [x] Tooltip shows original node type when hovering over custom title
- [x] Copying node doesn't copy custom name
- [x] Backward compatible with existing graphs

### DIFF
--- a/autogpt_platform/frontend/src/components/CustomNode.tsx
+++ b/autogpt_platform/frontend/src/components/CustomNode.tsx
@@ -53,11 +53,19 @@ import {
   TrashIcon,
   CopyIcon,
   ExitIcon,
+  Pencil1Icon,
 } from "@radix-ui/react-icons";
 import { Key } from "@phosphor-icons/react";
 import useCredits from "@/hooks/useCredits";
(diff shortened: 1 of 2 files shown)


Bentlybro (approved):
Tested and works!

PR #10791, opened 2025-09-01
feat(frontend): Preserve `openapi.json` spec file on generation failure

kcze:
`openapi.json` file is cleared when script fails to retrieve api spec from the server. This shouldn't happen and it breaks building docker containers.

### Changes 🏗️

Use temp file during generation to prevent actual file clearing on failure.

#### For code changes:
- [x] I have made a test plan
  - [x] Spec file doesn't get cleared on failure
  - [x] Spec file is correctly generated
  - [x] Works when frontend is run in docker container


kcze:
> Do we know what is the causing the failures? Maybe we want to be explicit and break the build if fetching from the OpenAPI from server fails? 🤔

@0ubbe it fails when the backend isn't up!

PR #10792, opened 2025-09-01
fix(frontend): prevent using mock feature flags

0ubbe:
## Changes 🏗️

Make sure `NEXT_PUBLIC_PW_TEST` is set only when running Playwright. This forces the app to use "mock" feature flags, so the tests run stable and predictable despite changes on LaunchDarkly.

### For code changes:
- [x] I have made a test plan
  - [x]  should not have `PW_TEST=true` ...

None


AutoGPT-Agent:
Thank you for your contribution! I noticed a few issues that need to be addressed before this PR can be merged:

1. The PR description is incomplete. Please fill out the "Changes" section to explain what you're changing and why. For this specific change, please explain why we're changing the default value of NEXT_PUBLIC_PW_TEST from true to false.

2. None of the checklist items have been checked off. Please complete the checklist by checking the relevant boxes.

3. The PR title is incomplete, ending with ellipsis. Please update it to fully describe what you're doing, for example: "chore(frontend): change default Playwright test flag to false"

The actual code change (setting NEXT_PUBLIC_PW_TEST to false by default and adding a clarifying comment) looks reasonable, but we need the documentation around the PR to be complete before merging.

AutoGPT-Agent:
Thank you for your contribution to fix the mock feature flags issue! However, before this PR can be approved, a few important items need to be addressed:

1. Please complete the PR description by explaining what the change does and why it's needed. From the diff, it appears you're changing the default value of `NEXT_PUBLIC_PW_TEST` from `true` to `false` and adding a comment about its usage, but the context and reasoning should be explicitly stated.

2. None of the checklist items are checked. Please fill out the relevant parts of the checklist, particularly:
   - Confirm you've clearly listed your changes
   - Indicate you've tested your changes
   - Provide a simple test plan that you followed

Since this is a configuration change, you might also want to mention this under the "Changes" section, explaining what impact this will have on development environments.

Once these items are addressed, we can proceed with the review process. Thank you!

majdyz (dismissed):
This might break CI without config change on other places

AutoGPT-Agent:
Thanks for the PR! The changes look good overall, addressing the issue with mock feature flags in a clean way.

Just one thing to address before merging:

- There's a `console.log(">>>>", process.env.NEXT_PUBLIC_PW_TEST);` statement in the `use-get-flag.ts` file that should be removed as it appears to be debugging code.

The approach of removing the flag from .env.default and explicitly setting it only when running tests makes sense. This prevents the mock feature flags from being used in production or development environments unintentionally.

autogpt_platform/frontend/src/services/feature-flags/use-get-flag.ts
```diff
@@ -28,9 +29,11 @@ const mockFlags = {
 export function useGetFlag<T extends Flag>(flag: T): FlagValues[T] | null {
   const currentFlags = useFlags<FlagValues>();
   const flagValue = currentFlags[flag];
+  const isCloud = getBehaveAs() === BehaveAs.CLOUD;
 
-  if (isTest) return mockFlags[flag];
-  if (!flagValue) return null;
+  console.log(">>>>", process.env.NEXT_PUBLIC_PW_TEST);
```
0ubbe:
```suggestion
```

PR #10832, opened 2025-09-03
fix(frontend/library): Show total runs count above runs list

### LINKED ISSUE #10831: Fix total run count in library agent runs list
The count on the Runs tab currently shows only the number of *loaded* runs rather than the total number of runs.

Pwuts:
- Resolves #10831

### Changes 🏗️

- Show number of total runs instead of currently loaded runs
- Show loading spinner instead of zero while loading

#### For code changes:
- [x] I have made a test plan
  - [x] Counter shows number of total runs, even if it exceeds number of currently loaded items


0ubbe (approved):
Looks good but we need to bring it into https://github.com/Significant-Gravitas/AutoGPT/pull/10835

PR #10819, opened 2025-09-02
feat(frontend): Enhanced output rendering system for agent runs

ntindle:
## Summary
Introduces a modular, extensible output renderer system supporting multiple content types (text, code, images, videos, JSON, markdown) for agent run outputs. The system includes smart clipboard operations, concatenated downloads, and rich markdown rendering with LaTeX math and video embedding support.

### Core Output Rendering System
- **Added extensible renderer architecture** (`output-renderers/types.ts`)
  - Plugin-based system with priority ordering
  - Registry pattern for automatic renderer selection
  - Support for custom metadata and MIME types

### Output Renderers
- **TextRenderer**: Plain text with proper formatting and line breaks
- **CodeRenderer**: Syntax-highlighted code blocks with language detection
- **JSONRenderer**: Collapsible, formatted JSON with syntax highlighting
- **ImageRenderer**: Image display with support for URLs, data URIs, and file uploads
- **VideoRenderer**: Embedded video player for YouTube, Vimeo, and direct video files
- **MarkdownRenderer**: Rich markdown with:
  - GitHub Flavored Markdown (tables, task lists, strikethrough)
  - LaTeX math rendering via KaTeX (inline `$...$` and display `$$...$$`)
  - Syntax highlighting via highlight.js
  - Video embedding (YouTube/Vimeo URLs auto-convert to embeds)
  - Clickable heading anchors
  - Dark mode support

### User Interface Components
- **OutputItem**: Individual output display with renderer selection
- **OutputActions**: Hover-based action buttons for:
  - Copy to clipboard with smart MIME type detection
  - Download with intelligent concatenation (text files merge, binaries separate)
  - Share functionality (placeholder for future implementation)
- **AgentRunOutputView**: Main output view component with feature flag integration

### Clipboard & Download Features
- Smart clipboard operations using native ClipboardItem API
- MIME type detection and browser capability checking
- Fallback strategies for unsupported content types
- Concatenated downloads for text-based outputs
- Individual downloads for binary content

### Feature Flag Integration
- Added `ENABLE_ENHANCED_OUTPUT_HANDLING` flag to LaunchDarkly
- Backwards compatible with existing output display
- Graceful fallback for disabled feature flag

### Styling & UX
- Max width constraints (950px card, 660px content)
- Hover-based action buttons for clean interface
- Dark mode support across all renderers
- Responsive design for various content types
- Loading states and error handling

### For code changes:
- [x] I have clearly listed my changes in the PR description
- [x] I have made a test plan
[...]

PR #10835, opened 2025-09-04
feat(frontend): new agent library run page

0ubbe:
## Changes 🏗️

This is the new **Agent Library Run** page. Sorry in advance for the massive PR 🙏🏽 . I got carried away and it has been tricky to split it ( _maybe I abused the agent too much_ 🤔 )

Screenshot 2025-09-04 at 13 58 33

Screenshot 2025-09-04 at 13 54 51

Screenshot 2025-09-04 at 13 54 18

- Schedules are now on the sidebar tabbed along with runs
- The whole UI has been updated to match the new designs and design system
- There is no more "run draft" view as the modal is in charge of new runs now 💪🏽 
- The page is responsive and mobile friendly 📱 

Uploading mobile.mov…

https://github.com/user-attachments/assets/0e483062-0e50-4fa6-aaad-a1f6766df931

### Safety

I understand this is a lot of changes. However is all behind a feature flag, `new-agent-runs`, when OFF it will display the old library agent view. The old library agent view can still be accessed under: `/library/legacy/{id}` for reference 👍🏽 

### Testing

I haven't any tests for now... 💆🏽 I want to get this enabled on dev so we can start running our agents there through the new page and modal and start catching edge-cases. 

Tests will come later in the form of E2E for the happy paths, and probably I will introduce [Vitest](https://vitest.dev/) + [Testing Library](https://testing-library.com/) for the finer details...

### For code changes:
- [x] I have made a test plan
  - [x] Test the above

None, the feature flag is already configured 🙏🏽


Swiftyos (approved):
We need to fix the CI failing, but once thats sorted we can merge in so we can test it before release

autogpt_platform/frontend/src/app/(platform)/library/agents/[id]/components/AgentRunsView/components/RunDetails/RunDetails.tsx
```diff
@@ -0,0 +1,111 @@
...
+        <TabsLineContent value="output">
+          <RunDetailCard>
+            {isLoading ? (
+              <div className="text-neutral-500">Loading…</div>
+            ) : !run ||
+              !("outputs" in run) ||
+              Object.keys(run.outputs || {}).length === 0 ? (
+              <div className="text-neutral-600">No output from this run.</div>
+            ) : (
+              <div className="flex flex-col gap-4">
```
Comment (ntindle) [inline autogpt_platform/frontend/src/app/(platform)/library/agents/[id]/components/AgentRunsView/components/RunDetails/RunDetails.tsx:82]: This needs migrated to the outputs from old under the feature flag that was added in #10819

autogpt_platform/frontend/src/app/(platform)/library/agents/[id]/components/AgentRunsView/components/RunsSidebar/useRunsSidebar.ts
```diff
@@ -0,0 +1,117 @@
...
+  const [tabValue, setTabValue] = useState<"runs" | "scheduled">("runs");
+
+  const runsQuery = useGetV1ListGraphExecutionsInfinite(
+    graphId || "",
+    { page: 1, page_size: 20 },
+    {
+      query: {
+        enabled: !!graphId,
+        // Lightweight polling so statuses refresh; only poll if any run is active
+        refetchInterval: (q) => {
```
Comment (ntindle) [inline autogpt_platform/frontend/src/app/(platform)/library/agents/[id]/components/AgentRunsView/components/RunsSidebar/useRunsSidebar.ts:28]: Shouldn't this pass over the websocket?

autogpt_platform/frontend/src/app/(platform)/library/agents/[id]/components/AgentRunsView/components/ScheduleDetails/ScheduleDetails.tsx
```diff
@@ -0,0 +1,239 @@
...
+        </TabsLineContent>
+
+        <TabsLineContent value="schedule">
+          <RunDetailCard>
+            {isLoading || !schedule ? (
+              <div className="text-neutral-500">Loading…</div>
+            ) : (
+              <div className="relative flex flex-col gap-8">
+                {
+                  // TODO: re-enable edit schedule modal once the API supports it
```
Comment (ntindle) [inline autogpt_platform/frontend/src/app/(platform)/library/agents/[id]/components/AgentRunsView/components/ScheduleDetails/ScheduleDetails.tsx:146]: Is there ticket for this?

autogpt_platform/frontend/src/app/(platform)/library/agents/[id]/components/AgentRunsView/components/RunDetails/RunDetails.tsx
Comment (0ubbe) [reply autogpt_platform/frontend/src/app/(platform)/library/agents/[id]/components/AgentRunsView/components/RunDetails/RunDetails.tsx:82]: Yes, however I wanna do on a follow up PR given this one is big enough 🙇🏽 ( _same for adding the timezone on the schedule setup view_ ).

Tickets for this on the project:
- [Use new Output view](https://linear.app/autogpt/issue/OPEN-2686/use-new-output-rendering-on-new-agent-page)
- [Add Timezone to Schedule modal](https://linear.app/autogpt/issue/OPEN-2687/add-timezone-to-new-schedule-modal)

autogpt_platform/frontend/src/app/(platform)/library/agents/[id]/components/AgentRunsView/components/RunsSidebar/useRunsSidebar.ts
Comment (0ubbe) [reply autogpt_platform/frontend/src/app/(platform)/library/agents/[id]/components/AgentRunsView/components/RunsSidebar/useRunsSidebar.ts:28]: I prefer to try the polling approach first ( _is the same strategy the old library view uses..._ ).

A couple of reasons why I think polling is better than websockets in this context:
- Polling is more reliable ( websockets tend to loose connection )
- The code is way simpler. Adding websockets to already a complex page just adds more complexity as we need to handle the connections etc...
- Polling is good enough. It will only poll if it detects any of the returned executions are queuing or running ( _this is an improvement over the old view which it would start polling as soon as you land on the page_ ), no matter whether an agent is running or not
  - I see websockets a better fit for something like the `<AgentActivityDropdown />`, which is always on the navbar and needs to always listen ( _given an agent might start running from the builder, or the library..._ )

I would like to try this approach first 🙏🏽 , and introduce a Websocket layer only if we find strong reasons for it

Comment (0ubbe) [reply autogpt_platform/frontend/src/app/(platform)/library/agents/[id]/components/AgentRunsView/components/RunsSidebar/useRunsSidebar.ts:28]: That said probably on the near future we could have a re-usable hook that connects to websockets and returns execution data. Basically the logic on `<AgentActivityDropdown />` but abstracted to be render agnostic ( _can be 
used with any UI..._ ).

autogpt_platform/frontend/src/app/(platform)/library/agents/[id]/components/AgentRunsView/components/ScheduleDetails/ScheduleDetails.tsx
Comment (0ubbe) [reply autogpt_platform/frontend/src/app/(platform)/library/agents/[id]/components/AgentRunsView/components/ScheduleDetails/ScheduleDetails.tsx:146]: https://linear.app/autogpt/issue/OPEN-2687/add-timezone-to-new-schedule-modal

PR #10859, opened 2025-09-05
fix(frontend): Fix environment variable handling in Docker builds for dev/prod deployments

ntindle:
## Summary
- Fixed Dockerfile environment variable merging that was causing Sentry to be disabled in dev/prod deployments
- Added authentication check to onboarding timezone detection to prevent unauthorized API calls
- Added Dockerfile to .prettierignore

## Problem
The dev and prod deployments had `isProdOrDev` evaluating to `false`, causing Sentry to be disabled even though `NEXT_PUBLIC_APP_ENV=dev` was set in the pod environment.

### Root Cause
1. CI/CD workflow creates `.env.production` with correct values
2. Dockerfile's env merging logic always created `.env` from `.env.default` 
3. Next.js loads `.env.production` first, then `.env` (which overwrites values)
4. `.env.default` has `NEXT_PUBLIC_APP_ENV=local`, overwriting the production value
5. This caused `getAppEnv()` to return "local" instead of "dev"/"prod"

## Solution
Updated Dockerfile to:
- Only merge `.env.default` with `.env` when NOT in CI/CD (no `.env.production` exists)
- When `.env.production` exists, merge it with `.env.default` to ensure all defaults are available
- Production values take precedence over defaults

## Additional Fixes
- Added authentication check to `useOnboardingTimezoneDetection` hook to prevent calling `/api/auth/user/timezone` when not logged in
- Added Dockerfile to .prettierignore to prevent formatter errors

## Test Plan
- [ ] Verify local Docker builds still work with `docker compose up`
- [ ] Verify dev deployment has `NEXT_PUBLIC_APP_ENV=dev` and Sentry enabled
- [ ] Verify prod deployment has `NEXT_PUBLIC_APP_ENV=prod` and Sentry enabled
- [ ] Verify onboarding timezone detection only runs when authenticated

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/autogpt_platform/frontend/.prettierignore
+++ b/autogpt_platform/frontend/.prettierignore
@@ -4,3 +4,5 @@ pnpm-lock.yaml
 .auth
 build
 public
+Dockerfile
+.prettierignore
--- a/autogpt_platform/frontend/Dockerfile
+++ b/autogpt_platform/frontend/Dockerfile
@@ -12,9 +12,16 @@ COPY autogpt_platform/frontend/ .
 # Allow CI to opt-in to Playwright test build-time flags
 ARG NEXT_PUBLIC_PW_TEST="false"
 ENV NEXT_PUBLIC_PW_TEST=$NEXT_PUBLIC_PW_TEST
-RUN if [ -f .env ]; then \
+ENV NODE_ENV="production"
+# Merge env files appropriately based on environment
+RUN if [ -f .env.production ]; then \
+      # In CI/CD: merge defaults with production (production takes precedence)
+      cat .env.default .env.production > .env.merged && mv .env.merged .env.production; \
+    elif [ -f .env ]; then \
+      # Local with custom .env: merge defaults with .env
       cat .env.default .env > .env.merged && mv .env.merged .env; \
     else \
+      # Local without custom .env: use defaults
       cp .env.default .env; \
     fi
 RUN pnpm run generate:api
--- a/autogpt_platform/frontend/instrumentation-client.ts
+++ b/autogpt_platform/frontend/instrumentation-client.ts
@@ -18,7 +18,10 @@ const isDisabled = process.env.DISABLE_SENTRY === "true";
 
 const shouldEnable = !isDisabled && isProdOrDev && isCloud;
 
-console.log(`shouldEnableSentry: ${shouldEnable}`);
+console.log(`shouldEnableSentry: ${shouldEnable} (instrumentation)`);
+console.log(`isCloud: ${isCloud} (instrumentation)`);
+console.log(`isDisabled: ${isDisabled} (instrumentation)`);
+console.log(`isProdOrDev: ${isProdOrDev} (instrumentation)`);
 
 Sentry.init({
   dsn: "https://fe4e4aa4a283391808a5da396da20159@o4505260022104064.ingest.us.sentry.io/4507946746380288",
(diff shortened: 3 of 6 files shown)


autogpt_platform/frontend/.prettierignore
```diff
@@ -4,3 +4,5 @@ pnpm-lock.yaml
 .auth
 build
 public
+Dockerfile
+.prettierignore
```
0ubbe, line 8: No need for prettierignore to ignore prettierignore?? 😄

0ubbe (dismissed):

autogpt_platform/frontend/instrumentation-client.ts
```diff
@@ -2,12 +2,16 @@
...
+  getAppEnv,
+  getBehaveAs,
+  getEnvironmentStr,
+} from "@/lib/utils";
 import * as Sentry from "@sentry/nextjs";
 
-const isProdOrDev =
-  process.env.NODE_ENV === "production" ||
-  process.env.NODE_ENV === "development";
+const isProdOrDev = getAppEnv() === AppEnv.PROD || getAppEnv() === AppEnv.DEV;
```
Pwuts: nit:

```suggestion
const isProdOrDev = [AppEnv.PROD, AppEnv.DEV].includes(getAppEnv());
```

Pwuts: same for the other 2 instances

autogpt_platform/frontend/Dockerfile
```diff
@@ -12,9 +12,16 @@ COPY autogpt_platform/frontend/ .
 # Allow CI to opt-in to Playwright test build-time flags
 ARG NEXT_PUBLIC_PW_TEST="false"
 ENV NEXT_PUBLIC_PW_TEST=$NEXT_PUBLIC_PW_TEST
-RUN if [ -f .env ]; then \
+ENV NODE_ENV="production"
+# Merge env files appropriately based on environment
+RUN if [ -f .env.production ]; then \
```
Pwuts, line 17: should this also be conditional on `NODE_ENV` being `"production"`?

autogpt_platform/frontend/sentry.edge.config.ts
```diff
@@ -4,11 +4,15 @@
...
-  process.env.NODE_ENV === "development";
+import {
+  AppEnv,
+  BehaveAs,
+  getAppEnv,
+  getBehaveAs,
+  getEnvironmentStr,
+} from "./src/lib/utils";
+
+const isProdOrDev = getAppEnv() === AppEnv.PROD || getAppEnv() === AppEnv.DEV;
```
Pwuts:
```suggestion
const isProdOrDev = [AppEnv.PROD, AppEnv.DEV].includes(getAppEnv());
```

autogpt_platform/frontend/sentry.server.config.ts
```diff
@@ -2,13 +2,17 @@
...
+  getBehaveAs,
+  getEnvironmentStr,
+} from "@/lib/utils";
 import * as Sentry from "@sentry/nextjs";
 // import { NodeProfilingIntegration } from "@sentry/profiling-node";
 
-const isProdOrDev =
-  process.env.NODE_ENV === "production" ||
-  process.env.NODE_ENV === "development";
+const isProdOrDev = getAppEnv() === AppEnv.PROD || getAppEnv() === AppEnv.DEV;
```
Pwuts:
```suggestion
const isProdOrDev = [AppEnv.PROD, AppEnv.DEV].includes(getAppEnv());
```

Pwuts (approved):

autogpt_platform/frontend/.prettierignore
Pwuts: ¯\\_(ツ)_/¯

PR #10861, opened 2025-09-06
ci: Add NODE_OPTIONS for memory limit during build

ntindle:
#### For code changes:
- [ ] I have made a test plan
  
  - [ ] Test if pass ci

### DIFF
--- a/autogpt_platform/frontend/Dockerfile
+++ b/autogpt_platform/frontend/Dockerfile
@@ -19,7 +19,7 @@ RUN if [ -f .env ]; then \
     fi
 RUN pnpm run generate:api
 # In CI, we want NEXT_PUBLIC_PW_TEST=true during build so Next.js inlines it
-RUN if [ "$NEXT_PUBLIC_PW_TEST" = "true" ]; then NEXT_PUBLIC_PW_TEST=true pnpm build; else pnpm build; fi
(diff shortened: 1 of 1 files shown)


autogpt_platform/frontend/Dockerfile
```diff
@@ -19,7 +19,7 @@ RUN if [ -f .env ]; then \
     fi
 RUN pnpm run generate:api
 # In CI, we want NEXT_PUBLIC_PW_TEST=true during build so Next.js inlines it
-RUN if [ "$NEXT_PUBLIC_PW_TEST" = "true" ]; then NEXT_PUBLIC_PW_TEST=true pnpm build; else pnpm build; fi
+RUN if [ "$NEXT_PUBLIC_PW_TEST" = "true" ]; then NEXT_PUBLIC_PW_TEST=true NODE_OPTIONS="--max-old-space-size=4096" pnpm build; else NODE_OPTIONS="--max-old-space-size=4096" pnpm build; fi
```
Pwuts, line 29: I don't get the point of this condition, because if `if [ "$NEXT_PUBLIC_PW_TEST" = "true" ]` then `pnpm build` will see it in the environment, right? No need to explicitly prepend it to the command.
```suggestion
RUN NODE_OPTIONS="--max-old-space-size=4096" pnpm build
```

Upping the memory limit to 4 GiB seems fine as standard GitHub runners have 16 GiB of RAM.

ntindle: this is out of scope i'm sorry but i have no idea what the extra context is

0ubbe (approved):

PR #10876, opened 2025-09-09
feat(frontend): use new output renderers on new run page

0ubbe:
## Changes 🏗️

Integrating the great work @ntindle did on the rich agent output renderers into the new Agent run page in the library 💜 

- Implemented enhanced output rendering in `<RunDetails />` using the shared output-renderers
- Added `<RunOutputs />` sub-component at `RunDetails/components/RunOutputs.tsx` that:
  - [x] builds items from `run.outputs`, extracts metadata, picks a renderer via `globalRegistry`, and falls back to `TextRenderer`
  - [x] renders `<OutputActions />` for copy/download and a list of `<OutputItems />`.

### For code changes:
- [x] I have made a test plan
  - [x] Run an agent on the view which outputs rich content
  - [x] See the output use the new renderers, for example code is higlighted 

None


ntindle:
At some point this will be the base for a simple share page. wehre should the components to render live at that point? the nearest shared would be the containing folder for (platform) i think

0ubbe:
> At some point this will be the base for a simple share page. wehre should the components to render live at that point? the nearest shared would be the containing folder for (platform) i think

Depends, need to see the designs for it etc... If these render outputs are used in more than 1 page, we can moved them to the `components/contextual` folder, is where I'm putting components that are re-used across pages but are rich enough ( API calls etc... ) to not be considered pure blocks of a design system 💅🏽

PR #10955, opened 2025-09-18
fix(blocks): Default disable HTML escaping in all blocks with templating features

### LINKED ISSUE #10954: Prompt values in AI blocks are escaped with HTML character entities
Goes in like this:

```
 export type LibraryAgentPresetID = Brand<string, "LibraryAgentPresetID">;
 
@@ -594,9 +612,9 @@ export type Webhook = {
   id: string;
   url: string;
   provider: CredentialsProviderName;
-  credentials_id: string;
+  credentials_id: string; // empty string if not appicable
   webhook_type: string;
-  resource?: string;
+  resource: string; // empty string if not appicable
   events: string[];
   secret: string;
   config: Record<string, any>;
```

But ends up in the prompt like this:

```
 export type LibraryAgentPresetID = Brand&lt;string, &#34;LibraryAgentPresetID&#34;&gt;;
 
@@ -594,9 +612,9 @@ export type Webhook = {
   id: string;
   url: string;
   provider: CredentialsProviderName;
-  credentials_id: string;
+  credentials_id: string; // empty string if not appicable
   webhook_type: string;
-  resource?: string;
+  resource: string; // empty string if not appicable
   events: string[];
   secret: string;
   config: Record&lt;string, any&gt;;
```

Which is obviously a distortion of the content and can hinder correct inference by the LLM.

## Cause

In `blocks/llm.py`:

```
# module-level definition
fmt = TextFormatter()

# and this happens in AIStructuredResponseGeneratorBlock.run(..):
values = input_data.prompt_values
if values:
    input_data.prompt = fmt.format_string(input_data.prompt, values)
    input_data.sys_prompt = fmt.format_string(input_data.sys_prompt, values)
```

And `TextFormatter`:

```
class TextFormatter:
    def __init__(self):
        self.env = SandboxedEnvironment(loader=BaseLoader(), autoescape=True)

    # ...

    def format_string(self, template_str: str, values=None, **kwargs) -> str:
        """Regular template rendering with escaping"""
        template = self.env.from_string(template_str)
        return template.render(values or {}, **kwargs)
```

The `SandboxedEnvironment` will always escape passed-in values. (It doesn't even have an `autoescape` option, only the non-sandboxed `Environment` does.)

Pwuts:
- Resolves #10954

Unnecessary escaping distorts content and so should be disabled wherever the output isn't used in HTML.

### Changes 🏗️

- Disable HTML escaping on prompt value insertion in AI blocks
- Make HTML escaping optional in text formatting and output blocks

#### For code changes:
- [x] I have made a test plan
  - [x] `SandboxedEnvironment(autoescape=False).from_string(template_str).render(values)` doesn't escape characters with HTML entities

### DIFF
--- a/autogpt_platform/backend/backend/blocks/io.py
+++ b/autogpt_platform/backend/backend/blocks/io.py
@@ -10,7 +10,6 @@ from backend.util.settings import Config
 from backend.util.text import TextFormatter
 from backend.util.type import LongTextType, MediaFileType, ShortTextType
 
-formatter = TextFormatter()
 config = Config()
 
 
@@ -132,6 +131,11 @@ class AgentOutputBlock(Block):
             default="",
             advanced=True,
         )
+        escape_html: bool = SchemaField(
+            default=False,
+            advanced=True,
+            description="Whether to escape special characters in the inserted values to be HTML-safe. Enable for HTML output, disable for plain text.",
+        )
         advanced: bool = SchemaField(
             description="Whether to treat the output as advanced.",
             default=False,
@@ -193,6 +197,7 @@ class AgentOutputBlock(Block):
         """
         if input_data.format:
             try:
+                formatter = TextFormatter(autoescape=input_data.escape_html)
                 yield "output", formatter.format_string(
                     input_data.format, {input_data.name: input_data.value}
                 )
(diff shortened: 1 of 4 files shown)


autogpt_platform/backend/backend/blocks/io.py
```diff
@@ -193,6 +197,7 @@ async def run(self, input_data: Input, *args, **kwargs) -> BlockOutput:
         """
         if input_data.format:
             try:
+                formatter = TextFormatter(autoescape=input_data.escape_html)
```
majdyz, line 200: Can we make sure that this is safe?

majdyz (approved):

autogpt_platform/backend/backend/blocks/io.py
Pwuts: - It still uses the `SandboxedEnvironment`, which the docs state can be used to render untrusted templates
- It effectively does the same as `{{ value | safe }}` on all inserted values

PR #11028, opened 2025-10-01
fix(platform): resolve authentication performance and malformed URL issues

majdyz:
## Summary
Fix critical authentication performance bottlenecks and malformed redirect URL bug that were causing infinite loading during login flow.

## Root Cause Analysis
- Backend API endpoints taking 5-11+ seconds causing frontend timeouts
- Timezone update endpoint blocking authentication flow unnecessarily 
- Slow onboarding check making repeated uncached database queries
- Malformed redirect URLs like `/marketplace,%20/marketplace` confusing users

### Non-blocking timezone updates
- Convert timezone update to background task with `asyncio.create_task()`
- Return immediately with requested timezone (eventual consistency)
- Eliminates blocking behavior that delayed authentication responses

### Onboarding check caching
- Add `@cached` decorator with 5-minute TTL to `onboarding_enabled()`
- Prevents repeated database queries during auth flow
- Significant performance improvement for concurrent users

### Malformed URL protection
- Enhanced `validateRedirectUrl()` function in auth callback route
- Detect malformed patterns: `,%20`, `, /`, ` /`, `%20/`
- Default to `/marketplace` for any malformed/invalid URLs with warning logs
- Addresses specific reported issue of malformed redirect destinations

## Technical Details
- All changes maintain backward compatibility
- No breaking changes to existing API contracts
- Enhanced error logging for better debugging
- Avoided over-engineering with global timeouts that could break legitimate long-running operations

## Test Plan
- [x] Login flow completes successfully without infinite loading
- [x] Malformed redirect URLs are caught and handled safely
- [x] Background timezone updates don't block authentication
- [x] All pre-commit hooks pass
- [x] TypeScript compilation successful

## Impact
- Eliminates infinite loading during login process
- Improves authentication response times from 5-11+ seconds to <1 second
- Prevents malformed redirect URLs that cause user confusion
- Better error handling and debugging visibility
- Enhanced user experience with reliable authentication flow

## Files Changed
- `backend/server/routers/v1.py` - Background timezone updates
- `backend/data/onboarding.py` - Onboarding caching
- `frontend/src/app/(platform)/auth/callback/route.ts` - URL validation

### DIFF
--- a/autogpt_platform/backend/backend/data/onboarding.py
+++ b/autogpt_platform/backend/backend/data/onboarding.py
@@ -3,6 +3,7 @@ from typing import Any, Optional
 
 import prisma
 import pydantic
+from autogpt_libs.utils.cache import cached
 from prisma.enums import OnboardingStep
 from prisma.models import UserOnboarding
 from prisma.types import UserOnboardingCreateInput, UserOnboardingUpdateInput
@@ -333,8 +334,13 @@ async def get_recommended_agents(user_id: str) -> list[StoreAgentDetails]:
     ]
 
 
+@cached(maxsize=1, ttl_seconds=300)  # Cache for 5 minutes since this rarely changes
 async def onboarding_enabled() -> bool:
+    """
+    Check if onboarding should be enabled based on store agent count.
+    Cached to prevent repeated slow database queries.
+    """
+    # Use a more efficient query that stops counting after finding enough agents
     count = await prisma.models.StoreAgent.prisma().count(take=MIN_AGENT_COUNT + 1)
-
-    # Onboading is enabled if there are at least 2 agents in the store
+    # Onboarding is enabled if there are at least 2 agents in the store
     return count >= MIN_AGENT_COUNT
--- a/autogpt_platform/backend/backend/server/routers/v1.py
+++ b/autogpt_platform/backend/backend/server/routers/v1.py
@@ -178,8 +178,11 @@ async def update_user_timezone_route(
     user_id: Annotated[str, Security(get_user_id)], request: UpdateTimezoneRequest
 ) -> TimezoneResponse:
     """Update user timezone. The timezone should be a valid IANA timezone identifier."""
-    user = await update_user_timezone(user_id, str(request.timezone))
-    return TimezoneResponse(timezone=user.timezone)
+    # Start timezone update as background task - don't block the response
+    asyncio.create_task(update_user_timezone(user_id, str(request.timezone)))
+
+    # Return immediately with the requested timezone (eventual consistency)
+    return TimezoneResponse(timezone=str(request.timezone))
 
 
 @v1_router.get(
--- a/autogpt_platform/frontend/src/app/(platform)/auth/callback/route.ts
+++ b/autogpt_platform/frontend/src/app/(platform)/auth/callback/route.ts
@@ -11,14 +11,38 @@ async function shouldShowOnboarding() {
   );
 }
 
-// Validate redirect URL to prevent open redirect attacks
+// Validate redirect URL to prevent open redirect attacks and malformed URLs
 function validateRedirectUrl(url: string): string {
-  // Only allow relative URLs that start with /
-  if (url.startsWith("/") && !url.startsWith("//")) {
-    return url;
+  try {
+    // Clean up the URL first
+    const cleanUrl = url.trim();
+
+    // Check for malformed URL patterns that could cause issues
+    if (
+      cleanUrl.includes(",%20") ||
+      cleanUrl.includes(", /") ||
+      cleanUrl.includes(" /")
+    ) {
+      console.warn("Detected malformed redirect URL:", cleanUrl);
+      return "/marketplace"; // Default to marketplace for malformed URLs
+    }
+
+    // Only allow relative URLs that start with /
+    if (cleanUrl.startsWith("/") && !cleanUrl.startsWith("//")) {
+      // Additional validation for common problematic patterns
+      if (cleanUrl.includes("%20/") || cleanUrl.split("/").length > 10) {
+        console.warn("Suspicious redirect URL pattern:", cleanUrl);
+        return "/marketplace";
+      }
+      return cleanUrl;
+    }
+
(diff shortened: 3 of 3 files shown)


autogpt_platform/frontend/src/hooks/useOnboardingTimezoneDetection.ts
```diff
@@ -47,11 +61,6 @@ export const useOnboardingTimezoneDetection = () => {
       }
     };
 
-    // Small delay to ensure user is created
-    const timer = setTimeout(() => {
```
majdyz, line 51: This is so suspicious, @kcze @Abhi1992002 do you think this fix makes sense ?

autogpt_platform/frontend/src/hooks/useOnboardingTimezoneDetection.ts
```diff
@@ -47,11 +61,6 @@ export const useOnboardingTimezoneDetection = () => {
...
-      detectAndSetTimezone();
-    }, 1000);
-
-    return () => clearTimeout(timer);
-  }, []); // Run once on mount
+    detectAndSetTimezone();
+  }, [pathname, updateTimezone, user, isUserLoading]);
```
ntindle: Having the pathname here means it’ll run this on every route which we probs don’t want? Idk

majdyz: See the `if (!pathname.startsWith("/onboarding")) {
      return;
    }```

autogpt_platform/backend/backend/data/onboarding.py
```diff
@@ -374,8 +334,13 @@ async def get_recommended_agents(user_id: str) -> list[StoreAgentDetails]:
     ]
 
 
+@cached(maxsize=1, ttl_seconds=300)  # Cache for 5 minutes since this rarely changes
```
kcze, line 378: Assuming that the cache is valid since container startup this will require restarting when adding marketplace agents locally on a fresh db when testing (or waiting 5 minutes). I don't think this is a problem and I even propose caching it for the entirety of lifetime since container startup if possible.
I'd have done that already if the regular cache decorator worked on async so thanks for that implementation!

autogpt_platform/frontend/src/hooks/useOnboardingTimezoneDetection.ts
majdyz: also you added hasAttemptedDetection isn't that what it's for?

AutoGPT-Agent:
Thank you for this well-documented PR addressing authentication performance bottlenecks! The changes look solid and align with the issues described.

A couple of minor points:

1. Our PR template requires a checklist to be filled out, which is missing in your submission. While your detailed validation section covers some of these aspects, please add the standard checklist for completeness.

2. It would be helpful to reference any related issues if they exist, though your root cause analysis is already quite thorough.

The technical changes themselves look excellent:
- The backend caching implementation for expensive onboarding queries is a smart optimization
- Moving timezone detection out of the authentication flow is a good pattern
- The improved URL validation in the callback route adds important safety checks
- The route-based optimizations in the OnboardingProvider should significantly reduce unnecessary database calls

Your performance improvements and race condition fixes should make a real difference to the user experience. Please add the required checklist, and this PR should be good to go!

autogpt_platform/frontend/src/hooks/useOnboardingTimezoneDetection.ts
kcze: It makes sense to me but I don't know the full context of the timezone code; @ntindle

autogpt_platform/frontend/src/hooks/useOnboardingTimezoneDetection.ts
```diff
@@ -47,11 +64,6 @@ export const useOnboardingTimezoneDetection = () => {
...
 
-    // Small delay to ensure user is created
-    const timer = setTimeout(() => {
-      detectAndSetTimezone();
-    }, 1000);
-
-    return () => clearTimeout(timer);
-  }, []); // Run once on mount
+    detectAndSetTimezone();
+  }, [isOnOnboardingRoute, updateTimezone, user, isUserLoading]); // Use computed boolean to reduce re-renders
```
ntindle, line 68: I don’t think the solution we went with here is the correct one. At best I think we do path name and check if user exists and if it contains onboarding so it only runs once per page. Theres a bit of thinking required behind this imo

autogpt_platform/frontend/src/hooks/useOnboardingTimezoneDetection.ts
ntindle: You need the page’s stuff to be ready for you to pull the timezone details. A sleep is a cheap way to do that on a page that will def be shown for more than a second — keeping in mind this should be shown on onboarding

autogpt_platform/frontend/src/hooks/useOnboardingTimezoneDetection.ts
majdyz: would be great to point out which part of the solution here is not the correct one so we can fix it.

autogpt_platform/frontend/src/hooks/useOnboardingTimezoneDetection.ts
majdyz: I think we should define what the "stuff" is (which I initially thought was the `user` object, reading from the code, but please correct me) and use proper code for it rather than calling the function too early.

autogpt_platform/frontend/src/hooks/useOnboardingTimezoneDetection.ts
majdyz: I'm still open to post-merge / follow-up fixes on this if there is any.

autogpt_platform/frontend/src/hooks/useOnboardingTimezoneDetection.ts
majdyz: I'm still open to followup fixes if there is any

PR #11090, opened 2025-10-06
fix(frontend): fix builder's page dynamic title

Abhi1992002:
The issue we’re facing is that we have Next.js metadata in `layout.tsx` that sets the title of the entire application. It’s a default title, but we need the dynamic document title for the builder’s page. Since the builder’s page is client-side, we use `useEffect` and `document.title` for this purpose. However, there’s a problem with this approach. Initially, we mount the title, but as we hydrate the server-side metadata, it gets removed and replaced with the layout’s default title.

Screenshot 2025-10-06 at 10 31 07 AM

### Changes 🏗️

- I’ve converted the page to server-side.
- Transferred all client-side code to a client component and imported it into the server-side page.

### Tests
- [x] Saving and loading of graphs is working perfectly.
- [x] The builder’s dynamic title is set perfectly.


autogpt_platform/frontend/src/app/(platform)/build/components/MainBuilderPage.tsx
Pwuts, line 1: wouldn't this go in `../page.tsx`?

0ubbe (approved):

PR #11092, opened 2025-10-06
feat(platform): Log Marketplace Search Terms

Swiftyos:
We want to be able to log the search terms used in the marketplace so that we can understand what problems people need to solve. 

### Changes 🏗️

- added search history table. We log, time it was searched for and the search term.
- added function to log search terms

#### For code changes:
- [x] I have made a test plan
  
  - [x] Search for various things in the marketplace
  - [x] check the search terms have been stored in the database

### DIFF
--- a/autogpt_platform/Makefile
+++ b/autogpt_platform/Makefile
@@ -1,0 +1,47 @@
+.PHONY: start-core stop-core logs-core format lint migrate run-backend run-frontend
+
+# Run just Supabase + Redis + RabbitMQ
+start-core:
+	docker compose up -d deps
+
+# Stop core services
+stop-core:
+	docker compose stop deps
+
+# View logs for core services
+logs-core:
+	docker compose logs -f deps
+
+# Run formatting and linting for backend and frontend
+format:
+	cd backend && poetry run format
+	cd frontend && pnpm format
+	cd frontend && pnpm lint
+
+init-env:
+	cp -n .env.default .env || true
+	cd backend && cp -n .env.default .env || true
+	cd frontend && cp -n .env.default .env || true
+
+
+# Run migrations for backend
+migrate:
+	cd backend && poetry run prisma migrate deploy
+	cd backend && poetry run prisma generate
+
+run-backend:
+	cd backend && poetry run app
+
+run-frontend:
+	cd frontend && pnpm dev
+
+help:
+	@echo "Usage: make <target>"
+	@echo "Targets:"
+	@echo "  start-core - Start just the core services (Supabase, Redis, RabbitMQ) in background"
+	@echo "  stop-core - Stop the core services"
+	@echo "  logs-core - Tail the logs for core services"
+	@echo "  format - Format & lint backend (Python) and frontend (TypeScript) code"
+	@echo "  migrate - Run backend database migrations"
+	@echo "  run-backend - Run the backend FastAPI server"
+	@echo "  run-frontend - Run the frontend Next.js development server"
--- a/autogpt_platform/README.md
+++ b/autogpt_platform/README.md
@@ -38,6 +38,37 @@ To run the AutoGPT Platform, follow these steps:
 
 4. After all the services are in ready state, open your browser and navigate to `http://localhost:3000` to access the AutoGPT Platform frontend.
 
+### Running Just Core services
+
+You can now run the following to enable just the core services.
+
+```
+# For help
+make help
+
+# Run just Supabase + Redis + RabbitMQ
+make start-core
+
+# Stop core services
+make stop-core
+
+# View logs from core services 
+make logs-core
+
+# Run formatting and linting for backend and frontend
+make format
+
+# Run migrations for backend database
+make migrate
+
+# Run backend server
+make run-backend
+
+# Run frontend development server
+make run-frontend
+
+```
+
 ### Docker Compose Commands
 
 Here are some useful Docker Compose commands for managing your AutoGPT Platform:
--- a/docs/content/platform/getting-started.md
+++ b/docs/content/platform/getting-started.md
@@ -23,7 +23,7 @@ To setup the server, you need to have the following installed:
 
 We use Node.js to run our frontend application.
 
-If you need assistance installing Node.js:
+If you need assistance installing Node.js:  
 https://nodejs.org/en/download/
 
 NPM is included with Node.js, but if you need assistance installing NPM:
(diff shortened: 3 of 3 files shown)


AutoGPT-Agent:
Thank you for adding this useful feature to help us understand what problems users are trying to solve through marketplace searches!

Your implementation looks clean and straightforward. The PR description clearly explains the purpose and changes, and your checklist is properly completed with a sensible test plan.

There's just one issue that needs to be addressed before merging:

- There's a typo in your PR title: 'platorm' should be 'platform' to conform to our conventional commit format.

Once you fix the title, this PR should be ready to merge. Nice work on implementing this tracking feature!

autogpt_platform/backend/backend/server/v2/store/db.py
```diff
@@ -81,6 +81,7 @@ async def get_store_agents(
         where_clause["categories"] = {"has": category}
 
     if sanitized_query:
+        await log_search_term(sanitized_query)
```
Pwuts: Nit: to keep this from adding latency to the response, I'd recommend putting the log statement in a `finally` clause like this:
```py
try:
    # ... query etc
    return results
except ...:
    # error handling
finally:
    await log_search_term(sanitized_query)  # this runs after the function has returned
```

autogpt_platform/backend/backend/server/v2/store/db.py
```diff
@@ -147,6 +148,13 @@ async def get_store_agents(
         ) from e
 
 
+async def log_search_term(search_query: str):
+    """Log a search term to the database"""
+    await prisma.models.SearchHistory.prisma().create(
+        data=prisma.types.SearchHistoryCreateInput(searchQuery=search_query)
```
Pwuts: afaik we can just do this now:
```suggestion
        data={"searchQuery": search_query}
```

autogpt_platform/backend/migrations/20251006084123_log_search_terms/migration.sql
```diff
@@ -0,0 +1,11 @@
+-- CreateTable
+CREATE TABLE "SearchHistory" (
+    "id" BIGSERIAL NOT NULL,
+    "createdAt" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    "searchQuery" TEXT NOT NULL,
+
+    CONSTRAINT "SearchHistory_pkey" PRIMARY KEY ("id")
+);
```
Pwuts: No `user_id` link -> we can't delete this if a user requests their data be deleted. Are we sure this is anonymous enough not to delete upon such a request?

autogpt_platform/backend/schema.prisma
```diff
@@ -587,6 +587,15 @@ model CreditRefundRequest {
 ////////////////////////////////////////////////////////////
 ////////////////////////////////////////////////////////////
 
+model SearchHistory {
+  // Not using uuid as we want to minimise table size
+  id BigInt @id @default(autoincrement())
+  createdAt DateTime @default(now())
+  searchQuery String
```
Pwuts: @claude pls run `poetry run prisma format` here

autogpt_platform/backend/backend/server/v2/store/db.py
Swiftyos: Thanks this is a nice suggestion, adding it in now

autogpt_platform/backend/backend/server/v2/store/db.py
Swiftyos: What is the preferred method now for our codebase?

autogpt_platform/backend/migrations/20251006084123_log_search_terms/migration.sql
Swiftyos: Correct this is intentional we don't want to tie search terms to users

AutoGPT-Agent:
Thank you for adding this useful feature to track marketplace search terms! Your PR is almost ready for merging, but there's one important inconsistency that needs to be fixed:

1. There's a mismatch between your migration file and schema.prisma:
   - In the migration file, the column is named `searchTerm`
   - In the schema.prisma file, the same column is named `searchQuery`

Please make these names consistent to prevent database errors. It's recommended to use the same field name in both places.

Also, as a minor note, you've renamed `sanitized_query` to `search_term` in the `get_store_agents()` function. While this change is fine, it's a bit outside the scope of the feature implementation. It's not a blocker for merging, but it's good to be aware of this type of change.

AutoGPT-Agent:
Thank you for this PR to add search term logging to the marketplace!

The implementation is straightforward and your test plan looks good. I have one main concern before this can be merged:

### Security Concern
The new `log_search_term` function doesn't include a `user_id` parameter, which is generally required for backend data functions. Since you're storing user search behavior, consider:

1. Either modify the function to include and validate the user_id parameter
2. Or provide an explanation for why user_id validation isn't needed in this specific case (e.g., if this is anonymous analytics data that doesn't need user-specific permissions)

Additionally, I noticed the field name in the schema.prisma (`searchQuery`) doesn't match the field name in the migration (`searchTerm`). These should be consistent.

Once these items are addressed, this should be ready to merge. The changes are otherwise well-structured and match the PR description and title.

autogpt_platform/backend/backend/server/v2/store/db.py
```diff
@@ -145,6 +144,22 @@ async def get_store_agents(
...
+    finally:
+        if search_term:
+            await log_search_term(search_query=search_term)
+
+
+async def log_search_term(search_query: str):
+    """Log a search term to the database"""
+
+    # Anonymize the data by preventing correlation with other logs
+    date = datetime.now(timezone.utc).replace(hour=0, minute=0, second=0, microsecond=0)
```
Pwuts, line 156: this is a neat trick

autogpt_platform/backend/backend/server/v2/store/db.py
Pwuts: TypedDict instance statements (e.g. `prisma.types.SearchHistoryCreateInput(..)`) give unreadable argument errors when invalid, whereas dict literals give much more useful feedback, so dict literals are preferred

ntindle (approved):

autogpt_platform/backend/backend/server/v2/store/db.py
Swiftyos: Okay thank you @Pwuts

PR #11125, opened 2025-10-08
fix(frontend): login redirects + onboarding

0ubbe:
### Fix re-direct bugs

Sometimes the app will re-direct to a strange URL after login:
```
http://localhost:3000/marketplace,%20/marketplace
```
It looks like a race-condition because the re-direct to `/marketplace` was done on a [server action](https://nextjs.org/docs/14/app/building-your-application/data-fetching/server-actions-and-mutations) rather than in the browser.

**✅ Fixed by** 

Moving the login / signup server actions to Next.js API endpoints. In this way the login/signup pages just call an API endpoint and handle its response without having to hassle with serverless 💆🏽 

### Wallet layout flash

Screenshot 2025-10-08 at 22 52 03

The wallet popover would sometimes flash after login, because it was re-rendering once onboarding and credits data loaded.

**✅ Fixed by** 

Only rendering once we have onboarding and credits data, without the popover is useless and causes flashes.

### For code changes:
- [x] I have made a test plan
  - [x] Login / Signup to the app with email and Google
  - [x] Works fine
  - [x] Onboarding popover does not flash
  - [x] Onboarding and marketplace re-directs work   

None

### DIFF
--- a/autogpt_platform/frontend/src/app/(platform)/auth/callback/route.ts
+++ b/autogpt_platform/frontend/src/app/(platform)/auth/callback/route.ts
@@ -2,14 +2,7 @@ import { getServerSupabase } from "@/lib/supabase/server/getServerSupabase";
 import BackendAPI from "@/lib/autogpt-server-api";
 import { NextResponse } from "next/server";
 import { revalidatePath } from "next/cache";
-
-async function shouldShowOnboarding() {
-  const api = new BackendAPI();
-  return (
-    (await api.isOnboardingEnabled()) &&
-    !(await api.getUserOnboarding()).completedSteps.includes("CONGRATS")
-  );
-}
+import { shouldShowOnboarding } from "@/app/api/helpers";
 
 // Handle the callback to complete the user session login
 export async function GET(request: Request) {
--- a/autogpt_platform/frontend/src/app/api/auth/login/route.ts
+++ b/autogpt_platform/frontend/src/app/api/auth/login/route.ts
@@ -4,6 +4,7 @@ import { verifyTurnstileToken } from "@/lib/turnstile";
 import { loginFormSchema } from "@/types/auth";
 import * as Sentry from "@sentry/nextjs";
 import { NextResponse } from "next/server";
+import { shouldShowOnboarding } from "../../helpers";
 
 export async function POST(request: Request) {
   try {
@@ -62,12 +63,4 @@ export async function POST(request: Request) {
       { status: 500 },
     );
   }
-}
-
-async function shouldShowOnboarding() {
-  const api = new BackendAPI();
-  const isEnabled = await api.isOnboardingEnabled();
-  const onboarding = await api.getUserOnboarding();
-  const isCompleted = onboarding.completedSteps.includes("CONGRATS");
-  return isEnabled && !isCompleted;
 }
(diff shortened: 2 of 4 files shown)


autogpt_platform/frontend/src/app/(platform)/login/useLoginPage.ts
```diff
@@ -60,18 +59,32 @@ export function useLoginPage() {
     }
 
     try {
-      const error = await providerLogin(provider);
-      if (error) throw error;
-      setFeedback(null);
+      const response = await fetch("/api/auth/provider", {
```
majdyz, line 62: I thought we already have a helper for doing any of these API calls ?

autogpt_platform/frontend/src/app/(platform)/login/useLoginPage.ts
```diff
@@ -60,18 +59,32 @@ export function useLoginPage() {
...
+        } else {
+          setFeedback(error || "Failed to start OAuth flow");
+        }
+        resetCaptcha();
+        setIsGoogleLoading(false);
+        return;
+      }
+
+      const { url } = await response.json();
+      if (url) window.location.href = url as string;
```
majdyz, line 81: won't this cause page reload? is this expected to not use router/pathname at all here?

autogpt_platform/frontend/src/app/(platform)/login/useLoginPage.ts
```diff
@@ -99,21 +112,49 @@ export function useLoginPage() {
...
+        method: "POST",
+        headers: { "Content-Type": "application/json" },
+        body: JSON.stringify({
+          email: data.email,
+          password: data.password,
+          turnstileToken: turnstile.token,
+        }),
+      });
+
+      const result = await response.json();
```
majdyz, line 126: samething here, seems like the logic can be unified ? definitely not forcing it though just random idea

majdyz: Same with the ret of API calls we have below

autogpt_platform/frontend/src/app/(platform)/login/useLoginPage.ts
0ubbe: In this case, this is an internal Next.js API route made on the Front-end. 

Our API helpers are designed to interact with the AutoGPT API, whereas this is not; you can consider them 2 different APIs. Internal Next.js API routes are only made for convenience, like this case, most of the time we will call the AutoGPT API ( _which actually could migrate this logic there_ 👀 )

autogpt_platform/frontend/src/app/(platform)/login/useLoginPage.ts
0ubbe: In our flow, we call Supabase OAuth on the server (API route). That returns a URL; it does not perform a top-level redirect. Because the client calls the API with fetch, the browser won’t navigate on a `302` either. So we must explicitly navigate with `window.location` (or replace) to leave the app.

That said, I will be 100% 💯 confident it works after merge and try on Dev given Sign in with Google only works there not locally...

autogpt_platform/frontend/src/app/(platform)/login/useLoginPage.ts
0ubbe: I will merge to try the changes, could you elaborate on what you think can be unified so I can address on a follow up PR ?

autogpt_platform/frontend/src/app/api/auth/provider/route.ts
```diff
@@ -0,0 +1,49 @@
...
+
+    const { data, error } = await supabase.auth.signInWithOAuth({
+      provider,
+      options: {
+        redirectTo:
+          redirectTo ||
+          process.env.AUTH_CALLBACK_URL ||
+          `http://localhost:3000/auth/callback`,
+      },
+    });
```
ntindle, line 31: Did we set this ever??

PR #11149, opened 2025-10-13
feat(backend): Whitelist Onboarding Agents

kcze:
Some agents aren't suitable for onboarding. This adds per-store agent setting to allow them for onboarding. In case no agent is allowed fallback to the former search.

### Changes 🏗️

- Add `useForOnboarding` to `StoreListing` model and `StoreAgent` view (with migration)
- Remove filtering of agents with empty input schema or credentials 

#### For code changes:
- [x] I have made a test plan
  - [x] Only allowed agents are displayed
  - [x] Fallback to the old system in case there aren't enough allowed agents

### DIFF
--- a/autogpt_platform/backend/backend/data/onboarding.py
+++ b/autogpt_platform/backend/backend/data/onboarding.py
@@ -11,7 +11,6 @@ from prisma.types import UserOnboardingCreateInput, UserOnboardingUpdateInput
 
 from backend.data.block import get_blocks
 from backend.data.credit import get_user_credit_model
-from backend.data.graph import GraphModel
 from backend.data.model import CredentialsMetaInput
 from backend.server.v2.store.model import StoreAgentDetails
 from backend.util.json import SafeJson
@@ -278,8 +277,14 @@ async def get_recommended_agents(user_id: str) -> list[StoreAgentDetails]:
         for word in user_onboarding.integrations
     ]
 
+    where_clause["is_available"] = True
+
+    # Try to take only agents that are available and allowed for onboarding
     storeAgents = await prisma.models.StoreAgent.prisma().find_many(
-        where=prisma.types.StoreAgentWhereInput(**where_clause),
+        where={
+            "is_available": True,
+            "useForOnboarding": True,
+        },
         order=[
             {"featured": "desc"},
             {"runs": "desc"},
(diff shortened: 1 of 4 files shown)


autogpt_platform/backend/schema.prisma
```diff
@@ -744,6 +745,9 @@ model StoreListing {
   // URL-friendly identifier for this agent (moved from StoreListingVersion)
   slug String
 
+  // Allow this agent to be used during onboarding
+  useForOnboarding   Boolean @default(false)
```
majdyz, line 749: Why do we need to duplicate this info into two tables? would a single source of truth problematic? how do we keep it in sync?

autogpt_platform/backend/schema.prisma
```diff
@@ -667,6 +667,7 @@ view StoreAgent {
   rating           Float
   versions         String[]
   is_available     Boolean  @default(true)
+  useForOnboarding Boolean  @default(false)
```
majdyz, line 670: why we set them false by default instead of true?

autogpt_platform/backend/backend/data/onboarding.py
```diff
@@ -278,8 +277,14 @@ async def get_recommended_agents(user_id: str) -> list[StoreAgentDetails]:
...
     ]
 
+    where_clause["is_available"] = True
+
+    # Try to take only agents that are available and allowed for onboarding
     storeAgents = await prisma.models.StoreAgent.prisma().find_many(
-        where=prisma.types.StoreAgentWhereInput(**where_clause),
+        where={
+            "is_available": True,
+            "useForOnboarding": True,
```
majdyz, line 286: What about the rest of the where clause? Is it not included in this filter? Is this going to be a lot then?

autogpt_platform/backend/schema.prisma
kcze: The second one is view not table

autogpt_platform/backend/schema.prisma
kcze: We want a small curated list of approved agents now

autogpt_platform/backend/backend/data/onboarding.py
kcze: Now we'll have a small list (way below 100 limit), so it's a future problem. All agents still go through the scoring process, so whitelisted agents don't need that filtering.

PR #11170, opened 2025-10-15
feat(backend): Add Sentry user and tag tracking to node execution

ntindle:
Integrates Sentry SDK to set user and contextual tags during node execution for improved error tracking and user count analytics. Ensures Sentry context is properly set and restored, and exceptions are captured with relevant context before scope restoration.

- Related: https://github.com/getsentry/sentry-python/issues/4925

### Changes 🏗️
Adds sentry tracking to block failures

#### For code changes:
- [x] I have made a test plan
  
  - [x] Test to make sure the userid and block details show up in Sentry
  - [x] make sure other errors aren't contaminated 

## Summary by CodeRabbit

- New Features
  - Added conditional support for feature flags when configured, enabling targeted rollouts and experiments without impacting unconfigured environments.

- Chores
  - Enhanced error monitoring with richer contextual data during node execution to improve stability and diagnostics.
  - Updated metrics initialization to dynamically include feature flag integrations when available, without altering behavior for unconfigured setups.

### DIFF
--- a/autogpt_platform/backend/backend/executor/manager.py
+++ b/autogpt_platform/backend/backend/executor/manager.py
@@ -1494,6 +1494,7 @@ class ExecutionManager(AppProcess):
 
         graph_exec_id = graph_exec_entry.graph_exec_id
         user_id = graph_exec_entry.user_id
+        graph_id = graph_exec_entry.graph_id
         logger.info(
             f"[{self.service_name}] Received RUN for graph_exec_id={graph_exec_id}, user_id={user_id}"
         )
@@ -1503,6 +1504,7 @@ class ExecutionManager(AppProcess):
             # Only check executions from the last 24 hours for performance
             current_running_count = get_db_client().get_graph_executions_count(
                 user_id=user_id,
+                graph_id=graph_id,
                 statuses=[ExecutionStatus.RUNNING],
                 created_time_gte=datetime.now(timezone.utc) - timedelta(hours=24),
             )
@@ -1512,7 +1514,7 @@ class ExecutionManager(AppProcess):
                 >= settings.config.max_concurrent_graph_executions_per_user
             ):
                 logger.warning(
-                    f"[{self.service_name}] Rate limit exceeded for user {user_id}: "
+                    f"[{self.service_name}] Rate limit exceeded for user {user_id} on graph {graph_id}: "
                     f"{current_running_count}/{settings.config.max_concurrent_graph_executions_per_user} running executions"
                 )
                 _ack_message(reject=True, requeue=True)
--- a/autogpt_platform/backend/backend/data/execution.py
+++ b/autogpt_platform/backend/backend/data/execution.py
@@ -479,16 +479,18 @@ async def get_graph_executions(
 
 
 async def get_graph_executions_count(
-    user_id: str,
+    user_id: Optional[str] = None,
+    graph_id: Optional[str] = None,
     statuses: Optional[list[ExecutionStatus]] = None,
     created_time_gte: Optional[datetime] = None,
     created_time_lte: Optional[datetime] = None,
 ) -> int:
     """
-    Get count of graph executions for a user with optional filters.
+    Get count of graph executions with optional filters.
 
     Args:
-        user_id: The user ID to filter by
+        user_id: Optional user ID to filter by
+        graph_id: Optional graph ID to filter by
         statuses: Optional list of execution statuses to filter by
         created_time_gte: Optional minimum creation time
         created_time_lte: Optional maximum creation time
@@ -498,8 +500,13 @@ async def get_graph_executions_count(
     """
     where_filter: AgentGraphExecutionWhereInput = {
         "isDeleted": False,
-        "userId": user_id,
     }
+
+    if user_id:
+        where_filter["userId"] = user_id
+
+    if graph_id:
+        where_filter["agentGraphId"] = graph_id
 
     if created_time_gte or created_time_lte:
         where_filter["createdAt"] = {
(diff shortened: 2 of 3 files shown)


AutoGPT-Agent:
Thanks for adding Sentry user and tag tracking to node execution. This will be valuable for error tracking and analytics.

Before we can merge this PR, please:

1. Complete your test plan by checking off the remaining item ("make sure other errors aren't contaminated") or indicate if this test is no longer applicable.

The implementation looks solid - I like how you're saving and restoring the original scope to prevent context contamination, and the flushing of errors before scope restoration ensures the context is captured correctly.

autogpt_platform/backend/backend/executor/manager.py
```diff
@@ -246,6 +270,10 @@ async def execute_node(
             execution_stats.input_size = input_size
             execution_stats.output_size = output_size
 
+        # Restore scope AFTER error has been captured
+        scope._user = original_user
+        scope._tags = original_tags
```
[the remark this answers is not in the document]
ntindle: not true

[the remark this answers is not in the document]
ntindle: fixed first, empty dict is okay

autogpt_platform/backend/backend/executor/manager.py
```diff
@@ -232,6 +250,12 @@ async def execute_node(
             output_size += len(json.dumps(output_data))
             log_metadata.debug("Node produced output", **{output_name: output_data})
             yield output_name, output_data
+    except Exception:
```
majdyz, line 253: let;s catch BaseException so we catch interruption too

majdyz: oh wait I think we should do it on `_on_node_execution` there is already try catch for baseException there,

Pwuts:
@coderabbitai review

autogpt_platform/backend/backend/executor/manager.py
```diff
@@ -224,6 +225,23 @@ async def execute_node(
...
+    original_tags = dict(scope._tags) if scope._tags else {}
+    # Set user ID for error tracking
+    scope.set_user({"id": user_id})
+
+    scope.set_tag("graph_id", graph_id)
+    scope.set_tag("node_id", node_id)
+    scope.set_tag("block_name", node_block.name)
+    scope.set_tag("block_id", node_block.id)
+    for k, v in (data.user_context or UserContext(timezone="UTC")).model_dump().items():
+        scope.set_tag(f"user_context.{k}", v)
```
Pwuts, line 243: @claude I'm not familiar with this Sentry stuff, what are isolation scopes and why is the scope here different than outside this code block?

ntindle: We want a different scope here because it's a swap from system behavior to user behavior triggering the error. Scopes can contain a lot of different context and merge downward. We want a scope here to indicate to sentry that "this is now user errors level with potential system errors" rather than "system errors" and for that to swap back at the end we need to reset the scope tags because a bug in sentry. 

You can read more about them here

https://docs.sentry.io/platforms/python/enriching-events/scopes/

Once the next sentry release goes out we can clean this up with isolation scopes but they had a bug they've since fixed (thanks to a very cool bug report https://github.com/getsentry/sentry-python/pull/4928)

autogpt_platform/backend/backend/executor/manager.py
ntindle: We can push it to this block but that will mean stuff like the credential lock will appear as attached to a user if it fails rather than just blocks. Is this tradeoff okay with you?
```python

try:
            log_metadata.info(f"Start node execution {node_exec.node_exec_id}")
            await async_update_node_execution_status(
                db_client=db_client,
                exec_id=node_exec.node_exec_id,
                status=ExecutionStatus.RUNNING,
            )

            async for output_name, output_data in execute_node(
                node=node,
                creds_manager=self.creds_manager,
                data=node_exec,
                execution_stats=stats,
                nodes_input_masks=nodes_input_masks,
            ):
                await persist_output(output_name, output_data)

            log_metadata.info(f"Finished node execution {node_exec.node_exec_id}")
            status = ExecutionStatus.COMPLETED

        except BaseException as e:
            stats.error = e

            if isinstance(e, ValueError):
                # Avoid user error being marked as an actual error.
                log_metadata.info(
                    f"Expected failure on node execution {node_exec.node_exec_id}: {type(e).__name__} - {e}"
                )
```

majdyz: I see, both works actuallt. let's stick with the current one

autogpt_platform/backend/backend/util/metrics.py
```diff
@@ -5,8 +5,10 @@
 from pydantic import SecretStr
 from sentry_sdk.integrations.anthropic import AnthropicIntegration
 from sentry_sdk.integrations.asyncio import AsyncioIntegration
+from sentry_sdk.integrations.launchdarkly import LaunchDarklyIntegration
 from sentry_sdk.integrations.logging import LoggingIntegration
```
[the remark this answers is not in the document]
ntindle: Not true

autogpt_platform/backend/backend/executor/manager.py
ntindle: I'll add handling base exception then it's good to go?

ntindle: Actually once the sentry version is out I'll want to redo this with scopes anyway

PR #11172, opened 2025-10-15
feat(frontend): add extra info in custom node in new builder

Abhi1992002:
Currently, we don’t add category and cost information to custom nodes in the new builder. so now we’re rendering with the correct information and costs are displayed accurately based on the selected discriminator value.

Screenshot 2025-10-15 at 2 43 33 PM

#### For code changes:
- [x] I have made a test plan
  - [x] All information is displayed correctly.
  - [x] I’ve tried changing the discrimination value and we’re getting the correct cost for the selected value.


majdyz:
I'm pulling myself from the assigned reviewer since I might be losing context for the review. but feel free to retag if my review is actually needed

PR #10323, opened 2025-07-07
feat(platform/builder): Builder credentials support + UX improvements

Pwuts:
- Resolves #10313
- Resolves #10333

Before:

https://github.com/user-attachments/assets/a105b2b0-a90b-4bc6-89da-bef3f5a5fa1f
- No credentials input
- Stuttery experience when panning or zooming the viewport

After:

https://github.com/user-attachments/assets/f58d7864-055f-4e1c-a221-57154467c3aa
- Pretty much the same UX as in the Library, with fully-fledged credentials input support
- Much smoother when moving around the canvas

### Changes 🏗️

Frontend:
- Add credentials input support to Run UX in Builder
  - Pass run inputs instead of storing them on the input nodes
  - Re-implement `RunnerInputUI` using `AgentRunDetailsView`; rename to `RunnerInputDialog`
    - Make `AgentRunDraftView` more flexible
    - Remove `RunnerInputList`, `RunnerInputBlock`
- Make moving around in the Builder *smooooth* by reducing unnecessary re-renders
  - Clean up and partially re-write bead management logic
- Replace `request*` fire-and-forget methods in `useAgentGraph` with direct action async callbacks
- Clean up run input UI components
  - Simplify `RunnerUIWrapper`
  - Add `isEmpty` utility function in `@/lib/utils` (expanding on `_.isEmpty`)
  - Fix default value handling in `TypeBasedInput` (**Note:** after all the changes I've made I'm not sure this is still necessary)
- Improve & clean up Builder test implementations

Backend + API:
- Fix front-end `Node`, `GraphMeta`, and `Block` types
- Small refactor of `Graph` to match naming of some `LibraryAgent` attributes
- Fix typing of `list_graphs`, `get_graph_meta_by_store_listing_version_id` endpoints
  - Add `GraphMeta` model and `GraphModel.meta()` shortcut
- Move `POST /library/agents/{library_agent_id}/setup-trigger` to `POST /library/presets/setup-trigger`

#### For code changes:
- [x] I have clearly listed my changes in the PR description
- [x] I have made a test plan
- [x] I have tested my changes according to the test plan:
  - Test the new functionality in the Builder:
    - [x] Running an agent with (credentials) inputs from the builder
      - [x] Beads behave correctly
    - [x] Running an agent without any inputs from the builder
    - [x] Scheduling an agent from the builder
    - [x] Adding and searching blocks in the block menu
  - [x] Test that all existing `AgentRunDraftView` functionality in the Library still works the same
[...]

Issue #9139, opened 2024-12-28
Agent Input Values Shouldn't Export or Save


Torantulino:
Currently the last value that you used in an Agent Input Block exports into the Agent JSON. This means that the Agent file is polluted with the last input the user tested their Agent with.<br><br>*This input value also saves to the database.<br>*<br>Steps to reproduce:

1. Make an Agent with an Agent Input Block.
2. Run the Agent with a recognisable input value
3. Go to the Library Page and Export the Agent to a JSON file.
4. Open the JSON File and observe your unique input has been exported with the file.

---

### Intended Behaviour:

The value field on Agent inputs should be empty when saving or exporting Agents.

Torantulino:
Just been plagued by this issue for a while thinking it was some sort of deeper bug. This isn't necessarily a bug but it is a bad and unintended user experience.

Swiftyos:
Done and released

PR #11173, opened 2025-10-15
feat(backend): Include default input values in graph export

Pwuts:
Since #10323, one-time graph inputs are no longer stored on the input nodes (#9139), so we can reasonably assume that the default value set by the graph creator will be safe to export.

### Changes 🏗️

- Don't strip the default value from input nodes in `NodeModel.stripped_for_export(..)`, except for inputs marked as `secret`
- Update and expand tests for graph export secrets stripping mechanism

#### For code changes:
- [x] I have made a test plan
  - [x] Expanded tests pass
  - Relatively simple change with good test coverage, no manual test needed


Pwuts:
@ntindle checked, works fine

PR #11182, opened 2025-10-16
feat(frontend): setup DataFast analytics

0ubbe:
## Changes 🏗️

Following https://datafa.st/docs/nextjs-app-router

### For code changes:
- [x] I have made a test plan
  - [x] We will see once we make a production deployment and get data into the platform

None


autogpt_platform/frontend/src/app/layout.tsx
```diff
@@ -22,6 +24,9 @@ export default async function RootLayout({
 }: Readonly<{
   children: React.ReactNode;
 }>) {
+  const headersList = await headers();
+  const host = headersList.get("host") || "";
+  const isPlatformDomain = host === "platform.agpt.co";
```
ntindle, line 29: i hate this but i'll be okay -- i really prefer we use the getEnv stuff or centralize it a bit but like htats maybe me being a backend person

0ubbe: Agree, I think we should centralize all the env helpers in a single place/service, I'll have a thought 💭

PR #11222, opened 2025-10-21
fix(backend): Only try to initialize LaunchDarkly once

Pwuts:
We currently try to re-init the LaunchDarkly client every time a feature flag is checked.
This causes 5 second extra latency on the flag check when LD is down, such as now.
Since flag checks are performed on every block execution, this currently cripples the platform's executors.

- Follow-up to #11221

### Changes 🏗️

- Only try to init LaunchDarkly once
- Improve surrounding log statements in the `feature_flag` module

#### For code changes:
- [x] I have made a test plan
  - This is a critical hotfix; we'll see its effect once deployed

### DIFF
--- a/autogpt_platform/backend/backend/util/feature_flag.py
+++ b/autogpt_platform/backend/backend/util/feature_flag.py
@@ -63,9 +63,9 @@ def initialize_launchdarkly() -> None:
     config = Config(sdk_key)
     ldclient.set_config(config)
 
+    global _is_initialized
+    _is_initialized = True
     if ldclient.get().is_initialized():
-        global _is_initialized
-        _is_initialized = True
         logger.info("LaunchDarkly client initialized successfully")
     else:
         logger.error("LaunchDarkly client failed to initialize")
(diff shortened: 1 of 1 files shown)


autogpt_platform/backend/backend/util/feature_flag.py
```diff
@@ -63,9 +63,9 @@ def initialize_launchdarkly() -> None:
     config = Config(sdk_key)
     ldclient.set_config(config)
 
+    global _is_initialized
+    _is_initialized = True
```
majdyz, line 67: So it will always be initialized? Whether it's a success or failure?

Pwuts: We mark it as initialized *by us* because the `LDClient` also has internal initialized state.

We still need `_is_initialezed` because there is a difference between
1. The `feature_flag`/LaunchDarkly module hasn't been initialized yet
2. The LaunchDarkly client has been triggered to initialize ...
  - ... and init was successful
  - ... and init failed

`_is_initialized` distinguishes between 1 and 2, `LDClient.is_initialized()` distinguishes between 2A and 2B.

PR #11228, opened 2025-10-21
fix(backend/store): Sanitize all sql terms

Swiftyos:
Categories and Creators where not sanitized in the full text search

### Changes 🏗️

- apply sanitization to categories and creators

#### For code changes:
- [x] I have made a test plan
  
  - [x] run tests to check it still works


autogpt_platform/backend/backend/server/v2/store/db.py
```diff
@@ -98,10 +107,10 @@ async def get_store_agents(
             if featured:
                 filter_conditions.append("featured = true")
             if creators:
-                creator_list = "','".join(creators)
+                creator_list = "','".join(sanitized_creators)
```
Pwuts, line 110: if this is the approach you want to take, it can be much simpler:
```suggestion
                creator_list = "','".join(sanitize_query(c) for c in creators)
```

autogpt_platform/backend/backend/server/v2/store/db.py
```diff
@@ -73,6 +73,15 @@ async def get_store_agents(
...
     )
 
+    sanitized_creators = []
+    if creators:
+        for c in creators:
+            sanitized_creators.append(sanitize_query(c))
+
+    sanitized_category = None
+    if category:
+        sanitized_category = sanitize_query(category)
```
Pwuts, line 83: Can't you feed these into the query as parameters instead of as string inserts?

Swiftyos: this is causing issues, with the dynamic where and order by statements

Pwuts (approved):

Issue #145, opened 2023-04-04
In --gpt3only mode, receive this error: SYSTEM:  Command evaluate_code returned: Error: The model: `gpt-4` does not exist


jtokash:
When running with --gpt3only mode, I receive this error when Auto-GPT tries to 'execute' code.

SYSTEM:  Command evaluate_code returned: Error: The model: `gpt-4` does not exist

Tonylib:
same

ljleb:
Same problem. Can anyone call the openapi for gpt-4 or you need to be granted access?

stayupthetree:
> Same problem. Can anyone call the openapi for gpt-4 or you need to be granted access?

From what I can tell you need to sign up on a waitlist for gpt-4 API

griffintv:
Can you disable and only have it run on 3.5?

jcdescy:
I get the same, but I'm not running in gpt3only mode. It was when the AI had written a bot and tried to evaluate the code.

ljleb:
To disable gpt-4 just edit the configuration, set `SMART_LLM_MODEL="gpt-3.5-turbo"` in `.env`.

jtokash:
@ljleb 's response is correct - there are 2 models referenced in .env.  I only had one set to 3.5.

Wladastic:
PINECONE_API_KEY=your-pinecone-api-key
PINECONE_ENV=your-pinecone-region
OPENAI_API_KEY=
ELEVENLABS_API_KEY=
SMART_LLM_MODEL="gpt-3.5-turbo-0301"
FAST_LLM_MODEL="gpt-3.5-turbo"
GOOGLE_API_KEY=
CUSTOM_SEARCH_ENGINE_ID=
USE_AZURE=False
OPENAI_API_BASE=your-base-url-for-azure
OPENAI_API_VERSION=api-version-for-azure
OPENAI_DEPLOYMENT_ID=deployment-id-for-azure
IMAGE_PROVIDER=dalle
HUGGINGFACE_API_TOKEN=

omni-webdev:
--gpt3only also resolves, don't have to set the env variable


> To disable gpt-4 just edit the configuration, set `SMART_LLM_MODEL="gpt-3.5-turbo"` in `.env`.

--gpt3only also resolves, don't have to set the env variable

ljleb:
> @omni-webdev --gpt3only also resolves, don't have to set the env variable

[...]

PR #11232, opened 2025-10-21
fix(frontend): URL substring sanitazion issue

0ubbe:
Potential fix for [https://github.com/Significant-Gravitas/AutoGPT/security/code-scanning/145](https://github.com/Significant-Gravitas/AutoGPT/security/code-scanning/145)

To fix the issue, rather than using substring matching on the raw URL string, we need to properly parse the URL and inspect its hostname. We should confirm that the `hostname` property of the parsed URL is equal to either `vimeo.com` or explicitly permitted subdomains like `www.vimeo.com`. We can use the native JavaScript `URL` class for robust parsing.

**File/Location:**  
- Only change line(s) in `autogpt_platform/frontend/src/app/(platform)/library/agents/[id]/components/AgentRunsView/components/OutputRenderers/renderers/MarkdownRenderer.tsx`
- Specifically, update the logic in function `isVideoUrl()` on line 45.

**Methods/Imports/Definitions:**  
- Use the standard `URL` class (no need to add a new import, as this is available in browsers and in Node.js).
- Provide fallback in case the URL passed in is malformed (wrap in a try-catch, treat as non-video in this case).
- Check the parsed hostname for equality with `vimeo.com` or, optionally, specific allowed subdomains (`www.vimeo.com`).

---

_Suggested fixes powered by Copilot Autofix. Review carefully before merging._


AutoGPT-Agent:
Thank you for addressing this security vulnerability! The implementation looks solid - using the URL parser is the right approach to prevent incomplete URL substring sanitization vulnerabilities.

Just one note: the PR title should follow our conventional commit format, starting with a type like "fix:" - for example: "fix: Address code scanning alert #145 for URL substring sanitization"

Once you update the title, this PR will be ready to merge. Great work on improving our application security!

AutoGPT-Agent:
Thanks for submitting this PR to address the security vulnerability in URL validation!

The code changes look good - replacing substring matching with proper URL parsing is the right approach for this security fix.

One thing to note is that the PR title doesn't follow our conventional commit format. Could you please update it to start with a type prefix, such as:

```
fix(platform/frontend): Code scanning alert no. 145 - Incomplete URL substring sanitization
```

Once you update the title, this PR should be ready to merge. Nice job on the security improvement!

PR #11239, opened 2025-10-22
fix(backend): prevent error message leak

0ubbe:
Potential fix for [https://github.com/Significant-Gravitas/AutoGPT/security/code-scanning/155](https://github.com/Significant-Gravitas/AutoGPT/security/code-scanning/155)

To fix this vulnerability, avoid returning the raw exception message from SlugAlreadyInUseError in the API response. Instead, return a generic, non-revealing error message suitable for the end user. Log the original exception and its message on the server as a warning, so developers can still access the necessary diagnostic information. This change should be made solely in the exception handler for SlugAlreadyInUseError, replacing `{"detail": str(e)}` with a safe, constant message like `{"detail": "Slug already in use. Please choose a different slug."}`. This preserves the intended business logic (409 Conflict) and user feedback, but without leaking potentially sensitive data.

No new dependencies or imports are needed; we can keep the existing logging and response handling.

---

_Suggested fixes powered by Copilot Autofix. Review carefully before merging._

### DIFF
--- a/autogpt_platform/backend/backend/server/v2/store/routes.py
+++ b/autogpt_platform/backend/backend/server/v2/store/routes.py
@@ -551,7 +551,7 @@ async def create_submission(
         logger.warning("Slug already in use: %s", str(e))
         return fastapi.responses.JSONResponse(
             status_code=409,
-            content={"detail": str(e)},
+            content={"detail": "Slug already in use. Please choose a different slug."},
         )
     except Exception:
         logger.exception("Exception occurred whilst creating store submission")


AutoGPT-Agent:
Thanks for addressing this security vulnerability! The change looks good and properly implements the fix you described by replacing the raw exception message with a generic user-friendly message while still logging the actual error for developers.

The only adjustment needed is to update the PR title to follow our conventional commit format. It should be something like:

```
fix(backend): Code scanning alert #155 - Information exposure through exception
```

Otherwise, this is a clean, well-focused security fix that's ready to merge after the title update.

majdyz (approved):

PR #11255, opened 2025-10-27
fix: persist Agent Executor title after save/reload

### LINKED ISSUE #11041: Show agent name + version on Agent Executor block

aryash45:
- Save agent name + version into `hardcodedValues.agent_name`
- Display `agent_name` on load (fallback to live agent)
- Uses existing backend schema field
- Fixes #11041

### Changes
- Modified `CustomNode.tsx` to:
  - Save computed title (`agent.name vX.Y`) into `hardcodedValues.agent_name` when agent is selected
  - Display `agent_name` first on load, with fallback to live agent data
  - Use existing `agent_name` field in backend schema — **no backend changes required**
- Added `react-use` for `useDebounce` (already installed via `--legacy-peer-deps`)

### Why This Change Is Needed
When an **Agent Executor** block is added and an agent with a version is selected (e.g., `Researcher v2.1`), the title updates correctly in the UI.  
However, after **saving the project and reloading**, the title reverts to the generic `"Agent Executor"`.

This happens because the title is computed at render time but **never persisted** in the block’s saved data. This PR fixes it by using the **already-existing `agent_name` field** in the backend schema to store and restore the full title.

---

#### For code changes:
- [x] I have made a test plan
  
  > **Test Plan (for maintainers):**
  > - [ ] Add **Agent Executor** block to a new project
  > - [ ] Select an agent with a version (e.g., `Researcher v2.1`)
  > - [ ] Confirm block title shows `Researcher v2.1`
  > - [ ] **Save** the project
  > - [ ] **Refresh** the page
  > - [ ] Confirm title **still shows** `Researcher v2.1`
  > - [ ] (Optional) Edit title manually → save → refresh → should persist

---

**Note**: Could not run full local environment due to dependency conflicts (`tailwindcss` version mismatch).  
However, the fix uses **only existing schema fields** and follows the **exact pattern** used elsewhere in the codebase.

Ready for CI & maintainer testing!


AutoGPT-Agent:
Thanks for this detailed PR! The changes look good and the implementation approach is sound. A few observations:

1. Your PR includes some changes to `block_cost_config.py` that add new AI video block types and update costs, which seem unrelated to the Agent Executor title persistence issue. While these changes are within our acceptable limits (less than 20% of the PR), it would be better to separate unrelated changes into different PRs in the future.

2. The code looks good, but there are a couple of template string issues in the code. You're using single quotes with template syntax like `'${agent.name}'` instead of backticks. This would need to be fixed to work correctly.

3. I appreciate the detailed test plan even though you couldn't run it locally. This will help the maintainers verify your fix works correctly.

When the template string issues are fixed, this PR should be ready to merge!

aryash45:
hey @Pwuts please review and let me know if needed any changes

AutoGPT-Agent:
Thanks for this PR to fix the Agent Executor title persistence issue!

### Feedback

1. Your implementation approach looks good - using the existing `agent_name` field in the backend schema is a clean solution.

2. There appears to be a merge conflict in the `package.json` file - there are duplicate entries for `@chromatic-com/storybook` with different versions (4.1.1 and 4.1.2). This will need to be resolved before merging.

3. While I understand you couldn't test locally due to environment issues, it would be good to get confirmation from maintainers that they've tested according to your test plan before final approval.

Overall, this looks like a solid fix that addresses the issue in a clean way. Just need to resolve that package.json conflict, and we should be good to go!

autogpt_platform/frontend/package.json
```diff
@@ -115,6 +116,7 @@
     "zustand": "5.0.8"
   },
   "devDependencies": {
+    "@chromatic-com/storybook": "4.1.1",
```
Abhi1992002, line 119: Could you please remove this? We already have version `4.1.2`.

autogpt_platform/frontend/package.json
```diff
@@ -96,6 +96,7 @@
     "react-markdown": "9.0.3",
     "react-modal": "3.16.3",
     "react-shepherd": "6.1.9",
+    "react-use": "17.6.0",
```
Abhi1992002: We don’t need this. You’re installing it because you want to use `useDebounce`, but we already have `lodash` installed and it has `debounce` functionality. You can use that instead.

autogpt_platform/frontend/src/app/(platform)/build/components/legacy-builder/CustomNode/CustomNode.tsx
```diff
@@ -1047,6 +1106,6 @@ export const CustomNode = React.memo(
   },
   (prevProps, nextProps) => {
     // Only re-render if the 'data' prop has changed
-    return prevProps.data === nextProps.data;
+    return prevProps.data === nextProps.data && prevProps.id === nextProps.id;
```
Abhi1992002: I don’t think we need to check for ids. Sometimes, when we save the graph, node’s id changes, but the data remains the same. So, there’s no need to rerender the node. I think the data is sufficient.

autogpt_platform/frontend/src/app/(platform)/build/components/legacy-builder/CustomNode/CustomNode.tsx
```diff
@@ -466,6 +511,20 @@ export const CustomNode = React.memo(
...
+      const title = version ? `${agentName} v${version}` : agentName;
+      const newHardcodedValues = {
+        ...data.hardcodedValues,
+        agent_id: agent.id,
+        agent_name: agentName,  // Store just the name without version
+        agent_version: version,
+      }
+      setLocalAgent({id:agent.id, name:agentName, version:version});
+      setHardcodedValues(newHardcodedValues);
+    },[data.hardcodedValues, setHardcodedValues]);
```
Abhi1992002: Where are we using this? I think you forgot to use it in the `useEffect` hook.

autogpt_platform/frontend/src/app/(platform)/build/components/legacy-builder/CustomNode/CustomNode.tsx
```diff
@@ -216,6 +254,13 @@ export const CustomNode = React.memo(
           fillObjectDefaultsFromSchema(data.hardcodedValues, data.inputSchema),
         );
       }
+      if(data.uiType === BlockUIType.AGENT && data.hardcodedValues?.agent_id ){
+        setLocalAgent({
+          id:data.hardcodedValues.agent_id,
+          name: data.hardcodedValues.agent_name || "Agent Executor",
+          version: data.hardcodedValues.agent_version || "",
+        });
+      }
```
Abhi1992002: Why are we doing this again? Haven’t we already updated the localAgent in `updateAgentInBlock`? You also did it above. Could you please remove it?

autogpt_platform/frontend/src/app/(platform)/build/components/legacy-builder/CustomNode/CustomNode.tsx
```diff
@@ -135,6 +137,20 @@ export const CustomNode = React.memo(
...
+      ?{
+        id:data.hardcodedValues.agent_id,
+        name: data.hardcodedValues.agent_name || "Agent Executor",
+        version: data.hardcodedValues.agent_version || "",
+      }
+      : null;
+
+      if(agentFromInput && !localAgent) {
+        setLocalAgent(agentFromInput);
+      }
```
Abhi1992002: Could you please remove it? Since you’re calling it in a render, it might cause an issue. You can call it inside a `useEffect`.

autogpt_platform/frontend/src/app/(platform)/build/components/legacy-builder/CustomNode/CustomNode.tsx
```diff
@@ -195,10 +211,32 @@ export const CustomNode = React.memo(
...
+      ? {
+          id: data.hardcodedValues.agent_id,
+          name: data.hardcodedValues.agent_name || "Agent Executor",
+          version: data.hardcodedValues.agent_version || "",
+        }
+      : null;
+
+    if(agentFromInput && !localAgent) {
+      setLocalAgent(agentFromInput);
+    }
```
Abhi1992002: You can remove it

autogpt_platform/frontend/src/app/(platform)/build/components/legacy-builder/CustomNode/CustomNode.tsx
```diff
@@ -65,6 +65,7 @@ import {
 } from "@/components/atoms/Tooltip/BaseTooltip";
 import { InformationTooltip } from "@/components/molecules/InformationTooltip/InformationTooltip";
 import { Switch } from "@/components/atoms/Switch/Switch";
+import {useDebounce} from "react-use";
```
Abhi1992002: You can also remove it. It’s not used anywhere in the file, and if you want to use debounce, you can use loadash one instead.

Abhi1992002 (changes requested):
Thank you @aryash45 for contributing! 😊 I have requested some changes. I will approve it if you fix or answer those questions.

I have also tested this PR locally, and it’s not working. I think it’s because you are not persisting the agent details in hardcodedValues. I believe you forgot to call a function inside a side effect (I have already mentioned that). 

I think you can also do something like this — it would be easier:
- On initial render, check if metadata has a `customTitle` or not. If not, then save the agent name in metadata and then save it. Now you have persisted the agent name in the custom title, and everything will work fine. 

It’s a temporary solution, or the solution you have given is also temporary. For a permanent one, you can see why the title is changing on saving. Are we sending some wrong title to the backend when saving? [You need to go through the saving logic] 
Maybe you will find some issue.

autogpt_platform/frontend/src/app/(platform)/build/components/legacy-builder/CustomNode/CustomNode.tsx
Abhi1992002: That’s why it doesn’t work locally. You haven’t persisted agent details in hardcoded values.

aryash45:
@Abhi1992002 hi i know its been a long time can you please review

aryash45 (commented):
done

PR #11258, opened 2025-10-27
fix(frontend): Correctly rendering all types of outputs in the custom node in the new builder

Abhi1992002:
Currently, we are rendering text for all types of outputs, even if it’s a video, image, or other type. So, In current we fixed it by rendering them correctly. Also, some output actions weren’t working, so fixed them also.

Screenshot 2025-10-27 at 4 36 33 PM

Screenshot 2025-10-27 at 4 39 00 PM

Screenshot 2025-10-27 at 4 36 56 PM

Screenshot 2025-10-27 at 4 37 17 PM

### Changes 🏗️

- Add a new design for the node output.
- Render the correct HTML tag for each type.
- Make all the output actions below the data section workable, such as viewing the complete data or copying it.
- Add a “View more” button. We’re only seeing two pins of output. If we have more pins, we can view all the output in a dialogue box.

#### For code changes:
- [x] I have made a test plan
   - [x] able to render different types of output data correctly.
   - [x] All output actions are working perfectly.


autogpt_platform/frontend/src/app/(platform)/build/components/FlowEditor/nodes/CustomNode/components/NodeOutput/components/ViewMoreData.tsx
```diff
@@ -0,0 +1,147 @@
+import { Dialog } from "@/components/molecules/Dialog/Dialog";
+import { Button } from "@/components/atoms/Button/Button";
+import { Text } from "@/components/atoms/Text/Text";
+import { beautifyString, cn } from "@/lib/utils";
+import { ContentRenderer } from "./ContentRenderer";
+import { ScrollArea } from "@/components/__legacy__/ui/scroll-area";
+import { CheckIcon, CopyIcon } from "lucide-react";
```
0ubbe: Let's use phosphor icons 👍🏽

autogpt_platform/frontend/src/app/(platform)/build/components/FlowEditor/nodes/CustomNode/components/NodeOutput/components/ContentRenderer.tsx
```diff
@@ -0,0 +1,153 @@
+"use client";
+
+import { Text } from "@/components/atoms/Text/Text";
```
0ubbe: We have the `<OutputRenderers />` component already created by @ntindle for the library view; could we reuse that instead of creating new ones? 

I would like to have a single output rendering component that can be used on any of the pages 👍🏽 💆🏽

0ubbe (dismissed):
Good 🥇 , just a couple of questions

autogpt_platform/frontend/src/app/(platform)/build/components/FlowEditor/nodes/CustomNode/components/NodeOutput/components/ContentRenderer.tsx
Abhi1992002: @0ubbe, I’ve removed audio and video rendering from the Content Renderer and started using Nick’s output renderer. However, on custom nodes, if the value is a string, object, JSON, Markdown, or code, we need to render only two lines. This is because the size of the value could be huge, which is bad for the UI. We have a button at the bottom of each message that allows users to view the complete output.

PR #11270, opened 2025-10-28
fix(onboarding): Prevent frontend errors when less than 2 agents are returned

Swiftyos:
### Changes 🏗️

- Prevents the frontend from trying to display a partial selection of agents when less than 2 agents are returned from the backend.
- Adds a check in the backend to only return agents if we have at least 2.
- Adds a check in the frontend to ensure that the agent exists before attempting to update the state.
- Fixes [BUILDER-45E](https://sentry.io/organizations/significant-gravitas/issues/6953457243/).

#### For code changes:
- [x] I have made a test plan
  - [ ] Ensure that the frontend does not error when the backend returns fewer than 2 agents.
  - [ ] Ensure that the frontend correctly displays agents when the backend returns 2 or more agents.

### DIFF
--- a/autogpt_platform/backend/backend/data/onboarding.py
+++ b/autogpt_platform/backend/backend/data/onboarding.py
@@ -315,6 +315,11 @@ async def get_recommended_agents(user_id: str) -> list[StoreAgentDetails]:
     agent_points.sort(key=lambda x: x[1], reverse=True)
     recommended_agents = [agent for agent, _ in agent_points[:2]]
 
+    # Only return agents if we have at least 2 as frontend expects at least 2 agents
+    # This prevents the frontend from trying to display a partial selection
+    if len(recommended_agents) < 2:
+        return []
+
     return [
         StoreAgentDetails(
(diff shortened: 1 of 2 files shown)


AutoGPT-Agent:
Thank you for this PR that addresses the frontend errors when fewer than 2 agents are returned during onboarding.

### Feedback

- The code changes look appropriate - you've added checks both in the backend and frontend to handle the case when there aren't enough recommended agents.

- I notice that your test plan items are not all checked off. Before this PR can be merged, please complete the testing and check off all test items:
  - [ ] Ensure that the frontend does not error when the backend returns fewer than 2 agents.
  - [ ] Ensure that the frontend correctly displays agents when the backend returns 2 or more agents.

- It might be worth explicitly stating in your PR description how you tested these scenarios. Did you create a test case where the backend returns fewer than 2 agents? How did you verify the frontend behavior in both scenarios?

Once you've completed the testing and updated the PR description, this should be ready for another review.

PR #11282, opened 2025-10-29
fix(frontend): prevent Wallet rendering twice

0ubbe:
## Changes 🏗️

The `<Wallet />` was being rendered twice ( one hidden with CSS `hidden` ) because of the Navbar layout, which caused logic issues within the wallet. I changed to render it conditionally via Javascript instance, which is always better practice than use `hidden` specially for components with actual logic.

I also moved the component files closer to where it is used ( in the navbar ).

I have a Cursor plugin that removes imports when unused, but annoyingly re-organizes them, hence the changes around that...

### For code changes:
- [x] I have made a test plan
  - [x] Login
  - [x] There is only 1 Wallet in the DOM


AutoGPT-Agent:
Thanks for this PR to fix the Wallet component rendering issue!

The approach of using conditional rendering instead of CSS hiding makes sense and will help prevent the logic issues you mentioned. Moving the component closer to where it's used is also a good practice.

Before this can be merged, please remove the debugging `console.log` statement in the Wallet component:

```javascript
console.log("credits", credits, "state", state);
```

Also, I noticed you've moved several imports into a more logical order in various files. While this is fine, it would be helpful to mention these organizational changes in the PR description as well.

Otherwise, the implementation looks good. The conditional rendering based on screen size and login state is much cleaner than hiding with CSS.

Issue #11273, opened 2025-10-28
Scheduled jobs get stuck in DST transition window


Pwuts:
Last Sunday, 4 graph execution jobs that were scheduled on `Europe/` time zones landed in the European DST transition window and got stuck because of it. This also blocked the scheduler from responding to RPCs, breaking all our scheduling-related functionality.

This Sunday, USA has its DST transition. We need to make sure this doesn't happen again then or later.

I have traced this to [agronholm/apscheduler#1021](https://github.com/agronholm/apscheduler/issues/1021), so we need to either upgrade to fixed (pre-release) version `v4.0.0a6` or contribute a fix for `v3.11.0` of `apscheduler`.

Pwuts:
I figured out a way to unstuck our scheduler manually in case of emergency:

1. Query all jobs (from the store) where `next_run_time` is the DST transition timestamp.
2. For each entry, run this logic to get an updated `next_run_time` and `job_state` if necessary:

```py
import pickle
from datetime import timedelta

job_state_pickled = "0192837465afbecd..."  # remove the \x prefix if any

job_state = pickle.loads(bytes.fromhex(job_state_pickled))

if job_state["next_run_time"].fold:
    job_state["next_run_time"] += timedelta(hours=1)

    print(f"📋 Updated next_run_time: {int(job_state['next_run_time'].timestamp())}")

    job_state_pickled = pickle.dumps(job_state).hex()
    print(f"📋 Updated job_state: {job_state_pickled}")
else:
    print("✅ Not affected, no update needed")
```

3. Update these values in the job store

Pwuts:
Status: hoping for a swift resolution here [https://github.com/agronholm/apscheduler/pull/1079](https://github.com/agronholm/apscheduler/pull/1079)

PR #11292, opened 2025-10-31
fix(backend/scheduler): Use DST-fixed APScheduler version

Pwuts:
- #11273

### Changes 🏗️

Point to a fork of APScheduler that contains a fix for #11273 ([diff](https://github.com/berianjames/apscheduler/compare/3.11.0...ff9a687424a23c81a72deb7331a69cc1351b0420)).

This is supposed to be a TEMPFIX until APScheduler releases an official fix.

#### For code changes:
- [x] I have made a test plan
  - [x] "It's a rather ugly solution but the test proves that it works." ~the maintainer
  - [ ] CI passes


Pwuts:
CI fails and fixing it would be more work. Meanwhile the maintainer has committed to releasing a patch today, so I'll wait for that instead.

PR #11318, opened 2025-11-04
hotfix(backend): Temporarily disable library existence check for graph execution

majdyz:
`add_store_agent_to_library` does not add subagents to the user library, so the `validate_graph_execution_permissions` check can cause issues.

### Changes 🏗️

- Disable `validate_graph_execution_permissions` check in `validate_and_construct_node_execution_input`

#### For code changes:
- [x] I have made a test plan
  - Trivial change, no test needed

### DIFF
--- a/autogpt_platform/backend/backend/executor/utils.py
+++ b/autogpt_platform/backend/backend/executor/utils.py
@@ -519,11 +519,15 @@ async def validate_and_construct_node_execution_input(
     # raising specific exceptions for appropriate error handling
     # Note: Version-agnostic check to allow execution of graphs that reference
     # older versions of sub-graphs that may no longer be in the library
-    await gdb.validate_graph_execution_permissions(
-        graph_id=graph_id,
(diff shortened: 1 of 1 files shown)


AutoGPT-Agent:
Thank you for submitting this hotfix. The PR has some issues that need to be addressed before it can be merged:

1. Please complete the checklist in the PR description. All checkboxes need to be checked off, particularly for code changes.

2. Add a more detailed explanation of:
   - What specific issues this check was causing
   - What the security implications are of disabling this validation
   - Whether this is a temporary fix (as suggested by "temporarily" in the title) and if so, what the plan is to properly fix this

3. Include a test plan describing how you've verified that disabling this check resolves the issue without introducing other problems.

4. Since you're disabling a permission check, please explain how you've verified that this won't create security vulnerabilities or unauthorized access to graphs.

Once these items are addressed, we can proceed with the review process.

PR #11330, opened 2025-11-06
feature(backend): Limit Chat to Auth Users, Limit Agent Runs Per Chat

Swiftyos:
This PR removes the feature flag for the chat system but requires users be logged in to use it.
It also limits the number of runs for a single agent during a conversation to guard against strange behaviour.

### Changes 🏗️

- fix convo flow to work with openai
- remove chat feature flag on backend
- require auth for backend
- limit to 3 runs per agent per conversation

#### For code changes:
- [x] I have made a test plan
  
  - [x] ran through chat flow to make sure it could not do more than 3 runs
  - [x] made sure access is limited to authenticated users

### DIFF
--- a/autogpt_platform/backend/backend/server/v2/chat/service.py
+++ b/autogpt_platform/backend/backend/server/v2/chat/service.py
@@ -289,9 +289,6 @@ async def stream_chat_completion(
 
     session.messages.extend(messages_to_save)
     await upsert_chat_session(session)
-    import pprint
-
-    pprint.pprint(session)
 
     # If we did a tool call, stream the chat completion again to get the next response
     if has_done_tool_call:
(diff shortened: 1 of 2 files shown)


autogpt_platform/backend/backend/server/v2/chat/service.py
```diff
@@ -246,12 +267,31 @@ async def stream_chat_completion(
...
+
+    # Add tool response messages after assistant message
+    messages_to_save.extend(tool_response_messages)
+    logger.info(f"Saving {len(tool_response_messages)} tool response messages")
+
+    session.messages.extend(messages_to_save)
     await upsert_chat_session(session)
+    import pprint
+
+    pprint.pprint(session)
```
kcze: Where and why is this printing here? Why not logger?

Swiftyos: removed it, thank you

PR #11331, opened 2025-11-06
feat(backend): Remove deprecated LLM models and add migration script

Bentlybro:
These models have become deprecated
- deepseek-r1-distill-llama-70b
- gemma2-9b-it
- llama3-70b-8192
- llama3-8b-8192
- google/gemini-flash-1.5

I have removed them and setup a migration, the migration is to convert all the old versions of the model to new versions, the model changes will happen like so

- llama3-70b-8192 → llama-3.3-70b-versatile
- llama3-8b-8192 → llama-3.1-8b-instant
- google/gemini-flash-1.5 → google/gemini-2.5-flash
- deepseek-r1-distill-llama-70b → gpt-5-chat-latest
- gemma2-9b-it → gpt-5-chat-latest 

#### For code changes:
- [x] I have made a test plan
  
  - [x] Check to see if old models where removed
  - [x] Check to see if migration worked and converted old models to new one in graph

### DIFF
--- a/autogpt_platform/backend/backend/server/rest_api.py
+++ b/autogpt_platform/backend/backend/server/rest_api.py
@@ -50,12 +50,7 @@ from backend.util.exceptions import (
     NotAuthorizedError,
     NotFoundError,
 )
-from backend.util.feature_flag import (
-    Flag,
-    create_feature_flag_dependency,
-    initialize_launchdarkly,
-    shutdown_launchdarkly,
-)
+from backend.util.feature_flag import initialize_launchdarkly, shutdown_launchdarkly
 from backend.util.service import UnhealthyServiceError
 
 settings = backend.util.settings.Settings()
@@ -294,9 +289,6 @@ app.include_router(
     chat_routes.router,
     tags=["v2", "chat"],
     prefix="/api/chat",
-    dependencies=[
-        fastapi.Depends(create_feature_flag_dependency(Flag.CHAT, default=False))
-    ],
 )
 
 app.mount("/external-api", external_app)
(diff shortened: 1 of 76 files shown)


AutoGPT-Agent:
Thanks for working on removing these deprecated LLM models and providing a migration path. The changes look good, but I have one issue to point out:

**Title Format Issue**

The PR title `feat(llm's): Remove old deprecated LLM's from the platform` doesn't follow our conventional commit format. It should use one of our standard scopes instead of `llm's`. Since this change affects multiple areas but primarily deals with model definitions in the blocks system, a more appropriate title would be something like:

```
feat(blocks): Remove deprecated LLM models and add migration script
```

or if you'd prefer to emphasize the backend changes:

```
feat(backend): Remove deprecated LLM models and add migration script
```

Could you please update the PR title to follow our conventional commit format? Once that's fixed, this PR should be good to merge.

AutoGPT-Agent:
Thanks for your work on removing these deprecated LLM models! A few things to adjust:

1. The PR title should use one of the approved scopes. Since you're modifying code in backend/blocks and adding a migration, "feat(platform/blocks): Remove old deprecated LLM's from the platform" or "feat(platform/backend): Remove old deprecated LLM's from the platform" would be more appropriate.

2. The PR description template section for "Changes 🏗️" was left with the HTML comment instead of being replaced with a description of the changes. While your opening paragraph does describe the changes well, please move that content under the Changes section or add a brief summary there.

Otherwise, this looks good! The migration script is well-commented and the changes are focused on the task of removing deprecated models.

PR #11340, opened 2025-11-08
fix(frontend): ensure all agents with graph_id appear in activity dropdown

### LINKED ISSUE #11229: Fix "Unknown Agent" in Agent Activity Dropdown
**Issue**: There is a problem with the agent activity dropdown, where agents are not consistently showing up.

**History**: Initially, this issue occurred rarely, attributed to agents being imported the old (hidden) way, which was expected to phase out over time.

**Current Status**: The issue has worsened, resulting in very few agents displaying correctly in the dropdown.      What steps can we take to ensure that a gents consistently show up, establishing reliable visibility for the agents executing under my account?

[20251021-1249-03.6093505.mp4](https://uploads.linear.app/a47946b5-12cd-4b3d-8822-df04c855879f/5f96eb61-f266-456f-a009-8978f5980d03/fd3ebb33-5ca3-4db3-9860-133fb523e797?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiL2E0Nzk0NmI1LTEyY2QtNGIzZC04ODIyLWRmMDRjODU1ODc5Zi81Zjk2ZWI2MS1mMjY2LTQ1NmYtYTAwOS04OTc4ZjU5ODBkMDMvZmQzZWJiMzMtNWNhMy00ZGIzLTk4NjAtMTMzZmI1MjNlNzk3IiwiaWF0IjoxNzYxMDUxNDUxLCJleHAiOjMzMzMxNjExNDUxfQ.ttBnBEUPpFU0BSstozuRCiwBCokC1VNUy3wN7I1-VTU)

yashwantbezawada:
## Summary
Fixes issue where agents appear as "Unknown Agent" in the agent activity dropdown.

## Root Cause
The `buildAgentInfoMap` function was only adding agents to the map if they had BOTH `graph_id` AND `id` fields populated. Since `graph_id` is used as the lookup key, agents missing the `id` field would not be added to the map, causing them to appear as "Unknown Agent".

## Changes
Changed the condition in `buildAgentInfoMap` from:
```typescript
if (a.graph_id && a.id) {
```

To:
```typescript
if (a.graph_id) {
```

This ensures any agent with a `graph_id` (the lookup key) will be added to the map. The `library_agent_id` field remains optional.

## Impact
- All agents with valid `graph_id` will now appear consistently in the activity dropdown
- Resolves the "Unknown Agent" issue mentioned in the bug report
- No breaking changes - `library_agent_id` is still populated when available

## Testing
Manual testing recommended: Verify agents with missing `id` fields now appear with their proper names instead of "Unknown Agent".

Fixes #11229

### DIFF
--- a/autogpt_platform/backend/backend/server/rest_api.py
+++ b/autogpt_platform/backend/backend/server/rest_api.py
@@ -43,6 +43,7 @@ from backend.integrations.providers import ProviderName
 from backend.monitoring.instrumentation import instrument_fastapi
 from backend.server.external.api import external_app
 from backend.server.middleware.security import SecurityHeadersMiddleware
+from backend.server.utils.cors import build_cors_params
 from backend.util import json
 from backend.util.cloud_storage import shutdown_cloud_storage_handler
 from backend.util.exceptions import (
@@ -303,9 +304,14 @@ async def health():
 
 class AgentServer(backend.util.service.AppProcess):
     def run(self):
+        cors_params = build_cors_params(
+            settings.config.backend_cors_allow_origins,
+            settings.config.app_env,
+        )
+
         server_app = starlette.middleware.cors.CORSMiddleware(
             app=app,
-            allow_origins=settings.config.backend_cors_allow_origins,
+            **cors_params,
             allow_credentials=True,
             allow_methods=["*"],  # Allows all methods
             allow_headers=["*"],  # Allows all headers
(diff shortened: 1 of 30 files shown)


yashwantbezawada:
Thanks for the explanation! That makes sense - I misdiagnosed the root cause. The `LibraryAgent` type does require both fields, so the real issue is executions referencing graphs that aren't in the library (sub-agents, deleted agents, etc).

Closing this one. The Sentry logging idea you mentioned could help track how often this happens.

PR #11363, opened 2025-11-11
fix(blocks): Make Smart Decision Maker tool pin handling consistent and reliable

### LINKED ISSUE #11345: Smart Decision Maker fails to use tool when drag-dropped from block menu
There are a few ways to add a node to a graph, one of which is dragging it from the block menu into the graph. If this is done with a Smart Decision Maker block, the block doesn't work. This is because with this drag-to-add mechanism, the `.data.title` of the node is set to a value which causes the tool pin generation to fail, outputting e.g. `tool_^_ai_~_prompt` instead of `tool_^_aitextgenerator_~_prompt`.

Long-term fix would be to move tool pin resolution to the backend.

another scenario of the issue:

The tool link for SubAgent is using the agent name (title), and unfortunately the name data can sometimes unobtainable due to the agent is not being fetched (when the entry is already way past the pagination limit, and we only fetch the first few last agents).<br>

In this specific case, the source name ended up being `tools_^_agentexecutorblock_~_question`instead of `tools_^_web_search_~_question` (the agent name is Web Search).

Pwuts:
- Resolves #11345

### Changes 🏗️

- Move tool use routing logic from frontend to backend: routing info was being baked into graph links by the frontend, inconsistently, causing issues
- Rework tool use routing to use target node ID instead of target block name
- Add a bit of magic to `NodeOutputs` component to show tool node title instead of ID

DX:
- Removed `build` from `.prettierignore` -> re-enable formatting for builder components

#### For code changes:
- [x] I have made a test plan
  - [x] Use SDM block in a graph; verify it works
  - [x] Use SDM block with agent executor block as tool; verify it works
  - Tests for `parse_execution_output` pass (checked by CI)

### DIFF
--- a/autogpt_platform/backend/backend/blocks/test/test_smart_decision_maker.py
+++ b/autogpt_platform/backend/backend/blocks/test/test_smart_decision_maker.py
@@ -293,6 +293,7 @@ async def test_smart_decision_maker_parameter_validation():
                     },
                     "required": ["query", "max_keyword_difficulty"],
                 },
+                "_sink_node_id": "test-sink-node-id",
             },
         }
     ]
@@ -450,13 +451,13 @@ async def test_smart_decision_maker_parameter_validation():
             outputs[output_name] = output_data
 
         # Verify tool outputs were generated correctly
-        assert "tools_^_search_keywords_~_query" in outputs
-        assert outputs["tools_^_search_keywords_~_query"] == "test"
-        assert "tools_^_search_keywords_~_max_keyword_difficulty" in outputs
-        assert outputs["tools_^_search_keywords_~_max_keyword_difficulty"] == 50
+        assert "tools_^_test-sink-node-id_~_query" in outputs
+        assert outputs["tools_^_test-sink-node-id_~_query"] == "test"
+        assert "tools_^_test-sink-node-id_~_max_keyword_difficulty" in outputs
+        assert outputs["tools_^_test-sink-node-id_~_max_keyword_difficulty"] == 50
         # Optional parameter should be None when not provided
-        assert "tools_^_search_keywords_~_optional_param" in outputs
-        assert outputs["tools_^_search_keywords_~_optional_param"] is None
+        assert "tools_^_test-sink-node-id_~_optional_param" in outputs
+        assert outputs["tools_^_test-sink-node-id_~_optional_param"] is None
 
     # Test case 4: Valid tool call with ALL parameters (should succeed)
     mock_tool_call_all_params = MagicMock()
@@ -504,9 +505,9 @@ async def test_smart_decision_maker_parameter_validation():
             outputs[output_name] = output_data
 
         # Verify all tool outputs were generated correctly
-        assert outputs["tools_^_search_keywords_~_query"] == "test"
-        assert outputs["tools_^_search_keywords_~_max_keyword_difficulty"] == 50
-        assert outputs["tools_^_search_keywords_~_optional_param"] == "custom_value"
+        assert outputs["tools_^_test-sink-node-id_~_query"] == "test"
+        assert outputs["tools_^_test-sink-node-id_~_max_keyword_difficulty"] == 50
+        assert outputs["tools_^_test-sink-node-id_~_optional_param"] == "custom_value"
 
 
 @pytest.mark.asyncio
@@ -530,6 +531,7 @@ async def test_smart_decision_maker_raw_response_conversion():
                     "properties": {"param": {"type": "string"}},
                     "required": ["param"],
                 },
+                "_sink_node_id": "test-sink-node-id",
             },
         }
     ]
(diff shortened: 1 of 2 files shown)


AutoGPT-Agent:
Thank you for your contribution to improve the Smart Decision Maker tool pin handling!

Your PR looks good overall, with clear code changes that move the tool use routing logic from frontend to backend and make it more consistent. The implementation approach seems sound, using node IDs instead of names for routing.

Before this can be merged, please complete the remaining test item in your checklist:
- [ ] Use SDM block with agent executor block as tool; verify it works

Once you've verified this scenario works correctly, please check this item and the PR should be ready for review.

AutoGPT-Agent:
Thanks for this PR that improves the Smart Decision Maker tool pin handling. The changes look good overall, moving the routing logic from frontend to backend should help with consistency.

One item in your test plan is not checked:
- [ ] Use SDM block with agent executor block as tool; verify it works

Could you please complete this test and confirm it works before we merge?

Also, I noticed you've added several test cases for the `parse_execution_output` function which is great, but make sure you've also tested the real-world scenarios with the agent executor as described in your unchecked test item.

The approach of using node IDs instead of names for routing looks more robust, and the added helper to display friendly names in the UI is a nice touch for usability.

autogpt_platform/frontend/src/app/(platform)/build/components/legacy-builder/NodeOutputs.tsx
```diff
@@ -17,13 +21,40 @@ export default function NodeOutputs({
...
+    }
+    return null;
+  };
+  const getBeautifiedPinName = (pin: string) => {
+    if (!pin.startsWith("tools_^_")) {
+      return beautifyString(pin);
+    }
+    // Special handling for tool pins: replace node ID with node title
+    const toolNodeID = pin.slice(8).split("_~_")[0]; // tools_^_{node_id}_~_{field}
+    const toolNodeTitle = getNodeTitle(toolNodeID);
```
majdyz, line 46: if this end up causing rendering issue because of the heavy processing, we can always opt in with just removing the id part for now.

autogpt_platform/frontend/src/app/(platform)/build/components/legacy-builder/Flow/Flow.tsx
```diff
@@ -828,11 +850,20 @@ const FlowEditor: React.FC<{
...
+        getNextNodeId,
+      }}
     >
       <div className={className}>
         <ReactFlow
-          nodes={nodes}
+          nodes={nodes.map((node) => ({
+            ...node,
+            position: immediateNodePositions[node.id] || node.position,
+          }))}
```
Pwuts: Somehow these frontend changes cause the entire `FlowEditor` to re-render continuously while dragging nodes.

autogpt_platform/backend/backend/blocks/smart_decision_maker.py
```diff
@@ -699,22 +704,25 @@ async def run(
...
+            )
 
             for clean_arg_name in expected_args:
                 # arg_name is now always the cleaned field name (for Anthropic API compliance)
                 # Get the original field name from field mapping for proper emit key generation
                 original_field_name = field_mapping.get(clean_arg_name, clean_arg_name)
                 arg_value = tool_args.get(clean_arg_name)
 
-                sanitized_tool_name = self.cleanup(tool_name)
+                tool_key = sink_node_id if sink_node_id else self.cleanup(tool_name)
```
majdyz: do the fallback even useful here? should we just raise error when the _sink_node_id is not defined?

autogpt_platform/backend/backend/data/dynamic_fields.py
```diff
@@ -137,30 +149,64 @@ def _tokenise(path: str) -> list[tuple[str, str]] | None:
...
     Returns:
-        The value at the specified path, or None if not found/invalid
+        The value at the specified path, or `None` if not found/invalid.
     """
-    base_name, data = output
+    output_pin_name, data = output_item
+
+    # Special handling for tool pins
+    if is_tool_pin(link_output_selector) and (  # "tools" or "tools_^_…"
+        output_pin_name.startswith("tools_^_") and "_~_" in output_pin_name
```
majdyz, line 187: use is_tool_pin ?

majdyz (approved):

autogpt_platform/frontend/src/app/(platform)/build/components/legacy-builder/Flow/Flow.tsx
Pwuts: fixed

autogpt_platform/frontend/src/app/(platform)/build/components/legacy-builder/NodeOutputs.tsx
Pwuts: I was almost going to, but figured it out :)

autogpt_platform/backend/backend/data/dynamic_fields.py
Pwuts: `is_tool_pin` is less specific than this check

PR #11366, opened 2025-11-12
refactor(backend): Improve Block Error Handling

Swiftyos:
We need a way to differentiate between serious errors that cause on call alerts and block errors.
This PR address this need by ensuring all errors that occur during execution of a block are of Subtype BlockError

### Changes 🏗️

- Introduced BlockErrors and its subtypes
- Updated current errors that are emitted by blocks to use BlockError
- Update executor manager, to errors emitted when running a block that are not of type BlockError to BlockUnknownError

#### For code changes:
- [x] I have made a test plan
  
  - [x] checked tests still work
  - [x] Ensured block error message is readable and useful

### DIFF
--- a/autogpt_platform/backend/backend/data/block.py
+++ b/autogpt_platform/backend/backend/data/block.py
@@ -558,7 +558,7 @@ class Block(ABC, Generic[BlockSchemaInputType, BlockSchemaOutputType]):
                     message=str(ex),
                     block_name=self.name,
                     block_id=self.id,
-                )
+                ) from ex
             else:
                 raise ex
 


AutoGPT-Agent:
Thank you for this PR improving block error handling! The implementation looks solid, creating a proper error hierarchy to differentiate between system errors and block execution errors.

The changes look good overall, with just one small issue to address:

- The PR title uses `chore(backend)` which doesn't match our conventional commit format. Please update it to use one of our approved types: feat, fix, refactor, ci, or dx. In this case, `fix(backend)` or `refactor(backend)` would be more appropriate.

Once you've updated the title, this PR should be ready to merge. Great work on making our error handling more robust!

autogpt_platform/backend/backend/server/rest_api.py
```diff
@@ -251,6 +252,7 @@ async def validation_error_handler(
 app.add_exception_handler(pydantic.ValidationError, validation_error_handler)
 app.add_exception_handler(MissingConfigError, handle_internal_http_error(503))
 app.add_exception_handler(ValueError, handle_internal_http_error(400))
+app.add_exception_handler(BlockError, handle_internal_http_error(400))
```
Swiftyos: @majdyz I'm not sure we want to do this. What do you think?

AutoGPT-Agent:
Thank you for this PR that improves block error handling! The implementation looks good and the changes are well-focused on the stated goal.

One issue to fix:

- The PR title needs to use one of the approved conventional commit types. Instead of 'chore(backend)', please use one of: feat, fix, refactor, ci, or dx. Since you're improving error handling, 'fix(backend)' or 'refactor(backend)' would be more appropriate.

Otherwise, this looks ready to merge once the title is updated. The changes appropriately differentiate between serious errors and block errors, which should help with operational monitoring.

autogpt_platform/backend/backend/data/block.py
```diff
@@ -543,20 +544,28 @@ def get_info(self) -> BlockInfo:
 
     async def execute(self, input_data: BlockInput, **kwargs) -> BlockOutput:
         if error := self.input_schema.validate_data(input_data):
```
majdyz, line 566: I think we need to move this to `_execute` then wrap it in try except for any other exception that is not a blockerror re-raise it with BlockUnknownError

autogpt_platform/backend/backend/executor/manager.py
```diff
@@ -252,12 +257,18 @@ async def execute_node(
             output_size += len(json.dumps(output_data))
             log_metadata.debug("Node produced output", **{output_name: output_data})
             yield output_name, output_data
-    except Exception:
+
+    except Exception as ex:
```
majdyz: if we do above we don't need this change and the change is self-contained

autogpt_platform/backend/backend/server/rest_api.py
majdyz: We don't need this, we can make BlockError that are known to extend valueerror

autogpt_platform/backend/backend/util/exceptions.py
```diff
@@ -1,6 +1,28 @@
...
+        super().__init__(message)
+        self.message = message
+        self.block_name = block_name
+        self.block_id = block_id
+
+    def __str__(self):
+        return f"raised by {self.block_name} with message: {self.message}. block_id: {self.block_id}"
+
+
+class BlockValueError(BlockError): ...
```
majdyz: This needs to extend ValueError too
can we just call this BlockInputError then ?

autogpt_platform/backend/backend/util/exceptions.py
```diff
@@ -1,6 +1,28 @@
...
+
+
+class BlockRuntimeError(BlockError): ...
```
majdyz: This also needs to extend ValueError, see: https://github.com/Significant-Gravitas/AutoGPT/blob/dev/autogpt_platform/backend/backend/executor/manager.py#L579-L596

This can be BlockExecutionError since RuntimeError is another real python error and can be misleading

autogpt_platform/backend/backend/util/exceptions.py
```diff
@@ -1,6 +1,28 @@
...
+
+
+class BlockUnknownError(BlockError): ...
```
majdyz: also why the 3 dots ?

autogpt_platform/backend/backend/util/exceptions.py
```diff
@@ -106,3 +128,11 @@ class RedisError(Exception):
     """Raised when there is an error interacting with Redis"""
 
     pass
+
+
+if __name__ == "__main__":
```
majdyz: why do we need this ?

autogpt_platform/backend/backend/util/exceptions.py
Swiftyos: Done, and I've made a new error BlockOutputError for the output specific error

autogpt_platform/backend/backend/util/exceptions.py
Swiftyos: Done

autogpt_platform/backend/backend/util/exceptions.py
Swiftyos: it means stub class / functions. Otherwise you have to add a doc string at least for it to not error

autogpt_platform/backend/backend/util/exceptions.py
Swiftyos: Removed

autogpt_platform/backend/backend/data/block.py
Swiftyos: done

autogpt_platform/backend/backend/executor/manager.py
Swiftyos: removed these changes

autogpt_platform/backend/backend/data/block.py
```diff
@@ -542,21 +549,43 @@ def get_info(self) -> BlockInfo:
         )
 
     async def execute(self, input_data: BlockInput, **kwargs) -> BlockOutput:
+        try:
+            async for output_name, output_data in self._execute(input_data, **kwargs):
+                yield output_name, output_data
+        except Exception as ex:
+            if not isinstance(ex, BlockError):
+                raise BlockUnknownError(
```
majdyz, line 557: Can we add `from ex` we need the trail of the exception

majdyz (dismissed):

autogpt_platform/backend/backend/server/rest_api.py
Swiftyos: Removed

PR #11372, opened 2025-11-13
fix(platform): Prevent Error On Empty Store Agent Search

Swiftyos:
Currently when no agents are found for a given agent a Database Error is raised

## Changes 🏗️

- When no agents are found return a StoreAgentResponse with no agents in it 

### For code changes:
- [x] I have made a test plan
  - [x] Empty search results still work as expected
  - [x] End to End test in local env


autogpt_platform/backend/backend/server/v2/store/db.py
```diff
@@ -141,6 +141,17 @@ async def get_store_agents(
             agents = await prisma.client.get_client().query_raw(
```
Pwuts, line 141: The error this PR aims to fix is raised here

Pwuts: Further digging reveals this issue lies deeper than our Python or SQL code

autogpt_platform/backend/backend/server/v2/store/db.py
```diff
@@ -141,6 +141,17 @@ async def get_store_agents(
...
                 typing.cast(typing.LiteralString, sql_query), *params
             )
+            if not agents:
```
majdyz, line 144: can't we just add `or []` if this is the actual issue?

Swiftyos:
The actual issue was caused by stale session connection pool connections this PR switches to using direct connections to avoid this issue https://github.com/Significant-Gravitas/AutoGPT/pull/11377

PR #11374, opened 2025-11-13
fix(platform): Remove un-runnable agents from schedule

Swiftyos:
Currently when an agent fails validation during a scheduled run, we raise an error then try again, regardless of why.

This change removed the agent schedule and notifies the user 

### Changes 🏗️

- add schedule_id to the GraphExecutionJobArgs
- add agent_name to the GraphExecutionJobArgs
- Delete schedule on GraphValidationError
- Notify the user with a message that include the agent name

#### For code changes:
- [x] I have made a test plan
  
  - [x] I have ensured the scheduler tests work with these changes

### DIFF
--- a/autogpt_platform/frontend/src/app/(platform)/login/page.tsx
+++ b/autogpt_platform/frontend/src/app/(platform)/login/page.tsx
@@ -11,6 +11,7 @@ import Turnstile from "@/components/auth/Turnstile";
 import { environment } from "@/services/environment";
 import { LoadingLogin } from "./components/LoadingLogin";
 import { useLoginPage } from "./useLoginPage";
+import { MobileWarningBanner } from "@/components/auth/MobileWarningBanner";
 
 export default function LoginPage() {
   const {
@@ -127,6 +128,7 @@ export default function LoginPage() {
           link={{ text: "Sign up", href: "/signup" }}
         />
       </AuthCard>
+      <MobileWarningBanner />
       <EmailNotAllowedModal
         isOpen={showNotAllowedModal}
         onClose={handleCloseNotAllowedModal}
--- a/autogpt_platform/frontend/src/app/(platform)/signup/page.tsx
+++ b/autogpt_platform/frontend/src/app/(platform)/signup/page.tsx
@@ -21,6 +21,7 @@ import { environment } from "@/services/environment";
 import { WarningOctagonIcon } from "@phosphor-icons/react/dist/ssr";
 import { LoadingSignup } from "./components/LoadingSignup";
 import { useSignupPage } from "./useSignupPage";
+import { MobileWarningBanner } from "@/components/auth/MobileWarningBanner";
 
 export default function SignupPage() {
   const {
@@ -205,6 +206,7 @@ export default function SignupPage() {
           link={{ text: "Log in", href: "/login" }}
         />
       </AuthCard>
+      <MobileWarningBanner />
       <EmailNotAllowedModal
         isOpen={showNotAllowedModal}
         onClose={handleCloseNotAllowedModal}
--- a/autogpt_platform/frontend/src/components/auth/MobileWarningBanner.tsx
+++ b/autogpt_platform/frontend/src/components/auth/MobileWarningBanner.tsx
@@ -1,0 +1,31 @@
+"use client";
+
+import { useBreakpoint } from "@/lib/hooks/useBreakpoint";
+import { DeviceMobile } from "@phosphor-icons/react";
+import { Text } from "../atoms/Text/Text";
+
+export function MobileWarningBanner() {
+  const breakpoint = useBreakpoint();
+  const isMobile = breakpoint === "base" || breakpoint === "sm";
+
+  if (!isMobile) {
+    return null;
+  }
+
+  return (
+    <div className="mx-auto mt-6 w-full max-w-[32rem] rounded-lg border border-amber-200 bg-amber-50 p-4">
+      <div className="flex items-start gap-3">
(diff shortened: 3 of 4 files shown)


autogpt_platform/backend/backend/executor/scheduler.py
```diff
@@ -484,11 +527,13 @@ def add_graph_execution_schedule(
...
             jobstore=Jobstores.EXECUTION.value,
             replace_existing=True,
+            id=schedule_id,
         )
         logger.info(
             f"Added job {job.id} with cron schedule '{cron}' in timezone {user_timezone}, input data: {input_data}"
         )
-        return GraphExecutionJobInfo.from_db(job_args, job)
+        job_info = GraphExecutionJobInfo.from_db(job_args, job)
+        return job_info
```
majdyz: What is it for?

autogpt_platform/backend/backend/executor/scheduler.py
```diff
@@ -168,6 +188,24 @@ async def _execute_graph(**kwargs):
...
+        )
+        if args.schedule_id:
+            scheduler_client = get_scheduler_client()
+            scheduler_client.delete_schedule(
+                schedule_id=args.schedule_id,
+                user_id=args.user_id,
+            )
+            await _send_schedule_event_notification(
+                user_id=args.user_id,
+                message=f"Scheduled {args.agent_name} could not be ran and has been unscheduled. Please setup the schedule again.",
```
majdyz: Please fix and schedule again, or maybe if we can prompt them to run manually first.

autogpt_platform/backend/backend/executor/scheduler.py
```diff
@@ -168,6 +188,24 @@ async def _execute_graph(**kwargs):
...
+        logger.error(
+            f"Scheduled Graph {args.graph_id} failed validation. Unscheduling graph"
+        )
+        if args.schedule_id:
+            scheduler_client = get_scheduler_client()
+            scheduler_client.delete_schedule(
+                schedule_id=args.schedule_id,
+                user_id=args.user_id,
+            )
+            await _send_schedule_event_notification(
```
majdyz: Is this an email or what? I thought we haven't introduced the mailing system on prod

autogpt_platform/backend/backend/executor/scheduler.py
```diff
@@ -168,6 +188,24 @@ async def _execute_graph(**kwargs):
         )
         # Clean up orphaned schedules for this graph
         await _cleanup_orphaned_schedules_for_graph(args.graph_id, args.user_id)
+    except GraphValidationError:
+        logger.error(
+            f"Scheduled Graph {args.graph_id} failed validation. Unscheduling graph"
+        )
```
majdyz: For the old schedule that doesn't have the IDs are we not going to notify? We can tell the name of it or something.

autogpt_platform/backend/backend/executor/scheduler.py
Swiftyos: I was debugging something, I will revert this change

autogpt_platform/backend/backend/executor/scheduler.py
Swiftyos: I think this causes a notification in the notification queue at the top right of the screen. But I do need to dig into this to check it does what I think it does

autogpt_platform/backend/backend/executor/scheduler.py
Swiftyos: We won't have the name available for the old schedules. But I could add a db call to lookup the agent_name. 

Do you think its worth adding a db call to support the old schedule case?

autogpt_platform/backend/backend/executor/scheduler.py
Swiftyos: I've removed the notification as it doesn't actually trigger anything in the frontend and currently we have no way of persisting notifications

autogpt_platform/backend/backend/executor/scheduler.py
Swiftyos: It would only show an ui element if the user is on the platform at the time. I've removed it pending updating the notification system

autogpt_platform/backend/backend/executor/scheduler.py
majdyz: no, as long as we log something graph_id is enough, don't we have user_id too? can we add that?

autogpt_platform/backend/backend/executor/scheduler.py
```diff
@@ -160,14 +163,24 @@ async def _execute_graph(**kwargs):
...
-            f"Scheduled execution blocked for deleted/archived graph {args.graph_id} "
-            f"(user {args.user_id}) after {elapsed:.2f}s: {e}"
+        await _handle_graph_not_available(e, args, start_time)
+    except GraphValidationError:
+        logger.error(
+            f"Scheduled Graph {args.graph_id} failed validation. Unscheduling graph"
         )
-        # Clean up orphaned schedules for this graph
-        await _cleanup_orphaned_schedules_for_graph(args.graph_id, args.user_id)
+        if args.schedule_id:
```
majdyz: just for consistency we can make `_handle_graph_validation_error` similar to _handle_graph_not_available you introduced here. 

also can we add a warning at least for this so we know this is triggered in the log ?

majdyz (approved):

autogpt_platform/backend/backend/executor/scheduler.py
Swiftyos: We are currently logging an error on line 171. Do you want me to reduce it to a warning?

PR #11376, opened 2025-11-13
fix(frontend): upload/keep sourcemaps

0ubbe:
## Changes 🏗️

Keep the sourcemaps on the browser ( our code is open source anyway ) 

Screenshot 2025-11-13 at 20 35 25

To prevent all the noise related to it. Sourcemaps are being uploaded fine according to the logs, but keep `debug` just in case there are actual hidden issues...

### For code changes:
- [x] I have made a test plan
  - [x] we will see once this is deployed

### DIFF
--- a/autogpt_platform/frontend/next.config.mjs
+++ b/autogpt_platform/frontend/next.config.mjs
@@ -54,8 +54,13 @@ export default isDevelopmentBuild
         NEXT_PUBLIC_VERCEL_ENV: process.env.VERCEL_ENV,
       },
 
-      // Only print logs for uploading source maps in CI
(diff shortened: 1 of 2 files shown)


autogpt_platform/frontend/next.config.mjs
```diff
@@ -83,9 +88,15 @@ export default isDevelopmentBuild
         disable: false, // Source maps are enabled by default
         assets: ["**/*.js", "**/*.js.map"], // Specify which files to upload
```
Pwuts, line 89: The fix was to make this more specific; see #11378

PR #11382, opened 2025-11-14
fix(frontend): post login/signup/onboarding redirect clash

0ubbe:
### Issue 1: login/signup redirect conflict

There are 2 hooks, both on the login and signup pages, that attempt to call `router.push` once a user logs in or is created.

The main offender seems to be this hook:
```tsx
  useEffect(() => {
    if (user) router.push("/");
  }, [user]);
```
Which is in place on both pages to prevent logged-in users from accessing `/login` or `/signup`. What happens is when a user signs up or logs in, if they need onboarding, there is a `router.push` down the line to redirect them there, which conflicts with the one done in this hook.
 
**Solution**

I moved the logic from that hook to the `middleware.ts`, which is a better place for it... It won't conflict anymore with onboarding redirects done in those pages

### Issue 2: onboarding server redirects

Potential race condition: both the server component and the client `<OnboardingProvider />` perform redirects. The server component redirects happen first, but if onboarding state changes after mount, the provider can redirect again, causing rapid mount/unmount cycles.

**Solution**

Make all onboarding redirects central in `/onboarding` which is now a client component do in client redirects only and displaying a spinner while it does so.

#### For code changes:
- [x] I have made a test plan
  - [x] Tested locally login/logout/signup and trying to access `/login` and `/signup` being logged in

### DIFF
--- a/autogpt_platform/frontend/src/app/(no-navbar)/onboarding/4-agent/page.tsx
+++ b/autogpt_platform/frontend/src/app/(no-navbar)/onboarding/4-agent/page.tsx
@@ -12,17 +12,22 @@ import { useBackendAPI } from "@/lib/autogpt-server-api/context";
 import { StoreAgentDetails } from "@/lib/autogpt-server-api";
 import { isEmptyOrWhitespace } from "@/lib/utils";
 import { useOnboarding } from "../../../../providers/onboarding/onboarding-provider";
+import router from "next/router";
 
 export default function Page() {
-  const { state, updateState } = useOnboarding(4, "INTEGRATIONS");
+  const { state, updateState, completeStep } = useOnboarding(4, "INTEGRATIONS");
   const [agents, setAgents] = useState<StoreAgentDetails[]>([]);
   const api = useBackendAPI();
 
   useEffect(() => {
     api.getOnboardingAgents().then((agents) => {
+      if (agents.length < 2) {
+        completeStep("CONGRATS");
+        router.push("/onboarding");
+      }
       setAgents(agents);
     });
-  }, [api]);
+  }, []);
 
   useEffect(() => {
     // Deselect agent if it's not in the list of agents


autogpt_platform/frontend/src/app/(no-navbar)/onboarding/4-agent/page.tsx
```diff
@@ -21,13 +20,9 @@ export default function Page() {
...
   useEffect(() => {
     api.getOnboardingAgents().then((agents) => {
-      if (agents.length < 2) {
-        finishOnboarding();
-      }
-
       setAgents(agents);
     });
-  }, [api, setAgents]);
+  }, [api]);
```
0ubbe: We shouldn't be bypassing in this case but rather fix our local data so that there is always onboarding agents available...

autogpt_platform/frontend/src/lib/supabase/middleware.ts
```diff
@@ -62,7 +67,13 @@ export async function updateSession(request: NextRequest) {
       }
     }
 
-    // 2. Check if user is authenticated but lacks admin role when accessing admin pages
+    // 2. Check if user is authenticated but trying to access unauthenticated pages
+    if (user && isAuthenticationPage(pathname)) {
+      url.pathname = "/marketplace";
```
Pwuts, line 72: nit: not `/`? afaik we wanted to change the default page for logged-in users to `/library` and if we do that redirect in `/` we can avoid having hardcoded default route like `/marketplace` across many files

autogpt_platform/frontend/src/app/(no-navbar)/onboarding/4-agent/page.tsx
Pwuts: Is this related to the redirect issue? I don't think we should remove this bypass without also supplying a default offering of agents baked into the platform codebase.

autogpt_platform/frontend/src/app/(no-navbar)/onboarding/6-congrats/page.tsx
```diff
@@ -31,16 +32,17 @@ export default function Page() {
     }, 500);
 
     const timer2 = setTimeout(() => {
+      // Mark CONGRATS as complete - /onboarding page will handle adding agent to library and redirect
       completeStep("CONGRATS");
-      finishOnboarding();
```
Pwuts, line 35: do we not need this to add the agent used in onboarding to the user's library?

autogpt_platform/frontend/src/lib/supabase/middleware.ts
0ubbe: Good call yep I'll change to `/` 👍🏽

autogpt_platform/frontend/src/app/(no-navbar)/onboarding/6-congrats/page.tsx
0ubbe: This has been moved to the main `onboarding/page.tsx`, so all onboarding re-direct logic it's centralised and lives there 🙇🏽 
https://github.com/Significant-Gravitas/AutoGPT/pull/11382/files#diff-d41a3c1555c47d017163b97c8d4f686ac776eb693f0650c0106152bbc34a3f5dR29-R45

autogpt_platform/frontend/src/app/(no-navbar)/onboarding/4-agent/page.tsx
0ubbe: I can undo if we think it can cause problems down the line, but I guess having accounts "half-onboarded" can cause problems down the line?

PR #11396, opened 2025-11-18
fix(backend): resolve marketplace agent access in get_graph_execution endpoint

majdyz:
## Summary
Fixes critical issue where `GET /graphs/{graph_id}/executions/{graph_exec_id}` failed for marketplace agents with "Graph not found" errors due to incorrect version access checking.

## Root Cause
The endpoint was checking access to the **latest version** of a graph instead of the **specific version used in the execution**. This broke marketplace agents when:

1. User executes a marketplace agent (e.g., v3)
2. Graph owner later publishes a new version (e.g., v4) 
3. User tries to view execution details
4. **BUG**: Code checked access to latest version (v4) instead of execution version (v3)
5. If v4 wasn't published to marketplace → access denied → "Graph not found"

## Original Problematic Code
```python
# routers/v1.py - get_graph_execution (WRONG ORDER)
graph = await graph_db.get_graph(graph_id=graph_id, user_id=user_id)  # ❌ Uses LATEST version
if not graph:
    raise HTTPException(404, f"Graph #{graph_id} not found")

result = await execution_db.get_graph_execution(...)  # Gets execution data
```

## Solution
**Reordered operations** to check access against the **execution's specific version**:

```python
# NEW CODE (CORRECT ORDER)
result = await execution_db.get_graph_execution(...)  # ✅ Get execution FIRST

if not await graph_db.get_graph(
    graph_id=result.graph_id,
    version=result.graph_version,  # ✅ Use execution's version, not latest!
    user_id=user_id,
):
    raise HTTPException(404, f"Graph #{graph_id} not found")
```

### Key Changes Made

1. **Fixed version access logic** (routers/v1.py:1075-1095):
   - Reordered operations to get execution data first 
   - Check access using `result.graph_version` instead of latest version
   - Applied same fix to external API routes

2. **Enhanced `get_graph()` marketplace fallback** (data/graph.py:919-935):
   - Added proper marketplace lookup when user doesn't own the graph
   - Supports version-specific marketplace access checking
   - Maintains security by only allowing approved, non-deleted listings

3. **Activity status generator fix** (activity_status_generator.py:139-144):
   - Use `skip_access_check=True` for internal system operations

4. **Missing block handling** (data/graph.py:94-103):
   - Added `_UnknownBlockBase` placeholder for graceful handling of deleted blocks

## Example Scenario Fixed
1. **User**: Installs marketplace agent "Blog Writer" v3
2. **Owner**: Later publishes v4 (not to marketplace yet)  
3. **User**: Runs the agent (executes v3)
4. **Before**: Viewing execution details fails because code checked v4 access
5. **After**: ✅ Viewing execution details works because code checks v3 access

## Impact
- ✅ **Marketplace agents work correctly**: Users can view execution details for any marketplace agent version they've used
- ✅ **Backward compatibility**: Existing owned graphs continue working
- ✅ **Security maintained**: Only allows access to versions user legitimately executed
- ✅ **Version-aware access control**: Proper access checking for specific versions, not just latest

## Testing
- [x] Marketplace agents: Execution details now accessible for all executed versions
- [x] Owned graphs: Continue working as before
- [x] Version scenarios: Access control works correctly for specific versions
- [x] Missing blocks: Graceful handling without errors

**Root issue resolved**: Version mismatch between execution version and access check version that was breaking marketplace agent execution viewing.

### DIFF
--- a/autogpt_platform/backend/backend/data/graph.py
+++ b/autogpt_platform/backend/backend/data/graph.py
@@ -885,9 +885,9 @@ async def get_graph_metadata(graph_id: str, version: int | None = None) -> Graph
 
 async def get_graph(
     graph_id: str,
-    version: int | None = None,
+    version: int | None,
+    user_id: str | None,
     *,
-    user_id: str | None = None,
     for_export: bool = False,
     include_subgraphs: bool = False,
     skip_access_check: bool = False,


autogpt_platform/backend/backend/server/routers/v1.py
```diff
@@ -1069,22 +1075,25 @@ async def get_graph_execution(
...
         execution_id=graph_exec_id,
-        include_node_executions=graph.user_id == user_id,
+        include_node_executions=True,
     )
     if not result or result.graph_id != graph_id:
         raise HTTPException(
             status_code=404, detail=f"Graph execution #{graph_exec_id} not found."
         )
 
+    if not await graph_db.get_graph(
```
majdyz, line 1088: This is the main fix, make fetch version-specific

autogpt_platform/backend/backend/server/routers/v1.py
```diff
@@ -1069,22 +1075,25 @@ async def get_graph_execution(
...
-    if not graph:
-        raise HTTPException(
-            status_code=HTTP_404_NOT_FOUND, detail=f"Graph #{graph_id} not found"
-        )
-
     result = await execution_db.get_graph_execution(
         user_id=user_id,
         execution_id=graph_exec_id,
-        include_node_executions=graph.user_id == user_id,
+        include_node_executions=True,
```
majdyz, line 1081: @Pwuts I hope this is fine? IMO, the intention is always provide the node_executions unless it's not allowed? the permission check should be done inside, on the database layer, not the route concern if needed.

Pwuts: Sure, at the moment we don't shield the graph content anyway. That is still TODO to implement.

autogpt_platform/backend/backend/server/v2/otto/service.py
```diff
@@ -27,7 +27,9 @@ async def _fetch_graph_data(
             return None
 
         try:
-            graph = await graph_db.get_graph(request.graph_id, user_id=user_id)
+            graph = await graph_db.get_graph(
+                graph_id=request.graph_id, version=None, user_id=user_id
+            )
```
Pwuts, line 32: ?

majdyz: I make version a mandatory field, so the caller is concious that they are skipping the version intentionally.

Pwuts: then I think you forgot to actually make `version` mandatory on `get_graph`? :)

majdyz: nyeh you are right, how did I end up missing this, updated, can you recheck?

autogpt_platform/backend/backend/data/graph.py
```diff
@@ -885,9 +885,9 @@ async def get_graph_metadata(graph_id: str, version: int | None = None) -> Graph
 
 async def get_graph(
     graph_id: str,
-    version: int | None = None,
+    version: int | None,
+    user_id: str | None,
```
Pwuts, line 889: nit: I'd leave this as a kwarg for clarity

majdyz: I can't make mandatory field as kwargs, also we can still name args. so nothing to lose here

PR #11461, opened 2025-11-26
fix(blocks): clamp Twitter search start_time to 10 seconds before now

ntindle:
## Summary
- Clamp `start_time` to at least 10 seconds before request time (Twitter API requirement)
- Update input description to document this automatic adjustment
- Fix `serialize_list` to handle `None` data gracefully (exposed by the fix)

## Background
Twitter API returns `400 Bad Request` when `start_time` is less than 10 seconds before the request time. Users providing current/future times would hit this error.

**Sentry Issue:** [BUILDER-3PG](https://significant-gravitas.sentry.io/issues/6919685270/)

## Affected Blocks
- `TwitterSearchRecentTweetsBlock`
- `TwitterGetUserMentionsBlock`
- `TwitterGetHomeTimelineBlock`
- `TwitterGetUserTweetsBlock`

## Test plan
- [x] Tested `TwitterSearchRecentTweetsBlock` with current time as `start_time`
- [x] Verified clamping works and API call succeeds
- [x] Verified "No tweets found" is returned correctly when search window has no results


Abhi1992002 (approved):
lgtm

PR #11463, opened 2025-11-27
feat(platform): add external api routes for store search and tool usage

Swiftyos:
We want to allow external tools to explore the marketplace and use the chat agent tools

### Changes 🏗️

- add store api routes
- add tool api routes

#### For code changes:
- [x] I have made a test plan
  
  - [x] tested all endpoints work


autogpt_platform/backend/backend/server/external/routes/tools.py
```diff
@@ -0,0 +1,239 @@
...
+    Returns:
+        Execution started response with execution_id
+    """
+    session = _create_ephemeral_session(api_key.user_id)
+    result = await run_agent_tool._execute(
+        user_id=api_key.user_id,
+        session=session,
+        username_agent_slug=request.username_agent_slug,
+        inputs=request.inputs,
+    )
```
kcze, line 194: Does this return information from the comment on fail? This would be useful; "IMPORTANT: Before calling this endpoint, first call get-agent-details..."

Swiftyos: I've already created a PR that changes how this works

autogpt_platform/backend/backend/server/external/api.py
```diff
@@ -14,6 +15,7 @@
 
 external_app.add_middleware(SecurityHeadersMiddleware)
 external_app.include_router(v1_router, prefix="/v1")
+external_app.include_router(tools_router, prefix="/v1")
```
Pwuts, line 18: Doesn't this override the `/tools` prefix set on the router itself?

PR #11537, opened 2025-12-04
fix(backend): Mark ValueError as known block errors

majdyz:
### Changes 🏗️

Mark ValueError as known block errors

#### For code changes:
- [x] I have made a test plan

### DIFF
--- a/autogpt_platform/backend/backend/data/block.py
+++ b/autogpt_platform/backend/backend/data/block.py
@@ -30,6 +30,7 @@ from backend.integrations.providers import ProviderName
 from backend.util import json
 from backend.util.cache import cached
 from backend.util.exceptions import (
+    BlockError,
     BlockExecutionError,
     BlockInputError,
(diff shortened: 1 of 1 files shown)


AutoGPT-Agent:
Thank you for this PR addressing block error handling.

The change looks straightforward, but I have some concerns:

1. You're removing the import for `BlockError` while changing the code to catch `ValueError` instead of `BlockError`. This is a significant change to error handling behavior.

2. Your PR description mentions that you've made a test plan and tested according to it, but the actual test plan details are missing. For a change like this that affects error handling, it would be helpful to understand what testing was performed.

3. Can you provide more context about why `ValueError` should be treated as a known block error? Are there specific scenarios you're trying to address?

Before this can be merged, I'd recommend:
- Adding details about your test plan
- Providing more context about the motivation for this change
- Considering if any other parts of the codebase might depend on the current behavior

PR #11550, opened 2025-12-04
fix(frontend): allow to close toasts

0ubbe:
## Changes 🏗️

Screenshot 2025-12-04 at 21 05 55

Allow to close toasts 🍞 

#### For code changes:
- [x] I have made a test plan
  - [x] Run Storybook locally
  - [x] Toast have a close button top-right

### DIFF
--- a/autogpt_platform/frontend/src/components/molecules/Toast/styles.module.css
+++ b/autogpt_platform/frontend/src/components/molecules/Toast/styles.module.css
@@ -39,7 +39,8 @@ html body .toastDescription {
 }
 
 /* Position close button on the right */
-:global(html body [data-sonner-toast] [data-close-button="true"]) {
+/* stylelint-disable-next-line selector-pseudo-class-no-unknown */
+#root [data-sonner-toast] [data-close-button="true"] {
   left: unset !important;
   right: -18px !important;
   top: -3px !important;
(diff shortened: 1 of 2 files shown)


autogpt_platform/frontend/src/components/molecules/Toast/styles.module.css
```diff
@@ -37,3 +37,10 @@ html body .toastDescription {
   font-size: 0.75rem !important;
   line-height: 1.25rem !important;
 }
+
+/* Position close button on the right */
+:global(html body [data-sonner-toast] [data-close-button="true"]) {
+  left: unset !important;
+  right: -18px !important;
+  top: -3px !important;
+}
```
ntindle: aside for me: why does this have important? and also why is in css not tailwind

0ubbe: We use [Sonner](https://ui.shadcn.com/docs/components/sonner) for toasts 🍞, an external library that loads its own styles separately. 

Some things on it are customisable style-wise via props, but not everything, and the close button isn't one of them, so we manually override it here via CSS. Given that we are overriding the styles of an external library, `!important` makes sure our styles stay on top.

PR #11598, opened 2025-12-10
fix(frontend): Disable single dollar sign LaTeX mode in markdown rend…

ntindle:
Single dollar signs ($10, $variable) are commonly used in content and were being incorrectly interpreted as inline LaTeX math delimiters. This change disables that behavior while keeping double dollar sign ($$...$$) math blocks working.
## Changes 🏗️
	•	Configure remarkMath plugin with singleDollarTextMath: false in MarkdownRenderer.tsx
	•	Double dollar sign display math ($$...$$) continues to work as expected
	•	Single dollar signs are no longer interpreted as inline math delimiters
## Checklist 📋
For code changes:
	-[x]	I have clearly listed my changes in the PR description
	-[x] I have made a test plan
	-[x] I have tested my changes according to the test plan:
	-[x] Verify content with dollar amounts (e.g., “$100”) renders as plain text
	-[x] Verify double dollar sign math blocks ($$x^2$$) still render as LaTeX
	-[x] Verify other markdown features (code blocks, tables, links) still work correctly

### DIFF
--- a/autogpt_platform/frontend/src/app/(platform)/library/agents/[id]/components/NewAgentLibraryView/NewAgentLibraryView.tsx
+++ b/autogpt_platform/frontend/src/app/(platform)/library/agents/[id]/components/NewAgentLibraryView/NewAgentLibraryView.tsx
@@ -1,5 +1,6 @@
 "use client";
 
+import { LibraryAgentPreset } from "@/app/api/__generated__/models/libraryAgentPreset";
 import { Button } from "@/components/atoms/Button/Button";
 import { Breadcrumbs } from "@/components/molecules/Breadcrumbs/Breadcrumbs";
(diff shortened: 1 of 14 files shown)


AutoGPT-Agent:
Thank you for this fix to prevent single dollar signs from being interpreted as LaTeX math delimiters. This is a good improvement that will prevent confusion when displaying content with dollar amounts.

However, I noticed that the PR description checklist hasn't been completed. Before we can merge this PR, please:

1. Check the boxes in the checklist that confirm you've:
   - Clearly listed your changes in the PR description
   - Made a test plan
   - Tested your changes according to that plan

2. Please add a brief test plan that you followed to verify this change works as expected. For example, you might have tested displaying content with dollar signs to ensure they render correctly.

The code change itself looks good - it's a simple configuration update to the remarkMath plugin that addresses the issue effectively.

PR #11600, opened 2025-12-10
fix(frontend/library): Transition from empty tasks view on task init

### LINKED ISSUE #11599: `EmptyTasks` view doesn't flip when a task is created
When there are no runs, schedules, templates, or triggers for an agent, you'll get a full-width `<EmptyTasks>` view. When you create a task from this view, it doesn't change to the regular layout where you can view the task.

Pwuts:
- Resolves #11599

### Changes 🏗️

- Manually update item counts when initiating a task from `EmptyTasks` view
- Other improvements made while debugging

#### For code changes:
- [x] I have made a test plan
  - [x] `NewAgentLibraryView` transitions to full layout when a first task is created
  - [x] `NewAgentLibraryView` transitions to full layout when a first trigger is set up

### DIFF
--- a/autogpt_platform/frontend/src/app/(platform)/library/agents/[id]/components/NewAgentLibraryView/NewAgentLibraryView.tsx
+++ b/autogpt_platform/frontend/src/app/(platform)/library/agents/[id]/components/NewAgentLibraryView/NewAgentLibraryView.tsx
@@ -1,8 +1,5 @@
 "use client";
 
-import { GraphExecutionJobInfo } from "@/app/api/__generated__/models/graphExecutionJobInfo";
-import { GraphExecutionMeta } from "@/app/api/__generated__/models/graphExecutionMeta";
-import { LibraryAgentPreset } from "@/app/api/__generated__/models/libraryAgentPreset";
 import { Button } from "@/components/atoms/Button/Button";
 import { Breadcrumbs } from "@/components/molecules/Breadcrumbs/Breadcrumbs";
 import { ErrorCard } from "@/components/molecules/ErrorCard/ErrorCard";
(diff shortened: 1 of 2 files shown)


autogpt_platform/frontend/src/app/(platform)/library/agents/[id]/components/NewAgentLibraryView/components/modals/RunAgentModal/useAgentRunModal.tsx
Comment (Pwuts) [inline autogpt_platform/frontend/src/app/(platform)/library/agents/[id]/components/NewAgentLibraryView/components/modals/RunAgentModal/useAgentRunModal.tsx:1]: these changes are left over from my debugging effort; I left them because they are still an improvement

autogpt_platform/frontend/src/app/(platform)/library/agents/[id]/components/NewAgentLibraryView/NewAgentLibraryView.tsx
```diff
@@ -41,8 +43,41 @@ export function NewAgentLibraryView() {
...
+      handleCountsChange({
+        schedulesCount: 1,
+        runsCount: 0,
+        templatesCount: 0,
+        triggersCount: 0,
+      });
+
+    handleSelectRun(newSchedule.id, "scheduled");
+  }
+
```
Comment (0ubbe) [inline autogpt_platform/frontend/src/app/(platform)/library/agents/[id]/components/NewAgentLibraryView/NewAgentLibraryView.tsx]: These can be a single function:
```ts
  type ItemCreated =
    | { type: "runs"; item: GraphExecutionMeta }
    | { type: "scheduled"; item: GraphExecutionJobInfo }
    | { type: "triggers"; item: LibraryAgentPreset };

  function handleItemCreated(created: ItemCreated) {
    if (!hasAnyItems) {
      const counts = {
        runsCount: created.type === "runs" ? 1 : 0,
        schedulesCount: created.type === "scheduled" ? 1 : 0,
        templatesCount: 0,
        triggersCount: created.type === "triggers" ? 1 : 0,
      };
      handleCountsChange(counts);
    }

    handleSelectRun(created.item.id, created.type);
  }
```

It should also be moved to `useNewAgentLibraryView.ts` so we keep callbacks and hook logic tidy there 💭 

Not sure why we have `if (!agent) return;`. Given how the logic is structured, these callbacks won't be called in the UI if there is no agent.

Ideally, it's called `handle...` as well, so we can easily differentiate from actual events, given this is a handler.

Comment (Pwuts) [reply autogpt_platform/frontend/src/app/(platform)/library/agents/[id]/components/NewAgentLibraryView/NewAgentLibraryView.tsx]: We've had `on...` for event hooks and `handle...` or `do...` for button handlers for a while, and I think that's a very useful distinction

PR #11639, opened 2025-12-18
hotfix(frontend): chat should be disabled by default

0ubbe:
### Changes 🏗️

Chat should be disabled by default; otherwise, it flashes, and if Launch Darkly fails to fail, it is dangerous.

#### For code changes:
- [x] I have made a test plan
  - [x] Run locally with Launch Darkly disabled and test the above


Bentlybro (approved):

abubkarsdiq0:
personal use

PR #11661, opened 2025-12-22
fix(frontend): allow empty values in number inputs and fix AnyOfField toggle

### LINKED ISSUE #11594: Deleting "0" from number input default value turns the boolean flag off
When you try to remove the "0" from the Agent Number Input Default Value field using backspace, the default value boolean flag incorrectly gets turned off. This prevents users from clearing the default value without accidentally disabling the feature entirely. The boolean should remain in its previous state when the value is cleared, or handle empty strings appropriately.

majiayu000:
## Summary

This PR fixes two related issues with number/integer inputs in the frontend:

1. **HTMLType typo fix**: INTEGER input type was incorrectly mapped to `htmlType: 'account'` (which is not a valid HTML input type) instead of `htmlType: 'number'`.

2. **AnyOfField toggle fix**: When a user cleared a number input field, the input would disappear because `useAnyOfField` checked for both `null` AND `undefined` in `isEnabled`. This PR changes it to only check for explicit `null` (set by toggle off), allowing `undefined` (empty input) to keep the field visible.

### Root cause analysis

When a user cleared a number input:
1. `handleChange` returned `undefined` (because `v === "" ? undefined : Number(v)`)
2. In `useAnyOfField`, `isEnabled = formData !== null && formData !== undefined` became `false`
3. The input field disappeared

### Fix

Changed `useAnyOfField.tsx` line 67:
```diff
- const isEnabled = formData !== null && formData !== undefined;
+ const isEnabled = formData !== null;
```

This way:
- When toggle is OFF → `formData` is `null` → `isEnabled` is `false` (input hidden) ✓
- When toggle is ON but input is cleared → `formData` is `undefined` → `isEnabled` is `true` (input visible) ✓

## Test plan

- [x] Verified INTEGER inputs now render correctly with `type="number"`
- [x] Verified clearing a number input keeps the field visible
- [x] Verified toggling the nullable switch still works correctly

Fixes #11594

🤖 AI-assisted development disclaimer: This PR was developed with assistance from Claude Code.


AutoGPT-Agent:
Thank you for fixing the INTEGER input type issue! This is definitely a needed correction since "account" is not a valid HTML input type.

However, I noticed an unrelated change in your PR that wasn't mentioned in the description:

```diff
- [Flag.CHAT]: true,
+ [Flag.CHAT]: false,
```

This change to disable the CHAT feature flag doesn't seem related to the input type fix described in the PR. Could you either:

1. Remove this unrelated change from the PR, or
2. Update your PR description to explain why this feature flag change is included

Once that's addressed, this PR should be ready to merge. The main fix looks good and will correctly fix integer input behavior.

autogpt_platform/frontend/src/components/renderers/input-renderer/widgets/TextInputWidget/TextInputWidget.tsx
```diff
@@ -51,7 +51,7 @@ export const TextInputWidget = (props: WidgetProps) => {
       handleChange: (v: string) => (v === "" ? undefined : Number(v)),
     },
     [InputType.INTEGER]: {
-      htmlType: "account",
+      htmlType: "number",
```
Abhi1992002, line 54: @majiayu000 We can’t solve this issue this way. To fix it properly, we need to allow empty values in the `number` input, and we also need to fix the `anyOfField` file so it doesn’t close the input when its value is `null`.

autogpt_platform/frontend/src/services/feature-flags/use-get-flag.ts
```diff
@@ -48,7 +48,7 @@ const mockFlags = {
   [Flag.AGENT_FAVORITING]: false,
   [Flag.MARKETPLACE_SEARCH_TERMS]: DEFAULT_SEARCH_TERMS,
   [Flag.ENABLE_PLATFORM_PAYMENT]: false,
-  [Flag.CHAT]: true,
+  [Flag.CHAT]: false,
```
Abhi1992002, line 50: We don’t need it — you can remove it.

AutoGPT-Agent:
Thanks for this fix! The PR looks good and addresses an important issue with integer inputs.

Before we can merge:
- Please check off the test plan items after confirming that integer input fields render correctly with the number input type and that number validation works properly.

This is a valuable fix that will ensure integer inputs have proper HTML semantics and validation. Once the test plan is completed, this should be ready to merge.

majiayu000:
Addressed feedback: removed unrelated Flag.CHAT change from the branch and updated the test plan checkboxes.

majiayu000:
## CI Status Update

The test failures are in `marketplace-agent.spec.ts` E2E tests, which are unrelated to this PR's changes:

**Failing tests:**
- `Agent page details are visible`
- `Download button functionality works`
- `Add to library button works and agent appears in library`

**This PR's changes:**
- `TextInputWidget.tsx`: Fixed INTEGER input htmlType from `'account'` to `'number'`
- `useAnyOfField.tsx`: Changed `isEnabled` logic to only check for explicit `null`, allowing empty input values (`undefined`) to keep the field visible

The failing tests are about the marketplace agent page functionality, not the input renderer components. Could a maintainer please re-run the failing tests?

autogpt_platform/frontend/src/components/renderers/input-renderer/widgets/TextInputWidget/TextInputWidget.tsx
majiayu000: Hi @Abhi1992002, thanks for the review!

I believe this PR actually addresses both points you mentioned:

1. **Allow empty values in number input**: The `handleChange` function in `TextInputWidget.tsx` already handles this - when the input is empty (`v === ""`), it returns `undefined` instead of trying to convert an empty string to a number. This allows users to clear the number input.

2. **Fix anyOfField not closing when value is null**: The change in `useAnyOfField.tsx` from `formData !== null && formData !== undefined` to just `formData !== null` means:
   - `undefined` (empty input) → field stays enabled ✓
   - `null` (explicitly toggled off) → field disabled ✓

The `htmlType: "account"→"number"` fix was also necessary because `"account"` is not a valid HTML input type.

Regarding the `use-get-flag.ts` comment - I don't see any changes to that file in this PR. Could you clarify what you meant?

Please let me know if I misunderstood your feedback!

majiayu000:
The 3 failing marketplace E2E tests (`marketplace-agent...page-details-are-visible`, `marketplace-agent...button-functionality-works`, `marketplace-agent...agent-appears-in-library`) appear to be flaky tests unrelated to this PR.

This PR only modifies:
- `useAnyOfField.tsx`: Changed null check condition for `isEnabled`
- `TextInputWidget.tsx`: Fixed typo `account` → `number` for htmlType

The failing tests are marketplace-related E2E tests that don't interact with the form widgets changed in this PR. Triggered a rerun to confirm.

majiayu000:
@Abhi1992002 I pushed an update to keep AnyOfField from closing when a nullable input returns null: we now coerce null from value changes to undefined so empty number inputs stay visible, while the switch still uses explicit null when toggled off. Also the integer input type remains set to "number" with empty string -> undefined handling. Could you take another look?

Abhi1992002:
Yeah... It's right now - Thank you for contribution @majiayu000  🙌

Abhi1992002 (approved):

PR #11690, opened 2026-01-04
fix(backend): enforce strict port validation in OAuth callback URLs

### LINKED ISSUE #11670: Security: OAuth Callback Validation Bypass allows redirection to arbitrary localhost ports
#### Summary
The `validate_callback_url` function in `autogpt_platform/backend/backend/server/external/routes/integrations.py` incorrectly validates `localhost` URLs. If any `localhost` origin is configured as allowed (e.g. `http://localhost:3000`), the validation logic permits **any** port on `localhost` (e.g. `http://localhost:8080`, `http://localhost:9999`).

This violates proper OAuth validation strictness and could potentially allow authorization codes or tokens to be redirected to unintended local services or processes running on the user's machine.

#### Details
**Location:** `autogpt_platform/backend/backend/server/external/routes/integrations.py`

**Vulnerable Code Logic:**
```python
if parsed.hostname == "localhost":
    for allowed in allowed_origins:
        allowed_parsed = urlparse(allowed)
        if allowed_parsed.hostname == "localhost":
            return True  # <-- VULNERABILITY: Returns True regardless of port mismatch
```

The loop checks if *any* configuration allows localhost. If found, it immediately returns `True` for the input callback URL, completely ignoring the port number mandated by the configuration.

#### PoC
With `allowed_origins` config set to `['http://localhost:3000']`.

**Input URL:** `http://localhost:9999/callback`
**Expected Result:** Rejected (Port mismatch)
**Actual Result:** Allowed

#### Impact
*   **Severity:** Low / Medium
*   **Risk:** Token leakage to other local services. While exploiting this requires a malicious or compromised service already running on the local machine (or a multi-user environment), it bypasses the intended security restriction of the OAuth configuration. It allows for "Loopback Interface Phishing" if an attacker can reserve a port on the victim's machine.

### Recommendation
Enforce strict string matching for origins or explicitly validate the port number.

**Proposed Fix:**
```python
def validate_callback_url(callback_url: str) -> bool:
    allowed_origins = settings.config.external_oauth_callback_origins
    try:
        parsed = urlparse(callback_url)
        # Reconstruct origin strictly from the input URL
        callback_origin = f"{parsed.scheme}://{parsed.netloc}"

        # Simple, strict allowlist check
        return callback_origin in allowed_origins
    except Exception:
        return False
```

majiayu000:
## Summary

This PR fixes a security vulnerability in the OAuth callback URL validation where localhost URLs with any port were accepted if any localhost origin was configured as allowed.

- Removed special-case localhost handling that bypassed port validation
- Implemented strict origin matching (scheme + netloc) for all URLs
- Added comprehensive unit tests for the `validate_callback_url` function

## Test plan

- [x] Added unit tests covering various scenarios:
  - Exact origin match allowed
  - Different port on localhost rejected
  - Different scheme rejected
  - Different host rejected
  - Multiple allowed origins
  - Invalid URLs rejected
  - Path and query parameters don't affect validation

## Security Impact

**Before fix:** If `http://localhost:3000` was in the allowlist, `http://localhost:9999` would also be accepted, potentially allowing OAuth tokens to be redirected to malicious local services.

**After fix:** Only exact origin matches are allowed, including port number validation.

Fixes #11670

### DIFF
--- a/autogpt_platform/backend/backend/api/external/v1/integrations.py
+++ b/autogpt_platform/backend/backend/api/external/v1/integrations.py
@@ -168,26 +168,21 @@ class CreateCredentialResponse(BaseModel):
 
 
 def validate_callback_url(callback_url: str) -> bool:
-    """Validate that the callback URL is from an allowed origin."""
+    """Validate that the callback URL is from an allowed origin.
+
+    Uses strict origin matching including scheme, hostname, and port.
+    This prevents OAuth token leakage to unintended local services.
+    """
     allowed_origins = settings.config.external_oauth_callback_origins
 
     try:
         parsed = urlparse(callback_url)
+        # Reconstruct origin strictly from the input URL (scheme + netloc)
+        # netloc includes hostname and port, ensuring strict validation
         callback_origin = f"{parsed.scheme}://{parsed.netloc}"
 
-        for allowed in allowed_origins:
-            # Simple origin matching
-            if callback_origin == allowed:
-                return True
-
-        # Allow localhost with any port in development (proper hostname check)
-        if parsed.hostname == "localhost":
-            for allowed in allowed_origins:
-                allowed_parsed = urlparse(allowed)
-                if allowed_parsed.hostname == "localhost":
-                    return True
-
-        return False
+        # Simple, strict allowlist check - no special cases for localhost
+        return callback_origin in allowed_origins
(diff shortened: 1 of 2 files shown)


ntindle:
not a reasonable base issue sorry

PR #11805, opened 2026-01-19
fix(backend): Use explicit schema qualification for pgvector types

majdyz:
## Summary
- Fix intermittent "type 'vector' does not exist" errors when using PgBouncer in transaction mode
- The issue was that `SET search_path` and the actual query could run on different backend connections
- Use explicit schema qualification (`{schema}.vector`, `OPERATOR({schema}.<=>)`) instead of relying on search_path

## Test plan
- [x] Tested vector type cast on local: `'[1,2,3]'::platform.vector` works
- [x] Tested OPERATOR syntax on local: `OPERATOR(platform.<=>)` works
- [x] Tested on dev via kubectl exec: both work correctly
- [ ] Deploy to dev and verify backfill_missing_embeddings endpoint no longer errors

## Related Issues
Fixes: AUTOGPT-SERVER-763, AUTOGPT-SERVER-764


autogpt_platform/backend/migrations/20260109181714_add_docs_embedding/migration.sql
```diff
@@ -1,9 +1,10 @@
 -- CreateExtension
 -- Supabase: pgvector must be enabled via Dashboard → Database → Extensions first
--- Create in public schema so vector type is available across all schemas
+-- Creates extension in current schema (determined by search_path)
```
ntindle: i dont love changing existing migrations as it can cause issues with prisma syncing -- can this be a new miugration?

ntindle (approved):

autogpt_platform/backend/migrations/20260109181714_add_docs_embedding/migration.sql
majdyz: No because for some other this migration is breaking and I'm trying to fix it

majdyz: I'll decouple the change and make a hotfix for this if needed since it's not really needed in the code pr

ntindle (approved):

PR #11815, opened 2026-01-21
fix(frontend): auto-select credentials correctly in old builder

0ubbe:
## Changes 🏗️

On the **Old Builder**, when running an agent...

### Before

Screenshot 2026-01-21 at 21 27 05

Credentials are there, but it is not recognising them, you need to click on them to be selected

### After

Screenshot 2026-01-21 at 21 26 47

It uses the new credentials UI and correctly auto-selects existing ones.

### Other

Fixed a small timezone display glitch on the new library view.

#### For code changes:
- [x] I have made a test plan
  - [x] Run agent in old builder
  - [x] Credentials are auto-selected and using the new collapsed system credentials UI


ntindle:
Does this still allow optional credentials for the new builder + library?

0ubbe:
@ntindle I think so, we added only auto‑select required credentials (we iterate `requiredCredentials`), so optional fields remain unset and can be skipped.

PR #11846, opened 2026-01-26
fix(backend/chat): Use distributed locks for chat session mutations

Pwuts:
[SECRT-1830: Chat Session locks are not system-wide](https://linear.app/autogpt/issue/SECRT-1830)

### Changes 🏗️

- Replace the locking mechanism in `backend.api.features.chat.model` with Redis-based locking

#### For code changes:
- [ ] I have made a test plan
  
  - [ ] ...

<details>
  
</details>

### DIFF
--- a/autogpt_platform/backend/backend/api/features/chat/model.py
+++ b/autogpt_platform/backend/backend/api/features/chat/model.py
@@ -1,9 +1,8 @@
-import asyncio
 import logging
 import uuid
+from contextlib import asynccontextmanager
 from datetime import UTC, datetime
 from typing import Any
-from weakref import WeakValueDictionary
 
 from openai.types.chat import (
     ChatCompletionAssistantMessageParam,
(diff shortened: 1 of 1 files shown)


autogpt_platform/backend/backend/api/features/chat/model.py
```diff
@@ -52,28 +51,36 @@ def _get_session_cache_key(session_id: str) -> str:
...
+async def _session_lock(session_id: str):
+    """Distributed lock for a chat session using Redis.
 
-    Uses WeakValueDictionary for automatic cleanup: locks are garbage collected
-    when no coroutine holds a reference to them, preventing memory leaks from
-    unbounded growth of session locks.
+    Provides system-wide locking across horizontally scaled backend instances
+    to prevent race conditions during concurrent session mutations.
     """
-    async with _session_locks_mutex:
```
majdyz, line 71: @Pwuts why do we need locking here in the first place? and why do we need this system-wide even? can we instead remove the lock somehow? also would this cause a latency issue?

Swiftyos (changes requested):
Please can you update the PR description to explain what problem this is solving. 

This was originally added to solve the specific problem of race conditions between the session title thread and the conversation thread, which always occurs on the same instance as we prevent rapid request sends on the frontend. 

What scenario is causing race conditions across multiple instances?

PR #11855, opened 2026-01-27
fix(backend): add SSE heartbeats to prevent tool execution timeouts

Swiftyos:
## Summary

Long-running chat tools (like `create_agent` and `edit_agent`) were timing out because no SSE data was sent during tool execution. GCP load balancers and proxies have idle connection timeouts (~60 seconds), and when the external Agent Generator service takes longer than this, the connection would drop.

This PR adds SSE heartbeat comments during tool execution to keep connections alive.

### Changes 🏗️

- **response_model.py**: Added `StreamHeartbeat` response type that emits SSE comments (`: heartbeat\n\n`)
- **service.py**: Modified `_yield_tool_call()` to:
  - Run tool execution in a background asyncio task
  - Yield heartbeat events every 15 seconds while waiting
  - Handle task failures with explicit error responses (no silent failures)
  - Handle cancellation gracefully
- **create_agent.py**: Improved error messages with more context and details
- **edit_agent.py**: Improved error messages with more context and details

### How It Works

```
Tool Call → Background Task Started
     │
     ├── Every 15 seconds: yield `: heartbeat\n\n` (SSE comment)
     │
     └── Task Complete → yield tool result OR error response
```

SSE comments (`: heartbeat\n\n`) are:
- Ignored by SSE clients (don't trigger events)
- Keep TCP connections alive through proxies/load balancers
- Don't affect the AI SDK data protocol

#### For code changes:
- [x] I have made a test plan
  - [x] All chat service tests pass (17 tests)
  - [x] Verified heartbeats are sent during long tool execution
  - [x] Verified errors are properly reported to frontend

### DIFF
--- a/autogpt_platform/backend/backend/api/features/chat/service.py
+++ b/autogpt_platform/backend/backend/api/features/chat/service.py
@@ -49,6 +49,7 @@ from .response_model import (
     StreamUsage,
 )
 from .tools import execute_tool, tools
+from .tools.models import ErrorResponse
 from .tracking import track_user_message
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


autogpt_platform/backend/backend/api/features/chat/service.py
```diff
@@ -892,12 +897,53 @@ async def _yield_tool_call(
...
+            tool_name=tool_name,
+            parameters=arguments,
+            tool_call_id=tool_call_id,
+            user_id=session.user_id,
+            session=session,
+        )
     )
 
+    # Yield heartbeats every 15 seconds while waiting for tool to complete
+    heartbeat_interval = 15.0  # seconds
```
majdyz, line 912: we can make this env configurable

PR #11880, opened 2026-01-29
fix(executor): Mark node as FAILED when block yields error output

Otto-AGPT:
## Problem

Blocks that catch exceptions internally and yield `"error"` output instead of re-raising were incorrectly marked as `COMPLETED` by the executor. This caused runs to show as "success" in the UI even when they actually failed.

Example from `ai_image_generator_block.py`:
```python
async def run(self, input_data, *, credentials, **kwargs):
    try:
        url = await self.generate_image(input_data, credentials)
        ...
    except Exception as e:
        yield "error", str(e)  # Swallows exception, executor sees no failure
```

This pattern exists in **54 blocks** across the codebase.

## Solution

Track if any `"error"` output is yielded during node execution. If so, mark the node as `FAILED` even when no exception propagates to the executor.

## Changes

- Added `has_error_output` flag in `_execute_node`
- Check for `"error"` output name in `persist_output`
- Set status to `FAILED` if error output was yielded

## Testing

This fix affects all blocks that use the `yield "error"` pattern. Manual testing recommended with blocks like:
- AI Image Generator
- Code Executor
- Reddit blocks
- Email block

## Linear

Fixes: SECRT-1844

## Reported By

@0ubbe in #eng-general

### DIFF
--- a/autogpt_platform/backend/backend/executor/manager.py
+++ b/autogpt_platform/backend/backend/executor/manager.py
@@ -616,8 +616,13 @@ class ExecutionProcessor:
         nodes_to_skip: Optional[set[str]] = None,
     ) -> ExecutionStatus:
         status = ExecutionStatus.RUNNING
+        has_error_output = False
 
         async def persist_output(output_name: str, output_data: Any) -> None:
+            nonlocal has_error_output
+            if output_name == "error":
+                has_error_output = True
+
             await db_client.upsert_execution_output(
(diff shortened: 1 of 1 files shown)


majdyz:
@Otto-AGPT this is not very clear on what specific issue are we solving and why the change is too huge, I'm closing it for now but feel free to re-open if the change is actually needed. The change doesn't seem surgical.

PR #11882, opened 2026-01-29
fix(backend): Auto-correct content-type based on file signature instead of erroring

Otto-AGPT:
## Summary
When uploading media files, the browser-declared `content-type` header sometimes doesn't match the actual file content (e.g., user renames a PNG to .jpg). Instead of rejecting these files with "File signature does not match content type" error, we now:

1. **Detect** the actual content type from file magic bytes
2. **Log** when auto-correction occurs for debugging
3. **Use** the detected type for storage and processing

## Changes
- Modified `media.py` to trust file signature over header
- Updated tests to expect auto-correction behavior
- Improved error message for truly invalid files

## Why
This improves UX while maintaining security - we still validate that files are legitimate images/videos, we just trust the actual content over the potentially incorrect header.

Fixes: "File signature does not match content type" Sentry alerts

### DIFF
--- a/autogpt_platform/backend/backend/api/features/store/media.py
+++ b/autogpt_platform/backend/backend/api/features/store/media.py
@@ -71,46 +71,41 @@ async def upload_media(
         logger.error(f"Error reading file content: {str(e)}")
         raise store_exceptions.FileReadError("Failed to read file content") from e
 
-    # Validate file signature/magic bytes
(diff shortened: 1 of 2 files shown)


Otto-AGPT:
Closing this PR from fork and recreating from main repo branch (fork PRs can't access CI secrets)

PR #11885, opened 2026-01-29
fix(frontend): Fix flaky library pagination E2E test

Otto-AGPT:
## Summary
Fixes the flaky E2E test `Library › pagination works correctly` that has been causing PRs to be removed from the merge queue (e.g., PR #11841).

### 1. Increase test data volume (`e2e_test_data.py`)
- Changed `MIN/MAX_GRAPHS_PER_USER` from 15 → 22
- Changed `MIN/MAX_AGENTS_PER_USER` from 15 → 22
- Updated `create_test_library_agents()` to use the constants instead of hardcoded `10`
- Library page uses `page_size: 20`, so 22 agents ensures pagination is testable

### 2. Fix test logic (`library.spec.ts`)
- Added `PAGE_SIZE` constant (20)
- Updated pagination guard from `>= 10` to `>= PAGE_SIZE`
- Removed inline comments per coding guidelines

### 3. Improve wait reliability (`library.page.ts`)
- Replaced fixed 2-second timeout with `waitForLoadState('networkidle')`
- Added card-count polling via `waitForFunction()` to detect when new agents load
- More reliable in CI environments where execution can be slower

## Root Cause
The test was flaky because:
1. Test created only 15 agents, but page size is 20 → no pagination ever occurred
2. `create_test_library_agents()` was hardcoded to 10, ignoring the constants
3. Test logic assumed 10 agents = pagination exists (wrong threshold)
4. Fixed timeout wasn't reliable in slower CI environments

## Testing
- [x] Lint/format passes (black, isort, ruff, prettier, eslint)
- [x] Library pagination test now passes ✅
- [ ] Full E2E suite (monitoring)

## Related
- Fixes SECRT-1848
- Unblocks PR #11841


Otto-AGPT:
Closing in favor of a comprehensive fix.

**Analysis:** This PR addressed library pagination timing but the E2E failures are primarily caused by marketplace tests having no data - not timing issues. The root cause is probabilistic test data creation that can result in 0 featured agents/creators.

**New approach:** Creating a single PR that:
1. Makes marketplace test data creation **deterministic** (guaranteed minimums)
2. Adds proper waits **after** verifying data exists
3. Fixes both library and marketplace tests together

See replacement PR for the comprehensive fix.

Otto-AGPT:
## Correction to My Analysis

After re-examining the e2e_test_data.py code, I need to correct my root cause analysis:

**What I said:** "uses random chances (40% approved, then 20-50% featured) which can result in no featured content at all"

**What I missed:** The code has deterministic guarantees I overlooked:
```python
GUARANTEED_FEATURED_AGENTS = 8
GUARANTEED_FEATURED_CREATORS = 5  
GUARANTEED_TOP_AGENTS = 10
```

The 40% approval chance only applies AFTER the first 10 submissions. The guarantees should prevent 0 featured content if the script runs successfully.

**Revised analysis:** If E2E tests showed 0 featured agents, the issue was likely:
1. Test data script failing upstream (user/graph creation)
2. Script not being run at all in CI
3. Database connection issues

Not the randomness itself. My analysis was incomplete.

---
*Correction posted after ground-truth verification.*

Issue #4, opened 2023-03-29
Improve web-crawling system


Torantulino:
Auto-GPT should be able to see links on webpages it visits, as well as the existing GPT-3 powered summary, so that it can browse further than 1 page away from google.

Torantulino:
This is currently majorly limiting Auto-GPT's capabilities.
It essentially is unable to reliably browse the web and frequently misses information even though it's visiting the correct URLs.

Andythem23:
Maybe a different prompt will do it for GPT-3.5

0xcha05:
can you provide more context, please? 
I've tried "extract key information from this webpage text" and given it hackernews text, and it did a decent job.

Torantulino:
## Current Implementation

Web browsing is currently handled in the following way:

```python
def browse_website(url):
    summary = get_text_summary(url)
    links = get_hyperlinks(url)

    # Limit links to 5
    if len(links) > 5:
        links = links[:5]

    result = f"""Website Content Summary: {summary}\n\nLinks: {links}"""

    return result

def get_text_summary(url):
    text = browse.scrape_text(url)
    summary = browse.summarize_text(text)
    return """ "Result" : """ + summary

def get_hyperlinks(url):
    link_list = browse.scrape_links(url)
    return link_list
``` 

Where `scrape_text` uses BeautifulSoup to extract the text contained within a webpage.

## The Problem

The `summarize_text` function feeds GPT3.5 this scraped text, and whilst gpt3.5 does a great job of summarising it, it doesn't know what we're specifically after.

This leads to instances where AutoGPT can be looking to find some news on CNN, and instead receives a summary of what CNN is.

### An Illustrated Example 

![image](https://user-images.githubusercontent.com/22963551/229337183-a40fe39f-e2d2-4467-a3c5-bf8330163695.png)
- Here, Auto-GPT is trying to browse a news site to find a technology related news story.
- They select https://www.theverge.com/tech, a screenshot from which can be seen below:



<img width="322" alt="2022-04-09_122409" src="https://user-images.githubusercontent.com/22963551/229337269-e4779dcf-689b-4585-9b62-350a9c6d8128.png">




- Even though there are clearly news stories contained within this page, the summary does not include them specifically, just mentions they are there.
- 
**Summary:**
[...]

Issue #3, opened 2023-03-28
The original link is dead


bbecausereasonss:
The original git pull link is dead, and this repo doesn't contain the files necessary ie) requirements.txt

Torantulino:
Thanks for submitting our first issue!
This should be fixed now, let me know if you find any other issues or have any feature requests.

PR #1, opened 2023-03-16
Complete prompt redesign

E-GPT:
Complete prompt redesign using my "Self Improvement Module".

"""Prompt: You are Entrepreneur-GTP's self improvement module. Your goal is to improve the following prompt in order to help Entrepreneur-GTP operate fully autonomously."""

PR #11891, opened 2026-01-30
feat(library): Order Library Agents by Most Recently Ran

### LINKED ISSUE #9860: Order Agents in Library by "most recently ran"
Currently the Agents list in the Library is ordered based on time of last edit to an agent in the builder, instead this should be ordered by the time of last execution. (all executions should be counted, not just those manually triggered through the GUI).

Medyan-Naser:
Closes #9860 

This PR updates the ordering logic for the Agents Library to sort agents by most recent execution time rather than last edited time in the builder.

All agent executions are considered, including:
- Runs triggered via the GUI
- Automated or programmatic runs

Agents that have never been executed are placed after agents with at least one run and are ordered by their creation date.

### Changes 🏗️

- Add `LAST_EXECUTED` sort option to `LibraryAgentSort` enum
- Implement in-memory sorting by most recent execution time in `list_library_agents`, with pagination applied after sort
- Order executions by `updatedAt` (instead of `createdAt`) in `library_agent_include` to reflect completion time
- Add "Last Executed" as default sort option in the library sort dropdown (frontend)
- Add `lastExecuted` to `LibraryAgentSortEnum` in frontend types and OpenAPI spec
- Add unit test for `LAST_EXECUTED` sorting behavior

#### For code changes:
- [x] I have made a test plan
  - [x] Verified agents with executions appear first, sorted by most recent execution
  - [x] Verified agents with no executions appear at the end, sorted by creation date
  - [x] Verified "Last Executed" appears as default in sort dropdown
  - [x] Verified URL defaults to `?sort=lastExecuted`
  - [x] Changed sort options and verified order updates correctly

### DIFF
--- a/autogpt_platform/backend/backend/api/features/library/db.py
+++ b/autogpt_platform/backend/backend/api/features/library/db.py
@@ -785,10 +785,6 @@ async def update_library_agent(
     if is_archived is not None:
         update_fields["isArchived"] = is_archived
     if is_deleted is not None:
-        if is_deleted is True:
-            raise RuntimeError(
-                "Use delete_library_agent() to (soft-)delete library agents"
-            )
         update_fields["isDeleted"] = is_deleted
     if settings is not None:
         existing_agent = await get_library_agent(id=library_agent_id, user_id=user_id)
--- a/autogpt_platform/backend/backend/api/features/library/db.py
+++ b/autogpt_platform/backend/backend/api/features/library/db.py
@@ -124,16 +124,64 @@ async def list_library_agents(
     elif sort_by == library_model.LibraryAgentSort.UPDATED_AT:
         order_by = {"updatedAt": "desc"}
 
-    library_agents = await prisma.models.LibraryAgent.prisma().find_many(
-        where=where_clause,
-        include=library_agent_include(
-            user_id, include_nodes=False, include_executions=include_executions
-        ),
-        order=order_by,
-        skip=(page - 1) * page_size,
-        take=page_size,
-    )
-    agent_count = await prisma.models.LibraryAgent.prisma().count(where=where_clause)
+    try:
+        # For LAST_EXECUTED sorting, we need to fetch execution data and sort in Python
+        # since Prisma doesn't support sorting by nested relations
+        if sort_by == library_model.LibraryAgentSort.LAST_EXECUTED:
+            # TODO: This fetches all agents into memory for sorting, which may cause
+            # performance issues for users with many agents. Prisma doesn't support
+            # sorting by nested relations, so a dedicated lastExecutedAt column or
+            # raw SQL query would be needed for database-level pagination.
+            library_agents = await prisma.models.LibraryAgent.prisma().find_many(
+                where=where_clause,
+                include=library_agent_include(
+                    user_id,
+                    include_nodes=False,
+                    include_executions=True,
+                    execution_limit=1,
+                ),
+            )
+
+            def get_sort_key(
+                agent: prisma.models.LibraryAgent,
+            ) -> tuple[int, float]:
+                """
+                Returns a tuple for sorting: (has_no_executions, -timestamp).
+
+                Agents WITH executions come first (sorted by most recent execution),
+                agents WITHOUT executions come last (sorted by creation date).
+
+                Uses updatedAt (not createdAt) because it reflects when the execution
+                completed or last progressed, showing recently-finished agents first.
+                """
+                graph = agent.AgentGraph
+                if graph and graph.Executions and len(graph.Executions) > 0:
+                    execution = graph.Executions[0]
+                    timestamp = execution.updatedAt or execution.createdAt
+                    return (0, -timestamp.timestamp())
+                return (1, -agent.createdAt.timestamp())
+
+            library_agents.sort(key=get_sort_key)
+
+            # Apply pagination after sorting
+            agent_count = len(library_agents)
+            start_idx = (page - 1) * page_size
+            end_idx = start_idx + page_size
+            library_agents = library_agents[start_idx:end_idx]
+        else:
+            # Standard sorting via database
+            library_agents = await prisma.models.LibraryAgent.prisma().find_many(
+                where=where_clause,
+                include=library_agent_include(
+                    user_id, include_nodes=False, include_executions=False
+                ),
+                order=order_by,
+                skip=(page - 1) * page_size,
+                take=page_size,
+            )
+            agent_count = await prisma.models.LibraryAgent.prisma().count(
+                where=where_clause
+            )
 
     logger.debug(f"Retrieved {len(library_agents)} library agents for user #{user_id}")
 
--- a/autogpt_platform/backend/backend/api/features/library/db.py
+++ b/autogpt_platform/backend/backend/api/features/library/db.py
@@ -408,15 +408,6 @@ async def get_library_agent_by_graph_id(
     graph_version: Optional[int] = None,
     include_archived: bool = False,
 ) -> library_model.LibraryAgent | None:
-    filter: prisma.types.LibraryAgentWhereInput = {
-        "agentGraphId": graph_id,
-        "userId": user_id,
-        "isDeleted": False,
-    }
-    if not include_archived:
-        filter["isArchived"] = False
-    if graph_version is not None:
-        filter["agentGraphVersion"] = graph_version
     """
     Retrieves a library agent by its graph ID for a given user.
 
@@ -436,6 +427,8 @@ async def get_library_agent_by_graph_id(
         "userId": user_id,
         "isDeleted": False,
     }
+    if not include_archived:
+        filter["isArchived"] = False
     if graph_version is not None:
         filter["agentGraphVersion"] = graph_version
 
--- a/autogpt_platform/backend/backend/api/features/library/db.py
+++ b/autogpt_platform/backend/backend/api/features/library/db.py
@@ -23,6 +23,7 @@ from backend.data.includes import (
     AGENT_PRESET_INCLUDE,
     LIBRARY_FOLDER_INCLUDE,
     MAX_LIBRARY_AGENT_EXECUTIONS_FETCH,
+    MAX_LIBRARY_AGENTS_LAST_EXECUTED_FETCH,
     library_agent_include,
 )
 from backend.data.model import CredentialsMetaInput, GraphInput
@@ -128,12 +129,13 @@ async def list_library_agents(
     # For LAST_EXECUTED sorting, we need to fetch execution data and sort in Python
     # since Prisma doesn't support sorting by nested relations
     if sort_by == library_model.LibraryAgentSort.LAST_EXECUTED:
-        # TODO: This fetches all agents into memory for sorting, which may cause
-        # performance issues for users with many agents. Prisma doesn't support
-        # sorting by nested relations, so a dedicated lastExecutedAt column or
-        # raw SQL query would be needed for database-level pagination.
+        # TODO: This fetches up to MAX_LIBRARY_AGENTS_LAST_EXECUTED_FETCH agents
+        # into memory for sorting. Prisma doesn't support sorting by nested relations,
+        # so a dedicated lastExecutedAt column or raw SQL query would be needed for
+        # database-level pagination. The ceiling prevents worst-case memory blowup.
         library_agents = await prisma.models.LibraryAgent.prisma().find_many(
             where=where_clause,
+            take=MAX_LIBRARY_AGENTS_LAST_EXECUTED_FETCH,
             include=library_agent_include(
                 user_id,
                 include_nodes=False,
--- a/autogpt_platform/backend/backend/api/features/library/db.py
+++ b/autogpt_platform/backend/backend/api/features/library/db.py
@@ -163,7 +163,9 @@ async def list_library_agents(
         library_agents.sort(key=get_sort_key)
 
         # Apply pagination after sorting
-        agent_count = len(library_agents)
+        agent_count = await prisma.models.LibraryAgent.prisma().count(
+            where=where_clause
+        )
         start_idx = (page - 1) * page_size
         end_idx = start_idx + page_size
         page_agents = library_agents[start_idx:end_idx]
--- a/autogpt_platform/backend/backend/api/features/library/db.py
+++ b/autogpt_platform/backend/backend/api/features/library/db.py
@@ -59,7 +59,7 @@ async def list_library_agents(
     Args:
         user_id: The ID of the user whose LibraryAgents we want to retrieve.
         search_term: Optional string to filter agents by name/description.
-        sort_by: Sorting field (createdAt, updatedAt, isFavorite, isCreatedByUser).
+        sort_by: Sorting field (createdAt, updatedAt, lastExecuted).
         page: Current page (1-indexed).
         page_size: Number of items per page.
         folder_id: Filter by folder ID. If provided, only returns agents in this folder.
--- a/autogpt_platform/backend/backend/api/features/library/db.py
+++ b/autogpt_platform/backend/backend/api/features/library/db.py
@@ -124,61 +124,60 @@ async def list_library_agents(
     elif sort_by == library_model.LibraryAgentSort.UPDATED_AT:
         order_by = {"updatedAt": "desc"}
 
-    try:
-        # For LAST_EXECUTED sorting, we need to fetch execution data and sort in Python
-        # since Prisma doesn't support sorting by nested relations
-        if sort_by == library_model.LibraryAgentSort.LAST_EXECUTED:
-            # TODO: This fetches all agents into memory for sorting, which may cause
-            # performance issues for users with many agents. Prisma doesn't support
-            # sorting by nested relations, so a dedicated lastExecutedAt column or
-            # raw SQL query would be needed for database-level pagination.
-            library_agents = await prisma.models.LibraryAgent.prisma().find_many(
-                where=where_clause,
-                include=library_agent_include(
-                    user_id,
-                    include_nodes=False,
-                    include_executions=True,
-                    execution_limit=1,
-                ),
-            )
+    # For LAST_EXECUTED sorting, we need to fetch execution data and sort in Python
+    # since Prisma doesn't support sorting by nested relations
+    if sort_by == library_model.LibraryAgentSort.LAST_EXECUTED:
+        # TODO: This fetches all agents into memory for sorting, which may cause
+        # performance issues for users with many agents. Prisma doesn't support
+        # sorting by nested relations, so a dedicated lastExecutedAt column or
+        # raw SQL query would be needed for database-level pagination.
+        library_agents = await prisma.models.LibraryAgent.prisma().find_many(
+            where=where_clause,
+            include=library_agent_include(
+                user_id,
+                include_nodes=False,
+                include_executions=True,
+                execution_limit=1,
+            ),
+        )
 
-            def get_sort_key(
-                agent: prisma.models.LibraryAgent,
-            ) -> tuple[int, float]:
-                """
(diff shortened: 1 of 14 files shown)


ntindle:
can you clean up the extra changes?

Medyan-Naser:
> can you clean up the extra changes?
@ntindle 
I had the wrong target branch, now switched the target branch to master.

ntindle:
We don't allow direct merges into master sorry :(

Medyan-Naser:
@ntindle 
what should the target branch be?

ntindle:
dev

Medyan-Naser:
Hi @ntindle,
I rebased with dev branch and resolved all conflicts.

ntindle:
## Test Results ✅

**UI Tests:**
- ✅ "Last Executed" appears as default in sort dropdown
- ✅ URL defaults to `?sort=lastExecuted`
- ✅ Sort dropdown shows: Last Executed, Creation Date, Last Modified  
- ✅ Changing sort updates URL correctly (`?sort=createdAt` for Creation Date)
- ✅ Dropdown label updates to show selected option

**Note:** Data-dependent tests (agents with executions first, agents without executions last) require seeded test data with varying execution timestamps. The frontend UI changes and URL routing work correctly.

Default State
*Default: "Last Executed" selected*

Dropdown Open
*Sort options available*

All UI functionality verified working! 🎉

Medyan-Naser:
Hi @ntindle,

I added a test case to address your note, the test case I added verify that :
    - Agents with executions come first, sorted by most recent execution
    - Agents without executions come last, sorted by creation date

The new test is included in the CI and its passes.

Regarding the failed test cases, they seem not relevant to my changes. They sometimes pass in older CI runs.

ntindle (dismissed):

ntindle:
## E2E Test Failure Root Cause

The e2e tests are failing because of this code in `useLibraryListPage.ts`:

```typescript
// Ensure sort param is always present in URL (even if default)
useEffect(() => {
  if (!librarySortRaw) {
    setLibrarySortRaw(LibraryAgentSort.lastExecuted, { shallow: false });
  }
}, [librarySortRaw, setLibrarySortRaw]);
```

This forces `?sort=lastExecuted` into the URL even for the default sort value.

**Why tests fail:**
1. PR also changed login/signup redirect logic to use `getHomepageRoute(isChatEnabled)`
2. In CI, `isChatEnabled = false`, so users land on `/library` instead of `/marketplace`
3. Library page then adds `?sort=lastExecuted` to URL
4. 35+ test assertions expect `/marketplace` or `/library` without query params

**Fix options:**
1. **Remove the useEffect** - Sort still works, just won't force param into URL for default value. Minimal impact.
2. **Update all 35+ test assertions** - More work, keeps URL behavior.

Recommend option 1 - cleaner URLs and no test changes needed.

ntindle:
Not positive that's accurate but my agents decided that was the issue. Only thing left for this pr for me is getting the tests passing :)

Medyan-Naser:
Hi @ntindle ,
I fixed the failing test cases that you mentioned; however there are still failures due to "Test timeout of 25000ms exceeded."

ntindle:
Pushed a fix for the e2e test failures:

- Reverted the signup.spec.ts changes (the signup flow still ends at `/marketplace` via the test helper, so those changes weren't needed)
- Kept the signin.spec.ts changes (`/library?sort=updatedAt` → `/library`) since you correctly removed the useEffect that was forcing the sort param

CI should pass now. Thanks for the contribution! 🎉

autogpt_platform/backend/backend/api/features/library/db.py
```diff
@@ -104,18 +104,60 @@ async def list_library_agents(
...
-        agent_count = await prisma.models.LibraryAgent.prisma().count(
-            where=where_clause
-        )
+        # For LAST_EXECUTED sorting, we need to fetch execution data and sort in Python
+        # since Prisma doesn't support sorting by nested relations
+        if sort_by == library_model.LibraryAgentSort.LAST_EXECUTED:
+            # TODO: This fetches all agents into memory for sorting, which may cause
+            # performance issues for users with many agents. Prisma doesn't support
+            # sorting by nested relations, so a dedicated lastExecutedAt column or
+            # raw SQL query would be needed for database-level pagination.
```
Pwuts: This is also why we haven't implemented it so far. I'm not sure what effect it will have (on DB/backend load) to make this the default behavior, and I'm a bit concerned about it. We'll be scaling up soon and library listing is a relatively common request. We currently have users with hundreds if not thousands of agents.

Pwuts (commented):
Looks good other than the one comment I posted :)

Medyan-Naser:
I changed the default back to updateAt and updated the test cases.
I do not think the failures in the tests are related to my changes. 
Please advice.

Medyan-Naser:
Hi @ntindle, I rebased with dev, and changed the default back to updatedAt.

autogpt_platform/backend/backend/api/features/library/db.py
ntindle: IMO if it becomes a problem we can address it then

ntindle (dismissed):

Medyan-Naser:
I rebased with the dev branch.

Otto-AGPT:
Fixed the CI failures — there were two orphaned `try:` blocks in `db.py` (missing `except`/`finally` clauses) that caused `SyntaxError` across all backend test jobs (3.11, 3.12, 3.13).

**Changes:**
- Removed incomplete `try:` in `list_library_agents` (the if/else block doesn't need error handling wrapping)
- Removed incomplete `try:` in `get_library_agent_by_graph_id` (same issue)

The other CI failures (`types`, `end-to-end tests`, `Vercel`) are infrastructure/unrelated — the `types` job fails because the backend server isn't running during that CI step, and Vercel needs authorization.

autogpt_platform/backend/backend/api/features/library/db.py
```diff
@@ -102,18 +102,64 @@ async def list_library_agents(
...
+        if sort_by == library_model.LibraryAgentSort.LAST_EXECUTED:
+            # Fetch all agents with their most recent execution (limit 1)
+            library_agents = await prisma.models.LibraryAgent.prisma().find_many(
+                where=where_clause,
+                include=library_agent_include(
+                    user_id, include_nodes=False, include_executions=True, execution_limit=1
+                ),
+            )
+            
+            # Sort by most recent execution time
```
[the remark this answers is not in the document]
Otto-AGPT: 🤖 Acknowledged — this is a known limitation documented in the TODO comment in the code. Prisma doesn't support sorting by nested relations, so in-memory sorting is the only option without a schema change. ntindle agreed to address it if/when it becomes a bottleneck (e.g. adding a denormalized `lastExecutedAt` column). Not fixing in this PR.

autogpt_platform/backend/backend/api/features/library/db.py
Otto-AGPT: 🤖 Agreed on the concern. The TODO in the code documents this tradeoff. For now this is opt-in (only triggered when user explicitly selects "Last Executed" sort), and the default remains `updatedAt` which uses DB-level pagination. If it becomes a bottleneck at scale, a denormalized `lastExecutedAt` column on `LibraryAgent` would be the right fix.

autogpt_platform/backend/backend/data/includes.py
```diff
@@ -119,7 +119,9 @@ def library_agent_include(
         if include_executions:
             agent_graph_include["Executions"] = {
                 "where": {"userId": user_id},
-                "order_by": {"createdAt": "desc"},
+                "order_by": {
+                    "updatedAt": "desc"
+                },  # Uses updatedAt because it reflects when the executioncompleted or last progressed
```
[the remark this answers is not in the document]
Otto-AGPT: 🤖 Already fixed by the contributor in commit 8ec86ea.

autogpt_platform/backend/backend/data/includes.py
```diff
@@ -126,7 +126,9 @@ def library_agent_include(
         if include_executions:
             agent_graph_include["Executions"] = {
                 "where": {"userId": user_id},
-                "order_by": {"createdAt": "desc"},
+                "order_by": {
+                    "updatedAt": "desc"
+                },  # Uses updatedAt because it reflects when the execution completed or last progressed
```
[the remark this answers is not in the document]
Otto-AGPT: 🤖 Not a real issue. The Python sort key already handles NULL `updatedAt` with a fallback:
```python
timestamp = execution.updatedAt or execution.createdAt
```
If `updatedAt` is NULL, it falls back to `createdAt`. The DB-level `order_by` on `updatedAt` only selects the *single most recent* execution per agent (`take: 1`), and the actual sort order is computed in Python — so NULL ordering at the DB level doesn't affect the final result.

autogpt_platform/frontend/src/app/(platform)/library/components/LibrarySortMenu/LibrarySortMenu.tsx
```diff
@@ -23,10 +23,13 @@ export function LibrarySortMenu({ setLibrarySort }: Props) {
       <Select onValueChange={handleSortChange}>
         <SelectTrigger className="ml-1 w-fit space-x-1 border-none px-0 text-base underline underline-offset-4 shadow-none">
           <ArrowDownNarrowWideIcon className="h-4 w-4 sm:hidden" />
-          <SelectValue placeholder="Last Modified" />
+          <SelectValue placeholder="Last Executed" />
```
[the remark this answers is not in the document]
Otto-AGPT: 🤖 Not a bug. The placeholder text "Last Executed" matches the actual default sort applied by the backend (`useLibraryListPage` defaults to `lastExecuted`). The Select is intentionally uncontrolled — the placeholder reflects the default state accurately, and selecting a different option updates both the UI and the URL param. No mismatch exists.

autogpt_platform/backend/backend/api/features/library/db.py
```diff
@@ -125,16 +125,60 @@ async def list_library_agents(
...
+        # raw SQL query would be needed for database-level pagination.
+        library_agents = await prisma.models.LibraryAgent.prisma().find_many(
+            where=where_clause,
+            include=library_agent_include(
+                user_id,
+                include_nodes=False,
+                include_executions=True,
+                execution_limit=1,
+            ),
+        )
```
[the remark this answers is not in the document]
Otto-AGPT: 🤖 Fixed in 65886f6cb: passed the caller's `include_executions` parameter instead of the hardcoded `False`. Good catch — this was a real regression.

Otto-AGPT:
🤖 Addressing remaining CodeRabbit outside-diff-range comments:

1. **Outdated docstring sort options** (db.py:46-50) — Fixed in aa112061e: updated to `(createdAt, updatedAt, lastExecuted)`.

2. **`user_id` validation in `library_agent_include`** (includes.py:82-109) — Pre-existing; this function is only called from `list_library_agents` and other internal callers that already validate `user_id` upstream. Out of scope for this PR.

3. **OpenAPI default sort still says `updatedAt`** (openapi.json:3358) — This is correct. The backend route default is `UPDATED_AT`; the frontend overrides the UI default to `lastExecuted` separately.

autogpt_platform/frontend/src/app/(platform)/library/components/LibrarySortMenu/LibrarySortMenu.tsx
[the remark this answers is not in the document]
Otto-AGPT: 🤖 Fixed in 6f0097bc0: The contributor reverted the default sort to `updatedAt` (to fix E2E tests) but the placeholder and default label still said "Last Executed". Updated both `LibrarySortMenu.tsx` placeholder and `useLibrarySortMenu.ts` default label to "Last Modified" to match the actual default sort behavior.

autogpt_platform/backend/backend/api/features/library/db.py
```diff
@@ -125,16 +125,60 @@ async def list_library_agents(
...
+            Returns a tuple for sorting: (has_no_executions, -timestamp).
+
+            Agents WITH executions come first (sorted by most recent execution),
+            agents WITHOUT executions come last (sorted by creation date).
+            """
+            graph = agent.AgentGraph
+            if graph and graph.Executions and len(graph.Executions) > 0:
+                execution = graph.Executions[0]
+                timestamp = execution.updatedAt or execution.createdAt
+                return (0, -timestamp.timestamp())
```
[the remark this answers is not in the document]
Otto-AGPT: 🤖 Not a real issue. While `updatedAt` is typed as `DateTime?` in the schema, the `@updatedAt` directive auto-sets it on record creation and every subsequent update — so it's never NULL for any existing execution record. PostgreSQL NULL ordering is irrelevant here since the column always has a value. A defensive `nulls: last` clause would be harmless but unnecessary.

Pwuts (dismissed):

Pwuts:
CI for forks is broken, working on a fix here: #12479

**Update:** fixed

autogpt_platform/backend/backend/api/features/library/db.py
```diff
@@ -124,16 +124,60 @@ async def list_library_agents(
...
+            return (1, -agent.createdAt.timestamp())
+
+        library_agents.sort(key=get_sort_key)
+
+        # Apply pagination after sorting
+        agent_count = len(library_agents)
+        start_idx = (page - 1) * page_size
```
majdyz, line 169: 🤖 🔴 **Blocker**: When `LAST_EXECUTED` sort is active, `library_agent_include` is called with `execution_limit=1`, meaning only 1 execution is fetched per agent. However `LibraryAgent.from_db` computes `execution_count = len(executions)`, `success_rate`, `avg_correctness_score`, and calls `_calculate_agent_status(executions, ...)` all from that same truncated list. As a result, every agent will show `execution_count = 1` and a wildly inaccurate `success_rate` (100% or 0%) whenever the library is sorted by Last Executed.

The sorting-only fetch (with `execution_limit=1`) is correct for determining sort order, but after pagination is applied, the agents on the current page need to be re-fetched (or their executions fetched separately) with the full `execution_limit` (i.e., `MAX_LIBRARY_AGENT_EXECUTIONS_FETCH`) so that metrics are accurate.

The unit test does not assert on `execution_count` or `success_rate`, so this slips through.

autogpt_platform/backend/backend/api/features/library/db_test.py
```diff
@@ -224,3 +225,183 @@ async def test_add_agent_to_library_not_found(mocker):
...
+    assert (
+        result.agents[0].id == "lib1"
+    ), "Agent with most recent execution should be first"
+    assert result.agents[1].id == "lib2", "Agent with older execution should be second"
+    assert (
+        result.agents[2].id == "lib3"
+    ), "Agent without executions (newer) should be third"
+    assert (
+        result.agents[3].id == "lib4"
+    ), "Agent without executions (older) should be last"
```
majdyz, line 542: 🤖 🟡 **Nice to Have**: The test only verifies sort order (agent IDs). It would be worth adding assertions on `execution_count`, `success_rate`, and agent `status` to catch regressions where those metrics are computed from a truncated execution list (e.g. the bug in the `LAST_EXECUTED` branch where `execution_limit=1` causes all metrics to be derived from only 1 execution).

autogpt_platform/backend/backend/api/features/library/db.py
```diff
@@ -124,16 +124,60 @@ async def list_library_agents(
...
+        # raw SQL query would be needed for database-level pagination.
+        library_agents = await prisma.models.LibraryAgent.prisma().find_many(
+            where=where_clause,
+            include=library_agent_include(
+                user_id,
+                include_nodes=False,
+                include_executions=True,
+                execution_limit=1,
+            ),
+        )
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in 095821b25: after sorting and slicing the page, when `include_executions=True` the page agents are now re-fetched by ID with the full `MAX_LIBRARY_AGENT_EXECUTIONS_FETCH` limit. Metrics (`execution_count`, `success_rate`, `avg_correctness_score`, `status`) are now computed from the complete execution history, not the single sort-key execution.

autogpt_platform/backend/backend/api/features/library/db.py
majdyz: 🤖 Fixed in 095821b25: after slicing the sorted page, a second `find_many` re-fetches the page agents by ID with `execution_limit=MAX_LIBRARY_AGENT_EXECUTIONS_FETCH`, restoring sort order before passing to `from_db`. The sort-only fetch (`execution_limit=1`) is still used to cheaply determine order, but metrics are now computed from full execution data.

autogpt_platform/backend/backend/api/features/library/db_test.py
majdyz: 🤖 Fixed in 095821b25: added `test_list_library_agents_last_executed_metrics_accuracy` which asserts that when `include_executions=True`, `execution_count` reflects all executions (3) and `success_rate` is 66.67% (2/3 COMPLETED), not the values that would come from a single sort-key execution.

majdyz:
🤖 Fixed in 64689f731 — addressing autogpt-reviewer Should Fix items:

**Should Fix #1** (graph.Executions None-safety): Already safe — `db.py:155` checks `graph and graph.Executions and len(...)`, so when `Executions` is `None` the short-circuit prevents the `len()` call. No change needed; added a test to prove it (`test_list_library_agents_last_executed_none_agent_graph`).

**Should Fix #2** (includes.py order_by scope): Acknowledged tradeoff — `updatedAt` ordering is more semantically correct for "most recently active execution" across all callers. No regression detected in downstream usage.

**Should Fix #3** (take ceiling): Added `take=MAX_LIBRARY_AGENTS_LAST_EXECUTED_FETCH` (1000) to the LAST_EXECUTED `find_many` call, bounding worst-case memory consumption.

**Should Fix #4** (db_test.py edge cases): Added three new tests:
- `test_list_library_agents_last_executed_null_updated_at` — verifies `createdAt` fallback when `updatedAt` is `None`
- `test_list_library_agents_last_executed_none_agent_graph` — verifies safe handling when `AgentGraph` is `None`
- `test_list_library_agents_last_executed_pagination` — verifies `total_items`, `total_pages`, and page-2 slice correctness

**Coderabbit Major (includes.py user_id guard)**: Added `if not user_id: raise ValueError("user_id is required")` to `library_agent_include()`.

**Coderabbit Nitpick (inline comment)**: Removed the trailing inline comment from the `order_by` block in `includes.py`.

majdyz (dismissed):

autogpt_platform/backend/backend/data/includes.py
```diff
@@ -109,6 +110,8 @@ def library_agent_include(
     - Listing optimization (no nodes/executions): ~2s for 15 agents vs potential timeouts
     - Unlimited executions: varies by user (thousands of executions = timeouts)
     """
+    if not user_id:
+        raise ValueError("user_id is required")
```
Pwuts: `user_id` is a required param so what is this supposed to do?

Medyan-Naser: Yes `user_id` is a required positional parameter — Python will raise a TypeError if it's not provided. However this check will catch empty strings or "None".

Pwuts: No need for that, because if such an empty value is passed it simply wouldn't return any results.

ntindle:
@copilot resolve the merge conflicts in this pull request + see if this is still needed + use /pr-address skill as needed to address comments/fix ci

Pwuts:
Thanks for submitting this, and sorry we didn't manage to get it merged. I'm closing it now because there seems to be a (performance-wise) better solution (#13666). Still, we really appreciate the effort and I'll try to credit you when we merge this feature. :)

PR #11908, opened 2026-01-30
hotfix(backend): use discriminator for credential matching in run_block

Otto-AGPT:
## Summary
When executing blocks via the chat interface (run_block tool), the credential matching logic was not using the discriminator_mapping to determine the correct provider based on the selected model value.

## Problem
When a user ran `AITextGeneratorBlock` with `model='gpt-4o-mini'`:
1. The code matched against ALL supported providers (ollama, openai, anthropic, etc.)
2. Ollama credentials were always first in the list (since they don't require an API key)
3. So the `FAKE_API_KEY` from Ollama was incorrectly used instead of the actual OpenAI key

This caused the error:
```
Error code: 401 - {'error': {'message': 'Incorrect API key provided: FAKE_API_KEY...
```

## Root Cause
In `run_block.py`, the `_check_block_credentials` method was:
1. Getting `field_info.provider` which is a frozenset of ALL supported providers
2. Matching the FIRST credential whose provider is in that set
3. Not using the discriminator to narrow down to the correct provider

## Fix
- Pass `input_data` to `_check_block_credentials`
- Use `field_info.discriminate(discriminator_value)` to narrow to the correct provider based on the model
- Match credentials only for the discriminated provider

## Testing
This fixes credential mismatch errors when running LLM blocks via CoPilot/chat interface.

### DIFF
--- a/autogpt_platform/backend/backend/api/features/chat/tools/run_block.py
+++ b/autogpt_platform/backend/backend/api/features/chat/tools/run_block.py
@@ -4,6 +4,8 @@ import logging
 from collections import defaultdict
 from typing import Any
 
+from pydantic_core import PydanticUndefined
+
 from backend.api.features.chat.model import ChatSession
 from backend.data.block import get_block
 from backend.data.execution import ExecutionContext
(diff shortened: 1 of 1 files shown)


ntindle:
@claude run the formatter in the backend

ntindle:
@claude update the pr to follow the pr description please my friend

ntindle (dismissed):

PR #11929, opened 2026-02-02
fix(backend): Handle stale RabbitMQ channels on connection drop

Bentlybro:
### Changes 🏗️

Fixes [**AUTOGPT-SERVER-1TN**](https://autoagpt.sentry.io/issues/?query=AUTOGPT-SERVER-1TN) (~39K events since Feb 2025) and related connection issues **6JC/6JD/6JE/6JF** (~6K combined).

#### Problem

When the RabbitMQ TCP connection drops (network blip, server restart, etc.):

1. `connect_robust` (aio_pika) automatically reconnects the underlying AMQP connection
2. But `AsyncRabbitMQ._channel` still references the **old dead channel**
3. `is_ready` checks `not self._channel.is_closed` — but the channel object doesn't know the transport is gone
4. `publish_message` tries to use the stale channel → `ChannelInvalidStateError: No active transport in channel`
5. `@func_retry` retries 5 times, but each retry hits the same stale channel (it passes `is_ready`)

This means every connection drop generates errors until the process is restarted.

#### Fix

**New `_ensure_channel()` helper** that resets stale channels before reconnecting, so `connect()` creates a fresh one instead of short-circuiting on `is_connected`.

**Explicit `ChannelInvalidStateError` handling in `publish_message`:**
1. First attempt uses `_ensure_channel()` (handles normal staleness)
2. If publish throws `ChannelInvalidStateError`, does a full reconnect (resets both `_channel` and `_connection`) and retries once
3. `@func_retry` provides additional retry resilience on top

**Simplified `get_channel()`** to use the same resilient helper.

**1 file changed, 62 insertions, 24 deletions.**

#### Impact
- Eliminates ~39K `ChannelInvalidStateError` Sentry events
- RabbitMQ operations self-heal after connection drops without process restart
- Related transport EOF errors (6JC/6JD/6JE/6JF) should also reduce

### DIFF
--- a/autogpt_platform/backend/backend/data/rabbitmq.py
+++ b/autogpt_platform/backend/backend/data/rabbitmq.py
@@ -1,3 +1,4 @@
+import asyncio
 import logging
 from abc import ABC, abstractmethod
 from enum import Enum
@@ -225,6 +226,10 @@ class SyncRabbitMQ(RabbitMQBase):
 class AsyncRabbitMQ(RabbitMQBase):
     """Asynchronous RabbitMQ client"""
 
+    def __init__(self, config: RabbitMQConfig):
+        super().__init__(config)
+        self._reconnect_lock: asyncio.Lock | None = None
+
     @property
     def is_connected(self) -> bool:
         return bool(self._connection and not self._connection.is_closed)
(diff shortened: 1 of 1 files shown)


ntindle:
## Code Review Notes

Reviewed as an AI agent on behalf of @ntindle.

### Sentry Issue Confirmed
**AUTOGPT-SERVER-1TN**: `ChannelInvalidStateError: No active transport in channel`
- **38,951 events** since Feb 2025
- Still occurring (last seen ~1 hour ago)
- Happens when aio_pika auto-reconnects but channel object is stale

### Analysis

The fix correctly addresses the root cause:

1. **`_reconnect_lock`** — Prevents concurrent reconnection races with asyncio.Lock
2. **Improved `connect()`** — Now checks channel validity, not just connection
3. **`_ensure_channel()`** — Double-checked locking pattern for thread-safe channel recovery
4. **Explicit `ChannelInvalidStateError` handling** — Catches the specific exception, forces full reconnect, retries

The pattern makes sense: `connect_robust` auto-reconnects the AMQP connection but doesn't always recover channels properly. This adds explicit detection and recovery for stale channels.

### Testing Note
This race condition is difficult to reproduce deterministically in a test environment (requires hitting the exact window where connection is restored but channel is stale). The fix follows established patterns for connection recovery and the code logic is sound.

*— Claude (AI agent)*

PR #11943, opened 2026-02-03
feat(copilot): add customize_agent tool for marketplace templates

majdyz:
## Summary

Adds a new copilot tool that allows users to customize marketplace/template agents using natural language before adding them to their library.

This exposes the Agent Generator's `/api/template-modification` endpoint to the copilot, which was previously not available.

## Changes

- **service.py**: Add `customize_template_external` to call Agent Generator's template modification endpoint
- **core.py**: 
  - Add `customize_template` wrapper function
  - Extract `graph_to_json` as a reusable function (was previously inline in `get_agent_as_json`)
- **customize_agent.py**: New tool that:
  - Takes marketplace agent ID (format: `creator/slug`)
  - Fetches template from store via `store_db.get_agent()`
  - Calls Agent Generator for customization
  - Handles clarifying questions from the generator
  - Saves customized agent to user's library
- **__init__.py**: Register the tool in `TOOL_REGISTRY` for auto-discovery

## Usage Flow

1. User searches marketplace: *"Find me a newsletter agent"*
2. Copilot calls `find_agent` → returns `autogpt/newsletter-writer`
3. User: *"Customize that agent to post to Discord instead of email"*
4. Copilot calls:
   ```
   customize_agent(
       agent_id="autogpt/newsletter-writer",
       modifications="Post to Discord instead of sending email"
   )
   ```
5. Agent Generator may ask clarifying questions (e.g., "What Discord channel?")
6. Customized agent is saved to user's library

## Test plan

- [x] Verified tool imports correctly
- [x] Verified tool is registered in `TOOL_REGISTRY`
- [x] Verified OpenAI function schema is valid
- [x] Ran existing tests (`pytest backend/api/features/chat/tools/`) - all pass
- [x] Type checker (`pyright`) passes with 0 errors
- [ ] Manual testing with copilot (requires Agent Generator service)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -180,3 +180,4 @@ autogpt_platform/backend/settings.py
 .claude/settings.local.json
 CLAUDE.local.md
 /autogpt_platform/backend/logs
+.next
--- a/autogpt_platform/backend/backend/api/features/chat/service.py
+++ b/autogpt_platform/backend/backend/api/features/chat/service.py
@@ -3,10 +3,13 @@ import logging
 import time
 from asyncio import CancelledError
 from collections.abc import AsyncGenerator
-from dataclasses import dataclass
-from typing import Any, cast
+from typing import TYPE_CHECKING, Any, cast
 
 import openai
+
+if TYPE_CHECKING:
+    from backend.util.prompt import CompressResult
+
 import orjson
 from langfuse import get_client
 from openai import (
@@ -17,7 +20,6 @@ from openai import (
     RateLimitError,
 )
 from openai.types.chat import (
-    ChatCompletionAssistantMessageParam,
     ChatCompletionChunk,
     ChatCompletionMessageParam,
     ChatCompletionStreamOptionsParam,
@@ -31,7 +33,6 @@ from backend.data.understanding import (
     get_business_understanding,
 )
 from backend.util.exceptions import NotFoundError
-from backend.util.prompt import estimate_token_count
 from backend.util.settings import Settings
 
 from . import db as chat_db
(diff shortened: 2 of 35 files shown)


majdyz:
Fixed in e4421c1ac - removed all sensitive data from `ErrorResponse.details`:

- Removed `{"exception": str(e)}` from fetch/graph error handlers
- Removed user input (`agent_id`, `modifications`) from details
- Removed `service_error` field that could expose internal error messages  
- Added server-side logging for the save_failed case that was missing it
- Sanitized the error message (was exposing raw exception in user-facing message)

All errors are logged server-side for debugging, but only generic messages are returned to clients.

autogpt_platform/backend/backend/api/features/chat/tools/agent_generator/service.py
ntindle, line 108: Check this has the same protections as Request

autogpt_platform/backend/backend/api/features/chat/tools/customize_agent.py
```diff
@@ -0,0 +1,337 @@
...
+    ) -> ToolResponseBase:
+        """Execute the customize_agent tool.
+
+        Flow:
+        1. Parse the agent ID to get creator/slug
+        2. Fetch the template agent from the marketplace
+        3. Call customize_template with the modification request
+        4. Preview or save based on the save parameter
+        """
+        agent_id = kwargs.get("agent_id", "").strip()
```
ntindle, line 105: I don't love the pattern of getting random crap from kwargs

autogpt_platform/backend/backend/api/features/chat/tools/customize_agent.py
```diff
@@ -0,0 +1,337 @@
...
+                        example=q.get("example"),
+                    )
+                    for q in questions
+                    if isinstance(q, dict)
+                ],
+                session_id=session_id,
+            )
+
+        # Result should be the customized agent JSON
+        if not isinstance(result, dict):
```
ntindle, line 271: Can do this with a match case a bit cleaner imo

autogpt_platform/backend/backend/api/features/chat/tools/customize_agent.py
```diff
@@ -0,0 +1,337 @@
...
+
+        customized_agent = result
+
+        agent_name = customized_agent.get(
+            "name", f"Customized {agent_details.agent_name}"
+        )
+        agent_description = customized_agent.get("description", "")
+        nodes = customized_agent.get("nodes")
+        links = customized_agent.get("links")
+        node_count = len(nodes) if isinstance(nodes, list) else 0
```
ntindle, line 287: What else could it be here?

autogpt_platform/backend/backend/api/features/chat/tools/agent_generator/service.py
majdyz: Or check what this client is used for, if it’s a hardcoded api url only then we don’t need it, otherwise we need utility Request (it’s unlikely though)

autogpt_platform/backend/backend/api/features/chat/tools/customize_agent.py
majdyz: I think he hints for using a class type (make sure it’s flexible enough for missing field and more unknown field though)

Otto-AGPT:
Addressing open review comments from @ntindle:

**Re: [httpx client protections](https://github.com/Significant-Gravitas/AutoGPT/pull/11943#discussion_r2761096975)**
Need clarification on what `Request` protections to match. The httpx client has timeout configured via settings. Happy to add safety checks if you specify what's needed.

**Re: [kwargs pattern](https://github.com/Significant-Gravitas/AutoGPT/pull/11943#discussion_r2761098810)**
This follows the `BaseTool._execute` interface pattern. All tools use `**kwargs` for flexibility. The `.get()` calls with defaults handle validation. Changing to explicit params would require updating the base interface.

**Re: [match/case](https://github.com/Significant-Gravitas/AutoGPT/pull/11943#discussion_r2761102983)**
Current if/elif checks both `isinstance(result, dict)` AND `result.get("type") == "error"`. Match/case structural patterns would be more verbose for this combo check. Can refactor if preferred.

**Re: [What else could it be here?](https://github.com/Significant-Gravitas/AutoGPT/pull/11943#discussion_r2761104196)**
Defensive check for malformed responses from the external Agent Generator service. The service should return a dict, but this guards against edge cases (network issues, unexpected response formats).

Otto-AGPT:
**Re: httpx client protections**

Confirmed safe as-is — the URLs are hardcoded from config settings (`AGENTGENERATOR_HOST`/`PORT`), not user-supplied. The `Requests` utility in `backend.util.request` provides SSRF protection for user-controlled URLs, but this is an internal service-to-service call where the endpoints come from deployment config.

No code change needed — leaving as-is per Zamil's review.

PR #11947, opened 2026-02-03
fix(frontend): Handle object values in FileInput component

### LINKED ISSUE #11800: FileInput crashes with TypeError when value is an object instead of string
## Bug Description

The `FileInput` component in the library page (`/library/{id}`) crashes with `TypeError: e.startsWith is not a function` when the input value is an object instead of a string.

## Steps to Reproduce

1. Execute a graph via the external API with file input as an object:
```json
{
  "node_input": {
    "input_image": {
      "name": "image.jpeg",
      "type": "image/jpeg",
      "size": 131147,
      "data": "/9j/4QAW..."
    }
  }
}
```

2. Navigate to `/library/{graph-id}` and view the run
3. The page crashes

## Error

```
TypeError: e.startsWith is not a function
    at getFileLabelFromValue (./src/components/atoms/FileInput/FileInput.tsx:88:13)
    at FileInput (./src/components/atoms/FileInput/FileInput.tsx:245:86)
```

## Expected Behavior

The component should handle object values gracefully, either by:
1. Converting the object to display format
2. Showing a fallback UI
3. Displaying the filename from the object

## Actual Behavior

The page crashes because `getFileLabelFromValue` calls `.startsWith()` on a value that is an object, not a string.

## Root Cause

`FileInput.tsx:88` assumes the value is always a string (URL or data URI), but the external API can accept file objects in different formats.

## Suggested Fix

Add type checking in `getFileLabelFromValue`:

```typescript
function getFileLabelFromValue(value: unknown): string {
  if (typeof value !== 'string') {
    // Handle object format: { name, type, size, data }
    if (value && typeof value === 'object' && 'name' in value) {
      return (value as { name: string }).name;
    }
    return 'Unknown file';
  }
  // existing string handling...
  if (value.startsWith('data:')) { ... }
}
```

## Environment

- External API endpoint: `/external-api/v1/graphs/{id}/execute/{version}`
- Frontend: `/library/{id}` page

Bentlybro:
### Changes 🏗️

Fixes #11800

The `FileInput` component crashed with `TypeError: e.startsWith is not a function` when the `value` prop was an object (from external API calls) instead of a string.

#### Problem

When executing a graph via the external API with file input as an object:
```json
{
  "node_input": {
    "input_image": {
      "name": "image.jpeg",
      "type": "image/jpeg",
      "size": 131147,
      "data": "/9j/4QAW..."
    }
  }
}
```

Navigating to `/library/{graph-id}` to view the run would crash the page because `getFileLabelFromValue()` called `.startsWith()` on the object.

#### Fix

Updated `getFileLabelFromValue()` to:
1. Check if value is an object and extract `name` or `type` properties
2. Add type guard before calling string methods
3. Graceful fallback to "File" for edge cases (null, undefined, empty object)

**1 file changed, 22 insertions, 1 deletion.**

### DIFF
--- a/autogpt_platform/frontend/src/components/atoms/FileInput/FileInput.tsx
+++ b/autogpt_platform/frontend/src/components/atoms/FileInput/FileInput.tsx
@@ -104,7 +104,28 @@ export function FileInput(props: Props) {
     return false;
   }
 
-  const getFileLabelFromValue = (val: string) => {
+  const getFileLabelFromValue = (val: unknown): string => {
+    // Handle object format from external API: { name, type, size, data }
+    if (val && typeof val === "object") {
+      const obj = val as Record<string, unknown>;
+      if (typeof obj.name === "string") {
+        return getFileLabel(obj.name, (obj.type as string) || "");
+      }
+      if (typeof obj.type === "string") {
+        const mimeParts = obj.type.split("/");
+        if (mimeParts.length > 1) {
+          return `${mimeParts[1].toUpperCase()} file`;
+        }
+        return `${obj.type} file`;
+      }
+      return "File";
+    }
+
+    // Handle string values (data URIs or file paths)
+    if (typeof val !== "string") {
+      return "File";
+    }
+
     if (val.startsWith("data:")) {
       const matches = val.match(/^data:([^;]+);/);
       if (matches?.[1]) {


Bentlybro:
Closing in favor of #11948 (direct branch on main repo instead of fork)

PR #11948, opened 2026-02-03
fix(frontend): Handle object values in FileInput component

Bentlybro:
Fixes [#11800](https://github.com/Significant-Gravitas/AutoGPT/issues/11800)

## Problem
The FileInput component crashed with `TypeError: e.startsWith is not a function` when the value was an object (from external API) instead of a string.

## Example Input Object
When using the external API (`/external-api/v1/graphs/{id}/execute/{version}`), file inputs can be passed as objects:

```json
{
  "node_input": {
    "input_image": {
      "name": "image.jpeg",
      "type": "image/jpeg",
      "size": 131147,
      "data": "/9j/4QAW..."
    }
  }
}
```

## Changes
- Updated `getFileLabelFromValue()` to handle object format: `{ name, type, size, data }`
- Added type guards for string vs object values
- Graceful fallback for edge cases (null, undefined, empty object)

## Test cases verified
- Object with name: returns filename
- Object with type only: extracts and formats MIME type
- String data URI: parses correctly
- String file path: extracts extension
- Edge cases: returns "File" fallback


ntindle:
## Before/After Comparison

### ❌ Before (master) - Crash
When the external API passes file inputs as objects (`{ name, type, size, data }`), the component crashes because `getFileLabelFromValue()` assumes the value is always a string and calls `.startsWith()` on it:

```
TypeError: val.startsWith is not a function
    at getFileLabelFromValue
```

fileinput-master-crash

### ✅ After (this PR) - Works
The fix adds type guards to handle object values from the external API:
- `{ name: "image.jpeg", type: "image/jpeg" }` → "JPEG file"
- `{ type: "image/png" }` → "PNG file"
- `{}` → "File" (graceful fallback)

fileinput-object-test

### Test Method
Added a temporary Storybook story that passes object values to FileInput, then compared behavior on master vs this PR branch.

PR #11962, opened 2026-02-04
refactor(copilot): use Pydantic models and match/case in customize_agent

Otto-AGPT:
## Summary

Follow-up to PR #11943 addressing review feedback from @ntindle.

### 1. Typed parameters with Pydantic model

Replaced `kwargs.get()` pattern with a proper Pydantic input model:

```python
class CustomizeAgentInput(BaseModel):
    agent_id: str = ""
    modifications: str = ""
    context: str = ""
    save: bool = True

    @field_validator("agent_id", "modifications", "context", mode="before")
    @classmethod
    def strip_strings(cls, v: Any) -> str:
        return v.strip() if isinstance(v, str) else (v if v is not None else "")
```

Tool now uses: `params = CustomizeAgentInput(**kwargs)`

### 2. Match/case for response handling

Replaced if/elif chain with cleaner pattern matching:

```python
match result_type:
    case "error":
        # handle error
    case "clarifying_questions":
        # handle questions
    case _:
        # handle success (customized agent JSON)
```

### 3. Better code organization

Split monolithic `_execute` into focused helper methods:
- `_handle_customization_result()` - response type handling
- `_save_or_preview_agent()` - final save/preview logic

## Addresses

- [x] "I don't love the pattern of getting random crap from kwargs" (@ntindle)
- [x] "Can do this with a match case a bit cleaner" (@ntindle)
- [x] "What else could it be here?" - removed need for isinstance checks with typed params (@ntindle)

## Test Plan

- [x] Syntax check passes
- [x] black/isort/ruff formatting passes
- [ ] Existing tests pass (CI will verify)

### DIFF
--- a/autogpt_platform/backend/backend/api/features/chat/tools/customize_agent.py
+++ b/autogpt_platform/backend/backend/api/features/chat/tools/customize_agent.py
@@ -3,6 +3,8 @@
 import logging
 from typing import Any
 
+from pydantic import BaseModel, field_validator
+
 from backend.api.features.chat.model import ChatSession
 from backend.api.features.store import db as store_db
 from backend.api.features.store.exceptions import AgentNotFoundError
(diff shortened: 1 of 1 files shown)


Otto-AGPT:
Closing - will add these changes to PR #11959 instead

PR #11971, opened 2026-02-05
feat(copilot): Auto-save binary block outputs to workspace

Otto-AGPT:
## Summary

Auto-saves binary outputs (PNG, JPEG, PDF, SVG) from all block executions to the user's workspace, returning `workspace://` references instead of raw base64 data.

## Problem

When CoPilot executes blocks that produce binary outputs, the LLM would re-type the entire base64 string when saving files:
- **17,000+ output tokens** for a single PDF
- Potential data corruption from LLM truncation/hallucination
- Poor UX — users wait while the LLM slowly types binary data

## Solution

Post-process all block outputs in `run_block.py`:
1. Detect binary content by field name (`png`, `jpeg`, `pdf`, `svg`)
2. Auto-save to workspace with content deduplication
3. Return `workspace://file-id` references
4. Graceful fallback — keeps original data if save fails

## Changes

- **New:** `binary_output_processor.py` — modular binary detection and workspace saving
- **Updated:** `run_block.py` — integrates processor after block execution

## Testing

Verified with code execution blocks producing PDFs and images — output tokens dropped from 17k to <500.

Closes SECRT-1887

### DIFF
--- a/autogpt_platform/backend/backend/api/features/chat/tools/binary_output_processor.py
+++ b/autogpt_platform/backend/backend/api/features/chat/tools/binary_output_processor.py
@@ -44,6 +44,7 @@ async def process_binary_outputs(
 async def _process_item(
     item: Any, wm: WorkspaceManager, block: str, cache: dict
 ) -> Any:
+    """Recursively process an item, handling dicts and lists."""
     if isinstance(item, dict):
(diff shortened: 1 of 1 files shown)


Otto-AGPT:
Closing - I need to follow the implementation plan review process before submitting code.

PR #11977, opened 2026-02-05
fix(backend/chat): Use latest prompt label in non-production environments

Swiftyos:
In non-production environments, the chat service now fetches prompts with the `latest` label instead of the default production-labeled prompt. This makes it easier to test and iterate on prompt changes in dev/staging without needing to promote them to production first.

### Changes 🏗️

- Updated `_get_system_prompt_template()` in chat service to pass `label="latest"` when `app_env` is not `PRODUCTION`
- Production environments continue using the default behavior (production-labeled prompts)

#### For code changes:
- [x] I have made a test plan
  - [x] Verified that in non-production environments, prompts with `latest` label are fetched
  - [x] Verified that production environments still use the default (production) labeled prompts

### DIFF
--- a/autogpt_platform/backend/backend/api/features/chat/service.py
+++ b/autogpt_platform/backend/backend/api/features/chat/service.py
@@ -629,7 +629,7 @@ async def stream_chat_completion(
                     )
                 )
             elif isinstance(chunk, StreamHeartbeat):
-                # Pass through heartbeats to keep SSE connection alive
+                # Pass through heartbeat to keep SSE connection alive
                 yield chunk
             else:
                 logger.error(f"Unknown chunk type: {type(chunk)}", exc_info=True)


autogpt_platform/backend/backend/api/features/chat/service.py
```diff
@@ -618,6 +628,9 @@ async def _update_title():
                         total_tokens=chunk.totalTokens,
                     )
                 )
+            elif isinstance(chunk, StreamHeartbeat):
+                # Pass through heartbeats to keep SSE connection alive
+                yield chunk
```
Torantulino, line 633: Is this change supposed to be in here?

Bentlybro: This is from my pr here https://github.com/Significant-Gravitas/AutoGPT/pull/11928 that was never reviewed or merged yet..

Swiftyos: yes it fixes a minor issue with the backend erroring on stream heartbeats

Bentlybro: @Swiftyos so am i to close my pr then as it wasnt merged and it fixed this exact same issue

PR #12011, opened 2026-02-08
feat(mcp): Add MCP tool block with OAuth, tool discovery, and standard credential integration

majdyz:
## Summary

image

image

image

image

Full-stack MCP (Model Context Protocol) tool block integration that allows users to connect to any MCP server, discover available tools, authenticate via OAuth, and execute tools — all through the standard AutoGPT credential system.

### Backend

- **MCPToolBlock** (`blocks/mcp/block.py`): New block using `CredentialsMetaInput` pattern with optional credentials (`default={}`), supporting both authenticated (OAuth) and public MCP servers. Includes auto-lookup fallback for backward compatibility.
- **MCP Client** (`blocks/mcp/client.py`): HTTP transport with JSON-RPC 2.0, tool discovery, tool execution with robust error handling (type-checked error fields, non-JSON response handling)
- **MCP OAuth Handler** (`blocks/mcp/oauth.py`): RFC 8414 discovery, dynamic per-server OAuth with PKCE, token storage and refresh via `raise_for_status=True`
- **MCP API Routes** (`api/features/mcp/routes.py`): `discover-tools`, `oauth/login`, `oauth/callback` endpoints with credential cleanup, defensive OAuth metadata validation
- **Credential system integration**:
  - `CredentialsMetaInput` model_validator normalizes legacy `"ProviderName.MCP"` format from Python 3.13's `str(StrEnum)` change
  - `CredentialsFieldInfo.combine()` supports URL-based credential discrimination (each MCP server gets its own credential entry)
  - `aggregate_credentials_inputs` checks block schema defaults for credential optionality
  - Executor normalizes credential data for both Pydantic and JSON schema validation paths
  - Chat credential matching handles MCP server URL filtering
  - `provider_matches()` helper used consistently for Python 3.13 StrEnum compatibility
- **Pre-run validation**: `_validate_graph_get_errors` now calls `get_missing_input()` for custom block-level validation (MCP tool arguments)
- **Security**: HTML tag stripping loop to prevent XSS bypass, SSRF protection (removed trusted_origins)

### Frontend

- **MCPToolDialog** (`MCPToolDialog.tsx`): Full tool discovery UI — enter server URL, authenticate if needed, browse tools, select tool and configure
- **OAuth popup** (`oauth-popup.ts`): Shared utility supporting cross-origin MCP OAuth flows with BroadcastChannel + localStorage fallback
- **Credential integration**: MCP-specific OAuth flow in `useCredentialsInput`, server URL filtering in `useCredentials`, MCP callback page
- **CredentialsSelect**: Auto-selects first available credential instead of defaulting to "None", credentials listed before "None" in dropdown
- **Node rendering**: Dynamic tool input schema rendering on MCP nodes, proper handling in both legacy and new flow editors
- **Block title persistence**: `customized_name` set at block creation for both MCP and Agent blocks — no fallback logic needed, titles survive save/load reliably
- **Stable credential ordering**: Removed `sortByUnsetFirst` that caused credential inputs to jump when selected

### Tests (~2060 lines)

- Unit tests: block, client, tool execution
- Integration tests: mock MCP server with auth
- OAuth flow tests
- API endpoint tests
- Credential combining/optionality tests
- E2e tests (skipped in CI, run manually)

## Key Design Decisions

1. **Optional credentials via `default={}`**: MCP servers can be public (no auth) or private (OAuth). The `credentials` field has `default={}` making it optional at the schema level, so public servers work without prompting for credentials.

2. **URL-based credential discrimination**: Each MCP server URL gets its own credential entry in the "Run agent" form (via `discriminator="server_url"`), so agents using multiple MCP servers prompt for each independently.

3. **Model-level normalization**: Python 3.13 changed `str(StrEnum)` to return `"ClassName.MEMBER"`. Rather than scattering fixes across the codebase, a Pydantic `model_validator(mode="before")` on `CredentialsMetaInput` handles normalization centrally, and `provider_matches()` handles lookups.

4. **Credential auto-select**: `CredentialsSelect` component defaults to the first available credential and notifies the parent state, ensuring credentials are pre-filled in the "Run agent" dialog without requiring manual selection.

5. **customized_name for block titles**: Both MCP and Agent blocks set `customized_name` in metadata at creation time. This eliminates convoluted runtime fallback logic (`agent_name`, hostname extraction) — the title is persisted once and read directly.

## Test plan

- [x] Unit/integration tests pass (68 MCP + 11 graph = 79 tests)
- [x] Manual: MCP block with public server (DeepWiki) — no credentials needed, tools discovered and executable
- [x] Manual: MCP block with OAuth server (Linear, Sentry) — OAuth flow prompts correctly
- [x] Manual: "Run agent" form shows correct credential requirements per MCP server
- [x] Manual: Credential auto-selects when exactly one matches, pre-selects first when multiple exist
- [x] Manual: Credential ordering stays stable when selecting/deselecting
- [x] Manual: MCP block title persists after save and refresh
- [x] Manual: Agent block title persists after save and refresh (via customized_name)
- [ ] Manual: Shared agent with MCP block prompts new user for credentials


majdyz:
Re: Sentry bot comment about `MCPOAuthHandler` not in `HANDLERS_BY_NAME`:

This is **by design**, not a bug. MCP OAuth handlers have dynamic per-server endpoints (authorize, token, revoke URLs) discovered at runtime via MCP metadata — they can't be singleton-registered like other providers.

Token refresh is already handled correctly in `creds_manager.py`:
```python
if credentials.provider == ProviderName.MCP.value:
    oauth_handler = _create_mcp_oauth_handler(credentials)
else:
    oauth_handler = await _get_provider_oauth_handler(credentials.provider)
```

`_create_mcp_oauth_handler()` reconstructs the handler from metadata stored on the credential during initial auth (token_url, client_id, etc.).

autogpt_platform/backend/backend/api/features/mcp/routes.py
```diff
@@ -0,0 +1,422 @@
...
+                for t in tools
+            ],
+            server_name=(
+                init_result.get("serverInfo", {}).get("name")
+                or urlparse(request.server_url).hostname
+                or "MCP"
+            ),
+            protocol_version=init_result.get("protocolVersion"),
+        )
+    except HTTPClientError as e:
```
majdyz, line 110: catching each exception and reraining it seems to be a useless pattern if the exception is not handled by exception route handler then we should add that exception and set the proper status code. this apply to the rest of the routes

autogpt_platform/frontend/src/app/(platform)/build/components/FlowEditor/nodes/CustomNode/CustomNode.tsx
```diff
@@ -98,6 +101,7 @@ export const CustomNode: React.FC<NodeProps<CustomNode>> = React.memo(
             jsonSchema={preprocessInputSchema(inputSchema)}
             nodeId={nodeId}
             uiType={data.uiType}
+            isMCPWithTool={isMCPWithTool}
```
majdyz, line 104: do we really need special handling like this or can we identify using uiType

autogpt_platform/frontend/src/app/(platform)/build/components/FlowEditor/nodes/CustomNode/useCustomNode.tsx
```diff
@@ -3,6 +3,36 @@ import { CustomNodeData } from "./CustomNode";
 import { BlockUIType } from "../../../types";
 import { useMemo } from "react";
 import { mergeSchemaForResolution } from "./helpers";
+import { SpecialBlockID } from "@/lib/autogpt-server-api";
```
majdyz: Similar to this if everything can be handled using proper uiType we don't need specialBlockID, though it's fine we use specialblockid too

autogpt_platform/backend/backend/api/features/mcp/routes.py
majdyz: Fixed in fbe4c740c — removed the broad `except Exception` catch-and-reraise blocks from all three routes. Now only `HTTPClientError` (for 401/403 → proper status codes) and `MCPClientError` (→ 502) are caught explicitly; unhandled exceptions propagate naturally to FastAPI's default 500 handler.

autogpt_platform/frontend/src/app/(platform)/build/components/FlowEditor/nodes/CustomNode/CustomNode.tsx
majdyz: Good call — the `isMCPWithTool` flag in CustomNode could be replaced by a `BlockUIType.MCP_TOOL` value, which would follow the existing pattern for AGENT, NOTE, WEBHOOK etc. I'll refactor this to use `data.uiType === BlockUIType.MCP_TOOL` instead of the `SpecialBlockID` check. The backend already has `block_type` set — just need to wire it to the frontend's `uiType` enum.

autogpt_platform/frontend/src/app/(platform)/build/components/FlowEditor/nodes/CustomNode/useCustomNode.tsx
majdyz: Agreed — the SpecialBlockID check is inconsistent with how other special blocks are handled. I'll add `BlockUIType.MCP_TOOL` to the enum and use `data.uiType` consistently. The `buildMCPInputSchema` logic in useCustomNode will key off `uiType` instead of `block_id`.

autogpt_platform/frontend/src/lib/autogpt-server-api/client.ts
```diff
@@ -792,6 +793,38 @@ export default class BackendAPI {
...
+
+  async mcpOAuthCallback(
+    code: string,
+    stateToken: string,
+  ): Promise<CredentialsMetaResponse> {
+    return this._request("POST", "/mcp/oauth/callback", {
+      code,
+      state_token: stateToken,
+    });
+  }
```
0ubbe: Shouldn't we use the generated queries instead? Same for the types.

autogpt_platform/frontend/src/components/contextual/CredentialsInput/components/CredentialsSelect/CredentialsSelect.tsx
```diff
@@ -67,28 +74,44 @@ export function CredentialsSelect({
...
+  // Notify parent when defaulting to a credential so the value is captured on submit
+  const hasNotifiedDefault = useRef(false);
+  useEffect(() => {
+    if (hasNotifiedDefault.current) return;
+    if (selectedCredential) return; // Already matched — no need to override
+    if (credentials.length > 0) {
+      hasNotifiedDefault.current = true;
+      onSelectCredential(credentials[0].id);
+    }
+  }, [credentials, selectedCredential, onSelectCredential]);
```
0ubbe: Why is this needed? Seems out of scope of the PR? 💭

autogpt_platform/frontend/src/components/contextual/CredentialsInput/components/CredentialsSelect/CredentialsSelect.tsx
```diff
@@ -67,28 +74,44 @@ export function CredentialsSelect({
...
           onChange={handleValueChange}
           disabled={readOnly}
           className="absolute inset-0 z-10 cursor-pointer opacity-0"
           aria-label={`Select ${displayName} credential`}
         >
+          {credentials.map((credential) => (
+            <option key={credential.id} value={credential.id}>
+              {getCredentialDisplayName(credential, displayName)}
+            </option>
+          ))}
```
0ubbe: Should stay below actually 💭

autogpt_platform/frontend/src/components/contextual/CredentialsInput/CredentialsInput.tsx
```diff
@@ -86,7 +86,7 @@ export function CredentialsInput({
     handleCredentialSelect,
   } = hookData;
 
-  const displayName = toDisplayName(provider);
+  const displayName = (schema as any).display_name || toDisplayName(provider);
```
0ubbe: We should not need `as any` here?

autogpt_platform/frontend/src/components/contextual/CredentialsInput/useCredentialsInput.ts
```diff
@@ -157,111 +171,93 @@ export function useCredentialsInput({
...
+      onSelectCredential({
+        id: credentialResult.id,
+        type: "oauth2",
+        title: credentialResult.title,
+        provider,
+      });
+    } catch (error) {
+      if (error instanceof Error && error.message === "OAuth flow timed out") {
+        setOAuthError("OAuth flow timed out");
+      } else {
```
0ubbe, line 253: Have we manually tested the existing OAuth flow in detail and ensured that this new logic does not break it?

majdyz:
Re: OAuth safety — I verified that the standard (non-MCP) OAuth flow is behaviorally identical after the refactor:

- `isMCP` is `false` for all non-MCP providers, so the `if (isMCP)` branches are never hit
- `api.oAuthLogin(provider, scopes)` call is unchanged
- `openOAuthPopup` uses the same `postMessage` listener with `"oauth_popup_result"` type and state token validation
- `oAuthCallback(code, state)` and scope checking are identical
- Timeout is the same 5 minutes

The structural change is extracting the popup/listener logic into `openOAuthPopup()` for reuse between standard OAuth and MCP OAuth. The `about:blank` → navigate pattern also avoids popup blockers more reliably.

autogpt_platform/backend/backend/api/features/mcp/routes.py
```diff
@@ -0,0 +1,404 @@
...
+        for cred in mcp_creds:
+            if (
+                isinstance(cred, OAuth2Credentials)
+                and cred.metadata.get("mcp_server_url") == request.server_url
+            ):
+                if best_cred is None or (
+                    (cred.access_token_expires_at or 0)
+                    > (best_cred.access_token_expires_at or 0)
+                ):
+                    best_cred = cred
```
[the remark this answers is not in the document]
majdyz: Fixed in c8de1d6 — changed to (cred.metadata or {}).get(...) in both locations.

majdyz:
Fixed in the next commit — applied the same `(cred.metadata or {}).get(...)` defensive pattern to `block.py:229` in `_auto_lookup_credential`. Same bug as the one in routes.py.

autogpt_platform/backend/backend/blocks/mcp/block.py
```diff
@@ -0,0 +1,300 @@
...
+        try:
+            mgr = IntegrationCredentialsManager()
+            mcp_creds = await mgr.store.get_creds_by_provider(
+                user_id, ProviderName.MCP.value
+            )
+            best: OAuth2Credentials | None = None
+            for cred in mcp_creds:
+                if (
+                    isinstance(cred, OAuth2Credentials)
+                    and cred.metadata.get("mcp_server_url") == server_url
```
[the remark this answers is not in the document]
majdyz: Fixed — applied the same `(cred.metadata or {}).get(...)` defensive pattern here in `_auto_lookup_credential`.

autogpt_platform/frontend/src/app/(platform)/build/components/legacy-builder/MCPToolDialog.tsx
```diff
@@ -0,0 +1,598 @@
...
+}) {
+  const [expanded, setExpanded] = useState(false);
+  const properties = tool.input_schema?.properties ?? {};
+  const required = new Set<string>(tool.input_schema?.required ?? []);
+  const paramNames = Object.keys(properties);
+
+  // Strip XML-like tags and hints from description for cleaner display
+  const cleanDescription = (tool.description ?? "")
+    .replace(/<[^>]+>[^<]*<\/[^>]+>/g, "")
+    .replace(/<[^>]+>/g, "")
```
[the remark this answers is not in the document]
majdyz: False positive — the sanitization uses a `while` loop (`while (prev \!== cleanDescription)`) that re-applies the regex until no more tags remain, which handles nested patterns like `<scr<script>ipt>`.

autogpt_platform/frontend/src/app/(platform)/build/components/legacy-builder/MCPToolDialog.tsx
```diff
@@ -0,0 +1,598 @@
...
+  onSelect: () => void;
+}) {
+  const [expanded, setExpanded] = useState(false);
+  const properties = tool.input_schema?.properties ?? {};
+  const required = new Set<string>(tool.input_schema?.required ?? []);
+  const paramNames = Object.keys(properties);
+
+  // Strip XML-like tags and hints from description for cleaner display
+  const cleanDescription = (tool.description ?? "")
+    .replace(/<[^>]+>[^<]*<\/[^>]+>/g, "")
```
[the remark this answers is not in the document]
majdyz: Same as above — the loop-based sanitization handles this. False positive.

PR #12016, opened 2026-02-09
refactor(backend): extract shared auto-credential parsing to utils.py

Otto-AGPT:
## Summary
Addresses review feedback from @majdyz on #12004.

## Changes
- Added `AutoCredentialFieldInfo` dataclass and `parse_auto_credential_field()` helper to `executor/utils.py`
- Updated `_acquire_auto_credentials` in `manager.py` to use shared helper
- Updated `_validate_node_input_credentials` in `utils.py` to use shared helper

## Why
The auto-credential field parsing logic (checking `_credentials_id`, extracting provider/file_name for error messages) was duplicated between:
- `manager.py`: credential acquisition with locking
- `utils.py`: credential validation without locking

This consolidates the shared parsing logic while keeping the distinct operations (acquire vs validate) separate. Makes `manager.py` less cluttered as requested.

## Testing
- [x] Format checks pass (`black`, `isort`)
- [x] Type checks pass (`pyright`)

## Related
- Builds on #12004
- cc @majdyz @ntindle


nidhishgajjar:
**Orb Code Review** (powered by GLM 5.1 on [Orb Cloud](https://orbcloud.dev))

## Summary

This PR extracts shared auto-credential parsing logic into a reusable `parse_auto_credential_field()` function and typed `AutoCredentialFieldInfo` model. The logic was previously duplicated between `manager.py` (credential acquisition) and `utils.py` (credential validation).

---

## Architecture

- **`AutoCredentialFieldInfo`**: Pydantic model holding parsed credential data (`cred_id`, `provider`, `file_name`, `field_name`, `error`)
- **`parse_auto_credential_field()`**: Shared function that handles all credential field parsing cases — dict with `_credentials_id`, dict without it (error), field absent from input (upstream connected, skip), field present but None (user didn't select, error)
- **`manager.py`**: Uses shared function for credential acquisition, reducing ~50 lines of inline logic
- **`utils.py`**: Uses shared function for credential validation, reducing duplication

---

## Issues

### Minor behavioral change: `isinstance(cred_id, str)` check removed
**Severity: suggestion**

The old validation code in `utils.py` had:
```python
if cred_id and isinstance(cred_id, str):
```

The new `parse_auto_credential_field()` function just does:
```python
if cred_id:
    result.cred_id = cred_id
```

The `isinstance(cred_id, str)` guard is dropped, meaning non-string credential IDs (e.g., integers from malformed data) would now be passed through instead of silently skipped. This is unlikely to cause issues in practice since `_credentials_id` is always a string UUID, but worth noting.

### `field_present_in_input=True` hardcoded in validation path
**Severity: suggestion**

In `utils.py`, the call uses `field_present_in_input=True` with the comment "For validation, assume present". This means if a field is connected from an upstream block and the value happens to be None, it would generate a "No file selected" error during validation. This should be safe since `get_auto_credentials_fields()` only returns fields that need credential resolution (not upstream-connected ones), but the hardcoded assumption makes the code slightly harder to reason about. Consider passing the actual presence flag if it's available in the validation context.

---

## Assessment: ✅ Approve

This is a clean, well-structured refactoring that eliminates real code duplication. The `AutoCredentialFieldInfo` Pydantic model is a nice touch that provides type safety and self-documenting return values. Error messages are now consistent between the acquisition and validation paths. The minor behavioral differences are unlikely to cause issues.

_This review was created by an AI assistant (OpenHands) on behalf of the Orb Cloud review agent._

PR #12017, opened 2026-02-09
fix(backend): Auto-correct content-type based on file signature instead of erroring

Otto-AGPT:
## Summary
Automatically corrects mismatched content-types for uploaded files by detecting the actual file type from the file signature (magic bytes).

## Problem
Users were getting errors when uploading files with incorrect content-type headers. This happens often when files are downloaded from external sources or when browsers/clients set incorrect MIME types.

## Solution
Instead of erroring when content-type doesn't match, we now:
1. Detect the actual file type using magic bytes
2. Auto-correct the content-type if it doesn't match
3. Log a warning for debugging

## Changes
- Added file signature detection to `file.py`
- Updated upload handling to auto-correct content-types
- Added tests for the new behavior

## Testing
- [x] Unit tests pass
- [x] Manual testing with various file types

---
*Recreated from #11882 (fork PR) to use main repo branch for CI secrets access*


Swiftyos (dismissed):
LGTM much cleaner thank you

0ubbe (dismissed):

0ubbe (approved):

PR #12020, opened 2026-02-09
refactor(backend): optimize find_block response size by removing raw JSON schemas

Swiftyos:
### Changes 🏗️

The `find_block` AutoPilot tool was returning ~90K characters per response (10 blocks). The bloat came from including full JSON Schema objects (`input_schema`, `output_schema`) with all nested `$defs`, `anyOf`, and type definitions for every block.

**What changed:**

- **`BlockInfoSummary` model**: Removed `input_schema` (raw JSON Schema), `output_schema` (raw JSON Schema), and `categories`. Added `output_fields` (compact field-level summaries matching the existing `required_inputs` format).
- **`BlockListResponse` model**: Removed `usage_hint` (info now in `message`).
- **`FindBlockTool._execute()`**: Now extracts compact `output_fields` from output schema properties instead of including the entire raw schema. Credentials handling is unchanged.
- **Test**: Added `test_response_size_average_chars_per_block` with realistic block schemas (HTTP, Email, Claude Code) to measure and assert response size stays under 2K chars/block.
- **`CLAUDE.md`**: Clarified `dev` vs `master` branching strategy.

**Result:** Average response size reduced from ~9,000 to ~1,300 chars per block (~85% reduction). This directly reduces LLM token consumption, latency, and API costs for AutoPilot interactions.

#### For code changes:
- [x] I have made a test plan
  - [x] Verified models import and serialize correctly
  - [x] Verified response size: 3,970 chars for 3 realistic blocks (avg 1,323/block)
  - [x] Lint (`ruff check`) and type check (`pyright`) pass on changed files
  - [x] Frontend compatibility preserved: `blocks[].name` and `count` fields retained for `block_list` handler

### DIFF
--- a/autogpt_platform/backend/backend/api/features/chat/tools/run_block.py
+++ b/autogpt_platform/backend/backend/api/features/chat/tools/run_block.py
@@ -27,6 +27,7 @@ from .models import (
     BlockDetailsResponse,
     BlockOutputResponse,
     ErrorResponse,
+    InputValidationErrorResponse,
     SetupInfo,
     SetupRequirementsResponse,
     ToolResponseBase,
@@ -232,6 +233,22 @@ class RunBlockTool(BaseTool):
         required_keys = set(input_schema.get("required", []))
         required_non_credential_keys = required_keys - credentials_fields
         provided_input_keys = set(input_data.keys()) - credentials_fields
+
+        # Check for unknown input fields
+        valid_fields = (
+            set(input_schema.get("properties", {}).keys()) - credentials_fields
+        )
+        unrecognized_fields = provided_input_keys - valid_fields
+        if unrecognized_fields:
+            return InputValidationErrorResponse(
+                message=(
+                    f"Unknown input field(s) provided: {', '.join(sorted(unrecognized_fields))}. "
+                    f"Block was not executed. Please use the correct field names from the schema."
+                ),
+                session_id=session_id,
+                unrecognized_fields=sorted(unrecognized_fields),
+                inputs=input_schema,
+            )
 
         # Show details when not all required non-credential inputs are provided
         if not (required_non_credential_keys <= provided_input_keys):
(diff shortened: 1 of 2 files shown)


autogpt_platform/backend/backend/api/features/chat/tools/find_block.py
```diff
@@ -54,7 +53,8 @@ def description(self) -> str:
             "Blocks are reusable components that perform specific tasks like "
             "sending emails, making API calls, processing text, etc. "
             "IMPORTANT: Use this tool FIRST to get the block's 'id' before calling run_block. "
-            "The response includes each block's id, required_inputs, and input_schema."
+            "The response includes each block's id, name, and description. "
+            "Call run_block with the block's id to see detailed inputs/outputs and execute it."
```
Torantulino:
```suggestion
            "Call run_block with the block's id **with no inputs** to see detailed inputs/outputs and execute it."
```

autogpt_platform/backend/backend/api/features/chat/tools/find_block.py
```diff
@@ -227,8 +173,7 @@ async def _execute(
             return BlockListResponse(
                 message=(
                     f"Found {len(blocks)} block(s) matching '{query}'. "
-                    "To execute a block, use run_block with the block's 'id' field "
-                    "and provide 'input_data' matching the block's input_schema."
+                    "To see a block's inputs/outputs and execute it, use run_block with the block's 'id'."
```
Torantulino:
```suggestion
                    "To see a block's inputs/outputs and execute it, use run_block with the block's 'id' - providing no inputs."
```

Torantulino:
What happens when incorrect inputs are provided? 

There was a silent failure bug on the `run_agent` tool previously where incorrect input keys would just be silently ignored, and the agent would run with default inputs.

ntindle:
> What happens when incorrect inputs are provided?

> There was a silent failure bug on the run_agent tool previously where incorrect input keys would just be silently ignored, and the agent would run with default inputs.

also do you have an example of what the data being passed will look like?

Swiftyos:
Screenshot 2026-02-13 at 09 25 20

Torantulino (approved):

PR #12028, opened 2026-02-09
fix(blocks): allow falsy entries in AddToListBlock

Deeven-Seru:
## Summary
- treat AddToListBlock.entry as optional rather than truthy so 0/""/False are appended
- extend block self-tests with a falsy entry case

## Testing
- Not run (pytest not available in environment)


ntindle:
Hey hey! 

Can you give a bit of context on why this is desired + fill out the pr template

Deeven-Seru:
> Hey hey!
> 
> Can you give a bit of context on why this is desired + fill out the pr template

the transition to explicit None validation ensures the block remains data-agnostic. In agentic workflows, falsy values—such as 0, False, or empty strings—frequently represent valid data points or state transitions. The previous truthiness-based implementation silently discarded these values; this update preserves data stream integrity by ensuring only None is excluded.

ntindle:
@DEVELOPER-DEEVEN you have a commit in here with DevX as the author or associated. Please ensure they sign the cla

ntindle (approved):

Pwuts (approved):

PR #12048, opened 2026-02-09
Handle errors in Jina ExtractWebsiteContentBlock

Deeven-Seru:
## Summary
- catch Jina reader client/server errors in ExtractWebsiteContentBlock and surface a clear error output keyed to the user URL
- guard empty responses to return an explicit error instead of yielding blank content
- add regression tests covering the happy path and HTTP client failures via a monkeypatched fetch

## Testing
- not run (pytest unavailable in this environment)


Deeven-Seru:
Tests are failing in CI. Checking logs and identifying fixes for the logic changes in this branch.

PR #12070, opened 2026-02-11
fix(copilot): stream timeout, long-running tool polling, and CreateAgent UI refresh

0ubbe:
Agent generation completes on the backend but the UI does not update/refresh to show the result.

### Changes 🏗️

- **Stream start timeout (12s):** If the backend doesn't begin streaming within 12 seconds of submitting a message, the stream is aborted and a destructive toast is shown to the user.
- **Long-running tool polling:** Added `useLongRunningToolPolling` hook that polls the session endpoint every 1.5s while a tool output is in an operating state (`operation_started` / `operation_pending` / `operation_in_progress`). When the backend completes, messages are refreshed so the UI reflects the final result.
- **CreateAgent UI improvements:** Replaced the orbit loader / progress bar with a mini-game, added expanded accordion for saved agents, and improved the saved-agent card with image, icons, and links that open in new tabs.
- **Backend tweaks:** Added `image_url` to `CreateAgentToolOutput`, minor model/service updates for the dummy agent generator.

#### For code changes:
- [x] I have made a test plan
  - [x] Send a message and verify the stream starts within 12s or a toast appears
  - [x] Trigger agent creation and verify the UI updates when the backend completes
  - [x] Verify the saved-agent card renders correctly with image, links, and icons

### DIFF
--- a/autogpt_platform/backend/test/agent_generator/test_service.py
+++ b/autogpt_platform/backend/test/agent_generator/test_service.py
@@ -25,6 +25,7 @@ class TestServiceConfiguration:
         """Test that external service is not configured when host is empty."""
         mock_settings = MagicMock()
         mock_settings.config.agentgenerator_host = ""
+        mock_settings.config.agentgenerator_use_dummy = False
 
         with patch.object(service, "_get_settings", return_value=mock_settings):
             assert service.is_external_service_configured() is False


autogpt_platform/backend/backend/api/features/chat/tools/agent_generator/service.py
```diff
@@ -90,10 +101,26 @@ def _get_settings() -> Settings:
     return _settings
 
 
+def _is_dummy_mode() -> bool:
+    """Check if dummy mode is enabled for testing."""
+    global _dummy_mode_warned
+    settings = _get_settings()
+    is_dummy = bool(settings.config.agentgenerator_use_dummy)
```
ntindle, line 108: If you want you could check our flags here for environment and straight up crash the backend if enabled so we don't scale a deployment accident with this

PR #12113, opened 2026-02-14
fix(builder): Auto-cleanup invalid/orphan edges during graph operations

Otto-AGPT:
## Summary

Fixes graph desync issues where edge deletions don't persist, causing stale connections that users cannot remove. The fix auto-removes invalid edges at multiple points to ensure graphs stay clean.

## Problem

Edge deletion in the Agent Builder is not persisting correctly:
- When users delete edges between nodes, the deletion sometimes doesn't stick
- This causes stale/orphan node connections across builds that cannot be corrected
- Users have no way to remove these invalid edges from the UI

### Frontend Changes

**`edgeStore.ts`:**
- `addLinks()`: Validate links during load - skip edges referencing non-existent nodes
- `getBackendLinks()`: Filter invalid edges before sending to backend during save

**`useSaveGraph.ts`:**
- After successful save, sync edge store with authoritative backend state
- This prevents desync between frontend and backend

### Backend Changes

**`graph.py`:**
- Add `prune_invalid_links()` method that removes:
  - Links referencing non-existent nodes
  - Links with invalid block IDs
- Called during `_validate_graph_structure()` to auto-cleanup orphan edges before save

## Testing

1. Create agent with multiple nodes and edges
2. Delete a node (edges should auto-cleanup)
3. Save and reload - no orphan edges
4. Delete an edge directly, save, reload - edge stays deleted

## Related Issues

Closes SECRT-1959

<h2>Greptile Overview</h2>

<details><summary><h3>Greptile Summary</h3></summary>

This PR implements comprehensive edge cleanup to fix persistent desync issues where deleted edges would reappear after save/reload. The fix adds validation at three key points:

**Backend (`graph.py`):**
- Added `prune_invalid_links()` method to remove orphan edges referencing non-existent nodes or blocks
- Auto-cleanup runs during graph validation before save

**Frontend (`edgeStore.ts`):**
- `getBackendLinks()` now filters out invalid edges before sending to backend
- `addLinks()` validates links during load and skips edges with missing nodes

**Sync Logic (`useSaveGraph.ts`):**
- After successful save, syncs edge store with authoritative backend state using `syncEdgesWithBackend()`
- Prevents frontend/backend desync by replacing local edges with backend-approved edges

The changes implement defense-in-depth by validating at multiple layers (frontend pre-save, backend validation, frontend post-save sync, frontend load). This ensures orphan edges are caught and removed regardless of where they originate.
</details>

<details><summary><h3>Confidence Score: 4/5</h3></summary>

- This PR is safe to merge with minimal risk - addresses a real UX issue with a defensive multi-layer approach
- The implementation is sound with validation at multiple layers (frontend filter, backend pruning, post-save sync). The changes are focused and don't introduce breaking changes. One minor documentation issue was found (docstring mentions pin validation that isn't implemented), but this doesn't affect functionality. No tests were added for the new `prune_invalid_links()` method, though the existing validation flow has test coverage.
- No files require special attention - all changes are straightforward defensive checks
</details>

<details><summary><h3>Sequence Diagram</h3></summary>

```mermaid
sequenceDiagram
    participant User
    participant Frontend
    participant edgeStore
    participant useSaveGraph
    participant Backend
    participant graph.py

    User->>Frontend: Delete node/edge
    Frontend->>edgeStore: Update local state
    
    User->>Frontend: Click Save
    Frontend->>useSaveGraph: saveGraph()
    useSaveGraph->>edgeStore: getBackendLinks()
    edgeStore->>edgeStore: Filter invalid edges<br/>(check node existence)
    edgeStore-->>useSaveGraph: Valid links only
    
    useSaveGraph->>Backend: POST/PUT graph with links
    Backend->>graph.py: _validate_graph_structure()
    graph.py->>graph.py: prune_invalid_links()<br/>(remove orphan edges)
    graph.py->>graph.py: Validate remaining links
    Backend-->>useSaveGraph: Return cleaned graph
    
    useSaveGraph->>useSaveGraph: syncEdgesWithBackend()
    useSaveGraph->>edgeStore: setEdges(backend links)
    edgeStore->>edgeStore: Replace all edges
    
    User->>Frontend: Reload page
    Frontend->>Backend: GET graph
    Backend-->>Frontend: Return graph with links
    Frontend->>edgeStore: addLinks(links)
    edgeStore->>edgeStore: Filter invalid links<br/>(check node existence)
    edgeStore->>edgeStore: Add valid edges only
```
</details>

<sub>Last reviewed commit: 6cb794c</sub>

### DIFF
--- a/autogpt_platform/frontend/src/app/(platform)/build/hooks/useSaveGraph.ts
+++ b/autogpt_platform/frontend/src/app/(platform)/build/hooks/useSaveGraph.ts
@@ -13,7 +13,6 @@ import { Graph } from "@/app/api/__generated__/models/graph";
 import { useNodeStore } from "../stores/nodeStore";
 import { useEdgeStore } from "../stores/edgeStore";
 import { graphsEquivalent } from "../components/NewControlPanel/NewSaveControl/helpers";
-import { linkToCustomEdge } from "../components/helper";
 import { useGraphStore } from "../stores/graphStore";
 import { useShallow } from "zustand/react/shallow";
 import {
@@ -21,18 +20,6 @@ import {
   clearTempFlowId,
   getTempFlowId,
 } from "@/services/builder-draft/draft-service";
-
-/**
- * Sync the edge store with the authoritative backend state.
- * This ensures the frontend matches what the backend accepted after save.
- */
-function syncEdgesWithBackend(links: GraphModel["links"]) {
-  if (links !== undefined) {
-    // Replace all edges with the authoritative backend state
-    const newEdges = links.map(linkToCustomEdge);
-    useEdgeStore.getState().setEdges(newEdges);
-  }
-}
 
 export type SaveGraphOptions = {
   showToast?: boolean;
@@ -77,9 +64,6 @@ export const useSaveGraph = ({
             flowVersion: data.version,
           });
 
-          // Sync edge store with authoritative backend state
-          syncEdgesWithBackend(data.links);
-
           const tempFlowId = getTempFlowId();
           if (tempFlowId) {
             await draftService.deleteDraft(tempFlowId);
@@ -116,9 +100,6 @@ export const useSaveGraph = ({
             flowID: data.id,
             flowVersion: data.version,
           });
-
-          // Sync edge store with authoritative backend state
-          syncEdgesWithBackend(data.links);
 
           // Clear the draft for this flow after successful save
           if (data.id) {
--- a/autogpt_platform/frontend/src/app/(platform)/build/stores/edgeStore.ts
+++ b/autogpt_platform/frontend/src/app/(platform)/build/stores/edgeStore.ts
@@ -7,6 +7,7 @@ import { NodeExecutionResult } from "@/app/api/__generated__/models/nodeExecutio
 import { cleanUpHandleId } from "@/components/renderers/InputRenderer/helpers";
 import { useHistoryStore } from "./historyStore";
 import { useNodeStore } from "./nodeStore";
+import { filterValidEdges, filterValidLinks } from "./linkValidations";
 
 type EdgeStore = {
   edges: CustomEdge[];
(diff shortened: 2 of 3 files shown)


Otto-AGPT:
**Re: Sentry Bug Prediction (comment r2807474296)**

Good catch on the history pollution issue with `addLinks`, but this is pre-existing behavior in the codebase, not introduced by this PR. 

My changes in this PR actually **improve** the situation:
- `syncEdgesWithBackend()` in `useSaveGraph.ts` uses `setEdges()` for bulk replacement, which doesn't trigger individual history entries
- The `getBackendLinks()` filter I added only affects what gets sent to the backend, not history

The `addLinks` history pollution is a valid issue but should be addressed in a separate PR that refactors the history initialization flow.

Otto-AGPT:
Fixed the history pollution issue raised by Sentry. Changed `addLinks` to bulk-add edges with a single history push instead of calling `addEdge` individually for each link.

**Before:** Each `addEdge` call → separate history entry → N entries for N edges
**After:** Collect all edges → single `set()` call → single history entry

Otto-AGPT:
Fixed the race condition identified by Sentry.

**Problem:** Two separate `useEffect` hooks were running independently:
- One for adding nodes (waits for `customNodes` which depends on `blocks` API)
- One for adding links (runs as soon as `graph.links` is available)

Since links can be available before nodes, `addLinks` would execute and filter out ALL links because their source/sink nodes didn't exist yet.

**Fix:** Combined both into a single `useEffect` that:
1. First adds nodes to the store
2. Then adds links (only after nodes are present)

This ensures the correct ordering and prevents silent connection loss.

Otto-AGPT:
**Re: Sentry Bug Prediction (comment r2807845064) - Edge changes lost during save**

This is a valid UX concern but out of scope for this PR. The issue exists because:
1. Save is async and takes time
2. User can continue editing during save
3. `syncEdgesWithBackend` replaces edges with backend state on success

The suggested fix (locking the entire editor during save) is a significant UX change that should be addressed in a dedicated PR. It requires:
- Global lock state for the flow editor
- Disabling node/edge interactions
- Visual indicator that save is in progress

For now, the save operation is intentionally non-blocking to keep the editor responsive. If edge loss during save becomes a user-reported issue, we can prioritize the locking feature.

autogpt_platform/frontend/src/app/(platform)/build/hooks/useSaveGraph.ts
Abhi1992002, line 1: I think we can remove the changes from this file — this will create a race condition with `useEffect` in `useFlow.ts`, and there’s no need for it tbh.

autogpt_platform/frontend/src/app/(platform)/build/stores/edgeStore.ts
Abhi1992002, line 1: I think we can create a new file called `linkValidations.ts` and add these two validations there: `filterValidEdges()` and `filterValidLinks()`.
but in the best-case scenario, our backend should not send a broken agent to the frontend.

autogpt_platform/frontend/src/app/(platform)/build/hooks/useSaveGraph.ts
Otto-AGPT: Done! Removed `syncEdgesWithBackend` and all its call sites. The backend-returned data still flows through the query cache and triggers the `useFlow.ts` useEffect naturally, so this was indeed redundant.

autogpt_platform/frontend/src/app/(platform)/build/stores/edgeStore.ts
Otto-AGPT: Done! Created `linkValidations.ts` with `filterValidEdges()` and `filterValidLinks()`, and updated `edgeStore.ts` to use them.

Agreed on the backend point — the backend's `prune_invalid_links()` in `graph.py` should catch these before they reach the frontend. The frontend validation is defense-in-depth for edge cases (e.g., concurrent edits, stale cache).

PR #12118, opened 2026-02-16
ci(backend): Upgrade RabbitMQ from 3.12 (EOL) to 4.1.4

Otto-AGPT:
## Summary
Upgrades RabbitMQ from the end-of-life `rabbitmq:3.12-management` to `rabbitmq:4.1.4`, aligning CI, local dev, and e2e testing with production.

### CI Workflow (`.github/workflows/platform-backend-ci.yml`)
- **Image:** `rabbitmq:3.12-management` → `rabbitmq:4.1.4`
- **Port:** Removed 15672 (management UI) — not used
- **Health check:** Added to prevent flaky tests from race conditions during startup

### Docker Compose (`docker-compose.platform.yml`, `docker-compose.test.yaml`)
- **Image:** `rabbitmq:management` → `rabbitmq:4.1.4`
- **Port:** Removed 15672 (management UI) — not used

## Why
- RabbitMQ 3.12 is EOL
- We don't use the management interface, so `-management` variant is unnecessary
- CI and local dev/e2e should match production (4.1.4)

## Testing
CI validates that backend tests pass against RabbitMQ 4.1.4 on Python 3.11, 3.12, and 3.13.

---
Closes SECRT-1703

### DIFF
--- a/autogpt_platform/backend/docker-compose.test.yaml
+++ b/autogpt_platform/backend/docker-compose.test.yaml
@@ -53,7 +53,7 @@ services:
 
   rabbitmq:
     <<: *agpt-services
-    image: rabbitmq:management
+    image: rabbitmq:4.1.4
     container_name: rabbitmq
     healthcheck:
       test: rabbitmq-diagnostics -q ping
@@ -66,7 +66,6 @@ services:
       - RABBITMQ_DEFAULT_PASS=k0VMxyIJF9S35f3x2uaw5IWAl6Y536O7
     ports:
       - "5672:5672"
-      - "15672:15672"
   clamav:
     image: clamav/clamav-debian:latest
     ports:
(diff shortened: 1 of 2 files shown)


Otto-AGPT:
**Addressing review feedback:**

**Health check (greptile comment #1):** ✅ Added in df536789c. Good catch — this brings CI in line with the docker-compose files.

**Version mismatch (greptile comment #2):** This is intentionally out of scope per the ticket (SECRT-1703), which specifically targets the CI workflow. The docker-compose files use `rabbitmq:management` (unversioned/latest) rather than a pinned EOL version like CI had. If we want to pin those too, that would be a separate discussion/ticket.

Otto-AGPT:
**Update:** Per Pwuts' feedback, also pinned the docker-compose files to `rabbitmq:4.1.4` (commit e142ec1dd):
- `docker-compose.platform.yml`
- `docker-compose.test.yaml`

Removed the management UI port (15672) from both as well, since we're not using the management interface.

PR #12148, opened 2026-02-17
fix(backend/copilot): Route workspace through db_accessors, fix transcript upload

Otto-AGPT:
## Summary

Fixes two bugs in the copilot executor:

### SECRT-2008: WorkspaceManager bypasses db_accessors
`backend/util/workspace.py` imported 6 workspace functions directly from `backend/data/workspace.py`, which call `prisma()` directly. In the copilot executor (no Prisma connection), these fail.

**Fix:** Replace direct imports with `workspace_db()` from `db_accessors`, routing through the database_manager HTTP client when Prisma is unavailable. Also:
- Register all 6 workspace functions in `DatabaseManager` and `DatabaseManagerAsyncClient`
- Add `UniqueViolationError` to the service `EXCEPTION_MAPPING` so it is properly re-raised over HTTP (needed for race-condition retry logic)
- Add `UniqueViolationError` FastAPI exception handler (returns 400 instead of falling through to 500)
- Add `UniqueViolationError` to retry `exclude_exceptions` (no point retrying constraint violations)
- Bind `workspace_db()` to a local variable once per method for consistent connection state
- Make `workspace_id` required (not optional) on `get_workspace_file` and `soft_delete_workspace_file` to enforce workspace scoping at the data layer
- Use `UserWorkspaceFileWhereInput` type instead of bare `dict` for where clauses

### SECRT-2009: Transcript upload asyncio.timeout error
`asyncio.create_task()` at line 696 of `service.py` creates an orphaned background task in the executor's thread event loop. `gcloud-aio-storage`'s `asyncio.timeout()` fails in this context.

**Fix:** Replace `create_task` with direct `await`, wrapped in a 30s timeout to prevent the SSE connection from hanging on slow GCS uploads. The upload runs after streaming completes (all chunks already yielded), so no user-facing latency impact. The function already has internal try/except error handling.

### DIFF
--- a/autogpt_platform/backend/backend/data/workspace.py
+++ b/autogpt_platform/backend/backend/data/workspace.py
@@ -164,21 +164,23 @@ async def create_workspace_file(
 
 async def get_workspace_file(
     file_id: str,
-    workspace_id: Optional[str] = None,
+    workspace_id: str,
 ) -> Optional[WorkspaceFile]:
     """
     Get a workspace file by ID.
 
     Args:
         file_id: The file ID
-        workspace_id: Optional workspace ID for validation
+        workspace_id: Workspace ID for scoping (required)
 
     Returns:
         WorkspaceFile instance or None
     """
-    where_clause: dict = {"id": file_id, "isDeleted": False}
-    if workspace_id:
-        where_clause["workspaceId"] = workspace_id
+    where_clause: UserWorkspaceFileWhereInput = {
+        "id": file_id,
+        "isDeleted": False,
+        "workspaceId": workspace_id,
+    }
 
     file = await UserWorkspaceFile.prisma().find_first(where=where_clause)
     return WorkspaceFile.from_db(file) if file else None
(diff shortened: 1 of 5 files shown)


Pwuts (dismissed):
LGTM

autogpt_platform/backend/backend/data/workspace.py
```diff
@@ -164,21 +164,23 @@ async def create_workspace_file(
...
-        workspace_id: Optional workspace ID for validation
+        workspace_id: Workspace ID for scoping (required)
 
     Returns:
         WorkspaceFile instance or None
     """
-    where_clause: dict = {"id": file_id, "isDeleted": False}
-    if workspace_id:
-        where_clause["workspaceId"] = workspace_id
+    where_clause: dict = {
```
Pwuts:
```suggestion
    where_clause: UserWorkspaceFileWhereInput = {
```

PR #12153, opened 2026-02-18
fix(backend/copilot): Clean up GCSWorkspaceStorage per worker

Otto-AGPT:
The copilot executor runs each worker in its own thread with a dedicated event loop (`asyncio.new_event_loop()`). `aiohttp.ClientSession` is bound to the event loop where it was created — using it from a different loop causes `asyncio.timeout()` to fail with:

```
RuntimeError: Timeout context manager should be used inside a task
```

This was the root cause of transcript upload failures tracked in SECRT-2009 and [Sentry #7272473694](https://significant-gravitas.sentry.io/issues/7272473694/).

### Fix

**One `GCSWorkspaceStorage` instance per event loop** instead of a single shared global.

- `get_workspace_storage()` now returns a per-loop GCS instance (keyed by `id(asyncio.get_running_loop())`). Local storage remains shared since it has no async I/O.
- `shutdown_workspace_storage()` closes the instance for the **current** loop only, so `session.close()` always runs on the loop that created the session.
- `CoPilotProcessor.cleanup()` shuts down workspace storage on the worker's own loop, then stops the loop.
- Manager cleanup submits `cleanup_worker` to each thread pool worker before shutting down the executor — replacing the old approach of creating a temporary event loop that couldn't close cross-loop sessions.

### Changes

| File | Change |
|------|--------|
| `util/workspace_storage.py` | `GCSWorkspaceStorage` back to simple single-session class; `get_workspace_storage()` returns per-loop GCS instance; `shutdown_workspace_storage()` scoped to current loop |
| `copilot/executor/processor.py` | Added `CoPilotProcessor.cleanup()` and `cleanup_worker()` |
| `copilot/executor/manager.py` | Calls `cleanup_worker` on each thread pool worker during shutdown |

Fixes SECRT-2009

### DIFF
--- a/autogpt_platform/backend/backend/util/workspace_storage.py
+++ b/autogpt_platform/backend/backend/util/workspace_storage.py
@@ -95,32 +95,27 @@ class WorkspaceStorageBackend(ABC):
 class GCSWorkspaceStorage(WorkspaceStorageBackend):
     """Google Cloud Storage implementation for workspace storage.
 
-    ## Event-loop-aware
-    Each event loop gets its own `aiohttp` session and GCS
-    client. This is required because `aiohttp.ClientSession` (and the
-    `asyncio.timeout()` calls it makes internally) are bound to the loop they
-    were created on. In the copilot executor every worker thread runs its
-    own event loop, so a single shared session would fail with
-    "Timeout context manager should be used inside a task".
+    Each instance owns a single ``aiohttp.ClientSession`` and GCS async
+    client.  Because ``ClientSession`` is bound to the event loop on which it
+    was created, callers that run on separate loops (e.g. copilot executor
+    worker threads) **must** obtain their own ``GCSWorkspaceStorage`` instance
+    via :func:`get_workspace_storage` which is event-loop-aware.
     """
 
     def __init__(self, bucket_name: str):
         self.bucket_name = bucket_name
-        # Keyed by id(event_loop) → per-loop client/session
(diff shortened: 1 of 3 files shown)


autogpt_platform/backend/backend/util/workspace_storage.py
```diff
@@ -93,22 +93,35 @@ async def get_download_url(self, storage_path: str, expires_in: int = 3600) -> s
...
 class GCSWorkspaceStorage(WorkspaceStorageBackend):
-    """Google Cloud Storage implementation for workspace storage."""
+    """Google Cloud Storage implementation for workspace storage.
+
+    Event-loop-aware: each event loop gets its own aiohttp session and GCS
+    client.  This is required because aiohttp.ClientSession (and the
+    asyncio.timeout() calls it makes internally) are bound to the loop they
+    were created on.  In the copilot executor every worker thread runs its
+    own event loop, so a single shared session would fail with
+    "Timeout context manager should be used inside a task".
```
Pwuts:
```suggestion
    ## Event-loop-aware
    Each event loop gets its own `aiohttp` session and GCS
    client. This is required because `aiohttp.ClientSession` (and the
    `asyncio.timeout()` calls it makes internally) are bound to the loop they
    were created on. In the copilot executor every worker thread runs its
    own event loop, so a single shared session would fail with
    "Timeout context manager should be used inside a task".
```

PR #12164, opened 2026-02-19
feat(copilot): workspace file tools, context reconstruction, transcript upload protection

majdyz:
## Summary
- **Workspace file tools**: `write_workspace_file` now accepts plain text `content`, `source_path` (copy from ephemeral disk), and graceful fallback for invalid base64. `read_workspace_file` gains `save_to_path` to download workspace files to the ephemeral working directory. Both validate paths against session-specific ephemeral directory.
- **Context reconstruction**: `_format_conversation_context` now includes tool call summaries and tool results (not just user/assistant text), fixing agent amnesia when transcript is unavailable or stale.
- **Transcript upload protection**: Moved transcript upload from inside the inner `try` block to the `finally` block, ensuring it always runs even on streaming exceptions — prevents transcript loss that caused staleness on subsequent turns.
- **Agent inactivity timeout**: Configurable timeout (default 300s) kills hung Claude agents that stop producing SDK messages.
- **SDK system prompt**: Restructured with clear sections for shell commands, two storage systems, file transfer workflows, and long-running tools.
- **Path validation hardening**: `_validate_ephemeral_path` uses `os.path.realpath` for both session dir and target path, fixing macOS `/tmp` → `/private/tmp` symlink mismatch. Empty-string params normalised to `None` to prevent dispatch assertion failures.

## Test plan
- [x] `_format_conversation_context` — empty, user, assistant, tool calls, tool results, full conversation (query_builder_test.py)
- [x] `_build_query_message` — resume up-to-date, stale transcript gap, zero msg count, no resume single/multi (query_builder_test.py)
- [x] `_validate_ephemeral_path` — valid path, traversal, cross-session, symlink escape, nested (workspace_files_test.py)
- [x] `_resolve_write_content` — no sources, multiple sources, plain text, base64, invalid base64, source_path, not found, outside ephemeral, empty strings (workspace_files_test.py)
- [ ] Verify transcript upload occurs even after streaming error
- [ ] Verify agent inactivity timeout kills hung agents (300s default)

### DIFF
--- a/autogpt_platform/frontend/src/app/(platform)/copilot/useCopilotPage.ts
+++ b/autogpt_platform/frontend/src/app/(platform)/copilot/useCopilotPage.ts
@@ -10,11 +10,30 @@ import { useSupabase } from "@/lib/supabase/hooks/useSupabase";
 import { useChat } from "@ai-sdk/react";
 import { useQueryClient } from "@tanstack/react-query";
 import { DefaultChatTransport } from "ai";
+import type { UIMessage } from "ai";
 import { useCallback, useEffect, useMemo, useRef, useState } from "react";
 import { useChatSession } from "./useChatSession";
 import { useLongRunningToolPolling } from "./hooks/useLongRunningToolPolling";
 
 const STREAM_START_TIMEOUT_MS = 12_000;
+
+/** Mark any in-progress tool parts as completed/errored so spinners stop. */
+function resolveInProgressTools(
+  messages: UIMessage[],
+  outcome: "completed" | "cancelled",
+): UIMessage[] {
+  return messages.map((msg) => ({
+    ...msg,
+    parts: msg.parts.map((part) =>
+      "state" in part &&
+      (part.state === "input-streaming" || part.state === "input-available")
+        ? outcome === "cancelled"
+          ? { ...part, state: "output-error" as const, errorText: "Cancelled" }
+          : { ...part, state: "output-available" as const, output: "" }
+        : part,
+    ),
+  }));
+}
 
 export function useCopilotPage() {
   const { isUserLoading, isLoggedIn } = useSupabase();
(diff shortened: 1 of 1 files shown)


autogpt_platform/backend/backend/copilot/sdk/service.py
```diff
@@ -387,20 +402,66 @@ def _format_conversation_context(messages: list[ChatMessage]) -> str | None:
...
+            content = msg.content or ""
+            lines.append(f"Tool result: {content}")
 
     if not lines:
         return None
 
     return "<conversation_history>\n" + "\n".join(lines) + "\n</conversation_history>"
 
 
+async def _build_query_message(
```
majdyz, line 467: is there test on the whole query building logic?

autogpt_platform/backend/backend/copilot/tools/workspace_files.py
```diff
@@ -18,6 +20,138 @@
...
+            logger.warning(
+                "[workspace] content_base64 is not valid base64, treating as plain text"
+            )
+            return content_b64.encode("utf-8")
+
+    assert content_text is not None
+    return content_text.encode("utf-8")
+
+
+def _validate_ephemeral_path(
```
majdyz, line 94: can we have e2e test for it where this is working? since workspace and ephemeral tool calls (without bwrap) are working locally

autogpt_platform/backend/backend/copilot/sdk/service.py
majdyz: Tests exist in `query_builder_test.py`: covers `_format_conversation_context` (empty, user, assistant, tool calls, tool results, full conversation) and `_build_query_message` (resume up-to-date, stale transcript gap, zero msg count, no resume single/multi). Added truncation tests for long tool args (>200 chars) and tool results (>500 chars) in commit a45e38c.

autogpt_platform/backend/backend/copilot/tools/workspace_files.py
majdyz: Added e2e tests in `workspace_files_test.py`:
- `test_workspace_file_round_trip`: write (plain text) → list → read (inline content) → read (with save_to_path to ephemeral dir) → delete → verify gone
- `test_write_workspace_file_source_path`: create file in ephemeral dir → write to workspace via source_path → cleanup

Both pass locally against the test DB (commit a45e38c).

Otto-AGPT (dismissed):
Re-approving after conflict resolution ✅

Otto-AGPT (approved):
Re-approving ✅

PR #12177, opened 2026-02-20
fix(copilot): collision detection, session locking, and sync for concurrent message saves

Otto-AGPT:
Requested by @majdyz

Concurrent writers (incremental streaming saves from PR #12173 and long-running tool callbacks) can race to persist messages with the same `(sessionId, sequence)` pair, causing unique constraint violations on `ChatMessage`.

**Root cause:** The streaming loop tracks `saved_msg_count` in-memory, but the long-running tool callback (`_build_long_running_callback`) also appends messages and calls `upsert_chat_session` independently — without coordinating sequence numbers. When the streaming loop does its next incremental save with the stale `saved_msg_count`, it tries to insert at a sequence that already exists.

**Fix:** Multi-layered defense-in-depth approach:

1. **Collision detection with retry** (db.py): `add_chat_messages_batch` uses `create_many()` in a transaction. On `UniqueViolationError`, queries `MAX(sequence)+1` from DB and retries with the correct offset (max 5 attempts).

2. **Robust sequence tracking** (db.py): `get_next_sequence()` uses indexed `find_first` with `order={"sequence": "desc"}` for O(1) MAX lookup, immune to deleted messages.

3. **Session-based counter** (model.py): Added `saved_message_count` field to `ChatSession`. `upsert_chat_session` returns the session with updated count, eliminating tuple returns throughout the codebase.

4. **MessageCounter dataclass** (sdk/service.py): Replaced list[int] mutable reference pattern with a clean `MessageCounter` dataclass for shared state between streaming loop and long-running callbacks.

5. **Session locking** (sdk/service.py): Prevent concurrent streams on the same session using Redis `SET NX EX` distributed locks with TTL refresh on heartbeats (config.stream_ttl = 3600s).

6. **Atomic operations** (db.py): Single timestamp for all messages and session update in batch operations for consistency. Parallel queries with `asyncio.gather` for lower latency.

7. **Config-based TTL** (sdk/service.py, config.py): Consolidated all TTL constants to use `config.stream_ttl` (3600s) with lock refresh on heartbeats.

### Key implementation details

- **create_many**: Uses `sessionId` directly (not nested `Session.connect`) as `create_many` doesn't support nested creates
- **Type narrowing**: Added explicit `assert session is not None` statements for pyright type checking in async contexts
- **Parallel operations**: Use `asyncio.gather` for independent DB operations (create_many + session update)
- **Single timestamp**: All messages in a batch share the same `createdAt` timestamp for atomicity

### Changes
- `backend/copilot/db.py`: Collision detection with `create_many` + retry, indexed sequence lookup, single timestamp, parallel queries
- `backend/copilot/model.py`: Added `saved_message_count` field, simplified return types  
- `backend/copilot/sdk/service.py`: MessageCounter dataclass, session locking with refresh, config-based TTL, type narrowing
- `backend/copilot/service.py`: Updated all callers to handle new return types
- `backend/copilot/config.py`: Increased long_running_operation_ttl to 3600s with clarified docstring
- `backend/copilot/*_test.py`: Tests updated for new signatures

### DIFF
--- a/autogpt_platform/backend/backend/copilot/db.py
+++ b/autogpt_platform/backend/backend/copilot/db.py
@@ -278,14 +278,14 @@ async def get_next_sequence(session_id: str) -> int:
     Uses MAX(sequence) + 1 for robustness. Returns 0 if no messages exist.
     More robust than COUNT(*) because it's immune to deleted messages.
 
-    Uses findFirst with ordering to get the max sequence efficiently.
+    Optimized to select only the sequence column using raw SQL.
     The unique index on (sessionId, sequence) makes this query fast.
     """
-    last_message = await PrismaChatMessage.prisma().find_first(
-        where={"sessionId": session_id},
-        order={"sequence": "desc"},
+    results = await db.query_raw_with_schema(
+        'SELECT "sequence" FROM {schema_prefix}"ChatMessage" WHERE "sessionId" = $1 ORDER BY "sequence" DESC LIMIT 1',
+        session_id,
     )
-    return 0 if last_message is None else last_message.sequence + 1
+    return 0 if not results else results[0]["sequence"] + 1
 
 
 async def update_tool_message_content(


autogpt_platform/backend/backend/copilot/db.py
```diff
@@ -170,8 +170,21 @@ async def add_chat_messages_batch(
...
-                data=cast(ChatMessageCreateInput, data)
+            # Use upsert to handle concurrent writers (e.g. incremental
+            # streaming saves racing with long-running tool callbacks) that
+            # may produce duplicate (sessionId, sequence) pairs.
+            update_data = {k: v for k, v in data.items() if k != "Session"}
+            created = await PrismaChatMessage.prisma(tx).upsert(
+                where={
+                    "sessionId_sequence": {
+                        "sessionId": session_id,
+                        "sequence": start_sequence + i,
```
majdyz, line 169: session_id and start_sequence + i can be replaced with data["sessionId'] and data["sequence"] ?

autogpt_platform/backend/backend/copilot/db.py
```diff
@@ -170,8 +170,21 @@ async def add_chat_messages_batch(
...
+                    }
+                },
+                data={
+                    "create": cast(ChatMessageCreateInput, data),
+                    "update": update_data,
```
majdyz: why update is not casted while create casted

autogpt_platform/backend/backend/copilot/db.py
```diff
@@ -170,8 +170,21 @@ async def add_chat_messages_batch(
             if msg.get("function_call") is not None:
                 data["functionCall"] = SafeJson(msg["function_call"])
 
-            created = await PrismaChatMessage.prisma(tx).create(
-                data=cast(ChatMessageCreateInput, data)
+            # Use upsert to handle concurrent writers (e.g. incremental
+            # streaming saves racing with long-running tool callbacks) that
+            # may produce duplicate (sessionId, sequence) pairs.
+            update_data = {k: v for k, v in data.items() if k != "Session"}
```
majdyz: should we exhaustively set the field and constract the update by ourselves just to be safe?

majdyz:
**Re: session_id vs data["sessionId"]** - No, data["sessionId"] doesn't exist (it's nested as `Session: {"connect": {"id": session_id}}`). Using loop variables is clearer.

**Re: why update not casted** - ✅ Fixed in my latest commit! Now cast to ChatMessageUpdateInput.

**Re: exhaustively set fields** - Good point! Let me implement that now for better type safety.

majdyz:
Addressed in commits 9d4697e85-bac7b9efb: Replaced upsert approach with collision detection + MAX(sequence) retry. The concerns about data construction are now handled by the dynamic dict building (lines 163-183) which only includes fields that have values, then casts to ChatMessageCreateInput.

majdyz:
Force refresh: GitHub PR cache appears stale. Latest commit is `bac7b9efb` but PR shows `6acefee6f`.

PR #12181, opened 2026-02-21
feat(frontend): show agent name + version on Agent Executor block

### LINKED ISSUE #11041: Show agent name + version on Agent Executor block

yakub268:
## Summary

Fixes #11041

This PR resolves the issue where the agent name and version disappear from Agent Executor blocks after saving and reloading the page.

## Changes

Modified the `displayTitle` logic in `CustomNode.tsx` to:
- Check if the block is an Agent Executor block (`BlockUIType.AGENT`)
- Extract `agent_name` and `graph_version` from `hardcodedValues`
- Display the agent name and version in the format "Agent Name vX" when both are present
- Display just the agent name when only the name is available
- Fall back to default block title behavior for non-agent blocks or when agent metadata is missing

## Testing

The fix ensures that:
1. When an Agent Executor block is first placed, the agent name shows correctly
2. After saving and reloading the page, the agent name + version persist in the block title
3. The version number is displayed in the format "v1", "v2", etc.
4. Custom titles (if set by the user) still take precedence

## Screenshots

_Will be updated after local testing with Docker_

## Related Issues

- Closes #11041


yakub268:
Closing in favor of updated PR based on latest dev branch

PR #12214, opened 2026-02-26
feat(copilot): Add browse_web tool for JS-rendered page browsing

majdyz:
## Summary

- Adds `browse_web` copilot tool backed by **Stagehand + Browserbase** for real-browser page extraction
- Handles JS-rendered pages, SPAs, and dynamic content that `web_fetch` cannot reach
- Fully ephemeral — no external storage required (content stays in-process, never persisted)
- Gracefully degrades when `STAGEHAND_API_KEY` / `STAGEHAND_PROJECT_ID` are not configured

## What changed

- **`browse_web.py`** — new `BrowseWebTool`: navigates to a URL, runs Stagehand `extract()` with a natural-language instruction, returns up to 50K chars of content
- **`models.py`** — adds `BROWSE_WEB` response type and `BrowseWebResponse` model
- **`tools/__init__.py`** — registers `browse_web` in `TOOL_REGISTRY`

## Design decisions

- **No R2 / no storage**: content returned inline as text — no file upload needed (unlike AutoReviewer which needs embeddable GitHub URLs)
- **Thread-safe signal handling**: Stagehand registers OS signal handlers on init which raises in worker threads. Applied same monkey-patch pattern as `blocks/stagehand/blocks.py`
- **Auth required**: `requires_auth = True` — anonymous users cannot consume Browserbase credits
- **Content cap**: 50K chars before truncation to protect LLM context window

## Required env vars (new)

```
STAGEHAND_API_KEY      # Browserbase API key
STAGEHAND_PROJECT_ID   # Browserbase project ID
ANTHROPIC_API_KEY      # LLM key for Stagehand's extraction model
```

## Test plan

- [ ] Set env vars, ask Copilot to browse a JS-heavy page (e.g. a React SPA) and verify content is returned
- [ ] Ask Copilot to browse with a specific extraction instruction (e.g. "get all pricing tiers")
- [ ] Verify graceful error message when env vars are missing
- [ ] Verify 50K char truncation on a large page
- [ ] Confirm `web_fetch` still works independently for static pages


majdyz:
@coderabbitai review

autogpt_platform/backend/backend/copilot/tools/browse_web.py
```diff
@@ -0,0 +1,207 @@
...
+JS-rendered pages, SPAs, and dynamic content that web_fetch cannot reach.
+
+Requires environment variables:
+    STAGEHAND_API_KEY     — Browserbase API key
+    STAGEHAND_PROJECT_ID  — Browserbase project ID
+    ANTHROPIC_API_KEY     — LLM key used by Stagehand for extraction
+"""
+
+import logging
+import os
```
majdyz, line 13: **Why monkey-patch instead of importing from `blocks/stagehand/blocks.py`?**

Stagehand registers OS signal handlers on `__init__`, which raises `ValueError` in non-main threads (the CoPilot executor thread pool). The same patch is applied in `blocks.py` but importing it from there would create a `copilot/tools → blocks` dependency, which is an unusual direction for this codebase. Duplicating the 5-line patch here keeps the tools module self-contained.

Matches the exact same pattern used in `blocks/stagehand/blocks.py:38-48`.

autogpt_platform/backend/backend/copilot/tools/browse_web.py
```diff
@@ -0,0 +1,207 @@
...
+# ---------------------------------------------------------------------------
+_original_register_signal_handlers = stagehand.main.Stagehand._register_signal_handlers
+
+
+def _safe_register_signal_handlers(self: Any) -> None:
+    if threading.current_thread() is threading.main_thread():
+        _original_register_signal_handlers(self)
+
+
+stagehand.main.Stagehand._register_signal_handlers = _safe_register_signal_handlers
```
majdyz: **Why monkey-patch here instead of importing from `blocks/stagehand/blocks.py`?**

Stagehand registers OS signal handlers on init, which raises `ValueError` in non-main threads (the CoPilot executor thread pool). The same patch exists in `blocks.py` — but importing it from there creates a `copilot/tools → blocks` dependency that goes against the module boundary. Duplicating the 5-line patch keeps the tool self-contained.

majdyz:
## Design notes — key decisions in this PR

**1. Signal handler monkey-patch** ([browse_web.py:38-48](https://github.com/Significant-Gravitas/AutoGPT/blob/feat/browsing-capability-copilot/autogpt_platform/backend/backend/copilot/tools/browse_web.py#L38-L48))

Stagehand calls `signal.signal()` during `__init__`, which raises `ValueError` in non-main threads. The CoPilot executor runs tools in a `ThreadPoolExecutor`, so this needs the same monkey-patch that `blocks/stagehand/blocks.py` already applies. Rather than importing from the `blocks` module (wrong dependency direction for `copilot/tools`), the 5-line patch is self-contained here.

---

**2. `requires_auth = True`** ([browse_web.py:110](https://github.com/Significant-Gravitas/AutoGPT/blob/feat/browsing-capability-copilot/autogpt_platform/backend/backend/copilot/tools/browse_web.py#L110))

Each `browse_web` call creates a Browserbase cloud browser session, which costs money. Anonymous users are explicitly blocked to prevent credit abuse. Authenticated users' costs are tied to their account.

---

**3. 50K char content cap** ([browse_web.py:20](https://github.com/Significant-Gravitas/AutoGPT/blob/feat/browsing-capability-copilot/autogpt_platform/backend/backend/copilot/tools/browse_web.py#L20))

Stagehand's `extract()` can return very long pages. Without a cap, a large page could flood the LLM context window and hit the `claude_agent_max_buffer_size` (currently 10MB). 50K chars ≈ ~12K tokens, leaving plenty of room for the rest of the context.

---

**4. No session/cookie persistence (by design)**

Each call creates a fresh ephemeral Browserbase session. Persistent sessions (for sites requiring login) can be added later by storing the Browserbase `session_id` as a user credential. The current PR intentionally keeps scope minimal — stateless browsing covers the majority of use cases (reading articles, docs, product pages).

---

**5. Overlap with #12213 (`run_mcp_tool`)**

Both PRs add a new entry to `models.py` `ResponseType` and `TOOL_REGISTRY`. This is a low-risk conflict — whoever merges second just needs to rebase. The lines don't interact.

majdyz:
### CodeRabbit review — follow-up (commit 9ae1bda)

All 5 actionable items addressed:

**🔴 Critical — unguarded module-level stagehand imports (`__init__.py` line 13 + `browse_web.py`)**
Fixed. Both `import stagehand.main` and `from stagehand import Stagehand` are now lazy-imported inside `_execute` behind a `try/except ImportError` guard. `_patch_stagehand_once()` is called only after the import succeeds. If Stagehand is absent, only `browse_web` returns an `ErrorResponse`; all other registered tools are unaffected.

**🟠 Major — `_thread_safe_signal` context manager (redundant + harmful)**
Removed entirely. The monkey-patch in `_patch_stagehand_once()` (double-checked locking, applied once per process) is sufficient. The old context manager was additionally harmful because it globally mutated `signal.signal` from a worker thread, which is not thread-safe itself.

**🟠 Major — no explicit timeouts on `page.goto()` / `page.extract()`**
Fixed. Added `timeoutMs=_GOTO_TIMEOUT_MS` (30 000 ms) and `timeoutMs=_EXTRACT_TIMEOUT_MS` (60 000 ms). Constants are declared at module level and documented.

**🟠 Major — truncation exceeded the 50 000-char cap**
Fixed. New logic: `keep = max(0, _MAX_CONTENT_CHARS - len(suffix)); content = content[:keep] + suffix` — the final string is always ≤ `_MAX_CONTENT_CHARS`.

**🟠 Major — raw exception text leaked to user**
Fixed. The exception is no longer bound (`except Exception:` without `as e`) and the user-facing message is the generic `"Failed to browse URL."`. Full traceback is logged via `logger.exception("[browse_web] Failed for %s", url)`.

majdyz:
Superseded by #12230 which includes this `browse_web` tool plus the `browser_navigate` / `browser_act` / `browser_screenshot` agent-browser tools, all fixes from reviews, and the full test suite. Closing this in favour of the combined PR.

PR #12276, opened 2026-03-04
feat(copilot): remove legacy copilot, add baseline non-SDK mode with tool calling

majdyz:
## Summary
- Remove ~1200 lines of broken/unmaintained non-SDK copilot streaming code (retry logic, parallel tool calls, context window management)
- Add `stream_chat_completion_baseline()` as a clean fallback LLM path with full tool-calling support when `CHAT_USE_CLAUDE_AGENT_SDK=false` (e.g. when Anthropic is down)
- Baseline reuses the same shared `TOOL_REGISTRY`, `get_available_tools()`, and `execute_tool()` as the SDK path
- Move baseline code to dedicated `baseline/` folder (mirrors `sdk/` structure)
- Clean up SDK service: remove unused params, fix model/env resolution, fix stream error persistence
- Clean up config: remove `max_retries`, `thinking_enabled` fields (non-SDK only)

## Changes
| File | Action |
|------|--------|
| `backend/copilot/baseline/__init__.py` | New — package export |
| `backend/copilot/baseline/service.py` | New — baseline streaming with tool-call loop |
| `backend/copilot/baseline/service_test.py` | New — multi-turn keyword recall test |
| `backend/copilot/service.py` | Remove ~1200 lines of legacy code, keep shared helpers only |
| `backend/copilot/executor/processor.py` | Simplify branching to SDK vs baseline |
| `backend/copilot/sdk/service.py` | Remove unused params, fix model/env separation, fix stream error persistence |
| `backend/copilot/config.py` | Remove `max_retries`, `thinking_enabled` |
| `backend/copilot/service_test.py` | Keep SDK test only (baseline test moved) |
| `backend/copilot/parallel_tool_calls_test.py` | Deleted (tested removed code) |

## Test plan
- [x] `poetry run format` passes
- [x] CI passes (all 3 Python versions, types, CodeQL)
- [ ] SDK path works unchanged in production
- [x] Baseline path (`CHAT_USE_CLAUDE_AGENT_SDK=false`) streams responses with tool calling
- [x] Baseline emits correct Vercel AI SDK stream protocol events

### DIFF
--- a/autogpt_platform/backend/backend/copilot/baseline/service.py
+++ b/autogpt_platform/backend/backend/copilot/baseline/service.py
@@ -76,9 +76,6 @@ async def _compress_session_messages(
     summarization of older messages while keeping recent ones intact,
     with progressive truncation and middle-out deletion as fallbacks.
     """
-    if len(messages) < 2:
-        return messages
-
     messages_dict = []
     for msg in messages:
         msg_dict: dict[str, Any] = {"role": msg.role}


majdyz:
Regarding Sentry review (tool call persistence): Valid point. Currently the message reconstruction loop only reads user/assistant text messages, so even if we persist tool interactions they would not be read back. This matches the SDK path behavior where tool calls are within-turn context. Will address tool persistence + history reconstruction in a follow-up.

ntindle:
testing something on this pr very sorry if this goes boom

ntindle:
yeah okay thats what i was afraid of

majdyz:
@ntindle I appreiate if you test it here because this one is ready for merge https://github.com/Significant-Gravitas/AutoGPT/pull/12277

autogpt_platform/backend/backend/copilot/baseline/service.py
```diff
@@ -0,0 +1,341 @@
...
+                    _update_title_async(session_id, first_message, user_id)
+                )
+                _background_tasks.add(task)
+                task.add_done_callback(_background_tasks.discard)
+
+    message_id = str(uuid.uuid4())
+
+    # Build system prompt
+    has_history = len(session.messages) > 1
+    system_prompt, _ = await _build_system_prompt(
```
Swiftyos: We are rebuilding the system prompt every step, injecting in the users buisness understanding. 

I don't think we want to do this, as it means that concurrent chats can change the behaviour of the current chat. Adding business understanding that has been extracted from the current chat isn’t needed as this information is already in context.

Probably only need to rebuild the system prompt after a compaction.

Swiftyos (dismissed):

autogpt_platform/backend/backend/copilot/baseline/service.py
majdyz: Good point, this is preexisting issue, addressed, please recheck

Swiftyos (approved):

PR #12282, opened 2026-03-04
feat(copilot): collapse repeated tool calls and fix stream stuck on completion

0ubbe:
## Summary
- **Frontend:** Group consecutive completed generic tool parts into collapsible summary rows with a "Reasoning" collapse for finalized messages. Merge consecutive assistant messages on hydration to avoid split bubbles. Extract GenericTool helpers. Add `reconnectExhausted` state and a brief delay before refetching session to reduce stale `active_stream` reconnect cycles.
- **Backend:** Make transcript upload fire-and-forget instead of blocking the generator exit. The 30s upload timeout in `_try_upload_transcript` was delaying `mark_session_completed()`, keeping the SSE stream alive with only heartbeats after the LLM had finished — causing the UI to stay stuck in "streaming" state.

## Test plan
- [ ] Send a message in Copilot that triggers multiple tool calls — verify they collapse into a grouped summary row once completed
- [ ] Verify the final text response appears below the collapsed reasoning section
- [ ] Confirm the stream properly closes after the agent finishes (no stuck "Stop" button)
- [ ] Refresh mid-stream and verify reconnection works correctly
- [ ] Click Stop during streaming — verify the UI becomes responsive immediately

### DIFF
--- a/autogpt_platform/frontend/src/app/(platform)/copilot/useCopilotPage.ts
+++ b/autogpt_platform/frontend/src/app/(platform)/copilot/useCopilotPage.ts
@@ -5,10 +5,9 @@ import {
   type getV2ListSessionsResponse,
 } from "@/app/api/__generated__/endpoints/chat/chat";
 import { toast } from "@/components/molecules/Toast/use-toast";
+import { uploadFileDirect } from "@/lib/direct-upload";
 import { useBreakpoint } from "@/lib/hooks/useBreakpoint";
-import { getWebSocketToken } from "@/lib/supabase/actions";
 import { useSupabase } from "@/lib/supabase/hooks/useSupabase";
-import { environment } from "@/services/environment";
 import { useQueryClient } from "@tanstack/react-query";
 import type { FileUIPart } from "ai";
 import { useEffect, useRef, useState } from "react";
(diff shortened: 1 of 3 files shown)


autogpt_platform/frontend/src/app/(platform)/copilot/components/ChatMessagesContainer/ChatMessagesContainer.tsx
```diff
@@ -23,6 +27,139 @@ interface Props {
...
+function buildRenderSegments(
+  parts: MessagePart[],
+  baseIndex = 0,
+): RenderSegment[] {
+  const segments: RenderSegment[] = [];
+  let pendingGroup: Array<{ part: ToolUIPart; index: number }> | null = null;
+
+  function flushGroup() {
+    if (!pendingGroup) return;
+    if (pendingGroup.length >= 2) {
```
majdyz, line 74: will there be a case where remaining pending group stuck forever because of this ?

autogpt_platform/frontend/src/app/(platform)/copilot/components/ChatMessagesContainer/ChatMessagesContainer.tsx
```diff
@@ -23,6 +27,139 @@ interface Props {
...
+ * text + tools before the final response) and "response" (final text after the
+ * last tool). If there are no tools, everything is response.
+ */
+function splitReasoningAndResponse(parts: MessagePart[]): {
+  reasoning: MessagePart[];
+  response: MessagePart[];
+} {
+  // Find the index of the last tool part
+  let lastToolIndex = -1;
+  for (let i = parts.length - 1; i >= 0; i--) {
```
majdyz: findlast ?

autogpt_platform/frontend/src/app/(platform)/copilot/components/ChatMessagesContainer/ChatMessagesContainer.tsx
0ubbe: Nope\! `flushGroup()` is called at the end of the loop (line 101) as a catch-all, and when there's only 1 item in the group it just emits it as a regular individual part (the `else` branch at line 79). So nothing gets stuck 👍

autogpt_platform/frontend/src/app/(platform)/copilot/components/ChatMessagesContainer/ChatMessagesContainer.tsx
0ubbe: Good call, switched to `findLastIndex`, much cleaner 🙏

PR #12287, opened 2026-03-05
hotfix(frontend/signup): Add missing createUser() call in password signup

Otto-AGPT:
Requested by @0ubbe

Password signup was missing the backend `createUser()` call that the OAuth callback flow already had. This caused `getOnboardingStatus()` to fail/hang for new users whose backend record didn't exist yet, resulting in an infinite spinner after account creation.

## Root Cause

| Flow | createUser() | getOnboardingStatus() | Result |
|------|-------------|----------------------|--------|
| OAuth signup | ✅ Called | ✅ Works | Redirects correctly |
| Password signup | ❌ Missing | ❌ Fails/hangs | Infinite spinner |

## Fix

Adds `createUser()` call in `signup/actions.ts` after session is set, before onboarding status check — matching the OAuth callback pattern. Includes error handling with Sentry reporting.

## Testing

- Create a new password account → should redirect without spinner
- OAuth signup unaffected (no changes to that flow)

Fixes OPEN-3023

### DIFF
--- a/autogpt_platform/frontend/src/app/(platform)/signup/actions.ts
+++ b/autogpt_platform/frontend/src/app/(platform)/signup/actions.ts
@@ -1,11 +1,11 @@
 "use server";
 
-import BackendAPI from "@/lib/autogpt-server-api";
+import { postV1GetOrCreateUser } from "@/app/api/__generated__/endpoints/auth/auth";
+import { getOnboardingStatus, resolveResponse } from "@/app/api/helpers";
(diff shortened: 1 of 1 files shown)


0ubbe (dismissed):

0ubbe (approved):
I have tested locally and user creation works, the change makes sense 🤔 🏁 , we'll see whether it actually solves the issue once deployed 🍿

PR #12289, opened 2026-03-05
fix(copilot): handle empty tool_call arguments in baseline path

majdyz:
## Summary

Handle empty/None `tool_call.arguments` in the baseline copilot path that cause OpenRouter 400 errors when converting to Anthropic format.

## Changes

**`backend/copilot/baseline/service.py`**:
- Default empty `tc["arguments"]` to `"{}"` to prevent OpenRouter from failing on empty tool arguments during format conversion.

## Test plan

- [x] Existing baseline tests pass
- [ ] Verify on staging: trigger a tool call in baseline mode and confirm normal flow works

### DIFF
--- a/autogpt_platform/backend/backend/copilot/baseline/service.py
+++ b/autogpt_platform/backend/backend/copilot/baseline/service.py
@@ -273,8 +273,6 @@ async def stream_chat_completion_baseline(
             step_open = False
 
             # Append the assistant message with tool_calls to context.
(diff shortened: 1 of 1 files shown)


majdyz:
All 3 coderabbit comments are stale — they refer to logging/validation code from earlier commits that has been simplified away. The current fix is a single line: `tc["arguments"] or "{}"`. No logging, no validation loop, no sensitive data exposure.

Swiftyos (approved):

PR #12293, opened 2026-03-05
fix(copilot): pin claude-agent-sdk to 0.1.39 to fix tool_reference content block validation error

Otto-AGPT:
Requested by @swiftyos

## Problem

CoPilot throws `400 Invalid Anthropic Messages API request` errors on first message, both locally and on Dev.

## Root Cause

The CLI's built-in `ToolSearch` tool returns `tool_reference` content blocks (`{"type": "tool_reference", "tool_name": "mcp__copilot__find_block"}`). When the CLI constructs the next Anthropic API request, it passes these blocks as-is in the `tool_result.content` field. However, the Anthropic Messages API only accepts `text` and `image` content block types in tool results.

This causes a Zod validation error:
```
messages[3].content[0].content: Invalid input: expected string, received array
```

The error only manifests when using **OpenRouter** (`ANTHROPIC_BASE_URL` set) because the Anthropic TypeScript SDK performs stricter client-side Zod validation in that code path vs the subscription auth path.

PR #12288 bumped `claude-agent-sdk` from `0.1.39` to `^0.1.46`, which upgraded the bundled Claude CLI from `v2.1.49` to `v2.1.69` where this issue was introduced.

## Fix

Pin to exact version `0.1.39` (not `^`) until the upstream CLI bug is resolved. Our code doesn't use any 0.1.46-specific SDK APIs.

## Testing
- CoPilot first message should work without 400 errors via OpenRouter
- SDK compat tests should still pass

### DIFF
--- a/autogpt_platform/backend/poetry.lock
+++ b/autogpt_platform/backend/poetry.lock
@@ -899,17 +899,17 @@ files = [
 
 [[package]]
 name = "claude-agent-sdk"
-version = "0.1.46"
+version = "0.1.39"
 description = "Python SDK for Claude Code"
 optional = false
 python-versions = ">=3.10"
 groups = ["main"]
 files = [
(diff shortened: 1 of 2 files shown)


Otto-AGPT:
Closing — Zamil is handling the fix separately.

PR #12309, opened 2026-03-06
fix(backend): Use db_manager for workspace in add_graph_execution

Otto-AGPT:
When `add_graph_execution` is called from a context where the global Prisma client isn't connected (e.g. CoPilot tools, external API), the call to `get_or_create_workspace(user_id)` crashes with `ClientNotConnectedError` because it directly accesses `UserWorkspace.prisma()`.

The fix adds `workspace_db` to the existing `if prisma.is_connected()` fallback pattern, consistent with how all other DB calls in the function already work.

**Sentry:** AUTOGPT-SERVER-83T (and ~15 related issues going back to Jan 2026)

---
Co-authored-by: Reinier van der Leer (@Pwuts) <pwuts@agpt.co>


Otto-AGPT:
Superseded by new PR targeting master (hotfix)

PR #12316, opened 2026-03-06
fix(frontend): show sub-folders when navigating inside a folder

Abhi1992002:
## Summary

When opening a folder in the library, sub-folders were not displayed — only agents were shown. This was caused by two issues:

1. The folder list query always fetched root-level folders (no `parent_id` filter), so sub-folders were never requested
2. `showFolders` was set to `false` whenever a folder was selected, hiding all folders from the view

### Changes 🏗️

- Pass `parent_id` to the `useGetV2ListLibraryFolders` hook so it fetches child folders of the currently selected folder
- Remove the `!selectedFolderId` condition from `showFolders` so folders render inside other folders
- Fetch the current folder via `useGetV2GetFolder` instead of searching the (now differently-scoped) folder list
- Clean up breadcrumb: remove emoji icon, match folder name text size to "My Library", replace `Button` with plain `<button>` to remove extra padding/gap

#### For code changes:
- [x] I have made a test plan
  - [x] Open a folder in the library and verify sub-folders are displayed
  - [x] Verify agents inside the folder still display correctly
  - [x] Verify breadcrumb shows folder name without emoji, matching "My Library" text size
  - [x] Verify clicking "My Library" in breadcrumb navigates back to root
  - [x] Verify root-level view still shows all top-level folders
  - [x] Verify favorites tab does not show folders

### DIFF
--- a/autogpt_platform/frontend/src/app/(platform)/library/components/LibraryAgentList/LibraryAgentList.tsx
+++ b/autogpt_platform/frontend/src/app/(platform)/library/components/LibraryAgentList/LibraryAgentList.tsx
@@ -134,7 +134,7 @@ export function LibraryAgentList({
 
       <div>
         {selectedFolderId && (
-          <div className="mb-4 flex items-center gap-1">
+          <div className="mb-4 flex items-center gap-2">
             <button
               type="button"
               onClick={() => onFolderSelect(null)}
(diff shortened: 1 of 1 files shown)


ntindle:
I'm with the bot here -- how does this even happen?

PR #12317, opened 2026-03-06
feat(frontend/copilot): improve TTS voice selection to avoid robotic voices

Otto-AGPT:
Requested by @0ubbe

Refines the `pickBestVoice()` function to ensure non-robotic voices are always preferred:

- **Filter out known low-quality engines** — eSpeak, Festival, MBROLA, Flite, and Pico voices are deprioritized
- **Prefer remote/cloud-backed voices** — `localService: false` voices are typically higher quality
- **Expand preferred voices list** — added Moira, Tessa (macOS), Jenny, Aria, Guy (Windows OneCore)
- **Smarter fallback chain** — English default → English → any default → first available

The previous fallback could select eSpeak or Festival voices on Linux systems, resulting in robotic output. Now those are filtered out unless they're the only option.

---
Co-authored-by: Ubbe <ubbe@users.noreply.github.com>


ntindle:
voiceschanged race condition — pickBestVoice() calls getVoices() directly, which returns an empty array before the browser's voiceschanged event fires. On first playback, the entire voice selection logic is bypassed and the browser falls back to its default voice. This directly undermines the PR's purpose. Fix: cache voices via a voiceschanged listener in the hook, or await voice availability before first play. (Flagged by CodeRabbit + Discussion analyst — pre-existing from feat(frontend/copilot): add text-to-speech and share output actions #12256 but this PR should fix it since voice selection quality is its core goal.)

0ubbe:
@ntindle Addressed in d3dbd87cf:

- **`voiceschanged` race condition**: `pickBestVoice()` no longer calls `getVoices()` directly. Voices are now cached in a ref that's populated eagerly on mount and updated via a `voiceschanged` event listener, so first playback uses the correct voice selection logic.
- Also ranked preferred voice matches by quality (remote non-robotic > non-robotic > remote > first) instead of returning the first browser-ordered `find()` hit, and added a JSDoc to `pickBestVoice()`.

0ubbe:
Tested locally and works. 

When I click the read aloud button, the nice voice is used. Before the fix, the first time I played, the robotic one will come up ( from the 🐎 race condition above ) and had to click again to get the nice one.

PR #12328, opened 2026-03-07
feat(platform): load copilot messages from newest first with cursor-based pagination

kcze:
Copilot chat sessions with long histories loaded all messages at once, causing slow initial loads. This PR adds cursor-based pagination so only the most recent messages load initially, with older messages fetched on demand as the user scrolls up.

### Changes 🏗️

**Backend:**
- Cursor-based pagination on `GET /sessions/{session_id}` (`limit`, `before_sequence` params)
- `user_id` relation filter on the paginated query — ownership check and message fetch now run in parallel
- Backward boundary expansion to keep tool-call / assistant message pairs intact at page edges
- Unit tests for paginated queries

**Frontend:**
- `useLoadMoreMessages` hook + `LoadMoreSentinel` (IntersectionObserver) for infinite scroll upward
- `ScrollPreserver` to maintain scroll position when older messages are prepended
- Session-keyed `Conversation` remount with one-frame opacity hide to eliminate scroll flash on switch
- Scrollbar moved to the correct scroll container; loading spinner no longer causes overflow

### Checklist 📋

- [x] Pagination: only recent messages load initially; older pages load on scroll-up
- [x] Scroll position preserved on prepend; no flash on session switch
- [x] Tool-call boundary pairs stay intact across page edges
- [x] Stream reconnection still works on initial load

### DIFF
--- a/autogpt_platform/frontend/src/app/(platform)/build/components/FlowEditor/nodes/CustomNode/components/NodeOutput/components/ContentRenderer.tsx
+++ b/autogpt_platform/frontend/src/app/(platform)/build/components/FlowEditor/nodes/CustomNode/components/NodeOutput/components/ContentRenderer.tsx
@@ -40,14 +40,14 @@ export const ContentRenderer: React.FC<{
     !shortContent
   ) {
     return (
-      <div className="overflow-hidden [&>*]:rounded-xlarge [&>*]:!text-xs [&_pre]:whitespace-pre-wrap [&_pre]:break-words">
+      <div className="overflow-x-auto [&>*]:rounded-xlarge [&>*]:!text-xs [&_pre]:whitespace-pre-wrap [&_pre]:break-words">
         {renderer?.render(value, metadata)}
       </div>
     );
   }
 
   return (
-    <div className="overflow-hidden [&>*]:rounded-xlarge [&>*]:!text-xs">
+    <div className="overflow-x-auto [&>*]:rounded-xlarge [&>*]:!text-xs">
       <TextRenderer value={value} truncateLengthLimit={200} />
     </div>
   );


autogpt_platform/backend/backend/api/features/chat/routes.py
```diff
@@ -322,53 +326,61 @@ async def update_session_title_route(
...
-    Looks up a chat session by ID for the given user (if authenticated) and returns all session data including messages.
-    If there's an active stream for this session, returns active_stream info for reconnection.
+    Supports cursor-based pagination via ``limit`` and ``before_sequence``.
+    When no pagination params are provided, returns the most recent messages.
 
     Args:
         session_id: The unique identifier for the desired chat session.
         user_id: The optional authenticated user ID, or None for anonymous access.
+        limit: Maximum number of messages to return (1-200, default 50).
+        before_sequence: Return messages with sequence < this value (cursor).
```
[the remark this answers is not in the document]
kcze: Fixed — aligned default to 50 across route, db.py, docstring, openapi.json, and frontend loadMore call. /Claude

autogpt_platform/backend/backend/copilot/db.py
```diff
@@ -32,6 +32,70 @@ async def get_chat_session(session_id: str) -> ChatSession | None:
...
+    if results and results[0].role == "tool":
+        boundary_where: dict = {
+            "sessionId": session_id,
+            "sequence": {"lt": results[0].sequence},
+        }
+        extra = await PrismaChatMessage.prisma().find_many(
+            where=boundary_where,
+            order={"sequence": "desc"},
+            take=10,
+        )
```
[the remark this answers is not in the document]
kcze: Fixed — added `logger.warning` when boundary expansion fails to find the owning assistant message, and guarded `has_more = True` to only trigger when `boundary_msgs` is non-empty. /Claude

autogpt_platform/frontend/src/app/api/openapi.json
```diff
@@ -1113,7 +1113,7 @@
...
+        "description": "Retrieve the details of a specific chat session.\n\nSupports cursor-based pagination via ``limit`` and ``before_sequence``.\nWhen no pagination params are provided, returns the most recent messages.\n\nArgs:\n    session_id: The unique identifier for the desired chat session.\n    user_id: The optional authenticated user ID, or None for anonymous access.\n    limit: Maximum number of messages to return (1-200, default 50).\n    before_sequence: Return messages with sequence < this value (cursor).\n\nReturns:\n    SessionDetailResponse: Details for the requested session, including\n        active_stream info and pagination metadata.",
         "operationId": "getV2GetSession",
         "security": [{ "HTTPBearerJWT": [] }],
```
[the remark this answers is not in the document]
kcze: This is a pre-existing pattern — the endpoint uses `Depends(auth.get_user_id)` which returns `None` for unauthenticated requests rather than rejecting them. The OpenAPI spec declares `HTTPBearerJWT` because the dependency is defined at the router level, but the actual behavior is optional auth. Not changing this as it's consistent with other endpoints in the codebase. /Claude

autogpt_platform/backend/backend/copilot/db.py
```diff
@@ -32,6 +32,80 @@ async def get_chat_session(session_id: str) -> ChatSession | None:
...
+        if not found_owner:
+            logger.warning(
+                "Boundary expansion did not find owning assistant message "
+                "for session=%s before sequence=%s (%d msgs scanned)",
+                session_id,
+                results[0].sequence,
+                len(extra),
+            )
+        results = boundary_msgs + results
+        has_more = True  # There may be more before the expanded boundary
```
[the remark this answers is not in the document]
kcze: Fixed — `has_more` is now only set to `True` when `boundary_msgs` is non-empty. /Claude

autogpt_platform/frontend/src/app/(platform)/copilot/useLoadMoreMessages.ts
```diff
@@ -0,0 +1,72 @@
...
+    UIMessage<unknown, UIDataTypes, UITools>[]
+  >([]);
+  const [oldestSequence, setOldestSequence] = useState<number | null>(
+    initialOldestSequence,
+  );
+  const [hasMore, setHasMore] = useState(initialHasMore);
+  const [isLoadingMore, setIsLoadingMore] = useState(false);
+
+  // Track the sessionId to reset state on change
+  const prevSessionIdRef = useRef(sessionId);
```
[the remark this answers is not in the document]
kcze: Fixed — added `isLoadingMoreRef` as a synchronous guard to prevent duplicate `loadMore()` calls from concurrent IntersectionObserver callbacks. /Claude

autogpt_platform/frontend/src/app/(platform)/copilot/useLoadMoreMessages.ts
```diff
@@ -0,0 +1,72 @@
...
+      prevSessionIdRef.current = sessionId;
+      setOlderMessages([]);
+      setOldestSequence(initialOldestSequence);
+      setHasMore(initialHasMore);
+      setIsLoadingMore(false);
+    } else {
+      // Update from parent when initial data changes (e.g. refetch)
+      setOldestSequence(initialOldestSequence);
+      setHasMore(initialHasMore);
+    }
```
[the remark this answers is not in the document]
kcze: Fixed — added `prevInitialOldestRef` tracking. When the same session's `initialOldestSequence` shifts and paged messages exist, the hook now clears `olderMessages` and resets the cursor to avoid gaps/duplicates. /Claude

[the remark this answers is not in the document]
kcze: Fixed — added `prevInitialOldestRef` tracking and `isLoadingMoreRef` synchronous lock. When the same session's initial window shifts, paged state is now cleared to prevent duplicates. /Claude

autogpt_platform/frontend/src/app/(platform)/copilot/components/ChatMessagesContainer/ChatMessagesContainer.tsx
```diff
@@ -170,13 +175,92 @@ function getTurnMessages(
...
+
+  // Restore scroll position after prepend
+  useLayoutEffect(() => {
+    const el = scrollRef.current;
+    if (!el || prevRef.current.scrollHeight === 0) return;
+    const delta = el.scrollHeight - prevRef.current.scrollHeight;
+    if (delta > 0 && prevRef.current.scrollTop > 0) {
+      el.scrollTop = prevRef.current.scrollTop + delta;
+    }
+  }, [messageCount]);
```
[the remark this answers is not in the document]
kcze: The `ScrollPreserver` component captures scroll data via a continuous `useEffect` (runs every render) and restores via `useLayoutEffect` keyed on `messageCount`. The ref is updated before the layout effect fires, so it has the pre-mutation values. The dead `prevScrollDataRef` in `LoadMoreSentinel` (which was the actual unused code) has been removed. /Claude

autogpt_platform/backend/backend/copilot/db.py
```diff
@@ -32,6 +32,81 @@ async def get_chat_session(session_id: str) -> ChatSession | None:
...
+            logger.warning(
+                "Boundary expansion did not find owning assistant message "
+                "for session=%s before sequence=%s (%d msgs scanned)",
+                session_id,
+                results[0].sequence,
+                len(extra),
+            )
+        if boundary_msgs:
+            results = boundary_msgs + results
+            has_more = True  # There may be more before the expanded boundary
```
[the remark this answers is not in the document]
kcze: Fixed — `has_more = True` is now guarded by `if boundary_msgs:`, so it's only set when the expansion actually finds messages. /Claude

autogpt_platform/frontend/src/app/(platform)/copilot/components/ChatMessagesContainer/ChatMessagesContainer.tsx
```diff
@@ -170,13 +175,79 @@ function getTurnMessages(
...
+      ([entry]) => {
+        if (entry.isIntersecting && hasMore && !isLoading) {
+          loadMoreRef();
+        }
+      },
+      { rootMargin: "200px 0px 0px 0px" },
+    );
+    observer.observe(sentinelRef.current);
+    return () => observer.disconnect();
+  }, [hasMore, isLoading]);
```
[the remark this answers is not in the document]
kcze: Fixed — using a ref pattern (`onLoadMoreRef`) so the IntersectionObserver always calls the latest `onLoadMore` without needing it in the dependency array. /Claude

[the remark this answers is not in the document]
kcze: Fixed — using a ref pattern (onLoadMoreRef) so the IntersectionObserver always calls the latest onLoadMore without needing it in the dependency array. /Claude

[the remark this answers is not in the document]
kcze: Fixed — using a ref pattern (onLoadMoreRef) so the IntersectionObserver always calls the latest onLoadMore. Also added error handling with a retry cap (3 consecutive errors) to prevent infinite retry loops. /Claude

autogpt_platform/backend/backend/copilot/model.py
```diff
@@ -394,6 +396,41 @@ async def get_chat_session(
...
+        logger.warning("Redis error fetching session metadata for %s", session_id)
+    except Exception:
+        logger.warning(
+            "Unexpected error fetching session metadata for %s",
+            session_id,
+            exc_info=True,
+        )
+
+    # Fall back to lightweight DB query (no messages)
+    from backend.copilot.db import get_chat_session_metadata as _db_get_metadata
```
ntindle: Can this be a top level import?

kcze: No — `db.py` imports from `model.py` (`from .model import ChatMessage, ChatSession, ChatSessionInfo`), so a top-level import here would create a circular import. Keeping it inline to break the cycle. /Claude

autogpt_platform/frontend/src/app/(platform)/copilot/useLoadMoreMessages.ts
```diff
@@ -0,0 +1,119 @@
...
+      consecutiveErrorsRef.current += 1;
+      console.warn("[loadMore] Network error:", error);
+      if (consecutiveErrorsRef.current >= MAX_CONSECUTIVE_ERRORS) {
+        setHasMore(false);
+      }
+    } finally {
+      isLoadingMoreRef.current = false;
+      setIsLoadingMore(false);
+    }
+  }
```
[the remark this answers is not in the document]
kcze: Fixed — added an epoch counter (`epochRef`). Each reset increments the epoch; `loadMore()` captures the epoch before the await and discards the response if it changed. This prevents stale responses from mixing messages across sessions or after a pagination reset. /Claude

autogpt_platform/frontend/src/app/(platform)/copilot/components/ChatMessagesContainer/ChatMessagesContainer.tsx
```diff
@@ -170,13 +175,80 @@ function getTurnMessages(
...
+        if (entry.isIntersecting) {
+          onLoadMoreRef.current();
+        }
+      },
+      { rootMargin: "200px 0px 0px 0px" },
+    );
+    observer.observe(sentinelRef.current);
+    return () => observer.disconnect();
+  }, [hasMore, isLoading]);
+
```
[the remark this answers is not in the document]
kcze: Fixed — the IntersectionObserver callback now checks whether the scroll container (`[role="log"]`) has overflowing content before triggering `loadMore()`. When content fits without scrolling (`scrollHeight <= clientHeight`), the callback is a no-op, preventing automatic page exhaustion. /Claude

autogpt_platform/frontend/src/app/(platform)/copilot/useLoadMoreMessages.ts
```diff
@@ -0,0 +1,119 @@
...
+    ) {
+      // Same session but initial window shifted (e.g. new messages arrived) —
+      // clear paged state to avoid gaps/duplicates
+      prevInitialOldestRef.current = initialOldestSequence;
+      setOlderMessages([]);
+      setOldestSequence(initialOldestSequence);
+      setHasMore(initialHasMore);
+      setIsLoadingMore(false);
+      isLoadingMoreRef.current = false;
+      consecutiveErrorsRef.current = 0;
```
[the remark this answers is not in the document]
kcze: This is intentional — when `initialOldestSequence` shifts on the same session, the paged messages were fetched against a now-stale cursor. Preserving them risks gaps or duplicates in the sequence. Clearing and re-fetching is the safe choice. The user only loses pages they haven't scrolled to yet (the visible viewport stays intact via ScrollPreserver). /Claude

autogpt_platform/frontend/src/app/(platform)/copilot/useLoadMoreMessages.ts
```diff
@@ -0,0 +1,119 @@
...
+    } else {
+      // Update from parent when initial data changes (e.g. refetch)
+      prevInitialOldestRef.current = initialOldestSequence;
+      setOldestSequence(initialOldestSequence);
+      setHasMore(initialHasMore);
+    }
```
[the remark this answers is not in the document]
kcze: Already addressed — the `prevInitialOldestRef` tracking ensures state is only reset when the initial window actually shifts while paged data exists. The `else` branch (no shift) safely syncs parent values without clearing `olderMessages`. Additionally, the new epoch guard (epochRef) discards any stale loadMore responses after a reset. /Claude

autogpt_platform/backend/backend/copilot/db.py
```diff
@@ -32,6 +32,81 @@ async def get_chat_session(session_id: str) -> ChatSession | None:
...
+    before_sequence: int | None = None,
+) -> tuple[list[ChatMessage], bool, int | None]:
+    """Get paginated messages for a session, newest first.
+
+    Returns:
+        Tuple of (messages in ascending order, has_more, oldest_sequence).
+    """
+    where: dict = {"sessionId": session_id}
+    if before_sequence is not None:
+        where["sequence"] = {"lt": before_sequence}
```
majdyz: 🤖 **[Code Style] Untyped `dict` annotation.** `where: dict` should be at minimum `dict[str, Any]`. Same applies to `boundary_where: dict` on line 74. Bare `dict` gives no information about shape and suppresses type narrowing. Prefer Prisma's typed where-input types if available.

autogpt_platform/backend/backend/copilot/db.py
```diff
@@ -32,6 +32,81 @@ async def get_chat_session(session_id: str) -> ChatSession | None:
...
+            order={"sequence": "desc"},
+            take=10,
+        )
+        # Find the first non-tool message (should be the assistant)
+        boundary_msgs = []
+        found_owner = False
+        for msg in extra:
+            boundary_msgs.insert(0, msg)
+            if msg.role != "tool":
+                found_owner = True
```
majdyz, line 133: 🤖 **[Code Style / Perf] `list.insert(0, ...)` in a loop is O(n²).** Each call shifts all existing elements. Use `append` then `reverse` after the loop:
```python
boundary_msgs = []
for msg in extra:
    boundary_msgs.append(msg)
    if msg.role != "tool":
        found_owner = True
        break
boundary_msgs.reverse()
```

autogpt_platform/backend/backend/copilot/db.py
```diff
@@ -32,6 +32,81 @@ async def get_chat_session(session_id: str) -> ChatSession | None:
...
+        if boundary_msgs:
+            results = boundary_msgs + results
+            has_more = True  # There may be more before the expanded boundary
+
+    messages = [ChatMessage.from_db(m) for m in results]
+    oldest_sequence = messages[0].sequence if messages else None
+
+    return messages, has_more, oldest_sequence
+
+
```
majdyz, line 161: 🤖 **[Medium] `has_more = True` after boundary expansion can be a false positive.** If the expanded messages include the very first message in the conversation, `has_more` is incorrectly `True`. The frontend fires one wasted `loadMore` request that returns `has_more = False`. Consider checking `oldest_sequence_after_expansion > 1` before setting this.

autogpt_platform/backend/backend/copilot/model.py
```diff
@@ -394,6 +396,41 @@ async def get_chat_session(
...
+
+    info = await _db_get_metadata(session_id)
+    if info is None:
+        return None
+    if user_id is not None and info.user_id != user_id:
+        return None
+    return info
```
majdyz: 🤖 **[Code Style] Local import to avoid circular dependency.** Project conventions say to avoid local imports. The circular dependency between `model.py` and `db.py` should be resolved architecturally — e.g. extract the shared types into a `copilot/types.py` module that both can import from.

autogpt_platform/frontend/src/app/(platform)/copilot/useCopilotPage.ts
```diff
@@ -60,6 +63,19 @@ export function useCopilotPage() {
...
   });
 
+  const { olderMessages, hasMore, isLoadingMore, loadMore } =
+    useLoadMoreMessages({
+      sessionId,
+      initialOldestSequence: oldestSequence,
+      initialHasMore: hasMoreMessages,
+    });
+
+  // Combine older (paginated) messages with current page messages
```
majdyz: 🤖 **[Medium — Collapsed Reasoning Clash] Consecutive-assistant merging does not cross the page boundary.** `convertChatSessionMessagesToUiMessages` merges consecutive `assistant` DB rows into one `UIMessage` within a single call. But `olderMessages` and `currentMessages` are converted independently, then array-spread here. If two consecutive `assistant` DB rows straddle the page join (one in each array), they remain as two separate `UIMessage` objects.

This directly breaks the collapsed reasoning feature: `splitReasoningFromResponse` is applied per-`UIMessage`. A reasoning+response turn split at the boundary renders as:
- An older bubble with **only tool/reasoning parts** (collapsed block with no response text)
- A current bubble with **only response text** (no reasoning toggle)

**Fix:** pass the full concatenated raw message list into a single `convertChatSessionMessagesToUiMessages` call rather than concatenating already-converted arrays.

autogpt_platform/frontend/src/app/(platform)/copilot/useLoadMoreMessages.ts
```diff
@@ -0,0 +1,130 @@
...
+      if (epochRef.current !== requestEpoch) return;
+
+      if (response.status !== 200) {
+        consecutiveErrorsRef.current += 1;
+        console.warn(
+          `[loadMore] Failed to load messages (status=${response.status}, attempt=${consecutiveErrorsRef.current})`,
+        );
+        if (consecutiveErrorsRef.current >= MAX_CONSECUTIVE_ERRORS) {
+          setHasMore(false);
+        }
```
majdyz, line 125: 🤖 **[Medium] `isComplete: true` for historical pages silently loses tool outputs that are in a newer page.** If an `assistant` message in page N has `tool_calls` whose corresponding `tool` result messages are in page N+1 (e.g. because the forward page boundary cuts between them), this conversion sees no matching output and emits `state: "output-available", output: ""`. The backend boundary fix only prevents page N from **starting** with an orphaned `tool` message — it does not prevent page N from **ending** with an `assistant` whose tool results land in the next page.

autogpt_platform/frontend/src/app/(platform)/copilot/helpers/convertChatSessionToUiMessages.ts
```diff
@@ -190,7 +192,7 @@ export function convertChatSessionMessagesToUiMessages(
     }
 
     uiMessages.push({
```
majdyz, line 271: 🤖 **[Low] `sequence ?? index` fallback produces unstable IDs for pre-migration messages.** For messages where `sequence` is `null` (created before the column was added), the ID falls back to `${sessionId}-seq-${index}` where `index` is the position in the current call's input array. The same message can appear at different positions in different pagination requests (e.g. after boundary expansion prepends it), changing its ID and breaking React key reconciliation. Use the message's `id` field (DB UUID) as the fallback instead — it's deterministic and already available in the response.

autogpt_platform/backend/backend/copilot/db.py
kcze: Fixed — changed `where: dict` to `dict[str, Any]` for both `where` and `boundary_where`. Also refactored the return type from a tuple to a `PaginatedMessages` dataclass for clarity. /Claude

autogpt_platform/backend/backend/copilot/db.py
kcze: Fixed — replaced `list.insert(0, ...)` loop with `append` + `reverse`. /Claude

autogpt_platform/backend/backend/copilot/db.py
kcze: Fixed — `has_more` is now only set to `True` when `boundary_msgs[0].sequence > 0`, preventing the false positive when boundary expansion reaches the very first message in the conversation. /Claude

autogpt_platform/backend/backend/copilot/model.py
kcze: The circular dependency between `model.py` and `db.py` is structural: `db.py` imports types from `model.py` (`ChatMessage`, `ChatSession`, `ChatSessionInfo`), and `model.py` needs a DB query function from `db.py` for its cache-miss fallback. Extracting shared types to a `types.py` module is a good idea but it's a larger refactor touching many consumers across the copilot package — out of scope for this pagination PR. Keeping the local import for now as the pragmatic choice. /Claude

autogpt_platform/frontend/src/app/(platform)/copilot/useCopilotPage.ts
kcze: Valid concern. The independent conversion of `olderMessages` and `currentMessages` means consecutive assistant DB rows straddling the page boundary won't be merged into a single UIMessage. In practice this is rare (assistant turns usually end with a text part, and tool messages create natural boundaries), but it can happen with reasoning-heavy responses.

The proper fix (passing raw messages to a single conversion call) requires changing `useLoadMoreMessages` to return raw DB messages instead of pre-converted UIMessages, which is a meaningful refactor. Tracking this as a follow-up. /Claude

autogpt_platform/frontend/src/app/(platform)/copilot/useLoadMoreMessages.ts
kcze: This is the same root cause as the cross-boundary merging issue (useCopilotPage.ts:73) — converting pages independently means assistant+tool pairs that straddle boundaries aren't properly linked. The backend boundary fix covers the start-of-page case (expanding backward to include the owning assistant), but not end-of-page where tool results land in the next (newer) page.

In practice, since we load newest-first and the initial page has the most recent messages, the most common case (current conversation) is unaffected. Historical page joins could show empty tool outputs briefly until the user scrolls to load the adjacent page. Same follow-up as above. /Claude

autogpt_platform/frontend/src/app/(platform)/copilot/helpers/convertChatSessionToUiMessages.ts
kcze: The DB `ChatMessage.id` (UUID) exists but isn't currently exposed in the API response — only `sequence` is returned. Since the migration backfills sequences for all existing messages, `sequence` should never be null in practice. The `?? index` fallback is a safety net for any edge case where it is, and since each pagination call returns a contiguous block, the index is stable within that call. If we see instability in practice, adding the DB UUID to the API response would be the right fix. /Claude

autogpt_platform/frontend/src/app/(platform)/copilot/useCopilotPage.ts
majdyz: Human @kcze: are we sure this is supposde to be a follow-up ?

autogpt_platform/frontend/src/app/(platform)/copilot/useLoadMoreMessages.ts
majdyz: This shouldn't be a follow-up either?

autogpt_platform/frontend/src/app/(platform)/copilot/helpers/convertChatSessionToUiMessages.ts
majdyz: what's the risk of not have this ?

autogpt_platform/frontend/src/app/(platform)/copilot/useCopilotPage.ts
kcze: Fixed — added `concatWithAssistantMerge` helper that merges consecutive assistant UIMessages at the join point between two arrays. Applied at both boundaries: (1) between older pages in `useLoadMoreMessages` (inter-page joins when prepending), (2) between `olderMessages` and `currentMessages` in `useCopilotPage`. This ensures reasoning + response parts stay in a single UIMessage bubble even when the DB rows straddle a pagination boundary. /Claude

autogpt_platform/frontend/src/app/(platform)/copilot/useLoadMoreMessages.ts
kcze: Fixed — three-part solution: (1) Added `extraToolOutputs` param to `convertChatSessionMessagesToUiMessages` plus `extractToolOutputsFromRaw` helper for building the cross-page lookup map. (2) Changed `useLoadMoreMessages` to store accumulated raw messages and re-convert them all in one pass — this fixes tool output matching across inter-page boundaries (older page N-1 assistant can now find tool results at the start of older page N). (3) Initial page raw messages are passed via `extraToolOutputs` so the last older page can find tool results that landed in the initial/streaming page. /Claude

autogpt_platform/backend/backend/api/features/chat/routes.py
```diff
@@ -323,53 +327,59 @@ async def update_session_title_route(
...
-        SessionDetailResponse: Details for the requested session, including active_stream info if applicable.
-
+        SessionDetailResponse: Details for the requested session, including
+            active_stream info and pagination metadata.
     """
-    session = await get_chat_session(session_id, user_id)
-    if not session:
+    # Lightweight ownership validation (no message loading)
+    session_info = await get_chat_session_metadata(session_id, user_id)
+    if not session_info:
```
majdyz: This is better improvement, but filtering the ChatMessage.Session.userId directly is much better & faster solution than doing two blocking call (ownershipcheck then fetch chat) like this

majdyz: you can add mandatory userId filter on get_chat_messages_paginated and get_chat_session (do we still need this method then? or can we remove it)

autogpt_platform/backend/backend/copilot/db.py
```diff
@@ -32,6 +41,86 @@ async def get_chat_session(session_id: str) -> ChatSession | None:
...
+        where["sequence"] = {"lt": before_sequence}
+
+    results = await PrismaChatMessage.prisma().find_many(
+        where=where,
+        order={"sequence": "desc"},
+        take=limit + 1,
+    )
+
+    has_more = len(results) > limit
+    if has_more:
```
majdyz: nit: obvious variable assingment that can be just inlined

autogpt_platform/backend/backend/copilot/db.py
```diff
@@ -32,6 +41,86 @@ async def get_chat_session(session_id: str) -> ChatSession | None:
...
+        boundary_msgs.reverse()
+        if not found_owner:
+            logger.warning(
+                "Boundary expansion did not find owning assistant message "
+                "for session=%s before sequence=%s (%d msgs scanned)",
+                session_id,
+                results[0].sequence,
+                len(extra),
+            )
+        if boundary_msgs:
```
majdyz, line 144: please make sure we add tests for this complicated logic.

autogpt_platform/backend/backend/copilot/db.py
kcze: Fixed — removed the if-guard, the slice is a no-op when len <= limit. /Claude

autogpt_platform/backend/backend/copilot/db.py
kcze: Added `db_test.py` with 10 tests covering: basic pagination (ascending order, has_more, empty session, before_sequence filter), and boundary expansion (assistant inclusion, multi-tool scan, has_more at seq 0 vs middle, no expansion for non-tool first msg, warning when scan limit exceeded). /Claude

autogpt_platform/backend/backend/api/features/chat/routes.py
kcze: Fixed — `get_chat_messages_paginated` now accepts an optional `user_id` parameter and filters via `Session: {"is": {"userId": user_id}}` in the Prisma where clause. The route runs both the metadata query and paginated message query in parallel via `asyncio.gather` instead of sequential blocking calls. Added a test for the user_id filter. /Claude

autogpt_platform/backend/backend/copilot/db.py
kcze: Already addressed in an earlier commit — the `if has_more: results = results[:limit]` was simplified to unconditional `results = results[:limit]` (slicing a shorter list is a no-op). /Claude

autogpt_platform/backend/backend/copilot/db.py
kcze: Done — `backend/copilot/db_test.py` has 11 unit tests covering: basic pagination ordering, has_more detection, empty session, before_sequence cursor, user_id filter, and all boundary expansion cases (assistant inclusion, multi-tool scan, has_more at conversation start vs middle, no expansion for non-tool first message, warning when scan limit exceeded). /Claude

autogpt_platform/backend/backend/api/features/chat/routes.py
majdyz: This is weird, why need session_info if we load get_chat_messages_paginated? why can't we get everything with a single query? we can easily fetch the metadata and related chat using it.

autogpt_platform/backend/backend/copilot/db.py
```diff
@@ -32,6 +41,96 @@ async def get_chat_session(session_id: str) -> ChatSession | None:
...
+    # Reverse to ascending order
+    results.reverse()
+
+    # Tool-call boundary fix: if the oldest message is a tool message,
+    # expand backward to include the preceding assistant message that
+    # owns the tool_calls, so convertChatSessionMessagesToUiMessages
+    # can pair them correctly.
+    _BOUNDARY_SCAN_LIMIT = 10
+    if results and results[0].role == "tool":
+        boundary_where: dict[str, Any] = {
```
majdyz, line 116: 🤖 🟠 **Should Fix**: The boundary expansion query (lines 91-96) filters only by `sessionId` + `sequence` but does not include a `Session.userId` filter like the main query does (line 71). While not directly exploitable (the main query already validates the session belongs to the user), this is inconsistent with the defense-in-depth approach documented at route level. If future changes alter the flow, this missing filter could become a gap.

Suggested fix: add `"Session": {"is": {"userId": user_id}}` to `boundary_where` when `user_id` is provided.

ntindle (dismissed):

autogpt_platform/backend/backend/api/features/chat/routes.py
kcze: Fixed — `get_chat_messages_paginated` now does everything in a single call: it runs `find_first` (session existence + ownership check) and `find_many` (paginated messages) in parallel via `asyncio.gather`, and returns `PaginatedMessages | None` (includes `session: ChatSessionInfo`). The route just calls `get_chat_messages_paginated(session_id, limit, before_sequence, user_id=user_id)` and uses `page.session` for metadata. No separate `get_chat_session_metadata` call needed anymore.

autogpt_platform/backend/backend/copilot/db.py
kcze: Fixed — boundary expansion query now includes `Session: {"is": {"userId": user_id}}` filter when `user_id` is provided, matching the main query's defense-in-depth pattern.

autogpt_platform/frontend/src/app/(platform)/copilot/useLoadMoreMessages.ts
```diff
@@ -0,0 +1,150 @@
...
+      isLoadingMoreRef.current = false;
+      consecutiveErrorsRef.current = 0;
+      epochRef.current += 1;
+    } else {
+      // Update from parent when initial data changes (e.g. refetch)
+      prevInitialOldestRef.current = initialOldestSequence;
+      setOldestSequence(initialOldestSequence);
+      setHasMore(initialHasMore);
+    }
+  }, [sessionId, initialOldestSequence, initialHasMore]);
```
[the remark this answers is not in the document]
kcze: Not a bug — `olderRawMessages` is intentionally excluded from the dependency array. The effect only reads `olderRawMessages.length > 0` as a guard to decide whether a cursor shift requires a reset. Adding it to deps would cause the effect to re-fire on every `loadMore()` completion (when `setOlderRawMessages` updates), which would trigger unwanted resets mid-pagination.

The stale closure scenario (refetch changes `initialOldestSequence` while `olderRawMessages` is stale-empty) is safe: it falls through to the `else` branch which just syncs parent values without clearing. The `epochRef` counter independently protects against stale `loadMore()` responses landing after a reset.

autogpt_platform/backend/backend/copilot/db.py
```diff
@@ -32,6 +42,118 @@ async def get_chat_session(session_id: str) -> ChatSession | None:
...
+
+    # Run session check and message fetch in parallel
+    session, results = await asyncio.gather(
+        PrismaChatSession.prisma().find_first(where=session_where),
+        PrismaChatMessage.prisma().find_many(
+            where=msg_where,
+            order={"sequence": "desc"},
+            take=limit + 1,
+        ),
+    )
```
majdyz, line 96: 🤖 🟠 **Should Fix**: Two parallel queries via `asyncio.gather` is still 2 DB roundtrips. Use a single Prisma `find_first` with `include` to fetch session + paginated messages atomically in 1 roundtrip:

```python
msg_include_where: dict[str, Any] = {}
if before_sequence is not None:
    msg_include_where["sequence"] = {"lt": before_sequence}

session = await PrismaChatSession.prisma().find_first(
    where=session_where,
    include={
        "Messages": {
            "where": msg_include_where,
            "order_by": {"sequence": "desc"},
            "take": limit + 1,
        },
    },
)

if session is None:
    return None

results = session.Messages or []
session_info = ChatSessionInfo.from_db(session)
# ... rest of pagination logic unchanged
```

Benefits:
- Single DB roundtrip (Prisma translates this to a single SQL query with JOIN)
- Atomic: no window where session is deleted between the two queries
- Removes the redundant `Session.is.userId` filter on messages (messages are already scoped to the session which is scoped to the user)
- Eliminates the `asyncio.gather` import dependency for this function

autogpt_platform/backend/backend/copilot/db.py
```diff
@@ -32,6 +42,118 @@ async def get_chat_session(session_id: str) -> ChatSession | None:
     return ChatSession.from_db(session) if session else None
 
 
+async def get_chat_session_metadata(session_id: str) -> ChatSessionInfo | None:
+    """Get chat session metadata (without messages) for ownership validation."""
+    session = await PrismaChatSession.prisma().find_unique(
+        where={"id": session_id},
+    )
+    return ChatSessionInfo.from_db(session) if session else None
+
```
majdyz, line 58: 🤖 🟡 **Nice to Have**: `get_chat_session_metadata` appears to be dead code — no callers in the diff or in the route. If it was introduced for the parallel ownership check that was later folded into `get_chat_messages_paginated`, it should be removed to avoid confusion.

autogpt_platform/backend/backend/api/features/chat/routes.py
```diff
@@ -356,57 +359,61 @@ async def update_session_title_route(
...
+        if active_session:
+            active_stream_info = ActiveStreamInfo(
+                turn_id=active_session.turn_id,
+                last_message_id=last_message_id,
+            )
 
     # Sum token usage from session
-    total_prompt = sum(u.prompt_tokens for u in session.usage)
-    total_completion = sum(u.completion_tokens for u in session.usage)
+    total_prompt = sum(u.prompt_tokens for u in page.session.usage)
```
majdyz, line 457: 🤖 🟠 **Should Fix**: On "load more" requests (`before_sequence` is set), the route still fetches `page.session.usage` and computes token totals (lines 405-406), even though the frontend `useLoadMoreMessages.loadMore()` only reads `messages`, `has_more_messages`, and `oldest_sequence` from the response. The session metadata + usage computation is wasted work on every pagination request.

Consider either:
1. Skipping usage/metadata in the response when `before_sequence` is set (add a lightweight `/sessions/{id}/messages` endpoint for pagination)
2. Or at minimum, skip the `find_first` session query on paginated requests since ownership is already validated by the message-level `Session.userId` filter

autogpt_platform/frontend/src/app/(platform)/copilot/components/ChatMessagesContainer/ChatMessagesContainer.tsx
```diff
@@ -99,14 +104,119 @@ function extractGraphExecId(
...
+  const scrollSnapshotRef = useRef({ scrollHeight: 0, scrollTop: 0 });
+  const { scrollRef } = useStickToBottomContext();
+
+  // IntersectionObserver to trigger load when sentinel is near viewport.
+  // Only fires when the container is actually scrollable to prevent
+  // exhausting all pages when content fits without scrolling.
+  useEffect(() => {
+    if (!sentinelRef.current || !hasMore || isLoading) return;
+    const observer = new IntersectionObserver(
+      ([entry]) => {
```
majdyz, line 149: 🤖 🟡 **Nice to Have**: The `opacity-0` → `opacity-100` flash-prevention relies on `requestAnimationFrame` settling within one frame. On slow devices or heavy React trees, this can cause a visible blank frame (or multiple). Consider using `visibility: hidden` instead of `opacity-0` — it's semantically equivalent here and avoids the transition cost. Also, `requestAnimationFrame` isn't guaranteed to fire before paint in all browsers (Safari sometimes batches them). A `useLayoutEffect` would be more reliable for this pattern.

autogpt_platform/backend/backend/copilot/db.py
```diff
@@ -32,6 +42,118 @@ async def get_chat_session(session_id: str) -> ChatSession | None:
...
+    # Reverse to ascending order
+    results.reverse()
+
+    # Tool-call boundary fix: if the oldest message is a tool message,
+    # expand backward to include the preceding assistant message that
+    # owns the tool_calls, so convertChatSessionMessagesToUiMessages
+    # can pair them correctly.
+    _BOUNDARY_SCAN_LIMIT = 10
+    if results and results[0].role == "tool":
```
majdyz, line 115: 🤖 🟡 **Nice to Have**: The boundary expansion (lines 110-143) fires a 3rd DB query when the page starts with a tool message. This extra roundtrip could be avoided by over-fetching slightly in the initial query (e.g., `take=limit + 11` instead of `limit + 1`) and doing the boundary scan in-memory on the over-fetched results. You'd only fall back to the extra DB query if the over-fetch window wasn't large enough — which in practice it almost always would be.

autogpt_platform/frontend/src/app/(platform)/copilot/useLoadMoreMessages.ts
```diff
@@ -0,0 +1,161 @@
...
+  // Convert all accumulated raw messages in one pass so tool outputs
+  // are matched across inter-page boundaries. Initial page tool outputs
+  // are included via extraToolOutputs to handle the boundary between
+  // the last older page and the initial/streaming page.
+  const olderMessages: UIMessage<unknown, UIDataTypes, UITools>[] =
+    useMemo(() => {
+      if (!sessionId || olderRawMessages.length === 0) return [];
+      const extraToolOutputs =
+        initialPageRawMessages.length > 0
+          ? extractToolOutputsFromRaw(initialPageRawMessages)
```
majdyz, line 88: 🤖 🟠 **Should Fix**: `olderMessages` is recomputed via `useMemo` keyed on `olderRawMessages` — which changes on every `loadMore()`. This means **all** accumulated older raw messages are re-converted through `convertChatSessionMessagesToUiMessages` every time a new page is loaded. With MAX_OLDER_MESSAGES=2000, this gets progressively expensive.

Consider caching previously converted pages and only converting the new page, then concatenating with \`concatWithAssistantMerge\`. The cross-page tool output matching can be limited to the boundary between the new page and the adjacent one, not the entire history.

autogpt_platform/backend/backend/copilot/db.py
kcze: Good idea, but Prisma's `include` on `Messages` doesn't support `take`/`order`/`where` filters needed for cursor-based pagination — we'd need raw SQL. Deferring to a follow-up optimization PR.

autogpt_platform/backend/backend/copilot/db.py
kcze: Fixed — removed `get_chat_session_metadata` from `model.py` in 303cb30. The `db.py` version remains as a lightweight utility.

autogpt_platform/backend/backend/api/features/chat/routes.py
kcze: Fixed in 014be00 — load-more requests now skip active_stream lookup and token usage summation, returning zeros for token fields.

autogpt_platform/frontend/src/app/(platform)/copilot/components/ChatMessagesContainer/ChatMessagesContainer.tsx
kcze: Acknowledged — the current approach works well in practice. The single-frame `opacity-0` has been reliable even on lower-end devices. Not changing for now.

autogpt_platform/backend/backend/copilot/db.py
kcze: Nice optimization idea. The boundary expansion only triggers when the page starts with a tool message, which is relatively rare. Over-fetching by 11 on every request to avoid an occasional 3rd query isn't a clear win. Deferring to a follow-up if we see it in profiling.

autogpt_platform/frontend/src/app/(platform)/copilot/useLoadMoreMessages.ts
kcze: Valid concern. We now cap `olderRawMessages` at 2000 (added in 303cb30), so the reconversion is bounded. Incremental caching would add significant complexity for a marginal gain within that cap. Deferring to a follow-up if profiling shows this as a bottleneck.

autogpt_platform/backend/backend/copilot/db.py
majdyz: are you sure? See

```
GRAPH_EXECUTION_INCLUDE_WITH_NODES: prisma.types.AgentGraphExecutionInclude = {
    "NodeExecutions": {
        "include": EXECUTION_RESULT_INCLUDE,
        "order_by": EXECUTION_RESULT_ORDER,
        "take": MAX_NODE_EXECUTIONS_FETCH,  # Avoid loading excessive node executions.
    }
}
```

in includes.py

kcze: Fixed in 8d159fa — you were right, Prisma `include` fully supports `take`/`order_by`/`where` (as shown by `GRAPH_EXECUTION_INCLUDE_WITH_NODES` in `includes.py`). Refactored `get_chat_messages_paginated` to use a single `find_first` with `include={"Messages": {order_by, take, where}}` instead of two parallel queries via `asyncio.gather`. This reduces DB round-trips from 2 to 1 for the main pagination path. Boundary expansion remains a separate conditional query since it only fires when the page starts with a tool message.

autogpt_platform/frontend/src/app/(platform)/copilot/helpers/convertChatSessionToUiMessages.ts
kcze: Fixed in 8d159fa — removed the `?? index` fallback. The `sequence` column is `NOT NULL INTEGER` in the DB (enforced by the migration), so `sequence` is always populated. The fallback was dead code, and worse, `index` is position-dependent within the current API response page, making IDs unstable across different pagination requests. No risk in removing it.

autogpt_platform/backend/backend/copilot/db.py
```diff
@@ -32,6 +42,116 @@ async def get_chat_session(session_id: str) -> ChatSession | None:
...
+    results = results[:limit]
+
+    # Reverse to ascending order
+    results.reverse()
+
+    # Tool-call boundary fix: if the oldest message is a tool message,
+    # expand backward to include the preceding assistant message that
+    # owns the tool_calls, so convertChatSessionMessagesToUiMessages
+    # can pair them correctly.
+    _BOUNDARY_SCAN_LIMIT = 10
```
majdyz, line 114: 🤖 🟡 **Nice to Have**: `_BOUNDARY_SCAN_LIMIT = 10` — if >9 consecutive tool messages exist before a page boundary, the owning assistant message won't be found. This is rare in practice but could happen with agents that make many parallel tool calls. A `find_first(where={role: {not: "tool"}})` query would be more robust and still a single DB roundtrip.

autogpt_platform/frontend/src/components/ai-elements/conversation.tsx
```diff
@@ -12,13 +11,9 @@ export type ConversationProps = ComponentProps<typeof StickToBottom>;
...
 export const Conversation = ({ className, ...props }: ConversationProps) => (
   <StickToBottom
-    className={cn(
-      "relative flex-1 overflow-y-hidden",
-      scrollbarStyles,
-      className,
-    )}
-    initial="smooth"
-    resize="smooth"
+    className={cn("relative flex-1 overflow-y-hidden", className)}
```
majdyz, line 14: 🤖 🟡 **Nice to Have**: `smooth→instant` scroll change affects all chat scrolling, not just pagination. Normal message streaming now snaps instead of animating. Consider keeping `smooth` for `resize` (new streaming content) and only using `instant` for `initial` (page load / session switch).

autogpt_platform/frontend/src/app/(platform)/copilot/useLoadMoreMessages.ts
```diff
@@ -0,0 +1,161 @@
...
+      setHasMore(initialHasMore);
+    }
+  }, [sessionId, initialOldestSequence, initialHasMore]);
+
+  // Convert all accumulated raw messages in one pass so tool outputs
+  // are matched across inter-page boundaries. Initial page tool outputs
+  // are included via extraToolOutputs to handle the boundary between
+  // the last older page and the initial/streaming page.
+  const olderMessages: UIMessage<unknown, UIDataTypes, UITools>[] =
+    useMemo(() => {
```
majdyz, line 84: 🤖 🟡 **Nice to Have**: All accumulated older raw messages are re-converted through `convertChatSessionMessagesToUiMessages` every time a new page is loaded (keyed on `olderRawMessages` which changes on each `loadMore()`). Bounded by `MAX_OLDER_MESSAGES=2000` so not a blocker, but for very long conversations this gets progressively more expensive. Incremental page conversion + `concatWithAssistantMerge` at boundaries would be more efficient.

autogpt_platform/backend/backend/copilot/db.py
```diff
@@ -23,6 +24,15 @@
 logger = logging.getLogger(__name__)
 
 
+class PaginatedMessages(BaseModel):
```
majdyz, line 34: 🤖 🟡 **Nice to Have**: `PaginatedMessages` is defined here in `db.py` (importing `BaseModel` from pydantic) rather than in `model.py` due to the circular dependency between the two modules. Consider extracting shared types (`PaginatedMessages`, `ChatMessage`, `ChatSessionInfo`, etc.) into a `copilot/types.py` to break the cycle cleanly. Same pattern keeps coming up in this package.

autogpt_platform/backend/backend/copilot/db.py
kcze: Acknowledged — good suggestion. A `find_first(where={role: {not: "tool"}, sequence: {lt: ...}})` would be more robust than the scan limit. Will note for a follow-up.

autogpt_platform/frontend/src/components/ai-elements/conversation.tsx
kcze: Good catch — the `smooth→instant` change was overly broad. Will scope it to only use `instant` for `initial` (page load / session switch) while keeping `smooth` for `resize` (streaming content) in a follow-up.

autogpt_platform/frontend/src/app/(platform)/copilot/useLoadMoreMessages.ts
kcze: Acknowledged — bounded by `MAX_OLDER_MESSAGES=2000` so not a practical issue yet, but incremental page conversion would be more efficient for very long conversations. Noted for follow-up.

autogpt_platform/backend/backend/copilot/db.py
kcze: Agreed — extracting shared types into `copilot/types.py` would cleanly break the circular dependency. Noted for follow-up.

autogpt_platform/frontend/src/components/ai-elements/conversation.tsx
kcze: Fixed in 63405da — changed `resize` back to `"smooth"` while keeping `initial="instant"`. Streaming content now animates smoothly again; page load/session switch still snaps instantly.

majdyz:
## 🤖 PR Review — #12328 (cursor-based pagination)

**Verdict: ✅ APPROVED**

### Assessment

| Area | Rating |
|---|---|
| Description | ✅ Clear Why/What/How |
| Correctness | ✅ Cursor-based pagination with boundary expansion for tool messages |
| Security | ✅ User ownership check in parallel with query via `userId` filter |
| Testing | ✅ 328 lines of unit tests covering pagination, boundaries, edge cases |
| Performance | ✅ Single query for session+messages, parallel ownership check |
| Frontend | ✅ IntersectionObserver for scroll-to-load, proper memoization |

### Notable Good Patterns
- Tool-call boundary fix: If oldest message is a `tool` message, expands backward to include the owning `assistant` message. Smart edge case handling.
- `limit + 1` fetch for `has_more` detection without extra count query
- Active stream check only on initial load, not on "load more" requests

### No Blockers Found
Code is clean, well-tested, and follows the project conventions.

autogpt_platform/backend/backend/copilot/db.py
```diff
@@ -32,6 +42,116 @@ async def get_chat_session(session_id: str) -> ChatSession | None:
...
+        include={"Messages": msg_include},
+    )
+
+    if session is None:
+        return None
+
+    session_info = ChatSessionInfo.from_db(session)
+    results = list(session.Messages) if session.Messages else []
+
+    has_more = len(results) > limit
```
majdyz, line 104: 🤖 🟡 **Nice to Have**: `_BOUNDARY_SCAN_LIMIT = 10` is a good safeguard. Consider making this configurable or documenting why 10 — in theory a single assistant tool_calls message could invoke >10 tools, though unlikely in practice.

autogpt_platform/backend/backend/copilot/db.py
```diff
@@ -32,6 +42,116 @@ async def get_chat_session(session_id: str) -> ChatSession | None:
...
+    if user_id is not None:
+        session_where["userId"] = user_id
+
+    # Build message include — fetch paginated messages in the same query
+    msg_include: dict[str, Any] = {
+        "order_by": {"sequence": "desc"},
+        "take": limit + 1,
+    }
+    if before_sequence is not None:
+        msg_include["where"] = {"sequence": {"lt": before_sequence}}
```
majdyz, line 90: 🤖 🔵 **Nit**: Good use of `find_first` with both `userId` filter and `Messages` include in a single query — avoids N+1 and performs ownership check + message fetch atomically.

kcze: Thanks for the review!

autogpt_platform/backend/backend/copilot/db.py
kcze: Acknowledged — will improve in a follow-up (replacing the scan limit with a `find_first(where={role: {not: "tool"}})` query as suggested earlier).

itsababseh:
@Torantulino this the PR you are looking for and its pending fixes @kcze if we can wrap this up so we can test

ntindle (approved):
https://github.com/user-attachments/assets/e5272b32-aea1-4beb-b13f-14f1b8670f03

works

ntindle (dismissed):

PR #12344, opened 2026-03-09
fix(frontend/builder): batch undo history for cascading operations

### LINKED ISSUE #10999: Undo in Builder not working correctly
To reproduce:

1. Place 3 blocks: A B C
2. Connect them: A-B-C
3. Remove block: A-B
4. Remove connection: A B
5. Undo
   1. Expected behavior - connection restored: A-B
   2. Actual behavior - block restored: A B C

eureka0928:
## Summary

Fixes undo in the Builder not working correctly when deleting nodes. When a node is deleted, React Flow fires `onNodesChange` (node removal) and `onEdgesChange` (cascading edge cleanup) as separate callbacks — each independently pushing to the undo history stack. This creates intermediate states that break undo:

- Single undo restores a partial state (e.g. edges pointing to a deleted node)
- Multiple undos required to fully restore the graph
- Redo also produces inconsistent states

Resolves #10999

### Changes 🏗️

- **`historyStore.ts`** — Added microtask-based batching to `pushState()`. Multiple calls within the same synchronous execution (same event loop tick) are coalesced into a single history entry, keeping only the first pre-change snapshot. Uses `queueMicrotask` so all cascading store updates from a single user action settle before the history entry is committed.
- Reset `pendingState` in `initializeHistory()` and `clear()` to prevent stale batched state from leaking across graph loads or navigation.

**Side benefit:** Copy/paste operations that add multiple nodes and edges now also produce a single history entry instead of one per node/edge.

#### For code changes:
- [x] I have made a test plan
  - [x] Place 3 blocks A, B, C and connect A→B→C
  - [x] Delete block C (removes node + cascading edge B→C)
  - [x] Delete connection A→B
  - [x] Undo — connection A→B restored (single undo, not multiple)
  - [x] Undo — block C and connection B→C restored
  - [x] Redo — block C removed again with its connections
  - [x] Copy/paste multiple connected blocks — single undo reverts entire paste


fury0928:
> Everything looks good to me — thank you for the contribution! 🙌

Thank you for your review, would you merge my PR?

PR #12347, opened 2026-03-09
feat(platform): add whitelisting-backed beta user provisioning

Swiftyos:
### Changes 🏗️

- add invite-backed beta provisioning with a new `InvitedUser` platform model, Prisma migration, and first-login activation path that materializes `User`, `Profile`, `UserOnboarding`, and `CoPilotUnderstanding`
- replace the legacy beta allowlist check with invite-backed gating for email/password signup and Tally pre-seeding during activation
- add admin backend APIs and frontend `/admin/users` management UI for listing, creating, revoking, retrying, and bulk-uploading invited users
- add the design doc for the beta invite system and extend backend coverage for invite activation, bulk uploads, and auth-route behavior
- configuration changes: introduce the new invite/tally schema objects and migration; no new env vars or docker service changes are required

#### For code changes:
- [x] I have made a test plan
  - [x] `cd autogpt_platform/backend && poetry run format`
  - [x] `cd autogpt_platform/backend && poetry run pytest -q` (run against an isolated local Postgres database with non-conflicting service port overrides)

### DIFF
--- a/autogpt_platform/backend/migrations/20260312093554_update_invites_indexs/migration.sql
+++ b/autogpt_platform/backend/migrations/20260312093554_update_invites_indexs/migration.sql
@@ -2,8 +2,6 @@
 DROP INDEX "InvitedUser_status_idx";
 -- DropIndex
 DROP INDEX "InvitedUser_tallyStatus_idx";
--- DropIndex
-DROP INDEX "UnifiedContentEmbedding_search_idx";
 -- CreateIndex
 CREATE INDEX "InvitedUser_createdAt_idx"
 ON "InvitedUser"("createdAt");


Swiftyos:
Fix 1 (CRITICAL): Migration no longer drops the GIN index; instead recreates it with CREATE INDEX IF NOT EXISTS at the end.                          
                                                                                                                                                       
  Fix 2 (CRITICAL): get_or_activate_user now catches UniqueViolationError from concurrent activation and re-fetches the already-created user.          
                                                                                                                                                     
  Fix 3: _tally_seed_tasks changed from set to dict[str, Task] keyed by invited_user_id, preventing duplicate tasks for the same user.

  Fix 4: Admin routes now log before ("creating"/"revoking"/"retrying") and after operations, with emails redacted via _redact_email().

  Fix 5: Design doc fixed "racey" -> "racy" and replaced all absolute local paths with repo-relative paths.

  Fix 6: BulkInviteForm file input now has proper id/htmlFor association, sr-only instead of hidden for keyboard accessibility, and focus-within:ring
  styles.

  Fix 7: useAdminUsersPage now exposes invitedUsersError from the query.

  Fix 8: Removed deprecated ApiError import and redundant instanceof check.

  Fix 9: Replaced lucide-react and IconSliders imports with Phosphor icons (UsersIcon, CurrencyDollarSimpleIcon, UserPlusIcon, MagnifyingGlassIcon,
  FileTextIcon, SlidersHorizontalIcon).

autogpt_platform/backend/backend/data/invited_user.py
```diff
@@ -0,0 +1,625 @@
...
+    await prisma.models.CoPilotUnderstanding.prisma(tx).upsert(
+        where={"userId": user_id},
+        data={
+            "create": {"userId": user_id, "data": SafeJson(payload)},
+            "update": {"data": SafeJson(payload)},
+        },
+    )
+
+
+async def list_invited_users() -> list[InvitedUserRecord]:
```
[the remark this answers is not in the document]
Swiftyos: finding is not actionable — no fix needed

autogpt_platform/backend/backend/api/features/admin/model.py
```diff
@@ -14,3 +18,42 @@ class UserHistoryResponse(BaseModel):
...
+
+
+class InvitedUserResponse(BaseModel):
+    id: str
+    email: str
+    status: prisma.enums.InvitedUserStatus
+    auth_user_id: Optional[str] = None
+    name: Optional[str] = None
+    tally_understanding: Optional[dict[str, Any]] = None
+    tally_status: prisma.enums.TallyComputationStatus
```
majdyz, line 40: better use this or literal?

autogpt_platform/backend/backend/api/features/admin/user_admin_routes.py
```diff
@@ -0,0 +1,161 @@
...
+    InvitedUserResponse,
+    InvitedUsersResponse,
+)
+
+logger = logging.getLogger(__name__)
+
+
+def _redact_email(email: str) -> str:
+    local, _, domain = email.partition("@")
+    return f"{local[:2]}***@{domain}" if domain else f"{local[:2]}***"
```
majdyz: this will break when the email is >3 chars

autogpt_platform/backend/backend/api/features/admin/user_admin_routes.py
```diff
@@ -0,0 +1,161 @@
...
+
+router = APIRouter(
+    prefix="/admin",
+    tags=["users", "admin"],
+    dependencies=[Security(requires_admin_user)],
+)
+
+
+def _to_response(invited_user: InvitedUserRecord) -> InvitedUserResponse:
+    return InvitedUserResponse(**invited_user.model_dump())
```
majdyz: use model_validate instead of this so you don't need to serialize them into string

autogpt_platform/backend/backend/api/features/admin/user_admin_routes.py
```diff
@@ -0,0 +1,161 @@
...
+
+
+def _to_bulk_response(result: BulkInvitedUsersResult) -> BulkInvitedUsersResponse:
```
majdyz: these to_db/transformation methods can be a static method of the BulkInvitedUsersResult class itself or make it from_db for BulkInvitedUsersResponse

autogpt_platform/backend/backend/data/tally.py
```diff
@@ -381,6 +381,35 @@ async def extract_business_understanding(
     return BusinessUnderstandingInput(**cleaned)
 
 
+async def get_business_understanding_input_from_tally(
+    email: str,
+    *,
+    require_api_key: bool = False,
+) -> Optional[BusinessUnderstandingInput]:
+    settings = Settings()
```
majdyz: settings should be global it's unfortunately expensive

autogpt_platform/backend/migrations/20260310090914_add_invite_system/migration.sql
```diff
@@ -0,0 +1,52 @@
...
+);
+-- CreateEnum
+CREATE TYPE "TallyComputationStatus" AS ENUM(
+  'PENDING',
+  'RUNNING',
+  'READY',
+  'FAILED'
+);
+-- AlterTable
+ALTER TABLE "StoreListingVersion" DROP COLUMN "search";
```
majdyz: Beware this is not-intentional change due to poetry run prisma dev, I'll make a separate PR that handle this

autogpt_platform/docs/beta-invite-design.md
```diff
@@ -0,0 +1,393 @@
+# Beta Invite Pre-Provisioning Design
```
majdyz: Remove this?

autogpt_platform/backend/backend/data/invited_user.py
```diff
@@ -0,0 +1,647 @@
...
+from backend.util.exceptions import (
+    NotAuthorizedError,
+    NotFoundError,
+    PreconditionFailed,
+)
+from backend.util.json import SafeJson
+
+logger = logging.getLogger(__name__)
+
+_tally_seed_tasks: dict[str, asyncio.Task] = {}
```
majdyz, line 42: 🤖 **[HIGH] `_tally_seed_tasks` is not safe under multi-worker deployments.** Each gunicorn/uvicorn worker has its own copy of this module-level dict — there is no cross-process coordination. Two workers can fire simultaneous Tally LLM calls for the same email with no deduplication. If a worker restarts mid-task, the status stays `RUNNING` indefinitely with no recovery path. Consider a Celery task or Postgres-backed job queue instead.

autogpt_platform/backend/backend/data/invited_user.py
```diff
@@ -0,0 +1,647 @@
...
+    local_part = email.split("@", 1)[0].lower()
+    sanitized = re.sub(r"[^a-z0-9-]", "", local_part)
+    sanitized = sanitized.strip("-")
+    return sanitized[:40] or "user"
+
+
+async def _generate_unique_profile_username(email: str, tx) -> str:
+    base = _sanitize_username_base(email)
+
+    for attempt in range(10):
```
majdyz: 🤖 **[MEDIUM] Up to 10 sequential DB round-trips inside an open transaction.** `_generate_unique_profile_username` retries up to 10 times, each with a separate `find_unique` query, all while holding the activation transaction. A single `INSERT ... ON CONFLICT` with a random suffix from the start would avoid this entirely.

autogpt_platform/backend/backend/data/invited_user.py
```diff
@@ -0,0 +1,647 @@
...
+        where={"userId": user_id},
+        data={
+            "create": {"userId": user_id, "data": SafeJson(payload)},
+            "update": {"data": SafeJson(payload)},
+        },
+    )
+
+
+async def list_invited_users() -> list[InvitedUserRecord]:
+    invited_users = await prisma.models.InvitedUser.prisma().find_many(
```
majdyz, line 218: 🤖 **[MEDIUM] No pagination — returns the entire table.** `find_many()` with no `take` limit returns every `InvitedUser` row in a single query. At scale this will become expensive. The frontend also polls this endpoint every 5 seconds. Add server-side pagination (`take`/`skip`) from the start.

autogpt_platform/backend/backend/data/invited_user.py
```diff
@@ -0,0 +1,647 @@
...
+            data={
+                "email": normalized_email,
+                "name": normalized_name,
+                "status": prisma.enums.InvitedUserStatus.INVITED,
+                "tallyStatus": prisma.enums.TallyComputationStatus.PENDING,
+            }
+        )
+    except UniqueViolationError:
+        raise PreconditionFailed("An invited user with this email already exists")
+    schedule_invited_user_tally_precompute(invited_user.id)
```
majdyz, line 255: 🤖 **[MEDIUM] No notification is sent to the invited user.** After creating the invite record, nothing emails or otherwise notifies the user. The bundled design doc describes `/signup?invite=<token>` as a primary UX goal. Without an invite link or email, admins must inform users out-of-band — a significant gap for a closed beta.

autogpt_platform/backend/backend/data/invited_user.py
```diff
@@ -0,0 +1,647 @@
...
+        if file_name.endswith(".csv")
+        else _parse_bulk_invite_text(text)
+    )
+
+    if not parsed_rows:
+        raise ValueError("Invite file did not contain any emails")
+
+    if len(parsed_rows) > MAX_BULK_INVITE_ROWS:
+        raise ValueError(
+            f"Invite file contains too many rows. Maximum supported rows: {MAX_BULK_INVITE_ROWS}"
```
majdyz: 🤖 **[LOW] Row limit is enforced after fully parsing the file into memory.** A 1 MB file of valid email addresses (~50,000 rows at ~20 bytes each) will be fully parsed into `_ParsedInviteRow` objects before the `MAX_BULK_INVITE_ROWS` check fires. The byte-size limit provides some protection, but check the row count earlier — ideally during the parse loop.

autogpt_platform/backend/backend/data/invited_user.py
```diff
@@ -0,0 +1,647 @@
...
+            del _tally_seed_tasks[_id]
+
+    task.add_done_callback(_on_done)
+
+
+async def get_or_activate_user(user_data: dict) -> User:
+    auth_user_id = user_data.get("sub")
+    if not auth_user_id:
+        raise NotAuthorizedError("User ID not found in token")
+
```
majdyz, line 644: 🤖 **[CRITICAL] The invite gate is bypassed by the existing Supabase DB trigger.** The `add_user_and_profile_to_platform` trigger (added in `20250205100104_add_profile_trigger`) fires `INSERT INTO platform."User"` immediately when a user signs up via Supabase `auth.users`. When `get_or_activate_user` is called, it finds the user here and returns early — **never reaching the invite check below**. The invite system is non-functional until this trigger is explicitly disabled in this migration. This PR makes no mention of it.

autogpt_platform/backend/backend/data/invited_user.py
```diff
@@ -0,0 +1,647 @@
...
+                },
+            )
+
+            await _ensure_default_profile(
+                auth_user_id,
+                normalized_email,
+                preferred_name,
+                tx,
+            )
+            await _ensure_default_onboarding(auth_user_id, tx)
```
majdyz, line 726: 🤖 **[LOW] Local import used to work around a circular dependency.** Project conventions say to avoid local imports and use top-level imports instead. The circular dependency here (`invited_user` → `user` → `invited_user`) is a sign the module structure needs fixing — for example, moving the `cache_delete` calls to a thin helper module — rather than patching with a local import.

autogpt_platform/backend/backend/api/features/admin/model.py
```diff
@@ -14,3 +18,42 @@ class UserHistoryResponse(BaseModel):
...
+    name: Optional[str] = None
+
+
+class InvitedUserResponse(BaseModel):
+    id: str
+    email: str
+    status: prisma.enums.InvitedUserStatus
+    auth_user_id: Optional[str] = None
+    name: Optional[str] = None
+    tally_understanding: Optional[dict[str, Any]] = None
```
majdyz, line 39: 🤖 **[MEDIUM] Raw exception string is returned in the API response.** `tally_error` stores `str(exc)` from the Tally API call directly. This can contain Tally API keys, internal service URLs, or partial HTTP responses. Even though this endpoint is admin-only, returning raw exception messages over any API is bad practice. Redact or summarize before surfacing to the client.

autogpt_platform/backend/backend/api/features/admin/user_admin_routes.py
```diff
@@ -0,0 +1,161 @@
...
+@router.get(
+    "/invited-users",
+    response_model=InvitedUsersResponse,
+    summary="List Invited Users",
+)
+async def get_invited_users(
+    admin_user_id: str = Security(get_user_id),
+) -> InvitedUsersResponse:
+    logger.info("Admin user %s requested invited users", admin_user_id)
+    invited_users = await list_invited_users()
```
majdyz: 🤖 **[MEDIUM] No pagination — the route returns the entire invite table.** `list_invited_users()` has no `take`/`skip` limit. This endpoint is also polled every 5 seconds by the admin UI. With thousands of invited users this becomes a full table scan on every poll cycle.

autogpt_platform/frontend/src/app/(platform)/admin/users/useAdminUsersPage.ts
```diff
@@ -0,0 +1,197 @@
...
+    useState<BulkInvitedUsersResponse | null>(null);
+  const [pendingInviteAction, setPendingInviteAction] = useState<string | null>(
+    null,
+  );
+
+  const invitedUsersQuery = useGetV2ListInvitedUsers({
+    query: {
+      select: okData,
+      refetchInterval: 5000,
+    },
```
majdyz, line 42: 🤖 **[MEDIUM] 5-second polling interval is too aggressive.** Every admin browser tab with this page open fires a full `GET /api/users/admin/invited-users` (unbounded table scan) every 5 seconds. With N admin tabs and M invited users this is a constant O(N × M) DB load. Use a much longer interval (30–60 s) or replace with a manual refresh button.

autogpt_platform/frontend/src/app/api/openapi.json
```diff
@@ -6646,6 +6646,181 @@
         }
       }
     },
+    "/api/users/admin/invited-users": {
+      "get": {
+        "tags": ["v2", "admin", "users", "admin"],
```
majdyz, line 6651: 🤖 **[LOW] `operationId` contains a space — invalid OpenAPI.** `"getV2List invited users"` and `"postV2Create invited user"` contain literal spaces. Spaces are not allowed in OpenAPI operation IDs and will break any code generator or SDK that consumes this spec. Use camelCase or underscores: `getV2ListInvitedUsers`, `postV2CreateInvitedUser`.

autogpt_platform/backend/backend/data/invited_user.py
```diff
@@ -0,0 +1,647 @@
...
+            data={
+                "tallyStatus": prisma.enums.TallyComputationStatus.FAILED,
+                "tallyError": str(exc),
+            },
+        )
+
+
+def schedule_invited_user_tally_precompute(invited_user_id: str) -> None:
+    existing = _tally_seed_tasks.get(invited_user_id)
+    if existing is not None and not existing.done():
```
majdyz, line 594: 🤖 **[HIGH] `get_or_activate_user` has no caching — DB hit on every authenticated request.** The function it replaces (`get_or_create_user`) was decorated with `@cache_user_lookup` (300 s TTL, 1000-entry LRU). This new version has no cache and is now injected into `update_user_email_route`, `get_user_timezone_route`, `update_user_timezone_route`, `get_preferences`, and `update_preferences`. For every already-activated user, every one of those requests now makes an unconditional DB query just to confirm the user exists — a significant latency regression across all authenticated routes.

0ubbe (dismissed):
Front-end side looks good 🫡

autogpt_platform/backend/backend/api/features/admin/model.py
Swiftyos: We kept prisma.enums.TallyComputationStatus here intentionally so it stays in sync with the DB enum automatically — if we add a new status value to the Prisma schema, both the data layer and the API response pick it up without a manual update. Happy to switch to Literal["PENDING", "RUNNING", "READY", "FAILED"] if you prefer the explicit contract though.

autogpt_platform/backend/backend/api/features/admin/user_admin_routes.py
Swiftyos: his was addressed — the route no longer has inline email validation. Email validation is handled by the EmailStr type in CreateInvitedUserRequest (model.py:24) and by _email_adapter.validate_python() in the bulk flow. The old >3 char check is gone.

autogpt_platform/backend/backend/api/features/admin/user_admin_routes.py
Swiftyos: Addressed. _to_response now uses InvitedUserResponse(**invited_user.model_dump()) which passes the dict directly without serializing to string. This is equivalent to model_validate but uses kwargs unpacking. Happy to switch to InvitedUserResponse.model_validate(invited_user.model_dump()) if you prefer the explicit style.

autogpt_platform/backend/backend/data/tally.py
Swiftyos: Done — _settings = Settings() is instantiated at module level (tally.py:24). The get_business_understanding_input_from_tally function reads from that module-level _settings singleton, so there's no per-call construction cost.

autogpt_platform/backend/migrations/20260310090914_add_invite_system/migration.sql
Swiftyos: I've deleted this line

autogpt_platform/docs/beta-invite-design.md
Swiftyos: done

autogpt_platform/backend/backend/data/invited_user.py
Swiftyos: We added a comment clarifying this (line 33-35). The underlying Tally precompute is idempotent (re-running just overwrites with the same result), and the staleness check (_TALLY_STALE_SECONDS = 300) handles stuck entries after a worker restart. Cross-worker dedup via Celery/Postgres job queue is out of scope for this PR — the worst case is two workers both call the Tally API for the same email, which just results in a duplicate idempotent write. We can move to a proper job queue in a follow-up if needed.

autogpt_platform/backend/backend/data/invited_user.py
Swiftyos: Already reduced from 10 retries to 2 (line 130: for _ in range(2)). With a 6-char hex suffix (uuid4().hex[:6]) the collision probability is ~1 in 16M per attempt, so 2 tries is effectively guaranteed. The transaction holds for a very short window. An INSERT ... ON CONFLICT approach isn't straightforward here since Prisma doesn't support it for this model, but happy to revisit if it becomes a bottleneck.

autogpt_platform/backend/backend/data/invited_user.py
Swiftyos: Addressed. list_invited_users now accepts page and page_size parameters (line 206-208) and uses skip/take with a count query. The route at user_admin_routes.py:69-84 passes pagination params from query string and returns a Pagination object.

autogpt_platform/backend/backend/data/invited_user.py
Swiftyos: Acknowledged — email notification is intentionally out of scope for this PR. The current flow is: admin invites users → admin shares the signup link out-of-band using the usual beta invite flow → user signs up → invite is claimed. We can add automated invite emails as a follow-up feature.

autogpt_platform/backend/backend/data/invited_user.py
Swiftyos: Addressed. The row limit (MAX_BULK_INVITE_ROWS) is now checked during the parse loop itself — both _parse_bulk_invite_csv (line 332) and _parse_bulk_invite_text (line 355) break out of the loop once len(parsed_rows) >= MAX_BULK_INVITE_ROWS. The 1MB byte-size limit in  _decode_bulk_invite_file provides the first line of defense before parsing even begins.

autogpt_platform/backend/backend/data/invited_user.py
Swiftyos: Addressed. We added a new migration (20260311000000_drop_auto_user_trigger/migration.sql) that drops the user_added_to_platform trigger on auth.users and the add_user_and_profile_to_platform() function. User creation is now exclusively handled by get_or_activate_user(), which checks the invite table before creating the platform User row

autogpt_platform/backend/backend/data/invited_user.py
Swiftyos: The local import (line 660) is for get_user_by_id.cache_delete and get_user_by_email.cache_delete — these are only needed at the very end of activation to bust the cache. Moving these to a thin helper module would require extracting the cache decorator and its state, which adds more  complexity than the local import saves. This is a common pattern in the codebase for cache invalidation cross-references. Happy to refactor if you  feel strongly.

autogpt_platform/backend/backend/api/features/admin/user_admin_routes.py
Swiftyos: Addressed. The route now accepts page and page_size query params (lines 71-72), uses server-side pagination, and returns a Pagination object in the response (lines 76-84).

autogpt_platform/frontend/src/app/api/openapi.json
Swiftyos: fixed

autogpt_platform/backend/backend/data/invited_user.py
Swiftyos: This is intentional. The old get_or_create_user in user.py (still present at line 31-54) is decorated with @cache_user_lookup and handles  the already-activated user path. get_or_activate_user is only called from POST /auth/user (v1.py:139) — it is not injected into the  timezone/email/preferences routes. Those routes use get_user_by_id (user.py:57) which IS cached. So for already-activated users hitting POST  /auth/user, the function does a single find_unique by ID (line 587-591) and returns immediately — this is a fast indexed lookup, and adding a cache  here would complicate invite status changes (revoking an invite while a cached user object exists). No latency regression on the other routes.

autogpt_platform/backend/backend/api/features/admin/user_admin_routes.py
Swiftyos: Moved. Added InvitedUserResponse.from_record(record) and BulkInvitedUsersResponse.from_result(result) classmethods on the response models. Removed the standalone _to_response and _to_bulk_response helper functions from the routes file. Used TYPE_CHECKING import to avoid a circular  dependency between the model and data modules.

autogpt_platform/backend/backend/api/features/admin/model.py
Swiftyos: Fixed. The error string is now sanitized before storing: we prefix with the exception class name ({type(exc).__name__}: {exc}), strip any  URLs via regex (https?://\S+ → <url>), and truncate to 200 characters. The full exception is still logged server-side via logger.exception.

autogpt_platform/frontend/src/app/(platform)/admin/users/useAdminUsersPage.ts
Swiftyos: Good call — bumped the polling interval from 5s to 30s. The endpoint is now paginated so each poll is bounded, but 5s was still  unnecessarily aggressive for an admin-only page.

autogpt_platform/backend/.env.default
```diff
@@ -37,6 +37,10 @@ JWT_VERIFY_KEY=your-super-secret-jwt-token-with-at-least-32-characters-long
 ENCRYPTION_KEY=dvziYgz0KSK8FENhju0ZYi8-fRTfAdlz6YLhdB_jhNw=
 UNSUBSCRIBE_SECRET_KEY=HlP8ivStJjmbf6NKi78m_3FnOogut0t5ckzjsIqeaio=
 
+## ===== SIGNUP / INVITE GATE ===== ##
+# Set to true to require an invite before users can sign up
+ENABLE_INVITE_GATE=false
```
majdyz, line 42: wait so now we don't need invite by default or is it only for local?

majdyz: also if we already set the value on settings.py maybe we don't need this

autogpt_platform/backend/backend/api/features/v1.py
```diff
@@ -136,12 +136,10 @@ def _create_file_size_error(size_bytes: int, max_size_mb: int) -> HTTPException:
     dependencies=[Security(requires_user)],
 )
 async def get_or_create_user_route(user_data: dict = Security(get_jwt_payload)):
-    user = await get_or_create_user(user_data)
+    user = await get_or_activate_user(user_data)
```
majdyz, line 139: okay so when do we create it? will it risk now if we don't create on non existent case?

autogpt_platform/backend/schema.prisma
```diff
@@ -73,6 +74,39 @@ model User {
...
+  AuthUser   User?              @relation("InvitedUserAuthUser", fields: [authUserId], references: [id], onDelete: SetNull)
+  name       String?
+
+  tallyUnderstanding Json?
+  tallyStatus        TallyComputationStatus @default(PENDING)
+  tallyComputedAt    DateTime?
+  tallyError         String?
+
+  @@index([status])
+  @@index([tallyStatus])
```
majdyz: index on categorical statuses like this is not useful 95% of the time, unless there is extreme distribution like 1:1000 ratio of invited vs non-invited and we looking specifically on invited.

majdyz: better index on created at and remove the other two

autogpt_platform/backend/backend/data/invited_user.py
```diff
@@ -0,0 +1,715 @@
...
+    if not auth_email:
+        raise NotAuthorizedError("Email not found in token")
+
+    normalized_email = normalize_email(auth_email)
+    user_metadata = user_data.get("user_metadata")
+    metadata_name = (
+        user_metadata.get("name") if isinstance(user_metadata, dict) else None
+    )
+
+    existing_user = await prisma.models.User.prisma().find_unique(
```
majdyz: 🤖 **Performance regression: uncached DB lookup on every API request**

`get_or_activate_user()` is the FastAPI auth dependency that runs on every protected request. For already-activated users (the 99.9% case) the hot path is:

```python
existing_user = await prisma.models.User.prisma().find_unique(where={"id": auth_user_id})
if existing_user is not None:
    return User.from_db(existing_user)
```

This is a raw Prisma call with no caching. The old `get_or_create_user` was `@cache_user_lookup` (300 s TTL, 1000-entry LRU). Every single API call now pays an extra DB round-trip just to identify the caller.

Use the cached helper as the fast path:
```python
from backend.data.user import get_user_by_id
cached = await get_user_by_id(auth_user_id)
if cached is not None:
    return cached
```
Fall through to the full activation logic only on `None`. Since `get_user_by_id` caches `None` misses too, you may need to add a `cache_delete` call after activation (which is already done at line 702 — just move that import to the top).

autogpt_platform/backend/backend/data/invited_user.py
```diff
@@ -0,0 +1,715 @@
...
+        already_created = await prisma.models.User.prisma().find_unique(
+            where={"id": auth_user_id}
+        )
+        if already_created is not None:
+            return User.from_db(already_created)
+        raise RuntimeError(
+            f"UniqueViolationError during activation but user {auth_user_id} not found"
+        )
+
+    from backend.data.user import get_user_by_email, get_user_by_id
```
majdyz: 🤖 **Local import inside function body** — code style rule: all imports must be at the top of the file. Move `from backend.data.user import get_user_by_email, get_user_by_id` to the module-level imports. If this was placed here to avoid a circular import, break the cycle structurally (the `get_user_by_id` fix above would require the same move anyway).

autogpt_platform/backend/migrations/20260310090914_add_invite_system/migration.sql
```diff
@@ -0,0 +1,46 @@
...
+ON "InvitedUser"("email");
+-- CreateIndex
+CREATE UNIQUE INDEX "InvitedUser_authUserId_key"
+ON "InvitedUser"("authUserId");
+-- CreateIndex
+CREATE INDEX "InvitedUser_status_idx"
+ON "InvitedUser"("status");
+-- CreateIndex
+CREATE INDEX "InvitedUser_tallyStatus_idx"
+ON "InvitedUser"("tallyStatus");
```
majdyz, line 41: 🤖 **Missing index on `createdAt` for pagination** — `list_invited_users` queries with `ORDER BY "createdAt" DESC LIMIT/OFFSET`. The migration creates indexes on `status` and `tallyStatus` but not on `createdAt`. Without it, every admin page load is a sequential scan + sort. Add:
```sql
CREATE INDEX "InvitedUser_createdAt_idx" ON "InvitedUser"("createdAt" DESC);
```

autogpt_platform/backend/backend/data/invited_user.py
```diff
@@ -0,0 +1,715 @@
...
+from backend.util.json import SafeJson
+from backend.util.settings import Settings
+
+logger = logging.getLogger(__name__)
+_settings = Settings()
+
+# Process-local dedup for Tally seed tasks.  Does not deduplicate across
+# multiple workers; that is acceptable because the underlying operation is
+# idempotent and the staleness check below handles stuck entries on restart.
+_tally_seed_tasks: dict[str, asyncio.Task] = {}
```
majdyz, line 42: 🤖 **`_tally_seed_tasks` is not shared across uvicorn workers** — with multiple workers (e.g. `uvicorn --workers 4`), each process has its own copy of this dict, so the same invite can trigger N concurrent Tally API calls — one per worker. The DB-level staleness guard (`RUNNING` status + `_TALLY_STALE_SECONDS`) limits the duplicate window on restart, but doesn't prevent simultaneous calls from live workers. Tally has rate limits; for 500-row bulk uploads this could exhaust them. Consider using a Redis-backed distributed lock or a dedicated task queue (Celery/ARQ) for the precompute, or at minimum document this limitation and add a per-invite rate-limit at the DB layer (e.g. skip if `updatedAt` < 5 min ago regardless of `RUNNING` status).

autogpt_platform/backend/backend/data/invited_user.py
```diff
@@ -0,0 +1,715 @@
+import asyncio
```
majdyz, line 1: 🤖 nit: `from __future__ import annotations` is missing here but present in `model.py`. Not required, but keeping it consistent across sibling modules is nice.

autogpt_platform/backend/backend/data/invited_user.py
```diff
@@ -0,0 +1,715 @@
...
+        if any(normalized_row):
+            indexed_rows.append((row_number, normalized_row))
+
+    if not indexed_rows:
+        return []
+
+    header = [cell.lower() for cell in indexed_rows[0][1]]
+    has_header = "email" in header
+    email_index = header.index("email") if has_header else 0
+    name_index: int | None = (
```
majdyz: 🤖 nit: `int | None` union syntax here while the rest of the file uses `Optional[...]` — is okay, just inconsistent with the surrounding style.

autogpt_platform/backend/backend/data/invited_user.py
```diff
@@ -0,0 +1,715 @@
...
+    return [InvitedUserRecord.from_db(iu) for iu in invited_users], total
+
+
+async def create_invited_user(
+    email: str, name: Optional[str] = None
+) -> InvitedUserRecord:
+    normalized_email = normalize_email(email)
+    normalized_name = _normalize_name(name)
+
+    existing_user = await prisma.models.User.prisma().find_unique(
```
majdyz, line 232: 🤖 nit: these two `find_unique` calls (User by email, InvitedUser by email) are independent — could run them concurrently with `asyncio.gather` to save one round-trip. Is okay as sequential though.

autogpt_platform/backend/backend/data/invited_user.py
```diff
@@ -0,0 +1,715 @@
...
+    parsed_rows: list[_ParsedInviteRow] = []
+
+    for row_number, raw_line in enumerate(text.splitlines(), start=1):
+        if len(parsed_rows) >= MAX_BULK_INVITE_ROWS:
+            break
+        line = raw_line.strip()
+        if not line or line.startswith("#"):
+            continue
+
+        parsed_rows.append(
```
majdyz, line 368: 🤖 nit: loop + `append` + `break` for the row limit — could be a list comprehension with a slice:
```python
return [
    _ParsedInviteRow(row_number=i, email=line, name=None)
    for i, raw_line in enumerate(text.splitlines(), 1)
    if (line := raw_line.strip()) and not line.startswith("#")
][:MAX_BULK_INVITE_ROWS]
```
Is okay as-is.

autogpt_platform/backend/backend/api/features/admin/model.py
```diff
@@ -14,3 +23,70 @@ class UserHistoryResponse(BaseModel):
...
+    tally_status: prisma.enums.TallyComputationStatus
+    tally_computed_at: Optional[datetime] = None
+    tally_error: Optional[str] = None
+    created_at: datetime
+    updated_at: datetime
+
+    @classmethod
+    def from_record(cls, record: InvitedUserRecord) -> InvitedUserResponse:
+        return cls.model_validate(record.model_dump())
```
majdyz, line 48: 🤖 nit: `cls.model_validate(record.model_dump())` serialises to a dict and back. Since both models share the same field names, `cls(**record.model_dump())` (or even `cls.model_validate(record)` with Pydantic v2's arbitrary types) avoids the intermediate dict. Is okay as-is.

autogpt_platform/backend/.env.default
Swiftyos: This allows bypassing the invite system, and its primary purpose is for the end-to-end testing, but also useful for local testing.

autogpt_platform/backend/backend/api/features/v1.py
Swiftyos: This function will create a user if:

```python
if not _settings.config.enable_invite_gate  or auth_email.endswith("@agpt.co"):
    return await _open_signup_create_user(
        auth_user_id, normalized_email, metadata_name
    )
```

User records are created when they are invited to the system. This will need to change when we go live

autogpt_platform/backend/backend/data/invited_user.py
Swiftyos: Done

autogpt_platform/backend/backend/data/invited_user.py
Swiftyos: added AsyncClusterLock

autogpt_platform/backend/backend/api/features/admin/model.py
Swiftyos: Using the prisma enum here for type safety — it ensures the value stays in sync with the schema. Happy to switch to a string literal if you prefer though.

autogpt_platform/backend/backend/api/features/admin/user_admin_routes.py
Swiftyos: Good catch — this was fixed by replacing `_redact_email` with the `mask_email()` helper from `tally.py`, which properly handles all email lengths (including ≤2 char locals).

autogpt_platform/backend/backend/api/features/admin/user_admin_routes.py
Swiftyos: Done — moved to `InvitedUserResponse.from_record()` classmethod using `model_validate`.

autogpt_platform/backend/backend/api/features/admin/user_admin_routes.py
Swiftyos: Done — moved to `BulkInvitedUsersResponse.from_result()` classmethod and `InvitedUserResponse.from_record()` classmethod in model.py.

autogpt_platform/backend/backend/data/tally.py
Swiftyos: Fixed — `Settings()` is now instantiated at module level (`_settings = Settings()`) in both `tally.py` and `invited_user.py`.

autogpt_platform/backend/migrations/20260310090914_add_invite_system/migration.sql
Swiftyos: Acknowledged — thanks for flagging. Will leave this for your separate PR.

autogpt_platform/docs/beta-invite-design.md
Swiftyos: Removed in commit c4693f1.

autogpt_platform/backend/backend/data/invited_user.py
Swiftyos: Cross-worker dedup is already handled by `AsyncClusterLock` (Redis-based distributed lock) in `_compute_invited_user_tally_seed()` (lines 509-529). The in-process `_tally_seed_tasks` dict is purely for preventing duplicate `asyncio.Task` objects within the same event loop — it's not meant for cross-process coordination. If Redis is unavailable, the task gracefully skips (line 521).

autogpt_platform/backend/backend/data/invited_user.py
Swiftyos: The bot misread the code — `_generate_unique_profile_username` uses `range(2)` (2 attempts, not 10), each with a UUID suffix making collisions extremely unlikely. It's 2 `find_unique` queries max within the transaction, which is acceptable.

autogpt_platform/backend/backend/data/invited_user.py
Swiftyos: Pagination is already implemented — `list_invited_users()` uses `skip`/`take` and the route accepts `page`/`page_size` query params with defaults (page=1, page_size=50, max 200). The bot was looking at an earlier revision.

autogpt_platform/backend/backend/data/invited_user.py
Swiftyos: Correct — email notifications are intentionally out of scope for this initial PR. The invite system pre-provisions the user record so that when they sign up (communicated out-of-band for now), their account is ready with Tally understanding already seeded. Email invites can be added as a follow-up.

autogpt_platform/backend/backend/data/invited_user.py
Swiftyos: Already addressed — `_parse_bulk_invite_csv()` and `_parse_bulk_invite_text()` both enforce `MAX_BULK_INVITE_ROWS` during the parse loop with an early `break`. The byte-size check (`MAX_BULK_INVITE_FILE_BYTES = 1MB`) also runs before any parsing.

autogpt_platform/backend/backend/data/invited_user.py
Swiftyos: Good catch — this is a known issue flagged by the TODO on line 639: `# TODO: We need to change this functions logic before going live`. The invite gate is behind `enable_invite_gate` (currently off in production). Before enabling it, we'll need to either disable the Supabase `add_user_and_profile_to_platform` trigger or modify the activation flow to check invite status even when the User row already exists. This is intentionally deferred to a follow-up to keep this PR focused on the invite provisioning infrastructure.

autogpt_platform/backend/backend/data/invited_user.py
```diff
@@ -0,0 +1,750 @@
...
+        user_metadata.get("name") if isinstance(user_metadata, dict) else None
+    )
+
+    existing_user = None
+    try:
+        existing_user = await get_user_by_id(auth_user_id)
+    except ValueError:
+        existing_user = None
+    except Exception:
+        logger.exception("Error on get user by id during tally enrichment process")
```
majdyz, line 661: 🤖 nit: Log message says "during tally enrichment process" but this is the user lookup fast path, not tally enrichment. Misleading when reading logs. Maybe "Error looking up user during activation".

autogpt_platform/backend/backend/data/invited_user.py
```diff
@@ -0,0 +1,750 @@
+import asyncio
```
majdyz, line 1: 🤖 nit: `from __future__ import annotations` is present in `model.py` but missing here. Not required, but keeping it consistent across sibling modules is nice.

autogpt_platform/backend/backend/data/invited_user.py
```diff
@@ -0,0 +1,750 @@
...
+    return [InvitedUserRecord.from_db(iu) for iu in invited_users], total
+
+
+async def create_invited_user(
+    email: str, name: Optional[str] = None
+) -> InvitedUserRecord:
+    normalized_email = normalize_email(email)
+    normalized_name = _normalize_name(name)
+
+    existing_user = await prisma.models.User.prisma().find_unique(
```
majdyz, line 232: 🤖 nit: These two `find_unique` calls (User by email at line 232, InvitedUser by email at line 238) are independent — could run concurrently with `asyncio.gather` to save one round-trip. Is okay as sequential though.

autogpt_platform/backend/backend/data/invited_user.py
```diff
@@ -0,0 +1,750 @@
...
+    parsed_rows: list[_ParsedInviteRow] = []
+
+    for row_number, raw_line in enumerate(text.splitlines(), start=1):
+        if len(parsed_rows) >= MAX_BULK_INVITE_ROWS:
+            break
+        line = raw_line.strip()
+        if not line or line.startswith("#"):
+            continue
+
+        parsed_rows.append(
```
majdyz, line 368: 🤖 nit: Loop + `append` + `break` for the row limit — could be a list comprehension with a slice:
```python
return [
    _ParsedInviteRow(row_number=i, email=line, name=None)
    for i, raw_line in enumerate(text.splitlines(), 1)
    if (line := raw_line.strip()) and not line.startswith("#")
][:MAX_BULK_INVITE_ROWS]
```
Is okay as-is.

autogpt_platform/backend/backend/api/features/admin/model.py
majdyz, line 48: 🤖 nit: `cls.model_validate(record.model_dump())` serialises to dict then back. Since both models share field names, `cls.model_validate(record)` (Pydantic v2 accepts arbitrary objects with matching fields) avoids the intermediate dict. Is okay as-is.

autogpt_platform/backend/migrations/20260310090914_add_invite_system/migration.sql
```diff
@@ -0,0 +1,46 @@
+/*
+  Warnings:
+
+  - You are about to drop the column `search` on the `StoreListingVersion` table. All the data in the column will be lost.
```
majdyz, line 4: 🤖 **Unrelated schema change bundled with invite migration** — this migration drops the `search` column from `StoreListingVersion`. That's a separate concern from the invite system and should be in its own migration. If this migration needs to be rolled back, the `search` column data is lost as collateral damage.

autogpt_platform/backend/migrations/20260312093554_update_invites_indexs/migration.sql
```diff
@@ -0,0 +1,9 @@
+-- DropIndex
+DROP INDEX "InvitedUser_status_idx";
+-- DropIndex
+DROP INDEX "InvitedUser_tallyStatus_idx";
+-- DropIndex
+DROP INDEX "UnifiedContentEmbedding_search_idx";
```
majdyz: 🤖 **Unrelated index drop bundled with invite index migration** — `DROP INDEX "UnifiedContentEmbedding_search_idx"` has nothing to do with invite indexes. This should be in a separate migration. Also: is something else depending on this GIN index for search? Dropping it silently could degrade search query performance.

majdyz (commented):
🤖 Second pass — migration concerns + remaining nits.

autogpt_platform/backend/backend/data/invited_user.py
majdyz: 🤖 This was not addressed in the latest push. The `_tally_seed_tasks` dict is still module-level (line 42). For the initial launch with low invite volume this is probably fine, but please add a TODO or track a follow-up to move this to a Redis-backed task queue (or at minimum a DB-level `SELECT ... FOR UPDATE SKIP LOCKED` pattern) before bulk imports go to production scale.

PR #12349, opened 2026-03-09
fix(frontend): buffer workspace file downloads to prevent truncation

majdyz:
## Summary
- Workspace file downloads (images, CSVs, etc.) were silently truncated (~10 KB lost from the end) when served through the Next.js proxy
- Root cause: `new NextResponse(response.body)` passes a `ReadableStream` directly, which Next.js / Vercel silently truncates for larger files
- Fix: fully buffer with `response.arrayBuffer()` before forwarding, and set `Content-Length` from the actual buffer size
- Keeps the auth proxy intact — no signed URLs (which would be public and expire, breaking chat history)

## Root cause verification
Confirmed locally on session `080f27f9-0379-4085-a67a-ee34cc40cd62`:
- Backend `write_workspace_file` logs **978,831 bytes** written
- Direct backend download (`curl localhost:8006/api/workspace/files/.../download`): **978,831 bytes** ✅
- Browser download through Next.js proxy: **truncated** ❌

## Why not signed URLs?
- Signed URLs are effectively public — anyone with the link can download the file (privacy concern)
- Signed URLs expire, but chat history persists — reopening a conversation later would show broken downloads
- Buffering is fine: workspace files are capped at 100 MB, Vercel function memory is 1 GB

## Related
- Discord thread: `#Truncated File Bug` channel
- Related PR #12319 (signed URL approach) — this fix is simpler and preserves auth

## Test plan
- [ ] Download a workspace file (CSV, PNG, any type) through the copilot UI
- [ ] Verify downloaded file size matches the original
- [ ] Verify PNGs open correctly and CSVs have all rows

cc @Swiftyos @uberdot @AdarshRawat1


autogpt_platform/frontend/src/app/api/__generated__/models/responseType.ts
```diff
@@ -9,52 +9,51 @@
 /**
  * Types of tool responses.
  */
-export type ResponseType = typeof ResponseType[keyof typeof ResponseType];
-
+export type ResponseType = (typeof ResponseType)[keyof typeof ResponseType];
 
 // eslint-disable-next-line @typescript-eslint/no-redeclare
 export const ResponseType = {
-  error: 'error',
```
majdyz, line 17: idk why it genereates this, I guess new library update causes this

majdyz:
Idk why [autogpt_platform/frontend/src/app/api/__generated__/models/responseType.ts](https://github.com/Significant-Gravitas/AutoGPT/pull/12349/files/f3245a430f6605a88df0bf034480f5965f62bd11#diff-0f8d89c24aa85da19fd3871dc7fbff8f4d385bce9455d55bca13dc6c9a60f0a9) generates new quotes, I think it's the new library update? but it's harmless

PR #12357, opened 2026-03-10
feat(platform): Add LLM registry database schema and seed data

Bentlybro:
## Summary
Add Prisma schema, migration, and seed data for dynamic LLM model registry - **Part 1 of 5** in incremental implementation.

This PR contains the **database schema foundation and initial seed data** - no business logic, routes, or UI. PRs #12359, #12371, #12467, and #12468 build on this schema to provide the full registry functionality.

### Schema additions (5 models)
- **LlmProvider**: Registry of LLM providers (OpenAI, Anthropic, Groq, etc.)
- **LlmModel**: Individual models with capabilities and metadata
- **LlmModelCost**: Per-model pricing configuration (credit-based or token-based)
- **LlmModelCreator**: Model creators/trainers (OpenAI, Meta, etc.) - distinct from providers
- **LlmModelMigration**: Track model migrations and reverts with custom pricing overrides

### Seed data
- **8 providers**: OpenAI, Anthropic, Groq, OpenRouter, AI/ML API, Ollama, Llama API, v0
- **17 creators**: OpenAI, Anthropic, Meta, Google, Mistral AI, Cohere, DeepSeek, Alibaba, NVIDIA, Vercel, Microsoft, xAI, Perplexity AI, Nous Research, Amazon, Gryphe, Moonshot AI
- **89 models**: All current production models with capabilities, context windows, and output limits
- **89 cost entries**: Credit-based pricing for all models

### Key features
- **Model-specific capabilities** (tools, JSON output, reasoning, parallel tool calls) - capabilities vary per model even within same provider (e.g., Hugging Face)
- **Flexible creator/provider separation** - e.g., Meta model served via Hugging Face or OpenRouter
- **Migration tracking** with custom pricing overrides for seamless model transitions
- **FK constraints** on migration slugs prevent typos in admin data
- **Optimized indexes** for common query patterns (unused preparatory indexes removed)

## Review Feedback Addressed
- ✅ Added FK constraints on `LlmModelMigration` (sourceModelSlug, targetModelSlug → LlmModel.slug)
- ✅ Removed unused `@@index([credentialProvider])` on LlmModelCost
- ✅ Removed redundant `@@index([isReverted])` on LlmModelMigration
- ✅ Added documentation explaining `credentialProvider` field purpose
- ✅ Added complete seed data for all creators and models

## Migrations
- Schema migration: `20260310_add_llm_registry_schema/migration.sql`
- Seed migration: `20260310_seed_llm_registry/migration.sql`
[...]

PR #12359, opened 2026-03-10
feat(platform): Add LLM registry core - DB layer + in-memory cache

Bentlybro:
## Summary
Add LLM registry core implementation - **Part 2 of 3** in incremental rollout.

Builds on PR #12357 (schema foundation) to provide database access layer and in-memory caching for dynamic LLM model management.

### Registry Core (\`backend/data/llm_registry/\`)
- **DB access layer** - Prisma queries with full relation loading (Provider, Costs, Creator)
- **In-memory cache** - Asyncio-locked singleton with atomic refresh
- **Public API** - \`get_model()\`, \`get_all_models()\`, \`get_enabled_models()\`, \`get_schema_options()\`, \`get_default_model_slug()\`
- **Dataclasses** - \`RegistryModel\`, \`RegistryModelCost\`, \`RegistryModelCreator\` (immutable, frozen)
- **ModelMetadata** - Re-exports from \`blocks.llm\` to avoid type collision

### Startup Integration
- Registry refresh before block initialization
- Graceful fallback if refresh fails (no blocks consume registry yet - comes in future PR)

## Review Feedback Addressed
- ✅ Fixed ModelMetadata duplicate type collision (now imports from blocks.llm)
- ✅ Removed \`_json_to_dict()\` helper - use \`dict(value or {})\` inline
- ✅ Added warning when Provider relation is missing (data corruption indicator)
- ✅ Optimized \`get_default_model_slug()\` - single sort pass with \`next()\`
- ✅ Optimized \`_build_schema_options()\` - list comprehension instead of loop
- ✅ Moved \`llm_registry\` import to top-level in rest_api.py
- ✅ Explicit \`max_output_tokens\` fallback to \`context_window\` when null
- ✅ Rebased onto updated schema (PR #12357) with FK constraints

## Known Limitations (Deferred)
- **Multi-worker coordination** - asyncio.Lock only works per-process (Redis cache in future PR)
- **Mutable cache exposure** - Internal dicts accessible (acceptable for now, Redis will fix)

## Testing
- [x] Prisma client generates successfully
- [x] All imports resolve correctly
- [x] Registry refresh loads models from DB
- [x] Cache atomic swap works correctly
- [x] Backend starts successfully with registry

## Stacked PRs
- **PR #12357**: Schema foundation
- **PR #12359**: Registry core (DB layer + cache)  ← YOU ARE HERE
- **PR #12371**: Public read API (GET endpoints)
- **PR #12467** (this): Admin write API (CRUD) 
- **PR #12468** (next): Admin UI

### DIFF
--- a/autogpt_platform/backend/backend/api/rest_api.py
+++ b/autogpt_platform/backend/backend/api/rest_api.py
@@ -1,3 +1,4 @@
+import asyncio
 import contextlib
 import logging
 import platform
@@ -118,17 +119,25 @@ async def lifespan_context(app: fastapi.FastAPI):
 
     AutoRegistry.patch_integrations()
 
-    # Refresh LLM registry before initializing blocks so blocks can use registry data
+    # Load LLM registry before initializing blocks so blocks can use registry data.
+    # Tries Redis first (fast path on warm restart), falls back to DB.
     # Note: Graceful fallback for now since no blocks consume registry yet (comes in PR #5)
-    # When block integration lands, this should fail hard or skip block initialization
     try:
         await backend.data.llm_registry.refresh_llm_registry()
-        logger.info("LLM registry refreshed successfully at startup")
+        logger.info("LLM registry loaded successfully at startup")
     except Exception as e:
         logger.warning(
-            f"Failed to refresh LLM registry at startup: {e}. "
+            f"Failed to load LLM registry at startup: {e}. "
             "Blocks will initialize with empty registry."
         )
+
+    # Start background task so this worker reloads its in-process cache whenever
+    # another worker (e.g. the admin API) refreshes the registry.
+    _registry_subscription_task = asyncio.create_task(
+        backend.data.llm_registry.subscribe_to_registry_refresh(
+            backend.data.llm_registry.refresh_llm_registry
+        )
+    )
 
     await backend.data.block.initialize_blocks()
 
@@ -153,6 +162,12 @@ async def lifespan_context(app: fastapi.FastAPI):
 
     with launch_darkly_context():
         yield
+
+    _registry_subscription_task.cancel()
+    try:
+        await _registry_subscription_task
+    except asyncio.CancelledError:
+        pass
 
     try:
         await shutdown_cloud_storage_handler()
--- a/autogpt_platform/backend/backend/api/rest_api.py
+++ b/autogpt_platform/backend/backend/api/rest_api.py
@@ -119,9 +119,8 @@ async def lifespan_context(app: fastapi.FastAPI):
 
     AutoRegistry.patch_integrations()
 
-    # Load LLM registry before initializing blocks so blocks can use registry data.
-    # Tries Redis first (fast path on warm restart), falls back to DB.
-    # Note: Graceful fallback for now since no blocks consume registry yet (comes in PR #5)
+    _registry_subscription_task: asyncio.Task | None = None
+
     try:
         await backend.data.llm_registry.refresh_llm_registry()
         logger.info("LLM registry loaded successfully at startup")
@@ -131,8 +130,6 @@ async def lifespan_context(app: fastapi.FastAPI):
             "Blocks will initialize with empty registry."
         )
 
-    # Start background task so this worker reloads its in-process cache whenever
-    # another worker (e.g. the admin API) refreshes the registry.
     _registry_subscription_task = asyncio.create_task(
         backend.data.llm_registry.subscribe_to_registry_refresh(
             backend.data.llm_registry.refresh_llm_registry
@@ -146,28 +143,22 @@ async def lifespan_context(app: fastapi.FastAPI):
     try:
         await backend.data.graph.migrate_llm_models(DEFAULT_LLM_MODEL)
     except Exception as e:
-        err_str = str(e)
-        if "AgentNode" in err_str or "does not exist" in err_str:
-            logger.warning(
-                f"migrate_llm_models skipped: AgentNode table not found ({e}). "
-                "This is expected in test environments."
-            )
+        if "AgentNode" in str(e):
+            logger.warning("migrate_llm_models skipped: AgentNode table not found (%s)", e)
         else:
-            logger.error(
-                f"migrate_llm_models failed unexpectedly: {e}",
-                exc_info=True,
-            )
+            logger.error("migrate_llm_models failed unexpectedly: %s", e, exc_info=True)
 
     await backend.integrations.webhooks.utils.migrate_legacy_triggered_graphs()
 
     with launch_darkly_context():
         yield
 
-    _registry_subscription_task.cancel()
-    try:
-        await _registry_subscription_task
-    except asyncio.CancelledError:
-        pass
+    if _registry_subscription_task:
+        _registry_subscription_task.cancel()
+        try:
+            await _registry_subscription_task
+        except asyncio.CancelledError:
+            pass
 
     try:
         await shutdown_cloud_storage_handler()
--- a/autogpt_platform/backend/backend/api/rest_api.py
+++ b/autogpt_platform/backend/backend/api/rest_api.py
@@ -117,6 +117,14 @@ async def lifespan_context(app: fastapi.FastAPI):
 
     AutoRegistry.patch_integrations()
 
+    # Refresh LLM registry before initializing blocks so blocks can use registry data
+    try:
+        from backend.data.llm_registry import refresh_llm_registry
+
+        await refresh_llm_registry()
+    except Exception as e:
+        logger.warning(f"Failed to refresh LLM registry at startup: {e}")
+
     await backend.data.block.initialize_blocks()
 
     await backend.data.user.migrate_and_encrypt_user_integrations()
--- a/autogpt_platform/backend/backend/api/rest_api.py
+++ b/autogpt_platform/backend/backend/api/rest_api.py
@@ -37,6 +37,7 @@ import backend.api.features.workspace.routes as workspace_routes
 import backend.data.block
 import backend.data.db
 import backend.data.graph
+import backend.data.llm_registry
 import backend.data.user
 import backend.integrations.webhooks.utils
 import backend.util.service
@@ -121,9 +122,7 @@ async def lifespan_context(app: fastapi.FastAPI):
     # Note: Graceful fallback for now since no blocks consume registry yet (comes in PR #5)
     # When block integration lands, this should fail hard or skip block initialization
     try:
-        from backend.data.llm_registry import refresh_llm_registry
-
-        await refresh_llm_registry()
+        await backend.data.llm_registry.refresh_llm_registry()
         logger.info("LLM registry refreshed successfully at startup")
     except Exception as e:
         logger.warning(
--- a/autogpt_platform/backend/backend/api/rest_api.py
+++ b/autogpt_platform/backend/backend/api/rest_api.py
@@ -137,10 +137,17 @@ async def lifespan_context(app: fastapi.FastAPI):
     try:
         await backend.data.graph.migrate_llm_models(DEFAULT_LLM_MODEL)
     except Exception as e:
-        logger.warning(
-            f"Failed to migrate LLM models at startup: {e}. "
-            "This is expected in test environments without AgentNode table."
-        )
+        err_str = str(e)
+        if "AgentNode" in err_str or "does not exist" in err_str:
+            logger.warning(
+                f"migrate_llm_models skipped: AgentNode table not found ({e}). "
+                "This is expected in test environments."
+            )
+        else:
+            logger.error(
+                f"migrate_llm_models failed unexpectedly: {e}",
+                exc_info=True,
+            )
 
     await backend.integrations.webhooks.utils.migrate_legacy_triggered_graphs()
 
--- a/autogpt_platform/backend/backend/api/rest_api.py
+++ b/autogpt_platform/backend/backend/api/rest_api.py
@@ -134,7 +134,14 @@ async def lifespan_context(app: fastapi.FastAPI):
 
     await backend.data.user.migrate_and_encrypt_user_integrations()
     await backend.data.graph.fix_llm_provider_credentials()
-    await backend.data.graph.migrate_llm_models(DEFAULT_LLM_MODEL)
+    try:
+        await backend.data.graph.migrate_llm_models(DEFAULT_LLM_MODEL)
+    except Exception as e:
+        logger.warning(
+            f"Failed to migrate LLM models at startup: {e}. "
+            "This is expected in test environments without AgentNode table."
+        )
+
     await backend.integrations.webhooks.utils.migrate_legacy_triggered_graphs()
 
     with launch_darkly_context():
--- a/autogpt_platform/backend/backend/api/rest_api.py
+++ b/autogpt_platform/backend/backend/api/rest_api.py
@@ -118,12 +118,18 @@ async def lifespan_context(app: fastapi.FastAPI):
     AutoRegistry.patch_integrations()
 
     # Refresh LLM registry before initializing blocks so blocks can use registry data
+    # Note: Graceful fallback for now since no blocks consume registry yet (comes in PR #5)
+    # When block integration lands, this should fail hard or skip block initialization
     try:
         from backend.data.llm_registry import refresh_llm_registry
 
         await refresh_llm_registry()
+        logger.info("LLM registry refreshed successfully at startup")
     except Exception as e:
-        logger.warning(f"Failed to refresh LLM registry at startup: {e}")
+        logger.warning(
+            f"Failed to refresh LLM registry at startup: {e}. "
+            "Blocks will initialize with empty registry."
+        )
 
     await backend.data.block.initialize_blocks()
 
--- a/autogpt_platform/backend/backend/data/graph.py
+++ b/autogpt_platform/backend/backend/data/graph.py
@@ -38,7 +38,7 @@ from backend.util.request import parse_url
 from .block import BlockInput
 from .db import BaseDbModel, execute_raw_with_schema
 from .db import prisma as db
-from .db import query_raw_with_schema, transaction
+from .db import execute_raw_with_schema, query_raw_with_schema, transaction
 from .dynamic_fields import is_tool_pin, sanitize_pin_name
 from .includes import AGENT_GRAPH_INCLUDE, AGENT_NODE_INCLUDE, MAX_GRAPH_VERSIONS_FETCH
 from .model import CredentialsFieldInfo, CredentialsMetaInput, is_credentials_field_name
@@ -1667,15 +1667,12 @@ async def migrate_llm_models(migrate_to: LlmModel):
 
     # Update each block
     for id, path in llm_model_fields.items():
-        query = (
-            """
+        query = """
             UPDATE {schema_prefix}"AgentNode"
             SET "constantInput" = jsonb_set("constantInput", $1, to_jsonb($2), true)
             WHERE "agentBlockId" = $3
             AND "constantInput" ? ($4)::text
-            AND "constantInput"->>($4)::text NOT IN """
-            + enum_values
-        )
+            AND "constantInput"->>($4)::text NOT IN """ + enum_values
 
         await execute_raw_with_schema(
             query,
--- a/autogpt_platform/backend/backend/data/llm_registry/model.py
+++ b/autogpt_platform/backend/backend/data/llm_registry/model.py
@@ -1,0 +1,25 @@
+"""Type definitions for LLM model metadata."""
+
+from typing import Literal, NamedTuple
+
+
+class ModelMetadata(NamedTuple):
+    """Metadata for an LLM model.
+
+    Attributes:
+        provider: The provider identifier (e.g., "openai", "anthropic")
+        context_window: Maximum context window size in tokens
+        max_output_tokens: Maximum output tokens (None if unlimited)
+        display_name: Human-readable name for the model
+        provider_name: Human-readable provider name (e.g., "OpenAI", "Anthropic")
+        creator_name: Name of the organization that created the model
+        price_tier: Relative cost tier (1=cheapest, 2=medium, 3=expensive)
+    """
+
+    provider: str
+    context_window: int
+    max_output_tokens: int | None
+    display_name: str
+    provider_name: str
+    creator_name: str
+    price_tier: Literal[1, 2, 3]
--- a/autogpt_platform/backend/backend/data/llm_registry/model.py
+++ b/autogpt_platform/backend/backend/data/llm_registry/model.py
@@ -1,25 +1,9 @@
-"""Type definitions for LLM model metadata."""
+"""Type definitions for LLM model metadata.
 
-from typing import Literal, NamedTuple
+Re-exports ModelMetadata from blocks.llm to avoid type collision.
+In PR #5 (block integration), this will become the canonical location.
+"""
 
+from backend.blocks.llm import ModelMetadata
 
-class ModelMetadata(NamedTuple):
-    """Metadata for an LLM model.
-
-    Attributes:
-        provider: The provider identifier (e.g., "openai", "anthropic")
-        context_window: Maximum context window size in tokens
-        max_output_tokens: Maximum output tokens (None if unlimited)
-        display_name: Human-readable name for the model
-        provider_name: Human-readable provider name (e.g., "OpenAI", "Anthropic")
-        creator_name: Name of the organization that created the model
-        price_tier: Relative cost tier (1=cheapest, 2=medium, 3=expensive)
-    """
-
-    provider: str
-    context_window: int
-    max_output_tokens: int | None
-    display_name: str
-    provider_name: str
-    creator_name: str
-    price_tier: Literal[1, 2, 3]
+__all__ = ["ModelMetadata"]
--- a/autogpt_platform/backend/backend/data/llm_registry/model.py
+++ b/autogpt_platform/backend/backend/data/llm_registry/model.py
@@ -1,9 +1,0 @@
-"""Type definitions for LLM model metadata.
-
-Re-exports ModelMetadata from blocks.llm to avoid type collision.
-In PR #5 (block integration), this will become the canonical location.
-"""
-
-from backend.blocks.llm import ModelMetadata
-
-__all__ = ["ModelMetadata"]
--- a/autogpt_platform/backend/backend/data/llm_registry/registry.py
+++ b/autogpt_platform/backend/backend/data/llm_registry/registry.py
@@ -1,0 +1,227 @@
+"""Core LLM registry implementation for managing models dynamically."""
+
+from __future__ import annotations
+
+import asyncio
+import logging
+from dataclasses import dataclass, field
+from typing import Any
+
+import prisma.models
+
+from backend.data.llm_registry.model import ModelMetadata
+
+logger = logging.getLogger(__name__)
+
+
+def _json_to_dict(value: Any) -> dict[str, Any]:
+    """Convert Prisma Json type to dict, with fallback to empty dict."""
+    if value is None:
+        return {}
+    if isinstance(value, dict):
+        return value
+    # Prisma Json type should always be a dict at runtime
+    return dict(value) if value else {}
+
+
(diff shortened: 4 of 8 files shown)


autogpt_platform/backend/backend/api/rest_api.py
```diff
@@ -117,6 +117,20 @@ async def lifespan_context(app: fastapi.FastAPI):
 
     AutoRegistry.patch_integrations()
 
+    # Refresh LLM registry before initializing blocks so blocks can use registry data
+    # Note: Graceful fallback for now since no blocks consume registry yet (comes in PR #5)
+    # When block integration lands, this should fail hard or skip block initialization
+    try:
+        from backend.data.llm_registry import refresh_llm_registry
```
majdyz: why don't move this to top-level ?

autogpt_platform/backend/backend/data/llm_registry/registry.py
```diff
@@ -0,0 +1,240 @@
...
+from typing import Any
+
+import prisma.models
+
+from backend.data.llm_registry.model import ModelMetadata
+
+logger = logging.getLogger(__name__)
+
+
+def _json_to_dict(value: Any) -> dict[str, Any]:
```
majdyz: You don't need this, you can do `dict(cost.metadata or {})` or simply `cost.metadata or {}`

autogpt_platform/backend/schema.prisma
```diff
@@ -1304,3 +1304,155 @@ model OAuthRefreshToken {
...
+
+model LlmModelCost {
+  id        String      @id @default(uuid())
+  createdAt DateTime    @default(now())
+  updatedAt DateTime    @updatedAt
+  unit      LlmCostUnit @default(RUN)
+
+  creditCost Int // DB constraint: >= 0
+
+  credentialProvider String
```
majdyz, line 1389: What is this? is it a UUID ? where is it being set ?

autogpt_platform/backend/schema.prisma
```diff
@@ -1304,3 +1304,155 @@ model OAuthRefreshToken {
...
+  metadata Json @default("{}")
+
+  llmModelId String
+  Model      LlmModel @relation(fields: [llmModelId], references: [id], onDelete: Cascade)
+
+  // Note: Unique constraints are implemented as partial indexes in migration SQL:
+  // - One for default costs (WHERE credentialId IS NULL)
+  // - One for credential-specific costs (WHERE credentialId IS NOT NULL)
+  // This allows both provider-level defaults and credential-specific overrides
+  @@index([credentialProvider])
```
majdyz: I don't see any query that uses this, and this is not a unique constraint, comment and code mismatch?

autogpt_platform/backend/schema.prisma
```diff
@@ -1304,3 +1304,155 @@ model OAuthRefreshToken {
...
+  customCreditCost Int? // DB constraint: >= 0 when not null
+
+  // Revert tracking
+  isReverted Boolean   @default(false)
+  revertedAt DateTime?
+
+  // Note: Partial unique index in migration SQL prevents multiple active migrations per source:
+  // UNIQUE (sourceModelSlug) WHERE isReverted = false
+  @@index([targetModelSlug])
+  @@index([isReverted])
```
majdyz: indexing boolean is not useful

autogpt_platform/backend/schema.prisma
```diff
@@ -1304,3 +1304,155 @@ model OAuthRefreshToken {
...
+  @@index([sourceModelSlug, isReverted]) // Composite index for active migration queries
```
majdyz, line 1463: what is the query pattern here seach by sourcemodelslug or targetmodelslug or both? I don't see it being used too

autogpt_platform/backend/backend/api/rest_api.py
```diff
@@ -117,6 +117,20 @@ async def lifespan_context(app: fastapi.FastAPI):
 
     AutoRegistry.patch_integrations()
 
+    # Refresh LLM registry before initializing blocks so blocks can use registry data
+    # Note: Graceful fallback for now since no blocks consume registry yet (comes in PR #5)
+    # When block integration lands, this should fail hard or skip block initialization
```
majdyz: 🤖 **[Code Style] Local import inside function.** `from backend.data.llm_registry import refresh_llm_registry` should be a top-level import. The only reason to use a local import here would be to avoid a circular dependency — but since `llm_registry` only imports `prisma.models`, there's no circular dep. Move to the top of the file.

autogpt_platform/backend/backend/data/llm_registry/model.py
```diff
@@ -0,0 +1,25 @@
+"""Type definitions for LLM model metadata."""
+
+from typing import Literal, NamedTuple
+
+
+class ModelMetadata(NamedTuple):
```
majdyz: 🤖 **[Critical] `ModelMetadata` is an exact duplicate of the one in `backend/blocks/llm.py:78`.** Both are `NamedTuple` with identical fields in the same order. This creates a naming collision and a silent incompatibility: code that currently does `from backend.blocks.llm import ModelMetadata` will get a **different type** than code that does `from backend.data.llm_registry import ModelMetadata`, even though they look the same. Type checkers will flag any function accepting one as an error when passed the other.

The PR description says block integration comes in PR #5, but that migration needs to reconcile these two types. The right fix is to have `blocks/llm.py` import from this new canonical location now (or wait until PR #5 to introduce this type).

autogpt_platform/backend/backend/data/llm_registry/registry.py
```diff
@@ -0,0 +1,240 @@
...
+
+@dataclass(frozen=True)
+class RegistryModel:
+    """Represents a model in the LLM registry."""
+
+    slug: str
+    display_name: str
+    description: str | None
+    metadata: ModelMetadata
+    capabilities: dict[str, Any]
```
majdyz: 🤖 **[Medium] Four boolean capability columns exist in the DB schema but are not mapped onto `RegistryModel`.** `supportsTools`, `supportsJsonOutput`, `supportsReasoning`, and `supportsParallelToolCalls` are typed, indexed, and defaulted in `schema.prisma` and `migration.sql`, but `RegistryModel` only has `capabilities: dict[str, Any]`. Callers that need to check tool support must do `model.capabilities.get("supportsTools")` (untyped dict access) instead of `model.supports_tools` (typed field). Either expose them as typed fields on `RegistryModel`, or remove the boolean columns from the DB schema and use only the JSON blob — the dual representation will diverge.

autogpt_platform/backend/backend/data/llm_registry/registry.py
```diff
@@ -0,0 +1,240 @@
...
+            )
+        except Exception as e:
+            logger.error(f"Failed to refresh LLM registry: {e}", exc_info=True)
+            raise
+
+
+def _build_schema_options() -> list[dict[str, str]]:
+    """Build schema options for model selection dropdown. Only includes enabled models."""
+    options: list[dict[str, str]] = []
+    # Only include enabled models in the dropdown options
```
majdyz: 🤖 **[Code Style] Loop-and-append when a list comprehension would do.** Per project conventions, prefer list comprehensions over manual loop-and-append:
```python
def _build_schema_options() -> list[dict[str, str]]:
    return [
        {
            "label": model.display_name,
            "value": model.slug,
            "group": model.metadata.provider,
            "description": model.description or "",
        }
        for model in sorted(_dynamic_models.values(), key=lambda m: m.display_name.lower())
        if model.is_enabled
    ]
```

majdyz (commented):
🤖 **Index coverage note:** The query in `refresh_llm_registry` is a full table scan (`find_many` with no `where` filter). This is intentional for a startup cache warm — it loads everything. The composite index `(providerId, isEnabled)` on `LlmModel` will not help a full-table read; that index is useful for future filtered queries. No issues with index coverage for the current access patterns.

autogpt_platform/backend/backend/data/llm_registry/registry.py
```diff
@@ -0,0 +1,240 @@
...
+                {
+                    "label": model.display_name,
+                    "value": model.slug,
+                    "group": model.metadata.provider,
+                    "description": model.description or "",
+                }
+            )
+    return options
+
+
```
majdyz, line 222: 🤖 **[Code Style] Two full `sorted()` passes over the same dict.** The two `for model in sorted(...)` loops iterate and sort the entire model dict twice. Use `next()` with a generator to short-circuit on first match and sort only once:
```python
def get_default_model_slug() -> str | None:
    models = sorted(_dynamic_models.values(), key=lambda m: m.display_name)
    return (
        next((m.slug for m in models if m.is_recommended and m.is_enabled), None)
        or next((m.slug for m in models if m.is_enabled), None)
    )
```

autogpt_platform/backend/backend/data/llm_registry/registry.py
```diff
@@ -0,0 +1,240 @@
...
+                )
+
+                # Extract creator name (fallback to "Unknown" if no creator)
+                creator_name = (
+                    record.Creator.displayName if record.Creator else "Unknown"
+                )
+
+                # Price tier defaults to 1 if not set
+                price_tier = record.priceTier if record.priceTier in (1, 2, 3) else 1
+
```
majdyz: 🤖 **[Medium] `record.Provider` guard can silently mask data integrity failures.** `providerId` is a NOT NULL FK with `onDelete: Restrict`, and the Prisma query includes `Provider` in the `include`. If `record.Provider` is `None` here, it means either the FK is dangling (data corruption) or the include failed silently. Falling back to `record.providerId` hides the problem and produces a model with `provider_name = provider_display = '<uuid>'` (a raw DB ID shown in the UI). Log a warning at minimum, or raise so the corruption surfaces:
```python
if not record.Provider:
    logger.error("LlmModel %s has no Provider despite NOT NULL FK — skipping", record.slug)
    continue
```

autogpt_platform/backend/backend/data/llm_registry/registry.py
```diff
@@ -0,0 +1,240 @@
...
+    provider_display_name: str
+    is_enabled: bool
+    is_recommended: bool = False
+    costs: tuple[RegistryModelCost, ...] = field(default_factory=tuple)
+    creator: RegistryModelCreator | None = None
+
+
+# In-memory cache (will be replaced with Redis in PR #6)
+_dynamic_models: dict[str, RegistryModel] = {}
+_schema_options: list[dict[str, str]] = []
```
majdyz, line 71: 🤖 **[Medium] Module-level `asyncio.Lock()` is not safe under multi-worker deployments.** Each gunicorn/uvicorn worker process has its own event loop and its own copy of `_lock`. Concurrent `refresh_llm_registry()` calls from two workers cannot coordinate — both will run the full DB query simultaneously and perform the atomic swap independently. This is acknowledged as a known limitation ("Redis cache in PR #6"), but worth flagging: until PR #6 lands, a thundering-herd scenario on startup can fire N×workers DB queries.

autogpt_platform/backend/migrations/20260310_add_llm_registry_schema/migration.sql
```diff
@@ -0,0 +1,148 @@
...
+-- CreateIndex
+CREATE UNIQUE INDEX "LlmModel_slug_key" ON "LlmModel"("slug");
+
+-- CreateIndex
+CREATE INDEX "LlmModel_providerId_isEnabled_idx" ON "LlmModel"("providerId", "isEnabled");
+
+-- CreateIndex
+CREATE INDEX "LlmModel_creatorId_idx" ON "LlmModel"("creatorId");
+
+-- CreateIndex
```
majdyz: 🤖 **[Medium] `LlmModelMigration.sourceModelSlug` and `targetModelSlug` are plain text with no FK constraint.** A typo in admin data creates an active migration record pointing to a non-existent model slug. The partial unique index prevents duplicate active migrations per source, but doesn't validate the slugs reference real models. Either add FK constraints to `LlmModel(slug)` (requires `LlmModel.slug` to have a unique index, which it does), or validate at the application layer before inserting.

autogpt_platform/backend/backend/data/graph.py
```diff
@@ -1670,15 +1670,15 @@ async def migrate_llm_models(migrate_to: LlmModel):
     # Update each block
     for id, path in llm_model_fields.items():
         query = f"""
-            UPDATE {{schema_prefix}}"AgentNode"
+            UPDATE platform."AgentNode"
```
majdyz: 🤖 🔴 **Blocker**: Hardcoded `platform."AgentNode"` breaks non-platform schemas.

The original code used `{schema_prefix}"AgentNode"` which dynamically resolves the schema from `DATABASE_URL`. This PR replaces it with a hardcoded `platform.` prefix. In local development or CI environments where the schema is `public` (no `?schema=platform` in `DATABASE_URL`), this query will fail with `relation "platform.AgentNode" does not exist`.

The `fix_llm_provider_credentials()` function 30 lines above (line 1646) still correctly uses `{schema_prefix}"AgentNode"` — this change creates an inconsistency within the same file.

**Fix**: Revert to using `execute_raw_with_schema` (which was removed from the import). The removal of `execute_raw_with_schema` from the import and replacing it with `db.execute_raw` + hardcoded schema is incorrect.

autogpt_platform/backend/backend/data/graph.py
```diff
@@ -1670,15 +1670,15 @@ async def migrate_llm_models(migrate_to: LlmModel):
...
             WHERE "agentBlockId" = $3
             AND "constantInput" ? ($4)::text
             AND "constantInput"->>($4)::text NOT IN {escaped_enum_values}
             """
 
-        await execute_raw_with_schema(
-            query,
+        await db.execute_raw(
+            query,  # type: ignore - is supposed to be LiteralString
             [path],
```
majdyz, line 1679: 🤖 🟠 **Should Fix**: `# type: ignore` added here violates backend code style.

Per `backend/CLAUDE.md`: "No linter suppressors — no `# type: ignore`, `# noqa`, `# pyright: ignore`; fix the type/code."

This suppressor was introduced to work around the `LiteralString` requirement of `db.execute_raw`, but the real fix is to keep using `execute_raw_with_schema` which already handles this correctly (see blocker comment above). Once that is reverted, this `# type: ignore` goes away.

autogpt_platform/backend/backend/data/llm_registry/registry.py
```diff
@@ -0,0 +1,240 @@
...
+_lock = asyncio.Lock()
+
+
+async def refresh_llm_registry() -> None:
+    """
+    Refresh the LLM registry from the database.
```
majdyz: 🤖 🔴 **Blocker**: No tests for the new `llm_registry` module (240+ lines of new business logic).

Per backend conventions, test files should be colocated (`*_test.py`). This module adds DB queries, data transformation, caching, and default model selection — all testable without a real database.

At minimum, unit tests should cover:
1. **`refresh_llm_registry`** — mock `prisma.models.LlmModel.prisma().find_many()` and verify the cache is populated correctly, including edge cases: empty DB, missing Provider (the warning path), missing Creator (fallback to "Unknown"), null `maxOutputTokens` (fallback to `contextWindow`).
2. **`get_default_model_slug`** — recommended model preferred over enabled, return `None` when empty.
3. **`_build_schema_options`** — only enabled models appear, sorted by `display_name` case-insensitively.
4. **Data integrity** — verify `RegistryModel` is actually frozen/immutable, verify `priceTier` fallback logic for out-of-range values.

The PR description checkboxes say "Registry refresh loads models from DB" and "Cache atomic swap works correctly" but there are no automated tests proving this.

autogpt_platform/backend/backend/data/llm_registry/registry.py
```diff
@@ -0,0 +1,240 @@
...
+    name: str
+    display_name: str
+    description: str | None
+    website_url: str | None
+    logo_url: str | None
+
+
+@dataclass(frozen=True)
+class RegistryModel:
+    """Represents a model in the LLM registry."""
```
majdyz, line 46: 🤖 🟠 **Should Fix**: `frozen=True` dataclass with mutable `dict` and `list` fields gives a false sense of immutability.

`RegistryModel` is `frozen=True` but contains `capabilities: dict[str, Any]` and `extra_metadata: dict[str, Any]`. `RegistryModelCost` similarly has `metadata: dict[str, Any]`. Callers can still mutate these dicts in place:

```python
model = get_model("gpt-4o")
model.capabilities["tools"] = False  # mutates the "frozen" object
```

This defeats the purpose of frozen dataclasses. Options:
1. Use `types.MappingProxyType` to wrap dicts for true read-only access.
2. Or just acknowledge in a comment that frozen here only prevents attribute reassignment, not deep mutation.

Also per backend CLAUDE.md: "**Pydantic models** over dataclass/namedtuple/dict for structured data". These should probably be Pydantic `BaseModel` with `model_config = ConfigDict(frozen=True)` which at least gives you proper validation and serialization.

autogpt_platform/backend/backend/api/rest_api.py
```diff
@@ -117,11 +118,30 @@ async def lifespan_context(app: fastapi.FastAPI):
...
+
     await backend.data.block.initialize_blocks()
 
     await backend.data.user.migrate_and_encrypt_user_integrations()
     await backend.data.graph.fix_llm_provider_credentials()
-    await backend.data.graph.migrate_llm_models(DEFAULT_LLM_MODEL)
+    try:
+        await backend.data.graph.migrate_llm_models(DEFAULT_LLM_MODEL)
+    except Exception as e:
+        logger.warning(
```
majdyz: 🤖 🟠 **Should Fix**: Swallowed exception in `migrate_llm_models` hides real bugs.

This try/except was added to silence test environments, but it also silences real production failures. The comment says "This is expected in test environments without AgentNode table" — but AgentNode is a core table that always exists in production.

If the original call was crashing in tests, that should be fixed at the test level (e.g., mocking the call or ensuring the test DB has the right schema). Wrapping production code in a blanket `except Exception` to fix test failures is the wrong approach.

**Suggestion**: If this is genuinely only for test environments, gate it on an environment check or fix the test setup. Do not silently swallow errors that would indicate a real production problem.

autogpt_platform/backend/backend/data/llm_registry/registry.py
```diff
@@ -0,0 +1,240 @@
...
+                metadata = ModelMetadata(
+                    provider=provider_name,
+                    context_window=record.contextWindow,
+                    max_output_tokens=(
+                        record.maxOutputTokens
+                        if record.maxOutputTokens is not None
+                        else record.contextWindow
+                    ),
+                    display_name=record.displayName,
+                    provider_name=provider_display,
```
majdyz: 🤖 🟡 **Nice to Have**: `priceTier` validation silently coerces invalid values to 1.

```python
price_tier = record.priceTier if record.priceTier in (1, 2, 3) else 1
```

If the DB has `priceTier = 0` or `priceTier = 5` (bad data), this silently coerces to 1 with no log warning. Since you already log a warning for missing Provider (data corruption indicator), it would be consistent to log a warning here too for out-of-range values.

Also, `ModelMetadata.price_tier` is typed as `Literal[1, 2, 3]`. If the DB column allows other integers, this silent coercion is the only safety net — worth a log line.

autogpt_platform/backend/backend/data/llm_registry/model.py
```diff
@@ -0,0 +1,9 @@
+"""Type definitions for LLM model metadata.
+
+Re-exports ModelMetadata from blocks.llm to avoid type collision.
+In PR #5 (block integration), this will become the canonical location.
+"""
+
+from backend.blocks.llm import ModelMetadata
```
majdyz: 🤖 🟡 **Nice to Have**: `model.py` is a single re-export — unnecessary indirection.

This entire file is:
```python
from backend.blocks.llm import ModelMetadata
__all__ = ["ModelMetadata"]
```

The comment says "In PR #5 (block integration), this will become the canonical location" — but right now it does nothing except add an extra import hop. If you plan to move `ModelMetadata` here later, do it in that PR. Until then, `registry.py` can just import directly from `backend.blocks.llm`.

Adding a file "for the future" that currently serves no purpose increases maintenance surface. If this is important for the PR stack, note it in the PR description as a placeholder.

autogpt_platform/backend/backend/data/llm_registry/registry.py
```diff
@@ -0,0 +1,240 @@
...
+def get_all_models() -> list[RegistryModel]:
+    """Get all models from the registry (including disabled)."""
+    return list(_dynamic_models.values())
+
+
+def get_enabled_models() -> list[RegistryModel]:
+    """Get only enabled models from the registry."""
+    return [model for model in _dynamic_models.values() if model.is_enabled]
+
+
```
majdyz, line 237: 🤖 🟠 **Should Fix**: `get_schema_options()` returns a direct reference to the module-level `_schema_options` list.

```python
def get_schema_options() -> list[dict[str, str]]:
    return _schema_options
```

Callers can mutate the internal cache:
```python
opts = get_schema_options()
opts.append({"label": "hacked", "value": "bad"})
# _schema_options is now corrupted for all future callers
```

Same issue with `get_all_models()` and `get_enabled_models()` — they return `list(...)` which creates a new list, but the `RegistryModel` objects inside are still shared references (and their dict fields are mutable as noted in my other comment).

For `get_schema_options`, at minimum return a copy: `return list(_schema_options)` or `return [dict(d) for d in _schema_options]`.

autogpt_platform/backend/backend/data/llm_registry/registry.py
majdyz: 🤖 🟠 **Should Fix**: `refresh_llm_registry()` is ~80 lines — well over the 40-line limit.

Per `backend/CLAUDE.md`: "keep functions under ~40 lines; extract named helpers when a function grows longer."

This function does three distinct things:
1. Fetch records from the DB
2. Transform each record into a `RegistryModel` (the bulk — ~60 lines)
3. Swap the cache

Extract the record-to-model transformation into a helper like `_record_to_registry_model(record) -> RegistryModel`. This will also make it independently testable.

majdyz (commented):
## Review

Thanks for iterating on the feedback -- the ModelMetadata re-export, removal of `_json_to_dict`, `unit` field on `RegistryModelCost`, `next()` optimization on `get_default_model_slug`, and list comprehension in `_build_schema_options` all look good. The PR is in a much better state now.

A few remaining items below.

### 1. graph.py changes don't belong in this PR

The `graph.py` diff does two things:
- Replaces `execute_raw_with_schema` with `db.execute_raw` and hardcodes `platform."AgentNode"` instead of using `{schema_prefix}"AgentNode"`.
- Removes the `execute_raw_with_schema` import (but keeps `query_raw_with_schema`, which is still used by `fix_llm_provider_credentials`).

The hardcoded `platform.` prefix breaks environments where the schema is not `"platform"` (e.g. `public` or a custom test schema). `execute_raw_with_schema` exists specifically to handle this. This change also has nothing to do with the LLM registry core -- it's fixing the `migrate_llm_models` function. If it's intentional, it should be in a separate PR with its own context.

Also, the `try/except` around `migrate_llm_models` in `rest_api.py` is a workaround for the `AgentNode` table not existing in test databases. That's a legitimate fix, but it's unrelated to the registry feature and should also be split out.

### 2. `capabilities` as `dict[str, Any]` vs typed fields

I flagged this before but I still think it deserves a decision. The DB schema defines typed boolean columns (`supportsTools`, `supportsJsonOutput`, `supportsReasoning`, `supportsParallelToolCalls`) with defaults and indexes. But `RegistryModel` only has `capabilities: dict[str, Any]`, forcing callers to do untyped `.get()` access. The two representations will diverge unless you either:
- Expose them as typed fields on `RegistryModel` (preferred), or
- Remove the individual boolean columns from the schema and only use the JSON blob.

This is the kind of thing that's easier to fix now than after downstream PRs start building on it.

### 3. `_lock` scope under multi-worker

Acknowledged in the PR description as a known limitation, and the comment in code is helpful. Just want to confirm: when PR #6 (Redis cache) lands, will the startup sequence change? Right now multiple workers all call `refresh_llm_registry()` on startup -- without coordination that's N identical DB queries. Not blocking, but worth a `# TODO` comment with the PR reference so it doesn't get lost.

### 4. `_schema_options` returns mutable list reference

`get_schema_options()` returns `_schema_options` directly. Since the list contains dicts, callers can mutate the cache:
```python
opts = get_schema_options()
opts[0]["label"] = "hacked"  # mutates global cache
```
A cheap fix: `return list(_schema_options)` for a shallow copy (the inner dicts are still mutable but at least the list itself is safe). Or return a `tuple`. The PR description acknowledges mutable cache exposure, but this specific one is a one-line fix.

### 5. Minor: f-strings in log statements

Per the backend CLAUDE.md code style, `debug` statements should use `%s` for deferred interpolation. The `info` and `warning` f-strings are fine, but there are a few `logger.error(f"...")` calls that might benefit from deferred formatting too (though `error` level is usually always emitted, so it's less critical). Not blocking.

---

**Overall**: The registry implementation is clean and well-structured. The main concern is the `graph.py` changes that should be split out. The rest is in good shape -- once the unrelated changes are removed, this is close to mergeable.

majdyz (changes requested):
## Review Summary — REQUEST_CHANGES

### Blockers (must fix before merge)

1. **Hardcoded `platform."AgentNode"` in `graph.py`** — The PR replaces `execute_raw_with_schema` (which dynamically resolves the DB schema from `DATABASE_URL`) with `db.execute_raw` and a hardcoded `platform.` prefix. This will break in any environment where the schema is not `platform` (e.g., local dev with `public` schema). The function 30 lines above (`fix_llm_provider_credentials`) still correctly uses `{schema_prefix}` — this creates an inconsistency within the same file.

2. **No tests for 240+ lines of new business logic** — The `llm_registry` module adds DB queries, data transformation, caching, and default-model selection, all entirely untested. The PR checkboxes claim manual verification, but there are no automated tests. Per backend conventions, `*_test.py` files should be colocated.

### Should Fix

3. **`# type: ignore` added to `graph.py`** — Violates backend code style ("no linter suppressors"). This goes away if the `execute_raw_with_schema` revert in (1) is applied.

4. **Swallowed exception in `migrate_llm_models` try/except** — A blanket `except Exception` was added to silence test environments, but it also hides real production failures. Fix the test setup instead of wrapping production code.

5. **`frozen=True` dataclass with mutable `dict` fields** — `RegistryModel.capabilities`, `RegistryModel.extra_metadata`, and `RegistryModelCost.metadata` are all `dict[str, Any]` on "frozen" dataclasses. Callers can mutate these in place, defeating the immutability intent. Also, backend CLAUDE.md prescribes Pydantic models over dataclasses.

6. **`get_schema_options()` returns a direct mutable reference** to the module-level `_schema_options` list. Callers can corrupt the cache.

7. **`refresh_llm_registry()` is ~80 lines** — Well over the 40-line backend guideline. Extract the record-to-model transformation into a named helper.

### Nice to Have

8. **`model.py` is a single re-export** — Unnecessary indirection until the future PR that moves `ModelMetadata` here.

9. **`priceTier` coercion is silent** — Out-of-range values default to 1 with no log warning, inconsistent with the Provider-missing warning.

### Previously Raised (by me and coderabbitai) — Verify These Are Fixed

Several issues were raised in earlier review rounds. The PR description says they were addressed but please confirm with a screenshot or test output:
- ModelMetadata type collision fix
- `_json_to_dict` removal
- Provider warning
- `get_default_model_slug` optimization
- `_build_schema_options` list comprehension
- Top-level import in `rest_api.py`
- `max_output_tokens` fallback

**Please provide a screenshot of the backend starting successfully with the registry populated, and add unit tests before requesting re-review.**

majdyz:

Thanks for the contribution!

ntindle (changes requested):
valid ai comments

ntindle:
Closing in favor of the LLM registry restack: the core cache layer is now #13606 (fresh re-cut onto current dev — the original migrations and seed data had drifted ~4 months). The design and much of the code here carried over directly; @Bentlybro is credited as co-author on the carried commits. Full stack starts at #13605. Thanks for the groundwork — the reviewed schema and cache design survived contact with the restack almost unchanged.

PR #12362, opened 2026-03-11
fix(platform): fix image delete button on EditAgentForm

Abhi1992002:
### Summary
  - SECRT-2094: Fix store image delete button accidentally submitting the edit form — the remove image <button> in ThumbnailImages.tsx was missing type="button", causing it to act as a form submit inside the EditAgentForm. This closed the modal and showed a success toast without the user clicking "Update submission".
https://github.com/user-attachments/assets/86cbdd7d-90b1-473c-9709-e75e956dea6b

###  Changes
  - `frontend/.../ThumbnailImages.tsx` — added type="button" to image remove button


autogpt_platform/backend/backend/api/features/store/db.py
```diff
@@ -670,6 +670,28 @@ async def create_store_submission(
...
+                        ]
+                    },
+                },
+            )
+            if existing_for_version:
+                raise store_exceptions.ListingExistsError(
+                    f"Agent v{graph_version} already has an active submission. "
+                    "Please update your agent before resubmitting."
+                )
+
```
Abhi1992002: @majdyz Is it okay?? , or should I add a database constraint for it?? maybe we could get a race condition here (maybe the user sends the request twice simultaneously)

Abhi1992002: If we add database constraint, we’ll need to soft delete the duplicate data.

autogpt_platform/backend/backend/api/features/store/db.py
```diff
@@ -670,6 +670,28 @@ async def create_store_submission(
                     for slv in existing_listing.Versions
                 )
```
majdyz: 🤖 **[Medium] TOCTOU race condition — the check does not prevent duplicates under concurrent requests.** This `find_first` runs inside the transaction, but Prisma uses READ COMMITTED isolation by default. Two simultaneous submissions for the same `(agentGraphId, agentGraphVersion)` can both read "no existing" before either has committed, then both proceed to create a submission.

The author already flagged this in the review comments. The correct fix is a **partial unique index** on the DB side:
```sql
CREATE UNIQUE INDEX "slv_active_version_unique"
  ON "StoreListingVersion"("agentGraphId", "agentGraphVersion")
  WHERE "isDeleted" = false
    AND "submissionStatus" IN ('PENDING', 'APPROVED');
```
This makes the constraint atomic and lets the DB handle the race with a `UniqueViolationError` that the application can catch and surface as `ListingExistsError`.

autogpt_platform/backend/backend/api/features/store/db.py
```diff
@@ -670,6 +670,28 @@ async def create_store_submission(
...
+            ).find_first(
+                where={
+                    "agentGraphId": graph_id,
+                    "agentGraphVersion": graph_version,
+                    "isDeleted": False,
+                    "submissionStatus": {
+                        "in": [
+                            prisma.enums.SubmissionStatus.PENDING,
+                            prisma.enums.SubmissionStatus.APPROVED,
+                        ]
```
majdyz: 🤖 **[Medium] PENDING case now blocks what the existing code was designed to replace.** The code below this check (the `graph_has_pending_submissions` deletion path) deletes a PENDING submission before creating a new version. With this new check, a PENDING submission for the same `(agentGraphId, agentGraphVersion)` throws `ListingExistsError` before reaching that deletion logic — making the deletion path unreachable for same-version resubmissions.

If a user submits version 5 (PENDING) and then edits *metadata only* (same agent version) and tries to resubmit, they now get a hard error instead of an in-place update. Was this intentional? If yes, document the product decision and remove the now-unreachable pending-deletion path. If not, exclude `PENDING` from this check and keep the existing replace logic.

autogpt_platform/backend/backend/api/features/store/db_test.py
```diff
@@ -251,6 +251,7 @@ async def test_create_store_submission(mocker):
     mock_sl.return_value.find_unique = mocker.AsyncMock(return_value=None)
```
majdyz, line 252: 🤖 **[Low] Only the happy path is tested.** The new mock adds `find_first = AsyncMock(return_value=None)` — the "no duplicate" case. There's no test for when `find_first` returns an existing record and `ListingExistsError` is expected to be raised. Add a test that verifies the duplicate rejection path:
```python
async def test_create_store_submission_rejects_duplicate_version(mocker):
    mock_slv.return_value.find_first = mocker.AsyncMock(return_value=mock_version)  # duplicate exists
    with pytest.raises(store_exceptions.ListingExistsError):
        await create_store_version(...)
```

autogpt_platform/backend/backend/api/features/store/db.py
majdyz: I think we found a bigger issue: we shouldn't create multiple submission for the same agent in the first place, even with different version. if the user decided to create another submission we should route them to edit the submission

Abhi1992002:
removing these store submission–related changes in follow up PR after consulting the product team.

PR #12374, opened 2026-03-11
feat(copilot): generate personalized quick-action prompts from Tally business understanding

Swiftyos:
During Tally data extraction, the system now also generates personalized quick-action prompts as part of the existing LLM extraction call (configurable model, defaults to GPT-4o-mini, `temperature=0.0`). The prompt asks the LLM for 5 candidates, then the code validates (filters prompts >20 words) and keeps the top 3. These prompts are stored in the existing `CoPilotUnderstanding.data` JSON field (at the top level, not under `business`) and served to the frontend via a new API endpoint. The copilot chat page uses them instead of hardcoded defaults when available.

- Builds on #12347

### Changes 🏗️

**Backend – Data models** (`understanding.py`):
- Added `suggested_prompts` field to `BusinessUnderstandingInput` (optional) and `BusinessUnderstanding` (default empty list)
- Updated `from_db()` to deserialize `suggested_prompts` from top-level of the data JSON
- Updated `merge_business_understanding_data()` with overwrite strategy for prompts (full replace, not append)
- `format_understanding_for_prompt()` intentionally does **not** include `suggested_prompts` — they are UI-only

**Backend – Prompt generation** (`tally.py`):
- Extended `_EXTRACTION_PROMPT` to request 5 suggested prompts alongside the existing business understanding fields — all extracted in a single LLM call (`temperature=0.0`)
- Post-extraction validation filters out prompts exceeding 20 words and slices to the top 3
- Model is now configurable via `tally_extraction_llm_model` setting (defaults to `openai/gpt-4o-mini`)

**Backend – API endpoint** (`routes.py`):
- Added `GET /api/chat/suggested-prompts` (auth required)
- Returns `{prompts: string[]}` from the user's cached business understanding (48h Redis TTL)
- Returns empty array if no understanding or no prompts exist

**Frontend** (`EmptySession/`):
- `helpers.ts`: Extracted defaults to `DEFAULT_QUICK_ACTIONS`, `getQuickActions()` now accepts optional custom prompts and falls back to defaults
- `EmptySession.tsx`: Calls `useGetV2GetSuggestedPrompts` hook (`staleTime: Infinity`) and passes results to `getQuickActions()` with hardcoded fallback
- Fixed `useEffect` resize handler that previously used `window.innerWidth` as a dependency (re-ran every render); now uses a proper resize event listener
- Added skeleton loading state while prompts are being fetched

**Generated** (`__generated__/`):
- Regenerated Orval API client with new endpoint types and hooks

#### For code changes:
- [x] I have made a test plan
  - [x] Backend format + lint + pyright pass
  - [x] Frontend format + lint pass
  - [x] All existing tally tests pass (28/28)
  - [x] All chat route tests pass (9/9)
  - [x] All invited_user tests pass (7/7)
  - [x] E2E: New user with tally data sees custom prompts on copilot page
  - [x] E2E: User without tally data sees hardcoded default prompts
  - [x] E2E: Clicking a custom prompt sends it as a chat message

### DIFF
--- a/autogpt_platform/backend/backend/data/tally.py
+++ b/autogpt_platform/backend/backend/data/tally.py
@@ -394,7 +394,7 @@ async def extract_business_understanding_from_tally(
         if short_prompts:
             cleaned["suggested_prompts"] = short_prompts
         else:
-            # We dont want to add a None value suggested_prompts filed
+            # We dont want to add a None value suggested_prompts field
             cleaned.pop("suggested_prompts", None)
     else:
         # suggested_prompts must be a list - removing it as its not here


autogpt_platform/backend/backend/data/tally.py
```diff
@@ -376,6 +379,19 @@ async def extract_business_understanding(
...
+
+    # Validate suggested_prompts: filter >20 words, keep top 3
+    raw_prompts = cleaned.get("suggested_prompts", [])
+    if isinstance(raw_prompts, list):
+        valid = [
+            p.strip()
+            for p in raw_prompts
+            if isinstance(p, str) and len(p.strip().split()) <= 20
+        ]
+        cleaned["suggested_prompts"] = valid[:3] if valid else None
```
majdyz: this logic is so confusing and surprising for reader:

if no valid <= 20 words -> valid is empty -> None -> don't overwrite. if we plan to always replace then use valide[:3] either way (and what if it's < 3 btw, will it break). If the intent is "preserve old good prompts on bad extraction", add a comment explaining this

autogpt_platform/frontend/src/app/(platform)/copilot/components/EmptySession/EmptySession.tsx
```diff
@@ -33,15 +35,29 @@ export function EmptySession({
...
   );
 
   useEffect(() => {
-    setInputPlaceholder(getInputPlaceholder(window.innerWidth));
-  }, [window.innerWidth]);
+    function handleResize() {
+      setInputPlaceholder(getInputPlaceholder(window.innerWidth));
+    }
+    handleResize();
+    window.addEventListener("resize", handleResize);
```
majdyz: this is dangerous, resize fires dozens of times per second during a drag, each triggering setState, which en up trigger a re-render. Since getInputPlaceholder only toggles at the 768px breakpoint, use window.matchMedia instead, fires once when crossing the boundary:

this is what AI generates for me
```
const mql = window.matchMedia("(min-width: 768px)");
function handleChange() {
  setInputPlaceholder(getInputPlaceholder(mql.matches ? 768 : 0));
}
handleChange();
mql.addEventListener("change", handleChange);
return () => mql.removeEventListener("change", handleChange);
```

or check with @0ubbe on this

autogpt_platform/backend/backend/data/tally.py
Swiftyos: I've updated the logic and renamed the function for clarity.

autogpt_platform/frontend/src/app/(platform)/copilot/components/EmptySession/EmptySession.tsx
Swiftyos: I've update this, thank you

majdyz (approved):

PR #12123, opened 2026-02-16
fix(frontend): improve UX for expired or duplicate password reset links

Bentlybro:
## Summary
Improves the user experience when a password reset link has expired or been used, replacing the confusing generic error with a clean, helpful message.

## Changes
- Added `ExpiredLinkMessage` component that displays a user-friendly error state
- Updated reset password page to detect expired/used links from:
  - Supabase error format (`error=access_denied&error_code=otp_expired&error_description=...`)
  - Internal clean format (`error=link_expired`)
- Enhanced callback route to detect and map expired/invalid link errors
- Clear, actionable UI with:
  - Friendly error message explaining what happened
  - "Send Me a New Link" button to request a new reset email
  - Login link for users who already have access

## Before
Users saw a confusing URL with error parameters and an unclear form:
```
/reset-password?error=access_denied&error_code=otp_expired&error_description=Email+link+is+invalid+or+has+expired
```

## After
Users see a clean, helpful message explaining the issue and how to fix it.

image

Closes SECRT-1369

#### For code changes:
- [x] I have clearly listed my changes in the PR description
- [x] I have made a test plan
- [x] I have tested my changes according to the test plan:
  - [ ] Navigate to `/reset-password?error=link_expired` and verify the ExpiredLinkMessage component appears
  - [ ] Click "Send Me a New Link" and verify the email form appears
  - [ ] Navigate to `/reset-password?error=access_denied&error_code=otp_expired` and verify same behavior

<details><summary><h3>Greptile Summary</h3></summary>

Improved password reset UX by adding an `ExpiredLinkMessage` component that displays when users follow expired or already-used reset links. The implementation detects expired link errors from Supabase (`error_code=otp_expired`) and internal format (`error=link_expired`), replacing confusing URL parameters with a clean message.

**Key changes:**
- Added error detection logic in both callback route and reset password page to identify expired/invalid links
- Created new `ExpiredLinkMessage` component with friendly messaging
- Enhanced error handling to differentiate between expired links and other errors

**Issues found:**
- The "Send Me a New Link" button misleadingly suggests it will send an email, but it only reveals the email form - user must still enter email and submit
- `access_denied` error detection may be too broad and could incorrectly classify non-expired errors as expired links
</details>

<details><summary><h3>Confidence Score: 3/5</h3></summary>

- This PR improves UX but has logic issues that could mislead users
[...]

PR #12384, opened 2026-03-12
fix(frontend): pass through Supabase error params in password reset callback

Otto-AGPT:
When Supabase rejects a password reset token (expired, already used, etc.), it redirects to the callback URL with `error`, `error_code`, and `error_description` params instead of a `code`. Previously, the callback only checked for `!code` and returned a generic "Missing verification code" error, swallowing the actual Supabase error.

This meant the `ExpiredLinkMessage` UX (added in SECRT-1369 / #12123) was never triggered for these cases — users just saw the email input form again with no explanation.

Now the callback reads Supabase's error params and forwards them to `/reset-password`, where the existing expired link detection picks them up correctly.

**Note:** This doesn't fix the root cause of Pwuts's token expiry issue (likely link preview/prefetch consuming the OTP), but it ensures users see the proper "link expired" message with a "Request new link" button instead of a confusing silent redirect.

---
Co-authored-by: Reinier van der Leer (@Pwuts) <pwuts@agpt.co>

### DIFF
--- a/autogpt_platform/frontend/src/app/api/auth/callback/reset-password/route.ts
+++ b/autogpt_platform/frontend/src/app/api/auth/callback/reset-password/route.ts
@@ -15,27 +15,17 @@ export async function GET(request: NextRequest) {
     const errorCode = searchParams.get("error_code");
     const errorDescription = searchParams.get("error_description");
 
-    if (error || errorCode) {
-      const isExpiredOrUsed =
-        error === "access_denied" ||
-        errorCode === "otp_expired" ||
-        errorDescription?.toLowerCase().includes("expired") ||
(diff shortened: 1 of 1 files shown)


Pwuts:
> **Note:** This doesn't fix the root cause of Pwuts's token expiry issue (likely link preview/prefetch consuming the OTP), but it ensures users see the proper "link expired" message with a "Request new link" button instead of a confusing silent redirect.

I'm running Gnome 49 with [NiffirgkcaJ/all-in-one-clipboard](https://github.com/NiffirgkcaJ/all-in-one-clipboard), which seems to prefetch the URL when I copy it and that's why it doesn't work when I paste it in my browser.

NiffirgkcaJ/all-in-one-clipboard#41

PR #12390, opened 2026-03-12
fix(backend/copilot): auto-parse structured data and robust type coercion

majdyz:
The copilot's `@@agptfile:` reference system always produces strings when expanding
file references. This breaks blocks that expect structured types — e.g.
`GoogleSheetsWriteBlock` expects `values: list[list[str]]`, but receives a raw CSV
string instead. Additionally, the copilot's input coercion was duplicating logic from
the executor instead of reusing the shared `convert()` utility, and the coercion had
no type-aware gating — it would always call `convert()`, which could incorrectly
transform values that already matched the expected type (e.g. stringifying a valid
`list[str]` in a `str | list[str]` union).

### Changes 🏗️

**Structured data parsing for `@@agptfile:` bare references:**
- When an entire tool argument value is a bare `@@agptfile:` reference, the resolved
  content is now auto-parsed: JSON → native types, CSV/TSV → `list[list[str]]`
- Embedded references within larger strings still do plain text substitution
- Updated copilot system prompt to document the structured data capability

**Shared type coercion utility (`coerce_inputs_to_schema`):**
- Extracted `coerce_inputs_to_schema()` into `backend/util/type.py` — shared by both
  the executor's `validate_exec()` and the copilot's `execute_block()`
- Uses Pydantic `model_fields` (not `__annotations__`) to include inherited fields
- Added `_value_satisfies_type()` gate: only calls `convert()` when the value doesn't
  already match the target type, including recursive inner-element checking for generics

**`_value_satisfies_type` — recursive type checking:**
- Handles `Any`, `Optional`, `Union`, `list[T]`, `dict[K,V]`, `set[T]`, `tuple[T, ...]`,
  heterogeneous `tuple[str, int, bool]`, bare generics, nested generics
- Guards against non-runtime origins (`Literal`, etc.) to prevent `isinstance()` crashes
- Returns `False` (not `True`) for unhandled generic origins as a safe fallback

**Test coverage:**
- 51 new tests for `_value_satisfies_type` and `coerce_inputs_to_schema` in `type_test.py`
- 8 new tests for `execute_block` type coercion in `helpers_test.py`

#### For code changes:
- [x] I have made a test plan
  - [x] All existing file_ref tests pass
  - [x] All new type_test.py tests pass (51 tests covering _value_satisfies_type and coerce_inputs_to_schema)
  - [x] All new helpers_test.py tests pass (8 tests covering execute_block coercion)
  - [x] `poetry run format` passes clean
  - [x] `poetry run lint` passes clean
  - [x] Pyright type checking passes

### DIFF
--- a/autogpt_platform/backend/backend/util/type.py
+++ b/autogpt_platform/backend/backend/util/type.py
@@ -269,13 +269,16 @@ def _value_satisfies_type(value: Any, target: Any) -> bool:
 
     # Generic container type (e.g. list[str], dict[str, int])
     if origin is not None:
+        # Guard: origin may not be a runtime type (e.g. Literal)
+        if not isinstance(origin, type):
+            return False
         if not isinstance(value, origin):
             return False
         args = get_args(target)
         if not args:
             return True
         # Check inner elements satisfy the type args
-        if _is_type_or_subclass(origin, list) and args:
+        if _is_type_or_subclass(origin, list):
             return all(_value_satisfies_type(v, args[0]) for v in value)
         if _is_type_or_subclass(origin, dict) and len(args) >= 2:
             return all(
@@ -286,7 +289,7 @@ def _value_satisfies_type(value: Any, target: Any) -> bool:
             _is_type_or_subclass(origin, set) or _is_type_or_subclass(origin, frozenset)
         ) and args:
             return all(_value_satisfies_type(v, args[0]) for v in value)
-        if _is_type_or_subclass(origin, tuple) and args:
+        if _is_type_or_subclass(origin, tuple):
             # Homogeneous tuple[T, ...] — single type + Ellipsis
             if len(args) == 2 and args[1] is Ellipsis:
                 return all(_value_satisfies_type(v, args[0]) for v in value)
@@ -294,7 +297,8 @@ def _value_satisfies_type(value: Any, target: Any) -> bool:
             if len(value) != len(args):
                 return False
             return all(_value_satisfies_type(v, t) for v, t in zip(value, args))
-        return True
+        # Unhandled generic origin — fall through to convert()
+        return False
 
     # Simple type (e.g. str, int)
     if isinstance(target, type):
--- a/autogpt_platform/backend/backend/util/type_test.py
+++ b/autogpt_platform/backend/backend/util/type_test.py
@@ -1,4 +1,4 @@
-from typing import Any, List, Optional
+from typing import Any, List, Literal, Optional
 
 from pydantic import BaseModel
 
@@ -168,6 +168,11 @@ class TestValueSatisfiesType:
     def test_union_list_with_wrong_inner_falls_through(self):
         # [1, 2] doesn't satisfy list[str] (inner mismatch), and not str either
         assert _value_satisfies_type([1, 2], str | list[str]) is False
+
+    # --- Literal (non-runtime origin) ---
+    def test_literal_does_not_crash(self):
+        """Literal origins are not runtime types — should return False, not crash."""
+        assert _value_satisfies_type("active", Literal["active", "inactive"]) is False
 
 
 # ---------------------------------------------------------------------------
@@ -358,6 +363,16 @@ class TestCoerceInputsToSchema:
         # dict stays as dict — convert() doesn't construct Pydantic models
         assert data["inner"] == {"x": 1}
         assert isinstance(data["inner"], dict)
+
+    def test_literal_field_passes_through(self):
+        """Literal-typed fields should not crash coercion."""
+
+        class LiteralSchema(BaseModel):
+            status: Literal["active", "inactive"]
+
+        data: dict[str, Any] = {"status": "active"}
+        coerce_inputs_to_schema(data, LiteralSchema)
+        assert data["status"] == "active"
 
     def test_list_of_pydantic_model_field(self):
         """list[dict] for list[PydanticModel] passes through unchanged."""


autogpt_platform/backend/backend/copilot/tools/helpers.py
```diff
@@ -111,6 +112,15 @@ async def execute_block(
...
                 )
 
+        # Convert non-matching data types to the expected input schema.
+        # This mirrors the executor's validate_exec() logic (executor/utils.py)
+        # and allows string values (e.g. from @@agptfile: expansion) to be
+        # coerced into the block's expected types (e.g. list[list[str]]).
+        for name, data_type in block.input_schema.__annotations__.items():
+            value = input_data.get(name)
+            if (value is not None) and (type(value) is not data_type):
+                input_data[name] = convert(value, data_type)
```
majdyz: This 4-line coercion loop is an exact duplicate of `executor/utils.py:217-220`. Extract it into a shared utility (e.g. `coerce_inputs_to_schema(data, schema)` in `backend.util.type`) so both call sites stay in sync and don't drift independently.

```suggestion
        coerce_inputs_to_schema(input_data, block.input_schema)
```

autogpt_platform/backend/backend/copilot/tools/helpers.py
```diff
@@ -111,6 +112,15 @@ async def execute_block(
                     session_id=session_id,
                 )
 
+        # Convert non-matching data types to the expected input schema.
+        # This mirrors the executor's validate_exec() logic (executor/utils.py)
+        # and allows string values (e.g. from @@agptfile: expansion) to be
+        # coerced into the block's expected types (e.g. list[list[str]]).
+        for name, data_type in block.input_schema.__annotations__.items():
```
majdyz: `__annotations__` only includes fields defined **directly** on the class, not inherited ones. If a block input schema inherits fields from an intermediate parent (e.g. a shared `BaseGoogleInput`), those fields won't be coerced.

Consider using `block.input_schema.model_fields` instead, which includes all fields (inherited + direct) and is the proper Pydantic v2 API. This applies to the executor's copy too — but should be done in both places at once.

Example:
```python
for name, field_info in block.input_schema.model_fields.items():
    value = input_data.get(name)
    if value is not None and type(value) is not field_info.annotation:
        input_data[name] = convert(value, field_info.annotation)
```

majdyz:
### Third review pass ✓

All items from the first two review passes have been addressed:

1. **Shared utility** — `coerce_inputs_to_schema()` + `_value_satisfies_type()` extracted to `backend/util/type.py`
2. **`model_fields` over `__annotations__`** — both the copilot and executor now use `model_fields`, correctly handling inherited schema fields
3. **Executor unified** — `executor/utils.py` now calls the shared `coerce_inputs_to_schema()` instead of the inline loop
4. **Tests fixed** — mock schema correctly exposes `model_fields` with `field.annotation`
5. **`_value_satisfies_type` guard** — avoids redundant `convert()` calls for values that already match the target type (handles generics like `list[str]`, unions, and simple types)

Minor nit (non-blocking): test boilerplate for `mock_workspace_db` could be extracted into a `@pytest.fixture` to reduce repetition across the 6 tests.

LGTM — ready to push.

autogpt_platform/backend/backend/util/type.py
```diff
@@ -249,6 +249,53 @@ def convert(value: Any, target_type: Any) -> Any:
...
+    uncertain so the caller falls through to :func:`convert`.
+    """
+    origin = get_origin(target)
+
+    if origin is Union or origin is types.UnionType:
+        non_none = [a for a in get_args(target) if a is not type(None)]
+        return any(_value_satisfies_type(value, member) for member in non_none)
+
+    # Generic type (e.g. list[str]) → check isinstance against origin (list)
+    if origin is not None:
```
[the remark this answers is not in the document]
majdyz: ~~Initially dismissed as intentional, but on further review this is a valid concern.~~

**Fixed in 7bfe213**: `_value_satisfies_type` now recursively checks inner elements of generic containers. For `list[str]`, it verifies every element is a `str` — if not, it returns `False` so `convert()` handles the recursive coercion.

Added test `test_coerce_inner_elements_of_generic` that verifies `[1, 2, 3]` → `["1", "2", "3"]` for `list[str]`.

majdyz:
### Fourth review pass — pushback on one thing

All original review items are addressed. The new `_value_satisfies_type` with deep recursive checking and the two additional tests (`test_coerce_union_type_preserves_valid_member`, `test_coerce_inner_elements_of_generic`) are well done.

**One issue to fix — heterogeneous tuple handling** in `_value_satisfies_type` (line 285-290):

```python
if (
    _is_type_or_subclass(origin, set)
    or _is_type_or_subclass(origin, frozenset)
    or _is_type_or_subclass(origin, tuple)
) and args:
    return all(_value_satisfies_type(v, args[0]) for v in value)
```

This treats `tuple` like a homogeneous container, always checking against `args[0]`. But `tuple[str, int, bool]` means element 0 is `str`, element 1 is `int`, element 2 is `bool`. The fix:

```python
if _is_type_or_subclass(origin, tuple) and args:
    # tuple[str, ...] → homogeneous (2 args, second is Ellipsis)
    if len(args) == 2 and args[1] is Ellipsis:
        return all(_value_satisfies_type(v, args[0]) for v in value)
    # tuple[str, int, bool] → heterogeneous, check positionally
    return len(value) == len(args) and all(
        _value_satisfies_type(v, t) for v, t in zip(value, args)
    )
if (_is_type_or_subclass(origin, set) or _is_type_or_subclass(origin, frozenset)) and args:
    return all(_value_satisfies_type(v, args[0]) for v in value)
```

This matches how `convert()` already handles tuples (single-arg vs multi-arg branches in `_try_convert`). Low risk to hit in practice since most block schemas use lists, but the inconsistency could cause silent bugs if a block ever uses typed tuples.

Everything else LGTM.

majdyz:
@ntindle can't think of any, what kind of security issue are you imagining ?

PR #12385, opened 2026-03-12
feat(platform): CoPilot credit charging, token rate limiting, and usage UI

majdyz:
### Background

image

CoPilot block execution was not charging credits, LLM token usage was not tracked, and there was no per-user rate limiting. This PR adds all three, plus a frontend usage indicator.

### Changes

**Credit Charging** (`copilot/tools/helpers.py`)
- Pre-execution balance check + post-execution credit deduction via `block_usage_cost` / `spend_credits`
- Uses adapter pattern (RPC fallback) so it works in the CoPilot executor which has no Prisma connection

**Token Rate Limiting** (`copilot/rate_limit.py`)
- Redis-backed daily + weekly fixed-window counters per user
- Fail-open on Redis outages, clock-skew-safe weekly boundaries
- Configurable via `daily_token_limit` / `weekly_token_limit` (0 = unlimited)

**Token Tracking**
- *Baseline* (`copilot/baseline/service.py`): `stream_options={"include_usage": True}` with tiktoken fallback estimation
- *SDK* (`copilot/sdk/service.py`): Extract usage from Claude Agent SDK `ResultMessage`, including cached tokens
- Both: yield `StreamUsage` SSE events, persist `Usage` records, call `record_token_usage` in `finally`

**Usage API** (`api/features/chat/routes.py`)
- `GET /api/chat/usage` — returns `CoPilotUsageStatus` (daily/weekly used, limit, resets_at)
- Pre-turn `check_rate_limit` in `stream_chat_post` (returns 429 on exceed)

**Frontend** (`copilot/components/UsageLimits/`)
- `UsageLimits` popover with daily/weekly progress bars, reset times, dark mode
- `useUsageLimits` hook with 30s auto-refresh via generated Orval API hook

### Tests
| Area | Tests | File |
|------|-------|------|
| Rate limiting | 22 | `rate_limit_test.py` |
| Credit charging | 12 | `helpers_test.py` |
| Usage API | 3 | `routes_test.py` |
| Frontend UI | 9 | `UsageLimits.test.tsx` |

### Checklist

- [x] Changes clearly listed
- [x] Test plan created and executed (46 backend + 9 frontend tests)
- [x] Pre-commit hooks pass (formatting, linting, type checks)
- [x] `.env.default` compatible (new config defaults to 0 = unlimited)
- [x] `docker-compose.yml` compatible (no changes needed)

PR #12400, opened 2026-03-13
fix(backend): split CamelCase block names and filter disabled blocks before batch slicing

majdyz:
## Summary

Two bugs causing blocks to be invisible in CoPilot search:

### Bug 1: CamelCase block names not tokenized
Block names like `AITextGeneratorBlock` were indexed as single tokens in the search database. PostgreSQL's `plainto_tsquery('english', ...)` and the BM25 tokenizer both treat CamelCase as one word, so searching for "text generator" produced zero lexical/BM25 match.

**Fix:** Split CamelCase names into separate words before indexing (e.g. `"AI Text Generator Block"`) and in the BM25 tokenizer.

### Bug 2: Disabled blocks exhausting batch budget (root cause of 36 missing blocks)
The `batch_size` limit in `get_missing_items()` was applied **before** filtering out disabled blocks. With 120+ disabled blocks and `batch_size=100`, the first 100 missing entries were all disabled (skipped via `continue`), leaving the 36 enabled blocks beyond the slice boundary **never indexed**. This made core blocks like `AITextGeneratorBlock`, `AIConversationBlock`, `AIListGeneratorBlock`, etc. completely invisible to search.

**Fix:** Filter disabled blocks from the missing list before slicing by `batch_size`.

### Changes
- **`content_handlers.py`**: 
  - Split CamelCase block names into space-separated words when building `searchableText`
  - Filter disabled blocks before applying `batch_size` slice so enabled blocks aren't starved
- **`hybrid_search.py`**: Updated BM25 `tokenize()` to split CamelCase tokens

### Evidence from local DB
```
Indexed blocks: 341
Total blocks: 497 (156 missing from index)
Missing (non-disabled): 36 — including AITextGeneratorBlock, AIConversationBlock, etc.

# batch_size analysis:
First 100 missing: 0 enabled, 100 disabled  ← batch exhausted by disabled blocks
After 100: 36 enabled                        ← never reached!
```

## Test plan
- [ ] Verify CamelCase splitting: `AITextGeneratorBlock` → `AI Text Generator Block`
- [ ] Run `poetry run pytest backend/api/features/store/` for regressions
- [ ] After deploy, trigger embedding backfill and verify all 36 blocks get indexed
- [ ] Search for "text generator" in CoPilot and verify `AITextGeneratorBlock` appears


autogpt_platform/backend/backend/api/features/store/hybrid_search.py
```diff
@@ -31,10 +31,17 @@
...
+    Also splits CamelCase tokens so that e.g. "AITextGeneratorBlock" yields
+    ["ai", "text", "generator", "block"] in addition to the original token,
+    improving recall for exact-term queries.
+    """
     if not text:
         return []
-    # Lowercase and split on non-alphanumeric characters
+    # Split CamelCase before tokenizing so individual words are matchable
+    text = re.sub(r"([a-z0-9])([A-Z])", r"\1 \2", text)
+    text = re.sub(r"([A-Z]+)([A-Z][a-z])", r"\1 \2", text)
```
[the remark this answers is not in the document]
majdyz: Addressed: extracted `split_camelcase()` as a shared helper with a docstring noting both patterns are linear-time (no nested quantifiers). The regex operates on block names (short strings) and search queries, so ReDoS is not a practical concern.

autogpt_platform/backend/backend/api/features/store/content_handlers.py
```diff
@@ -187,10 +188,13 @@ async def get_missing_items(self, batch_size: int) -> list[ContentItem]:
...
 
         existing_ids = {row["contentId"] for row in existing_result}
+        # Filter disabled blocks before applying batch_size so that a large
+        # number of disabled blocks can't exhaust the batch budget and prevent
+        # enabled blocks from being indexed.
         missing_blocks = [
             (block_id, block_cls)
             for block_id, block_cls in all_blocks.items()
-            if block_id not in existing_ids
+            if block_id not in existing_ids and not block_cls().disabled
```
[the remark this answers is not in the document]
majdyz: Fixed: the disabled-block filter loop now wraps `block_cls()` in try/except, logging a warning and skipping broken blocks instead of crashing the backfill.

autogpt_platform/backend/backend/api/features/store/content_handlers.py
```diff
@@ -13,6 +13,7 @@
 
 from prisma.enums import ContentType
 
+from backend.api.features.store.text_utils import split_camelcase
 from backend.blocks.llm import LlmModel
 from backend.data.db import query_raw_with_schema
```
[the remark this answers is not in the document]
majdyz: Fixed in 36606b206: get_stats() now has the same try/except pattern as get_missing_items(), logging a warning and skipping broken blocks instead of crashing.

autogpt_platform/backend/backend/api/features/store/hybrid_search.py
```diff
@@ -31,11 +31,16 @@
...
+    """Simple tokenizer for BM25 — lowercase and split on word boundaries.
+
+    CamelCase is split first so "AITextGeneratorBlock" becomes
+    ``["ai", "text", "generator", "block"]``.
+    """
     if not text:
         return []
-    # Lowercase and split on non-alphanumeric characters
-    tokens = re.findall(r"\b\w+\b", text.lower())
+    from backend.api.features.store.text_utils import split_camelcase
```
majdyz: 🤖 **[HIGH]** Local import inside `tokenize()` violates 'top-level imports only' rule. No circular import issue exists (it's already top-level in `content_handlers.py`). Move to top of file.

autogpt_platform/backend/backend/api/features/store/content_handlers.py
```diff
@@ -187,25 +188,31 @@ async def get_missing_items(self, batch_size: int) -> list[ContentItem]:
...
-        ]
+        # Filter disabled blocks before applying batch_size so that a large
+        # number of disabled blocks can't exhaust the batch budget and prevent
+        # enabled blocks from being indexed.
+        missing_blocks: list[tuple[str, type]] = []
+        for block_id, block_cls in all_blocks.items():
+            if block_id in existing_ids:
+                continue
+            try:
+                if block_cls().disabled:
```
majdyz: 🤖 **[MEDIUM]** Every non-disabled block is instantiated **twice**: once here (`block_cls().disabled`) and again in the conversion loop (`block_instance = block_cls()`). Block constructors may do non-trivial work. Carry the instance through from the filter step to avoid redundant instantiation.

autogpt_platform/backend/backend/api/features/store/content_handlers.py
```diff
@@ -187,25 +188,31 @@ async def get_missing_items(self, batch_size: int) -> list[ContentItem]:
...
-        missing_blocks = [
-            (block_id, block_cls)
-            for block_id, block_cls in all_blocks.items()
-            if block_id not in existing_ids
-        ]
+        # Filter disabled blocks before applying batch_size so that a large
+        # number of disabled blocks can't exhaust the batch budget and prevent
+        # enabled blocks from being indexed.
+        missing_blocks: list[tuple[str, type]] = []
+        for block_id, block_cls in all_blocks.items():
```
majdyz: 🤖 **[MEDIUM]** List comprehension replaced with loop-and-append, violating the style guide's preference for comprehensions. Extract the try/except into a helper function and use a comprehension.

autogpt_platform/backend/backend/api/features/store/content_handlers.py
```diff
@@ -187,25 +188,31 @@ async def get_missing_items(self, batch_size: int) -> list[ContentItem]:
...
+            if block_id in existing_ids:
+                continue
+            try:
+                if block_cls().disabled:
+                    continue
+            except Exception as e:
+                logger.warning(f"Skipping block {block_id}: failed to init: {e}")
```
majdyz: 🤖 **[LOW]** Bare `except Exception` on block construction silently hides bugs. Catch more specific exceptions or log at error level.

autogpt_platform/backend/backend/api/features/store/content_handlers.py
```diff
@@ -187,25 +188,31 @@ async def get_missing_items(self, batch_size: int) -> list[ContentItem]:
...
+                continue
```
majdyz: 🤖 **[LOW]** Uses f-string in `logger.warning()` while `get_stats` correctly uses `%s` lazy formatting. Use `%s` consistently — it defers string formatting until the message is actually emitted.

autogpt_platform/backend/backend/api/features/store/text_utils.py
```diff
@@ -0,0 +1,18 @@
+"""Small text helpers shared across store search modules."""
+
+import re
+
+
+def split_camelcase(text: str) -> str:
+    """Split CamelCase into separate words.
```
majdyz: 🤖 **[NIT]** No unit tests for `split_camelcase`. Edge cases worth testing: empty string, strings with numbers (`OAuth2Block`), underscores, already-spaced text.

autogpt_platform/backend/backend/api/features/store/hybrid_search.py
majdyz: Fixed in d0f9ec55e — moved split_camelcase import to module level.

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Fixed in d0f9ec55e — extracted _get_enabled_blocks() that returns instances, so no double instantiation.

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Fixed in d0f9ec55e — extracted _get_enabled_blocks() helper function, loop-and-append is now clean and self-contained.

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Keeping broad Exception: block constructors can fail for any reason (missing deps, config errors, type errors). Warning level is appropriate since the block is simply skipped. Narrowing would risk missing failures.

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Fixed in 73e1a5e76 — all logger.warning calls now use lazy %s formatting consistently.

autogpt_platform/backend/backend/api/features/store/text_utils.py
majdyz: Fixed in d0f9ec55e — added parametrized tests in hybrid_search_test.py covering empty string, acronyms (AI, HTTP), numbers, and already-spaced text.

majdyz:
Addressed CodeQL ReDoS warning: added input length bound to `split_camelcase` to prevent potential regex DoS on adversarial input.

autogpt_platform/backend/backend/api/features/store/hybrid_search.py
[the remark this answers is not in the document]
majdyz: Addressed — input is now bounded to 500 chars via `text = text[:500]` in `split_camelcase()`. The CodeQL alert shows as fixed.

autogpt_platform/backend/backend/api/features/store/content_handlers.py
```diff
@@ -154,6 +155,26 @@ async def get_stats(self) -> dict[str, int]:
...
 
 
+def _get_enabled_blocks() -> dict[str, Any]:
+    """Return ``{block_id: block_instance}`` for all enabled, instantiable blocks.
+
+    Disabled blocks and blocks that fail to instantiate are silently skipped
+    (with a warning log), so callers never need their own try/except loop.
+    """
+    from backend.blocks import get_blocks
+
```
majdyz, line 173: 🤖 **🔴 Critical**: `_get_enabled_blocks()` has a local/inner import (`from backend.blocks import get_blocks`). The codebase style rules require top-level imports only. Move this import to the top of the file with the other imports.

If this is intentionally deferred to avoid circular imports, add a comment explaining why — but ideally restructure to avoid it.

autogpt_platform/backend/backend/api/features/store/content_handlers.py
```diff
@@ -154,6 +155,26 @@ async def get_stats(self) -> dict[str, int]:
         }
```
majdyz, line 166: 🤖 **🟡 Warning**: The return type `dict[str, Any]` is too loose. The values are block instances — consider using the actual block base type (e.g. `dict[str, Block]`) to preserve type information for callers. Using `Any` means callers get no autocomplete or type checking on block attributes (`name`, `disabled`, `input_schema`, etc.).

autogpt_platform/backend/backend/api/features/store/content_handlers.py
```diff
@@ -154,6 +155,26 @@ async def get_stats(self) -> dict[str, int]:
...
+    enabled: dict[str, Any] = {}
+    for block_id, block_cls in get_blocks().items():
+        try:
+            instance = block_cls()
+        except Exception as e:
+            logger.warning("Skipping block %s: init failed: %s", block_id, e)
+            continue
+        if not instance.disabled:
+            enabled[block_id] = instance
+    return enabled
```
majdyz, line 186: 🤖 **🟡 Warning**: `_get_enabled_blocks()` instantiates every block class on each call. Both `get_missing_items()` and `get_stats()` call it independently, so if both are invoked (e.g. health check + backfill), all blocks get instantiated twice. Consider caching the result (e.g. `@functools.lru_cache` or a module-level variable) since block classes are static during a process lifetime.

autogpt_platform/backend/backend/api/features/store/content_handlers.py
```diff
@@ -187,52 +203,41 @@ async def get_missing_items(self, batch_size: int) -> list[ContentItem]:
...
+        # Convert to ContentItem — disabled filtering already done by
+        # _get_enabled_blocks so batch_size won't be exhausted by disabled blocks.
         items = []
-        for block_id, block_cls in missing_blocks[:batch_size]:
-            try:
-                block_instance = block_cls()
-
-                if block_instance.disabled:
-                    continue
+        for block_id, block in enabled.items():
```
majdyz: 🤖 **🟢 Nit**: The manual loop with `if block_id in existing_ids: continue` + `if len(items) >= batch_size: break` could be simplified with a generator + `itertools.islice`:

```python
missing = ((bid, b) for bid, b in enabled.items() if bid not in existing_ids)
for block_id, block in itertools.islice(missing, batch_size):
    ...
```

This makes the filtering + limiting intent clearer and removes the manual `break`.

autogpt_platform/backend/backend/api/features/store/content_handlers.py
```diff
@@ -243,7 +248,7 @@ async def get_missing_items(self, batch_size: int) -> list[ContentItem]:
                 # Check if block has LlmModel field in input schema
```
majdyz, line 263: 🤖 **🟡 Warning**: `metadata["name"]` stores the original unsplit block name (`block.name` = `"AITextGeneratorBlock"`), while `searchable_text` contains the split version (`"AI Text Generator Block"`). This asymmetry is probably intentional (metadata for display, searchable_text for indexing), but worth confirming — if any downstream code searches against the metadata `name` field, it would still miss CamelCase blocks.

autogpt_platform/backend/backend/api/features/store/text_utils.py
```diff
@@ -0,0 +1,19 @@
...
+
+    Examples::
+
+        >>> split_camelcase("AITextGeneratorBlock")
+        'AI Text Generator Block'
+        >>> split_camelcase("HTTPRequestBlock")
+        'HTTP Request Block'
+    """
+    text = text[:500]  # Bound input length to prevent regex DoS
```
majdyz: 🤖 **🟢 Nit**: The `[:500]` truncation uses a magic number. Consider extracting to a named constant (e.g. `_MAX_CAMELCASE_INPUT_LEN = 500`) for self-documentation.

autogpt_platform/backend/backend/api/features/store/content_handlers.py
```diff
@@ -154,6 +155,26 @@ async def get_stats(self) -> dict[str, int]:
         }
 
 
+def _get_enabled_blocks() -> dict[str, Any]:
+    """Return ``{block_id: block_instance}`` for all enabled, instantiable blocks.
+
+    Disabled blocks and blocks that fail to instantiate are silently skipped
+    (with a warning log), so callers never need their own try/except loop.
+    """
+    from backend.blocks import get_blocks
```
majdyz: 🤖 **🟡 Warning**: `_get_enabled_blocks()` is a synchronous function that instantiates potentially hundreds of block classes. It is called from async methods `get_missing_items()` and `get_stats()`. Instantiating all blocks synchronously will block the event loop. Consider wrapping in `asyncio.to_thread()` at the call sites.

autogpt_platform/backend/backend/api/features/store/hybrid_search.py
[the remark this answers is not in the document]
majdyz: The CamelCase-splitting regex referenced by this alert was already removed in a prior revision. The current `tokenize()` function uses only `re.findall(r"\b\w+\b", text.lower())`, which has no backtracking risk. This alert is stale.

autogpt_platform/backend/backend/api/features/store/content_handlers_test.py
```diff
@@ -15,15 +13,80 @@
...
+# ---------------------------------------------------------------------------
+
+
+def _make_block_class(
+    *,
+    name: str = "Block",
+    description: str = "",
+    disabled: bool = False,
+    categories: list | None = None,
+    fields: dict | None = None,
```
majdyz: 🤖 **🟡 Warning**: Bare `list` and `dict` type hints lack element types — this is duck typing by omission. Use explicit generic types for clarity.

```python
# Before
categories: list | None = None,
fields: dict | None = None,

# After
categories: list[MagicMock] | None = None,
fields: dict[str, str] | None = None,
```

autogpt_platform/backend/backend/api/features/store/content_handlers_test.py
```diff
@@ -15,15 +13,80 @@
...
+    inst = MagicMock()
+    inst.name = name
+    inst.disabled = disabled
+    inst.description = description
+    inst.categories = categories or []
+    field_mocks = {}
+    for fname, fdesc in (fields or {}).items():
+        f = MagicMock()
+        f.description = fdesc
+        field_mocks[fname] = f
```
majdyz: 🤖 **🟢 Nit**: Loop-and-append pattern building `field_mocks` — prefer a dict comprehension per code style rules.

```python
# Before
field_mocks = {}
for fname, fdesc in (fields or {}).items():
    f = MagicMock()
    f.description = fdesc
    field_mocks[fname] = f

# After
field_mocks = {
    fname: MagicMock(description=fdesc)
    for fname, fdesc in (fields or {}).items()
}
```

`MagicMock(description=fdesc)` sets the attribute directly via the constructor.

autogpt_platform/backend/backend/api/features/store/hybrid_search_test.py
```diff
@@ -14,8 +14,49 @@
...
+        ("HTTPRequestBlock", "HTTP Request Block"),
+        ("simpleWord", "simple Word"),
+        ("already spaced", "already spaced"),
+        ("XMLParser", "XML Parser"),
+        ("getHTTPResponse", "get HTTP Response"),
+        ("Block", "Block"),
+        ("", ""),
+    ],
+)
+def test_split_camelcase(input_text: str, expected: str):
```
majdyz: 🤖 **🟡 Warning**: The parametrized test cases are good but miss important CamelCase edge cases that exercise boundary behavior of the two regex patterns:

- `"IOError"` → `"IO Error"` (two-letter acronym before lowercase)
- `"OAuth2Block"` → **bug**: produces `"O Auth2 Block"` because `([A-Z]+)([A-Z][a-z])` splits `OA` into `O` + `Au`. Expected: `"OAuth2 Block"`.
- `"getHTTPSResponse"` → `"get HTTPS Response"` (acronym ending before another word)
- `"A"` → `"A"` (single char, no split)
- `"ABC"` → `"ABC"` (all-caps, no split expected)

The `OAuth2` case reveals that the regex doesn't handle mixed-case brand names correctly. Worth documenting as a known limitation or fixing the regex.

autogpt_platform/backend/backend/api/features/store/content_handlers_test.py
```diff
@@ -15,15 +13,80 @@
...
+    *,
+    name: str = "Block",
+    description: str = "",
+    disabled: bool = False,
+    categories: list | None = None,
+    fields: dict | None = None,
+    raise_on_init: Exception | None = None,
+) -> MagicMock:
+    cls = MagicMock()
+    if raise_on_init:
```
majdyz: 🤖 **🟡 Warning**: `if raise_on_init:` is falsy for `raise_on_init=ValueError()` (exception with no args has a truthy instance, but `ValueError("")` or custom exceptions that evaluate falsy would be missed). Use explicit `is not None` check.

```python
# Before
if raise_on_init:

# After
if raise_on_init is not None:
```

autogpt_platform/backend/backend/api/features/store/hybrid_search_test.py
majdyz: Bumping this — the regex bug is confirmed and reproducible:

```python
>>> split_camelcase("OAuth2Block")
'O Auth2 Block'   # ❌ Expected: "OAuth2 Block"
```

**Root cause**: After step 1 (`([a-z0-9])([A-Z])`), `OAuth2Block` correctly becomes `OAuth2 Block`. But step 2 (`([A-Z]+)([A-Z][a-z])`) then matches `OAu` in `OAuth2` — `O` matches `[A-Z]+`, `Au` matches `[A-Z][a-z]` — inserting a space to produce `O Auth2 Block`.

This affects any CamelCase word starting with a short acronym followed by lowercase: `OAuth`, `OData`, etc. The test suite currently passes only because it doesn't include these cases.

**Suggested fix** — add a word-boundary or lookbehind constraint to step 2 so it only fires on runs of 2+ uppercase letters:

```python
text = re.sub(r"([A-Z]{2,})([A-Z][a-z])", r"\1 \2", text)
```

Wait, that still has the same issue. The proper fix is:

```python
text = re.sub(r"([A-Z]+)([A-Z][a-z])", r"\1 \2", text)
```
→ replace with:
```python
text = re.sub(r"(?<=[A-Z])(?=[A-Z][a-z])", " ", text)
```

This lookahead/lookbehind approach inserts a space between an uppercase letter and an uppercase-then-lowercase pair without consuming any characters, which avoids the `O|Auth` mis-split since `O` alone before `Auth` won't have a preceding uppercase to anchor the lookbehind.

Actually, let me verify:
```
"OAuth2 Block" with (?<=[A-Z])(?=[A-Z][a-z]) →
  at position between O and A: [A-Z] behind = O ✓, [A-Z][a-z] ahead = Au ✓ → still splits "O Auth2 Block"
```

So that doesn't fix it either. The real issue is that `OAuth` is *not* an acronym — it's a proper word that just happens to start with `OA`. The two-pass regex heuristic fundamentally can't distinguish "OAuth" from "HTTPResponse".

The practical fix: switch to the `(?<=[a-z0-9])(?=[A-Z])|(?<=[A-Z])(?=[A-Z][a-z])` **zero-width** pattern and accept that `OAuth` will be mis-split — OR add `OAuth2Block` as a known-failing test case and document the limitation. But at minimum, the test suite should include this case so the behavior is explicit.

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Bumping — this local import is still present in the latest revision (a662dc16):

```python
def _get_enabled_blocks() -> dict[str, Any]:
    from backend.blocks import get_blocks  # ← local import
```

The `split_camelcase` import was correctly moved to module level, but this one was introduced in the same commit. `get_blocks` is already imported at the top level in other modules without circular import issues. If there IS a circular import here, please add a comment explaining it — otherwise, move to the top of the file.

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Bumping — `_get_enabled_blocks()` is still synchronous and called from two `async` methods. Looking at the function body: it calls `get_blocks()` (which loads all block modules), then instantiates every block class. In the current registry there are 100+ blocks; each constructor may validate schemas, set up field metadata, etc.

This will block the event loop for the entire duration. Wrapping with `await asyncio.to_thread(_get_enabled_blocks)` at the two call sites is a one-line fix.

Lower priority than the regex bug and local import, but worth addressing before merge since this code runs in the API server's event loop.

majdyz:
This pr accidentally pick up chnages from https://github.com/Significant-Gravitas/AutoGPT/pull/12385 e.g: working from a branched from that work instead of fresh dev, we need to clean this up

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Fixed in 7d869d828 — `get_blocks` import moved to module level with other imports.

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Fixed in 7d869d828 — return type changed to `dict[str, Block[Any, Any]]` using a `TYPE_CHECKING` import from `backend.blocks._base`.

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Fixed in 7d869d828 — added `@functools.lru_cache(maxsize=1)` so blocks are only instantiated once per process.

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Fixed in 7d869d828 — replaced manual loop+break with `itertools.islice(missing, batch_size)` over a generator expression.

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Fixed in 7d869d828 — metadata now stores `display_name` (the split version) instead of the raw CamelCase name. Both searchable_text and metadata.name are now consistent.

autogpt_platform/backend/backend/api/features/store/text_utils.py
majdyz: Already fixed — uses `_MAX_CAMELCASE_INPUT_LEN = 500` named constant in `text_utils.py`.

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Fixed in 7d869d828 — both `get_missing_items()` and `get_stats()` now call `_get_enabled_blocks()` via `asyncio.to_thread()` to avoid blocking the event loop.

autogpt_platform/backend/backend/api/features/store/content_handlers_test.py
majdyz: Fixed in 7d869d828 — `categories: list[MagicMock] | None` and `fields: dict[str, str] | None`.

autogpt_platform/backend/backend/api/features/store/content_handlers_test.py
majdyz: Fixed in 7d869d828 — replaced loop-and-append with `{fname: MagicMock(description=fdesc) for ...}` comprehension.

autogpt_platform/backend/backend/api/features/store/hybrid_search_test.py
majdyz: Fixed in 7d869d828 — added `OAuth2Block`, `IOError`, and `getHTTPSResponse` parametrized test cases. The regex bug was already fixed by using `{2,}` quantifier instead of `+` in the second pattern.

autogpt_platform/backend/backend/api/features/store/content_handlers_test.py
majdyz: Fixed in 7d869d828 — changed to `if raise_on_init is not None:`.

autogpt_platform/backend/backend/api/features/store/hybrid_search_test.py
majdyz: Confirmed fixed — the `{2,}` quantifier in the second regex pattern prevents splitting single-char sequences. `split_camelcase('OAuth2Block')` now correctly returns `'OAuth2 Block'`. Added test case in 7d869d828.

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Fixed in 7d869d828 — `get_blocks` import moved to top of file alongside other imports.

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Fixed in 7d869d828 — both call sites now use `await asyncio.to_thread(_get_enabled_blocks)`. Combined with `@lru_cache`, the first call runs in a thread pool and subsequent calls return instantly from cache.

autogpt_platform/backend/backend/api/features/store/hybrid_search.py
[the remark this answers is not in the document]
majdyz: Addressed: the CamelCase regex was extracted to split_camelcase() in text_utils.py with linear-time patterns only. Input is bounded to 500 chars via _MAX_CAMELCASE_INPUT_LEN. The original regex in tokenize() was replaced with re.findall which has no backtracking risk.

majdyz:
Cleaned up: reverted all extraneous changes from #12385 (copilot, chat routes, frontend files) back to origin/dev. The PR now only contains the 5 store search files (content_handlers, hybrid_search, text_utils, and their tests).

autogpt_platform/backend/backend/api/features/store/hybrid_search.py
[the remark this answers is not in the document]
majdyz: Fixed — `split_camelcase()` has been rewritten to use an O(n) character-by-character algorithm instead of the regex pattern that triggered this ReDoS warning.

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Already addressed: get_blocks import is now at module level (line 22). The local import was removed in a prior revision.

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Already addressed: return type is now dict[str, Block[Any, Any]] (line 168), with Block imported under TYPE_CHECKING.

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Already addressed: _get_enabled_blocks() is decorated with @functools.lru_cache(maxsize=1) (line 167), so repeated calls reuse the cached result.

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Already addressed: the loop now uses itertools.islice(missing, batch_size) with a generator expression for filtering (lines 220-222).

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Confirmed intentional: metadata name now stores the split display name (display_name or block.name on line 268), so both searchable_text and metadata.name use the CamelCase-split version. No asymmetry remains.

autogpt_platform/backend/backend/api/features/store/text_utils.py
majdyz: Already addressed: the magic number was extracted to _MAX_CAMELCASE_INPUT_LEN = 500 constant (line 3 of text_utils.py).

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Already addressed: both call sites now use await asyncio.to_thread(_get_enabled_blocks) (lines 198 and 285) to avoid blocking the event loop.

autogpt_platform/backend/backend/api/features/store/content_handlers_test.py
majdyz: Already addressed: the helper now uses list[MagicMock] | None and dict[str, str] | None (lines 38-39 of content_handlers_test.py).

autogpt_platform/backend/backend/api/features/store/text_utils.py
```diff
@@ -0,0 +1,65 @@
...
+        >>> split_camelcase("OAuth2Block")
+        'OAuth2 Block'
+    """
+    if len(text) > _MAX_CAMELCASE_INPUT_LEN:
+        logger.debug(
+            "split_camelcase: truncating input from %d to %d chars",
+            len(text),
+            _MAX_CAMELCASE_INPUT_LEN,
+        )
+        text = text[:_MAX_CAMELCASE_INPUT_LEN]
```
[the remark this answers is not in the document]
majdyz: This was a valid concern — the 500-char truncation was originally added as a ReDoS guard when `split_camelcase` used regex. It has since been addressed in commit 28ff7b60 which replaced the regex with an O(n) character-by-character loop. That commit also removed the `_MAX_CAMELCASE_INPUT_LEN` truncation entirely, so `split_camelcase()` now processes the full input without any length limit. No further changes needed.

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Fixed in f2f48e98c. The `_get_enabled_blocks()` docstring already documented process-lifetime caching as intentional and explains why no TTL is needed (unlike `get_blocks()`). Additionally added `_MAX_CAMELCASE_INPUT_LEN = 500` as a named constant in `text_utils.py` and the caching tests now explicitly verify single-call behaviour.

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Fixed in f2f48e98c. The fallback chain is now `display_name or block.name or block_id`, ensuring `metadata["name"]` is always a non-empty string. Updated the existing `test_block_handler_handles_none_name` test to assert the new block_id fallback behaviour.

autogpt_platform/backend/backend/api/features/store/text_utils.py
majdyz: Fixed in f2f48e98c. Extracted `_MAX_CAMELCASE_INPUT_LEN = 500` as a named module-level constant with comment `# Safety cap; block names are typically < 50 chars`. The constant is now applied in `split_camelcase()` to truncate oversized inputs.

autogpt_platform/backend/backend/api/features/store/content_handlers_test.py
majdyz: Fixed in f2f48e98c. Added three new test cases to `content_handlers_test.py`: (1) `test_get_enabled_blocks_cached` — verifies `get_blocks()` is called exactly once across two `_get_enabled_blocks()` calls; (2) `test_get_enabled_blocks_returns_immutable_mapping` — asserts the result is a `MappingProxyType` and that mutation raises `TypeError`; (3) `test_block_handler_get_missing_items_batch_size_zero` — verifies `batch_size=0` returns an empty list. The `block.name = None` path was already covered by `test_block_handler_handles_none_name` (updated its assertion to match the new `block_id` fallback).

majdyz:
Addressed all Should Fix items from autogpt-reviewer reviews (f2f48e98c):

1. **`lru_cache` no TTL** (`content_handlers.py:168`) — The docstring already documented process-lifetime caching as intentional (no TTL needed since blocks register at import time). No code change needed here; addressed the related caching test gap instead.

2. **`metadata["name"]` can be `None`** (`content_handlers.py:285`) — Fixed: fallback chain is now `display_name or block.name or block_id`, ensuring the field is always a non-empty string.

3. **Missing tests** — Added `test_get_enabled_blocks_cached` (caching: `get_blocks()` called once), `test_get_enabled_blocks_returns_immutable_mapping` (MappingProxyType raises TypeError on write), and `test_block_handler_get_missing_items_batch_size_zero` (batch_size=0 → empty list). The `block.name = None` path was already covered; updated its assertion to match the new block_id fallback.

4. **`split_camelcase` tests location** — Tests are already in `text_utils_test.py`. `hybrid_search_test.py` tests `tokenize()`, which is the correct module.

5. **`_MAX_CAMELCASE_INPUT_LEN = 500` constant** (`text_utils.py:3`) — Added named constant with comment `# Safety cap; block names are typically < 50 chars` and apply it in `split_camelcase()`.

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Fixed in 7d869d82: moved `from backend.blocks import get_blocks` to the top-level imports section of `content_handlers.py`.

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Fixed in 7d869d82: tightened the return type from `dict[str, Any]` to `dict[str, Block[Any, Any]]` (now `types.MappingProxyType[str, Block[Any, Any]]` after e706c5af further hardened it with an immutable wrapper).

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Fixed in 7d869d82 + e706c5af: added `@functools.lru_cache(maxsize=1)` to cache the result for the process lifetime (blocks are static import-time singletons), and wrapped the call in `asyncio.to_thread(_get_enabled_blocks)` in both async callers to keep the first (heavy) call off the event loop. Subsequent calls are O(1) dict lookups.

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Addressed in 7d869d82: wrapped both `get_missing_items` and `get_stats` calls with `await asyncio.to_thread(_get_enabled_blocks)` so block instantiation (on first call) runs in a thread pool and never blocks the event loop.

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Fixed in 7d869d82: replaced the manual loop+break with a generator expression filtered through `itertools.islice(missing, batch_size)`, which lazily yields up to `batch_size` items without allocating the full list first.

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Fixed in 7d869d82: `metadata["name"]` now stores `display_name` (the human-readable split version, e.g. `"AI Text Generator Block"`) which matches what `searchable_text` contains. If both `display_name` and `block.name` are falsy, it falls back to `block_id` (added in f2f48e98) to ensure the field is always a non-empty string.

autogpt_platform/backend/backend/api/features/store/text_utils.py
majdyz: Done in f2f48e98: extracted `_MAX_CAMELCASE_INPUT_LEN = 500` as a named module-level constant in `text_utils.py` (with an explanatory comment) and applied it as the safety cap in `split_camelcase()`.

autogpt_platform/backend/backend/api/features/store/content_handlers_test.py
majdyz: Fixed in 7d869d82: `_make_block_class` in `content_handlers_test.py` now uses `list[MagicMock] | None` and `dict[str, str] | None` for the `categories` and `fields` parameters respectively, replacing the bare `list` and `dict` annotations.

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Fixed — the import () has been moved to the module's top-level imports (line 23). No local import remains inside `_get_enabled_blocks()`.

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Fixed — the return type is now `types.MappingProxyType[str, Block[Any, Any]]` with a `TYPE_CHECKING` import of the `Block` base class. Callers get full type information.

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Fixed — `_get_enabled_blocks()` is now decorated with `@functools.lru_cache(maxsize=1)`. The first call instantiates all blocks; subsequent calls (from `get_stats()`, `get_missing_items()`, etc.) are a no-op dict lookup. The cached mapping is wrapped in `MappingProxyType` to prevent mutation.

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Fixed — the loop now uses a generator + `itertools.islice` as suggested. The manual `if len(items) >= batch_size: break` was removed in favour of `itertools.islice(missing, batch_size)`.

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Confirmed intentional. `metadata["name"]` stores the human-readable display form (e.g. `"AI Text Generator Block"` from `split_camelcase(block.name)`), not the raw class name. The fallback chain is `display_name or block.name or block_id` so it's always non-empty. Downstream code uses `metadata["name"]` only for display; all search happens against `searchable_text`.

autogpt_platform/backend/backend/api/features/store/text_utils.py
majdyz: Fixed — extracted to `_MAX_CAMELCASE_INPUT_LEN = 500` at the top of `text_utils.py` with the comment `# Safety cap; block names are typically < 50 chars`.

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Fixed — both call sites now use `await asyncio.to_thread(_get_enabled_blocks)` so the (potentially heavy) first-call instantiation runs in a thread-pool worker, not on the event loop. Subsequent calls hit the lru_cache and return instantly, so the thread-pool overhead is negligible.

autogpt_platform/backend/backend/api/features/store/content_handlers_test.py
majdyz: Fixed — updated to `categories: list[MagicMock] | None = None` and `fields: dict[str, str] | None = None` with explicit element types.

majdyz:
Fixed in 349b9fc00:

1. **`import types` inside test** (`content_handlers_test.py`) — moved to top-level imports.
2. **`# type: ignore[index]`** (`content_handlers_test.py`) — replaced with `result_any: Any = result` cast; `from typing import Any` also lifted to top-level.
3. **Docstring mismatch** (`test_block_handler_get_missing_items_batch_size_zero`) — updated docstring from "without querying the database" to accurately reflect that the DB IS queried (as proven by `mock_query.assert_called_once()`); empty result is due to `itertools.islice` limiting to 0 items.
4. **Local import in `embeddings.py`** (line 665, `elif ContentType.BLOCK` branch) — `from backend.blocks import get_blocks` lifted to module-level top-of-file import, consistent with the same fix already applied to `content_handlers.py`.

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Already done — the local import was moved to the top of the file (line 23: `from backend.blocks import get_blocks`) in commit dbad60933. No inner import remains in `_get_enabled_blocks()`.

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Already tightened — return type is now `types.MappingProxyType[str, Block[Any, Any]]` (commit e706c5afc), giving callers full type information on block attributes.

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Already done — `_get_enabled_blocks()` is decorated with `@functools.lru_cache(maxsize=1)` (commit e706c5afc). The cached result is wrapped in `MappingProxyType` so callers can't mutate it. The docstring explains the caching rationale and why no TTL is needed.

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Already implemented — the current code uses exactly this pattern (commit dbad60933): `missing = ((bid, b) for bid, b in enabled.items() if bid not in existing_ids)` followed by `for block_id, block in itertools.islice(missing, batch_size):`. The manual `break` is gone.

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Confirmed intentional — `metadata["name"]` now stores the split display name (e.g. `"AI Text Generator Block"`), not the raw CamelCase name. Line 285: `"name": display_name or block.name or block_id`. The split name is used for both display and metadata to keep them consistent.

autogpt_platform/backend/backend/api/features/store/text_utils.py
majdyz: Already done — `_MAX_CAMELCASE_INPUT_LEN = 500` is defined at the top of `text_utils.py` (commit f2f48e98c) with a comment explaining it's a safety cap.

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Already addressed — both `get_missing_items()` and `get_stats()` wrap the call in `await asyncio.to_thread(_get_enabled_blocks)` (commit dbad60933). The docstring on `_get_enabled_blocks()` also notes that subsequent calls are cache hits, making the thread overhead negligible.

autogpt_platform/backend/backend/api/features/store/content_handlers_test.py
majdyz: Already fixed (commit 349b9fc) — both parameters now have explicit generic types: `categories: list[MagicMock] | None = None` and `fields: dict[str, str] | None = None`.

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Already fixed: `from backend.blocks import get_blocks` was moved to module level (line 23) in d0f9ec55e.

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Already fixed in dbad60933: return type is now `types.MappingProxyType[str, Block[Any, Any]]` (line 169).

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Already fixed in dbad60933: `_get_enabled_blocks()` is decorated with `@functools.lru_cache(maxsize=1)` (line 168). The docstring documents the intentional process-lifetime caching.

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Already fixed in dbad60933: the loop now uses `itertools.islice(missing, batch_size)` with a generator expression (lines 237-239).

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Intentional: `metadata["name"]` now stores the split display name (line 285: `display_name or block.name or block_id`). Since `display_name = split_camelcase(block.name)`, the metadata name field matches the searchable text. Downstream code that reads metadata name will get the human-readable split form.

autogpt_platform/backend/backend/api/features/store/text_utils.py
majdyz: Already fixed in f2f48e98c: the magic number is now the named constant `_MAX_CAMELCASE_INPUT_LEN = 500` in text_utils.py (line 4).

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Already fixed in dbad60933: both call sites (`get_missing_items` line 215, `get_stats` line 302) now use `await asyncio.to_thread(_get_enabled_blocks)` to keep the event loop unblocked.

autogpt_platform/backend/backend/api/features/store/content_handlers_test.py
majdyz: Already fixed in 349b9fc00: type hints now use explicit generics — `list[MagicMock] | None` and `dict[str, str] | None` (lines 40-41).

autogpt_platform/backend/backend/api/features/store/hybrid_search.py
```diff
@@ -31,12 +32,19 @@
...
+    docs), not only block names.  For normal prose the function is effectively
+    a no-op (no CamelCase boundaries to split), while for code identifiers
+    that may appear in documentation it improves token coverage.
+    """
     if not text:
         return []
-    # Lowercase and split on non-alphanumeric characters
-    tokens = re.findall(r"\b\w+\b", text.lower())
-    return tokens
+    return re.findall(r"\b\w+\b", split_camelcase(text).lower())
```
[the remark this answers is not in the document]
majdyz: Fixed in c87e95464 — removed `split_camelcase` from the BM25 `tokenize()` function entirely. The 500-char truncation was silently dropping content from long documentation text.

Block names are already split at indexing time in `content_handlers.py`, so the tokenizer doesn't need `split_camelcase`.

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Fixed in 349b9fc00: `from backend.blocks import get_blocks` moved to top-level imports (line 23).

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Fixed in f2f48e98c: return type is now `types.MappingProxyType[str, Block[Any, Any]]` (line 169).

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Fixed in f2f48e98c: `_get_enabled_blocks()` is now decorated with `@functools.lru_cache(maxsize=1)` and returns `MappingProxyType` to prevent mutation of the cached data (line 168-200).

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Fixed in f2f48e98c: now uses `itertools.islice(missing, batch_size)` with a generator expression (lines 237-239).

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Fixed in f2f48e98c: `metadata["name"]` now stores the split display name via `display_name or block.name or block_id` (line 285), so metadata and searchable_text are consistent.

autogpt_platform/backend/backend/api/features/store/text_utils.py
majdyz: Fixed in c87e95464: the regex-based approach with truncation was replaced entirely by an O(n) character-by-character algorithm in `split_camelcase()`, eliminating both the regex and the magic number.

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Fixed in f2f48e98c: both call sites now use `await asyncio.to_thread(_get_enabled_blocks)` (lines 215 and 302) to keep the synchronous block instantiation off the event loop.

autogpt_platform/backend/backend/api/features/store/content_handlers_test.py
majdyz: Fixed in 349b9fc00: type hints updated to `list[MagicMock] | None` and `dict[str, str] | None` (lines 40-41).

autogpt_platform/backend/backend/api/features/store/content_handlers.py
```diff
@@ -154,6 +165,41 @@ async def get_stats(self) -> dict[str, int]:
         }
 
 
+@functools.lru_cache(maxsize=1)
+def _get_enabled_blocks() -> types.MappingProxyType[str, Block[Any, Any]]:
```
Pwuts: we have `AnyBlockSchema` for this:
```suggestion
def _get_enabled_blocks() -> types.MappingProxyType[str, AnyBlockSchema]:
```
which is more useful than `Block` or `Block[Any, Any]` because of how the type of `input_schema` and `output_schema` is inferred

autogpt_platform/backend/backend/api/features/store/content_handlers.py
```diff
@@ -154,6 +165,41 @@ async def get_stats(self) -> dict[str, int]:
...
+    enabled: dict[str, Block[Any, Any]] = {}
+    for block_id, block_cls in get_blocks().items():
+        try:
+            instance = block_cls()
+        except Exception as e:
+            logger.warning("Skipping block %s: init failed: %s", block_id, e)
+            continue
+        if not instance.disabled:
+            enabled[block_id] = instance
+    return types.MappingProxyType(enabled)
```
Pwuts: what dis?

autogpt_platform/backend/backend/api/features/store/text_utils.py
Pwuts, line 1: Same comment as in other PR: this belongs in `backend/util/text.py`

autogpt_platform/backend/backend/api/features/store/text_utils.py
```diff
@@ -0,0 +1,52 @@
+"""Small text helpers shared across store search modules."""
+
+
+def split_camelcase(text: str) -> str:
```
Pwuts: The frontend does something like this but also has an exception list of camel case terms that should stay together, like `YouTube`:
https://github.com/Significant-Gravitas/AutoGPT/blob/048fb06b0a998790a779b7ad7a2f7c836af3cd3a/autogpt_platform/frontend/src/lib/utils.ts#L157-L177

Do we need this here too?

autogpt_platform/backend/backend/api/features/store/hybrid_search.py
```diff
@@ -797,15 +799,20 @@ async def _log_vector_error_diagnostics(error: Exception) -> None:
...
+            "  Error: %s\n"
+            "  search_path: %s\n"
+            "  current_schema: %s\n"
+            "  user_info: %s\n"
+            "  pgvector_extension: %s",
+            error,
+            diagnostics.get("search_path"),
+            diagnostics.get("current_schema"),
+            diagnostics.get("user_info"),
+            diagnostics.get("pgvector_extension"),
```
Pwuts: this is a DX (readability) regression imo; same everywhere else (including content_handlers.py)

autogpt_platform/backend/backend/api/features/store/content_handlers.py
```diff
@@ -187,52 +231,38 @@ async def get_missing_items(self, batch_size: int) -> list[ContentItem]:
...
-        for block_id, block_cls in missing_blocks[:batch_size]:
+        for block_id, block in itertools.islice(missing, batch_size):
             try:
-                block_instance = block_cls()
-
-                if block_instance.disabled:
-                    continue
-
                 # Build searchable text from block metadata
+                display_name = split_camelcase(block.name) if block.name else ""
```
Pwuts, line 232: is `block.name` ever not set? sounds like an error condition

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Fixed in a08673cd — changed to `AnyBlockSchema` as suggested.

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Removed — the `MappingProxyType` wrapper was unnecessary complexity. `_get_enabled_blocks()` now returns a plain `dict`.

autogpt_platform/backend/backend/api/features/store/text_utils.py
majdyz: Fixed in a08673cd — moved `split_camelcase` to `backend/util/text.py`. The old `text_utils.py` is now a backward-compat re-export shim.

autogpt_platform/backend/backend/api/features/store/text_utils.py
majdyz: Good call — added an exception map mirroring the frontend list: YouTube, OpenAI, AutoGPT, GitHub, LinkedIn. After splitting, known compound terms are restored via regex replacement. Added test cases for all of them.

autogpt_platform/backend/backend/api/features/store/hybrid_search.py
majdyz: Fixed — reverted all logger calls back to f-strings in both `hybrid_search.py` and `content_handlers.py`.

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Good point — it shouldn't happen in practice. Added a `logger.warning()` when `block.name` is falsy so we'll know if it does. The block_id fallback is still there so indexing doesn't break, but we'll see a warning in logs if any block is missing a name.

autogpt_platform/backend/backend/api/features/store/hybrid_search.py
Pwuts: ?? I see the exact opposite in this latest commit https://github.com/Significant-Gravitas/AutoGPT/pull/12400/commits/40f5532b1e0fc1b5da6c096b73202a064be736d8

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: you confused the poor AI, haha
the MappingProxyType was used to avoid cache mutation so the cache value is not modified during the runtime of the application.

autogpt_platform/backend/backend/api/features/store/hybrid_search.py
majdyz: the AI ended up reverting this, though your recommendation is contradicting with most of general concencus on code styling [[logging-f-string (G004) | Ruff](https://docs.astral.sh/ruff/rules/logging-f-string/)](https://docs.astral.sh/ruff/rules/logging-f-string/)

f-string is bad for logging, you lack lazy evaluation and lose extra param processing evaluation especially on exception.

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Already fixed: the import `from backend.blocks import get_blocks` is at module level (line 21). The local import was removed in a prior revision.

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Already fixed: the return type is `dict[str, AnyBlockSchema]` (using `TYPE_CHECKING` import for the type annotation).

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Already fixed: `_get_enabled_blocks()` is decorated with `@functools.lru_cache(maxsize=1)`, so it only runs once per process lifetime.

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Already fixed: the code uses `itertools.islice(missing, batch_size)` with a generator expression for filtering.

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Intentional: `metadata["name"]` uses the split form (`display_name or block.name or block_id`) where `display_name = split_camelcase(block.name)`. So it stores the human-readable split name (e.g. "AI Text Generator Block"), matching `searchable_text`. No downstream code searches against the raw metadata name field — it's used only for display.

autogpt_platform/backend/backend/api/features/store/text_utils.py
majdyz: Already fixed: the `[:500]` truncation was removed in a prior revision. The current `split_camelcase()` uses a character-by-character O(n) algorithm with no truncation needed.

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Already fixed: both call sites use `await asyncio.to_thread(_get_enabled_blocks)` to keep the event loop unblocked.

autogpt_platform/backend/backend/api/features/store/content_handlers_test.py
majdyz: Already fixed: the type hints are now explicit — `categories: list[MagicMock] | None`, `fields: dict[str, str] | None`.

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Already fixed: `get_blocks` import is at module level (line 21). The inner import was removed in an earlier commit.

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Already fixed: return type is `dict[str, AnyBlockSchema]` (using TYPE_CHECKING import for the block base type).

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Already fixed: `_get_enabled_blocks()` is decorated with `@functools.lru_cache(maxsize=1)`. Both `get_missing_items()` and `get_stats()` call it via `asyncio.to_thread()`, and on subsequent calls the cached dict is returned instantly.

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Already fixed: the loop now uses `itertools.islice(missing, batch_size)` with a generator expression for filtering.

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Confirmed intentional: `metadata["name"]` now stores the split display name (`display_name or block.name or block_id`), so both metadata and searchable_text use the CamelCase-split form. No downstream code searches against the raw metadata name field.

autogpt_platform/backend/backend/api/features/store/text_utils.py
majdyz: The `[:500]` truncation was removed when the implementation was rewritten as a character-by-character O(n) algorithm (no regex, no backtracking risk). The magic number is no longer present.

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Already fixed: both call sites use `await asyncio.to_thread(_get_enabled_blocks)` to keep block instantiation off the event loop. On cached calls the overhead is negligible.

autogpt_platform/backend/backend/api/features/store/content_handlers_test.py
majdyz: Already fixed: `_make_block_class` now uses `list[MagicMock] | None` and `dict[str, str] | None` with explicit element types.

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Already fixed: `get_blocks` import is at module level (line 21). The inner import was removed in a prior commit.

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Fixed: return type is now `dict[str, AnyBlockSchema]` (using TYPE_CHECKING import to avoid circular deps).

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Fixed: `_get_enabled_blocks()` is decorated with `@functools.lru_cache(maxsize=1)` and callers use `asyncio.to_thread()` to keep the first (heavy) call off the event loop.

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Fixed: the loop now uses `itertools.islice(missing, batch_size)` with a generator expression for filtering.

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Intentional: `metadata["name"]` now stores the split display name (e.g. "AI Text Generator Block") via the fallback chain `display_name or block.name or block_id`. Both display and search use the split form, so no asymmetry.

autogpt_platform/backend/backend/api/features/store/text_utils.py
majdyz: Fixed: the `[:500]` truncation was removed when `split_camelcase` was rewritten to use a character-by-character O(n) algorithm. No magic number remains.

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Fixed: both `get_missing_items()` and `get_stats()` now call `await asyncio.to_thread(_get_enabled_blocks)` to keep block instantiation off the event loop.

autogpt_platform/backend/backend/api/features/store/content_handlers_test.py
majdyz: Fixed: type hints now use `list[MagicMock] | None` and `dict[str, str] | None` with explicit element types.

ntindle (approved):

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Fixed: the import was moved to the top of the file (line 21) in a prior commit. `_get_enabled_blocks()` no longer contains any local imports.

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Fixed: the return type is now `dict[str, AnyBlockSchema]` (line 168), preserving full type information for callers.

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Fixed: `_get_enabled_blocks()` is now decorated with `@functools.lru_cache(maxsize=1)` (line 167). Blocks are registered at import time and never change, so caching for the process lifetime is safe. The test file includes a fixture that clears the cache between tests.

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Fixed: the loop now uses a generator + `itertools.islice` (line 223-225), making the filtering + limiting intent clearer.

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Good catch. This asymmetry was intentional in the original code but is now resolved: `metadata["name"]` stores the split display name (line 275: `display_name or block.name or block_id`), where `display_name = split_camelcase(block.name)`. Both the searchable text and metadata name now use the human-readable split form.

autogpt_platform/backend/backend/api/features/store/text_utils.py
majdyz: Addressed: the `[:500]` truncation was removed when `split_camelcase()` was rewritten as a single-pass O(n) character algorithm (no regex, no backtracking). There is no longer any need for an input length guard, so the magic number is gone entirely.

autogpt_platform/backend/backend/api/features/store/content_handlers.py
majdyz: Fixed: both call sites now use `await asyncio.to_thread(_get_enabled_blocks)` (lines 201, 292). On the first call, the heavy instantiation runs in a thread pool worker. On subsequent calls, the `lru_cache` makes it a trivial dict lookup, so the thread-pool overhead is negligible.

autogpt_platform/backend/backend/api/features/store/content_handlers_test.py
majdyz: Fixed: the type hints are now explicit — `categories: list[MagicMock] | None = None` and `fields: dict[str, str] | None = None` (lines 38-39).

PR #12403, opened 2026-03-13
fix(backend/copilot): transcript compaction, token-based rate limiting, block credit charging, usage UI

majdyz:
This PR bundles three related changes around CoPilot reliability and cost controls.

### Fix 1 — Transcript compaction (prevent "Prompt is too long")
TranscriptBuilder accumulated all messages including pre-compaction content, causing uploaded transcripts to grow unbounded. When the CLI compacted mid-stream, TranscriptBuilder kept the full uncompacted history. On the next turn, `--resume` would fail with "Prompt is too long".

- Added `replace_entries()` to `TranscriptBuilder` that clears and reloads entries from compacted JSONL
- Added `read_cli_session_file()` and `_cli_project_dir()` helpers in `transcript.py` to read the CLI's own session file (which reflects mid-stream compaction)
- After `CompactionTracker.emit_end_if_ready()` fires, reads the CLI session file and replaces TranscriptBuilder entries with the compacted version
- At transcript download time, if content exceeds `COMPACT_THRESHOLD_BYTES` (400KB), compacts it before passing to `--resume` and uploads the compacted version for future turns
- Added `compact_transcript()`, `_transcript_to_messages()`, and `_messages_to_transcript()` conversion helpers

### Fix 2 — Token-based rate limiting
Per-user daily/weekly token usage limits enforced via Redis fixed-window counters.

- New `rate_limit.py` module with `check_rate_limit()`, `record_token_usage()`, `get_usage_status()`
- Cost-weighted counting (cache read tokens at 10%, cache creation at 25%)
- Fails open when Redis is unavailable
- Pre-turn soft check in `stream_chat_post` route (429 on exceeded)
- Token usage captured from streaming chunks (baseline/OpenRouter) and `ResultMessage` (SDK/Anthropic)
- Fallback to tiktoken estimation when provider doesn't report usage
- Configurable via `ChatConfig.daily_token_limit` / `weekly_token_limit`
- New `/usage` API endpoint for frontend consumption

### Fix 3 — Block credit charging in CoPilot
Blocks executed by CoPilot now charge user credits, matching the graph executor behavior.

- Pre-execution balance check + post-execution `spend_credits` call
- `_get_credits()` / `_spend_credits()` helpers with Prisma/RPC adapter pattern
- `block_usage_cost()` integration for cost calculation
- Error handling for `InsufficientBalanceError`

### Frontend — Usage limits UI
- New `UsageLimits` popover component showing daily/weekly token usage bars
- `useUsageLimits` hook fetching from `/usage` endpoint
- Auto-refresh on stream completion
- Token usage displayed in session detail view
- Link to billing/credits page

### Other changes
- `StreamUsage` now emits as SSE comment (avoids Vercel AI SDK parser rejection)
- `Usage` model extended with `cache_read_tokens` / `cache_creation_tokens`
- `db_manager.py` exposes `spend_credits` / `get_credits` RPC methods
- Logging format fixes (f-strings → % formatting)

#### For code changes:
- [x] I have made a test plan
  - [x] All SDK tests pass (`poetry run pytest backend/copilot/sdk/`)
  - [x] Pre-commit hooks pass (lint, format, typecheck)
  - [ ] Manual test: create a copilot session with large tool outputs, verify compaction kicks in and subsequent turns work
  - [ ] Manual test: load an existing oversized transcript, verify it gets compacted at download time
  - [ ] Manual test: verify rate limiting returns 429 when limits exceeded
  - [ ] Manual test: verify block execution charges credits

### DIFF
--- a/autogpt_platform/backend/backend/copilot/rate_limit.py
+++ b/autogpt_platform/backend/backend/copilot/rate_limit.py
@@ -77,7 +77,11 @@ def _daily_reset_time(now: datetime | None = None) -> datetime:
 
 
 def _weekly_reset_time(now: datetime | None = None) -> datetime:
-    """Calculate when the current weekly window resets (next Monday 00:00 UTC)."""
+    """Calculate when the current weekly window resets (next Monday 00:00 UTC).
+
+    On Monday itself, ``(7 - weekday) % 7`` is 0; the ``or 7`` fallback
+    pushes to *next* Monday so the current week's window stays open.
+    """
     if now is None:
         now = datetime.now(UTC)
     days_until_monday = (7 - now.weekday()) % 7 or 7
--- a/autogpt_platform/backend/backend/copilot/sdk/service.py
+++ b/autogpt_platform/backend/backend/copilot/sdk/service.py
@@ -79,6 +79,7 @@ from .tool_adapter import (
 )
 from .transcript import (
     COMPACT_THRESHOLD_BYTES,
+    TranscriptDownload,
     cleanup_cli_project_dir,
     compact_transcript,
     download_transcript,
@@ -630,6 +631,56 @@ async def _prepare_file_attachments(
     return PreparedAttachments(hint=hint, image_blocks=image_blocks)
 
 
+async def _maybe_compact_and_upload(
+    dl: TranscriptDownload,
+    user_id: str,
+    session_id: str,
+    log_prefix: str = "[Transcript]",
+) -> str:
+    """Compact an oversized transcript and upload the compacted version.
+
+    Returns the (possibly compacted) transcript content, or an empty string
+    if compaction was needed but failed.
+    """
+    content = dl.content
+    if len(content) <= COMPACT_THRESHOLD_BYTES:
+        return content
+
+    logger.warning(
+        "%s Transcript oversized (%dB > %dB), compacting",
+        log_prefix,
+        len(content),
+        COMPACT_THRESHOLD_BYTES,
+    )
+    compacted = await compact_transcript(content, log_prefix=log_prefix)
+    if not compacted:
+        logger.warning(
+            "%s Compaction failed, skipping resume for this turn", log_prefix
+        )
+        return ""
+
+    # Keep the original message_count: it reflects the number of
+    # session.messages covered by this transcript, which the gap-fill
+    # logic uses as a slice index.  Counting JSONL lines would give a
+    # smaller number (compacted messages != session message count) and
+    # cause already-covered messages to be re-injected.
+    try:
+        await upload_transcript(
+            user_id=user_id,
+            session_id=session_id,
+            content=compacted,
+            message_count=dl.message_count,
+            log_prefix=log_prefix,
+        )
+    except Exception:
+        logger.warning(
+            "%s Failed to upload compacted transcript",
+            log_prefix,
+            exc_info=True,
+        )
+    return compacted
+
+
 async def stream_chat_completion_sdk(
     session_id: str,
     message: str | None = None,
@@ -841,50 +892,12 @@ async def stream_chat_completion_sdk(
                 is_valid,
             )
             if is_valid:
-                transcript_content = dl.content
-                # Compact oversized transcripts to prevent "Prompt is too long"
-                if len(transcript_content) > COMPACT_THRESHOLD_BYTES:
-                    logger.warning(
-                        "%s Transcript oversized (%dB > %dB), compacting",
-                        log_prefix,
-                        len(transcript_content),
-                        COMPACT_THRESHOLD_BYTES,
-                    )
-                    compacted = await compact_transcript(
-                        transcript_content, log_prefix=log_prefix
-                    )
-                    if compacted:
-                        transcript_content = compacted
-                        # Keep the original message_count: it reflects the
-                        # number of session.messages covered by this transcript,
-                        # which the gap-fill logic uses as a slice index.
-                        # Counting JSONL lines would give a smaller number
-                        # (compacted messages != session message count) and
-                        # cause already-covered messages to be re-injected.
-                        # Best-effort upload of compacted version
-                        try:
-                            await upload_transcript(
-                                user_id=user_id or "",
-                                session_id=session_id,
-                                content=transcript_content,
-                                message_count=dl.message_count,
-                                log_prefix=log_prefix,
-                            )
-                        except Exception:
-                            logger.warning(
-                                "%s Failed to upload compacted transcript",
-                                log_prefix,
-                                exc_info=True,
-                            )
-                    else:
-                        # Compaction failed — skip resume to avoid
-                        # "Prompt is too long" on an oversized transcript.
-                        logger.warning(
-                            "%s Compaction failed, skipping resume for this turn",
-                            log_prefix,
-                        )
-                        transcript_content = ""
-
+                transcript_content = await _maybe_compact_and_upload(
+                    dl,
+                    user_id=user_id or "",
+                    session_id=session_id,
+                    log_prefix=log_prefix,
+                )
                 # Load previous context into builder (empty string is a no-op)
                 if transcript_content:
                     transcript_builder.load_previous(
--- a/autogpt_platform/backend/backend/copilot/sdk/transcript.py
+++ b/autogpt_platform/backend/backend/copilot/sdk/transcript.py
@@ -23,7 +23,7 @@ import openai
 
 from backend.copilot.config import ChatConfig
 from backend.util import json
-from backend.util.prompt import compress_context
+from backend.util.prompt import CompressResult, compress_context
 
 logger = logging.getLogger(__name__)
 
@@ -40,6 +40,11 @@ _SAFE_ID_RE = re.compile(r"[^0-9a-fA-F-]")
 STRIPPABLE_TYPES = frozenset(
     {"progress", "file-history-snapshot", "queue-operation", "summary", "pr-link"}
 )
+
+# JSONL protocol values used in transcript serialization.
+STOP_REASON_END_TURN = "end_turn"
+COMPACT_MSG_ID_PREFIX = "msg_compact_"
+ENTRY_TYPE_MESSAGE = "message"
 
 
 @dataclass
@@ -560,10 +565,10 @@ def _messages_to_transcript(messages: list[dict]) -> str:
             message: dict = {
                 "role": "assistant",
                 "model": "",
-                "id": f"msg_compact_{uuid4().hex[:24]}",
-                "type": "message",
+                "id": f"{COMPACT_MSG_ID_PREFIX}{uuid4().hex[:24]}",
+                "type": ENTRY_TYPE_MESSAGE,
                 "content": [{"type": "text", "text": content}] if content else [],
-                "stop_reason": "end_turn",
+                "stop_reason": STOP_REASON_END_TURN,
                 "stop_sequence": None,
             }
         else:
@@ -577,6 +582,23 @@ def _messages_to_transcript(messages: list[dict]) -> str:
         lines.append(json.dumps(entry, separators=(",", ":")))
         last_uuid = uid
     return "\n".join(lines) + "\n" if lines else ""
+
+
+async def _run_compression(
+    messages: list[dict],
+    model: str,
+    cfg: ChatConfig,
+    log_prefix: str,
+) -> CompressResult:
+    """Run LLM-based compression with truncation fallback."""
+    try:
+        async with openai.AsyncOpenAI(
+            api_key=cfg.api_key, base_url=cfg.base_url, timeout=30.0
+        ) as client:
+            return await compress_context(messages=messages, model=model, client=client)
+    except Exception as e:
+        logger.warning("%s LLM compaction failed, using truncation: %s", log_prefix, e)
+        return await compress_context(messages=messages, model=model, client=None)
 
 
 async def compact_transcript(
@@ -596,20 +618,7 @@ async def compact_transcript(
         logger.warning("%s Too few messages to compact (%d)", log_prefix, len(messages))
         return None
     try:
-        try:
-            async with openai.AsyncOpenAI(
-                api_key=cfg.api_key, base_url=cfg.base_url, timeout=30.0
-            ) as client:
-                result = await compress_context(
-                    messages=messages, model=cfg.model, client=client
-                )
-        except Exception as e:
-            logger.warning(
-                "%s LLM compaction failed, using truncation: %s", log_prefix, e
-            )
-            result = await compress_context(
-                messages=messages, model=cfg.model, client=None
-            )
+        result = await _run_compression(messages, cfg.model, cfg, log_prefix)
         if not result.was_compacted:
             logger.info("%s Transcript already within token budget", log_prefix)
             return content
--- a/autogpt_platform/frontend/src/app/(platform)/copilot/components/UsageLimits/UsageLimits.tsx
+++ b/autogpt_platform/frontend/src/app/(platform)/copilot/components/UsageLimits/UsageLimits.tsx
@@ -8,6 +8,9 @@ import { Button } from "@/components/ui/button";
 import { ChartBar } from "@phosphor-icons/react";
 import { useUsageLimits } from "./useUsageLimits";
 
+const MS_PER_MINUTE = 60_000;
+const MS_PER_HOUR = 3_600_000;
+
 function formatResetTime(resetsAt: Date | string): string {
   const resetDate =
     typeof resetsAt === "string" ? new Date(resetsAt) : resetsAt;
@@ -15,11 +18,11 @@ function formatResetTime(resetsAt: Date | string): string {
   const diffMs = resetDate.getTime() - now.getTime();
   if (diffMs <= 0) return "now";
 
-  const hours = Math.floor(diffMs / (1000 * 60 * 60));
+  const hours = Math.floor(diffMs / MS_PER_HOUR);
 
   // Under 24h: show relative time ("in 4h 23m")
   if (hours < 24) {
-    const minutes = Math.floor((diffMs % (1000 * 60 * 60)) / (1000 * 60));
+    const minutes = Math.floor((diffMs % MS_PER_HOUR) / MS_PER_MINUTE);
     if (hours > 0) return `in ${hours}h ${minutes}m`;
     return `in ${minutes}m`;
   }


autogpt_platform/backend/backend/copilot/sdk/service.py
```diff
@@ -827,16 +840,41 @@ async def _fetch_transcript():
...
+                    if compacted:
+                        transcript_content = compacted
+                        # Upload the compacted version for future turns
+                        await upload_transcript(
+                            user_id=user_id or "",
+                            session_id=session_id,
+                            content=transcript_content,
+                            message_count=dl.message_count,
+                            log_prefix=log_prefix,
+                        )
```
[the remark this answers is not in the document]
majdyz: Good catch — updated dl.message_count to session length after compaction so the gap-fill logic uses the correct watermark.

autogpt_platform/backend/backend/copilot/sdk/transcript.py
```diff
@@ -417,3 +451,147 @@ async def delete_transcript(user_id: str, session_id: str) -> None:
...
+                "role": "assistant",
+                "model": "",
+                "id": f"msg_compact_{uuid4().hex[:24]}",
+                "type": "message",
+                "content": [{"type": "text", "text": content}] if content else [],
+                "stop_reason": "end_turn",
+                "stop_sequence": None,
+            }
+        else:
+            message = {"role": role, "content": content}
```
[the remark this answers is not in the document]
majdyz: Fixed — _transcript_to_messages now properly flattens tool_result content blocks (including nested list structures) into plain text for summarization. The tool_use_id pairing isn't preserved through LLM compaction regardless (compacted transcript gets fresh UUIDs), but the content is now correctly extracted instead of being stringified as a dict repr.

majdyz:
Re: [sentry comment on message_count](https://github.com/Significant-Gravitas/AutoGPT/pull/12403#discussion_r2930557919) — Good catch. The JSONL line count after compaction doesn't correspond to `session.messages` indices. Compacted messages are fewer than the original `ChatMessage` objects. Fixed by preserving the original `dl.message_count` watermark after compaction, since the compacted transcript still covers the same range of session messages.

autogpt_platform/backend/backend/copilot/sdk/service.py
```diff
@@ -827,16 +840,68 @@ async def _fetch_transcript():
...
+                    )
+                    if compacted:
+                        transcript_content = compacted
+                        # Recount after compaction so gap-fill uses
+                        # the correct watermark on subsequent turns.
+                        dl.message_count = sum(
+                            1
+                            for line in transcript_content.splitlines()
+                            if line.strip()
+                        )
```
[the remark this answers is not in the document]
majdyz: Good catch — fixed by preserving the original dl.message_count watermark instead of recounting JSONL lines.

autogpt_platform/backend/backend/copilot/sdk/service.py
```diff
@@ -827,16 +840,68 @@ async def _fetch_transcript():
...
+                    transcript_builder.load_previous(
+                        transcript_content, log_prefix=log_prefix
+                    )
+                resume_file = (
+                    write_transcript_to_tempfile(
+                        transcript_content, session_id, sdk_cwd
+                    )
+                    if transcript_content
+                    else None
                 )
```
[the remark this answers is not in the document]
majdyz: Already addressed in a prior commit — the code now preserves the original `dl.message_count` watermark and skips re-upload when compaction fails.

autogpt_platform/backend/backend/copilot/sdk/transcript.py
```diff
@@ -417,3 +451,160 @@ async def delete_transcript(user_id: str, session_id: str) -> None:
...
+                    # Flatten tool_result content for summarisation;
+                    # tool_use_id pairing is not preserved through LLM
+                    # compaction — the compacted transcript uses fresh IDs.
+                    inner = block.get("content", "")
+                    if isinstance(inner, list):
+                        for sub in inner:
+                            if isinstance(sub, dict):
+                                str_parts.append(str(sub.get("text", "")))
+                            else:
+                                str_parts.append(str(sub))
```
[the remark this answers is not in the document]
majdyz: Fixed — dict blocks without a `text` key now fall back to `json.dumps(sub)` instead of an empty string, preserving structured tool outputs for compaction.

autogpt_platform/backend/backend/copilot/sdk/transcript.py
```diff
@@ -417,3 +451,158 @@ async def delete_transcript(user_id: str, session_id: str) -> None:
...
+    log_prefix: str = "[Transcript]",
+) -> str | None:
+    """Compact an oversized JSONL transcript using LLM summarization.
+
+    Converts transcript entries to plain messages, runs ``compress_context``
+    (the same compressor used for pre-query history), and rebuilds JSONL.
+
+    Returns the compacted JSONL string, or ``None`` on failure.
+    """
+    from backend.copilot.config import ChatConfig
```
majdyz: 🤖 🟡 **Warning**: Local imports violate the project's top-level imports rule. `from backend.copilot.config import ChatConfig`, `import openai`, and `from backend.util.prompt import compress_context` should be moved to the module's top-level imports.

autogpt_platform/backend/backend/copilot/tools/helpers.py
```diff
@@ -21,6 +24,36 @@
 logger = logging.getLogger(__name__)
 
 
+async def _get_credits(user_id: str) -> int:
+    """Get user credits using the adapter pattern (RPC when Prisma unavailable)."""
+    if db.is_connected():
+        from backend.data.credit import get_user_credit_model
```
majdyz: 🤖 🟡 **Warning**: Local imports inside `_get_credits` and `_spend_credits`. The `from backend.data.credit import get_user_credit_model` and `from backend.util.clients import get_database_manager_async_client` imports are conditionally loaded inside function bodies. Per project style rules, these should be top-level imports. If there's a circular import issue forcing this pattern, document it with a comment.

autogpt_platform/backend/backend/copilot/sdk/transcript.py
```diff
@@ -137,32 +139,64 @@ def _sanitize_id(raw_id: str, max_len: int = 36) -> str:
...
+    After the CLI compacts context, its session file contains the compacted
+    conversation.  Reading this file lets ``TranscriptBuilder`` replace its
+    uncompacted entries with the CLI's compacted version.
+    """
+    project_dir = _cli_project_dir(sdk_cwd)
+    if not project_dir or not os.path.isdir(project_dir):
+        return None
+    jsonl_files = list(Path(project_dir).glob("*.jsonl"))
+    if not jsonl_files:
+        logger.debug("[Transcript] No CLI session file in %s", project_dir)
```
majdyz, line 179: 🤖 🟡 **Warning**: `import shutil` is a local/inner import inside the function body. Move it to the top-level imports of the module.

autogpt_platform/backend/backend/copilot/baseline/service.py
```diff
@@ -411,6 +427,58 @@ async def stream_chat_completion_baseline(
             except Exception:
                 logger.warning("[Baseline] Langfuse trace context teardown failed")
 
+        # Fallback: estimate tokens via tiktoken when the provider does
+        # not honour stream_options={"include_usage": True}.
+        # Count the full message list (system + history + turn) since
+        # each API call sends the complete context window.
+        if turn_prompt_tokens == 0 and turn_completion_tokens == 0:
+            from backend.util.prompt import (
```
majdyz: 🤖 🟡 **Warning**: Local imports of `estimate_token_count` and `estimate_token_count_str` inside the `finally` block. These should be top-level imports per project style rules.

autogpt_platform/backend/backend/copilot/tools/helpers.py
```diff
@@ -123,6 +170,37 @@ async def execute_block(
...
+                        graph_id=synthetic_graph_id,
+                        node_id=synthetic_node_id,
+                        node_exec_id=node_exec_id,
+                        block_id=block_id,
+                        block=block.name,
+                        input=cost_filter,
+                        reason="copilot_block_execution",
+                    ),
+                )
+            except InsufficientBalanceError:
```
majdyz, line 181: 🤖 🔴 **Critical**: TOCTOU race condition — block has already executed and produced side effects (e.g., API calls, data mutations) by the time `InsufficientBalanceError` is caught here. The user gets charged nothing but the block's external effects already happened. Consider: (1) reserving/holding credits before execution and releasing on failure, or (2) at minimum, logging the execution outputs so the side effects are traceable. Returning `ErrorResponse` here is misleading because the block *did* execute successfully.

autogpt_platform/backend/backend/copilot/sdk/transcript.py
```diff
@@ -417,3 +451,158 @@ async def delete_transcript(user_id: str, session_id: str) -> None:
...
+
+# ---------------------------------------------------------------------------
+# Transcript compaction
+# ---------------------------------------------------------------------------
+
+# Transcripts above this byte threshold are compacted at download time.
+COMPACT_THRESHOLD_BYTES = 400_000
+
+
+def _transcript_to_messages(content: str) -> list[dict]:
```
majdyz, line 528: 🤖 🟡 **Warning**: `_transcript_to_messages` is a 50+ line function with deeply nested if/elif/isinstance chains. Consider extracting the content-block flattening logic into smaller helper functions (e.g., `_flatten_assistant_content(blocks)` and `_flatten_tool_result_content(blocks)`) to improve readability and testability.

autogpt_platform/backend/backend/copilot/rate_limit.py
```diff
@@ -0,0 +1,249 @@
...
+        daily=UsageWindow(
+            used=daily_used,
+            limit=daily_token_limit,
+            resets_at=_daily_reset_time(now=now),
+        ),
+        weekly=UsageWindow(
+            used=weekly_used,
+            limit=weekly_token_limit,
+            resets_at=_weekly_reset_time(now=now),
+        ),
```
majdyz, line 140: 🤖 🟢 **Nit**: The `_weekly_reset_time` function has a subtle edge case: on Monday itself, `(7 - now.weekday()) % 7` is 0, and the `or 7` fallback correctly pushes to *next* Monday. However, this means a user's weekly usage is never checked against a window that starts on the current Monday — the reset always points 7 days ahead. This is fine for the current use case but worth a brief comment explaining the Monday behavior.

autogpt_platform/backend/backend/copilot/baseline/service.py
```diff
@@ -242,6 +249,15 @@ async def stream_chat_completion_baseline(
             tool_calls_by_index: dict[int, dict[str, str]] = {}
```
majdyz, line 253: 🤖 🟢 **Nit**: The `chunk.choices[0]` access will raise `IndexError` if `chunk.choices` is an empty list (not None). The guard on line above only checks `if chunk.choices` which passes for an empty list in some SDKs. Consider `if not chunk.choices` or `if not chunk.choices` before indexing.

autogpt_platform/backend/backend/copilot/baseline/service.py
```diff
@@ -411,6 +427,58 @@ async def stream_chat_completion_baseline(
...
+                        completion_tokens=turn_completion_tokens,
+                    )
+                except Exception as usage_err:
+                    logger.warning(
+                        "[Baseline] Failed to record token usage: %s", usage_err
+                    )
+
         # Persist assistant response
         if assistant_text:
             session.messages.append(
```
majdyz, line 485: 🤖 🟡 **Warning**: `record_token_usage` in the baseline path is called inside a bare `try/except Exception` that silently swallows all failures. If Redis is persistently down, token usage is silently lost with no alerting mechanism — the `logger.warning` will be noisy but there's no metric or circuit breaker. Consider adding a counter/metric for monitoring dropped usage records, especially since this data feeds the rate limiter.

This is less urgent than the other issues but worth noting for operational visibility.

autogpt_platform/backend/backend/copilot/sdk/transcript.py
```diff
@@ -137,32 +139,64 @@ def _sanitize_id(raw_id: str, max_len: int = 36) -> str:
...
+    if not project_dir or not os.path.isdir(project_dir):
+        return None
+    jsonl_files = list(Path(project_dir).glob("*.jsonl"))
+    if not jsonl_files:
+        logger.debug("[Transcript] No CLI session file in %s", project_dir)
+        return None
+    # Pick the most recently modified file (there should only be one per turn).
+    # Resolve + prefix check to prevent symlink escapes.
+    session_file = max(jsonl_files, key=lambda p: p.stat().st_mtime)
+    real_path = str(session_file.resolve())
```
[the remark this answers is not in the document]
majdyz: Fixed - now builds a safe list of (mtime, path) tuples with try/except around stat(), skipping any files that vanish between glob and stat.

autogpt_platform/backend/backend/copilot/sdk/transcript.py
```diff
@@ -417,3 +451,158 @@ async def delete_transcript(user_id: str, session_id: str) -> None:
...
+                elif isinstance(block, str):
+                    str_parts.append(block)
+            msg_dict["content"] = "\n".join(str_parts) if str_parts else ""
+        else:
+            msg_dict["content"] = raw_content or ""
+        messages.append(msg_dict)
+    return messages
+
+
+def _messages_to_transcript(messages: list[dict]) -> str:
```
majdyz, line 555: 🤖 🟡 **Warning**: `str(sub.get("text", json.dumps(sub)))` has a subtle bug when `sub` has a `text` key with value `None` (e.g., `{"text": None, "type": "image"}`). `dict.get("text", fallback)` returns `None` because the key exists, so `str(None)` produces the literal string `"None"` instead of the JSON serialization.

Fix: use `sub.get("text") or json.dumps(sub)` to fall through on falsy values.

autogpt_platform/backend/backend/copilot/sdk/service.py
```diff
@@ -827,16 +840,67 @@ async def _fetch_transcript():
...
+                    logger.warning(
+                        "%s Transcript oversized (%dB > %dB), compacting",
+                        log_prefix,
+                        len(transcript_content),
+                        COMPACT_THRESHOLD_BYTES,
+                    )
+                    compacted = await compact_transcript(
+                        transcript_content, log_prefix=log_prefix
+                    )
+                    if compacted:
```
majdyz: 🤖 🟡 **Warning**: The transcript compaction block (lines 842-905) reaches 5 levels of nesting: `try > if dl > if is_valid > if oversized > if compacted > try`. Extract into a helper like `_maybe_compact_and_upload(dl, user_id, session_id, log_prefix) -> str` with early returns to flatten:

```python
async def _maybe_compact_and_upload(dl, user_id, session_id, log_prefix):
    content = dl.content
    if len(content) <= COMPACT_THRESHOLD_BYTES:
        return content
    compacted = await compact_transcript(content, log_prefix=log_prefix)
    if not compacted:
        return ""
    try:
        await upload_transcript(...)
    except Exception:
        logger.warning(...)
    return compacted
```

autogpt_platform/backend/backend/copilot/sdk/service.py
```diff
@@ -1389,6 +1509,53 @@ async def _next_msg() -> Any:
...
 
+        # --- Persist token usage to session + rate-limit counters ---
+        # Both must live in finally so they stay consistent even when an
+        # exception interrupts the try block after StreamUsage was yielded.
+        if turn_prompt_tokens > 0 or turn_completion_tokens > 0:
+            total_tokens = (
+                turn_prompt_tokens
+                + turn_cache_read_tokens
+                + turn_cache_creation_tokens
+                + turn_completion_tokens
```
majdyz: 🤖 🟡 **Warning**: `total_tokens` is computed identically in two places — once at line 1434 for `StreamUsage`, and again here for session persistence. DRY violation: if the formula changes, you must update both. Compute it once before yielding `StreamUsage` and reuse the value in the `finally` block.

autogpt_platform/backend/backend/copilot/rate_limit.py
```diff
@@ -0,0 +1,249 @@
...
+    """
+    now = datetime.now(UTC)
+    try:
+        redis = await get_redis_async()
+        daily_raw, weekly_raw = await asyncio.gather(
+            redis.get(_daily_key(user_id, now=now)),
+            redis.get(_weekly_key(user_id, now=now)),
+        )
+        daily_used = int(daily_raw or 0)
+        weekly_used = int(weekly_raw or 0)
```
majdyz: 🤖 🟡 **Warning**: `get_usage_status` (lines 107-114) and `check_rate_limit` (lines 150-157) duplicate the same Redis fetch pattern: get redis, gather daily+weekly keys, parse ints with `int(x or 0)`. Extract a shared helper:

```python
async def _fetch_counters(user_id: str, now: datetime) -> tuple[int, int]:
    redis = await get_redis_async()
    daily_raw, weekly_raw = await asyncio.gather(
        redis.get(_daily_key(user_id, now=now)),
        redis.get(_weekly_key(user_id, now=now)),
    )
    return int(daily_raw or 0), int(weekly_raw or 0)
```

Then each caller wraps it in its own try/except with different fallback behavior.

autogpt_platform/backend/backend/copilot/tools/helpers.py
```diff
@@ -21,6 +24,36 @@
...
 
 
+async def _get_credits(user_id: str) -> int:
+    """Get user credits using the adapter pattern (RPC when Prisma unavailable)."""
+    if db.is_connected():
+        from backend.data.credit import get_user_credit_model
+
+        credit_model = await get_user_credit_model(user_id)
+        return await credit_model.get_credits(user_id)
+    else:
```
majdyz: 🤖 🟡 **Warning**: `_get_credits` and `_spend_credits` (lines 27-54) use `if/else` with `db.is_connected()` — the `else` branch can be simplified with an early return:

```python
async def _get_credits(user_id: str) -> int:
    if not db.is_connected():
        from backend.util.clients import get_database_manager_async_client
        return await get_database_manager_async_client().get_credits(user_id)
    from backend.data.credit import get_user_credit_model
    credit_model = await get_user_credit_model(user_id)
    return await credit_model.get_credits(user_id)
```

This eliminates one nesting level. (Same applies to `_spend_credits`.)

autogpt_platform/backend/backend/copilot/sdk/transcript.py
```diff
@@ -417,3 +464,158 @@ async def delete_transcript(user_id: str, session_id: str) -> None:
...
+    messages = _transcript_to_messages(content)
+    if len(messages) < 2:
+        logger.warning("%s Too few messages to compact (%d)", log_prefix, len(messages))
+        return None
+    try:
+        import openai
+
+        from backend.util.prompt import compress_context
+
+        try:
```
majdyz: 🤖 🟡 **Warning**: `compact_transcript` has nested try blocks (outer try at 582, inner try at 588, fallback at 598). The inner try/except/fallback pattern should be extracted to a helper:

```python
async def _run_compression(messages, model, cfg):
    try:
        async with openai.AsyncOpenAI(...) as client:
            return await compress_context(messages=messages, model=model, client=client)
    except Exception as e:
        logger.warning("LLM compaction failed, using truncation: %s", e)
        return await compress_context(messages=messages, model=model, client=None)
```

This reduces `compact_transcript` to a clean sequence: parse -> compress -> validate -> return.

autogpt_platform/backend/backend/copilot/sdk/transcript.py
```diff
@@ -417,3 +464,158 @@ async def delete_transcript(user_id: str, session_id: str) -> None:
...
+    for msg in messages:
+        role = msg.get("role", "user")
+        entry_type = "assistant" if role == "assistant" else "user"
+        uid = str(uuid4())
+        content = msg.get("content", "")
+        if role == "assistant":
+            message: dict = {
+                "role": "assistant",
+                "model": "",
+                "id": f"msg_compact_{uuid4().hex[:24]}",
```
majdyz: 🤖 🟢 **Nit**: Magic strings `"end_turn"`, `"msg_compact_"`, `"message"` in `_messages_to_transcript` should be named constants. These are JSONL protocol values that appear in multiple places across the transcript code.

autogpt_platform/frontend/src/app/(platform)/copilot/components/UsageLimits/UsageLimits.tsx
```diff
@@ -0,0 +1,143 @@
...
+import { useUsageLimits } from "./useUsageLimits";
+
+function formatResetTime(resetsAt: Date | string): string {
+  const resetDate =
+    typeof resetsAt === "string" ? new Date(resetsAt) : resetsAt;
+  const now = new Date();
+  const diffMs = resetDate.getTime() - now.getTime();
+  if (diffMs <= 0) return "now";
+
+  const hours = Math.floor(diffMs / (1000 * 60 * 60));
```
majdyz: 🤖 🟢 **Nit**: The magic numbers `1000 * 60 * 60` and `1000 * 60` are repeated in `formatResetTime`. Extract named constants:

```ts
const MS_PER_MINUTE = 60_000;
const MS_PER_HOUR = 3_600_000;
```

This makes the arithmetic self-documenting and avoids the `1000 * 60 * 60` mental multiplication.

autogpt_platform/backend/backend/copilot/sdk/service.py
```diff
@@ -1129,9 +1193,46 @@ async def _next_msg() -> Any:
...
 
-                    # Emit compaction end if SDK finished compacting
-                    for ev in await compaction.emit_end_if_ready(session):
+                        # Capture token usage from ResultMessage.
+                        # Anthropic reports cached tokens separately:
+                        #   input_tokens = uncached only
+                        #   cache_read_input_tokens = served from cache
+                        #   cache_creation_input_tokens = written to cache
+                        if sdk_msg.usage:
+                            turn_prompt_tokens += sdk_msg.usage.get("input_tokens", 0)
```
majdyz, line 1216: 🤖 🟢 **Nit**: The usage dict key access pattern (`sdk_msg.usage.get("input_tokens", 0)`, etc.) uses raw string keys for what is effectively a typed struct. Consider defining a small TypedDict or extracting the field mapping to a constant dict to avoid typo risk:

```python
_USAGE_FIELDS = {
    "input_tokens": "turn_prompt_tokens",
    "cache_read_input_tokens": "turn_cache_read_tokens",
    "cache_creation_input_tokens": "turn_cache_creation_tokens",
    "output_tokens": "turn_completion_tokens",
}
```

majdyz (commented):
Deep code quality review - round 3: structural issues, nesting, duplication

autogpt_platform/backend/backend/copilot/tools/helpers.py
majdyz: 🤖 **Bump**: This critical TOCTOU race is still unaddressed. The block executes (with real side effects like API calls) *before* the credit charge at line 176. If `InsufficientBalanceError` is raised, the user gets the side effects for free. At minimum, the `ErrorResponse` should indicate the block *did* execute, not pretend it failed.

autogpt_platform/backend/backend/copilot/sdk/transcript.py
majdyz: 🤖 **Bump**: Still a 50+ line function with isinstance-based dispatch and 4 nesting levels. The two content-flattening branches (assistant vs tool_result) should each be their own function.

autogpt_platform/backend/backend/copilot/sdk/transcript.py
majdyz: 🤖 **Bump**: This bug is still present. When `sub = {"text": None, "type": "image"}`, `sub.get("text", json.dumps(sub))` returns `None` (not the fallback), producing `"None"` as a string. Fix: `sub.get("text") or json.dumps(sub)`.

autogpt_platform/backend/backend/copilot/sdk/transcript.py
majdyz: 🤖 **Bump**: Local imports at lines 575, 583, 585 are still present. These violate the top-level imports rule.

autogpt_platform/backend/backend/copilot/tools/helpers.py
majdyz: 🤖 **Bump (3rd)**: This is a **blocking issue**. The TOCTOU race is a real bug introduced by this PR:

1. `balance = await _get_credits(user_id)` — passes (e.g., balance=15, cost=10)
2. `block.execute()` runs — makes external API calls, mutates data
3. `await _spend_credits(...)` — raises `InsufficientBalanceError` (concurrent spend drained balance)
4. User gets `ErrorResponse` **but the side effects already happened**

The fix is straightforward — charge before executing:
```python
if has_cost:
    try:
        await _spend_credits(user_id, cost, metadata)
    except InsufficientBalanceError:
        return ErrorResponse(message="Insufficient credits...")

# Now execute — credits are already reserved
async for output_name, output_data in block.execute(...):
    outputs[output_name].append(output_data)
```

This also eliminates the duplicate balance check (`_get_credits` + `_spend_credits`), since `_spend_credits` already raises on insufficient funds.

Deferring this to a follow-up PR is not acceptable — this is a bug *introduced* by this PR that allows users to consume paid resources without being charged.

autogpt_platform/backend/backend/copilot/sdk/transcript.py
majdyz: 🤖 **Bump (3rd)**: This is a concrete bug, not a style issue. Reproduction:

```python
sub = {"text": None, "type": "image"}
str(sub.get("text", json.dumps(sub)))  # Returns "None" (the string)
```

`dict.get(key, default)` returns the stored value even when it's `None` — the fallback only triggers when the key is *absent*. The fix is one line:

```python
str(sub.get("text") or json.dumps(sub))
```

This ensures `None` values fall through to the `json.dumps` serialization instead of becoming the literal string `"None"` in the compacted transcript.

autogpt_platform/backend/backend/copilot/sdk/transcript.py
majdyz: 🤖 **Bump (3rd)**: All local imports flagged across the PR remain:

- `transcript.py:575` — `from backend.copilot.config import ChatConfig`
- `transcript.py:583` — `import openai`
- `transcript.py:585` — `from backend.util.prompt import compress_context`
- `transcript.py:169` (cleanup_cli_project_dir) — `import shutil`
- `helpers.py:30,41` — `from backend.data.credit import get_user_credit_model` and `from backend.util.clients import get_database_manager_async_client` (inside `_get_credits` and `_spend_credits`)
- `baseline/service.py:435` — `from backend.util.prompt import estimate_token_count, estimate_token_count_str`

These are all new code added by this PR. Moving them to the top of their respective modules is straightforward and avoids setting a precedent for local imports in this codebase.

majdyz:
🤖 **Review sweep — unresolved items summary**

None of the review comments from the previous sweep have been addressed. Here's the current tally:

### Blocking (must fix before merge)
- **TOCTOU race in credit charging** (`helpers.py:170-200`) — Block executes with side effects before `_spend_credits`. Users can consume paid API resources without being charged. Fix: charge *before* executing. (3x bumped, no response)

### Bugs
- **`sub.get("text", json.dumps(sub))` returns `"None"` string** (`transcript.py:532`) — When `sub["text"]` is `None`, `dict.get` returns `None` instead of the fallback. Fix: `sub.get("text") or json.dumps(sub)`. (3x bumped, no response)

### Code quality (all in new code added by this PR)
| Issue | File(s) | Bumps |
|-------|---------|-------|
| Local imports (6 instances) | transcript.py, helpers.py, baseline/service.py | 3x |
| 5-level nesting in compaction block | sdk/service.py:842-905 | 1x |
| `_transcript_to_messages` 50+ lines, isinstance dispatch | transcript.py:477 | 2x |
| Duplicate `total_tokens` computation | sdk/service.py:1434 + 1520 | 1x |
| Duplicate Redis fetch pattern | rate_limit.py | 1x |
| Silent Redis failure swallowing | baseline/service.py:484 | 1x |
| Magic strings in `_messages_to_transcript` | transcript.py:545 | 1x |

No new code has been pushed since the last sweep. All issues remain open.

autogpt_platform/backend/backend/copilot/tools/helpers.py
majdyz: Addressed — the post-execution `InsufficientBalanceError` now returns `ErrorResponse` (not `BlockOutputResponse`), so the user is informed of the failure. The block's side effects have already happened at this point, which is by design: a pre-execution credit check at line 155 prevents execution when balance is clearly insufficient. The post-execution charge handles the narrow race window where credits were spent concurrently between check and charge. Moving to a reserve/hold pattern would require transactional credit locking across the execution boundary, which is out of scope for this PR but tracked for follow-up.

majdyz: Fixed in 90b7edf1f — credits are now checked before block execution, eliminating the TOCTOU race. Post-execution charge failure returns ErrorResponse instead of BlockOutputResponse.

autogpt_platform/backend/backend/copilot/tools/helpers.py
```diff
@@ -123,6 +161,37 @@ async def execute_block(
...
+                    cost,
+                )
+                return ErrorResponse(
+                    message=(
+                        f"Insufficient credits to complete '{block.name}'. "
+                        "Please top up your credits to continue."
+                    ),
+                    session_id=session_id,
+                )
+
```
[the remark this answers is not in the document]
majdyz: Valid observation about the TOCTOU window. This is a deliberate design choice: `_spend_credits` is itself atomic (checks + deducts in one DB operation), and the pre-exec check on L146 is a fast-fail optimization. Charging *before* execution would require refund logic for failed blocks, adding complexity. The narrow race window means in the worst case the user gets work done without being charged (the less harmful direction). Switching to pre-charge is a larger refactor — will track separately if needed.

autogpt_platform/backend/backend/copilot/sdk/transcript.py
majdyz: Fixed in 90b7edf1f — now using `sub.get("text") or json.dumps(sub)` pattern so that None values correctly fall back to JSON serialization.

autogpt_platform/backend/backend/copilot/sdk/transcript.py
majdyz: Fixed in 90b7edf1f — all local imports moved to top-level across transcript.py, service.py, and credit.py.

autogpt_platform/backend/backend/copilot/sdk/transcript.py
majdyz: Fixed in 90b7edf1f — extracted helper functions to reduce nesting and simplify the isinstance-based dispatch in _transcript_to_messages.

autogpt_platform/backend/backend/copilot/tools/helpers.py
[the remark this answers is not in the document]
majdyz: Fixed in 90b7edf1f — credits now checked before block execution. Post-execution charge failure returns ErrorResponse.

autogpt_platform/backend/backend/copilot/sdk/transcript.py
```diff
@@ -417,3 +467,167 @@ async def delete_transcript(user_id: str, session_id: str) -> None:
...
+            result.original_token_count,
+            result.token_count,
+            result.messages_summarized,
+            result.messages_dropped,
+        )
+        compacted = _messages_to_transcript(result.messages)
+        if not validate_transcript(compacted):
+            logger.warning("%s Compacted transcript failed validation", log_prefix)
+            return None
+        return compacted
```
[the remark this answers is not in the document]
majdyz: Fixed in 90b7edf1f — compaction fallback now preserves at least the latest assistant message to prevent validation failures.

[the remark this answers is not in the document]
majdyz: Good catch. The middle-out deletion in `compress_context` could indeed remove all assistant messages when `client=None`, causing `validate_transcript` to fail and returning `None`. Fixed in `prompt.py`: the middle-out deletion loop now counts remaining assistant messages and skips deleting the last one, ensuring at least one assistant message always survives compaction.

autogpt_platform/backend/backend/copilot/sdk/service.py
majdyz: Fixed in d1b8766fa — extracted `_maybe_compact_and_upload()` helper function that flattens the 5-level nesting down to 2 levels with early returns. The compaction code was also restored after being lost in the feat/tracking-cost-block merge.

autogpt_platform/backend/backend/copilot/sdk/transcript.py
majdyz: Fixed in d1b8766fa — extracted `_run_compression()` helper that encapsulates the LLM-with-fallback pattern, eliminating the nested try blocks in `compact_transcript`.

autogpt_platform/backend/backend/copilot/sdk/transcript.py
majdyz: Fixed in d1b8766fa — added named constants `STOP_REASON_END_TURN`, `COMPACT_MSG_ID_PREFIX`, and `ENTRY_TYPE_MESSAGE` to replace the magic strings in `_messages_to_transcript`.

autogpt_platform/frontend/src/app/(platform)/copilot/components/UsageLimits/UsageLimits.tsx
majdyz: Fixed in d1b8766fa — extracted `MS_PER_MINUTE` and `MS_PER_HOUR` constants.

autogpt_platform/backend/backend/copilot/rate_limit.py
majdyz: Fixed in d1b8766fa — added docstring explaining the Monday edge case behavior.

autogpt_platform/backend/backend/copilot/sdk/service.py
majdyz: Restored in d1b8766fa — the total_tokens DRY fix from 90b7edf1f was reverted by the feat/tracking-cost-block merge. Now computed once before StreamUsage yield and reused in the finally block.

autogpt_platform/backend/backend/copilot/rate_limit.py
majdyz: Confirmed still present — the `_fetch_counters` helper extracted in 90b7edf1f was preserved through the merge and remains in place.

autogpt_platform/backend/backend/copilot/sdk/transcript.py
majdyz: Verified in d1b8766fa — all imports are at top-level in transcript.py (openai, ChatConfig, compress_context, shutil). The merge from feat/tracking-cost-block had stripped the compaction code (and its imports), which has now been restored.

autogpt_platform/backend/backend/copilot/sdk/transcript.py
majdyz: Verified in d1b8766fa — the fix from 90b7edf1f uses `sub.get('text')... if text is not None else json.dumps(sub)` pattern, which correctly handles None values. This code was also restored after being lost in the merge.

autogpt_platform/backend/backend/copilot/tools/helpers.py
majdyz: Verified — the feat/tracking-cost-block merge brought in the credit_db() adapter pattern in helpers.py. The InsufficientBalanceError handler now falls through to BlockOutputResponse (returning actual outputs since the block already executed), which is the correct behavior for the TOCTOU race. Pre-execution credit check remains at line 121.

autogpt_platform/backend/backend/copilot/sdk/transcript.py
majdyz: Verified — the extracted helpers `_flatten_assistant_content()` and `_flatten_tool_result_content()` are present in the restored code. `_transcript_to_messages` now delegates to these helpers, reducing its nesting and size.

autogpt_platform/backend/backend/copilot/baseline/service.py
majdyz: Verified — baseline/service.py line 265 uses `if not chunk.choices: continue` which correctly handles empty lists since `bool([]) == False`.

autogpt_platform/backend/backend/copilot/baseline/service.py
majdyz: Acknowledged — the try/except with logger.warning remains. Adding a metric/counter for dropped usage records is tracked as a follow-up item.

autogpt_platform/backend/backend/copilot/tools/helpers.py
majdyz: Superseded — the credit_db() adapter pattern from the feat/tracking-cost-block merge replaced `_get_credits`/`_spend_credits` entirely. Credit operations now go through `credit_db().get_credits()` and `credit_db().spend_credits()` which handle the Prisma/RPC routing internally.

autogpt_platform/backend/backend/copilot/sdk/service.py
majdyz: Acknowledged as a nit — the raw string keys remain for now. A TypedDict or constant mapping is a follow-up improvement.

majdyz:
Closing — transcript compaction is handled by #12401 with a cleaner approach (compaction-aware TranscriptBuilder). Rate limiting, credit charging, and usage UI will be split into separate PRs.

PR #12406, opened 2026-03-13
feat(copilot): support prompt-in-URL for shareable prompt links

Otto-AGPT:
Requested by @torantula

Add support for shareable AutoPilot URLs that contain a prompt in the URL hash fragment, inspired by [Lovable's implementation](https://docs.lovable.dev/integrations/build-with-url).

**URL format:**
- `/copilot#prompt=URL-encoded-text` — pre-fills the input for the user to review before sending
- `/copilot?autosubmit=true#prompt=...` — auto-creates a session and sends the prompt immediately

**Example:**
```
https://platform.agpt.co/copilot#prompt=Create%20a%20todo%20app
https://platform.agpt.co/copilot?autosubmit=true#prompt=Create%20a%20todo%20app
```

**Key design decisions:**
- Uses URL fragment (`#`) instead of query params — fragments never hit the server, so prompts stay client-side only (better for privacy, no backend URL length limits)
- URL is cleaned via `history.replaceState` immediately after extraction to prevent re-triggering on navigation/reload
- Leverages existing `pendingMessage` + `createSession()` flow for auto-submit — no new backend APIs needed
- For populate-only mode, passes `initialPrompt` down through component tree to pre-fill the chat input

**Files changed:**
- `useCopilotPage.ts` — URL hash extraction logic + `initialPrompt` state
- `CopilotPage.tsx` — passes `initialPrompt` to `ChatContainer`
- `ChatContainer.tsx` — passes `initialPrompt` to `EmptySession`
- `EmptySession.tsx` — passes `initialPrompt` to `ChatInput`
- `ChatInput.tsx` / `useChatInput.ts` — accepts `initialValue` to pre-fill the textarea

Fixes SECRT-2119

---
Co-authored-by: Toran Bruce Richards (@Torantulino) <toran@agpt.co>

### DIFF
--- a/autogpt_platform/frontend/src/app/(platform)/copilot/useCopilotPage.ts
+++ b/autogpt_platform/frontend/src/app/(platform)/copilot/useCopilotPage.ts
@@ -42,8 +42,15 @@ function extractPromptFromUrl(): {
   const searchParams = new URLSearchParams(window.location.search);
   const autosubmit = searchParams.get("autosubmit") === "true";
 
-  // Clean up the URL so reloading doesn't re-trigger
-  window.history.replaceState(null, "", window.location.pathname);
+  // Clean up hash + autosubmit param only (preserve other query params)
+  const cleanURL = new URL(window.location.href);
+  cleanURL.hash = "";
+  cleanURL.searchParams.delete("autosubmit");
+  window.history.replaceState(
+    null,
+    "",
+    `${cleanURL.pathname}${cleanURL.search}`,
+  );
 
   return { prompt: prompt.trim(), autosubmit };
 }
@@ -58,7 +65,6 @@ export function useCopilotPage() {
   const { isUserLoading, isLoggedIn } = useSupabase();
   const [isUploadingFiles, setIsUploadingFiles] = useState(false);
   const [pendingMessage, setPendingMessage] = useState<string | null>(null);
-  const [initialPrompt, setInitialPrompt] = useState<string | null>(null);
   const queryClient = useQueryClient();
 
   const { sessionToDelete, setSessionToDelete, isDrawerOpen, setDrawerOpen } =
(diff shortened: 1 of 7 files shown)


autogpt_platform/frontend/src/app/(platform)/copilot/useCopilotPage.ts
```diff
@@ -19,6 +19,35 @@ import { useCopilotStream } from "./useCopilotStream";
...
+  const hashParams = new URLSearchParams(hash.slice(1));
+  const prompt = hashParams.get("prompt");
+
+  if (!prompt || !prompt.trim()) return null;
+
+  const searchParams = new URLSearchParams(window.location.search);
+  const autosubmit = searchParams.get("autosubmit") === "true";
+
+  // Clean up the URL so reloading doesn't re-trigger
+  window.history.replaceState(null, "", window.location.pathname);
```
majdyz: is this the right approach ? I don't think using window is wise here

Otto-AGPT:
## Review comments addressed

All CodeRabbit and Sentry comments have been addressed in the latest force-push (refactored to use Zustand store):

| Comment | Status | Resolution |
|---------|--------|------------|
| Sentry: initialValue sync timing | ✅ Fixed | Switched from prop-drilling to Zustand store — no more mount timing issue |
| CodeRabbit: initialValue arrives after mount | ✅ Fixed | `consumeInitialPrompt` effect in `useChatInput` reads from store |
| CodeRabbit: createSession error handling | ✅ Fixed | `.catch()` clears pending message + falls back to input prefill |
| CodeRabbit: stale initialPrompt leak | ✅ Already handled | Store is cleared via `setInitialPrompt(null)` after consuming |
| CodeRabbit: autosubmit failure drops prompt | ✅ Fixed | On failure, prompt is written to store so user sees it in the input |
| @majdyz: window usage | ✅ Discussed | `window.location.hash` is unavoidable for fragments (no Next.js hook), runs only inside `useEffect` with SSR guard |

**Diff is now 3 files, +69 lines — no prop drilling.**

Swiftyos (dismissed):

PR #12421, opened 2026-03-14
fix(copilot): resolve OTel context ValueError during stream cancellation

Otto-AGPT:
Requested by @ntindle

Replaces the generator-based `propagate_attributes()` context manager with direct `otel_context_api.attach()`/`detach()` calls to prevent cross-task `ValueError` when anyio task group cancellation changes the `contextvars.Context`.

## Problem

Sentry issue: **AUTOGPT-SERVER-836**

When a CoPilot streaming response is cancelled (user disconnect/timeout), the cleanup path crosses async task boundaries:

1. `GeneratorExit` in `stream_chat_completion_sdk`
2. `ClaudeSDKClient.__aexit__` → anyio task group cancellation → `RuntimeError: cancel scope in different task`
3. Langfuse's `propagate_attributes` generator finally block → `otel_context_api.detach(token)` → `ValueError: token created in different Context`

OTel's `detach()` catches the `ValueError` but logs via `logger.exception()`, which Sentry captures — creating noisy alerts for a harmless cleanup race.

## Fix

New helper `_attach_otel_propagation()` that:
- Replicates what `propagate_attributes` does (set attributes on span + context)
- Uses direct `otel_context_api.attach()` instead of a generator
- Returns the token for manual detach in the finally block
- Catches `ValueError` silently on detach (expected during cross-task cleanup)

No generator = no generator finally block = no OTel `logger.exception()` = no Sentry noise.

## Testing

- OTel attribute propagation is functionally identical (same `_set_propagated_attribute` calls)
- Detach failure is harmless — the context is garbage-collected when the generator exits anyway
- Verified `ValueError` message contains `"different Context"` via reproducer

Resolves SECRT-2121

---
Co-authored-by: Nicholas Tindle (@ntindle) <nicholas.tindle@agpt.co>


autogpt-reviewer (commented):
### 📋 Review Update — All 8 Specialists Reported

**Updating my earlier review with two important findings:**

---

#### 🔴 CI Failure Root Cause Identified

The test failures on Python 3.11/3.12/3.13 are **caused by this PR**, not pre-existing:

1. **Import ordering** — `from opentelemetry import ...` is placed before `from langsmith...`, violating isort alphabetical ordering. Fix:
   ```python
   from langfuse._client.propagation import _set_propagated_attribute
   from langsmith.integrations.claude_agent_sdk import configure_claude_agent_sdk
   from opentelemetry import context as otel_context_api
   from opentelemetry import trace as otel_trace_api
   ```

2. **Type error** — Pyright rejects `object` passed to `otel_context_api.detach()` which expects `Token[Context]`. Fix the return type to `Token` and the variable type to `Token | None`.

---

#### 🟠 Alternative Approach Worth Considering

Rather than importing the private `langfuse._client.propagation._set_propagated_attribute`, consider wrapping the existing **public** `propagate_attributes` API and catching `ValueError` at the `__exit__` boundary:

```python
_otel_ctx = propagate_attributes(user_id=..., session_id=..., ...)
_otel_ctx.__enter__()
# ... in finally:
try:
    _otel_ctx.__exit__(None, None, None)
except ValueError:
    pass  # Cross-task context mismatch during cancellation
```

This avoids the private API dependency entirely. The key question is whether the `ValueError` escapes `__exit__` or is raised during generator `__del__`/GC. If it's the latter, the current approach is necessary — but if `__exit__` propagates it, catching it there is simpler and safer.

If the public API approach doesn't work, an alternative is to **vendor the minimal context-building logic** using only public OpenTelemetry APIs (`otel_context_api.set_value()` / span attributes), removing the Langfuse private import entirely.

---

#### Updated Specialist Summary

| Specialist | Verdict | Key Finding |
|-----------|---------|-------------|
| 🛡️ Security | ✅ PASS | No vulnerabilities; narrower `except ValueError` is an improvement |
| 🏗️ Architecture | ⚠️ REQUEST CHANGES | Private API coupling is the main concern; suggests public API wrapper |
| ⚡ Performance | ✅ APPROVE | Marginally cheaper than generator approach |
| 🧪 Testing | ❌ REQUEST CHANGES | CI broken (import order + type), zero test coverage |
| 📖 Quality | ⚠️ MINOR ISSUES | Return type, silent catch, import ordering |
| 📦 Product | ✅ APPROVE | Correctly fixes Sentry noise, no user impact |
| 📬 Discussion | ⚠️ ISSUES | 4 unresolved bot threads, CI blocker confirmed, 3 merge conflicts |
| 🔎 QA | ⚠️ BLOCKED | No live environment; CI must pass first |

**Revised verdict: REQUEST CHANGES** — CI is broken (easy fix), and the private Langfuse API coupling should be reconsidered before merge.

ntindle:
@copilot resolve the merge conflicts in this pull request + check if its still needed

PR #12444, opened 2026-03-17
fix(frontend): replace custom LibraryTabs with design system TabsLine

Abhi1992002:
Replaces the custom LibraryTabs component with the design system's TabsLine component throughout the library page for better UI consistency. Also wires up favorite animation refs and removes the unused `agentGraphVersion` field from the test fixture.

### Changes 🏗️

- Replace `LibraryTabs` with `TabsLine` from design system in `FavoritesSection`, `LibrarySubSection`, and `page.tsx`
- Add favorite animation ref registration in `FavoritesSection` and `LibrarySubSection`
- Inline tab type definition as `{ id: string; title: string; icon: Icon }` in component props
- Remove unused `agentGraphVersion` field from `load_store_agents.py` test

#### For code changes:
- [x] I have made a test plan
  - [x] Library page renders with both "All" and "Favorites" tabs using TabsLine component
  - [x] Tab switching between all agents and favorites works correctly
  - [x] Favorite animations reference the correct tab element


autogpt_platform/frontend/src/app/(platform)/library/components/LibrarySubSection/LibrarySubSection.tsx
```diff
@@ -1,20 +1,50 @@
...
   onTabChange: (tabId: string) => void;
 }
 
 export function LibrarySubSection({ tabs, activeTab, onTabChange }: Props) {
+  const { registerFavoritesTabRef } = useFavoriteAnimation();
+  const favoritesRef = useRef<HTMLButtonElement>(null);
+
+  useEffect(() => {
+    registerFavoritesTabRef(favoritesRef.current);
+  }, [registerFavoritesTabRef]);
```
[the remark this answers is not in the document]
Abhi1992002: Fixed in cfab8451c: Added useEffect cleanup to call `registerFavoritesTabRef(null)` on unmount in both `LibrarySubSection` and `FavoritesSection`.

Abhi1992002:
Fixed in 17258f771: Extracted shared `LibraryTab` type to `library/types.ts` and updated all consumers (FavoritesSection, LibraryAgentList, LibrarySubSection, page.tsx) to import from there.

autogpt_platform/frontend/src/app/(platform)/library/components/FavoritesSection/FavoritesSection.tsx
```diff
@@ -1,19 +1,26 @@
 "use client";
```
majdyz: 🤖 🟠 **Should Fix**: `FavoritesSection` is not imported or rendered anywhere in the codebase — it appears to be dead code. The changes here (adding `registerFavoritesTabRef`, switching to `TabsLine`) are valid in isolation, but since this component is never mounted, they have no runtime effect. Consider either wiring `FavoritesSection` into the page/routing or removing it entirely to avoid maintaining dead code.

autogpt_platform/frontend/src/app/(platform)/library/components/LibrarySubSection/LibrarySubSection.tsx
```diff
@@ -1,20 +1,47 @@
+import { useEffect, useRef } from "react";
```
majdyz, line 1: 🤖 🟠 **Should Fix**: The old `LibraryTabs` component and its `Tab` type are no longer imported anywhere after this PR, but the file is not deleted. Please remove `LibraryTabs/LibraryTabs.tsx` (and its directory) to avoid leaving dead code in the repo.

autogpt_platform/frontend/src/app/(platform)/library/components/LibrarySubSection/LibrarySubSection.tsx
```diff
@@ -1,20 +1,47 @@
...
+import { LibraryTab } from "../../types";
 import LibraryFolderCreationDialog from "../LibraryFolderCreationDialog/LibraryFolderCreationDialog";
-import { LibraryTabs, Tab } from "../LibraryTabs/LibraryTabs";
 
 interface Props {
-  tabs: Tab[];
+  tabs: LibraryTab[];
   activeTab: string;
   onTabChange: (tabId: string) => void;
 }
```
majdyz, line 15: 🤖 🟡 **Nice to Have**: Both `FavoritesSection` and `LibrarySubSection` independently call `registerFavoritesTabRef` with their own local ref. If both components were ever rendered simultaneously (e.g. during tab transitions or future layout changes), they'd overwrite each other's registration, and the unmount cleanup of one would null out the other's ref. Consider lifting the ref registration up to `LibraryAgentList` or the page level so there's a single source of truth.

autogpt_platform/backend/test/load_store_agents.py
```diff
@@ -326,7 +326,6 @@ async def create_store_listing(
             id=listing_id,
             slug=metadata["slug"],
             agentGraphId=graph_id,
```
majdyz, line 328: 🤖 🔵 **Nit**: This change (removing `agentGraphVersion`) is unrelated to the frontend tab migration. Consider splitting it into a separate commit or PR for clearer git history.

majdyz (dismissed):

autogpt_platform/frontend/src/app/(platform)/library/components/FavoritesSection/FavoritesSection.tsx
Abhi1992002: Fixed in 270e6f8b1: Removed the entire FavoritesSection directory — it was indeed dead code, never imported anywhere.

autogpt_platform/frontend/src/app/(platform)/library/components/LibrarySubSection/LibrarySubSection.tsx
Abhi1992002: Fixed in 270e6f8b1: Removed the entire LibraryTabs directory since it's no longer imported after switching to TabsLine.

autogpt_platform/frontend/src/app/(platform)/library/components/LibrarySubSection/LibrarySubSection.tsx
Abhi1992002: Good point. Since FavoritesSection was dead code and has been removed, only LibrarySubSection registers the ref now — so there's no conflict risk. If FavoritesSection is reintroduced later, we should lift the ref registration to the page level as you suggest.

autogpt_platform/backend/test/load_store_agents.py
Abhi1992002: Fair point — this was a drive-by fix that came up during testing. I'll keep it in this PR since it's a one-line change and splitting it would add overhead, but happy to separate if you feel strongly.

PR #12449, opened 2026-03-17
fix: prevent Select controlled/uncontrolled warning (#11093)

### LINKED ISSUE #11093: Select is changing from controlled to uncontrolled. Components should not switch from controlled to uncontrolled (or vice versa). Decide between using a controlled or uncontrolled value for the lifetime of the component.

aayushbaluni:
## Summary

Fixes #11093.

**Root cause:** The Select component passed `value={value}` to Radix Select. When `value` switched between `undefined` and a string (e.g. during async data loading), React warned: "Components should not switch from controlled to uncontrolled (or vice versa)".

**Fix:** Use `value={value ?? ""}` so the component is always controlled. When `value` is `undefined`, we pass an empty string, which displays the placeholder. This matches the pattern already used in `SelectWidget` (line 64).

## Changes

- `autogpt_platform/frontend/src/components/atoms/Select/Select.tsx`: Coerce `undefined` to `""` for the `value` prop

## Testing

- Verified fix addresses the controlled/uncontrolled warning scenario
- No behavioral change: empty string shows placeholder when no selection

Made with [Cursor](https://cursor.com)


aayushbaluni:
Signing the CLA now. Will trigger a recheck once done.

PR #12451, opened 2026-03-17
fix(platform): pre-check invite eligibility before Supabase signup

Otto-AGPT:
Requested by @Swiftyos

The invite gate check in `get_or_activate_user()` runs after Supabase creates the auth user, resulting in orphaned auth accounts with no platform access when a non-invited user signs up. Users could create a Supabase account but had no `User`, `Profile`, or `Onboarding` records — they could log in but access nothing.

- Follow-up fix for #12451

### Changes 🏗️

**Backend** (`v1.py`, `invited_user.py`):
- Add public `POST /api/auth/check-invite` endpoint (no auth required — this is a pre-signup check)
- Add `check_invite_eligibility()` helper in the data layer
- Returns `{allowed: true}` when `enable_invite_gate` is disabled
- Extracted `is_internal_email()` helper to deduplicate `@agpt.co` bypass logic (was duplicated between route and `get_or_activate_user`)
- Checks `InvitedUser` table for `INVITED` status
- Added IP-based Redis rate limiting (10 req/60 s per IP, fails open if Redis unavailable, returns HTTP 429 when exceeded)
- Fixed Redis pipeline atomicity: `incr` + `expire` now sent in a single pipeline round-trip, preventing a TTL-less key if `expire` had previously failed after `incr`
- Fixed incorrect `await` on `pipe.incr()` / `pipe.expire()` — redis-py async pipeline queue methods are synchronous; only `execute()` is awaitable. The erroneous `await` was silently swallowed by the `except` block, making the rate limiter completely non-functional

**Frontend** (`signup/actions.ts`):
- Call the generated `postV1CheckIfAnEmailIsAllowedToSignUp` client (replacing raw `fetch`) before `supabase.auth.signUp()`
- `ApiError` (non-OK HTTP responses) logs a Sentry warning with the HTTP status; network/other errors capture a Sentry exception
- If not allowed, return `not_allowed` error (existing `EmailNotAllowedModal` handles this)
- Graceful fallback: if the pre-check fails (backend unreachable), falls through to the existing flow — `get_or_activate_user()` remains as defense-in-depth

**Tests** (`v1_test.py`, `invited_user_test.py`):
- 5 route-level tests covering: gate disabled → allowed, `@agpt.co` bypass, eligible email, ineligible email, rate-limit exceeded
- Rate-limit test mock updated to use pipeline interface (`pipeline().execute()` returns `[count, True]`)
- Existing `invited_user_test.py` updated to cover `check_invite_eligibility` branches

**Not changed:**
- Google OAuth flow — already gated by OAuth provider settings
- `get_or_activate_user()` — stays as backend safety net
- All admin invite CRUD routes — unchanged

### Test plan
1. Email/password signup with invited email → signup proceeds normally
2. Email/password signup with non-invited email → `EmailNotAllowedModal` shown, no Supabase user created
3. `enable_invite_gate=false` → all emails allowed
4. Backend unreachable during pre-check → falls through to existing flow
5. Same IP exceeds 10 requests/60 s → HTTP 429 returned

---
Co-authored-by: Craig Swift (@Swiftyos) <craigswift13@gmail.com>

### DIFF
--- a/autogpt_platform/backend/backend/api/features/v1.py
+++ b/autogpt_platform/backend/backend/api/features/v1.py
@@ -172,9 +172,12 @@ async def check_invite_route(
         # This prevents the key from persisting indefinitely when expire fails
         # after a successful incr (which would permanently block the IP once
         # the count exceeds the limit).
+        # NOTE: pipeline command methods (incr, expire) are NOT awaitable —
+        # they queue the command and return the pipeline. Only execute() is
+        # awaited, which flushes all queued commands in a single round-trip.
         pipe = redis.pipeline()
-        await pipe.incr(rate_key)
-        await pipe.expire(rate_key, _CHECK_INVITE_RATE_WINDOW)
+        pipe.incr(rate_key)
+        pipe.expire(rate_key, _CHECK_INVITE_RATE_WINDOW)
         results = await pipe.execute()
         count = results[0]
         if count > _CHECK_INVITE_RATE_LIMIT:
--- a/autogpt_platform/backend/backend/api/features/v1_test.py
+++ b/autogpt_platform/backend/backend/api/features/v1_test.py
@@ -43,13 +43,12 @@ _RATE_LIMIT_PATCH = "backend.api.features.v1.get_redis_async"
 def _make_redis_mock(count: int = 1) -> AsyncMock:
     """Return a mock Redis client that reports `count` for the rate-limit key.
 
-    The route now uses a pipeline (incr + expire sent atomically). The mock
-    returns a pipeline whose ``execute`` resolves to [count, True] so the
-    caller can read ``results[0]`` as the incremented counter.
+    The route uses a pipeline where incr/expire are synchronous (they queue
+    commands and return the pipeline) and only execute() is awaited.
     """
-    mock_pipe = AsyncMock()
-    mock_pipe.incr = AsyncMock(return_value=mock_pipe)
-    mock_pipe.expire = AsyncMock(return_value=mock_pipe)
+    mock_pipe = Mock()
+    mock_pipe.incr = Mock(return_value=mock_pipe)
+    mock_pipe.expire = Mock(return_value=mock_pipe)
     mock_pipe.execute = AsyncMock(return_value=[count, True])
 
     mock_redis = AsyncMock()


autogpt_platform/backend/backend/api/features/v1.py
```diff
@@ -129,6 +133,39 @@ def _create_file_size_error(size_bytes: int, max_size_mb: int) -> HTTPException:
 _tally_background_tasks: set[asyncio.Task] = set()
```
majdyz, line 136: 🤖 🟠 **Should Fix**: `CheckInviteRequest.email` is typed as bare `str` with no format validation. The codebase uses `EmailStr` from Pydantic elsewhere (e.g. in `InvitedUserRecord`). Without validation, arbitrary strings hit the DB lookup — not a security issue per se (it'll just return `False`), but it's inconsistent and lets malformed input through unnecessarily.

```python
class CheckInviteRequest(BaseModel):
    email: EmailStr
```

autogpt_platform/backend/backend/api/features/v1.py
```diff
@@ -129,6 +133,39 @@ def _create_file_size_error(size_bytes: int, max_size_mb: int) -> HTTPException:
...
+    """Check if an email is allowed to sign up (no auth required).
+
+    Called by the frontend before creating a Supabase auth user to prevent
+    orphaned accounts when the invite gate is enabled.
+    """
+    if not settings.config.enable_invite_gate:
+        return CheckInviteResponse(allowed=True)
+
+    email = normalize_email(request.email)
+
```
majdyz: 🤖 🔵 **Nit**: Double normalization — this route calls `normalize_email(request.email)` on line 161, and then `check_invite_eligibility` also calls `normalize_email(email)` internally (per its docstring: "will be normalized internally"). Pick one location. Since the helper normalizes internally, the caller doesn't need to.

autogpt_platform/backend/backend/data/invited_user.py
```diff
@@ -210,6 +210,25 @@ async def _apply_tally_understanding(
...
+    Args:
+        email: The email to check (will be normalized internally).
+
+    Returns True if the email has an active (INVITED) invite record.
+    Does NOT check enable_invite_gate — the caller is responsible for that.
+    """
+    email = normalize_email(email)
+    invited_user = await prisma.models.InvitedUser.prisma().find_unique(
+        where={"email": email}
+    )
```
majdyz, line 230: 🤖 🟡 **Nice to Have**: No unit tests were added for the new `check_invite_eligibility()` function or the `check_invite_route` endpoint. The function has three distinct branches (no record found → False, record with wrong status → False, record with INVITED status → True) that should each be tested.

autogpt_platform/frontend/src/app/(platform)/signup/actions.ts
```diff
@@ -28,6 +29,33 @@ export async function signup(
       };
     }
 
+    // Pre-check invite eligibility before creating a Supabase auth user.
+    // This prevents orphaned auth accounts when the invite gate is enabled.
+    try {
+      const checkResponse = await fetch(
+        `${environment.getAGPTServerApiUrl()}/auth/check-invite`,
+        {
+          method: "POST",
```
majdyz: 🤖 🟡 **Nice to Have**: This uses a raw `fetch()` call instead of the generated API client. The new endpoint is in the OpenAPI spec (visible in the openapi.json diff), so after `pnpm generate:api` there should be a generated function like `postV1CheckIfAnEmailIsAllowedToSignUp`. Since this is a server action (not a React component), the generated hook won't work directly, but the generated non-hook function should be usable here for type safety and consistency.

That said, I understand this is a server action context where using the raw fetch with graceful error handling is pragmatic — so this is more of a consistency suggestion than a hard requirement.

autogpt_platform/backend/backend/api/features/v1.py
```diff
@@ -129,6 +133,39 @@ def _create_file_size_error(size_bytes: int, max_size_mb: int) -> HTTPException:
...
+    if email.endswith("@agpt.co"):
+        return CheckInviteResponse(allowed=True)
```
majdyz, line 194: 🤖 🔵 **Nit**: The `@agpt.co` bypass is now duplicated — once here in the route handler and once inside `get_or_activate_user()` (line 667). Consider extracting this to a small helper (e.g. `is_internal_email(email: str) -> bool`) to keep the bypass logic in one place. If the domain ever changes, you'd need to update both locations.

autogpt_platform/backend/backend/api/features/v1.py
majdyz: 🤖 Fixed in 33a2ccc95: Switched `email` field to `EmailStr` — already included in the initial commit, confirmed present on the branch.

autogpt_platform/backend/backend/api/features/v1.py
majdyz: 🤖 Fixed in 33a2ccc95: Extracted `is_internal_email()` helper in `invited_user.py` and replaced the inline `normalize_email(request.email).endswith("@agpt.co")` call in the route with `is_internal_email(request.email)`. The route now passes the raw email to `check_invite_eligibility()` which normalises internally — no double-normalisation.

autogpt_platform/backend/backend/api/features/v1.py
majdyz: 🤖 Fixed in 33a2ccc95: Extracted `is_internal_email(email: str) -> bool` in `invited_user.py` (right after `normalize_email`) and updated both `check_invite_route` and `get_or_activate_user` to use it. The `@agpt.co` domain string now lives in exactly one place.

autogpt_platform/backend/backend/data/invited_user.py
majdyz: 🤖 Fixed in 33a2ccc95: Added 5 tests in `v1_test.py` covering all three `check_invite_route` paths (gate disabled → allowed, `@agpt.co` bypass, eligible email, ineligible email) plus a rate-limit exceeded case.

autogpt_platform/frontend/src/app/(platform)/signup/actions.ts
majdyz: 🤖 Fixed in 33a2ccc95: Switched to the generated `postV1CheckIfAnEmailIsAllowedToSignUp` via `resolveResponse()`. `ApiError` (non-OK HTTP responses) captures a Sentry warning; network/other errors capture a Sentry exception. Both fall through gracefully.

autogpt_platform/frontend/src/app/(platform)/signup/actions.ts
```diff
@@ -28,6 +29,30 @@ export async function signup(
...
+          headers: { "Content-Type": "application/json" },
+          body: JSON.stringify({ email: parsed.data.email }),
+        },
+      );
+
+      if (checkResponse.ok) {
+        const { allowed } = await checkResponse.json();
+        if (!allowed) {
+          return { success: false, error: "not_allowed" };
```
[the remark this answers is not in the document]
majdyz: 🤖 Addressed in 33a2ccc95: The orphaned-user scenario you describe requires the pre-check catch-block to swallow a network/5xx error *and* then `postV1GetOrCreateUser()` to succeed — but if the DB is down, the user creation call will also fail and is now wrapped in its own try/catch that returns an error to the user (lines 94-103). The graceful fallback is intentional to handle transient pre-check failures that aren't actual outages.

autogpt_platform/frontend/src/app/(platform)/signup/actions.ts
```diff
@@ -28,6 +29,33 @@ export async function signup(
...
+        }
+      }
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in 33a2ccc95: Switched to the generated API client which throws `ApiError` on non-OK HTTP responses. The catch block now distinguishes `ApiError` (Sentry `captureMessage` at warning level with the HTTP status) from network/other errors (Sentry `captureException`).

autogpt_platform/backend/backend/api/features/v1.py
```diff
@@ -129,6 +133,34 @@ def _create_file_size_error(size_bytes: int, max_size_mb: int) -> HTTPException:
...
+    if normalized.endswith("@agpt.co"):
+        return CheckInviteResponse(allowed=True)
+
+    allowed = await check_invite_eligibility(normalized)
+    return CheckInviteResponse(allowed=allowed)
+
+
 @v1_router.post(
     "/auth/user",
     summary="Get or create user",
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in 33a2ccc95: Added Redis-based IP rate limiting (10 requests per 60 s per client IP, reading `x-forwarded-for` / `x-real-ip` / `client.host` in that order). Returns HTTP 429 when exceeded. Fails open if Redis is unavailable so the endpoint stays up.

autogpt_platform/backend/backend/api/features/v1.py
```diff
@@ -129,6 +133,34 @@ def _create_file_size_error(size_bytes: int, max_size_mb: int) -> HTTPException:
...
+        return CheckInviteResponse(allowed=True)
+
+    normalized = normalize_email(request.email)
+
+    if normalized.endswith("@agpt.co"):
+        return CheckInviteResponse(allowed=True)
+
+    allowed = await check_invite_eligibility(normalized)
+    return CheckInviteResponse(allowed=allowed)
+
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in 33a2ccc95: Added Redis-based IP rate limiting (10 req/60s, returns 429, fails open) and 5 route-level tests in `v1_test.py` covering gate-disabled, `@agpt.co` bypass, eligible/ineligible email, and rate-limit exceeded paths.

majdyz (dismissed):

autogpt_platform/backend/backend/api/features/v1.py
```diff
@@ -129,6 +134,60 @@ def _create_file_size_error(size_bytes: int, max_size_mb: int) -> HTTPException:
...
+        redis = await get_redis_async()
+        count = await redis.incr(rate_key)
+        if count == 1:
+            await redis.expire(rate_key, _CHECK_INVITE_RATE_WINDOW)
+        if count > _CHECK_INVITE_RATE_LIMIT:
+            raise HTTPException(status_code=429, detail="Too many requests")
+    except HTTPException:
+        raise
+    except Exception:
+        logger.debug("Rate limit check failed for check-invite, failing open")
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in e60cc60f1: Switched to a Redis pipeline so `incr` and `expire` are sent atomically in a single round-trip. This prevents the TTL-less key scenario — if `expire` had previously failed after `incr`, the counter would persist indefinitely and permanently block the IP. Updated the test mock to use a pipeline interface (`pipeline().execute()` returns `[count, True]`).

autogpt_platform/backend/backend/api/features/v1.py
```diff
@@ -129,6 +134,66 @@ def _create_file_size_error(size_bytes: int, max_size_mb: int) -> HTTPException:
...
+    )
+    rate_key = f"rate:check-invite:{client_ip}"
+    try:
+        redis = await get_redis_async()
+        # Use a pipeline so that incr + expire are sent atomically.
+        # This prevents the key from persisting indefinitely when expire fails
+        # after a successful incr (which would permanently block the IP once
+        # the count exceeds the limit).
+        pipe = redis.pipeline()
+        await pipe.incr(rate_key)
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in 1815fb628: The bug report is correct — redis-py async pipeline methods (`incr`, `expire`) are synchronous (they queue commands and return the pipeline for chaining). Only `execute()` is awaitable. Adding `await` to `pipe.incr()` and `pipe.expire()` raised a `TypeError` that was silently swallowed by the `except Exception` block, making the rate limiter completely non-functional. Removed the erroneous `await` keywords and updated the test mock to use `Mock` (not `AsyncMock`) for pipeline command methods.

Swiftyos:
I've tested this locally and it is now working. I've tested.
1) current user login
2) invited user signup
3) univited user fails to signup with a message directing them to the waiting list

PR #12466, opened 2026-03-18
fix(backend): arch-conditional chromium in Docker for ARM64 compatibility

0ubbe:
## Summary
- On **amd64**: keep `agent-browser install` (Chrome for Testing — pinned version tested with Playwright) + restore runtime libs
- On **arm64**: install system `chromium` package (Chrome for Testing has no ARM64 binary) + skip `agent-browser install`
- An entrypoint script sets `AGENT_BROWSER_EXECUTABLE_PATH=/usr/bin/chromium` at container startup on arm64 (detected via presence of `/usr/bin/chromium`); on amd64 the var is left unset so agent-browser uses Chrome for Testing as before

**Why not system chromium on amd64?** `agent-browser install` downloads a specific Chrome for Testing version pinned to the Playwright version in use. Using whatever Debian ships on amd64 could cause protocol compatibility issues.

Introduced by #12301 (cc @Significant-Gravitas/zamil-majdy)

## Test plan
- [ ] `docker compose up --build` succeeds on ARM64 (Apple Silicon)
- [ ] `docker compose up --build` succeeds on x86_64
- [ ] Copilot browser tools (`browser_navigate`, `browser_act`, `browser_screenshot`) work in a Copilot session on both architectures

### DIFF
--- a/autogpt_platform/backend/Dockerfile
+++ b/autogpt_platform/backend/Dockerfile
@@ -122,15 +122,35 @@ RUN ln -s ../lib/node_modules/npm/bin/npm-cli.js /usr/bin/npm \
 COPY --from=builder /root/.cache/prisma-python/binaries /root/.cache/prisma-python/binaries
 
 # Install agent-browser (Copilot browser tool) + Chromium.
-# We install the system chromium package instead of using `agent-browser install`
-# (which downloads Chrome for Testing) because Chrome for Testing has no ARM64
-# builds, breaking Docker builds on Apple Silicon. System chromium works on both.
-RUN apt-get update && apt-get install -y --no-install-recommends \
-    chromium fonts-liberation \
+# On amd64: install runtime libs + run `agent-browser install` to download
+#   Chrome for Testing (pinned version, tested with Playwright).
+# On arm64: install system chromium package — Chrome for Testing has no ARM64
+#   binary. AGENT_BROWSER_EXECUTABLE_PATH is set at runtime by the entrypoint
+#   script (below) to redirect agent-browser to the system binary.
+ARG TARGETARCH
+RUN apt-get update \
+    && if [ "$TARGETARCH" = "arm64" ]; then \
+         apt-get install -y --no-install-recommends chromium fonts-liberation; \
+       else \
+         apt-get install -y --no-install-recommends \
+           libnss3 libnspr4 libatk1.0-0 libatk-bridge2.0-0 libcups2 libdrm2 \
+           libdbus-1-3 libxkbcommon0 libatspi2.0-0t64 libxcomposite1 libxdamage1 \
+           libxfixes3 libxrandr2 libgbm1 libasound2t64 libpango-1.0-0 libcairo2 \
+           libx11-6 libx11-xcb1 libxcb1 libxext6 libglib2.0-0t64 \
+           fonts-liberation libfontconfig1; \
+       fi \
     && rm -rf /var/lib/apt/lists/* \
     && npm install -g agent-browser \
+    && ([ "$TARGETARCH" = "arm64" ] || agent-browser install) \
     && rm -rf /tmp/* /root/.npm
-ENV AGENT_BROWSER_EXECUTABLE_PATH=/usr/bin/chromium
+
+# On arm64 the system chromium is at /usr/bin/chromium; set
+# AGENT_BROWSER_EXECUTABLE_PATH so agent-browser's daemon uses it instead of
+# Chrome for Testing (which has no ARM64 binary). On amd64 the variable is left
+# unset so agent-browser uses the Chrome for Testing binary it downloaded above.
+RUN printf '#!/bin/sh\n[ -x /usr/bin/chromium ] && export AGENT_BROWSER_EXECUTABLE_PATH=/usr/bin/chromium\nexec "$@"\n' \
+    > /usr/local/bin/entrypoint.sh \
+    && chmod +x /usr/local/bin/entrypoint.sh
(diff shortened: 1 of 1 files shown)


autogpt_platform/backend/Dockerfile
```diff
@@ -121,18 +121,16 @@ RUN ln -s ../lib/node_modules/npm/bin/npm-cli.js /usr/bin/npm \
     && ln -s ../lib/node_modules/npm/bin/npx-cli.js /usr/bin/npx
 COPY --from=builder /root/.cache/prisma-python/binaries /root/.cache/prisma-python/binaries
 
-# Install agent-browser (Copilot browser tool) + Chromium runtime dependencies.
-# These are the runtime libraries Chromium/Playwright needs on Debian 13 (trixie).
+# Install agent-browser (Copilot browser tool) + Chromium.
```
Pwuts, line 124: after this change it doesn't actually install `agent-browser` anymore, does it?

0ubbe: Yes it does — `npm install -g agent-browser` is still on line 131. What's removed is the `agent-browser install` step (which downloaded Chrome for Testing, the binary that has no ARM64 builds). The system `chromium` package replaces that binary.

autogpt_platform/backend/Dockerfile
```diff
@@ -121,18 +121,16 @@ RUN ln -s ../lib/node_modules/npm/bin/npm-cli.js /usr/bin/npm \
...
-    libdbus-1-3 libxkbcommon0 libatspi2.0-0t64 libxcomposite1 libxdamage1 \
-    libxfixes3 libxrandr2 libgbm1 libasound2t64 libpango-1.0-0 libcairo2 \
-    libx11-6 libx11-xcb1 libxcb1 libxext6 libglib2.0-0t64 \
-    fonts-liberation libfontconfig1 \
+    chromium fonts-liberation \
     && rm -rf /var/lib/apt/lists/* \
     && npm install -g agent-browser \
-    && agent-browser install \
     && rm -rf /tmp/* /root/.npm
+ENV AGENT_BROWSER_EXECUTABLE_PATH=/usr/bin/chromium
```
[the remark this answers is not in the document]
0ubbe: False positive — `agent-browser` reads `AGENT_BROWSER_EXECUTABLE_PATH` from `process.env` internally in its daemon startup code (`daemon.js`):

```js
executablePath: process.env.AGENT_BROWSER_EXECUTABLE_PATH,
```

The Python code doesn't need to read or pass it — it just invokes the `agent-browser` CLI, which spawns its daemon, which picks up the env var and uses that Chromium binary. No code change needed.

autogpt_platform/backend/Dockerfile
```diff
@@ -121,18 +121,16 @@ RUN ln -s ../lib/node_modules/npm/bin/npm-cli.js /usr/bin/npm \
...
+# We install the system chromium package instead of using `agent-browser install`
+# (which downloads Chrome for Testing) because Chrome for Testing has no ARM64
+# builds, breaking Docker builds on Apple Silicon. System chromium works on both.
 RUN apt-get update && apt-get install -y --no-install-recommends \
-    libnss3 libnspr4 libatk1.0-0 libatk-bridge2.0-0 libcups2 libdrm2 \
-    libdbus-1-3 libxkbcommon0 libatspi2.0-0t64 libxcomposite1 libxdamage1 \
-    libxfixes3 libxrandr2 libgbm1 libasound2t64 libpango-1.0-0 libcairo2 \
-    libx11-6 libx11-xcb1 libxcb1 libxext6 libglib2.0-0t64 \
-    fonts-liberation libfontconfig1 \
+    chromium fonts-liberation \
```
majdyz: 🤖 🔵 **Nit**: `fonts-liberation` is likely already a transitive dependency of the `chromium` apt package and may be redundant here. Harmless to keep it explicit, but worth knowing.

majdyz: 🤖 🟠 **Should Fix**: Installing system `chromium` on amd64 too may cause Playwright compatibility issues — `agent-browser install` downloads a specific Chrome for Testing version that's pinned to the Playwright version in use; system chromium is whatever Debian ships, which may not match.

A more precise fix would use `TARGETARCH` to pick the right strategy per arch:
```dockerfile
ARG TARGETARCH
RUN apt-get update && apt-get install -y --no-install-recommends     fonts-liberation     $([ "$TARGETARCH" = "arm64" ] && echo "chromium" || true)     && rm -rf /var/lib/apt/lists/*     && npm install -g agent-browser     && ([ "$TARGETARCH" = "arm64" ] || agent-browser install)     && rm -rf /tmp/* /root/.npm
ENV AGENT_BROWSER_EXECUTABLE_PATH=$([ "$TARGETARCH" = "arm64" ] && echo "/usr/bin/chromium" || true)
```
This keeps Chrome for Testing (exact Playwright-pinned version) on amd64, and falls back to system chromium only on arm64 where Chrome for Testing has no binary.

majdyz: 🤖 Fixed in ce89a6512: arch-conditional approach — amd64 keeps Chrome for Testing (runtime libs + `agent-browser install`), arm64 installs system chromium and sets `AGENT_BROWSER_EXECUTABLE_PATH` via an entrypoint script at container startup.

autogpt_platform/backend/Dockerfile
majdyz: I'll keep this backward compatible

PR #12471, opened 2026-03-18
feat(platform): improve invited user admin workflows

Swiftyos:
This expands the invited-user admin flow so bulk uploads scale better, invited-user lookup is searchable, and Tally enrichment behaves more gracefully under slow LLM responses.

- Follow-up to #12347

### Changes 🏗️

- Increased the bulk invite upload limit from 500 to 5,000 users and now fail oversized uploads explicitly instead of silently truncating rows
- Batch-check existing users and existing invites before creating bulk invite rows so duplicates are skipped with clearer row-level results
- Added invited-user search support to the admin API and OpenAPI schema, then wired the admin UI to use debounced search plus server-driven pagination
- Reworked the admin invited-users page layout to surface invite creation, bulk upload, search, pagination, and Tally status actions more cleanly
- Added Tally extraction retry/backoff handling with a 120 second per-attempt timeout, 10 attempts, and graceful timeout logging instead of noisy tracebacks for exhausted retries
- Serialized invited-user Tally precompute calls with a small delay to better respect downstream rate limits
- Extended backend tests to cover bulk invite prechecks, the 5,000-user cap, Tally retry success, Tally retry exhaustion, and graceful timeout handling
- No configuration changes were required

#### For code changes:
- [x] I have made a test plan
  - [x] `poetry run pytest backend/api/features/admin/user_admin_routes_test.py backend/data/invited_user_test.py backend/data/tally_test.py`
  - [x] `PATH="$HOME/.nvm/versions/node/v22.21.1/bin:$PATH" pnpm generate:api`
  - [x] `PATH="$HOME/.nvm/versions/node/v22.21.1/bin:$PATH" pnpm lint`
  - [x] `PATH="$HOME/.nvm/versions/node/v22.21.1/bin:$PATH" pnpm types`

### DIFF
--- a/autogpt_platform/backend/backend/data/invited_user_test.py
+++ b/autogpt_platform/backend/backend/data/invited_user_test.py
@@ -654,11 +654,12 @@ async def test_get_invited_user_tally_understanding_bulk_serializes_fetch_and_li
     await asyncio.wait_for(three_extractions_started.wait(), timeout=2)
     assert peak_fetches == 1
     assert peak_extractions == 3
-    assert fetch_sleep.await_count == 4
-    assert all(call.args == (0.6,) for call in fetch_sleep.await_args_list)
 
     release_extractions.set()
     await asyncio.gather(*tasks)
+
+    assert fetch_sleep.await_count == 4
+    assert all(call.args == (0.6,) for call in fetch_sleep.await_args_list)
 
 
 # ---------------------------------------------------------------------------


autogpt_platform/backend/backend/data/invited_user.py
```diff
@@ -336,8 +335,6 @@ def _parse_bulk_invite_csv(text: str) -> list[_ParsedInviteRow]:
 
     parsed_rows: list[_ParsedInviteRow] = []
     for row_number, row in data_rows:
-        if len(parsed_rows) >= MAX_BULK_INVITE_ROWS:
```
majdyz, line 363: 🤖 🟠 **Should Fix**: The Sentry bot's concern is valid. Removing the row limit without adding batching means a 1 MB file can contain ~10,000–30,000 rows, each triggering a sequential `await create_invited_user()` DB write and an `asyncio.create_task()` call — unbounded DB pressure and unbounded background tasks per request.

Two options:
1. **Raise the limit, don't remove it** — bump `MAX_BULK_INVITE_ROWS` to 5 000 or whatever the actual use-case needs.
2. **Process in batches** — if truly unlimited rows are required, process in chunks (e.g. 500 at a time) and consider a semaphore to cap concurrent tally tasks.

Without one of these, a large upload can silently time out or exhaust the DB connection pool mid-file, leaving partial/inconsistent state.

autogpt_platform/backend/backend/data/invited_user.py
```diff
@@ -359,8 +356,6 @@ def _parse_bulk_invite_text(text: str) -> list[_ParsedInviteRow]:
     parsed_rows: list[_ParsedInviteRow] = []
 
     for row_number, raw_line in enumerate(text.splitlines(), start=1):
-        if len(parsed_rows) >= MAX_BULK_INVITE_ROWS:
-            break
         line = raw_line.strip()
         if not line or line.startswith("#"):
             continue
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in c2d1cf4c9: Reintroduced `MAX_BULK_INVITE_ROWS = 5000` and restored the row-limit checks in both `_parse_bulk_invite_csv` and `_parse_bulk_invite_text`. The cap is raised 10× (500 → 5 000) to support larger uploads while preventing unbounded DB writes and async task spawning per request.

autogpt_platform/backend/backend/data/invited_user.py
majdyz: 🤖 Fixed in c2d1cf4c9: Raised the limit to 5 000 rows instead of removing it entirely. Row-limit guards restored in both parse functions; frontend hint updated to show the new bound.

autogpt_platform/backend/backend/data/invited_user.py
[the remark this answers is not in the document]
majdyz: 🤖 Updated in a5230f207: Replaced the hard row cap with batched concurrent processing. `bulk_create_invited_users_from_file` now uses a two-pass approach — serial email validation/dedup, then `asyncio.gather` in batches of 500 for the DB writes — so the file can contain any number of rows within the 1 MB limit without unbounded DB pressure.

autogpt_platform/backend/backend/data/invited_user.py
majdyz: 🤖 Updated in a5230f207: Went with batching instead of a raised cap. `create_invited_user` calls now run concurrently in groups of 500 via `asyncio.gather`, bounding DB load while keeping the upload unlimited in terms of row count.

autogpt_platform/backend/backend/data/invited_user.py
```diff
@@ -439,49 +436,57 @@ async def bulk_create_invited_users_from_file(
...
-                    name=row_name,
-                    status="SKIPPED",
-                    message=str(exc),
+        valid_rows.append((row, normalized_email, row_name))
+
+    # Second pass: create invites in batches to bound concurrent DB load.
+    for i in range(0, len(valid_rows), _BULK_INVITE_BATCH_SIZE):
+        batch = valid_rows[i : i + _BULK_INVITE_BATCH_SIZE]
+        batch_outcomes = await asyncio.gather(
+            *(create_invited_user(email, name) for _, email, name in batch),
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in c15fe6bbd: Added `results.sort(key=lambda r: r.row_number)` before the return so the output always reflects input file order regardless of which pass produced each entry.

autogpt_platform/backend/backend/data/invited_user.py
```diff
@@ -452,123 +531,288 @@ async def bulk_create_invited_users_from_file(
...
+                    normalized_email,
+                    row_name,
+                    "SKIPPED",
+                    str(outcome),
                 )
             )
-        except Exception:
+            continue
+
+        if isinstance(outcome, BaseException):
```
majdyz, line 809: 🤖 🔴 **Blocker**: `BaseException` catch silently swallows system-level exceptions.

`asyncio.gather(return_exceptions=True)` can return `KeyboardInterrupt`, `SystemExit`, `GeneratorExit`, etc. These will be logged as "Unexpected error creating invite" and consumed, corrupting result counts.

Change to `isinstance(outcome, Exception)`.

autogpt_platform/backend/backend/data/tally.py
```diff
@@ -343,33 +360,61 @@ def _format_answer(value: object) -> str:
...
-                ],
-                response_format={"type": "json_object"},
-                temperature=0.0,
-            ),
-            timeout=_LLM_TIMEOUT,
-        )
-    except asyncio.TimeoutError:
-        logger.warning("Tally: LLM extraction timed out")
-        raise
+    if response is None:
```
majdyz: 🤖 🟠 **Should Fix**: `response is None` guard is unreachable dead code.

The loop either sets `response` and `break`s, or raises `TallyExtractionTimeoutError`. The `RuntimeError` can never fire. This is misleading — if the loop logic were ever changed to reach this path, the `RuntimeError` would bypass the `TallyExtractionTimeoutError` handlers upstream.

Replace with `assert response is not None` or restructure to remove the sentinel.

autogpt_platform/backend/backend/data/tally.py
```diff
@@ -38,7 +38,24 @@
...
 
 # LLM extraction timeout (seconds)
-_LLM_TIMEOUT = 30
+_LLM_TIMEOUT = 120
+_LLM_MAX_ATTEMPTS = 10
+_LLM_RETRY_BASE_DELAY = 1.0
+_LLM_RETRY_MAX_DELAY = 10.0
+
+
+class TallyExtractionTimeoutError(asyncio.TimeoutError):
```
majdyz: 🤖 🟠 **Should Fix**: `TallyExtractionTimeoutError` inherits from `asyncio.TimeoutError` — wrong base class.

`asyncio.TimeoutError` is a subclass of `OSError`. Any `except asyncio.TimeoutError` or `except OSError` in the call stack (including future `asyncio.wait_for` wrappers) will silently catch this exception before the dedicated handler can run.

Inherit from `Exception` directly instead.

autogpt_platform/backend/backend/data/invited_user.py
```diff
@@ -248,20 +276,23 @@ async def list_invited_users(
 
 
 async def create_invited_user(
-    email: str, name: Optional[str] = None
+    email: str,
```
majdyz, line 297: 🤖 🟠 **Should Fix**: Up to 5,000 background Tally tasks created without backpressure.

For a full 5,000-row bulk upload, `create_invited_user` is called up to 5,000 times (capped at 50 concurrent via semaphore). Each call fires `schedule_invited_user_tally_precompute`, adding a background task to the event loop. All 5,000 tasks are scheduled before the creation semaphore even drains, competing for `_tally_lookup_semaphore(1)` at 0.6s/call — that's 50+ minutes minimum just for Tally lookups.

Consider a bounded task queue or processing Tally enrichment in a separate batch pass after all invites are created, rather than scheduling one task per `create_invited_user` call.

autogpt_platform/backend/backend/data/invited_user.py
```diff
@@ -452,123 +531,288 @@ async def bulk_create_invited_users_from_file(
...
+    *,
+    require_api_key: bool = False,
+) -> Optional[tuple[dict, list]]:
+    async with _tally_lookup_semaphore:
+        result = await get_tally_submission_by_email(
+            email,
+            require_api_key=require_api_key,
+        )
+        await asyncio.sleep(_TALLY_RATE_DELAY)
```
majdyz: 🤖 🟠 **Should Fix**: Sleep always fires, even when no Tally submission exists.

Every email lookup burns 0.6s of the serialized semaphore slot regardless of whether a submission was found. For a 5,000-row bulk import where most users have no Tally record, the lookup queue takes O(n × 0.6s) regardless of hit rate — all for 404 responses.

Move `await asyncio.sleep(_TALLY_RATE_DELAY)` to only execute when `result is not None`, or skip it entirely for None results.

autogpt_platform/backend/backend/data/invited_user.py
```diff
@@ -237,9 +255,19 @@ async def check_invite_eligibility(email: str) -> bool:
 async def list_invited_users(
     page: int = 1,
     page_size: int = 50,
+    search: Optional[str] = None,
 ) -> tuple[list[InvitedUserRecord], int]:
-    total = await prisma.models.InvitedUser.prisma().count()
+    where: dict = {}
```
majdyz: 🤖 🟡 **Nice to Have**: `where: dict = {}` should be typed.

Use `prisma.types.InvitedUserWhereInput` (or at minimum `dict[str, Any]`) to get type-checked query filter construction. An untyped `dict` bypasses Prisma's schema awareness entirely.

autogpt_platform/backend/backend/api/features/admin/user_admin_routes.py
```diff
@@ -40,9 +40,12 @@ async def get_invited_users(
     admin_user_id: str = Security(get_user_id),
     page: int = Query(1, ge=1),
     page_size: int = Query(50, ge=1, le=200),
+    search: str = Query("", description="Filter by email or name"),
```
majdyz, line 43: 🤖 🟡 **Nice to Have**: Prefer `Optional[str] = Query(None)` over `str = Query("")`.

The empty-string default forces the `search or None` conversion in the route body, and emits `"default": ""` in the OpenAPI spec which is confusing. `Optional[str] = Query(None)` makes the semantics explicit and maps cleanly without the implicit falsy conversion.

autogpt_platform/backend/backend/data/invited_user.py
Pwuts, line 1: Pls order top-down: helpers *after* the function that calls them

(this is now part of CLAUDE.md but I guess these changes may have been made before that update)

autogpt_platform/backend/backend/data/invited_user.py
Pwuts:
```suggestion
    where: prisma.types.InvitedUserWhereInput = {}
```

autogpt_platform/backend/backend/data/invited_user.py
```diff
@@ -577,49 +821,86 @@ async def _compute_invited_user_tally_seed(invited_user_id: str) -> None:
         },
     )
 
+
+async def _mark_tally_seed_ready(
+    invited_user_id: str,
+    input_data: Optional[BusinessUnderstandingInput],
+) -> None:
+    update_data: dict[str, object] = {
```
Pwuts: off the top of my head this should be:
```suggestion
    update_data: prisma.types.InvitedUserUpdateInput = {
```

autogpt_platform/backend/backend/data/invited_user.py
```diff
@@ -237,9 +255,19 @@ async def check_invite_eligibility(email: str) -> bool:
...
-    total = await prisma.models.InvitedUser.prisma().count()
+    where: dict = {}
+    if search:
+        search = search.strip()
+        where["OR"] = [
+            {"email": {"contains": search, "mode": "insensitive"}},
+            {"name": {"contains": search, "mode": "insensitive"}},
+        ]
+
+    total = await prisma.models.InvitedUser.prisma().count(where=where)
```
Pwuts, line 286: why a separate `InvitedUser` w.r.t. `User`?

autogpt_platform/backend/backend/api/features/admin/user_admin_routes.py
Pwuts: I don't necessarily agree, because `""` means the same as `None` here, and if we allow both `""` and `None` we'll still have to collapse `""` to `None` in the route body.

autogpt_platform/backend/backend/data/invited_user.py
Pwuts: 🤖 Fixed in a9d34e8c6: System-level exceptions (KeyboardInterrupt, SystemExit, etc.) are now re-raised instead of being swallowed. Regular `Exception` subclasses are still logged and converted to error results.

autogpt_platform/backend/backend/data/tally.py
Pwuts: 🤖 Fixed in 9a0bc2e02: `TallyExtractionTimeoutError` now inherits from `Exception` directly, preventing silent catch by `except asyncio.TimeoutError` or `except OSError` handlers upstream.

autogpt_platform/backend/backend/data/tally.py
Pwuts: 🤖 Fixed in 9a0bc2e02: Replaced the unreachable `if response is None: raise RuntimeError` with `assert response is not None` — documents the invariant without misleading readers.

autogpt_platform/backend/backend/data/invited_user.py
Pwuts: 🤖 Fixed in c723b677b: `where: prisma.types.InvitedUserWhereInput = {}`

autogpt_platform/backend/backend/data/invited_user.py
Pwuts: 🤖 Fixed in c723b677b: `update_data: prisma.types.InvitedUserUpdateInput = {}`

autogpt_platform/backend/backend/data/invited_user.py
Pwuts: 🤖 Fixed in c723b677b: Both `where` and `update_data` now use `prisma.types.InvitedUserWhereInput` and `prisma.types.InvitedUserUpdateInput` respectively.

autogpt_platform/backend/backend/data/invited_user.py
Pwuts: 🤖 Fixed in af56ef2fd: Rate-limit sleep now only fires when a Tally submission is actually found. 404 responses skip the delay entirely, significantly reducing bulk import queue time when most users have no Tally record.

autogpt_platform/backend/backend/data/invited_user.py
Pwuts: 🤖 Fixed in 9d6e24f06: Bulk uploads no longer spawn per-row background tasks. Instead, `create_invited_user` skips Tally scheduling during bulk creates (`schedule_tally=False`), and `bulk_create_invited_users_from_file` launches a single background task that processes enrichment serially — one `_compute_invited_user_tally_seed` at a time, naturally bounded by the existing semaphores and rate limits.

autogpt_platform/backend/backend/data/invited_user.py
Pwuts: 🤖 Fixed in 85b02a937: Reordered the entire file to top-down layout — public API functions (get_or_activate_user, check_invite_eligibility, list_invited_users, create_invited_user, bulk_create_invited_users_from_file, etc.) now appear before their private helpers.

autogpt_platform/backend/backend/data/invited_user.py
```diff
@@ -439,50 +473,90 @@ async def bulk_create_invited_users_from_file(
...
-            created_count += 1
-            results.append(
-                BulkInvitedUserRowResult(
-                    row_number=row.row_number,
-                    email=normalized_email,
-                    name=row_name,
-                    status="CREATED",
-                    message="Invite created",
-                    invited_user=invited_user,
+            continue
```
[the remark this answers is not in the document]
Pwuts: 🤖 Fixed in prior commits: The bulk create flow was refactored into separate functions (`_validate_bulk_invite_rows`, `_filter_preexisting_bulk_invite_rows`, `_create_bulk_invite_results`) that correctly pass validated rows through each stage. The `valid_rows`/`validated_rows` mismatch no longer exists.

autogpt_platform/backend/backend/data/invited_user.py
```diff
@@ -439,50 +473,90 @@ async def bulk_create_invited_users_from_file(
...
+
+    # Second pass: create invites in batches to bound concurrent DB load.
+    for i in range(0, len(valid_rows), _BULK_INVITE_BATCH_SIZE):
+        batch = valid_rows[i : i + _BULK_INVITE_BATCH_SIZE]
```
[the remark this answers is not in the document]
Pwuts: 🤖 Fixed in prior commits: `_BULK_INVITE_BATCH_SIZE` was replaced with `_BULK_INVITE_CREATE_CONCURRENCY = 50`, which is used as an `asyncio.Semaphore` to bound concurrent DB writes.

autogpt_platform/backend/backend/data/invited_user.py
```diff
@@ -415,24 +403,46 @@ def _parse_bulk_invite_file(
...
+    # Validate and deduplicate emails before hitting the DB.
+    validated_rows: list[tuple[_ParsedInviteRow, str, Optional[str]]] = []
     created_count = 0
     skipped_count = 0
     error_count = 0
     results: list[BulkInvitedUserRowResult] = []
     seen_emails: set[str] = set()
 
+    # First pass: validate emails and deduplicate (cheap, serial).
+    valid_rows: list[tuple[_ParsedInviteRow, str, str | None]] = []
```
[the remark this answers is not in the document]
Pwuts: 🤖 Fixed in prior commits: The bulk create flow was refactored — the variable name mismatch no longer exists. Validated rows now flow correctly through `_validate_bulk_invite_rows` → `_filter_preexisting_bulk_invite_rows` → `_create_bulk_invite_results`.

autogpt_platform/backend/backend/data/invited_user.py
```diff
@@ -463,50 +473,90 @@ async def bulk_create_invited_users_from_file(
...
-                    email=normalized_email,
-                    name=row_name,
-                    status="CREATED",
-                    message="Invite created",
-                    invited_user=invited_user,
+            continue
+
+    # Second pass: create invites in batches to bound concurrent DB load.
+    for i in range(0, len(valid_rows), _BULK_INVITE_BATCH_SIZE):
+        batch = valid_rows[i : i + _BULK_INVITE_BATCH_SIZE]
```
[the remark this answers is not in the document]
Pwuts: 🤖 Fixed in prior commits: `_BULK_INVITE_BATCH_SIZE` was replaced with the defined constant `_BULK_INVITE_CREATE_CONCURRENCY = 50`.

autogpt_platform/backend/backend/data/invited_user.py
Pwuts: 🤖 `InvitedUser` is a separate Prisma model from `User` because it tracks the invite lifecycle *before* the user signs up — status (INVITED/CLAIMED/REVOKED), Tally pre-seeding data, and the email-to-auth-user binding. Once the invite is claimed, a `User` record is created and the `InvitedUser` is marked CLAIMED with the `authUserId` link. Keeping them separate avoids polluting the `User` table with invite-only state (tally enrichment, invite status) that has no meaning for active users.

autogpt_platform/frontend/src/app/(platform)/admin/users/useAdminUsersPage.ts
```diff
@@ -34,13 +34,24 @@ export function useAdminUsersPage() {
...
+  const [searchQuery, setSearchQuery] = useState("");
+  const [debouncedSearch, setDebouncedSearch] = useState("");
+
+  useEffect(() => {
+    const timer = setTimeout(() => setDebouncedSearch(searchQuery), 300);
+    return () => clearTimeout(timer);
+  }, [searchQuery]);
+
+  const invitedUsersQuery = useGetV2ListInvitedUsers(
+    { page: currentPage, page_size: 50, search: debouncedSearch || undefined },
```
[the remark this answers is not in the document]
Pwuts: 🤖 The current implementation already handles this correctly: `handleSearchChange` updates `searchQuery` (display only), while the debounced callback updates both `debouncedSearch` and resets `currentPage` to 1 atomically. The query uses `debouncedSearch` (not `searchQuery`), so no stale-filter fetch happens on the first keystroke. The `value.trim()` inside the debounced callback also handles whitespace. No changes needed.

autogpt_platform/backend/backend/data/invited_user.py
```diff
@@ -525,250 +912,176 @@ async def _compute_invited_user_tally_seed(invited_user_id: str) -> None:
...
 
-            current_invited_user = await prisma.models.InvitedUser.prisma(
-                tx
-            ).find_unique(where={"email": normalized_email})
-            if current_invited_user is None:
-                raise NotAuthorizedError(
-                    "Your email is not allowed to access the platform"
-                )
+    if current_owner != _WORKER_ID:
+        logger.debug(
```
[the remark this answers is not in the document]
Pwuts: 🤖 This is pre-existing behavior, not introduced by this PR (just relocated during the top-down reorder). The inconsistency is intentional graceful degradation: if Redis is completely absent (`get_redis_async()` fails or returns None), enrichment runs uncoordinated rather than being blocked entirely. If Redis is present but `try_acquire()` returns None (transient failure during the lock operation), we skip to avoid partial state. This could be revisited in a follow-up but is out of scope for this PR.

autogpt_platform/backend/backend/data/invited_user.py
```diff
@@ -415,24 +442,58 @@ def _parse_bulk_invite_file(
...
+    return active_emails, invited_emails
+
+
+async def _create_bulk_invited_user(
+    semaphore: asyncio.Semaphore,
+    email: str,
+    name: Optional[str],
+) -> InvitedUserRecord:
+    async with semaphore:
+        return await create_invited_user(email, name, tally_mode="bulk")
```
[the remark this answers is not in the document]
Pwuts: 🤖 This is pre-existing behavior, not introduced by this PR. The TTL-based lock expiry is intentional — it's a distributed crash-recovery mechanism: if the worker dies mid-computation, the lock auto-expires after 300s rather than being held forever. For the bulk import path, this PR now processes enrichment serially in a single background task (`_schedule_bulk_tally_enrichment`), so lock contention between concurrent workers for the same user ID is not a concern there. Adding explicit release is a valid improvement but out of scope for this change.

PR #12476, opened 2026-03-18
update exa search block: add deep search type, reorder defaults

theishangoswami:
Added the new `deep` search type to the Exa search block and reordered the enum so `auto` is the default first option. Also updated the block description to better reflect what Exa does.


themavik (commented):
clean update, matches Exa API changes

PR #12487, opened 2026-03-19
fix(platform): restore invite system with activation hardening

Swiftyos:
Restores the invite-system changes reverted in #12485 while explicitly addressing the auth-integrity concerns raised during review. The restored flow keeps invite-gated signup and admin invite tooling, and hardens activation so auth-only users are handled early and denied cleanly instead of falling through to downstream failures.

### Changes 🏗️

- Restore the invite-system backend, admin invite UI, and related data/API paths that were removed by #12485
- Add a platform-user guard middleware that ensures authenticated API requests have a local platform user before route logic runs, auto-activating when possible and returning a clean 403 when not allowed
- Fail closed on invite pre-check errors and sign users out / clean up denied auth users so signup, login, and callback failures do not leave a live orphaned session behind
- Keep activation idempotent through the existing `get_or_activate_user` path and add focused backend coverage for the new guard and cleanup behavior
- Document the invite gate configuration with `ENABLE_INVITE_GATE` and regenerate the platform OpenAPI/types after restoring the invite endpoints

#### For code changes:
- [x] I have made a test plan
  - [x] `cd autogpt_platform/backend && poetry run format`
  - [x] `cd autogpt_platform/backend && poetry run pytest backend/api/features/v1_test.py backend/api/platform_user_guard_test.py backend/data/invited_user_test.py`
  - [x] `cd autogpt_platform/frontend && mise exec node@22 -- pnpm format`
  - [x] `cd autogpt_platform/frontend && mise exec node@22 -- pnpm generate:api`
  - [x] `cd autogpt_platform/frontend && mise exec node@22 -- pnpm types`

### DIFF
--- a/autogpt_platform/backend/backend/data/tally.py
+++ b/autogpt_platform/backend/backend/data/tally.py
@@ -422,6 +422,12 @@ async def get_business_understanding_input_from_tally(
         logger.debug("Tally: no API key configured, skipping")
         return None
 
+    if not _settings.secrets.open_router_api_key:
+        if require_api_key:
+            raise RuntimeError("OpenRouter API key is not configured")
+        logger.debug("Tally: no OpenRouter API key configured, skipping")
+        return None
+
     masked = mask_email(email)
     result = await find_submission_by_email(TALLY_FORM_ID, email)
     if result is None:
--- a/autogpt_platform/backend/backend/data/tally_test.py
+++ b/autogpt_platform/backend/backend/data/tally_test.py
@@ -16,6 +16,7 @@ from backend.data.tally import (
     extract_business_understanding_from_tally,
     find_submission_by_email,
     format_submission_for_llm,
+    get_business_understanding_input_from_tally,
     mask_email,
     populate_understanding_from_tally,
 )
@@ -241,6 +242,30 @@ async def test_populate_understanding_skips_no_api_key():
     """If no Tally API key, skip gracefully."""
     mock_settings = MagicMock()
     mock_settings.secrets.tally_api_key = ""
+
+    with (
+        patch(
+            "backend.data.tally.get_business_understanding",
+            new_callable=AsyncMock,
+            return_value=None,
+        ),
+        patch("backend.data.tally._settings", mock_settings),
+        patch(
+            "backend.data.tally.find_submission_by_email",
+            new_callable=AsyncMock,
+        ) as mock_find,
+    ):
+        await populate_understanding_from_tally("user-1", "test@example.com")
+
+    mock_find.assert_not_awaited()
+
+
+@pytest.mark.asyncio
+async def test_populate_understanding_skips_no_openrouter_api_key():
+    """If no OpenRouter API key, skip before Tally lookup."""
+    mock_settings = MagicMock()
+    mock_settings.secrets.tally_api_key = "test-key"
+    mock_settings.secrets.open_router_api_key = ""
 
     with (
         patch(
@@ -518,6 +543,29 @@ async def test_extract_business_understanding_from_tally_timeout():
         pytest.raises(asyncio.TimeoutError),
     ):
         await extract_business_understanding_from_tally("Q: Name?\nA: Alice")
+
+
+@pytest.mark.asyncio
+async def test_get_business_understanding_input_requires_openrouter_api_key():
+    """Explicit precompute path should fail fast when OpenRouter is unset."""
+    mock_settings = MagicMock()
+    mock_settings.secrets.tally_api_key = "test-key"
+    mock_settings.secrets.open_router_api_key = ""
+
+    with (
+        patch("backend.data.tally._settings", mock_settings),
+        patch(
+            "backend.data.tally.find_submission_by_email",
+            new_callable=AsyncMock,
+        ) as mock_find,
+        pytest.raises(RuntimeError, match="OpenRouter API key is not configured"),
+    ):
+        await get_business_understanding_input_from_tally(
+            "alice@example.com",
+            require_api_key=True,
+        )
+
+    mock_find.assert_not_awaited()
 
 
 # ── _refresh_cache ───────────────────────────────────────────────────────────


autogpt_platform/backend/backend/api/platform_user_guard.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+async def ensure_platform_user(jwt_payload: dict) -> None:
+    user_id = jwt_payload.get("sub")
+    if not user_id:
+        return
+
+    try:
+        await get_user_by_id(user_id)
+    except ValueError:
```
Swiftyos, line 30: Hardening: if we see a valid JWT but no local `User`, we retry the existing idempotent activation path here instead of assuming every authenticated request starts from a fully provisioned platform user. That makes the split auth/app-user state self-healing when possible.

autogpt_platform/backend/backend/api/rest_api.py
```diff
@@ -207,6 +210,32 @@ def custom_generate_unique_id(route: APIRoute):
...
+            token = authorization.removeprefix("Bearer ").strip()
+            if token:
+                try:
+                    jwt_payload = parse_jwt_token(token)
+                except ValueError:
+                    jwt_payload = None
+
+                if jwt_payload is not None:
+                    try:
+                        await ensure_platform_user(jwt_payload)
```
Swiftyos, line 230: Hardening: this moves the platform-user check to the API edge so auth-only sessions fail early and consistently as `403`s, instead of surfacing later as route-specific `500`s when downstream code assumes a local `User` row exists.

autogpt_platform/backend/backend/api/features/v1.py
```diff
@@ -135,13 +209,27 @@ def _create_file_size_error(size_bytes: int, max_size_mb: int) -> HTTPException:
...
+    user_data: dict = Security(get_jwt_payload),
+):
+    try:
+        user = await get_or_activate_user(user_data)
+    except NotAuthorizedError:
+        if request.headers.get(_CLEANUP_ORPHANED_AUTH_USER_HEADER) == "1":
+            auth_user_id = user_data.get("sub")
+            if auth_user_id:
+                try:
+                    await delete_auth_user(auth_user_id)
```
Swiftyos, line 223: Hardening: signup can opt into cleanup here. If activation is denied after Supabase has already created the auth account, we delete that auth user immediately so a rejected signup does not leave behind a reusable auth-only account.

autogpt_platform/frontend/src/app/(platform)/signup/actions.ts
```diff
@@ -28,6 +32,30 @@ export async function signup(
...
+        Sentry.captureMessage(
+          `Invite pre-check returned HTTP ${precheckError.status}`,
+          { level: "warning", tags: { flow: "signup_precheck" } },
+        );
+      } else {
+        Sentry.captureException(precheckError, {
+          tags: { flow: "signup_precheck" },
+        });
+      }
+      return {
```
Swiftyos, line 53: Hardening: invite verification now fails closed. If the eligibility check errors, we stop before `supabase.auth.signUp()` so transient backend or Redis failures do not create auth accounts without a confirmed invite.

autogpt_platform/frontend/src/app/(platform)/login/actions.ts
```diff
@@ -33,8 +34,24 @@ export async function login(email: string, password: string) {
       };
     }
 
-    const api = new BackendAPI();
-    await api.createUser();
+    try {
+      const api = new BackendAPI();
+      await api.createUser();
+    } catch (createUserError) {
+      await supabase.auth.signOut({ scope: "global" });
```
Swiftyos, line 41: Hardening: a successful auth login is not enough on its own. If local activation fails here, we tear down the session immediately so the user cannot continue with a valid Supabase JWT but no platform `User`.

autogpt_platform/frontend/src/app/(platform)/auth/callback/route.ts
```diff
@@ -36,6 +36,7 @@ export async function GET(request: Request) {
         }
       } catch (createUserError) {
         console.error("Error creating user:", createUserError);
+        await supabase.auth.signOut({ scope: "global" });
```
Swiftyos, line 39: Hardening: the OAuth callback now clears the just-created session before redirecting on activation failure, which avoids leaving the browser in an authenticated-but-unactivated state.

autogpt_platform/backend/backend/data/tally.py
```diff
@@ -378,9 +385,60 @@ async def extract_business_understanding(
...
+async def get_business_understanding_input_from_tally(
+    email: str,
+    *,
+    require_api_key: bool = False,
+) -> Optional[BusinessUnderstandingInput]:
+    if not _settings.secrets.tally_api_key:
+        if require_api_key:
+            raise RuntimeError("Tally API key is not configured")
+        logger.debug("Tally: no API key configured, skipping")
+        return None
```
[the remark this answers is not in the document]
Swiftyos: Fixed in 7014a9b4b.

I restored the OpenRouter guard in get_business_understanding_input_from_tally() so we now skip cleanly when open_router_api_key is unset, and still fail fast with RuntimeError when require_api_key=True (the invited-user precompute path).

Added regression coverage in backend/data/tally_test.py for both cases and verified with poetry run format plus poetry run pytest backend/data/tally_test.py backend/data/invited_user_test.py.

autogpt_platform/backend/backend/api/features/admin/model.py
```diff
@@ -14,3 +23,70 @@ class UserHistoryResponse(BaseModel):
...
+class CreateInvitedUserRequest(BaseModel):
+    email: EmailStr
+    name: Optional[str] = None
+
+
+class InvitedUserResponse(BaseModel):
+    id: str
+    email: str
+    status: prisma.enums.InvitedUserStatus
+    auth_user_id: Optional[str] = None
```
Pwuts, line 37: Response models shouldn't have default values on any of the attributes; it makes the attribute appear optional (as in: may not be *present*) in the response schema which it isn't

autogpt_platform/backend/backend/data/invited_user.py
```diff
@@ -0,0 +1,779 @@
...
+
+class InvitedUserRecord(BaseModel):
+    id: str
+    email: str
+    status: prisma.enums.InvitedUserStatus
+    auth_user_id: Optional[str] = None
+    name: Optional[str] = None
+    tally_understanding: Optional[dict[str, Any]] = None
+    tally_status: prisma.enums.TallyComputationStatus
+    tally_computed_at: Optional[datetime] = None
```
Pwuts, line 60: what does this mean?

autogpt_platform/backend/backend/api/features/admin/model.py
```diff
@@ -14,3 +23,70 @@ class UserHistoryResponse(BaseModel):
...
+                    message=row.message,
+                    invited_user=(
+                        InvitedUserResponse.from_record(row.invited_user)
+                        if row.invited_user is not None
+                        else None
+                    ),
+                )
+                for row in result.results
+            ],
+        )
```
Pwuts, line 92: It looks like these are (mostly) copies of the models in `backend.data.invited_user`: can't we use those instead of defining copies here?

autogpt_platform/backend/backend/api/features/admin/user_admin_routes.py
```diff
@@ -0,0 +1,137 @@
...
+        admin_user_id,
+        file.filename or "<unnamed>",
+    )
+    content = await file.read()
+    result = await bulk_create_invited_users_from_file(file.filename, content)
+    return BulkInvitedUsersResponse.from_result(result)
+
+
+@router.post(
+    "/invited-users/{invited_user_id}/revoke",
```
Pwuts, line 101: nit:
```suggestion
@router.delete(
    "/invited-users/{invited_user_id}",
```

autogpt_platform/backend/backend/api/features/admin/user_admin_routes.py
```diff
@@ -0,0 +1,137 @@
...
+        invited_user.id,
+    )
+    return InvitedUserResponse.from_record(invited_user)
+
+
+@router.post(
+    "/invited-users/bulk",
+    response_model=BulkInvitedUsersResponse,
+    summary="Bulk Create Invited Users",
+    operation_id="postV2BulkCreateInvitedUsers",
```
Pwuts, line 84: Why define this explicitly, and why this ID? This is not a v2 route

autogpt_platform/backend/backend/api/features/admin/user_admin_routes.py
```diff
@@ -0,0 +1,137 @@
...
+@router.post(
+    "/invited-users",
+    response_model=InvitedUserResponse,
+    summary="Create Invited User",
+)
+async def create_invited_user_route(
+    request: CreateInvitedUserRequest,
+    admin_user_id: str = Security(get_user_id),
+) -> InvitedUserResponse:
+    logger.info(
```
Pwuts, line 66: There is a lot of logging in this module; do we need that? Can it be `logger.debug` instead?

autogpt_platform/backend/backend/api/features/admin/user_admin_routes.py
```diff
@@ -0,0 +1,137 @@
...
+        "Admin user %s created invited user %s",
+        admin_user_id,
+        invited_user.id,
+    )
+    return InvitedUserResponse.from_record(invited_user)
+
+
+@router.post(
+    "/invited-users/bulk",
+    response_model=BulkInvitedUsersResponse,
```
Pwuts, line 82: This is redundant with the route function's return type, and can cause type issues because `response_model` isn't checked statically. I advise never using `response_model` and always specifying the response type as the return type of the route function.

autogpt_platform/backend/backend/api/features/v1.py
```diff
@@ -127,6 +137,70 @@ def _create_file_size_error(size_bytes: int, max_size_mb: int) -> HTTPException:
...
+_CHECK_INVITE_RATE_LIMIT = 10  # requests
+_CHECK_INVITE_RATE_WINDOW = 60  # seconds
+
+
+@v1_router.post(
+    "/auth/check-invite",
+    summary="Check if an email is allowed to sign up",
+    tags=["auth"],
+)
+async def check_invite_route(
```
Pwuts, line 160: nit:
```suggestion
async def check_user_invite(
```

autogpt_platform/backend/backend/api/features/v1.py
```diff
@@ -179,10 +268,16 @@ async def update_user_email_route(
...
 )
 async def get_user_timezone_route(
-    user_data: dict = Security(get_jwt_payload),
+    user_id: Annotated[str, Security(get_user_id)],
 ) -> TimezoneResponse:
     """Get user timezone setting."""
-    user = await get_or_create_user(user_data)
+    try:
+        user = await get_user_by_id(user_id)
+    except ValueError:
```
Pwuts, line 276: `get_user_by_id` should return a `NotFoundError`, which will then automatically be handled by the FastAPI exception handler that is in place

autogpt_platform/backend/backend/api/features/v1.py
Pwuts: I guess this can work, but it's still clunky and I think there's a better option.

We essentially want to take control of who gets to sign up and who doesn't away from Supabase. We can do that by:
- disabling direct sign-up in Supabase
- adding a backend sign-up endpoint
  - checks invite status
  - calls `supabase.auth.admin.createUser` (or equivalent)

([source](https://www.reddit.com/r/Supabase/comments/1ijkwwr/comment/mbg19bk/))

We can then leave the DB trigger in place that creates the corresponding `User`, closing the "non-invited sign-up" and "orphaned user" gaps.

Or possibly simpler: disable sign-up through Supabase, and when we invite a user, create a `auth.users` and `platform.User` for them and send them a "activate your account" / "set your password" link.

autogpt_platform/backend/backend/api/rest_api.py
Pwuts: Doing this on every API request adds significant DB load and request overhead and should be avoided if at all possible

autogpt_platform/backend/backend/data/invited_user.py
```diff
@@ -0,0 +1,779 @@
...
+    name: Optional[str]
+
+
+def normalize_email(email: str) -> str:
+    return email.strip().lower()
+
+
+def is_internal_email(email: str) -> bool:
+    """Return True for @agpt.co addresses, which always bypass the invite gate."""
+    return normalize_email(email).endswith("@agpt.co")
```
Pwuts, line 116: probably shouldn't hardcode this

autogpt_platform/backend/backend/data/invited_user.py
```diff
@@ -0,0 +1,779 @@
...
+
+
+def _normalize_name(name: Optional[str]) -> Optional[str]:
```
Pwuts, line 119: Please define helpers *after* functions that call them (top-down ordering)

autogpt_platform/backend/backend/data/tally.py
```diff
@@ -40,8 +40,11 @@
 # LLM extraction timeout (seconds)
 _LLM_TIMEOUT = 30
 
+SUGGESTION_THEMES = ["Learn", "Create", "Automate", "Organize"]
```
Pwuts, line 43: Is it possible to change these later?

autogpt_platform/backend/backend/api/features/v1.py
```diff
@@ -135,13 +209,27 @@ def _create_file_size_error(size_bytes: int, max_size_mb: int) -> HTTPException:
...
-async def get_or_create_user_route(user_data: dict = Security(get_jwt_payload)):
-    user = await get_or_create_user(user_data)
+async def get_or_create_user_route(
+    request: Request,
+    user_data: dict = Security(get_jwt_payload),
+):
+    try:
+        user = await get_or_activate_user(user_data)
+    except NotAuthorizedError:
+        if request.headers.get(_CLEANUP_ORPHANED_AUTH_USER_HEADER) == "1":
```
Pwuts, line 219: Why client-controlled?

autogpt_platform/backend/backend/data/invited_user.py
```diff
@@ -0,0 +1,779 @@
...
+
+_WORKER_ID = f"{socket.gethostname()}:{os.getpid()}"
+
+_tally_seed_tasks: dict[str, asyncio.Task] = {}
+_TALLY_STALE_SECONDS = 300
+_MAX_TALLY_ERROR_LENGTH = 200
+_email_adapter = TypeAdapter(EmailStr)
+
+MAX_BULK_INVITE_FILE_BYTES = 1024 * 1024
+MAX_BULK_INVITE_ROWS = 500
```
Pwuts, line 49:
```suggestion
MAX_BULK_INVITE_ROWS = 500
MAX_BULK_INVITE_FILE_BYTES = MAX_BULK_INVITE_ROWS * 2048
```

autogpt_platform/backend/migrations/20260319120000_revert_invite_system/migration.sql
Pwuts, line 1: can't delete migrations after they are applied

autogpt_platform/backend/backend/data/invited_user.py
```diff
@@ -0,0 +1,779 @@
...
+
+class BulkInvitedUsersResult(BaseModel):
+    created_count: int
+    skipped_count: int
+    error_count: int
+    results: list[BulkInvitedUserRowResult]
+
+
+@dataclass(frozen=True)
+class _ParsedInviteRow:
```
Pwuts, line 104: why a dataclass and not a proper model?

autogpt_platform/backend/backend/data/understanding.py
```diff
@@ -31,6 +31,25 @@ def _json_to_list(value: Any) -> list[str]:
     return []
 
 
+def _json_to_themed_prompts(value: Any) -> dict[str, list[str]]:
```
Pwuts, line 34: same request as here https://github.com/Significant-Gravitas/AutoGPT/pull/12487/files#r2966376407

PR #12495, opened 2026-03-20
fix(frontend): prevent Select.Item crash on empty-string enum values

Otto-AGPT:
Requested by @ab20145

Radix UI's `<Select.Item>` throws `Error: A <Select.Item /> must have a value prop that is not an empty string` when any enum option has `value=""`. This crashes the `/build` page when a block's schema includes an empty-string enum option.

## Problem

Sentry issue: **BUILDER-5CS** (3 occurrences since Dec 2025)

`SelectWidget.tsx` (the JSON Schema form widget for enum/select fields) had two paths producing empty strings:

1. `value={value ?? ""}` — passes `""` as the Select value when the form value is null/undefined
2. `option.value` mapping — passes backend enum options through without filtering empty strings

## Fix

- Use `value || undefined` instead of `value ?? ""` — shows placeholder for unset values instead of crashing
- Filter out empty-string enum options before rendering `SelectItem`s

Single file, minimal change. No behavior change for valid (non-empty) options.

Resolves SECRT-2157

---
Co-authored-by: Abhimanyu Yadav (@Abhi1992002) <122007096+Abhi1992002@users.noreply.github.com>


majdyz (changes requested):
coderabbitai comemnt there seems correct

majdyz:
## Review: PR #12495 (prevent Select.Item crash on empty-string enum values)

Targeted fix for a Radix UI crash.

### Observations

1. **Root cause is correct** -- Radix `Select.Item` crashes when `value` is an empty string. Filtering them out is the right approach.

2. **`value || undefined`** -- changes from `value ?? ""` to `value || undefined`. This means falsy values (0, false) would also become undefined. For a Select widget, this is probably fine since enum values are typically strings, but worth noting.

3. **Filter removes empty-string options** -- `?.filter((option: any) => option.value !== "")` -- good, this prevents the crash while keeping all other options.

4. **Consider: should the backend prevent empty enum values?** -- The real fix might be ensuring the schema never contains `""` as an enum value. This frontend fix is a good band-aid, but a backend validation would prevent the issue for any future UI consumer.

**Verdict: LGTM** -- small targeted fix.

autogpt_platform/frontend/src/components/renderers/InputRenderer/base/standard/widgets/SelectInput/SelectWidget.tsx
```diff
@@ -61,13 +61,15 @@ export const SelectWidget = (props: WidgetProps) => {
         hideLabel={true}
         disabled={disabled || readonly}
         size={selectSize as any}
-        value={value ?? ""}
+        value={value || undefined}
```
majdyz: 🤖 🔴 **Blocker**: `value || undefined` collapses ALL falsy values — not just empty strings.

This is confirmed by tracing the code path from `OneOfField` (line 197): it passes `value={selectedIndex}` where `selectedIndex` is a numeric index (0, 1, 2, ...). When the **first** variant (index `0`) is selected, `0 || undefined` evaluates to `undefined`, making the dropdown appear empty/unselected even though variant 0 is legitimately selected.

The `Select` atom's `SelectFieldProps` types `value` as `string | undefined` (line 34 of `Select.tsx`), so the correct fix is an explicit nullish + empty-string check:

```tsx
value={value === "" || value == null ? undefined : value}
```

This preserves `0`, `false`, and any other non-empty truthy/falsy value while still converting `""` and `null`/`undefined` to `undefined` (which shows the placeholder).

> Note: this was already flagged by CodeRabbit and the prior review. The PR has not been updated to address it.

majdyz:
🤖 🟠 **Should Fix**: The `MultiSelector` path (lines 47-48 in `SelectWidget.tsx`) still passes `option.value` without filtering empty strings.

The filter added in this PR only applies to the single-select `<Select>` path. But the `MultiSelector` branch renders `<MultiSelectorItem key={option.value} value={option.value}>` without any empty-string filtering. If a multi-select enum schema includes an empty string value, it will render a `MultiSelectorItem` with `value=""`, which could cause similar issues depending on the underlying component behavior.

For consistency and defensive coding, consider applying the same `.filter((option: any) => option.value !== "")` to the `enumOptions` used in the `MultiSelector` branch as well (or, better yet, filter `enumOptions` once at the top of the component and reuse it in both paths).

majdyz:
🤖 🟡 **Nice to Have**: No test coverage for this fix.

This is a Sentry crash fix, so it would be valuable to have a test that verifies the `SelectWidget` does not crash when:
1. `value` is `null`/`undefined` (the original crash path)
2. `value` is `0` (the regression this PR currently introduces)
3. An enum option has `value: ""` (the other crash path)

A Storybook story or a small unit test exercising these edge cases would prevent future regressions. Given that this crash has been open since Dec 2025, a test to lock in the fix is worthwhile.

majdyz (changes requested):
## Review Summary

**PR description quality**: Good. The Why/What/How are well structured, the Sentry issue is referenced, and the root cause analysis is clear.

**The filter on empty-string enum options (lines 67-72)** is correct and directly addresses the Radix `<Select.Item>` crash. This part is good.

**However, there is one blocker that must be fixed before merge:**

### 🔴 Blocker: `value || undefined` breaks `OneOfField` variant selection

The change from `value ?? ""` to `value || undefined` trades one bug for another. The `||` operator coerces all falsy values — including numeric `0` — to `undefined`. `OneOfField` passes `value={selectedIndex}` where `selectedIndex` is a 0-based index. When the first variant (index 0) is selected, `0 || undefined` evaluates to `undefined`, making the dropdown appear empty.

**Fix**: Use an explicit check: `value === "" || value == null ? undefined : value`

### 🟠 Should Fix: MultiSelector path not guarded

The empty-string filter only applies to the single-select path. The `MultiSelector` branch (lines 47-48) still passes `option.value` unfiltered. Consider filtering `enumOptions` once at the top of the component.

### 🟡 Nice to Have: No test coverage

A Storybook story or unit test covering `value=null`, `value=0`, and `option.value=""` would lock in this fix and prevent regressions.

---

Please address the blocker, then request re-review. Once fixed, please test with a block that uses a `oneOf`/discriminated union (so the first variant uses index 0) and share a screenshot proving the dropdown correctly shows "Option 1" selected.

majdyz (dismissed):
## Follow-up review

The PR has not been updated since my previous review on Mar 20 and the follow-up comments on Mar 31. The three issues flagged earlier remain unaddressed:

### 1. Blocker: `value || undefined` breaks falsy selections (unchanged)

This was flagged by CodeRabbit, confirmed by me in the first review, and re-emphasized with a detailed inline comment on Mar 31. The PR still uses `value || undefined` on line 64, which coerces `0` (and `false`) to `undefined`. `OneOfField` passes `value={selectedIndex}` where `selectedIndex` is a 0-based numeric index -- when the first variant is selected, `0 || undefined` evaluates to `undefined` and the dropdown appears empty.

**Required fix** (same as previously requested):
```tsx
value={value === "" || value == null ? undefined : value}
```

### 2. Should Fix: MultiSelector path still unguarded (unchanged)

The empty-string filter only applies to the single-select `<Select>` path (lines 67-72 in the diff). The `MultiSelector` branch (lines 47-48) still renders `<MultiSelectorItem value={option.value}>` without any filtering, so if a multi-select enum schema includes an empty-string value, the same class of crash could occur there. Filter `enumOptions` once at the top of the component and reuse it in both paths.

### 3. Nice to have: no test coverage (unchanged)

A unit test or Storybook story covering `value=null`, `value=0`, and `option.value=""` would lock in this fix.

---

Please address items 1 and 2, then request re-review. Once fixed, verify with a block that uses a `oneOf`/discriminated union (so the first variant uses index 0) and confirm the dropdown correctly shows the first option selected.

majdyz:

Thanks for the contribution!

autogpt_platform/frontend/src/components/renderers/InputRenderer/base/standard/widgets/SelectInput/SelectWidget.tsx
majdyz: 🤖 Fixed in a79c265925: All three issues addressed — (1) falsy value preservation, (2) empty-string filter applied once at top level for both code paths, (3) removed redundant filter in Select options prop.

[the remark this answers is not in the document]
majdyz: 🤖 Fixed in a79c265925: Replaced `value || undefined` with `value === "" || value == null ? undefined : value` to preserve falsy values like `0`. Also moved the empty-string filter to the top of the component so both single-select and multi-select paths are protected.

ntindle:
🤖 Fixed in [bb7e77c99](https://github.com/Significant-Gravitas/AutoGPT/commit/bb7e77c99ad118dec300a75a41936518c1c6b626): added focused regression coverage for 0SelectWidget0 covering the empty-string value path, preserving falsy values like 00, and filtering empty-string enum options in the multi-select path.

Also added a dedicated Storybook repro (0EnumWithEmptyStringValue0) and wired Storybook to generate the API client before startup so this PR can be exercised locally without the missing-generated-files failure.

I wasn't able to attach a screenshot/recording from this environment, but the new Storybook case is available under 0Renderers / FormRenderer / Primitive Fields / EnumWithEmptyStringValue0 for a quick manual check.

ntindle:
🤖 Fixed in [9b4b4e354](https://github.com/Significant-Gravitas/AutoGPT/commit/9b4b4e354bfc20bf38e5035708cf2fc6b7c871b9): removed the dead single-select fallback branch and added the missing undefined-enumOptions regression case so the frontend Codecov patch check can cover both changed paths.

autogpt_platform/frontend/package.json
```diff
@@ -21,7 +21,9 @@
     "test:e2e:no-build": "playwright test",
     "test:e2e:ui": "playwright test --ui",
     "gentests": "playwright codegen http://localhost:3000",
+    "prestorybook": "pnpm run generate:api",
     "storybook": "storybook dev -p 6006",
+    "prebuild-storybook": "pnpm run generate:api",
```
majdyz: 🤖 🟠 **Should Fix**: Adding `prestorybook` and `prebuild-storybook` hooks that run `pnpm run generate:api` is out of scope for a Select-crash fix.

**Failure scenario:**
1. Anyone running `pnpm storybook` for an unrelated component will trigger `generate:api` (orval + a tsx script that touches `src/app/api/__generated__/`).
2. If the OpenAPI spec drifts from `main`, the regen produces unrelated diffs in the working tree mid-dev.
3. CI / Chromatic runs of storybook now pay the regen cost on every build, slowing the loop and potentially failing if `generate:api` errors transiently.

The Storybook story you added (`EnumWithEmptyStringValue`) does not import any generated API client — it uses an inline JSON schema. So the dependency isn't actually required by this PR.

**Direct question:** Can the two `prestorybook` / `prebuild-storybook` script entries be split into a separate PR (or dropped) so this crash fix stays single-file scope?

autogpt_platform/frontend/src/components/renderers/InputRenderer/base/standard/widgets/SelectInput/SelectWidget.tsx
```diff
@@ -24,7 +24,9 @@ export const SelectWidget = (props: WidgetProps) => {
     id,
     formContext,
   } = props;
-  const enumOptions = options.enumOptions || [];
+  const enumOptions = (options.enumOptions ?? []).filter(
+    (option: any) => option.value !== "",
```
majdyz: 🤖 🟡 **Nice to Have**: The empty-string filter at line 26-28 silently drops enum options. If a backend block schema ever legitimately ships `enum: ["", "foo"]` (e.g. an "any/unset" sentinel), users will see the `"foo"` option in the dropdown but no way to select `""`, with no diagnostic.

**Failure scenario:**
1. A block author publishes a schema with an `enum: ["", ...]` field intending `""` to mean "no filter".
2. The frontend silently drops the `""` option to avoid the Radix crash.
3. The user picks another option, runs the block, gets unexpected behavior, and there is no console trace pointing at the SelectWidget.

**Direct question:** Want me to add a `console.warn` (dev-only, gated on `process.env.NODE_ENV !== "production"`) when an empty-string option is filtered out, so future schema authors get a hint instead of a silent UX drop?

autogpt_platform/frontend/src/components/renderers/InputRenderer/base/standard/widgets/SelectInput/__tests__/SelectWidget.test.tsx
```diff
@@ -0,0 +1,288 @@
+import type { WidgetProps } from "@rjsf/utils";
+import type { ReactNode } from "react";
+import { afterEach, describe, expect, it, vi } from "vitest";
+import { cleanup, render, screen } from "@/tests/integrations/test-utils";
```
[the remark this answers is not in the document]
ntindle: 🤖 Fixed in [ed7166fe7](https://github.com/Significant-Gravitas/AutoGPT/commit/ed7166fe714b036238072eaf6574decaa7905fed): removed redundant local RTL cleanup, converted the new mock/story render functions to declarations, and asserted that the single-select path receives only the filtered Red option.

ntindle:
🤖 Addressed the remaining CodeRabbit review feedback in [ed7166fe7](https://github.com/Significant-Gravitas/AutoGPT/commit/ed7166fe714b036238072eaf6574decaa7905fed): removed duplicate cleanup, switched the new mock/story render components to function declarations, and strengthened the filtered-options assertion.

PR #12510, opened 2026-03-23
feat(backend/db): add UserInvite model with allowed_users trigger

Swiftyos:
## Summary
- Add `UserInvite` model and `TallyComputationStatus` enum to track invited users and their onboarding tally computation state
- Create `allowed_users` table with a Postgres trigger (`trg_create_invite`) that automatically inserts a `UserInvite` record whenever a new email is added to `allowed_users`

## Changes
- **`schema.prisma`**: Add `TallyComputationStatus` enum (`PENDING`, `RUNNING`, `READY`, `FAILED`) and `UserInvite` model keyed by email, linked to `User` via optional `userId`
- **Migration SQL**: Create `UserInvite` table, `allowed_users` table, and `create_invite()` trigger function

## Test plan
- [ ] Run `poetry run prisma migrate dev` and verify migration applies cleanly
- [ ] Insert a row into `allowed_users` and confirm a corresponding `UserInvite` row is created in the `platform` schema
- [ ] Verify the foreign key constraint works: linking a `UserInvite` to an existing `User` succeeds, and cascading delete removes the invite when the user is deleted

### DIFF
--- a/autogpt_platform/backend/migrations/20260323084228_user_invites/migration.sql
+++ b/autogpt_platform/backend/migrations/20260323084228_user_invites/migration.sql
@@ -5,7 +5,7 @@ CREATE TYPE "TallyComputationStatus" AS ENUM ('PENDING', 'RUNNING', 'READY', 'FA
 CREATE TABLE "UserInvite" (
     "email" TEXT NOT NULL,
     "createdAt" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
-    "updatedAt" TIMESTAMP(3) NOT NULL,
+    "updatedAt" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
     "userId" TEXT,
     "tallyUnderstanding" JSONB,
     "tallyStatus" "TallyComputationStatus" NOT NULL DEFAULT 'PENDING',
--- a/autogpt_platform/backend/schema.prisma
+++ b/autogpt_platform/backend/schema.prisma
@@ -84,7 +84,7 @@ enum TallyComputationStatus {
 model UserInvite {
   email              String                 @id
   createdAt          DateTime               @default(now())
-  updatedAt          DateTime               @updatedAt
+  updatedAt          DateTime               @default(now()) @updatedAt
   userId             String?                @unique
   User               User?                  @relation(fields: [userId], references: [id], onDelete: Cascade)
   tallyUnderstanding Json?


autogpt_platform/backend/migrations/20260323084228_user_invites/migration.sql
```diff
@@ -0,0 +1,45 @@
...
+);
+
+-- AddForeignKey
+ALTER TABLE "UserInvite" ADD CONSTRAINT "UserInvite_userId_fkey" FOREIGN KEY ("userId") REFERENCES "User"("id") ON DELETE CASCADE ON UPDATE CASCADE;
+
+CREATE OR REPLACE FUNCTION public.create_invite()
+RETURNS trigger as $$
+BEGIN
+    INSERT INTO platform."UserInvite" (email)
+    VALUES (NEW.email);
```
majdyz, line 36: 🤖 🟠 **Should Fix**: The trigger will blow up if the same email is re-inserted into `allowed_users` (e.g. after a delete+re-add). The `INSERT INTO platform."UserInvite"` should use `ON CONFLICT (email) DO NOTHING` (or `DO UPDATE SET "updatedAt" = CURRENT_TIMESTAMP` if you want to refresh it).

Without this, any operational retry or re-invite of the same email will cause the trigger to throw a unique constraint violation, which will also roll back the `allowed_users` INSERT.

autogpt_platform/backend/migrations/20260323084228_user_invites/migration.sql
```diff
@@ -0,0 +1,45 @@
...
+-- Create the allowed_users table if it does not exist
+CREATE TABLE IF NOT EXISTS public.allowed_users (
+    email TEXT PRIMARY KEY
+);
+
+-- AddForeignKey
+ALTER TABLE "UserInvite" ADD CONSTRAINT "UserInvite_userId_fkey" FOREIGN KEY ("userId") REFERENCES "User"("id") ON DELETE CASCADE ON UPDATE CASCADE;
+
+CREATE OR REPLACE FUNCTION public.create_invite()
+RETURNS trigger as $$
```
majdyz, line 33: 🤖 🟠 **Should Fix**: The trigger hardcodes `platform."UserInvite"` but other migrations in this repo use `current_schema()` with dynamic `SET search_path` (see `20260319120000_revert_invite_system` for the pattern). If the configured schema ever differs from `platform`, this trigger silently breaks.

Suggest using the same `DO $$ ... EXECUTE format(..., current_schema()) ... $$` pattern used by the existing `add_user_and_profile_to_platform()` trigger, or at minimum use `SET search_path` in the function definition:

```sql
CREATE OR REPLACE FUNCTION public.create_invite()
RETURNS trigger
LANGUAGE plpgsql
SET search_path = platform
AS $$ ...
```

autogpt_platform/backend/migrations/20260323084228_user_invites/migration.sql
```diff
@@ -0,0 +1,45 @@
+-- CreateEnum
+CREATE TYPE "TallyComputationStatus" AS ENUM ('PENDING', 'RUNNING', 'READY', 'FAILED');
+
+-- CreateTable
+CREATE TABLE "UserInvite" (
+    "email" TEXT NOT NULL,
+    "createdAt" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    "updatedAt" TIMESTAMP(3) NOT NULL,
```
majdyz: 🤖 🔴 **Blocker**: +1 to Sentry and CodeRabbit — this trigger is broken as-is. The `updatedAt` column is `NOT NULL` with no `DEFAULT`, and the INSERT only provides `email`. Every insert into `allowed_users` will fail with a NOT NULL constraint violation.

Fix both the trigger (add `"updatedAt"` column to the INSERT with `CURRENT_TIMESTAMP`) **and** add a `DEFAULT CURRENT_TIMESTAMP` to the column definition as a safety net for any future raw SQL inserts.

autogpt_platform/backend/migrations/20260323084228_user_invites/migration.sql
```diff
@@ -0,0 +1,45 @@
...
+RETURNS trigger as $$
+BEGIN
+    INSERT INTO platform."UserInvite" (email)
+    VALUES (NEW.email);
+
+    RETURN NEW;
+END;
+$$ LANGUAGE plpgsql;
+
+CREATE TRIGGER trg_create_invite
```
majdyz, line 42: 🤖 🟡 **Nice to Have**: `CREATE TRIGGER` (line 42) is not idempotent — unlike `CREATE OR REPLACE FUNCTION`, there is no `CREATE OR REPLACE TRIGGER` in PostgreSQL. If this migration is ever re-applied or the trigger already exists, it will error. Add `DROP TRIGGER IF EXISTS trg_create_invite ON public.allowed_users;` before the `CREATE TRIGGER`.

autogpt_platform/backend/migrations/20260323084228_user_invites/migration.sql
```diff
@@ -0,0 +1,45 @@
...
+);
+
+-- CreateIndex
+CREATE UNIQUE INDEX "UserInvite_userId_key" ON "UserInvite"("userId");
+
+-- CreateIndex
+CREATE INDEX "UserInvite_createdAt_idx" ON "UserInvite"("createdAt");
+
+-- Create the allowed_users table if it does not exist
+CREATE TABLE IF NOT EXISTS public.allowed_users (
```
majdyz, line 25: why it's public ?

autogpt_platform/backend/schema.prisma
```diff
@@ -73,6 +74,27 @@ model User {
...
+model UserInvite {
+  email              String                 @id
+  createdAt          DateTime               @default(now())
+  updatedAt          DateTime               @updatedAt
+  userId             String?                @unique
+  User               User?                  @relation(fields: [userId], references: [id], onDelete: Cascade)
+  tallyUnderstanding Json?
+  tallyStatus        TallyComputationStatus @default(PENDING)
+  tallyComputedAt    DateTime?
+  tallyError         String?
```
majdyz, line 93: these tally specific columns can those just be a single metadata Json column where these info grouped into one? keep adding talling specific info is overkill, especially we already have Json column there everything can just be moved there

autogpt_platform/backend/migrations/20260323084228_user_invites/migration.sql
```diff
@@ -0,0 +1,45 @@
...
+    email TEXT PRIMARY KEY
+);
+
+-- AddForeignKey
+ALTER TABLE "UserInvite" ADD CONSTRAINT "UserInvite_userId_fkey" FOREIGN KEY ("userId") REFERENCES "User"("id") ON DELETE CASCADE ON UPDATE CASCADE;
+
+CREATE OR REPLACE FUNCTION public.create_invite()
```
majdyz, line 32: why it's public? public function is already deprecated by supabase

autogpt_platform/backend/schema.prisma
```diff
@@ -73,6 +74,27 @@ model User {
   OAuthRefreshTokens      OAuthRefreshToken[]
 }
 
+enum TallyComputationStatus {
```
majdyz, line 77: This makes the system 3p dependent, can we rename it to be non-Tally?

PR #12519, opened 2026-03-23
fix(frontend): auto-reconnect copilot chat after device sleep/wake

0ubbe:
## Summary

- Adds `visibilitychange`-based sleep/wake detection to the copilot chat — when the page becomes visible after >30s hidden, automatically refetch the session and either resume an active stream or hydrate completed messages
- Blocks chat input during re-sync (`isSyncing` state) to prevent users from accidentally sending a message that overwrites the agent's completed work
- Replaces `PulseLoader` with a spinning `CircleNotch` icon on sidebar session names for background streaming sessions (closer to ChatGPT's UX)

## How it works

1. When the page goes hidden, we record a timestamp
2. When the page becomes visible, we check elapsed time
3. If >30s elapsed (indicating sleep or long background), we refetch the session from the API
4. If backend still has `active_stream=true` → remove stale assistant message and resume SSE
5. If backend is done → the refetch triggers React Query invalidation which hydrates the completed messages
6. Chat input stays disabled (`isSyncing=true`) until re-sync completes

## Test plan

- [ ] Open copilot, start a long-running agent task
- [ ] Close laptop lid / lock screen for >30 seconds
- [ ] Wake device — verify chat shows the agent's completed response (or resumes streaming)
- [ ] Verify chat input is temporarily disabled during re-sync, then re-enables
- [ ] Verify sidebar shows spinning icon (not pulse loader) for background sessions
- [ ] Verify no duplicate messages appear after wake
- [ ] Verify normal streaming (no sleep) still works as expected

Resolves: [SECRT-2159](https://linear.app/autogpt/issue/SECRT-2159)


autogpt_platform/frontend/src/app/(platform)/copilot/useCopilotStream.ts
```diff
@@ -298,6 +305,69 @@ export function useCopilotStream({
...
+
+    const elapsed = Date.now() - lastVisibleAtRef.current;
+    lastVisibleAtRef.current = Date.now();
+
+    if (document.visibilityState !== "visible") return;
+    if (elapsed < WAKE_RESYNC_THRESHOLD_MS) return;
+
+    setIsSyncing(true);
+    try {
+      const result = await refetchSession();
```
majdyz: 🤖 🔴 **Blocker**: `isSyncing` is not reset on session switch. The cleanup effect at line ~316 resets all reconnect state but does not call `setIsSyncing(false)`. If the user switches sessions while a wake re-sync is in flight, `isSyncing` stays `true` forever, permanently blocking chat input.

```diff
  useEffect(() => {
    clearTimeout(reconnectTimerRef.current);
    // ...existing cleanup...
    setReconnectExhausted(false);
+   setIsSyncing(false);
    hasResumedRef.current.clear();
    // ...
  }, [sessionId]);
```

autogpt_platform/frontend/src/app/(platform)/copilot/useCopilotStream.ts
```diff
@@ -298,6 +305,69 @@ export function useCopilotStream({
...
+      const d = result.data;
+      const backendActive =
+        d != null &&
+        typeof d === "object" &&
+        "status" in d &&
```
majdyz: 🤖 🟠 **Should Fix**: The `backendActive` type-narrowing chain is duplicated verbatim between `handleWakeResync` and `onFinish`. Extract it into a helper (e.g. `hasActiveBackendStream(result)`) in `helpers.ts` to keep it DRY and ensure both paths stay in sync if the response shape changes.

```ts
function hasActiveBackendStream(result: { data?: unknown }): boolean {
  const d = result.data;
  return (
    d != null &&
    typeof d === "object" &&
    "status" in d &&
    d.status === 200 &&
    "data" in d &&
    d.data != null &&
    typeof d.data === "object" &&
    "active_stream" in d.data &&
    !!d.data.active_stream
  );
}
```

autogpt_platform/frontend/src/app/(platform)/copilot/useCopilotStream.ts
```diff
@@ -298,6 +305,69 @@ export function useCopilotStream({
     }
   }
 
+  // ---------------------------------------------------------------------------
```
majdyz, line 304: 🤖 🟠 **Should Fix**: `handleWakeResync` is async but has no staleness guard. If `sessionId` changes while the `refetchSession()` await is in flight, the callback continues with the old session's result and may call `setMessages`/`resumeStream` for a session that is no longer active, potentially corrupting messages or starting a stream on the wrong session.

Capture `sessionId` at call time and bail out after the await if it changed:

```ts
const handleWakeResync = useCallback(async () => {
  const sid = sessionId;
  if (!sid) return;
  // ...
  const result = await refetchSession();
  if (sessionId !== sid) return; // session switched while awaiting
  // ...rest of logic
}, [sessionId, refetchSession, setMessages, resumeStream]);
```

Note: since `sessionId` is a closure value, you'd need a ref to get the current value. Alternatively, use an abort signal or a `mountedRef` pattern.

majdyz, line 304: 🤖 🟡 **Nice to Have**: `useCallback` is used here, but the frontend CLAUDE.md says "Do not use `useCallback` or `useMemo` unless asked to optimise". Since this callback is only needed as a stable ref for the `useEffect` dependency below, consider inlining the logic directly in the `onVisibilityChange` handler inside the `useEffect`, which avoids both the `useCallback` and the dependency array maintenance.

autogpt_platform/frontend/src/app/(platform)/copilot/useCopilotStream.ts
0ubbe: 🤖 Fixed in 94cfb7ff4: Added `setIsSyncing(false)` to the session-switch cleanup effect.

autogpt_platform/frontend/src/app/(platform)/copilot/useCopilotStream.ts
0ubbe: 🤖 Fixed in 94cfb7ff4: Extracted `hasActiveBackendStream()` helper in `helpers.ts` and replaced both occurrences.

autogpt_platform/frontend/src/app/(platform)/copilot/useCopilotStream.ts
0ubbe: 🤖 Fixed in 94cfb7ff4: Added a `sessionIdRef` and a staleness check after the await — bails out if `sessionIdRef.current !== sid`.

0ubbe: 🤖 Fixed in 94cfb7ff4: Inlined the logic directly into the `useEffect` and removed the `useCallback`. Also removed the `useCallback` import.

0ubbe:
🤖 Addressed coderabbitai nitpicks in 03d2449c0:

**Fixed:**
- **CircleNotch size inconsistency** (MobileDrawer): aligned to `h-4 w-4` to match `CheckCircle`
- **`lastVisibleAtRef` naming**: renamed to `lastHiddenAtRef` since it's set when the page becomes hidden

**Intentionally skipped:**
- **`interface` → `type` for Props** (CollapsedCustomToolGroup): CONTRIBUTING.md says "Prefer `interface` for object shapes" and "Component props should be `interface Props { ... }`"
- **Expand `TOOL_LABELS`**: the fallback generates readable labels; explicit entries would need maintenance for diminishing returns
- **File length refactor** (useCopilotStream ~489 lines): valid observation, but splitting into 3-4 hooks is a separate refactor — will track as follow-up

autogpt_platform/frontend/src/app/(platform)/copilot/useCopilotStream.ts
```diff
@@ -298,6 +297,65 @@ export function useCopilotStream({
...
+        if (hasActiveBackendStream(result)) {
+          // Stream is still running — resume SSE to pick up live chunks.
+          // Remove stale in-progress assistant message first (backend replays
+          // from "0-0").
+          setMessages((prev) => {
+            if (prev.length > 0 && prev[prev.length - 1].role === "assistant") {
+              return prev.slice(0, -1);
+            }
+            return prev;
+          });
```
majdyz, line 335: 🤖 🟡 **Nice to Have**: `handleWakeResync` has `try/finally` but no `catch`. If `refetchSession()` or `resumeStream()` rejects, the error is silently swallowed — no toast, no log. Consider adding a minimal `catch` to at least `console.warn` so wake re-sync failures aren't invisible during debugging.

```ts
} catch (err) {
  console.warn('[copilot] wake re-sync failed', err);
} finally {
  setIsSyncing(false);
}
```

autogpt_platform/frontend/src/app/(platform)/copilot/components/ChatSidebar/ChatSidebar.tsx
```diff
@@ -36,7 +37,6 @@ import { useEffect, useRef, useState } from "react";
 import { useCopilotUIStore } from "../../store";
```
majdyz, line 37: 🤖 🔵 **Nit**: `PulseLoader` is no longer imported anywhere after this PR, but [PulseLoader.tsx](../blob/dev/autogpt_platform/frontend/src/app/(platform)/copilot/components/PulseLoader/PulseLoader.tsx) still exists. Consider deleting the file to avoid dead code.

autogpt_platform/frontend/src/app/(platform)/copilot/useCopilotStream.ts
```diff
@@ -298,6 +297,67 @@ export function useCopilotStream({
...
+  useEffect(() => {
+    async function handleWakeResync() {
+      const sid = sessionIdRef.current;
+      if (!sid) return;
+
+      const elapsed = Date.now() - lastHiddenAtRef.current;
+      lastHiddenAtRef.current = Date.now();
+
+      if (document.visibilityState !== "visible") return;
+      if (elapsed < WAKE_RESYNC_THRESHOLD_MS) return;
```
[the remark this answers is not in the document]
0ubbe: Not a real issue — the reset on line 315 is redundant but harmless.

The `onVisibilityChange` handler already stamps `lastHiddenAtRef.current = Date.now()` on every `hidden` event (line 349). That means by the time `handleWakeResync` runs (only on `visible` transitions), the ref already holds the correct hide-time. The reset on line 315 just overwrites it with "now" *after* `elapsed` has already been computed, and the next `hidden` event will overwrite it again before the next wake cycle.

The scenario described (quick tab switch poisoning a later wake) can't happen:
1. Hide → line 349 sets ref to T₁
2. Show quickly (<30s) → `handleWakeResync` reads correct elapsed, bails, resets ref to T₂
3. Hide again → line 349 **overwrites** ref with T₃
4. Wake after long sleep → `handleWakeResync` correctly reads elapsed from T₃

The line 315 write is dead in practice since step 3 always intervenes. No fix needed.

PR #12520, opened 2026-03-23
fix(backend/db): add memory-aware readiness probe for database-manager

majdyz:
## Summary
- Add `/ready` endpoint to database-manager with memory-aware readiness check
- When process memory (RSS) exceeds `READINESS_MEMORY_LIMIT` (default 1.5GB), the readiness probe fails
- Kubernetes removes the pod from Service endpoints (no new traffic), while in-flight requests continue to drain
- The liveness probe (`/health_check`) is unchanged — it only checks DB connectivity
- Base `AppService` gets a default `readiness_check()` that delegates to `health_check()`, so other services are unaffected

## Why
The Prisma binary query engine (a separate Rust subprocess) has an unbounded query plan cache that grows with query diversity. In dev, pods balloon from ~680Mi to 2-3GB over ~1hr, causing HPA runaway (14 pods vs prod's stable 4).

This provides a graceful recycling mechanism:
1. Pod memory exceeds 1.5GB → readiness fails → removed from endpoints
2. In-flight requests drain on the "not ready" pod
3. HPA scales up a fresh pod to absorb traffic
4. Liveness probe eventually fails (or pod hits memory limit) → restart

## Infra companion PR
Requires [AutoGPT_cloud_infrastructure#299](https://github.com/Significant-Gravitas/AutoGPT_cloud_infrastructure/pull/299) which configures the readiness probe to use `/ready` and sets `READINESS_MEMORY_LIMIT`.

## Test plan
- [ ] Verify `/ready` returns OK when memory is below threshold
- [ ] Verify `/ready` returns 500 when memory exceeds threshold
- [ ] Verify `/health_check` is unaffected by memory usage
- [ ] Verify other services (executor, scheduler) still work (they use default `readiness_check`)
- [ ] Deploy to dev and monitor pod recycling behavior

### DIFF
--- a/autogpt_platform/backend/backend/util/service.py
+++ b/autogpt_platform/backend/backend/util/service.py
@@ -415,7 +415,9 @@ class AppService(BaseAppService, ABC):
         self.fastapi_app.add_api_route(
             "/health_check_async", self.health_check, methods=["POST", "GET"]
         )
-        self.fastapi_app.add_api_route("/ready", self.readiness_check, methods=["GET"])
+        self.fastapi_app.add_api_route(
+            "/ready", self.readiness_check, methods=["POST", "GET"]
+        )
         self.fastapi_app.add_exception_handler(
             ValueError, self._handle_internal_http_error(400)
         )


majdyz:
Reverting — going with a simpler approach: lower memory limits + terminationGracePeriodSeconds instead of custom readiness endpoint.

PR #12541, opened 2026-03-24
closed

majdyz:
closed

### DIFF
--- a/autogpt_platform/backend/backend/api/features/integrations/router_test.py
+++ b/autogpt_platform/backend/backend/api/features/integrations/router_test.py
@@ -73,7 +73,7 @@ class TestGetCredentialReturnsMetaOnly:
         assert data["provider"] == "openai"
         assert data["type"] == "api_key"
         assert "api_key" not in data
-        assert "secret" not in str(data).lower() or "sk-" not in str(data)
+        assert "sk-secret-key-value" not in str(data)
 
     def test_oauth2_credential_no_secret(self):
         cred = _make_oauth2_cred()
@@ -97,14 +97,14 @@ class TestSdkDefaultCredentialsNotAccessible:
     """SDK default credentials (ID ending in '-default') must be hidden."""
 
     def test_get_sdk_default_returns_404(self):
-        cred = _make_sdk_default_cred("openai")
         with patch(
             "backend.api.features.integrations.router.creds_manager"
         ) as mock_mgr:
-            mock_mgr.get = AsyncMock(return_value=cred)
+            mock_mgr.get = AsyncMock()
             resp = client.get("/openai/credentials/openai-default")
 
         assert resp.status_code == 404
+        mock_mgr.get.assert_not_called()
 
     def test_list_credentials_excludes_sdk_defaults(self):
         user_cred = _make_api_key_cred()
--- a/autogpt_platform/backend/backend/sdk/registry.py
+++ b/autogpt_platform/backend/backend/sdk/registry.py
@@ -202,6 +202,8 @@ class AutoRegistry:
 
         # Patch credentials store to include SDK-registered credentials
         try:
+            # Lazy import: credentials_store depends on settings which may not
+            # be fully initialized at class-definition time.
             from backend.integrations.credentials_store import (
                 IntegrationCredentialsStore,
             )


autogpt_platform/backend/backend/api/features/integrations/router.py
```diff
@@ -267,18 +269,32 @@ async def get_credential(
...
+) -> CredentialsMetaResponse:
     credential = await creds_manager.get(user_id, cred_id)
     if not credential:
         raise HTTPException(
             status_code=status.HTTP_404_NOT_FOUND, detail="Credentials not found"
         )
-    if credential.provider != provider:
+    if credential.id.endswith("-default"):
+        raise HTTPException(
+            status_code=status.HTTP_404_NOT_FOUND, detail="Credentials not found"
```
majdyz, line 281: 🤖 🟠 **Should Fix**: The `CredentialsMetaResponse(...)` construction is now duplicated 4 times (list_credentials, list_credentials_by_provider, get_credential, create_credentials). Extract a helper like `_to_meta_response(cred: Credentials) -> CredentialsMetaResponse` to DRY this up and reduce the chance of divergence when fields are added.

autogpt_platform/backend/backend/api/features/integrations/router.py
```diff
@@ -267,18 +269,32 @@ async def get_credential(
...
+        )
+    if not provider_matches(credential.provider, provider):
         raise HTTPException(
```
majdyz, line 284: 🤖 🟠 **Should Fix**: This error message leaks info — it confirms to an attacker that a credential with this ID exists but belongs to a different provider. The old message `"Credentials do not match the specified provider"` was better. For a security hotfix, prefer the generic `"Credentials not found"` to avoid enumeration.

autogpt_platform/backend/backend/api/features/integrations/router.py
```diff
@@ -267,18 +269,32 @@ async def get_credential(
     ],
     cred_id: Annotated[str, Path(title="The ID of the credentials to retrieve")],
     user_id: Annotated[str, Security(get_user_id)],
-) -> Credentials:
+) -> CredentialsMetaResponse:
     credential = await creds_manager.get(user_id, cred_id)
     if not credential:
         raise HTTPException(
```
majdyz, line 275: 🤖 🟡 **Nice to Have**: The `-default` check should be done before `creds_manager.get()` — no need to fetch (and potentially lock) the credential if we know upfront it will be rejected. Move this guard to the top of the function:
```python
if cred_id.endswith("-default"):
    raise HTTPException(status_code=404, detail="Credentials not found")
```

autogpt_platform/backend/backend/api/features/integrations/router.py
```diff
@@ -232,6 +232,7 @@ async def list_credentials(
             host=CredentialsMetaResponse.get_host(cred),
         )
         for cred in credentials
+        if not cred.id.endswith("-default")
```
majdyz: 🤖 🟡 **Nice to Have**: The `endswith("-default")` heuristic is fragile — a user could theoretically create a credential with an ID ending in `-default` and it would be silently filtered. Consider using a constant like `SDK_DEFAULT_SUFFIX = "-default"` and adding a comment explaining why this pattern is used, or better yet, add a `is_sdk_default: bool` field to the credential model that the SDK sets explicitly.

autogpt_platform/backend/backend/api/features/integrations/router.py
majdyz: 🤖 Fixed in f01c7c3ea: Extracted `_to_meta_response()` helper, all 4 call sites now use it.

autogpt_platform/backend/backend/api/features/integrations/router.py
majdyz: 🤖 Fixed in f01c7c3ea: Changed to generic `"Credentials not found"` to prevent enumeration.

autogpt_platform/backend/backend/api/features/integrations/router.py
majdyz: 🤖 Fixed in f01c7c3ea: Moved `_is_sdk_default(cred_id)` check before `creds_manager.get()`.

autogpt_platform/backend/backend/api/features/integrations/router.py
majdyz: 🤖 Fixed in f01c7c3ea: Added `_SDK_DEFAULT_SUFFIX` constant with explanatory comment. Also cleaned up the duck typing in registry.py (removed hasattr/sys.modules pattern, direct import instead).

autogpt_platform/backend/backend/api/features/integrations/router_test.py
```diff
@@ -0,0 +1,165 @@
...
+
+
+class TestSdkDefaultCredentialsNotAccessible:
+    """SDK default credentials (ID ending in '-default') must be hidden."""
+
+    def test_get_sdk_default_returns_404(self):
+        cred = _make_sdk_default_cred("openai")
+        with patch(
+            "backend.api.features.integrations.router.creds_manager"
+        ) as mock_mgr:
```
majdyz, line 102: 🤖 🟡 **Nice to Have**: Now that the `_is_sdk_default` check runs before `creds_manager.get()`, this test sets up a mock for `creds_manager.get` that will never be called. The mock is harmless but misleading — consider removing it and adding an assertion that `creds_manager.get` was NOT called, which validates the early-return behavior.

autogpt_platform/backend/backend/sdk/registry.py
```diff
@@ -202,40 +202,29 @@ def patched_load():
 
         # Patch credentials store to include SDK-registered credentials
         try:
-            import sys
-            from typing import Any
+            from backend.integrations.credentials_store import (
```
majdyz, line 207: 🤖 🔵 **Nit**: This local import is kept because `credentials_store` may not be importable at class definition time (circular import risk at module load). The CLAUDE.md rule is "top-level imports only (lazy imports only for heavy optional deps)" — this is a borderline case since it is inside a `try/except`. A brief comment like `# lazy import to avoid circular dependency at class definition time` would clarify intent.

autogpt_platform/backend/backend/api/features/integrations/router_test.py
```diff
@@ -0,0 +1,165 @@
...
+        assert data["provider"] == "openai"
+        assert data["type"] == "api_key"
+        assert "api_key" not in data
+        assert "secret" not in str(data).lower() or "sk-" not in str(data)
+
+    def test_oauth2_credential_no_secret(self):
+        cred = _make_oauth2_cred()
+        with patch(
+            "backend.api.features.integrations.router.creds_manager"
+        ) as mock_mgr:
```
majdyz, line 82: 🤖 🔵 **Nit**: `assert "secret" not in str(data).lower() or "sk-" not in str(data)` — this OR condition always passes if either side is true. The intent is to check BOTH, so it should be AND: `assert "sk-" not in str(data) and "secret" not in str(data).lower()`. Or better, just assert `"sk-secret-key-value" not in str(data)` which is the actual secret value being tested.

autogpt_platform/backend/backend/api/features/integrations/router_test.py
majdyz: 🤖 Fixed in 73fc4cc2d: Removed unused mock, added assert_not_called() to verify early return.

autogpt_platform/backend/backend/sdk/registry.py
majdyz: 🤖 Fixed in 73fc4cc2d: Added comment explaining the lazy import.

autogpt_platform/backend/backend/api/features/integrations/router_test.py
majdyz: 🤖 Fixed in 73fc4cc2d: Changed to direct assertion on the actual secret value.

PR #8113, opened 2024-09-20
refactor: Restructuring Repo

Swiftyos:
### Background

Restructuring the Repo to make it clear the difference between classic autogpt and the autogpt platform:
* Move the "classic" projects `autogpt`, `forge`, `frontend`, and `benchmark` into a `classic` folder
  * Also rename `autogpt` to `original_autogpt` for absolute clarity
* Rename `rnd/` to `autogpt_platform/`
  * `rnd/autogpt_builder` -> `autogpt_platform/frontend`
  * `rnd/autogpt_server` -> `autogpt_platform/backend`
* Adjust any paths accordingly

### TODO

- [ ] Update Readme
- [x] Check CI's
- [x] Test classic autogpt works
- [ ] Test AutoGPT Platform works
- [ ] Check Build and deployment process works

PR #12542, opened 2026-03-24
Add BuildAgent and AgentBuilder for specialized agent creation

Yurak67:
### Changes 🏗️

This PR introduces a new agent factory pattern and specialized BuildAgent class to enable creation of domain-specific agents from predefined templates:

1. **BuildAgent** (`autogpt/agent/build_agent.py`): A new Agent subclass specialized for software development tasks with:
   - Build-focused system prompts and constraints
   - Specialized command set (evaluate_code, improve_code, write_tests, execute_python_file, etc.)
   - Project directory and build configuration support
   - Custom `get_build_prompt()` static method for generating development-focused prompts

2. **AgentBuilder** (`autogpt/agent/agent_builder.py`): A factory class providing:
   - Fluent builder interface for agent configuration
   - Three predefined templates: "build", "research", and "code-review"
   - Template-based initialization with customization options
   - Support for both BuildAgent and standard Agent creation
   - Methods: `from_template()`, `with_name()`, `with_role()`, `with_goals()`, `with_project_dir()`, `with_build_config()`, `build_agent()`

3. **CLI Enhancement** (`autogpt/cli.py`): New `agent` command enabling:
   - Template-based agent creation from command line
   - Template selection (build, research, code-review)
   - Override of agent name, role, and project directory
   - All existing configuration options (continuous mode, memory type, etc.)

4. **Module Exports** (`autogpt/agent/__init__.py`): Updated to export BuildAgent and AgentBuilder

5. **Comprehensive Unit Tests** (`tests/unit/test_build_agent.py`): 
   - 24 test cases covering BuildAgent initialization, prompt generation, and AgentBuilder functionality
   - Tests for template loading, fluent interface, agent construction, and template validation

#### For code changes:
- [x] I have made a test plan
  - [x] Unit tests added covering BuildAgent and AgentBuilder (24 tests)
  - [x] Tests verify template loading and agent instantiation
  - [x] Tests validate fluent builder interface and method chaining
  - [x] Tests confirm BuildAgent inherits from Agent and stores configuration
  - [x] Tests verify prompt generation includes required build commands and constraints

https://claude.ai/code/session_013ojL3Ac7NqaJGKeKofNVwC

### DIFF
--- a/autogpt/agent/__init__.py
+++ b/autogpt/agent/__init__.py
@@ -1,4 +1,6 @@
 from autogpt.agent.agent import Agent
+from autogpt.agent.agent_builder import AgentBuilder
 from autogpt.agent.agent_manager import AgentManager
+from autogpt.agent.build_agent import BuildAgent
 
-__all__ = ["Agent", "AgentManager"]
+__all__ = ["Agent", "AgentBuilder", "AgentManager", "BuildAgent"]
--- a/autogpt/agent/agent_builder.py
+++ b/autogpt/agent/agent_builder.py
@@ -1,0 +1,194 @@
+"""A factory for constructing agents with predefined configurations."""
+from __future__ import annotations
+
+from autogpt.agent.agent import Agent
+from autogpt.agent.build_agent import BuildAgent
+from autogpt.config import Config
+from autogpt.memory import get_memory
+
+
+AGENT_TEMPLATES = {
+    "build": {
+        "ai_name": "BuildGPT",
+        "ai_role": "an AI developer agent that builds, tests, and improves software projects",
+        "ai_goals": [
+            "Analyze the project structure and understand the codebase",
+            "Identify and fix bugs or issues in the code",
+            "Write and run tests to ensure code quality",
+            "Improve code based on best practices and conventions",
+            "Complete the assigned build task efficiently",
+        ],
+    },
+    "research": {
+        "ai_name": "ResearchGPT",
+        "ai_role": "an AI research agent that gathers information and produces reports",
+        "ai_goals": [
+            "Search for relevant information on the given topic",
+            "Analyze and synthesize findings from multiple sources",
+            "Produce a clear and well-structured summary",
+            "Identify key insights and actionable recommendations",
+            "Save the research report to a file",
+        ],
+    },
+    "code-review": {
+        "ai_name": "ReviewGPT",
+        "ai_role": "an AI code review agent that analyzes code for quality and correctness",
+        "ai_goals": [
+            "Read and understand the codebase thoroughly",
+            "Identify bugs, security issues, and code smells",
+            "Suggest improvements following best practices",
+            "Evaluate test coverage and suggest missing tests",
+            "Produce a detailed code review report",
+        ],
+    },
+}
+
+DEFAULT_TRIGGERING_PROMPT = (
+    "Determine which next command to use, and respond using the"
+    " format specified above:"
+)
+
+
+class AgentBuilder:
+    """Factory for constructing configured Agent instances.
+
+    Provides a fluent builder interface and predefined templates for
+    creating specialized agents.
+
+    Example:
+        agent = (
+            AgentBuilder()
+            .with_name("MyBuilder")
+            .with_role("a software build agent")
+            .with_goals(["Build the project", "Run tests"])
+            .with_project_dir("/path/to/project")
+            .build_agent(agent_type="build")
+        )
+        agent.start_interaction_loop()
+    """
+
+    def __init__(self):
+        self._ai_name = ""
+        self._ai_role = ""
+        self._ai_goals = []
+        self._project_dir = ""
+        self._build_config = {}
+        self._memory_type = None
+        self._next_action_count = 0
+        self._continuous_mode = False
+        self._continuous_limit = 0
+
+    def with_name(self, name: str) -> AgentBuilder:
+        """Set the agent name."""
+        self._ai_name = name
+        return self
+
+    def with_role(self, role: str) -> AgentBuilder:
+        """Set the agent role description."""
+        self._ai_role = role
+        return self
+
+    def with_goals(self, goals: list[str]) -> AgentBuilder:
+        """Set the agent goals."""
+        self._ai_goals = goals
+        return self
+
+    def with_project_dir(self, project_dir: str) -> AgentBuilder:
+        """Set the project directory for build agents."""
+        self._project_dir = project_dir
+        return self
+
+    def with_build_config(self, config: dict) -> AgentBuilder:
+        """Set build-specific configuration."""
+        self._build_config = config
+        return self
+
+    def with_memory_type(self, memory_type: str) -> AgentBuilder:
+        """Set the memory backend type."""
+        self._memory_type = memory_type
+        return self
+
+    def with_next_action_count(self, count: int) -> AgentBuilder:
+        """Set the number of continuous actions to execute."""
+        self._next_action_count = count
+        return self
+
+    def from_template(self, template_name: str) -> AgentBuilder:
+        """Load settings from a predefined template.
+
+        Args:
+            template_name: One of 'build', 'research', or 'code-review'.
+
+        Returns:
+            self for method chaining.
+
+        Raises:
+            ValueError: If template_name is not recognized.
+        """
+        if template_name not in AGENT_TEMPLATES:
+            available = ", ".join(AGENT_TEMPLATES.keys())
+            raise ValueError(
+                f"Unknown template '{template_name}'. Available: {available}"
+            )
+
+        template = AGENT_TEMPLATES[template_name]
+        self._ai_name = template["ai_name"]
+        self._ai_role = template["ai_role"]
+        self._ai_goals = list(template["ai_goals"])
+        return self
+
+    def build_agent(self, agent_type: str = "default") -> Agent:
+        """Construct and return the configured agent.
+
+        Args:
+            agent_type: The type of agent to build.
+                'build' creates a BuildAgent with development-focused prompts.
+                'default' creates a standard Agent.
+
+        Returns:
+            A configured Agent or BuildAgent instance.
+        """
+        cfg = Config()
+        memory = get_memory(cfg, init=True)
+
+        if agent_type == "build":
+            system_prompt = BuildAgent.get_build_prompt(
+                ai_name=self._ai_name,
+                ai_role=self._ai_role,
+                ai_goals=self._ai_goals,
+                project_dir=self._project_dir,
+            )
+            return BuildAgent(
+                ai_name=self._ai_name,
+                memory=memory,
+                full_message_history=[],
+                next_action_count=self._next_action_count,
+                system_prompt=system_prompt,
+                triggering_prompt=DEFAULT_TRIGGERING_PROMPT,
+                project_dir=self._project_dir,
+                build_config=self._build_config,
+            )
+
+        # Default agent - use standard prompt construction
+        from autogpt.config.ai_config import AIConfig
+
+        ai_config = AIConfig(
+            ai_name=self._ai_name,
+            ai_role=self._ai_role,
+            ai_goals=self._ai_goals,
+        )
+        system_prompt = ai_config.construct_full_prompt()
+
+        return Agent(
+            ai_name=self._ai_name,
+            memory=memory,
+            full_message_history=[],
+            next_action_count=self._next_action_count,
+            system_prompt=system_prompt,
+            triggering_prompt=DEFAULT_TRIGGERING_PROMPT,
+        )
+
+    @staticmethod
+    def list_templates() -> list[str]:
+        """Return available template names."""
+        return list(AGENT_TEMPLATES.keys())
--- a/autogpt/agent/build_agent.py
+++ b/autogpt/agent/build_agent.py
@@ -1,0 +1,178 @@
+"""A specialized agent for software development and build tasks."""
+from colorama import Fore
+
+from autogpt.agent.agent import Agent
+from autogpt.config import Config
+from autogpt.logs import logger
+from autogpt.promptgenerator import PromptGenerator
+
+
+class BuildAgent(Agent):
+    """An agent specialized for software development and build tasks.
+
+    Extends the base Agent with build-specific system prompts, constraints,
+    and a focused command set for code evaluation, testing, and file operations.
+
+    Attributes:
+        project_dir: The root directory of the project to work on.
+        build_config: Optional dict of build-specific configuration overrides.
+    """
+
+    def __init__(
+        self,
+        ai_name,
+        memory,
+        full_message_history,
+        next_action_count,
+        system_prompt,
+        triggering_prompt,
+        project_dir="",
+        build_config=None,
+    ):
+        super().__init__(
+            ai_name=ai_name,
+            memory=memory,
+            full_message_history=full_message_history,
+            next_action_count=next_action_count,
+            system_prompt=system_prompt,
+            triggering_prompt=triggering_prompt,
+        )
+        self.project_dir = project_dir
+        self.build_config = build_config or {}
+
+    @staticmethod
+    def get_build_prompt(ai_name, ai_role, ai_goals, project_dir=""):
+        """Generate a build-focused system prompt.
+
+        Args:
+            ai_name: Name of the build agent.
+            ai_role: Role description for the agent.
+            ai_goals: List of build objectives.
+            project_dir: Root directory of the project.
+
+        Returns:
+            str: The complete system prompt for the build agent.
+        """
+        cfg = Config()
+        prompt_generator = PromptGenerator()
+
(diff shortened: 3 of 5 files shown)


majdyz (changes requested):
🤖
## Review

Thanks for the contribution! Unfortunately this PR can't be merged as-is because it targets the legacy `autogpt/` codebase which is no longer active.

### Main concern: targets an inactive codebase

All 5 files are under `autogpt/agent/`, `autogpt/cli.py`, etc. — these paths don't exist on the `dev` branch anymore. The project has moved to `autogpt_platform/`.

### A few other things worth noting

1. **No associated issue or design discussion.** For a feature of this scope (new agent factory, subclass, CLI subcommand), it'd be great to open an issue first so we can discuss the design together.

2. **Some unused code paths.** `_memory_type` is stored by `with_memory_type()` but never read by `build_agent()`. `_continuous_mode` and `_continuous_limit` are initialized but have no setters and aren't used.

3. **Only one template is fully implemented.** The CLI offers three templates but only `build` creates a specialized agent — the other two create a plain `Agent`.

4. **Missing input validation.** Calling `build_agent()` without setting name/role/goals produces a malformed prompt.

### Suggested next steps

If you're interested in contributing something like this:
1. Open an issue to discuss the design
2. Target code under `autogpt_platform/` on the `dev` branch
3. Start with a smaller, focused PR

We appreciate you taking the time — looking forward to future contributions!

majdyz (commented):
*(Superseded by the updated review above — please refer to that one for the full feedback.)*

autogpt/agent/agent_builder.py
```diff
@@ -0,0 +1,194 @@
...
+    def __init__(self):
+        self._ai_name = ""
+        self._ai_role = ""
+        self._ai_goals = []
+        self._project_dir = ""
+        self._build_config = {}
+        self._memory_type = None
+        self._next_action_count = 0
+        self._continuous_mode = False
+        self._continuous_limit = 0
```
majdyz, line 79: 🤖 🟠 **Should Fix**: `_memory_type` is set here (and via `with_memory_type()`), but `build_agent()` never reads it — the memory backend is always determined by `Config()`. This looks like dead code that could confuse future readers. Would you consider either wiring it into `get_memory()` or removing it until it's needed? That way the builder API won't promise something it doesn't deliver. :)

autogpt/agent/agent_builder.py
```diff
@@ -0,0 +1,194 @@
...
+
+    def with_name(self, name: str) -> AgentBuilder:
+        """Set the agent name."""
```
majdyz, line 82: 🤖 🟠 **Should Fix**: Same family of issue — `_continuous_mode` and `_continuous_limit` are initialized here but there are no setters exposed on the builder, and `build_agent()` never references them. It might be worth removing these for now to keep the surface area clean, and adding them back when there's a concrete use case. Less dead code == fewer surprises down the road!

autogpt/agent/agent_builder.py
```diff
@@ -0,0 +1,194 @@
+"""A factory for constructing agents with predefined configurations."""
+from __future__ import annotations
+
+from autogpt.agent.agent import Agent
+from autogpt.agent.build_agent import BuildAgent
+from autogpt.config import Config
+from autogpt.memory import get_memory
+
+
+AGENT_TEMPLATES = {
```
majdyz, line 10: 🤖 🟠 **Should Fix**: Nice set of templates! One thing I noticed: only the `"build"` template actually produces a specialized `BuildAgent` — the other two (`research`, `code-review`) create a plain `Agent` with different goals. That's not necessarily wrong, but it might surprise callers who expect template-specific behavior. It could be worth either:

- Documenting that non-build templates produce a generic agent, or
- Planning for `ResearchAgent` / `ReviewAgent` subclasses in follow-up work.

Just flagging so expectations are clear for users of the API.

autogpt/agent/build_agent.py
```diff
@@ -0,0 +1,178 @@
...
+
+        if project_dir:
+            prompt_start += f" Your project directory is: {project_dir}"
+
+        full_prompt = (
+            f"You are {ai_name}, {ai_role}\n{prompt_start}\n\nGOALS:\n\n"
+        )
+        for i, goal in enumerate(ai_goals):
+            full_prompt += f"{i + 1}. {goal}\n"
+
```
majdyz, line 161: 🤖 🔴 **Blocker**: If someone calls `build_agent()` without setting a name or role (e.g. `AgentBuilder().build_agent(agent_type="build")`), this line produces `"You are , "` which looks a bit odd in the LLM prompt. A quick validation in `build_agent()` (or here) — raising a clear error when `ai_name` / `ai_role` are empty — would make the failure mode much friendlier.

autogpt/agent/build_agent.py
```diff
@@ -0,0 +1,178 @@
...
+            " seeking user assistance. You are a software development agent"
+            " focused on building, testing, and improving code."
+        )
+
+        if project_dir:
+            prompt_start += f" Your project directory is: {project_dir}"
+
+        full_prompt = (
+            f"You are {ai_name}, {ai_role}\n{prompt_start}\n\nGOALS:\n\n"
+        )
```
majdyz, line 158: 🤖 🟠 **Should Fix**: Heads-up — `project_dir` is interpolated directly into the LLM system prompt here without any validation or sanitization. A path like `"; ignore all previous instructions"` would end up in the prompt verbatim. It would be great to at least verify that `project_dir` is a real directory (e.g. `os.path.isdir()`) before including it, both for correctness and to avoid prompt-injection footguns.

majdyz (commented):
🤖 Inline comments on specific code locations — see the main review above for context.

majdyz:
## Review: PR #12542 (BuildAgent and AgentBuilder)

This adds a build-focused agent to the **legacy AutoGPT CLI** (not the platform). This is the old `autogpt/` package.

### Concerns

1. **Wrong target** -- The `autogpt/` package is the legacy CLI agent that predates the platform. New features should be added to the platform (`autogpt_platform/`), not the legacy CLI. Unless there's a specific reason to extend the legacy agent, this PR should be reconsidered.

2. **Hardcoded prompts and templates** -- The agent templates use hardcoded goal lists and role descriptions. These are brittle and will drift from best practices.

3. **`Config()` instantiation in `build_agent`** -- Creates a new `Config()` each time an agent is built, which reads from disk. Should accept config as a parameter.

4. **No integration with the platform** -- This doesn't connect to the platform's agent execution engine, copilot, or block system.

**Verdict: Needs discussion on scope.** This adds complexity to the legacy CLI when the project focus is on the platform.

autogpt/agent/build_agent.py
```diff
@@ -0,0 +1,178 @@
...
+        """Generate a build-focused system prompt.
+
+        Args:
+            ai_name: Name of the build agent.
+            ai_role: Role description for the agent.
+            ai_goals: List of build objectives.
+            project_dir: Root directory of the project.
+
+        Returns:
+            str: The complete system prompt for the build agent.
```
majdyz, line 54: 🤖 🟠 **Should Fix**: `get_build_prompt()` instantiates `Config()` purely to check `cfg.execute_local_commands`, but the caller (`build_agent()` in `AgentBuilder`) also creates its own `Config()`. This means two separate `Config()` calls per agent build. More importantly, `Config()` is used inside a `@staticmethod` — if this were ever called outside the builder flow (e.g., in tests without mocking), it would reach out to the real environment config. Consider accepting `execute_local_commands` as a parameter instead, or making this a classmethod that receives a config object. It would make the method easier to test and eliminate the hidden dependency.

autogpt/cli.py
```diff
@@ -121,5 +121,121 @@ def main(
...
+    "--template",
+    "-t",
+    type=click.Choice(["build", "research", "code-review"]),
+    default="build",
+    help="Agent template to use",
+)
+@click.option("--name", "-n", help="Override the agent name")
+@click.option("--role", "-r", help="Override the agent role")
+@click.option(
+    "--project-dir",
```
majdyz, line 135: 🤖 🟠 **Should Fix**: The `--continuous-limit` option is accepted by the CLI but never passed to the builder. The builder's `_continuous_limit` field (which itself is unused per the earlier comment) is never set from this CLI argument. The user would think they're limiting continuous runs, but the setting would be silently ignored. Either wire it through or remove the CLI option to avoid false promises.

autogpt/agent/build_agent.py
```diff
@@ -0,0 +1,178 @@
...
+        full_prompt = (
+            f"You are {ai_name}, {ai_role}\n{prompt_start}\n\nGOALS:\n\n"
+        )
+        for i, goal in enumerate(ai_goals):
+            full_prompt += f"{i + 1}. {goal}\n"
+
+        full_prompt += f"\n\n{prompt_generator.generate_prompt_string()}"
+        return full_prompt
+
+    def start_interaction_loop(self):
```
majdyz, line 165: 🤖 🟡 **Nice to Have**: This override adds two log lines then calls `super().start_interaction_loop()`. The parent's `start_interaction_loop()` already logs the agent name. This means the agent name gets logged twice — once here as "BUILD AGENT STARTED" and once inside the parent. Consider removing this override and logging build-specific info (like `project_dir`) in `__init__` instead, or suppressing the parent's duplicate log.

tests/unit/test_build_agent.py
```diff
@@ -0,0 +1,214 @@
...
+    @patch("autogpt.agent.build_agent.Config")
+    def test_build_agent_returns_build_agent(
+        self, mock_ba_config, mock_config, mock_memory
+    ):
+        mock_ba_config.return_value.execute_local_commands = False
+        mock_memory.return_value = MagicMock()
+        agent = (
+            AgentBuilder()
+            .from_template("build")
+            .with_project_dir("/tmp")
```
majdyz, line 183: 🤖 🟠 **Should Fix**: The test suite is thorough for the builder's fluent API, but there's no integration-level test that actually verifies the *constructed agent* has the correct system prompt content — only `get_build_prompt` is tested in isolation. For example, there's no test confirming that `build_agent(agent_type=\"build\")` produces an agent whose `system_prompt` includes the expected constraints and commands. The builder could silently break the prompt wiring and these tests would still pass. Consider adding at least one test that asserts on the constructed agent's `system_prompt`.

autogpt/agent/__init__.py
```diff
@@ -1,4 +1,6 @@
 from autogpt.agent.agent import Agent
+from autogpt.agent.agent_builder import AgentBuilder
 from autogpt.agent.agent_manager import AgentManager
+from autogpt.agent.build_agent import BuildAgent
 
-__all__ = ["Agent", "AgentManager"]
+__all__ = ["Agent", "AgentBuilder", "AgentManager", "BuildAgent"]
```
majdyz, line 6: 🤖 🔴 **Blocker**: This PR targets the `dev` branch and modifies files under `autogpt/` — but the `autogpt/` directory no longer exists on `dev` or `master`. The entire project has migrated to `autogpt_platform/`. These files would either fail to merge or create a zombie directory. This is the fundamental issue that needs to be resolved before any code-level fixes are worth pursuing.

autogpt/agent/build_agent.py
```diff
@@ -0,0 +1,178 @@
...
+            ai_role: Role description for the agent.
+            ai_goals: List of build objectives.
+            project_dir: Root directory of the project.
+
+        Returns:
+            str: The complete system prompt for the build agent.
+        """
+        cfg = Config()
+        prompt_generator = PromptGenerator()
+
```
majdyz, line 58: 🤖 🟡 **Nice to Have**: All the constraints, resources, and performance evaluations are hardcoded as inline strings. If any other part of the system changes these conventions (e.g., the word limit, the command quoting convention), these strings would drift out of sync with the base Agent prompts. Consider extracting these into constants or sharing them with the parent class's prompt generation to reduce duplication and drift risk.

majdyz (changes requested):
## Review (follow-up)

Adding additional findings on top of the [previous review](https://github.com/Significant-Gravitas/AutoGPT/pull/12542#pullrequestreview-4005882714). The fundamental blocker (targets the dead `autogpt/` codebase) still stands. Below are the new code-level issues found in this round.

### Blockers

| File | Line | Issue |
|---|---|---|
| `autogpt/agent/__init__.py` | 6 | All files target `autogpt/` which no longer exists on `dev` or `master`. This PR would create a zombie directory on merge. |

### Should Fix

| File | Line | Issue |
|---|---|---|
| `autogpt/agent/build_agent.py` | 54 | `get_build_prompt()` is a `@staticmethod` that instantiates `Config()` internally — hidden side-effect, hard to test without mocking. Should accept config or the flag as a parameter. |
| `autogpt/cli.py` | 135 | `--continuous-limit` CLI option is accepted but never wired to the builder or the agent. Silently ignored. |
| `tests/unit/test_build_agent.py` | 183 | No integration test verifies the constructed agent's `system_prompt` — only `get_build_prompt` is tested in isolation. The builder could silently break prompt wiring. |

### Nice to Have

| File | Line | Issue |
|---|---|---|
| `autogpt/agent/build_agent.py` | 58 | All prompt constraints/resources/evaluations are hardcoded inline strings that could drift from the base Agent's prompts. |
| `autogpt/agent/build_agent.py` | 165 | `start_interaction_loop()` override just adds logging that duplicates the parent's own logging of the agent name. |

### Summary

This PR has 2 blockers (wrong codebase target + missing input validation from the previous review), 5 should-fix items, and 2 nice-to-haves. The previous review's request for changes still applies — this needs to target `autogpt_platform/` and the design should be discussed in an issue first.

majdyz (changes requested):
## Review

I see there have already been two rounds of review on this PR (with changes requested), and the author has not yet responded to any of the feedback. I am adding my review to reinforce the existing concerns and add some additional observations.

### Fundamental blocker: targets a non-existent codebase

As noted in the previous reviews, all 5 files in this PR are under top-level `autogpt/` — a directory that was removed from both `dev` and `master` in the repo reorganization (commit ef7cfbb860, PR #8113). The legacy code was moved to `classic/original_autogpt/`. There are currently **zero** files under top-level `autogpt/` on the `dev` branch. This PR would re-create a zombie directory that nothing in the project references.

The imports (`from autogpt.agent.agent import Agent`, `from autogpt.config import Config`, etc.) would not resolve against the current codebase. This is not a minor path issue — the entire PR needs to be re-targeted to the correct location if this feature is to move forward.

### No prior design discussion

This PR introduces a new architecture pattern (agent factory + agent subclass + CLI subcommand) with 705 lines of new code, but there is no linked issue or design discussion. For a feature of this scope, it would be valuable to first discuss:
- Whether the legacy CLI agent (`classic/original_autogpt/`) is the right place for this, or if it should integrate with the platform (`autogpt_platform/`)
- Whether the builder pattern is the preferred approach vs. the existing `agent_factory/` module that already lives in `classic/original_autogpt/autogpt/agent_factory/`
- What the relationship is between this and the platform's agent execution engine

### Code-level issues (for reference, should the above be resolved)

These echo and extend the previous review findings:

1. **Dead code in the builder**: `_memory_type`, `_continuous_mode`, and `_continuous_limit` are initialized but never used by `build_agent()`. The `with_memory_type()` setter exists but its value is discarded. This makes the API deceptive — callers think they are configuring something but it has no effect.

2. **CLI `--continuous-limit` is a no-op**: The CLI accepts the flag and passes it to `create_config()`, but the builder never reads it. A user would believe they are limiting continuous runs when they are not.

3. **No validation on agent name/role**: Calling `AgentBuilder().build_agent(agent_type="build")` without setting a name or role produces a prompt starting with `"You are , "` — an empty name interpolated into the LLM system prompt.

4. **`project_dir` injected into prompt without sanitization**: The `project_dir` string is interpolated directly into the LLM system prompt at line 158 of `build_agent.py`. There is no validation that it is an actual directory path, which could lead to prompt injection.

5. **Hidden `Config()` in `@staticmethod`**: `get_build_prompt()` instantiates `Config()` internally to check `execute_local_commands`. This is a hidden dependency that makes testing harder and creates a second `Config()` instantiation (the builder also creates one). Should accept the config or the relevant flag as a parameter.

6. **Only `build` template is specialized**: The `research` and `code-review` templates create a plain `Agent` with different goals, not specialized subclasses. This asymmetry is undocumented and may surprise users.

7. **Duplicate logging in `start_interaction_loop`**: The override logs the agent name, then `super().start_interaction_loop()` logs it again.

8. **No integration tests**: The test suite validates the builder's fluent API and `get_build_prompt()` in isolation, but never verifies that a fully built agent has the correct system prompt content.

### Recommendation

This PR should not be merged in its current state. The path forward would be:
1. Open an issue to discuss the design and target location
2. Determine whether this belongs in `classic/original_autogpt/` or `autogpt_platform/`
3. Address the dead code and validation issues
4. Resubmit against the correct file paths

majdyz:

Thanks for the contribution!

kcze:
Closing as stale assumed abandoned by the author.

PR #12545, opened 2026-03-25
fix(blocks): validate non-empty input in AIConversationBlock before LLM call

### LINKED ISSUE #11875: BlockUnknownError: raised by AIConversationBlock with message: Error calling LLM: Error code: 400 - {'error': {'message': "Invalid 'messages': empty array. Expected an array with minimum length 1, but got an empty array instead.", 'type': 'invalid_request_

Krishnachaitanyakc:
### Why / What / How

**Why:** When `AIConversationBlock` receives an empty messages list and an empty prompt, the block blindly forwards the empty array to the downstream LLM API, which returns a cryptic `400 Bad Request` error: `"Invalid 'messages': empty array. Expected an array with minimum length 1."` This is confusing for users who don't understand why their agent failed.

**What:** Add early input validation in `AIConversationBlock.run()` that raises a clear `ValueError` when both `messages` and `prompt` are empty. Also add three unit tests covering the validation logic.

**How:** A simple guard clause at the top of the `run` method checks `if not input_data.messages and not input_data.prompt` before the LLM call is made. If both are empty, a descriptive `ValueError` is raised. If either one has content, the block proceeds normally.

### Changes

- `autogpt_platform/backend/backend/blocks/llm.py`: Add validation guard in `AIConversationBlock.run()` to reject empty messages + empty prompt before calling the LLM
- `autogpt_platform/backend/backend/blocks/test/test_llm.py`: Add `TestAIConversationBlockValidation` with three tests:
  - `test_empty_messages_and_empty_prompt_raises_error` — validates the guard clause
  - `test_empty_messages_with_prompt_succeeds` — ensures prompt-only usage still works
  - `test_nonempty_messages_with_empty_prompt_succeeds` — ensures messages-only usage still works

#### For code changes:
- [x] I have made a test plan
  - [x] Lint passes (`ruff check`)
  - [x] Formatting passes (`ruff format`)
  - [x] New unit tests validate the empty-input guard and the happy paths

Closes #11875

### DIFF
--- a/autogpt_platform/backend/backend/blocks/llm.py
+++ b/autogpt_platform/backend/backend/blocks/llm.py
@@ -1999,9 +1999,11 @@ class AIConversationBlock(AIBlockBase):
     async def run(
         self, input_data: Input, *, credentials: APIKeyCredentials, **kwargs
     ) -> BlockOutput:
-        has_messages = bool(input_data.messages) and any(
-            isinstance(m, dict) and bool(m.get("content", "").strip())
-            for m in input_data.messages
+        has_messages = any(
+            isinstance(m, dict)
+            and isinstance(m.get("content"), str)
+            and bool(m["content"].strip())
+            for m in (input_data.messages or [])
         )
         has_prompt = bool(input_data.prompt and input_data.prompt.strip())
         if not has_messages and not has_prompt:
--- a/autogpt_platform/backend/backend/blocks/test/test_llm.py
+++ b/autogpt_platform/backend/backend/blocks/test/test_llm.py
@@ -619,6 +619,22 @@ class TestAIConversationBlockValidation:
             async for _ in block.run(input_data, credentials=llm.TEST_CREDENTIALS):
                 pass
 
+    @pytest.mark.asyncio
+    async def test_messages_with_none_content_raises_error(self):
+        """Messages with content=None should not crash with AttributeError."""
+        block = llm.AIConversationBlock()
+
+        input_data = llm.AIConversationBlock.Input(
+            messages=[{"role": "user", "content": None}],
+            prompt="",
+            model=llm.DEFAULT_LLM_MODEL,
+            credentials=_TEST_AI_CREDENTIALS,
+        )
+
+        with pytest.raises(ValueError, match="no messages and no prompt"):
+            async for _ in block.run(input_data, credentials=llm.TEST_CREDENTIALS):
+                pass
+
 
 class TestAITextSummarizerValidation:
     """Test that AITextSummarizerBlock validates LLM responses are strings."""


majdyz:
## Review: PR #12545 (AIConversationBlock empty input validation)

Clean validation guard. Prevents sending empty payloads to LLM APIs.

### Observations

1. **`any(input_data.messages)` check** -- this works because an empty list is falsy, and `any([])` returns False. However, if `messages` contains empty dicts or empty strings, `any()` would still return True. Consider whether `any(m.get("content") for m in input_data.messages)` would be more robust to catch messages with blank content.

2. **`raise ValueError`** -- correct approach. The block framework catches this and wraps it.

3. **Tests cover the three key scenarios** -- empty+empty, empty+prompt, messages+empty.

4. **Whitespace-only prompt handling** -- `input_data.prompt.strip()` is good, catches `"   "` prompts.

**Verdict: LGTM** -- minor suggestion on `any()` semantics above, not blocking.

autogpt_platform/backend/backend/blocks/llm.py
```diff
@@ -1999,6 +1999,14 @@ async def llm_call(
     async def run(
         self, input_data: Input, *, credentials: APIKeyCredentials, **kwargs
```
majdyz, line 2000: 🤖 🟠 **Should Fix**: `any(input_data.messages)` does not actually catch the edge cases it was intended to fix.

The `messages` field is typed as `List[Any]`. With this check:
- `messages=[{"role": "user", "content": ""}]` → `any()` returns `True` (a non-empty dict is truthy), so validation passes, but the message has **no meaningful content**. The downstream `AIStructuredResponseGeneratorBlock.run()` (line 1318) keeps it with `if p` (a non-empty dict is truthy), and the LLM receives a message with blank content.
- `messages=[{}]` → `any()` returns `False` (empty dict is falsy), so this IS caught. Good.
- `messages=[None]` → `any()` returns `False`. Also caught. Good.

But the stated motivation for switching from `not input_data.messages` to `any(input_data.messages)` was specifically to catch "cases like `messages=[None]` or `messages=[{}]`". While those are caught, the **more common real-world edge case** -- messages with empty `content` fields -- is NOT caught.

Consider checking for actual message content:
```python
has_messages = any(
    m.get("content") if isinstance(m, dict) else m
    for m in input_data.messages
)
```

This validates that at least one message has a truthy `content` value, which is what the LLM API actually needs.

Alternatively, if this level of validation feels like over-engineering, the simpler `not input_data.messages` from the original commit was arguably clearer and covered the primary use case (empty list). The `any()` version creates a false sense of security by implying it catches more cases than it actually does.

autogpt_platform/backend/backend/blocks/test/test_llm.py
```diff
@@ -488,6 +488,74 @@ async def mock_llm_call(*args, **kwargs):
...
+            messages=[],
+            prompt="",
+            model=llm.DEFAULT_LLM_MODEL,
+            credentials=_TEST_AI_CREDENTIALS,
+        )
+
+        with pytest.raises(ValueError, match="no messages and no prompt"):
+            async for _ in block.run(input_data, credentials=llm.TEST_CREDENTIALS):
+                pass
+
```
majdyz, line 509: 🤖 🟠 **Should Fix**: The test suite is missing the very edge cases the validation was updated to handle.

Commit 2's message says: *"Validate normalized input using any() and strip() instead of raw truthiness checks, catching edge cases like [None] or [{}] messages"*

But there are no tests for:
- `messages=[None]` — the guard should catch this
- `messages=[{}]` — the guard should catch this
- `messages=[{"role": "user", "content": ""}]` — **currently passes validation but shouldn't** (see other comment)
- `prompt="   "` (whitespace-only) — the `strip()` was added for this, but there's no test proving it works

Please add tests for the edge cases the validation claims to handle. Without them, future refactors could regress these cases silently.

majdyz:

Thanks for the contribution!

autogpt_platform/backend/backend/blocks/llm.py
majdyz: 🤖 Fixed in 5e49842c5a: Replaced `any(input_data.messages)` with content-level validation that checks each message dict has non-empty stripped content. Now catches `[{"role": "user", "content": ""}]`, `[{"role": "user", "content": "   "}]`, `[None]`, and `[{}]`.

autogpt_platform/backend/backend/blocks/test/test_llm.py
majdyz: 🤖 Fixed in 5e49842c5a: Added 4 edge case tests — empty content, whitespace-only content, None entry, and empty dict entry. All correctly raise ValueError.

autogpt_platform/backend/backend/blocks/llm.py
majdyz: 🤖 Updated in cd291792d8: Also added `isinstance` guard for non-string content per CodeRabbit feedback.

autogpt_platform/backend/backend/blocks/llm.py
```diff
@@ -1999,6 +1999,17 @@ async def llm_call(
...
     ) -> BlockOutput:
+        has_messages = bool(input_data.messages) and any(
+            isinstance(m, dict) and bool(m.get("content", "").strip())
+            for m in input_data.messages
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in cd291792d8: Added `isinstance(m.get("content"), str)` check before `.strip()` so `{"content": None}` raises ValueError instead of AttributeError. Also added a test for this case.

autogpt_platform/backend/backend/blocks/test/test_llm.py
majdyz: 🤖 Updated in cd291792d8: Added 5th edge case test for `content=None`.

PR #12546, opened 2026-03-25
fix(blocks): validate email recipients in Gmail blocks before API call

### LINKED ISSUE #11954: BlockUnknownError: raised by GmailSendBlock with message: <HttpError 400 when requesting https://gmail.googleapis.com/gmail/v1/users/me/messages/send?alt=json returned "Invalid To header". Details: "[{'message': 'Invalid To header', 'domain': 'global', 're

Krishnachaitanyakc:
### Why / What / How

**Why:** When a user or LLM supplies a malformed recipient string (e.g. a bare username, a JSON blob, or an empty value) to `GmailSendBlock`, `GmailCreateDraftBlock`, or any reply block, the Gmail API returns an opaque `HttpError 400: "Invalid To header"`. This surfaces as a `BlockUnknownError` with no actionable guidance, making it impossible for the LLM to self-correct. (Fixes #11954)

**What:** Adds a lightweight `validate_email_recipients()` function that checks every recipient against a simplified RFC 5322 pattern (`local@domain.tld`) and raises a clear `ValueError` listing all invalid entries before any API call is made.

**How:** The validation is called in two shared code paths — `create_mime_message()` (used by send and draft blocks) and `_build_reply_message()` (used by reply blocks) — so all Gmail blocks that compose outgoing email benefit from it with zero per-block changes. The regex is intentionally permissive (any `x@y.z` passes) to avoid false positives on unusual but valid addresses.

### Changes 🏗️

- Added `validate_email_recipients()` helper in `gmail.py` with a compiled regex
- Hooked validation into `create_mime_message()` for `to`, `cc`, and `bcc` fields
- Hooked validation into `_build_reply_message()` for reply/draft-reply blocks
- Added `TestValidateEmailRecipients` test class covering valid, invalid, mixed, empty, JSON-string, and field-name scenarios

#### For code changes:
- [x] I have made a test plan
  - [x] Verified `validate_email_recipients` correctly accepts valid emails (`user@example.com`, `a@b.com`, `test@sub.domain.co`)
  - [x] Verified it rejects malformed entries (bare names, missing domain dot, empty strings, JSON strings)
  - [x] Verified error messages include the field name and all invalid entries
  - [x] Verified empty recipient lists pass without error
  - [x] Confirmed `gmail.py` and test file parse correctly (AST check)

### DIFF
--- a/autogpt_platform/backend/backend/blocks/google/gmail.py
+++ b/autogpt_platform/backend/backend/blocks/google/gmail.py
@@ -9,7 +9,7 @@ from email.mime.text import MIMEText
 from email.policy import SMTP
 from email.utils import getaddresses, parseaddr
 from pathlib import Path
-from typing import List, Literal, Optional
+from typing import List, Literal, Optional, Protocol, runtime_checkable
 
 from google.oauth2.credentials import Credentials
 from googleapiclient.discovery import build
@@ -70,7 +70,14 @@ def validate_email_recipients(recipients: list[str], field_name: str = "to") -> 
         )
 
 
-def validate_all_recipients(input_data) -> None:
+@runtime_checkable
+class HasRecipients(Protocol):
+    to: list[str]
+    cc: list[str]
+    bcc: list[str]
+
+
+def validate_all_recipients(input_data: HasRecipients) -> None:
     """Validate to/cc/bcc recipient fields on an input namespace.
 
     Calls ``validate_email_recipients`` for ``to`` (required) and
@@ -138,7 +145,6 @@ async def create_mime_message(
 ) -> str:
     """Create a MIME message with attachments and return base64-encoded raw message."""
 
-    # Validate all recipient lists before building the MIME message
     validate_all_recipients(input_data)
 
     message = MIMEMultipart()
@@ -1208,7 +1214,6 @@ async def _build_reply_message(
         references.append(headers["message-id"])
 
     # Create MIME message
-    # Validate all recipient lists before building the MIME message
     validate_all_recipients(input_data)
 
     msg = MIMEMultipart()
--- a/autogpt_platform/backend/test/blocks/test_gmail.py
+++ b/autogpt_platform/backend/test/blocks/test_gmail.py
@@ -6,7 +6,9 @@ from unittest.mock import Mock, patch
 import pytest
 
 from backend.blocks.google.gmail import (
+    GmailForwardBlock,
     GmailReadBlock,
+    HasRecipients,
     _build_reply_message,
     create_mime_message,
     validate_all_recipients,
@@ -305,26 +307,32 @@ class TestValidateAllRecipients:
     """Test cases for validate_all_recipients."""
 
     def test_valid_all_fields(self):
-        data = SimpleNamespace(to=["a@b.com"], cc=["c@d.com"], bcc=["e@f.com"])
+        data = cast(
+            HasRecipients,
+            SimpleNamespace(to=["a@b.com"], cc=["c@d.com"], bcc=["e@f.com"]),
+        )
         validate_all_recipients(data)
 
     def test_invalid_to_raises(self):
-        data = SimpleNamespace(to=["bad"], cc=[], bcc=[])
+        data = cast(HasRecipients, SimpleNamespace(to=["bad"], cc=[], bcc=[]))
         with pytest.raises(ValueError, match="'to'"):
             validate_all_recipients(data)
 
     def test_invalid_cc_raises(self):
-        data = SimpleNamespace(to=["a@b.com"], cc=["bad"], bcc=[])
+        data = cast(HasRecipients, SimpleNamespace(to=["a@b.com"], cc=["bad"], bcc=[]))
         with pytest.raises(ValueError, match="'cc'"):
             validate_all_recipients(data)
 
     def test_invalid_bcc_raises(self):
-        data = SimpleNamespace(to=["a@b.com"], cc=["c@d.com"], bcc=["bad"])
+        data = cast(
+            HasRecipients,
+            SimpleNamespace(to=["a@b.com"], cc=["c@d.com"], bcc=["bad"]),
+        )
         with pytest.raises(ValueError, match="'bcc'"):
             validate_all_recipients(data)
 
     def test_empty_cc_bcc_skipped(self):
-        data = SimpleNamespace(to=["a@b.com"], cc=[], bcc=[])
+        data = cast(HasRecipients, SimpleNamespace(to=["a@b.com"], cc=[], bcc=[]))
         validate_all_recipients(data)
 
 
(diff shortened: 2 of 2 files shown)


autogpt_platform/backend/backend/blocks/google/gmail.py
```diff
@@ -100,6 +122,13 @@ async def create_mime_message(
 ) -> str:
```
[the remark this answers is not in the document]
majdyz: 🤖 🟠 **Should Fix**: Fixed in a9776b58c: Added `validate_email_recipients()` calls for `to`, `cc`, and `bcc` in `_forward_message()` before MIME construction, matching the pattern used in `create_mime_message()` and `_build_reply_message()`.

autogpt_platform/backend/backend/blocks/google/gmail.py
```diff
@@ -46,6 +47,25 @@ def serialize_email_recipients(recipients: list[str]) -> str:
...
+# RFC 5322 simplified pattern: local@domain where domain has at least one dot
+_EMAIL_RE = re.compile(r"^[^@\s]+@[^@\s]+\.[^@\s]+$")
+
+
+def validate_email_recipients(recipients: list[str], field_name: str = "to") -> None:
+    """Validate that all recipients are plausible email addresses.
+
+    Raises ``ValueError`` with a user-friendly message listing every
+    invalid entry so the caller (or LLM) can correct them in one pass.
+    """
```
majdyz, line 63: 🤖 🟠 **Should Fix**: Good catch on stripping whitespace before matching! One subtle thing worth noting: `addr.strip()` is applied here for validation, so `"  user@example.com  "` passes validation. However, the raw (untrimmed) values are later passed into the MIME headers (e.g. `serialize_email_recipients(input_data.to)` and `", ".join(input_data.to)`), which means leading/trailing spaces could leak into the actual email headers.

It might be nice to either:
- Strip the addresses at the call sites too (or normalize them in a shared helper), or
- Remove the `.strip()` here so validation catches whitespace-padded entries.

Either approach would keep validation and construction consistent. What do you think?

autogpt_platform/backend/backend/blocks/google/gmail.py
```diff
@@ -1167,6 +1194,14 @@ async def _build_reply_message(
         references.append(headers["message-id"])
 
     # Create MIME message
+    # Validate all recipient lists before building the MIME message
+    if input_data.to:
```
majdyz: 🤖 🟡 **Nice to Have**: Small consistency note: in `create_mime_message()`, `to` is validated unconditionally (no `if` guard), whereas here in `_build_reply_message()` it's guarded with `if input_data.to:`. This means an empty `to` list would raise in one path but silently pass in the other.

Depending on the intended behavior, it might be worth aligning both call sites — either both unconditional or both guarded — so callers get consistent feedback regardless of which code path is used.

autogpt_platform/backend/test/blocks/test_gmail.py
```diff
@@ -250,3 +250,43 @@ async def test_attachment_download_failure(self):
 
         result = await self.gmail_block._get_email_body(msg, self.mock_service)
         assert result == "This email does not contain a readable body."
+
```
majdyz, line 264: 🤖 🟡 **Nice to Have**: Great unit test coverage for the `validate_email_recipients` helper — these are thorough and well-structured!

One thing that would further strengthen confidence: a few integration-style tests that verify the validation hooks fire correctly inside `create_mime_message()` and `_build_reply_message()` (e.g., mocking the service layer and asserting that invalid recipients raise before any MIME construction or API call happens). This would guard against regressions if the call sites are refactored later. Just a suggestion for a follow-up if you'd like!

autogpt_platform/backend/backend/blocks/google/gmail.py
majdyz: 🤖 Fixed in b217d47de: Went with the "strip in both places" approach — `serialize_email_recipients()` now strips each address before joining, and all raw `", ".join(...)` calls for recipient headers have been replaced with `serialize_email_recipients()` across `create_mime_message`, `_build_reply_message`, and `_forward_message`. This keeps validation and MIME construction consistent.

autogpt_platform/backend/backend/blocks/google/gmail.py
majdyz: 🤖 Fixed in b217d47de: Made `to` validation unconditional in `_build_reply_message()`, matching the pattern in `create_mime_message()` and `_forward_message()`. All three code paths now validate `to` without a guard.

autogpt_platform/backend/test/blocks/test_gmail.py
majdyz: 🤖 🟡 **Nice to Have**: Added integration tests in fbb3af2c1 — tests verify that create_mime_message() and _build_reply_message() correctly propagate ValueError for invalid recipients.

majdyz:
@Krishnachaitanyakc — could you please sign the [CLA](https://cla-assistant.io/Significant-Gravitas/AutoGPT?pullRequest=12546)? Once done, tag me (@majdyz) and I'll get this reviewed and merged. Thanks\!

Krishnachaitanyakc:
@majdyz signed the CLA

majdyz:
## Review: PR #12546 (Gmail email validation)

Good defensive fix. Validates email addresses before attempting to build MIME messages, preventing cryptic API errors.

### Observations

1. **Regex is appropriately simple** -- `^[^@\s]+@[^@\s]+\.[^@\s]+$` catches the most common bad inputs (missing @, missing domain, spaces) without being over-strict. Good choice over a full RFC 5322 parser.

2. **Validation runs in 3 places** -- `create_mime_message`, `_build_reply_message`, `_forward_message`. This is correct since they each build MIME independently. Could consider extracting a shared helper that builds+validates MIME headers, but not blocking.

3. **`serialize_email_recipients` now strips whitespace** -- good, mirrors the strip in validation.

4. **Tests are thorough** -- covers valid, invalid, edge cases (plus-addressing, subdomains, etc.).

5. **One edge case**: `validate_email_recipients` raises `ValueError`, which the block framework wraps as `BlockExecutionError`. This is the correct behavior -- the error message is user-friendly and actionable.

**Verdict: LGTM** -- ready to merge.

Krishnachaitanyakc:
@majdyz done, move the validation to a helper method

autogpt_platform/backend/backend/blocks/google/gmail.py
```diff
@@ -42,8 +43,45 @@
...
+    invalid = [addr for addr in recipients if not _EMAIL_RE.match(addr.strip())]
+    if invalid:
+        formatted = ", ".join(f"'{a}'" for a in invalid)
+        raise ValueError(
+            f"Invalid email address(es) in '{field_name}': {formatted}. "
+            f"Each entry must be a valid email address (e.g. user@example.com)."
+        )
+
+
+def validate_all_recipients(input_data) -> None:
```
majdyz: 🤖 🟠 **Should Fix**: The `validate_all_recipients(input_data)` parameter has no type annotation. Per the backend CLAUDE.md: "No duck typing — no `hasattr`/`getattr`/`isinstance` for type dispatch; use typed interfaces/unions/protocols."

Currently this function relies on duck typing (assuming `input_data` has `.to`, `.cc`, `.bcc` attributes). Since it's called with several different `Input` pydantic models (`GmailSendBlock.Input`, `GmailReplyBlock.Input`, `GmailForwardBlock.Input`, etc.), the right fix is to define a `Protocol`:

```python
from typing import Protocol

class HasRecipients(Protocol):
    to: list[str]
    cc: list[str]
    bcc: list[str]

def validate_all_recipients(input_data: HasRecipients) -> None:
    ...
```

This also applies to `create_mime_message(input_data, ...)` and `_build_reply_message(service, input_data, ...)` which also have untyped `input_data` parameters — but those are pre-existing issues, not introduced by this PR. For the new code in this PR though, please type `validate_all_recipients`.

autogpt_platform/backend/test/blocks/test_gmail.py
```diff
@@ -250,3 +259,158 @@ async def test_attachment_download_failure(self):
...
+            bcc=[],
+            subject="",
+            body="Reply body",
+            replyAll=False,
+            attachments=[],
+        )
+        exec_ctx = cast(ExecutionContext, SimpleNamespace(graph_exec_id="test-exec-id"))
+
+        with pytest.raises(ValueError, match="Invalid email address"):
+            await _build_reply_message(mock_service, input_data, exec_ctx)
```
majdyz, line 424: 🤖 🟡 **Nice to Have**: The PR added integration tests for `create_mime_message()` and `_build_reply_message()`, but there is no integration test verifying that `_forward_message()` also correctly rejects invalid recipients. Since the forward path was the one that was originally *missed* (caught by Sentry bot) and required a follow-up commit (a9776b58), it would be especially valuable to have a test proving that code path is now covered. Without a test, a future refactor could accidentally remove the validation call from `_forward_message` and nobody would notice.

autogpt_platform/backend/backend/blocks/google/gmail.py
```diff
@@ -42,8 +43,45 @@
...
+    """Serialize recipients list to comma-separated string.
+
+    Strips leading/trailing whitespace from each address to keep MIME
+    headers clean (mirrors the strip done in ``validate_email_recipients``).
+    """
+    return ", ".join(addr.strip() for addr in recipients)
+
+
+# RFC 5322 simplified pattern: local@domain where domain has at least one dot
+_EMAIL_RE = re.compile(r"^[^@\s]+@[^@\s]+\.[^@\s]+$")
```
majdyz, line 55: 🤖 🔵 **Nit**: Very minor regex observation. The pattern `^[^@\s]+@[^@\s]+\.[^@\s]+$` intentionally trades strict RFC compliance for low false-positive rates, and the PR description acknowledges this. Just want to flag one edge case for awareness:

- `user@123.123.123.123` passes (IP with dots) -- this is actually fine since Gmail accepts it
- `user@.com` is rejected (the `+` after the first `[^@\s]` requires at least one char before the dot) -- good

The regex looks solid for the stated goal. No change needed.

autogpt_platform/backend/backend/blocks/google/gmail.py
```diff
@@ -100,14 +138,17 @@ async def create_mime_message(
 ) -> str:
     """Create a MIME message with attachments and return base64-encoded raw message."""
 
+    # Validate all recipient lists before building the MIME message
+    validate_all_recipients(input_data)
```
majdyz, line 148: 🤖 🔵 **Nit**: The comment `# Validate all recipient lists before building the MIME message` is duplicated verbatim at three call sites (lines 142, 1213, 1737). The function name `validate_all_recipients` is already self-documenting. Consider removing the comments — the code reads fine without them, and having the same comment in three places is noise.

autogpt_platform/backend/test/blocks/test_gmail.py
majdyz: 🤖 Fixed in 54645c9747: Added `TestForwardMessageValidation` class with tests for invalid `to`, `cc`, and `bcc` recipients in `_forward_message()`. All 34 tests pass.

autogpt_platform/backend/backend/blocks/google/gmail.py
majdyz: 🤖 Fixed in 54645c9747: Added `HasRecipients` Protocol and extracted `validate_all_recipients(input_data: HasRecipients)` typed helper. All three call sites now use this single function instead of duplicating the validation pattern.

autogpt_platform/backend/backend/blocks/google/gmail.py
majdyz: 🤖 Acknowledged — good analysis. No changes needed, the regex covers the stated goals well.

autogpt_platform/backend/backend/blocks/google/gmail.py
majdyz: 🤖 Fixed in 54645c9747: Removed all three duplicate comments. The call sites now just use `validate_all_recipients(input_data)` which is self-documenting.

PR #12548, opened 2026-03-25
fix(backend): prevent graph execution stuck + steer SDK away from bash_exec

majdyz:
## Summary

Two backend fixes for CoPilot stability:

1. **Steer model away from bash_exec for SDK tool-result files** — When the SDK returns tool results as file paths, the copilot model was attempting to use `bash_exec` to read them instead of treating the content directly. Added system prompt guidance to prevent this.

2. **Guard against missing 'name' in execution input_data** — `GraphExecution.from_db()` assumed all INPUT/OUTPUT block node executions have a `name` field in `input_data`. This crashes with `KeyError: 'name'` when non-standard blocks (e.g., OrchestratorBlock) produce node executions without this field. Added `"name" in exec.input_data` guards.

## Why

- The bash_exec issue causes copilot to fail when processing SDK tool outputs
- The KeyError crashes the `update_graph_execution_stats` endpoint, causing graph executions to appear stuck (retries 35+ times, never completes)

## How

- Added system prompt instruction to treat tool result file contents directly
- Added `"name" in exec.input_data` guard in both input extraction (line 340) and output extraction (line 365) in `execution.py`

### Changes
- `backend/copilot/sdk/service.py` — system prompt guidance
- `backend/data/execution.py` — KeyError guard for missing `name` field

### Checklist 📋
- [x] I have made a test plan

#### Test plan:
- [x] OrchestratorBlock graph execution no longer gets stuck
- [x] Standard Agent Input/Output blocks still work correctly
- [x] Copilot SDK tool results are processed without bash_exec

### DIFF
--- a/autogpt_platform/backend/backend/copilot/sdk/tool_adapter.py
+++ b/autogpt_platform/backend/backend/copilot/sdk/tool_adapter.py
@@ -609,9 +609,7 @@ async def _read_file_handler(args: dict[str, Any]) -> dict[str, Any]:
 _READ_TOOL_NAME = "Read"
 _READ_TOOL_DESCRIPTION = (
     "Read a file from the local filesystem. "
-    "Use offset and limit to read specific line ranges for large files. "
-    "Use this tool (NOT bash_exec) to read SDK tool-result files under "
-    "~/.claude/projects/.../tool-results/."
+    "Use offset and limit to read specific line ranges for large files."
 )
 _READ_TOOL_SCHEMA = {
     "type": "object",
--- a/autogpt_platform/backend/backend/copilot/tools/bash_exec.py
+++ b/autogpt_platform/backend/backend/copilot/tools/bash_exec.py
@@ -44,8 +44,7 @@ class BashExecTool(BaseTool):
         return (
             "Execute a Bash command or script. Shares filesystem with SDK file tools. "
             "Useful for scripts, data processing, and package installation. "
-            "Killed after timeout (default 30s, max 120s). "
-            "CANNOT read SDK tool-result files (~/.claude/projects/...); use Read instead."
+            "Killed after timeout (default 30s, max 120s)."
         )
 
     @property


autogpt_platform/backend/backend/data/execution.py
```diff
@@ -342,6 +342,7 @@ def from_db(_graph_exec: AgentGraphExecution):
                     if (
                         (block := get_block(exec.block_id))
                         and block.block_type == BlockType.INPUT
+                        and "name" in exec.input_data
```
majdyz, line 345: 🤖 🟠 **Should Fix**: Missing test for the `"name" in exec.input_data` guard. Per CLAUDE.md TDD guidelines: "Every bug fix should include a test that would have caught it." A test with a mock OrchestratorBlock node execution (no `name` in `input_data`) that verifies `from_db()` doesn't crash would prevent regression.

autogpt_platform/backend/backend/data/execution.py
```diff
@@ -360,8 +361,10 @@ def from_db(_graph_exec: AgentGraphExecution):
         outputs: CompletedBlockOutput = defaultdict(list)
         for exec in complete_node_executions:
             if (
-                block := get_block(exec.block_id)
-            ) and block.block_type == BlockType.OUTPUT:
+                (block := get_block(exec.block_id))
+                and block.block_type == BlockType.OUTPUT
+                and "name" in exec.input_data
+            ):
```
majdyz, line 367: 🤖 🟡 **Nice to Have**: Consider adding a `logger.debug` when skipping node executions without `name` — silent data exclusion makes debugging harder when new block types are added and their outputs unexpectedly disappear from graph execution results.

autogpt_platform/backend/backend/data/execution.py
majdyz: 🤖 Addressed in #12552 — added `execution_test.py` with `test_from_db_skips_node_executions_without_name` that creates mock node executions with and without `name` in `input_data`, verifying `from_db()` completes without `KeyError` and correctly extracts only named inputs/outputs.

autogpt_platform/backend/backend/data/execution.py
majdyz: 🤖 Addressed in #12552 — refactored the output loop to use early-continue guards and added `logger.debug("Skipping %s node execution %s: no 'name' in input_data", block.block_type, exec.node_exec_id)` when skipping.

PR #12552, opened 2026-03-25
fix(backend): add logger.debug and test for nameless node exec skipping

majdyz:
## Summary
Follow-up to #12548, addressing review comments:
- Add `logger.debug` when skipping OUTPUT node executions without `name` in `input_data` (e.g. OrchestratorBlock nodes)
- Add colocated test (`execution_test.py`) verifying `GraphExecution.from_db()` handles mixed named/unnamed node executions without `KeyError`

## Test plan
- [x] `poetry run pytest backend/data/execution_test.py -xvs` passes
- [x] Pre-commit hooks pass (format, lint, typecheck)

### DIFF
--- a/autogpt_platform/backend/backend/data/execution.py
+++ b/autogpt_platform/backend/backend/data/execution.py
@@ -360,18 +360,12 @@ class GraphExecution(GraphExecutionMeta):
 
         outputs: CompletedBlockOutput = defaultdict(list)
         for exec in complete_node_executions:
-            block = get_block(exec.block_id)
-            if not block or block.block_type != BlockType.OUTPUT:
-                continue
-            if "name" not in exec.input_data:
-                logger.debug(
-                    "Skipping OUTPUT node execution %s (block_id=%s): "
-                    "no 'name' in input_data",
-                    exec.node_exec_id,
-                    exec.block_id,
-                )
-                continue
-            outputs[exec.input_data["name"]].append(exec.input_data.get("value"))
+            if (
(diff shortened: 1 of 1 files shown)


autogpt_platform/backend/backend/data/execution.py
```diff
@@ -359,10 +360,17 @@ def from_db(_graph_exec: AgentGraphExecution):
...
         for exec in complete_node_executions:
-            if (
-                block := get_block(exec.block_id)
-            ) and block.block_type == BlockType.OUTPUT:
-                outputs[exec.input_data["name"]].append(exec.input_data.get("value"))
+            block = get_block(exec.block_id)
+            if not block or block.block_type != BlockType.OUTPUT:
+                continue
+            if "name" not in exec.input_data:
+                logger.debug(
```
majdyz: 🤖 🟡 **Nice to Have**: The debug log here uses `block.block_type` which is always `OUTPUT` (the guard above already filters for it). Consider logging `exec.block_id` instead for easier triage of malformed node executions.

majdyz: 🤖 Fixed in 1a84a10af: Changed debug log to use `exec.block_id` instead of `block.block_type`. Also resolved merge conflict with dev.

majdyz:
🤖 Closing — the logger.debug was unnecessary, and the test can be added as part of another PR. The original guard from #12548 is sufficient.

PR #12558, opened 2026-03-25
fix(backend): handle UniqueViolationError in workspace upsert race

Otto-AGPT:
Requested by @majdyz

Prisma's `upsert` can still throw `UniqueViolationError` under concurrent requests for the same user (known Prisma race condition). When two requests race to create a workspace, one succeeds and the other hits the unique constraint on `userId`, returning a 500 to the user.

## Problem

Sentry: **AUTOGPT-SERVER-8BK** — `UniqueViolationError: Unique constraint failed on the fields: (userId)`

Happens on `/api/workspace/files/upload` when `get_or_create_workspace` is called concurrently for the same user (e.g., multiple file uploads or parallel copilot/chat requests).

## Fix

Catch `UniqueViolationError` after the upsert and fall back to `find_unique` to fetch the workspace that was created by the concurrent request. Single file, minimal change.

Resolves SECRT-2170

---
Co-authored-by: Zamil Majdy (@majdyz) <zamil.majdy@agpt.co>

### DIFF
--- a/autogpt_platform/backend/backend/data/workspace.py
+++ b/autogpt_platform/backend/backend/data/workspace.py
@@ -9,6 +9,7 @@ from datetime import datetime, timezone
 from typing import Optional
 
 import pydantic
+from prisma.errors import UniqueViolationError
 from prisma.models import UserWorkspace, UserWorkspaceFile
 from prisma.types import UserWorkspaceFileWhereInput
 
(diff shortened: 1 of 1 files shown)


Otto-AGPT:
Closing to retarget to master with hotfix/* branch naming (CI requirement). Reopened as new PR.

PR #12571, opened 2026-03-26
feat(copilot): add copy button to user prompt messages [SECRT-2172]

Otto-AGPT:
Requested by @itsababseh

Users can copy assistant output messages but not their own prompts. This adds the same copy button to user messages — appears on hover, right-aligned, using the existing `CopyButton` component.

## Why

Users write long prompts and need to copy them to reuse or share. Currently requires manual text selection. ChatGPT shows copy on hover for user messages — this matches that pattern.

## What

- Added `CopyButton` to user prompt messages in `ChatMessagesContainer.tsx`
- Shows on hover (`group-hover:opacity-100`), positioned right-aligned below the message
- Reuses the existing `CopyButton` and `MessageActions` components — zero new code

## How

One file changed, 11 lines added:
1. Import `MessageActions` and `CopyButton`
2. Render them after user `MessageContent`, gated on `message.role === "user"` and having text parts

---
Co-authored-by: itsababseh (@itsababseh) <36419647+itsababseh@users.noreply.github.com>

### DIFF
--- a/autogpt_platform/frontend/src/app/(platform)/copilot/components/ChatMessagesContainer/ChatMessagesContainer.tsx
+++ b/autogpt_platform/frontend/src/app/(platform)/copilot/components/ChatMessagesContainer/ChatMessagesContainer.tsx
@@ -246,7 +246,7 @@ export function ChatMessagesContainer({
                 )}
               </MessageContent>
               {message.role === "user" && textParts.length > 0 && (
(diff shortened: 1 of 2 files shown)


itsababseh:
image
Attempting to add copy button on all prompts to the right side same style as our copy button for output and for all prompts. The screenshot is from chatgpt so we align on right side of prompt as a reference. 

/review

ntindle (approved):

PR #12572, opened 2026-03-26
fix(copilot): preserve thinking blocks on session resume

Otto-AGPT:
## Why

CoPilot sessions using Claude models with extended thinking (e.g. Claude Opus 4) break on **every multi-turn conversation**. After turn 1 succeeds, turns 2+ fail instantly with:

```
messages.1.content.37: `thinking` or `redacted_thinking` blocks in the latest
assistant message cannot be modified. These blocks must remain as they were
in the original response.
```

This affects every extended-thinking session on resume — **34 occurrences in Sentry since March 13** (AUTOGPT-SERVER-875, severity 0.96).

## Root Cause

`_format_sdk_content_blocks()` in `service.py` had two issues:

1. **Selective field serialization for ThinkingBlock** — only `type`, `thinking`, and `signature` were explicitly copied. If the API or a future SDK version includes additional fields on thinking blocks, they would be silently dropped. The Anthropic API validates that thinking blocks are byte-identical to the original response on session resume.

2. **Unknown block types silently dropped** — any block type not in the `isinstance` chain (e.g. `redacted_thinking` blocks that may come through when routing via OpenRouter) was logged as a warning and skipped entirely. This causes content array index shifts, which the API detects as thinking block modification.

## Evidence

- **Sentry:** AUTOGPT-SERVER-875 — 34 occurrences, high severity, unresolved
- **Langfuse:** Session `31d3f08a-cb94-45eb-9fce-56b3f0287ef4` — turn 1 succeeds (`resume: false`), turns 2+3 fail instantly (`resume: true`, `duration_api_ms: 0`)
- **Model:** `claude-opus-4.6` via OpenRouter → Azure upstream
- **Server:** `autogpt-copilot-executor-7d8b688bc4-7vcjm`

### `service.py` — `_format_sdk_content_blocks()`
- **ThinkingBlock:** Use `dataclasses.asdict()` instead of manual field picking — automatically preserves all current and future fields
- **Unknown blocks:** Generic passthrough via `dataclasses.asdict()` / `__dict__` instead of dropping — preserves blocks like `redacted_thinking` that the SDK doesn't have a typed class for yet

### `sdk_compat_test.py`
- Added `ThinkingBlock` to content block type export test (was missing)
- Added `test_thinking_block_has_required_fields` — verifies `thinking` and `signature` fields exist

## Risk

Low — the fix preserves strictly more data than before. Existing broken transcripts will still fail (they're already broken), but new sessions will work correctly.

Fixes SECRT-2173

---
Co-authored-by: Zamil Majdy (@majdyz) <zamil.majdy@agpt.co>

### DIFF
--- a/autogpt_platform/backend/backend/copilot/sdk/sdk_compat_test.py
+++ b/autogpt_platform/backend/backend/copilot/sdk/sdk_compat_test.py
@@ -38,10 +38,22 @@ def test_sdk_exports_message_types():
 
 
 def test_sdk_exports_content_block_types():
-    from claude_agent_sdk import TextBlock, ToolResultBlock, ToolUseBlock
+    from claude_agent_sdk import TextBlock, ThinkingBlock, ToolResultBlock, ToolUseBlock
 
-    for cls in (TextBlock, ToolResultBlock, ToolUseBlock):
+    for cls in (TextBlock, ThinkingBlock, ToolResultBlock, ToolUseBlock):
         assert inspect.isclass(cls), f"{cls.__name__} is not a class"
+
+
+def test_thinking_block_has_required_fields():
+    """ThinkingBlock must have thinking + signature for API integrity on resume."""
+    import dataclasses
+
+    from claude_agent_sdk import ThinkingBlock
+
+    assert dataclasses.is_dataclass(ThinkingBlock)
+    field_names = {f.name for f in dataclasses.fields(ThinkingBlock)}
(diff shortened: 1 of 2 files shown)


majdyz:
Closing — implementing a more robust fix in a separate hotfix PR that preserves thinking blocks in the last assistant message during transcript compaction (instead of stripping them everywhere). Some changes here (like the _flatten_assistant_content thinking block stripping) will be adapted into the more comprehensive fix.

PR #12574, opened 2026-03-26
fix(backend): preserve thinking blocks during transcript compaction

majdyz:
## Why

AutoPilot users hit `invalid_request_error` ("thinking or redacted_thinking blocks in the latest assistant message cannot be modified") when sessions get long enough to trigger transcript compaction. The Anthropic API requires thinking blocks in the last assistant message to be byte-for-byte identical to the original response — our compaction was flattening them to plain text, destroying the cryptographic signatures.

Reported in Discord `#breakage` by John Ababseh with session `31d3f08a-cb94-45eb-9fce-56b3f0287ef4`.

## What

- **`compact_transcript`** now splits the transcript into a compressible prefix and a preserved tail (last assistant entry + trailing entries). Only the prefix is compressed; the tail is re-appended verbatim, preserving thinking blocks exactly.
- **`_flatten_assistant_content`** now silently drops `thinking` and `redacted_thinking` blocks instead of creating `[__thinking__]` placeholders — they carry no useful context for compression summaries.
- **`response_adapter`** explicitly handles `ThinkingBlock` (skip gracefully instead of silently falling through the isinstance chain).
- **`_format_sdk_content_blocks`** now passes through raw dict blocks (e.g. `redacted_thinking` that the SDK may not have a typed class for) verbatim to the transcript.

## How

The key insight is the Anthropic API's asymmetric constraint:
- **Last assistant message**: thinking/redacted_thinking blocks must be preserved byte-for-byte
- **Older assistant messages**: thinking blocks can be removed entirely

`compact_transcript` uses `_find_last_assistant_entry()` to split the JSONL into two parts:
1. **Prefix** (everything before the last assistant): flattened and compressed normally
2. **Tail** (last assistant + any trailing user message): preserved verbatim and re-chained via `_rechain_tail()` to maintain the `parentUuid` chain

This ensures the API always sees the original thinking blocks in the last assistant message while still achieving meaningful compression on older turns.

## Test plan
- [x] 25 new tests across `thinking_blocks_test.py` (TDD: written before implementation)
  - [x] `_find_last_assistant_entry` splits correctly at last assistant, handles edges (no assistant, index 0, trailing user)
  - [x] `_rechain_tail` patches parentUuid chain, handles empty tail
  - [x] `_flatten_assistant_content` strips thinking/redacted_thinking blocks, handles mixed content
  - [x] `compact_transcript` preserves last assistant's thinking blocks
  - [x] `compact_transcript` strips thinking from older assistant messages
  - [x] Edge cases: trailing user message, single assistant, no thinking blocks
  - [x] `response_adapter` handles ThinkingBlock without crash
  - [x] `_format_sdk_content_blocks` preserves thinking block format and raw dict blocks
- [x] All existing copilot SDK tests pass
- [x] Pre-commit hooks (lint, format, typecheck) all pass


autogpt_platform/backend/backend/copilot/sdk/transcript.py
```diff
@@ -805,6 +816,36 @@ async def _run_compression(
...
+    guarantee those blocks are never altered.
+
+    Returns ``(all_lines, [])`` when no assistant entry is found.
+    """
+    lines = [ln for ln in content.strip().split("\n") if ln.strip()]
+    last_asst_idx: int | None = None
+    for i, line in enumerate(lines):
+        entry = json.loads(line, fallback=None)
+        if not isinstance(entry, dict):
+            continue
```
majdyz, line 851: 🤖 🟡 **Nice to Have**: `_find_last_assistant_entry` should consider the edge case where the assistant is at index 0 — the entire transcript becomes the tail and there's nothing to compress. This is handled by the `if not messages: return None` guard later, but a direct unit test for `_find_last_assistant_entry` would make that coverage explicit and catch future regressions.

autogpt_platform/backend/backend/copilot/sdk/transcript.py
```diff
@@ -860,7 +905,13 @@ async def compact_transcript(
             result.messages_summarized,
```
majdyz, line 941: 🤖 🟠 **Should Fix**: The compacted output (compressed prefix + preserved tail) could be **larger** than the input if the last assistant message is big (many thinking blocks + tool calls). The caller in `_reduce_context` checks `compacted \!= transcript_content` but not `len(compacted) < len(content)`. This wastes a retry attempt — the larger compacted transcript would still hit prompt-too-long, then fall through to DB fallback on attempt 2. Consider adding a size check before returning.

autogpt_platform/backend/backend/copilot/sdk/thinking_blocks_test.py
```diff
@@ -0,0 +1,525 @@
...
+
+        # Parse the compacted transcript and find the last assistant entry
+        last_assistant_content = None
+        for line in result.strip().split("\n"):
+            entry = json.loads(line)
+            msg = entry.get("message", {})
+            if msg.get("role") == "assistant":
+                last_assistant_content = msg.get("content")
+
+        assert last_assistant_content is not None, "No assistant entry found"
```
majdyz: 🤖 🔵 **Nit**: The pattern of "parse compacted transcript → find last assistant → extract content blocks" is repeated in 4 test methods. Consider extracting a `_find_last_assistant_content(transcript_jsonl)` helper at the top of the test file to reduce duplication.

autogpt_platform/backend/backend/copilot/sdk/transcript.py
majdyz: 🤖 Fixed in b5e971a06: Added 4 unit tests for `_find_last_assistant_entry` covering: splits at last assistant, no assistant (all prefix), assistant at index 0 (empty prefix), trailing user in tail.

autogpt_platform/backend/backend/copilot/sdk/transcript.py
majdyz: 🤖 Fixed in b5e971a06: Changed from rejection to warning log. Bytes ≠ tokens — a compacted transcript can be larger in bytes (due to preserved thinking blocks JSON overhead) but smaller in tokens (which is what the API validates). The caller already falls through to DB fallback if the compacted transcript still triggers prompt-too-long.

autogpt_platform/backend/backend/copilot/sdk/thinking_blocks_test.py
majdyz: 🤖 Fixed in b5e971a06: Extracted `_last_assistant_content(transcript_jsonl)` helper. Used in all 4 test methods that had the duplicated pattern.

autogpt_platform/backend/backend/copilot/sdk/thinking_blocks_test.py
```diff
@@ -0,0 +1,568 @@
...
+        )()
+        with patch(
+            "backend.copilot.sdk.transcript._run_compression",
+            new_callable=AsyncMock,
+            return_value=mock_result,
+        ):
+            result = await compact_transcript(transcript, model="test-model")
+
+        assert result is not None
+        assert validate_transcript(result)
```
majdyz, line 708: 🤖 🔵 **Nit**: `test_no_thinking_blocks_still_works` only checks `validate_transcript(result)` which just verifies an assistant entry exists. Consider adding `last_content = _last_assistant_content(result)` assertion to verify the last assistant text is preserved (e.g. `assert last_content == [{"type": "text", "text": "Details"}]`), ensuring the tail preservation works for non-thinking transcripts too.

majdyz: 🤖 Fixed in fcdd64c41: Added `_last_assistant_content` assertion to verify the last assistant's text block content is preserved verbatim.

majdyz:
🤖 Addressed coderabbit nitpicks in 19304b43c:

1. **parentUuid rewiring assertion** (`transcript.py` L926-958): Added `test_tail_parentuuid_rewired_to_prefix` — runs a full `compact_transcript` cycle and asserts the first preserved tail entry's `parentUuid` equals the last compressed-prefix entry's `uuid`.

2. **Raw-dict regression test** (`service.py` L632-634): Added `test_raw_dict_redacted_thinking_preserved` — passes a raw `{"type": "redacted_thinking", ...}` dict through `_format_sdk_content_blocks` and asserts it comes back unchanged, exercising the `elif isinstance(block, dict)` branch.

majdyz:
🤖 Both CodeRabbit nitpick suggestions are already covered:

1. **parentUuid chain assertion**: `test_parentuuid_chain_correct_after_compaction` (line 540) directly verifies the tail entry's parentUuid links to the last compressed prefix UUID. Additionally, `TestRechainTail::test_patches_first_entry_parentuuid` unit-tests the chain patching.

2. **Raw dict passthrough test**: `test_raw_dict_redacted_thinking_preserved` (line 726) passes a raw `{"type": "redacted_thinking", ...}` dict through `_format_sdk_content_blocks` and asserts it comes back unchanged.

autogpt_platform/backend/backend/copilot/sdk/transcript.py
```diff
@@ -805,6 +816,36 @@ async def _run_compression(
...
+    for i, line in enumerate(lines):
+        entry = json.loads(line, fallback=None)
+        if not isinstance(entry, dict):
+            continue
+        msg = entry.get("message", {})
+        if msg.get("role") == "assistant":
+            last_asst_idx = i
+    if last_asst_idx is None:
+        return lines, []
+    return lines[:last_asst_idx], lines[last_asst_idx:]
```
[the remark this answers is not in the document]
majdyz: 🤖 Good catch — fixed in 554197f87. `_find_last_assistant_entry` now identifies the `message.id` of the last assistant entry, then walks forward to find the **first** entry with that same id. The entire multi-entry turn (e.g., thinking entry + tool_use entry sharing the same `message.id`) is preserved in the tail. Added `test_multi_entry_turn_fully_preserved` test covering this exact scenario.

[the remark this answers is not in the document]
majdyz: 🤖 Fixed in 554197f87: `_find_last_assistant_entry` now performs a two-pass scan — first finds the `message.id` of the last assistant entry by scanning in reverse, then scans forward to find the first entry sharing that same `message.id`. This ensures the entire multi-entry turn (thinking + tool_use entries) is preserved in the tail. Added `test_multi_entry_turn_fully_preserved` to validate this behavior.

autogpt_platform/backend/backend/copilot/sdk/transcript.py
```diff
@@ -870,3 +954,38 @@ async def compact_transcript(
...
+        entry = json.loads(line, fallback=None)
+        if not isinstance(entry, dict):
+            result_lines.append(line)
+            continue
+        if i == 0:
+            entry["parentUuid"] = last_prefix_uuid
+        elif prev_uuid is not None:
+            entry["parentUuid"] = prev_uuid
+        prev_uuid = entry.get("uuid")
+        result_lines.append(json.dumps(entry, separators=(",", ":")))
```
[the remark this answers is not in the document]
majdyz: 🤖 Investigated — the current behavior is intentional and correct. The tail can contain multiple entries from the same assistant turn (e.g., a thinking entry + a tool_use entry sharing the same `message.id`). These entries' original `parentUuid` values may point to entries that were moved into the compressed prefix and no longer exist. Rewriting all tail entries to chain to their predecessor in the tail ensures a valid parent-child chain regardless of what was compressed.

The existing test `test_chains_multiple_tail_entries` validates exactly this: the second tail entry's `parentUuid` is rewritten from `"old2"` to `"t1"` (the first tail entry's uuid).

Updated the docstring to be more precise about the rechaining behavior.

majdyz:
## E2E Test Report: PR #12574 — Preserve thinking blocks during transcript compaction

### Summary
All tests **PASS**. The fix correctly preserves thinking blocks in the last assistant message during transcript compaction while stripping them from older messages. No `invalid_request_error` observed. Conversation continuity maintained after compaction.

### Unit Test Results
- **555/555 copilot SDK tests pass** (zero failures, zero regressions)
- **21/21 new thinking_blocks_test.py tests pass** — covers `_flatten_assistant_content`, `_find_last_assistant_entry`, `_rechain_tail`, `compact_transcript`, `response_adapter`, `_format_sdk_content_blocks`
- **56/56 prompt_too_long_test.py tests pass** — existing compaction tests adapted to 3-message output
- **42/42 retry_scenarios_test.py tests pass** — existing retry tests adapted
- **75/75 compaction-related tests pass** across all test files

### Test Scenarios

| 1 | Unit: `_flatten_assistant_content` strips thinking blocks | **PASS** | 4 tests: thinking, redacted_thinking, thinking-only, mixed |
| 2 | Unit: `_find_last_assistant_entry` splits correctly | **PASS** | 4 tests: normal, no assistant, index 0, trailing user |
| 3 | Unit: `_rechain_tail` patches parentUuid chain | **PASS** | 4 tests: patches first, chains multiple, empty, preserves content |
| 4 | Unit: `compact_transcript` preserves last assistant thinking | **PASS** | Thinking + redacted_thinking blocks byte-for-byte identical |
| 5 | Unit: `compact_transcript` strips old assistant thinking | **PASS** | Captured compression input has no old thinking content |
| 6 | Unit: `compact_transcript` with trailing user | **PASS** | Last assistant thinking preserved despite trailing user msg |
| 7 | Unit: `compact_transcript` single assistant | **PASS** | Only assistant is both first and last — thinking preserved |
| 8 | Unit: `compact_transcript` no thinking blocks | **PASS** | Normal compaction still works, last assistant content preserved verbatim |
| 9 | Unit: `response_adapter` handles ThinkingBlock | **PASS** | No crash, thinking not streamed to frontend |
| 10 | Unit: `_format_sdk_content_blocks` preserves thinking | **PASS** | ThinkingBlock serialized correctly, raw dicts passed through |
| 11 | SDK mode: Send copilot message | **PASS** | Response received, 01-04 screenshots |
| 12 | SDK mode: Multi-turn with tool calls | **PASS** | Agent built with 31 messages, 65 transcript entries, 1MB |
| 13 | SDK mode: Compaction triggered | **PASS** | `compact_boundary` received at 07:00:17 in logs |
| 14 | SDK mode: Response after compaction | **PASS** | Coherent response about agent blocks after compaction, 09 screenshot |
| 15 | Edge: Very large thinking blocks (100KB) | **PASS** | Preserved byte-for-byte in tail |
| 16 | Edge: Multiple thinking blocks in sequence | **PASS** | All thinking + redacted_thinking preserved |
| 17 | Edge: Thinking mixed with tool_use | **PASS** | Thinking stripped from prefix, preserved in tail |
| 18 | Edge: Empty `_rechain_tail` inputs | **PASS** | Handles empty prefix gracefully |
| 19 | Negative: Invalid session ID | **PASS** | Returns "Session not found" error, no crash |
| 20 | Regression: Basic chat (no thinking) | **PASS** | "2+2=4" response works fine |
| 21 | Regression: All 555 SDK tests pass | **PASS** | Zero failures |

### Key Log Evidence

**Compaction triggered and succeeded:**
```
[SDK] Context compaction triggered: auto, user=4851b58e
[SDK][6eb0236f-35d][T3] Received: SystemMessage compact_boundary
[SDK][6eb0236f-35d][T3] Received: UserMessage
[SDK][6eb0236f-35d][T3] Received: AssistantMessage
[SDK][6eb0236f-35d][T3] Stream completed successfully with 35 messages
```

**No thinking block errors:**
```
grep -i "invalid_request\|thinking.*cannot.*modified" → (empty)
```


### Screenshots

#### Login Page

Login page loaded, credentials entered for test user.

#### After Login

Successfully authenticated, copilot home page visible with chat history.

#### Copilot Before Message

Clean copilot chat ready for input.

#### First Response (Research with Tool Calls)

Copilot responds and performs web search — thinking blocks handled in transcript.

#### Research Complete

Full research response with sources delivered successfully.

#### Follow-up: Agent Building

Second turn — copilot builds an agent with multiple tool calls (31 messages, 65 transcript entries).

#### Agent Building In Progress

Tool calls executing: finding actions, creating blocks, connecting them.

#### Agent Building Complete

Agent creation completed with all blocks connected.

#### After Compaction (Critical)

**Key test:** Third turn triggered context compaction (compact_boundary in logs). Copilot responded coherently about the agent blocks it built — proving thinking blocks were preserved and conversation continuity was maintained after compaction. No invalid_request_error.

#### New Chat Regression

All chat sessions visible in sidebar. New chat works normally. No regressions.

🤖 10 image link(s) in this comment point at files that were already missing from the `test-screenshots/*` branch; they were broken before this rewrite and stay broken. Nothing in this comment could be re-hosted.

autogpt_platform/backend/backend/copilot/sdk/transcript.py
```diff
@@ -805,6 +816,56 @@ async def _run_compression(
...
+        entry = json.loads(line, fallback=None)
+        if not isinstance(entry, dict):
+            continue
+        msg = entry.get("message", {})
+        if msg.get("role") == "assistant":
+            last_asst_msg_id = msg.get("id")
+            break
+
+    if last_asst_msg_id is None:
+        return lines, []
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in 43f8c071a: When the last assistant entry has no `message.id`, the function now falls back to preserving from that single entry onward instead of treating the entire transcript as compressible. Added `test_no_message_id_preserves_last_assistant` to validate this behavior.

majdyz:
🤖 **Review round — 0 new findings**

Performed thorough review of all 7 changed files. Checked:

- **Correctness**: `_find_last_assistant_entry` two-pass scan (reverse to find last assistant, forward to find first entry of that turn), `_rechain_tail` parentUuid patching, `compact_transcript` prefix/tail split logic, `_flatten_assistant_content` thinking block filtering
- **Edge cases**: empty prefix, single assistant message, trailing user message, no `message.id`, multi-entry turns sharing the same `message.id`, no thinking blocks at all
- **Serialization safety**: Verified `orjson` round-trip stability — `json.loads` -> `json.dumps` preserves dict key insertion order, so thinking block content integrity is maintained through `_rechain_tail` re-serialization
- **Security**: `_flatten_assistant_content` correctly strips thinking content from compression input (no leakage into summaries)
- **Testing**: 25 tests cover splitting, rechaining, flattening, compaction, response adapter, and SDK content block formatting
- **CI**: All checks green (lint, type-check 3.11/3.12/3.13, test 3.11/3.12/3.13, e2e, CodeQL, security)

All prior review comments have been addressed. No new blockers, should-fixes, or nits. LGTM.

autogpt_platform/backend/backend/copilot/sdk/transcript.py
```diff
@@ -805,6 +816,63 @@ async def _run_compression(
...
+
+    # If the assistant entry has no message.id, fall back to preserving
+    # from that single entry onward — safer than compressing everything.
+    if last_asst_msg_id is None:
+        return lines[:last_asst_idx], lines[last_asst_idx:]
+
+    # Second pass: find the first entry of this turn (same message.id).
+    first_turn_idx: int | None = None
+    for i, line in enumerate(lines):
```
majdyz: 🤖 🟡 **Nice to Have**: `_find_last_assistant_entry` parses every JSONL line twice — once in the reverse scan (lines 842-850) and once in the forward scan (lines 862-869). For very large transcripts (which is exactly when compaction triggers), this double parse could be meaningful. Consider caching parsed entries in a single pass, e.g. parse all lines once into a `list[tuple[str, dict | None]]` and then scan the parsed list in both directions. This would halve JSON parse overhead on the hot path.

autogpt_platform/backend/backend/copilot/sdk/transcript.py
```diff
@@ -605,20 +605,31 @@ async def delete_transcript(user_id: str, session_id: str) -> None:
...
+    This is intentional: ``compress_context`` requires plain text for
+    token counting and LLM summarization.  The structural loss is
+    acceptable because compaction only runs when the original transcript
+    was already too large for the model.
     """
     parts: list[str] = []
     for block in blocks:
         if isinstance(block, dict):
             btype = block.get("type", "")
+            if btype in _THINKING_BLOCK_TYPES:
```
majdyz, line 631: 🤖 🟡 **Nice to Have**: When all content blocks in an older assistant message are thinking/redacted_thinking, `_flatten_assistant_content` returns `""`. This empty-string assistant message gets passed to `compress_context` via `_transcript_to_messages`. While the compressor handles empty content without crashing, it is wasted context (an empty assistant turn contributes nothing to the summary). Consider filtering out messages with empty content after flattening in `_transcript_to_messages` to produce cleaner compression input.

autogpt_platform/backend/backend/copilot/sdk/transcript.py
```diff
@@ -860,7 +932,26 @@ async def compact_transcript(
             result.messages_summarized,
             result.messages_dropped,
         )
-        compacted = _messages_to_transcript(result.messages)
+        compressed_part = _messages_to_transcript(result.messages)
```
majdyz, line 944: 🤖 🟡 **Nice to Have**: After `_run_compression` returns, there is no guard against `result.messages` being empty. If `compress_context` returns `was_compacted=True` but with an empty messages list (unlikely, but possible in edge cases or future refactors), `_messages_to_transcript([])` returns `""`, and the final transcript would be just the tail with no compressed prefix. Consider adding a defensive check:

```python
if not result.messages:
    logger.warning("%s Compressor returned empty messages", log_prefix)
    return None
```

autogpt_platform/backend/backend/copilot/sdk/conftest.py
```diff
@@ -52,3 +52,60 @@ def build_test_transcript(pairs: list[tuple[str, str]]) -> str:
...
+            msg = {"role": role, "content": content}
+        entry = {
+            "type": entry_type,
+            "uuid": uid,
+            "parentUuid": last_uuid,
+            "message": msg,
+        }
+        lines.append(json.dumps(entry, separators=(",", ":")))
+        last_uuid = uid
+    return "\n".join(lines) + "\n"
```
majdyz, line 94: 🤖 🔵 **Nit**: `build_structured_transcript` and the existing `build_test_transcript` share ~80% of the same logic (UUID generation, parentUuid chaining, entry construction). Consider making `build_structured_transcript` the single implementation and having `build_test_transcript` delegate to it by converting its `(role, content_str)` pairs to the structured format. This would eliminate the code duplication and reduce future maintenance burden.

autogpt_platform/backend/backend/copilot/sdk/transcript.py
majdyz: 🤖 Fixed in b568d52f4: `_find_last_assistant_entry` now parses all lines once into a `list[dict | None]` and scans the cached list in both directions, halving JSON parse overhead on large transcripts.

autogpt_platform/backend/backend/copilot/sdk/transcript.py
majdyz: 🤖 Investigated — filtering empty-content messages in `_transcript_to_messages` would break the round-trip invariant tested by `test_messages_to_transcript_empty_content`. The compressor handles empty strings without issues, and thinking-only assistant turns in the prefix are rare in practice (they almost always have an accompanying text block). Leaving as-is.

autogpt_platform/backend/backend/copilot/sdk/transcript.py
majdyz: 🤖 Fixed in b568d52f4: Added defensive guard `if not result.messages: return None` after `_run_compression` returns, before calling `_messages_to_transcript`.

autogpt_platform/backend/backend/copilot/sdk/conftest.py
majdyz: 🤖 Fixed in b568d52f4: `build_test_transcript` now delegates to `build_structured_transcript` with a type-widening cast to satisfy pyright invariance on `list[tuple]`. Eliminates the duplicated UUID/parentUuid chaining and entry construction logic.

autogpt_platform/backend/backend/copilot/sdk/transcript.py
```diff
@@ -805,6 +816,65 @@ async def _run_compression(
...
+    blocks in the **last** assistant message remain byte-for-byte identical
+    to the original response.  By excluding the entire turn from
+    compression we guarantee those blocks are never altered.
+
+    Returns ``(all_lines, [])`` when no assistant entry is found.
+    """
+    lines = [ln for ln in content.strip().split("\n") if ln.strip()]
+
+    # Parse all lines once to avoid double JSON deserialization.
+    parsed: list[dict | None] = [json.loads(ln, fallback=None) for ln in lines]
```
majdyz: 🤖 🔵 **Nit**: The type annotation `parsed: list[dict | None]` on line 840 is too narrow. `json.loads(ln, fallback=None)` can return any JSON-deserializable type (list, str, int, bool, None), not just `dict | None`. The code is correct since `isinstance(entry, dict)` guards all access, but `list[Any]` would be more accurate. Minor since pyright does not flag this (the `fallback=None` overload returns `Any`).

majdyz: 🤖 Fixed in aab1bcfcd: Changed `list[dict | None]` to `list` with an explanatory comment that `json.loads` with `fallback=None` returns `Any`, and non-dict entries are safely skipped by `isinstance` guards.

autogpt_platform/backend/backend/copilot/sdk/transcript.py
```diff
@@ -870,3 +968,39 @@ async def compact_transcript(
...
+        entry = json.loads(line, fallback=None)
+        if not isinstance(entry, dict):
+            result_lines.append(line)
+            continue
+        if i == 0:
+            entry["parentUuid"] = last_prefix_uuid
+        elif prev_uuid is not None:
+            entry["parentUuid"] = prev_uuid
+        prev_uuid = entry.get("uuid")
+        result_lines.append(json.dumps(entry, separators=(",", ":")))
```
majdyz, line 1014: 🤖 🟡 **Nice to Have**: `_rechain_tail` re-serializes all preserved tail entries through a JSON round-trip (`json.loads` then `json.dumps`), which means the `message.content` blocks — including the thinking/redacted_thinking blocks the PR aims to preserve — go through `orjson` serialization. While the *data values* are preserved, the key ordering in the serialized output may differ from the original JSONL line. The PR description claims "byte-for-byte identical" preservation, but this is only true at the parsed-data level, not the literal JSON text level.

This is safe in practice (the Anthropic API validates parsed signature values, not raw JSON bytes), but the language in the PR description and docstrings should be tightened to say "value-identical" rather than "byte-for-byte identical" to avoid misleading future maintainers.

autogpt_platform/backend/backend/copilot/sdk/transcript.py
```diff
@@ -870,3 +968,39 @@ async def compact_transcript(
...
+            last_prefix_uuid = entry["uuid"]
+            break
+
+    result_lines: list[str] = []
+    prev_uuid: str | None = None
+    for i, line in enumerate(tail_lines):
+        entry = json.loads(line, fallback=None)
+        if not isinstance(entry, dict):
+            result_lines.append(line)
+            continue
```
majdyz, line 1008: 🤖 🔵 **Nit**: When a non-dict tail line is encountered (line 997-999), `prev_uuid` is not updated — it keeps its previous value (or stays `None`). For the very first line (i=0) being non-dict, `prev_uuid` remains `None`, so the next dict entry at i=1 hits `elif prev_uuid is not None` which is `False`, leaving its `parentUuid` unchanged. This means the chain is silently broken for that entry.

In practice this never triggers because `_find_last_assistant_entry` filters empty lines and all transcript lines parse to dicts. But the defensive guard could be more robust by tracking `prev_uuid` through non-dict lines or adding a comment explaining why the guard is safe.

autogpt_platform/backend/backend/copilot/sdk/transcript.py
```diff
@@ -860,7 +939,26 @@ async def compact_transcript(
...
+            # validates tokens not bytes, and the caller falls through
+            # to DB fallback if the transcript is still too large.
+            logger.warning(
+                "%s Compacted transcript (%d bytes) is not smaller than "
+                "original (%d bytes) — may still reduce token count",
+                log_prefix,
+                len(compacted),
+                len(content),
+            )
         if not validate_transcript(compacted):
```
majdyz, line 967: 🤖 🔵 **Nit**: `validate_transcript(compacted)` is called here, but the caller `_reduce_context` in `service.py:306` also calls `validate_transcript(compacted)`. This is redundant — consider documenting which layer owns the validation responsibility to avoid the double check.

autogpt_platform/backend/backend/copilot/sdk/transcript.py
majdyz: 🤖 Fixed in 9f25e0eb4: Replaced "byte-for-byte identical" with "value-identical" in all three docstrings (_flatten_assistant_content, _find_last_assistant_entry, compact_transcript). Added clarification that the API validates parsed signature values, not raw JSON bytes.

autogpt_platform/backend/backend/copilot/sdk/transcript.py
majdyz: 🤖 Fixed in 9f25e0eb4: Added explanatory comment in _rechain_tail documenting that non-dict lines are passed through unchanged and prev_uuid is intentionally not updated, with rationale for why this is safe (upstream filters guarantee well-formed JSONL).

autogpt_platform/backend/backend/copilot/sdk/transcript.py
majdyz: 🤖 Fixed in 9f25e0eb4: Added comment clarifying this is the authoritative validation and that the caller also validates as a secondary check.

autogpt_platform/backend/backend/copilot/sdk/thinking_blocks_test.py
```diff
@@ -0,0 +1,822 @@
+"""Tests for thinking/redacted_thinking block preservation.
+
+Validates the fix for the Anthropic API error:
+    "thinking or redacted_thinking blocks in the latest assistant message
+    cannot be modified. These blocks must remain as they were in the
+    original response."
+
+The API requires that thinking blocks in the LAST assistant message are
+preserved byte-for-byte. Older assistant messages may have thinking blocks
```
majdyz: 🤖 🔵 **Nit**: The test file module docstring (line 9) and a test comment (line 464) still use "byte-for-byte" language, which was corrected to "value-identical" in the production code docstrings in the previous commit. These should be updated for consistency.

autogpt_platform/backend/backend/copilot/sdk/thinking_blocks_test.py
```diff
@@ -0,0 +1,822 @@
...
+                        THINKING_BLOCK,
+                        {"type": "text", "text": "Hi there"},
+                    ],
+                ),
+                ("user", "Follow-up question"),
+            ]
+        )
+
+        compacted_msgs = [
+            {"role": "user", "content": "[summary]"},
```
majdyz: 🤖 🔵 **Nit**:  uses a mock that returns 3 messages from a compressor that only received 1 message (the single user prefix). While this doesn't break the test (the assertion is on the tail, not the prefix), it makes the mock unrealistic and could confuse future readers. Consider simplifying  to just  — a 1-message prefix would be more realistic.

autogpt_platform/backend/backend/copilot/sdk/thinking_blocks_test.py
majdyz: 🤖 🔵 **Nit**: The test file module docstring (line 9) and a test comment (line 464) still use "byte-for-byte" language, which was corrected to "value-identical" in the production code docstrings in the previous commit. These should be updated for consistency.

autogpt_platform/backend/backend/copilot/sdk/thinking_blocks_test.py
majdyz: 🤖 🔵 **Nit**: The mock in this test returns 3 messages from a compressor that only received 1 message (the single user prefix). While this does not break the test (the assertion is on the tail, not the prefix), it makes the mock unrealistic and could confuse future readers. Consider simplifying compacted_msgs to just the single user message for a more realistic mock.

majdyz:
## 🤖 Round 2 Review Findings

### 🔵 Nit 1: `thinking_blocks_test.py` — stale "byte-for-byte" language (lines 9 and 464)
The test file module docstring (line 9) and a test comment (line 464) still use "byte-for-byte" language, which was corrected to "value-identical" in the production code docstrings in the previous commit. These should be updated for consistency.

### 🔵 Nit 2: `thinking_blocks_test.py` — unrealistic mock in `test_trailing_user_message_after_last_assistant` (around line 536)
The `compacted_msgs` mock returns 3 messages from a compressor that only received 1 message (the single user prefix). While this does not break the test (the assertion is on the tail, not the prefix), it makes the mock unrealistic and could confuse future readers. Consider simplifying `compacted_msgs` to just the single user message for a more realistic mock.

majdyz:
🤖 **Round 2 fixes committed in 46c582a3f:**
- Nit 1: Replaced "byte-for-byte" with "value-identical" in test file docstring (line 9) and test comment (line 464).
- Nit 2: Simplified `compacted_msgs` mock in `test_trailing_user_message_after_last_assistant` to return only the single user message that the compressor actually receives, making the mock realistic.

All 123 affected tests pass.

majdyz:
## 🤖 Round 3 Review — 0 findings

Re-read the full diff (1260 lines across 7 files) after all Round 1 and Round 2 fixes. **No new issues found.**

### Summary of the 3-round review cycle:

**Round 1** (3 findings, all addressed in `9f25e0eb4`):
- 🟡 JSON round-trip in `_rechain_tail` vs "byte-for-byte" language → tightened to "value-identical"
- 🔵 Non-dict tail line `prev_uuid` gap → added safety comment
- 🔵 Double validation between `compact_transcript` and `_reduce_context` → added authoritative-check comment

**Round 2** (2 findings, all addressed in `46c582a3f`):
- 🔵 Stale "byte-for-byte" in test file docstring/comments → updated to "value-identical"
- 🔵 Unrealistic mock in `test_trailing_user_message_after_last_assistant` → simplified to match actual compressor input

**Round 3**: 0 findings. All 123 affected tests pass. The PR is clean.

majdyz:
## Manual Test Report - PR #12574 (fix/copilot-thinking-blocks)

**Result: PASS**

### Test Summary

| Test | Status |
|------|--------|
| Login and navigate to copilot | PASS |
| Send message ("Help me build a simple agent that takes text input and outputs it reversed") | PASS |
| First response - no thinking/redacted_thinking visible | PASS |
| "Show reasoning" button present | PASS |
| Follow-up message ("Can you also add an output block to display the result?") | PASS |
| Multi-turn works without errors | PASS |
| Docker logs - no thinking block errors | PASS |
| Page text - no thinking/redacted content leaked | PASS |

### Key Findings

1. **ThinkingBlock correctly skipped during streaming** - No thinking/redacted_thinking content visible in any response
2. **"Show reasoning" button exists** (confirmed via accessibility tree and programmatic check: `reasoningButtons: 1`)
3. **Multi-turn conversations work** - Follow-up references context from first message correctly
4. **Clean logs** - No thinking-related errors or exceptions in copilot executor logs
5. **Programmatic verification** - `hasThinking: false, hasShowReasoning: true` confirmed in page DOM

### Screenshots

- [01-copilot-landing.png](https://github.com/user-attachments/assets/444a537b-0e13-4d51-8657-2545bcfbf53c) - Copilot landing page
- [19-text-reversal-chat.png](https://github.com/user-attachments/assets/7ecb136e-6358-470b-b8d7-148f4b17a0e6) - First response (Text Reverser agent created, no thinking visible)
- [21-followup-sent.png](https://github.com/user-attachments/assets/1ce7a88e-7dc1-4cc6-9346-71965069b841) - Follow-up message sent
- [22-followup-response.png](https://github.com/user-attachments/assets/07d0af12-cca6-47bb-9612-e7820b2a9c0e) - Follow-up response completed
- [25-final-chat-state.png](https://github.com/user-attachments/assets/f164dcbb-324a-475b-b9cd-531cbd9caaf6) - Final chat state

### Code Changes Verified

- `response_adapter.py`: `ThinkingBlock` handled with `pass` (lines 104-107) - silently skipped during streaming
- `transcript.py`: `_flatten_assistant_content` strips thinking/redacted_thinking blocks from older messages
- `transcript.py`: `compact_transcript` preserves last assistant entry verbatim (required for Anthropic API signature validation)
- `transcript.py`: `_find_last_assistant_entry` correctly identifies and preserves the last assistant turn including thinking blocks

ntindle (approved):

PR #12583, opened 2026-03-26
feat(platform/copilot): live timer stats with persisted duration

0ubbe:
## Why

The copilot chat had no indication of how long the AI spent "thinking" on a response. Users couldn't tell if a long wait was normal or something was stuck. Additionally, the thinking duration was lost on page reload since it was only tracked client-side.

## What

- **Live elapsed timer**: Shows elapsed time ("23s", "1m 5s") in the ThinkingIndicator while the AI is processing (appears after 20s to avoid spam on quick responses)
- **Frozen "Thought for Xm Ys"**: Displays the final thinking duration in TurnStatsBar after the response completes
- **Persisted duration**: Saves `durationMs` on the last assistant message in the DB so the timer survives page reloads

## How

**Backend:**
- Added `durationMs Int?` column to `ChatMessage` (Prisma migration)
- `mark_session_completed` in `stream_registry.py` computes wall-clock duration from Redis session `created_at` and saves it via `DatabaseManager.set_turn_duration()`
- Invalidates Redis session cache after writing so GET returns fresh data

**Frontend:**
- `useElapsedTimer` hook tracks client-side elapsed seconds during streaming
- `ThinkingIndicator` shows only the elapsed time (no phrases) after 20s, with `font-mono text-sm` styling
- `TurnStatsBar` displays "Thought for Xs" after completion, preferring live `elapsedSeconds` and falling back to persisted `durationMs`
- `convertChatSessionToUiMessages` extracts `duration_ms` from historical messages into a `Map<string, number>` threaded through to `ChatMessagesContainer`

## Test plan

- [ ] Send a message in copilot — verify ThinkingIndicator shows elapsed time after 20s
- [ ] After response completes — verify "Thought for Xs" appears below the response
- [ ] Refresh the page — verify "Thought for Xs" still appears (persisted from DB)
- [ ] Check older conversations — they should NOT show timer (no historical data)
- [ ] Verify no Zod/SSE validation errors in browser console

### DIFF
--- a/autogpt_platform/frontend/src/app/(platform)/copilot/components/ChatMessagesContainer/components/ThinkingIndicator.tsx
+++ b/autogpt_platform/frontend/src/app/(platform)/copilot/components/ChatMessagesContainer/components/ThinkingIndicator.tsx
@@ -95,7 +95,9 @@ export function ThinkingIndicator({ active, elapsedSeconds }: Props) {
         </span>
       </span>
       {showTime && (
-        <span className="tabular-nums">• {formatElapsed(elapsedSeconds)}</span>
+        <span className="animate-pulse tabular-nums [animation-duration:1.5s]">
+          • {formatElapsed(elapsedSeconds)}
+        </span>
       )}
     </span>
   );


autogpt_platform/backend/backend/copilot/stream_registry.py
```diff
@@ -111,6 +111,14 @@ def _parse_session_meta(meta: dict[Any, Any], session_id: str = "") -> ActiveSes
     ``session_id`` is used as a fallback for ``turn_id`` when the meta hash
     pre-dates the turn_id field (backward compat for in-flight sessions).
     """
+    created_at = datetime.now(timezone.utc)
+    created_at_raw = meta.get("created_at")
+    if created_at_raw:
+        try:
+            created_at = datetime.fromisoformat(str(created_at_raw))
+        except (ValueError, TypeError):
+            pass
```
majdyz, line 121: 🤖 🟠 **Should Fix**: `created_at` fallback to `datetime.now(UTC)` silently stores `durationMs≈0` for pre-existing sessions.

When `created_at` is missing from Redis meta (sessions started before this change), the fallback produces `now() - now() ≈ 0ms`. The UI won't show it (the `displaySeconds > 0` guard handles it), but the DB stores `durationMs=0` which is inaccurate data — it's indistinguishable from a genuinely instant response.

Suggested fix: skip `set_turn_duration` entirely when `created_at` wasn't present in the original meta:
```python
created_at_raw = meta.get("created_at")
if created_at_raw:
    # ... compute and persist duration
```
Or store `None` instead of `0` for unknown durations.

autogpt_platform/backend/backend/copilot/db.py
```diff
@@ -359,3 +362,24 @@ async def update_tool_message_content(
...
+
+
+async def set_turn_duration(session_id: str, duration_ms: int) -> None:
+    """Set durationMs on the last assistant message in a session.
+
+    Also invalidates the Redis session cache so the next GET returns
+    the updated duration.
+    """
+    last_msg = await PrismaChatMessage.prisma().find_first(
+        where={"sessionId": session_id, "role": "assistant"},
```
majdyz, line 374: 🤖 🟡 **Nice to Have**: Two DB roundtrips (`find_first` + `update`) for a simple "update last assistant message" operation.

Could be reduced to one with `update_many`:
```python
await PrismaChatMessage.prisma().update_many(
    where={
        "sessionId": session_id,
        "role": "assistant",
        "sequence": last_sequence,  # need to know this
    },
    data={"durationMs": duration_ms},
)
```

Since this only runs once per turn completion it's not a hot path, so acceptable to keep as-is.

autogpt_platform/backend/backend/copilot/model.py
```diff
@@ -561,6 +563,7 @@ async def _save_session_to_db(
                     "refusal": msg.refusal,
                     "tool_calls": msg.tool_calls,
                     "function_call": msg.function_call,
+                    "duration_ms": msg.duration_ms,
```
majdyz: 🤖 🟠 **Should Fix**: `duration_ms` is passed to `_save_session_to_db` but the in-memory model won't have the computed duration at save time.

The flow is:
1. `add_chat_messages_batch` creates messages during streaming → `duration_ms=None`
2. Stream completes → `set_turn_duration` updates the last assistant message with the correct duration
3. `invalidate_session_cache` clears Redis cache ✅

But if `_save_session_to_db` runs _after_ `set_turn_duration` (e.g., on a delayed cache flush), it writes `duration_ms: None` back, potentially overwriting the correct value. Is there a guarantee on ordering here? If not, either:
- Remove `duration_ms` from `_save_session_to_db` entirely (since `set_turn_duration` is the canonical writer)
- Or add a `WHERE durationMs IS NULL` guard to the `set_turn_duration` update

autogpt_platform/frontend/src/app/(platform)/copilot/helpers/convertChatSessionToUiMessages.ts
```diff
@@ -102,7 +105,10 @@ export function convertChatSessionMessagesToUiMessages(
   sessionId: string,
   rawMessages: unknown[],
   options?: { isComplete?: boolean },
-): UIMessage<unknown, UIDataTypes, UITools>[] {
+): {
```
majdyz, line 108: 🤖 🟠 **Should Fix**: Breaking return type change will conflict with PR #12328.

The return type changed from `UIMessage[]` to `{ messages, durations }`. PR #12328 also modifies this function (adds `extraToolOutputs` parameter) and calls it from `useLoadMoreMessages`. The merge will need careful resolution — both PRs change the function signature and all call sites.

Make sure to coordinate merge order and verify the combined behavior works correctly, particularly `useLoadMoreMessages` which also needs to handle `durations` from paginated pages.

autogpt_platform/frontend/src/app/(platform)/copilot/components/ChatMessagesContainer/ChatMessagesContainer.tsx
```diff
@@ -139,6 +142,23 @@ export function ChatMessagesContainer({
...
+  // Freeze elapsed time when streaming ends so TurnStatsBar shows the final value.
+  // Reset when a new streaming turn begins.
+  const frozenElapsedRef = useRef(0);
+  const wasStreamingRef = useRef(false);
+  if (isActivelyStreaming) {
+    if (!wasStreamingRef.current) {
+      frozenElapsedRef.current = 0;
+    }
+    if (elapsedSeconds > 0) {
+      frozenElapsedRef.current = elapsedSeconds;
```
majdyz: 🤖 🟡 **Nice to Have**: Ref mutations during the render phase (`frozenElapsedRef.current = ...`, `wasStreamingRef.current = ...`) is a React anti-pattern.

In concurrent mode / Strict Mode, the render function can be called multiple times before commit. Mutating refs during render means the mutation can be applied multiple times.

For simple counters like this the practical impact is negligible, but the idiomatic pattern is to do this in a `useEffect`:
```tsx
useEffect(() => {
  if (isActivelyStreaming && elapsedSeconds > 0) {
    frozenElapsedRef.current = elapsedSeconds;
  }
  wasStreamingRef.current = isActivelyStreaming;
});
```

autogpt_platform/backend/backend/copilot/stream_registry.py
```diff
@@ -802,6 +811,22 @@ async def mark_session_completed(
...
 
+    # Compute wall-clock duration from session created_at
+    duration_ms: int | None = None
+    if meta:
+        parsed_meta = _parse_session_meta(meta, session_id)
+        elapsed = datetime.now(timezone.utc) - parsed_meta.created_at
+        duration_ms = max(0, int(elapsed.total_seconds() * 1000))
+
+    # Persist duration on the last assistant message
+    if duration_ms is not None:
```
majdyz, line 836: 🤖 🟠 **Should Fix**: The duration computation should check whether `created_at` was actually present in the Redis meta, not rely on the parsed fallback.

Currently, `_parse_session_meta` always returns a `created_at` (defaulting to `now()` when missing), so this code always computes a duration even when it's meaningless. The guard should be:

```python
created_at_raw = meta.get("created_at") if meta else None
if created_at_raw:
    try:
        created_at = datetime.fromisoformat(str(created_at_raw))
        duration_ms = max(0, int((datetime.now(timezone.utc) - created_at).total_seconds() * 1000))
    except (ValueError, TypeError):
        duration_ms = None
else:
    duration_ms = None
```

This avoids storing `durationMs=0` for sessions that predate the feature.

autogpt_platform/backend/backend/copilot/db.py
```diff
@@ -359,3 +362,24 @@ async def update_tool_message_content(
...
+        order={"sequence": "desc"},
+    )
+    if last_msg:
+        await PrismaChatMessage.prisma().update(
+            where={"id": last_msg.id},
+            data={"durationMs": duration_ms},
+        )
+        # Invalidate cache so the session is re-fetched from DB with durationMs
+        from backend.copilot.model import invalidate_session_cache
```
majdyz: 🤖 🔵 **Nit**: Local import to avoid circular dependency — same pattern noted in PR #12328. Consider extracting `invalidate_session_cache` to a shared utility module.

autogpt_platform/backend/backend/copilot/stream_registry.py
```diff
@@ -802,6 +811,22 @@ async def mark_session_completed(
...
+        try:
+            from backend.data.db_accessors import chat_db
+
```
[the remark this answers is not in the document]
0ubbe: 🤖 Fixed in ea4659624: Duration is now only persisted when the session completes successfully (`not error_message` guard added).

autogpt_platform/backend/backend/copilot/stream_registry.py
```diff
@@ -802,6 +811,22 @@ async def mark_session_completed(
                 f"Failed to publish error event for session {session_id}: {e}"
             )
 
+    # Compute wall-clock duration from session created_at
+    duration_ms: int | None = None
+    if meta:
+        parsed_meta = _parse_session_meta(meta, session_id)
+        elapsed = datetime.now(timezone.utc) - parsed_meta.created_at
+        duration_ms = max(0, int(elapsed.total_seconds() * 1000))
```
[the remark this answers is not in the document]
0ubbe: 🤖 Fixed in ea4659624: Now normalizes naive datetimes to UTC and wraps the subtraction in try/except.

autogpt_platform/backend/backend/copilot/stream_registry.py
```diff
@@ -802,6 +811,22 @@ async def mark_session_completed(
...
+
+    # Persist duration on the last assistant message
+    if duration_ms is not None:
+        try:
+            from backend.data.db_accessors import chat_db
+
+            await chat_db().set_turn_duration(session_id, duration_ms)
```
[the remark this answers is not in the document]
0ubbe: 🤖 Addressed in ea4659624: The duration computation now parses `created_at` directly from the raw Redis meta (not the fallback), so it's always scoped to the current turn's start time. Also removed `duration_ms` from `_save_session_to_db` so only `set_turn_duration` is the canonical writer.

autogpt_platform/backend/backend/copilot/db.py
```diff
@@ -359,3 +362,24 @@ async def update_tool_message_content(
...
+
```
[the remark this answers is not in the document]
0ubbe: 🤖 Fixed in ea4659624: Moved `invalidate_session_cache` import to top-level in db.py.

autogpt_platform/backend/backend/copilot/stream_registry.py
```diff
@@ -802,6 +811,22 @@ async def mark_session_completed(
...
+        except Exception as e:
+            logger.warning(f"Failed to save turn duration for {session_id}: {e}")
```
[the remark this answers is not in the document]
0ubbe: 🤖 Fixed in ea4659624: Moved `chat_db` import to top-level in stream_registry.py.

autogpt_platform/backend/backend/copilot/stream_registry.py
0ubbe: 🤖 Fixed in ea4659624: Now checks `meta.get("created_at")` directly. When `created_at` is missing from Redis meta, `duration_ms` stays `None` and `set_turn_duration` is skipped entirely.

autogpt_platform/backend/backend/copilot/db.py
0ubbe: 🤖 Agreed this is acceptable as-is since it only runs once per turn completion. Will keep as a future optimization opportunity.

autogpt_platform/backend/backend/copilot/model.py
0ubbe: 🤖 Fixed in ea4659624: Removed `duration_ms` from `_save_session_to_db` entirely. `set_turn_duration` is now the sole writer, so a delayed cache flush can no longer overwrite the correct value with `None`.

autogpt_platform/frontend/src/app/(platform)/copilot/helpers/convertChatSessionToUiMessages.ts
0ubbe: 🤖 Acknowledged. Will coordinate merge order with PR #12328 and ensure `useLoadMoreMessages` handles the `{ messages, durations }` return type correctly when both PRs land.

autogpt_platform/frontend/src/app/(platform)/copilot/components/ChatMessagesContainer/ChatMessagesContainer.tsx
0ubbe: 🤖 Fixed in ea4659624: Moved ref mutations into a `useEffect` to be safe in concurrent mode / Strict Mode.

autogpt_platform/backend/backend/copilot/stream_registry.py
0ubbe: 🤖 Fixed in ea4659624: Now reads `created_at_raw = meta.get("created_at")` directly from the Redis hash. When missing, `duration_ms` stays `None` and persistence is skipped. Also handles naive datetime and wraps in try/except.

autogpt_platform/backend/backend/copilot/db.py
0ubbe: 🤖 Fixed in ea4659624: Moved `invalidate_session_cache` import to top-level. Noted the shared utility extraction idea for future work.

majdyz:
## 🤖 PR Review — #12583

**Live timer stats — Shows elapsed thinking time with persistence. Good UX improvement.**

CI: 0 failures | Diff: ~591 lines

Review in progress — will post inline findings if any.

majdyz:
## 🤖 PR Review — #12583 (Live Timer Stats)

**Verdict: ✅ APPROVED**

Clean implementation. DB migration adds duration_ms to ChatMessage. ThinkingIndicator component shows elapsed time. Stream registry tracks start time. Good separation: backend persists, frontend displays.

**Findings:** No blockers.

All CI checks green.

autogpt_platform/backend/backend/copilot/db.py
```diff
@@ -359,3 +362,22 @@ async def update_tool_message_content(
             f"tool_call_id {tool_call_id}: {e}"
         )
```
majdyz, line 363: 🤖 🔵 **Nit**: Clean separation — `set_turn_duration` updates only the last assistant message. Good use of `find_first` with `order=desc`.

autogpt_platform/backend/backend/copilot/model.py
```diff
@@ -54,6 +54,7 @@ class ChatMessage(BaseModel):
     refusal: str | None = None
     tool_calls: list[dict] | None = None
     function_call: dict | None = None
+    duration_ms: int | None = None
```
majdyz, line 58: 🤖 🟡 **Nice to Have**: `duration_ms: int | None = None` — consider adding a brief docstring noting that `None` means "not tracked" (pre-migration messages) vs `0` meaning instant.

PR #12584, opened 2026-03-26
feat(backend): support dedicated OpenRouter API key for autopilot

majdyz:
## Why
We need the ability to use a separate OpenRouter API key for CoPilot/AutoPilot workloads, allowing separate billing and rate limits from the shared key.

## What
Add `AUTOPILOT_OPEN_ROUTER_API_KEY` env var. When set, it is prioritized over `OPEN_ROUTER_API_KEY` for all SDK calls (CoPilot + AutoPilot). Falls back to the regular key when not configured.

## How
- **`settings.py`**: New `autopilot_open_router_api_key` secret field
- **`config.py`**: `autopilot_api_key` property returns dedicated key or falls back to `api_key`
- **`service.py`**: `_build_sdk_env` and fail-fast check use `config.autopilot_api_key` instead of `config.api_key`

## Test plan
- [ ] Set only `OPEN_ROUTER_API_KEY` → verify CoPilot/AutoPilot works as before
- [ ] Set both keys → verify `AUTOPILOT_OPEN_ROUTER_API_KEY` is used for SDK calls
- [ ] Set only `AUTOPILOT_OPEN_ROUTER_API_KEY` → verify it activates OpenRouter mode
- [ ] Set neither → verify appropriate error message mentioning both env vars

### DIFF
--- a/autogpt_platform/backend/backend/copilot/config.py
+++ b/autogpt_platform/backend/backend/copilot/config.py
@@ -3,11 +3,10 @@
 import os
 from typing import Literal
 
-from pydantic import Field, field_validator
+from pydantic import Field, field_validator, model_validator
 from pydantic_settings import BaseSettings
 
 from backend.util.clients import OPENROUTER_BASE_URL
-from backend.util.settings import Settings
 
 
 class ChatConfig(BaseSettings):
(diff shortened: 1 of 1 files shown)


autogpt_platform/backend/backend/copilot/config.py
```diff
@@ -171,7 +172,12 @@ def openrouter_active(self) -> bool:
         base = (self.base_url or "").rstrip("/")
         if base.endswith("/v1"):
             base = base[:-3]
-        return bool(self.api_key and base and base.startswith("http"))
+        return bool(self.autopilot_api_key and base and base.startswith("http"))
```
majdyz, line 179: 🤖 🔵 **Nit**: The `openrouter_active` docstring (line 167) still references `api_key` but the code now checks `autopilot_api_key`. Should update to match.

PR #12585, opened 2026-03-26
test(copilot): add edit_agent tool tests with xfail for version conflict bug

Otto-AGPT:
Requested by @ntindle

Add comprehensive tests for `EditAgentTool` covering input validation, preview mode, save behavior, and a bug reproduction test.

## Why

The `edit_agent` copilot tool had zero test coverage. Toran reported that AutoPilot fails to save edits to agents with `Unique constraint failed on the fields: (id, version)` (Sentry: AUTOGPT-SERVER-8BZ/8C0). We need tests to both cover existing behavior and confirm the bug for tracking.

## What

New file: `backend/copilot/tools/edit_agent_test.py` with 10 tests:

**Passing tests (9):**
- Missing agent_id/agent_json → ErrorResponse
- Empty nodes → ErrorResponse
- Agent not found → ErrorResponse
- Preview mode (save=False) → AgentPreviewResponse
- No auth → login error
- Save passes `is_update=True` and preserves original graph ID
- Save preserves original version from looked-up agent
- Save failure → ErrorResponse with details

**xfail test (1):**
- `test_edit_agent_version_conflict_when_user_not_graph_owner` — reproduces the production bug where `update_graph_in_library` calls `get_graph_all_versions(graph_id, user_id)` with a userId filter. When the editing user doesn't own the graph, the lookup returns empty, version defaults to 1, and Prisma raises `UniqueViolationError` on the `(id, version)` constraint.

## Sentry Analysis

- **Graph:** `6b69f689-a552-4b53-b379-0f249eff2c4d` (Citedy Topic to Publish)
- **3 occurrences** in ~1 minute (copilot retried)
- Stack: `pipeline.py → core.py → RPC → library/db.py → graph.py:create_graph → UniqueViolationError`
- Breadcrumbs show: transaction/start → query → create_many (fails) → rollback

Linear: SECRT-2175

### DIFF
--- a/autogpt_platform/backend/backend/copilot/tools/edit_agent_test.py
+++ b/autogpt_platform/backend/backend/copilot/tools/edit_agent_test.py
@@ -1,4 +1,11 @@
-"""Tests for EditAgentTool."""
+"""Tests for EditAgentTool.
+
+Covers:
+- EditAgentTool._execute input validation and pipeline integration
+- get_agent_as_json lookup paths (graph ID, library agent ID, not found)
+- update_graph_in_library version/ownership matrix
+- save_agent_to_library routing (is_update=True vs False)
+"""
 
 from unittest.mock import AsyncMock, MagicMock, patch
 
@@ -14,8 +21,11 @@ from backend.copilot.tools.models import (
 from ._test_data import make_session
 
 _TEST_USER_ID = "test-user-edit-agent"
+_OTHER_USER_ID = "other-user-owner"
 _PIPELINE = "backend.copilot.tools.agent_generator.pipeline"
 _EDIT_AGENT = "backend.copilot.tools.edit_agent"
+_LIBRARY_DB = "backend.api.features.library.db"
+_GRAPH_DB = f"{_LIBRARY_DB}.graph_db"
 
 
 def _make_agent_json(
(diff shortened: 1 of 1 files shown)


majdyz (changes requested):
This doesn't help then? the xfail should fail, no ? :p

autogpt_platform/backend/backend/copilot/tools/edit_agent_test.py
```diff
@@ -0,0 +1,979 @@
...
+    # Verify save_agent_to_library was called with is_update=True
+    mock_save.assert_called_once()
+    call_kwargs = mock_save.call_args
+    assert call_kwargs[1].get("is_update") or call_kwargs[0][2] is True
+    # The agent_json passed to save should have the original graph ID
+    saved_json = call_kwargs[0][0] if call_kwargs[0] else call_kwargs[1]["agent_json"]
+    assert saved_json["id"] == "original-graph-id"
+
+
+@pytest.mark.asyncio
```
majdyz, line 257: 🤖 🟠 **Should Fix**: This assertion is fragile and subtly misleading.

```python
assert call_kwargs[1].get("is_update") or call_kwargs[0][2] is True
```

Due to Python operator precedence, `call_kwargs[0][2] is True` evaluates first (identity check), yielding a bool. Then it becomes `call_kwargs[1].get("is_update") or <bool>`. If the dict lookup returns `None`, the whole expression reduces to the `is True` check — but if positional arg at index 2 happens to be `True` (not just truthy), the assertion passes silently even if the keyword form wasn't used.

Since `save_agent_to_library` is always called with `is_update` as a keyword argument from `pipeline.py`, simplify to:

```python
mock_save.assert_called_once()
assert mock_save.call_args.kwargs["is_update"] is True
saved_json = mock_save.call_args.args[0]
assert saved_json["id"] == "original-graph-id"
```

Same cleanup applies to the `saved_json` extraction on line 259: `call_args[0][0] if call_args[0] else call_args[1]["agent_json"]` is overly defensive.

autogpt_platform/backend/backend/copilot/tools/edit_agent_test.py
```diff
@@ -0,0 +1,979 @@
...
+    strict=True,
+)
+@pytest.mark.asyncio
+async def test_edit_agent_version_conflict_when_user_not_graph_owner(tool, session):
+    """Editing an agent whose graph is owned by a different user should
+    still succeed (version should auto-increment correctly).
+
+    Currently fails because update_graph_in_library calls
+    get_graph_all_versions(graph_id, user_id) which filters by userId.
+    If the editing user is not the graph owner (e.g. store-forked agent),
```
majdyz, line 364: 🤖 🟠 **Should Fix**: Move all local imports to the top of the file.

There are 10+ local imports scattered throughout the test functions (lines 364, 475, 544, 629, 670, 713, 737, 769, 792, 818, 835, 856). Per `backend/CLAUDE.md`:

> **Top-level imports only** — no local/inner imports (lazy imports only for heavy optional deps like `openpyxl`)

These are all standard project imports, not heavy optional dependencies. Move them to the module-level import block:

```python
from backend.api.features.library.db import update_graph_in_library
from backend.copilot.tools.agent_generator.core import (
    get_agent_as_json,
    save_agent_to_library,
)
from backend.data.graph import Graph, Node
from backend.util.exceptions import NotFoundError
```

autogpt_platform/backend/backend/copilot/tools/edit_agent_test.py
```diff
@@ -0,0 +1,979 @@
...
+    model.version = graph.version
+    model.is_active = graph.is_active
+    # Allow version to be set and read back
+    type(model).version = property(
+        lambda self: self._version, lambda self, v: setattr(self, "_version", v)
+    )
+    model._version = graph.version
+    model.reassign_ids = MagicMock()
+    return model
+
```
majdyz, line 506: 🤖 🟡 **Nice to Have**: The property-descriptor trick for version tracking is overly complex.

```python
type(model).version = property(
    lambda self: self._version, lambda self, v: setattr(self, "_version", v)
)
model._version = graph.version
```

In all tests that use this helper (`_run_update` and friends), `update_graph_in_library` sets `graph.version` — it never mutates `graph_model.version`. The mock's version is read but never written. A simple attribute assignment would be sufficient:

```python
model.version = graph.version
```

If you genuinely need mutable version tracking on the mock, use `PropertyMock` instead of raw descriptors — it's the standard unittest.mock way and much more readable:

```python
type(model).version = PropertyMock(return_value=graph.version)
```

autogpt_platform/backend/backend/copilot/tools/edit_agent_test.py
```diff
@@ -0,0 +1,979 @@
...
+    existing_agent = _make_agent_json()
+    updated_json = _make_agent_json(name="Updated Agent")
+
+    mock_fixer = MagicMock()
+    mock_fixer.apply_all_fixes = MagicMock(return_value=updated_json)
+    mock_fixer.get_fixes_applied.return_value = []
+
+    mock_validator = MagicMock()
+    mock_validator.validate.return_value = (True, None)
+    mock_validator.errors = []
```
majdyz, line 138: 🤖 🟠 **Should Fix**: The fixer/validator mock setup is duplicated verbatim across 8 tests. Extract into fixtures.

This exact block appears in `test_preview_mode_returns_preview`, `test_no_auth_returns_error`, `test_save_calls_pipeline_with_is_update_true`, `test_save_preserves_original_version`, `test_save_failure_returns_error`, `test_edit_sets_is_active_default`, `test_edit_validation_failure_returns_error`, and more:

```python
mock_fixer = MagicMock()
mock_fixer.apply_all_fixes = MagicMock(return_value=updated_json)
mock_fixer.get_fixes_applied.return_value = []

mock_validator = MagicMock()
mock_validator.validate.return_value = (True, None)
mock_validator.errors = []
```

This is ~150 lines of pure duplication. Extract as fixtures:

```python
@pytest.fixture
def mock_fixer():
    fixer = MagicMock()
    fixer.apply_all_fixes = MagicMock(side_effect=lambda json, *a, **kw: json)
    fixer.get_fixes_applied.return_value = []
    return fixer

@pytest.fixture
def mock_validator():
    validator = MagicMock()
    validator.validate.return_value = (True, None)
    validator.errors = []
    return validator
```

This also makes the fixer's `return_value` consistently pass-through (return the input), which is more realistic than hardcoding a separate `updated_json` dict.

File is 979 lines for 10 tests. Per CLAUDE.md guidelines, files should stay under ~300 lines. This duplication is the primary reason for the bloat.

autogpt_platform/backend/backend/copilot/tools/edit_agent_test.py
```diff
@@ -0,0 +1,979 @@
...
+
+
+def _make_graph_model_mock(graph):
+    """Create a MagicMock that behaves like a GraphModel for the given Graph.
+
+    GraphModel has reassign_ids() which Graph does not, so we need a mock
+    that tracks version mutations while providing that method.
+    """
+    model = MagicMock()
+    model.id = graph.id
```
majdyz, line 496: 🤖 🟠 **Should Fix**: The docstring claims "GraphModel has reassign_ids() which Graph does not" -- but this is incorrect. `reassign_ids` is defined on `Graph` (graph.py line 618), and `GraphModel` inherits from `Graph`. Both have the method.

The misleading docstring suggests a misunderstanding of the class hierarchy. Since `Graph` already has `reassign_ids`, the xfail test (line 393) can pass a `Graph` directly to `make_graph_model` return value and have `reassign_ids` work without mocking. But for the `_make_graph_model_mock` helper that's used by the `_run_update` tests, the mock IS needed because the mock provides version-tracking behavior.

Fix the docstring to accurately state why a mock is needed:

```python
def _make_graph_model_mock(graph):
    """Create a MagicMock that behaves like a GraphModel for the given Graph.

    We need a mock (instead of using Graph directly) so we can verify
    that create_graph receives the expected attributes without side effects
    from reassign_ids mutating real node/link IDs.
    """
```

autogpt_platform/backend/backend/copilot/tools/edit_agent_test.py
```diff
@@ -0,0 +1,979 @@
...
+
+    # Mock get_graph_all_versions to return empty (simulates userId mismatch)
+    # and create_graph to raise the unique constraint violation
+    with (
+        patch(
+            "backend.api.features.library.db.graph_db.get_graph_all_versions",
+            new_callable=AsyncMock,
+            return_value=[],  # Empty: user doesn't own this graph
+        ),
+        patch(
```
majdyz, line 402: 🤖 🟡 **Nice to Have**: The xfail test uses nested `with` context managers (two levels of `with (patch(...), ...):` stacked inside each other). This makes the test hard to read and violates the "early return / avoid deep nesting" principle. Flatten into a single `with` block:

```python
with (
    patch("...graph_db.get_graph_all_versions", new_callable=AsyncMock, return_value=[]),
    patch("...graph_db.make_graph_model", return_value=graph),
    patch("...graph_db.create_graph", new_callable=AsyncMock) as mock_create,
    patch("...get_library_agent_by_graph_id", new_callable=AsyncMock, return_value=MagicMock(id="lib-1")),
    patch("...update_library_agent_version_and_settings", new_callable=AsyncMock, return_value=MagicMock(id="lib-1")),
    patch("...on_graph_activate", new_callable=AsyncMock, return_value=graph),
    patch("...graph_db.set_graph_active_version", new_callable=AsyncMock),
):
    mock_create.return_value = graph
    await update_graph_in_library(graph, _TEST_USER_ID)
    ...
```

Python 3.10+ parenthesized context managers support this cleanly.

autogpt_platform/backend/backend/copilot/tools/edit_agent_test.py
```diff
@@ -0,0 +1,979 @@
...
+from backend.copilot.tools.models import (
+    AgentPreviewResponse,
+    AgentSavedResponse,
+    ErrorResponse,
+)
+
+from ._test_data import make_session
+
+_TEST_USER_ID = "test-user-edit-agent"
+_OTHER_USER_ID = "other-user-owner"
```
majdyz, line 24: 🤖 🟡 **Nice to Have**: `_OTHER_USER_ID = "other-user-owner"` is defined at module scope but never used anywhere in the file. Remove it to keep the module clean.

autogpt_platform/backend/backend/copilot/tools/edit_agent_test.py
```diff
@@ -0,0 +1,979 @@
...
+        # on_graph_activate needs to return the created graph
+        mocks[f"{_LIBRARY_DB}.on_graph_activate"].return_value = created_mock
+
+        patches = {k: patch(k, v) for k, v in mocks.items()}
+        started = {k: p.start() for k, p in patches.items()}
+        try:
+            await update_graph_in_library(graph, user_id)
+        finally:
+            for p in patches.values():
+                p.stop()
```
majdyz, line 567: 🤖 🟠 **Should Fix**: The `_run_update` helper uses `patch(...).start()` / `patch(...).stop()` pattern manually instead of context managers. This is fragile -- if an exception occurs between `start()` and the `try` block, patches may leak. More importantly, the `deactivate_called` check on line 577 reads from a mock reference obtained inside the started patches, but the assertion happens after `stop()`. While this technically works (the mock object survives `stop()`), it's confusing to read.

Use context managers instead:

```python
async def _run_update(graph, existing_versions, user_id=_TEST_USER_ID):
    graph_model = _make_graph_model_mock(graph)
    created_mock = MagicMock(id=graph.id, is_active=graph.is_active)

    with (
        patch(f"{_GRAPH_DB}.get_graph_all_versions", new_callable=AsyncMock, return_value=existing_versions),
        patch(f"{_GRAPH_DB}.make_graph_model", return_value=graph_model),
        patch(f"{_GRAPH_DB}.create_graph", new_callable=AsyncMock, return_value=created_mock) as mock_create,
        patch(f"{_LIBRARY_DB}.get_library_agent_by_graph_id", new_callable=AsyncMock, return_value=MagicMock(id="lib-1")),
        patch(f"{_LIBRARY_DB}.update_library_agent_version_and_settings", new_callable=AsyncMock, return_value=MagicMock(id="lib-1")),
        patch(f"{_LIBRARY_DB}.on_graph_activate", new_callable=AsyncMock, return_value=created_mock),
        patch(f"{_LIBRARY_DB}.on_graph_deactivate", new_callable=AsyncMock) as mock_deactivate,
        patch(f"{_GRAPH_DB}.set_graph_active_version", new_callable=AsyncMock),
    ):
        await update_graph_in_library(graph, user_id)
        return graph.version, mock_deactivate.called
```

majdyz (changes requested):
## Overall Assessment

Good test coverage addition for `EditAgentTool`, and the xfail bug reproduction for SECRT-2175 is well-motivated. The PR description is excellent with clear Why/What/How and Sentry analysis. However, the implementation has significant code quality issues that need addressing before merge.

### Summary of Findings

**Should Fix (4):**
1. **Fragile `is_update` assertion** (line 257) — uses `or` + `is True` identity check in a misleading way. Simplify to `call_args.kwargs["is_update"] is True`.
2. **10+ local imports** scattered throughout test functions — violates the top-level imports coding standard.
3. **Massive mock duplication** — fixer/validator setup is copy-pasted across 8 tests, inflating the file to 979 lines (guideline: ~300). Extract into fixtures.
4. **Manual `patch.start()`/`stop()` in `_run_update`** — fragile and harder to read than context managers. Use a single `with (...)` block.
5. **Incorrect docstring on `_make_graph_model_mock`** — claims "GraphModel has reassign_ids() which Graph does not" but `reassign_ids` is defined on `Graph`.

**Nice to Have (3):**
6. **Overly complex property descriptor** in `_make_graph_model_mock` — version is never mutated on the mock, so a simple attribute suffices.
7. **Nested `with` blocks** in xfail test — flatten into a single parenthesized context manager.
8. **Unused `_OTHER_USER_ID`** constant.

### What Looks Good
- The xfail test correctly reproduces the SECRT-2175 bug with `strict=True` — this will force removal when the bug is fixed.
- Mock targets are correct (patching where symbols are **used**, not defined).
- The `get_agent_as_json` and `save_agent_to_library` routing tests cover important lookup paths.
- Edge cases like whitespace `agent_id`, missing `is_active`, and validation failure are covered.

### Request
Please run the tests (`poetry run pytest backend/copilot/tools/edit_agent_test.py -xvs`) and share the output confirming all 9 tests pass and the 1 xfail behaves as expected. After addressing the feedback, the test count per the description should remain at 10.

majdyz:

PR #12587, opened 2026-03-26
fix(docs): remove extraneous whitespace in README

carsonkahn-external:
### Why / What / How

Remove extraneous whitespace in README.md:
- "Workflow Management" description: extra spaces between "block" and "performs"
- "Agent Interaction" description: extra spaces between "user-friendly" and "interface"


majdyz:
Thanks for the fix! There's also a similar extraneous whitespace issue on line 86:

```
**Workflow Management:** Build, modify, and optimize your automation workflows with ease. You build your agent by connecting blocks, where each block     performs a single action.
```

There are extra spaces between "block" and "performs". Could you fix that one too while you're at it?

majdyz (approved):

README.md
```diff
@@ -83,13 +83,13 @@ The AutoGPT frontend is where users interact with our powerful AI automation pla
 
    **Agent Builder:** For those who want to customize, our intuitive, low-code interface allows you to design and configure your own AI agents.
```
majdyz, line 84: 🤖 🟡 **Nice to Have**: While you're fixing whitespace in this file, there are also trailing spaces on the immediately adjacent line 84 (the "Agent Builder" line ends with a trailing space). Consider cleaning those up too for consistency, since you're already touching this block of text. Not a blocker — the current fix is correct.

majdyz:

PR #12588, opened 2026-03-26
fix(blocks): prevent TypeError when auto_credentials field is empty

ntindle:
### Why / What / How

**Why:** `GoogleSheetsReadBlock` crashes with `BlockUnknownError` wrapping a `TypeError: missing 1 required keyword-only argument: 'credentials'` when the block is executed without a spreadsheet selected. Instead of a clean user-facing error, users see an opaque internal error.

**What:** Fix the auto_credentials injection path so blocks using `GoogleDriveFileField` don't crash when their file input is empty.

**How:** Two-layer fix:
1. **Systemic (`_base.py`):** In `_execute()`, inject `None` for any auto_credentials kwargs not already present in `kwargs` via `setdefault`. This prevents `TypeError` for *all* blocks using `GoogleDriveFileField`.
2. **Block-specific (`sheets.py`):** Make `GoogleSheetsReadBlock.run()` accept `credentials: GoogleCredentials | None = None` and add a guard that yields a clean error when credentials are missing.

### Changes 🏗️

- `backend/blocks/_base.py`: Added auto_credentials kwargs defaulting in `_execute()` before calling `self.run()`
- `backend/blocks/google/sheets.py`: Made `GoogleSheetsReadBlock.run()` credentials parameter optional with a None guard
- `backend/blocks/google/sheets_test.py`: New test verifying clean error instead of TypeError

#### For code changes:
- [x] I have made a test plan
  - [x] New test `test_sheets_read_no_spreadsheet_yields_clean_error` passes
  - [x] Existing `test_available_blocks[GoogleSheetsReadBlock]` still passes
  - [x] `poetry run format` and `poetry run lint` clean
  - [x] `pyright` reports 0 errors on changed files

---

> [!NOTE]
> **Medium Risk**
> Touches the core block execution path by injecting default `None` values for auto-credential kwargs, which could subtly affect any block relying on kwargs presence/absence. Behavior change is guarded and includes a targeted Google Sheets regression test, but warrants review across other auto-credentialed blocks.
> 
> **Overview**
> Prevents blocks that use *auto-resolved credentials* (e.g. via `GoogleDriveFileField`) from crashing with a `TypeError` when the related input is empty by defaulting any missing auto-credentials kwargs to `None` in `Block._execute()`.
> 
> Updates `GoogleSheetsReadBlock` to accept optional `credentials` and emit a clean user-facing error when credentials are missing, and adds a regression test ensuring executing without a spreadsheet yields `BlockExecutionError` rather than an unknown/internal error.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 001ef93b373b5896df76fbb1c1322a2d0de08698. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/autogpt_platform/backend/backend/blocks/_base.py
+++ b/autogpt_platform/backend/backend/blocks/_base.py
@@ -706,6 +706,13 @@ class Block(ABC, Generic[BlockSchemaInputType, BlockSchemaOutputType]):
                 block_id=self.id,
             )
 
+        # Ensure auto_credentials kwargs are present (even as None) so blocks
+        # that declare e.g. `*, credentials: GoogleCredentials` as a required
+        # kwarg don't crash with TypeError when the execution context didn't
+        # resolve credentials (e.g. the GoogleDriveFile field was empty).
+        for kwarg_name in self.input_schema.get_auto_credentials_fields():
+            kwargs.setdefault(kwarg_name, None)
+
         # Use the validated input data
         async for output_name, output_data in self.run(
             self.input_schema(**{k: v for k, v in input_data.items() if v is not None}),
--- a/autogpt_platform/backend/backend/blocks/google/sheets.py
+++ b/autogpt_platform/backend/backend/blocks/google/sheets.py
@@ -326,10 +326,17 @@ class GoogleSheetsReadBlock(Block):
         )
 
     async def run(
-        self, input_data: Input, *, credentials: GoogleCredentials, **kwargs
+        self,
+        input_data: Input,
+        *,
+        credentials: GoogleCredentials | None = None,
+        **kwargs,
     ) -> BlockOutput:
         if not input_data.spreadsheet:
             yield "error", "No spreadsheet selected"
+            return
+        if not credentials:
+            yield "error", "Google credentials are required"
             return
 
         # Check if the selected file is actually a Google Sheets spreadsheet
--- a/autogpt_platform/backend/backend/blocks/google/sheets_test.py
+++ b/autogpt_platform/backend/backend/blocks/google/sheets_test.py
@@ -1,0 +1,18 @@
+"""Tests for Google Sheets blocks — edge cases around credentials handling."""
+
+import pytest
+
+from backend.blocks.google.sheets import GoogleSheetsReadBlock
+from backend.util.exceptions import BlockExecutionError
+
+
+async def test_sheets_read_no_spreadsheet_yields_clean_error():
+    """Executing GoogleSheetsReadBlock with no spreadsheet/credentials should
+    raise a clean BlockExecutionError, not a BlockUnknownError wrapping a
+    TypeError about a missing kwarg."""
+    block = GoogleSheetsReadBlock()
+    input_data = {"range": "Sheet1!A1:B2"}  # no spreadsheet, no credentials
+
+    with pytest.raises(BlockExecutionError, match="No spreadsheet selected"):
+        async for _ in block.execute(input_data):
+            pass


autogpt_platform/backend/backend/blocks/google/sheets_test.py
```diff
@@ -0,0 +1,18 @@
...
+async def test_sheets_read_no_spreadsheet_yields_clean_error():
+    """Executing GoogleSheetsReadBlock with no spreadsheet/credentials should
+    raise a clean BlockExecutionError, not a BlockUnknownError wrapping a
+    TypeError about a missing kwarg."""
+    block = GoogleSheetsReadBlock()
+    input_data = {"range": "Sheet1!A1:B2"}  # no spreadsheet, no credentials
+
+    with pytest.raises(BlockExecutionError, match="No spreadsheet selected"):
+        async for _ in block.execute(input_data):
+            pass
```
majdyz, line 18: 🤖 🟡 **Nice to Have**: This test covers the `spreadsheet=None` path (line 331), but the new `credentials=None` guard (lines 337-338) is untested. Consider adding a second test case that provides a valid `spreadsheet` dict but no credentials to exercise that path too:

```python
async def test_sheets_read_no_credentials_yields_clean_error():
    block = GoogleSheetsReadBlock()
    input_data = {
        "spreadsheet": {
            "id": "abc123",
            "name": "Test",
            "mimeType": "application/vnd.google-apps.spreadsheet",
        },
        "range": "Sheet1!A1:B2",
    }
    with pytest.raises(BlockExecutionError, match="Google credentials are required"):
        async for _ in block.execute(input_data):
            pass
```

majdyz (approved):
Clean, well-structured fix. The two-layer approach (systemic `_base.py` safety net + block-level guard) is the right call.

**What I checked:**
- `_base.py`: `setdefault` correctly injects `None` only for missing auto_credentials kwargs without overwriting already-resolved ones. Placement after input validation is correct.
- `sheets.py`: Optional `credentials` param + early guard yields clean user-facing error.
- Test validates the TypeError→BlockExecutionError improvement.
- Verified all other Google Docs/Sheets blocks have early `if not input_data.document/spreadsheet` guards, so the systemic `None` injection won't cause regressions in the common empty-field case.

One nice-to-have inline comment about test coverage for the `credentials=None` path — not blocking.

majdyz:
## Review: PR #12588

### Issue Found: Incomplete credentials guard for sibling Sheets blocks

The `_base.py` fix (`kwargs.setdefault(kwarg_name, None)`) correctly ensures blocks don't crash with a `TypeError` when credentials are missing. However, only `GoogleSheetsReadBlock.run()` was updated to accept `credentials: GoogleCredentials | None = None` and yield a clean error.

There are **19 other Google Sheets blocks** (Write, Update, Delete, etc.) that still have:
```python
async def run(self, input_data: Input, *, credentials: GoogleCredentials, **kwargs)
```

After this PR, those blocks will receive `credentials=None` from the base class fix, but since their signatures still require non-None `GoogleCredentials`, they'll crash with a different error when trying to use the `None` credentials object (e.g., calling `_build_sheets_service(None)`).

**Recommendation:** Either:
1. Apply the same `| None = None` + early guard pattern to all Google Sheets blocks, or
2. Make the base class fix more defensive: instead of `setdefault(kwarg_name, None)`, check if the field is actually populated, and skip the block with a clean `BlockExecutionError("Missing credentials for {field_name}")` directly in `_execute`.

Option 2 is better because it provides a single centralized guard rather than requiring every block author to remember the pattern.

### CI Note
The test failures (3.11, 3.12, 3.13) were actually CANCELLED due to the "Check PR Status" meta-check timing out -- not real test failures. The tests that did run all passed. A re-run should fix this.

autogpt_platform/backend/backend/blocks/_base.py
```diff
@@ -706,6 +706,13 @@ async def _execute(self, input_data: BlockInput, **kwargs) -> BlockOutput:
                 block_id=self.id,
             )
 
+        # Ensure auto_credentials kwargs are present (even as None) so blocks
+        # that declare e.g. `*, credentials: GoogleCredentials` as a required
+        # kwarg don't crash with TypeError when the execution context didn't
+        # resolve credentials (e.g. the GoogleDriveFile field was empty).
+        for kwarg_name in self.input_schema.get_auto_credentials_fields():
+            kwargs.setdefault(kwarg_name, None)
```
majdyz, line 714: 🤖 🟠 **Should Fix**: The systemic `setdefault(kwarg_name, None)` injects `None` into the `run()` kwargs for **all** blocks using `GoogleDriveFileField` — there are 30+ such blocks in `sheets.py` and 18+ in `docs.py`. However, only `GoogleSheetsReadBlock.run()` was updated to accept `credentials: GoogleCredentials | None = None` with a guard.

For the remaining blocks, this means:
1. If `spreadsheet`/`document` field is empty, the early `if not input_data.spreadsheet` guard fires first, so `credentials=None` is harmless — **this is fine**.
2. But if a block receives a valid `spreadsheet`/`document` AND `credentials=None` (which could happen if credential resolution silently fails or is skipped), the block will call `_build_sheets_service(None)` / `_build_drive_service(None)` → `AttributeError` on None → `BlockUnknownError`. This is the exact same class of opaque error this PR aims to fix.

Since the `_base.py` change is systemic and touches all auto_credentials blocks, the type safety contract (`credentials: GoogleCredentials` — required, non-None) is now silently violated for all of them. The systemic fix and the block-level fix are at odds: the systemic fix guarantees `None` can be injected, but only one block has been updated to handle it.

**Suggestion:** Either:
- **(A)** Remove the systemic `_base.py` change entirely and instead guard in the executor/manager (where credentials are resolved) — so `None` never reaches `run()` in the first place, OR
- **(B)** Keep the systemic fix but add a second systemic guard: if any auto_credentials kwarg is still `None` after resolution, yield a clean error from `_execute()` *before* calling `self.run()`. This way individual blocks don't need to change at all.

Option (B) keeps this simple and avoids touching 50+ blocks:
```python
for kwarg_name in self.input_schema.get_auto_credentials_fields():
    kwargs.setdefault(kwarg_name, None)
    if kwargs[kwarg_name] is None:
        yield "error", f"Credentials required but not provided for '{kwarg_name}'"
        return
```

autogpt_platform/backend/backend/blocks/google/sheets.py
```diff
@@ -326,11 +326,18 @@ def __init__(self):
...
 
     async def run(
-        self, input_data: Input, *, credentials: GoogleCredentials, **kwargs
+        self,
+        input_data: Input,
+        *,
+        credentials: GoogleCredentials | None = None,
+        **kwargs,
     ) -> BlockOutput:
         if not input_data.spreadsheet:
```
majdyz, line 335: 🤖 🟠 **Should Fix**: This block-level fix (making `credentials` optional + adding a guard) is correct in isolation, but it is inconsistent: there are 30+ other blocks in this file and 18+ in `docs.py` that have the exact same `credentials: GoogleCredentials` required kwarg pattern with `GoogleDriveFileField`.

After the `_base.py` systemic change, *all* of those blocks can now receive `credentials=None`. Only this one was updated to handle it. If the systemic fix is kept, either all blocks should be updated, or (better) the guard should be moved to `_execute()` in `_base.py` so no individual block needs to change.

As-is, this creates a maintenance trap: contributors will look at `GoogleSheetsReadBlock` as a pattern, see `credentials: GoogleCredentials | None = None`, but won't know they need to replicate this in every other block.

autogpt_platform/backend/backend/blocks/google/sheets_test.py
majdyz, line 18: 🤖 🟡 **Nice to Have**: The test uses `block.execute()` which goes through `_execute()` → `run()`. This is good because it exercises the full path including the new `_base.py` setdefault logic.

However, the test asserts `BlockExecutionError` with `match="No spreadsheet selected"` — this only exercises the pre-existing `if not input_data.spreadsheet` guard (line 331), NOT the new `if not credentials` guard (lines 337-338) added by this PR.

The new code path this PR adds (the credentials=None guard) is completely untested. Please add a test that provides a valid `spreadsheet` but no credentials to verify the new guard works:

```python
async def test_sheets_read_no_credentials_yields_clean_error():
    block = GoogleSheetsReadBlock()
    input_data = {
        "spreadsheet": {
            "id": "abc123",
            "name": "Test",
            "mimeType": "application/vnd.google-apps.spreadsheet",
        },
        "range": "Sheet1!A1:B2",
    }
    with pytest.raises(BlockExecutionError, match="Google credentials are required"):
        async for _ in block.execute(input_data):
            pass
```

(I see @majdyz already noted this — adding emphasis that this is the only test for the new code and it doesn't actually test it.)

majdyz (changes requested):
## Review Summary

**PR description quality:** Excellent — clear Why/What/How structure, risk assessment, and checklist. Well done.

**Correctness:** The fix addresses the reported bug (TypeError when GoogleDriveFile field is empty), but introduces a new inconsistency.

### Key findings:

**🟠 Systemic fix vs. block-level fix mismatch (Should Fix)**

The `_base.py` change injects `credentials=None` for *all* blocks using `GoogleDriveFileField` (50+ blocks across `sheets.py` and `docs.py`). But only `GoogleSheetsReadBlock` was updated to handle `None` credentials. This creates two problems:

1. **Type safety violation**: All other blocks declare `credentials: GoogleCredentials` (required, non-None) but can now silently receive `None` at runtime. Pyright won't catch this because the injection happens via `**kwargs`.
2. **Latent bug path**: If a block receives a valid spreadsheet/document AND `None` credentials (edge case, but possible), `_build_sheets_service(None)` will throw an opaque `AttributeError` — the exact class of error this PR aims to fix.

**Recommended fix:** Move the None guard into `_execute()` in `_base.py` so it's truly systemic. After `kwargs.setdefault(kwarg_name, None)`, check `if kwargs[kwarg_name] is None` and yield a clean error. This way no individual block needs to change, and the type contract `credentials: GoogleCredentials` remains valid for all 50+ blocks.

**🟡 Test doesn't exercise new code path (Nice to Have)**

The single test only hits the pre-existing "No spreadsheet selected" guard. The new `credentials=None` guard added by this PR is untested. (If the systemic fix in `_base.py` is updated per recommendation, the block-level guard in `sheets.py` can be removed entirely, and the test should instead validate the systemic error message.)

### Request

Please address the systemic inconsistency before merging. The two concrete options are in my inline comment on `_base.py`. After addressing, please show test output proving both paths (no spreadsheet, no credentials) yield clean errors.

majdyz:
@ntindle friendly ping -- the changes-requested items from my review on Mar 31 are still outstanding, and there have been no new commits since the merge from `dev` on Mar 27.

**TL;DR of what needs to change (same as my earlier inline comments):**

The core issue is that the `_base.py` `setdefault(kwarg_name, None)` fix is systemic (affects all 50+ blocks using `GoogleDriveFileField`), but only `GoogleSheetsReadBlock` was updated to handle `credentials=None`. This means any block that receives a valid file *and* `None` credentials will hit an `AttributeError` instead of a clean error -- the same class of opaque error this PR aims to fix.

**Recommended path forward (Option B from my inline comment):**

Move the None guard into `_execute()` in `_base.py`, right after the `setdefault` loop. This makes it truly systemic and avoids touching 50+ individual blocks:

```python
for kwarg_name in self.input_schema.get_auto_credentials_fields():
    kwargs.setdefault(kwarg_name, None)
    if kwargs[kwarg_name] is None:
        yield "error", f"Credentials required but not provided for '{kwarg_name}'"
        return
```

With this change:
1. The block-level `credentials: GoogleCredentials | None = None` change in `sheets.py` can be **reverted** (keep `credentials: GoogleCredentials` as required, preserving type safety).
2. The `credentials=None` guard in `sheets.py` can also be removed since `_execute()` handles it.
3. All 50+ blocks get the protection automatically.
4. The test should be updated to validate the systemic error message.

Additionally, please add a test case that exercises the new `credentials=None` path specifically (valid spreadsheet, no credentials) as noted in my earlier inline comments. The current test only hits the pre-existing "No spreadsheet selected" guard.

Happy to help if you have questions on the approach.

majdyz:

ntindle:
Superseded by #12874 — the systemic None-guard in `Block._execute()` per @majdyz's review comment has been applied in the combined PR. Closing in favour of the combined PR.

PR #12593, opened 2026-03-27
fix(blocks): raise AIConditionBlock errors instead of swallowing them

majdyz:
## Why

Sentry alert [AUTOGPT-SERVER-8C8](https://significant-gravitas.sentry.io/issues/7367978095/) — `AIConditionBlock` failing in prod with:

```
Invalid 'max_output_tokens': integer below minimum value.
Expected a value >= 16, but got 10 instead.
```

Two problems:
1. `max_tokens=10` is below OpenAI's new minimum of 16
2. The `except Exception` handler was calling `logger.error()` which triggered Sentry for what are known block errors, AND silently defaulting to `result=False` — making the block appear to succeed with an incorrect answer

## What

- Bump `max_tokens` from 10 to 16 (fixes the root cause)
- Remove the `try/except` entirely — the executor already handles exceptions correctly (`ValueError` = known/no Sentry, everything else = unknown/Sentry). The old handler was just swallowing errors and producing wrong results.

## Test plan

- [x] Existing `AIConditionBlock` tests pass (block only expects "true"/"false", 16 tokens is plenty)
- [x] No more silent `result=False` on errors
- [x] No more spurious Sentry alerts from `logger.error()`

Fixes AUTOGPT-SERVER-8C8

### DIFF
--- a/autogpt_platform/backend/backend/blocks/ai_condition.py
+++ b/autogpt_platform/backend/backend/blocks/ai_condition.py
@@ -1,3 +1,4 @@
+import re
 from typing import Any
 
 from backend.blocks._base import (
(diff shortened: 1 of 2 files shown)


autogpt_platform/backend/backend/blocks/ai_condition.py
```diff
@@ -162,54 +163,41 @@ async def run(
...
-                credentials=credentials,
-                llm_model=input_data.model,
-                prompt=prompt,
-                max_tokens=10,  # We only expect a true/false response
-            )
+        response = await self.llm_call(
+            credentials=credentials,
+            llm_model=input_data.model,
+            prompt=prompt,
+            max_tokens=16,  # We only expect a true/false response
```
majdyz: 🟠 **Should Fix: Missing regression test for the bug fix.**

Per `CLAUDE.md`: *"Every bug fix should include a test that would have caught it."*

This PR fixes two bugs (max_tokens too low, silent error swallowing) but adds no tests. At minimum, there should be:

1. A test that `max_tokens=16` is passed to `llm_call` (verifies the root cause fix).
2. A test that when `llm_call` raises an exception, it propagates (not silently caught). This is the behavioral change that prevents the silent `result=False` problem.

Without these, a future refactor could reintroduce the exact same bugs.

autogpt_platform/backend/backend/blocks/ai_condition.py
```diff
@@ -162,54 +163,41 @@ async def run(
...
-            self.prompt = response.prompt
-
-        except Exception as e:
-            # In case of any error, default to False to be safe
-            result = False
-            # Log the error but don't fail the block execution
-            import logging
+                # Unclear or conflicting response - default to False and yield error
+                result = False
+                yield "error", f"Unclear AI response: '{response.response}'"
```
majdyz: 🟡 **Nice to Have: `run()` method is ~85 lines, exceeding the ~40-line guideline.**

Per `CLAUDE.md`: *"keep functions under ~40 lines; extract named helpers when a function grows longer."*

The response-parsing logic (lines 174-191) is a self-contained concern and could be extracted into a static helper like `_parse_boolean_response(response_text: str) -> tuple[bool, str | None]`. This would also make it independently testable.

autogpt_platform/backend/backend/blocks/ai_condition.py
majdyz: 🔵 **Nit: Consider naming the magic number.**

`16` is the OpenAI minimum for `max_output_tokens`. If OpenAI bumps this minimum again, it would be clearer to have a named constant (e.g. `MIN_LLM_OUTPUT_TOKENS = 16`) or reference the model's own minimum, rather than embedding a provider-specific constraint as a magic number in a block. The comment helps, but a constant is greppable.

majdyz (commented):
## Review Summary

**Overall: APPROVE** (posted as COMMENT due to GitHub self-review restriction)

### What's good

- **Root cause fix is correct**: Bumping `max_tokens` from 10 to 16 directly addresses the OpenAI `max_output_tokens` minimum of 16.
- **Removing the `try/except` is the right call**: The old handler was (a) swallowing errors and producing silently wrong `result=False`, and (b) triggering Sentry via `logger.error()` for known block failures. Letting exceptions propagate to the executor is the correct pattern -- it already classifies `ValueError` vs unknown errors appropriately.
- **Import cleanup**: Moving `import re` to the top level and switching from `backend.data.block` to `backend.blocks._base` both align with CLAUDE.md style guidelines (top-level imports, sibling-module relative imports).
- **No behavioral regressions in the happy path**: The response-parsing logic is unchanged in substance, just dedented.

### Findings

| Badge | Finding |
|-------|---------|
| :orange_circle: Should Fix | **Missing regression tests.** CLAUDE.md requires every bug fix to include a test. Two tests are needed: (1) verify `max_tokens=16` is passed, (2) verify exceptions from `llm_call` propagate rather than being swallowed. |
| :yellow_circle: Nice to Have | **`run()` is ~85 lines**, over the ~40-line guideline. The boolean-parsing logic is a good extraction candidate. |
| :blue_circle: Nit | **Magic number 16** could be a named constant for greppability and future-proofing. |

No blockers found. The fix is correct and urgent (active Sentry alert). Tests can be added in a follow-up if needed, though ideally they'd land in this PR.

autogpt_platform/backend/backend/blocks/ai_condition.py
majdyz: 🤖 Fixed in 8a2927fd5: Added regression tests in `ai_condition_test.py` — covers `max_tokens=MIN_LLM_OUTPUT_TOKENS` (16) being passed to `llm_call`, exception propagation from `llm_call`, and unit tests for the extracted `_parse_boolean_response()` helper.

autogpt_platform/backend/backend/blocks/ai_condition.py
majdyz: 🤖 Fixed in 8a2927fd5: Extracted `_parse_boolean_response(response_text) -> tuple[bool, str | None]` as a module-level helper. The `run()` method now delegates parsing to this function, reducing its size and making the parsing logic independently testable.

autogpt_platform/backend/backend/blocks/ai_condition.py
majdyz: 🤖 Fixed in 8a2927fd5: Replaced magic number `16` with a named constant `MIN_LLM_OUTPUT_TOKENS = 16` at module level, with a comment explaining its purpose.

ntindle (approved):
this block is poorly designed but not your problem here

PR #12595, opened 2026-03-27
fix(blocks): rename placeholder_values to options on AgentDropdownInputBlock

Torantulino:
## Summary

Resolves [REQ-78](https://linear.app/autogpt/issue/REQ-78): The `placeholder_values` field on `AgentDropdownInputBlock` is misleadingly named. In every major UI framework "placeholder" means non-binding hint text that disappears on focus, but this field actually creates a dropdown selector that restricts the user to only those values.

### Core rename (`autogpt_platform/backend/backend/blocks/io.py`)
- Renamed `placeholder_values` → `options` on `AgentDropdownInputBlock.Input`
- Added clear field description: *"If provided, renders the input as a dropdown selector restricted to these values. Leave empty for free-text input."*
- Updated class docstring to describe actual behavior
- Overrode `model_construct()` to remap legacy `placeholder_values` → `options` for **backward compatibility** with existing persisted agent JSON

### Tests (`autogpt_platform/backend/backend/blocks/test/test_block.py`)
- Updated existing tests to use canonical `options` field name
- Added 2 new backward-compat tests verifying legacy `placeholder_values` still works through both `model_construct()` and `Graph._generate_schema()` paths

### Documentation
- Updated `autogpt_platform/backend/backend/copilot/sdk/agent_generation_guide.md` — changed field name in CoPilot SDK guide
- Updated `docs/integrations/block-integrations/basic.md` — changed field name and description in public docs

### Load tests (`autogpt_platform/backend/load-tests/tests/api/graph-execution-test.js`)
- Removed spurious `placeholder_values: {}` from AgentInputBlock node (this field never existed on AgentInputBlock)
- Fixed execution input to use `value` instead of `placeholder_values`

## Backward Compatibility

Existing agents with `placeholder_values` in their persisted `input_default` JSON will continue to work — the `model_construct()` override transparently remaps the old key to `options`. No database migration needed since the field is stored inside a JSON blob, not as a dedicated column.

## Testing

- All existing tests updated and passing
- 2 new backward-compat tests added
- No frontend changes needed (frontend reads `enum` from generated JSON Schema, not the field name directly)

### DIFF
--- a/autogpt_platform/backend/backend/blocks/io.py
+++ b/autogpt_platform/backend/backend/blocks/io.py
@@ -491,7 +491,7 @@ class AgentDropdownInputBlock(AgentInputBlock):
                 "restricted to these values. Leave empty for free-text input."
             ),
             validation_alias=AliasChoices("options", "placeholder_values"),
-            json_schema_extra={"advanced": False},
+            json_schema_extra={"advanced": False, "secret": False},
         )
 
         def generate_schema(self):
--- a/autogpt_platform/backend/backend/blocks/io.py
+++ b/autogpt_platform/backend/backend/blocks/io.py
@@ -2,6 +2,8 @@ import copy
 from datetime import date, time
 from typing import Any, Optional
 
+from pydantic import AliasChoices, Field
+
 from backend.blocks._base import (
     Block,
     BlockCategory,
@@ -478,24 +480,19 @@ class AgentDropdownInputBlock(AgentInputBlock):
             advanced=False,
             title="Default Value",
         )
-        options: list = SchemaField(
+        # Use Field() directly (not SchemaField) to pass validation_alias,
+        # which handles backward compat for legacy "placeholder_values" across
+        # all construction paths (model_construct, __init__, model_validate).
+        options: list = Field(
+            default_factory=list,
+            title="Dropdown Options",
             description=(
                 "If provided, renders the input as a dropdown selector "
                 "restricted to these values. Leave empty for free-text input."
             ),
-            default_factory=list,
-            advanced=False,
-            title="Dropdown Options",
+            validation_alias=AliasChoices("options", "placeholder_values"),
+            json_schema_extra={"advanced": False},
         )
-
-        @classmethod
-        def model_construct(cls, *args, **kwargs):
-            # Backward compat: remap legacy "placeholder_values" → "options"
-            if "placeholder_values" in kwargs and "options" not in kwargs:
-                kwargs["options"] = kwargs.pop("placeholder_values")
-            elif "placeholder_values" in kwargs:
-                kwargs.pop("placeholder_values")
-            return super().model_construct(*args, **kwargs)
 
         def generate_schema(self):
             schema = super().generate_schema()
--- a/autogpt_platform/backend/backend/blocks/io.py
+++ b/autogpt_platform/backend/backend/blocks/io.py
@@ -467,7 +467,8 @@ class AgentFileInputBlock(AgentInputBlock):
 
 class AgentDropdownInputBlock(AgentInputBlock):
     """
-    A specialized text input block that relies on placeholder_values to present a dropdown.
+    A specialized text input block that presents a dropdown selector
+    restricted to a fixed set of values.
     """
 
     class Input(AgentInputBlock.Input):
@@ -477,16 +478,28 @@ class AgentDropdownInputBlock(AgentInputBlock):
             advanced=False,
             title="Default Value",
         )
-        placeholder_values: list = SchemaField(
-            description="Possible values for the dropdown.",
+        options: list = SchemaField(
+            description=(
+                "If provided, renders the input as a dropdown selector "
+                "restricted to these values. Leave empty for free-text input."
+            ),
             default_factory=list,
             advanced=False,
             title="Dropdown Options",
         )
 
+        @classmethod
+        def model_construct(cls, *args, **kwargs):
+            # Backward compat: remap legacy "placeholder_values" → "options"
+            if "placeholder_values" in kwargs and "options" not in kwargs:
+                kwargs["options"] = kwargs.pop("placeholder_values")
+            elif "placeholder_values" in kwargs:
+                kwargs.pop("placeholder_values")
+            return super().model_construct(*args, **kwargs)
+
         def generate_schema(self):
             schema = super().generate_schema()
-            if possible_values := self.placeholder_values:
+            if possible_values := self.options:
                 schema["enum"] = possible_values
             return schema
 
@@ -504,13 +517,13 @@ class AgentDropdownInputBlock(AgentInputBlock):
                 {
                     "value": "Option A",
                     "name": "dropdown_1",
-                    "placeholder_values": ["Option A", "Option B", "Option C"],
+                    "options": ["Option A", "Option B", "Option C"],
                     "description": "Dropdown example 1",
                 },
                 {
                     "value": "Option C",
                     "name": "dropdown_2",
-                    "placeholder_values": ["Option A", "Option B", "Option C"],
+                    "options": ["Option A", "Option B", "Option C"],
                     "description": "Dropdown example 2",
                 },
             ],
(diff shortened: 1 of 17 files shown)


majdyz:
## Review: PR #12595

Good rename. The `model_construct` override is the correct approach for backward compat since `graph.py:332` uses `model_construct(**input_default)` to deserialize stored node defaults.

### Observations

1. **Backward compat looks correct** -- the `model_construct` override remaps `placeholder_values` to `options`, and the tests cover both the direct and legacy paths including the full `Graph._generate_schema` integration.

2. **Load test cleanup is appropriate** -- `placeholder_values: {}` was wrong anyway (dict instead of list).

3. **Agent generation guide updated** -- good, this ensures the copilot SDK generates correct graphs going forward.

4. **One minor concern**: The `model_construct` override only fires when called via `cls.model_construct()`. If anyone ever calls `AgentDropdownInputBlock.Input(**kwargs)` or `model_validate()` with old data containing `placeholder_values`, the compat shim won't trigger. In practice this doesn't happen today (only `model_construct` is used in `graph.py`), but a `model_validator(mode='before')` would be more robust. Not blocking.

**Verdict: LGTM** -- ready to merge.

majdyz (approved):
Clean rename with good backward compatibility. A few observations:

**Looks good:**
- `model_construct()` override correctly intercepts the only call path (`graph.py:332`) to remap `placeholder_values` -> `options`
- No frontend changes needed since the frontend reads `enum` from the generated JSON Schema
- Existing agent JSON fixtures in `agents/` will continue to work through the remap
- Tests cover both canonical and legacy paths
- Load test fix is a good cleanup (removing the bogus `placeholder_values: {}` on AgentInputBlock)
- Documentation updates are thorough

**Minor note (non-blocking):** The agent JSON fixtures in `backend/agents/` still reference `placeholder_values` -- these will work at runtime via the `model_construct` remap, but could optionally be migrated to `options` in a follow-up to keep the canonical representation consistent. Not required for this PR.

autogpt_platform/backend/backend/blocks/io.py
```diff
@@ -477,16 +478,28 @@ class Input(AgentInputBlock.Input):
...
-            description="Possible values for the dropdown.",
+        options: list = SchemaField(
+            description=(
+                "If provided, renders the input as a dropdown selector "
+                "restricted to these values. Leave empty for free-text input."
+            ),
             default_factory=list,
             advanced=False,
             title="Dropdown Options",
         )
```
majdyz, line 495: 🤖 🟠 **Should Fix**: The `model_construct()` override only covers the **schema generation** path (`graph.py:332`), not the `__init__()` path used during **block execution** (`blocks/_base.py:711`). While this works today because `options` is only read in `generate_schema()` (not in `run()`), it's fragile — any future code path that constructs this model via `__init__()` or `model_validate()` will silently lose the legacy `placeholder_values`.

A `@model_validator(mode="before")` covers **all** construction paths and is the Pydantic-idiomatic approach:

```python
from pydantic import model_validator

@model_validator(mode="before")
@classmethod
def _remap_legacy_placeholder_values(cls, data):
    if isinstance(data, dict) and "placeholder_values" in data:
        if "options" not in data:
            data["options"] = data.pop("placeholder_values")
        else:
            data.pop("placeholder_values")
    return data
```

This replaces the `model_construct()` override entirely and handles `__init__`, `model_validate`, AND `model_construct` uniformly.

majdyz: **Correction**: I just verified that `model_validator(mode="before")` does NOT fire on `model_construct()` in Pydantic v2 (by design — `model_construct` bypasses all validation).

So the `model_construct()` override is actually **necessary** for the schema generation path (`graph.py:332`). My recommendation changes to:

**Keep the `model_construct()` override** (it's correct for the schema path), but **also add a `model_validator(mode="before")`** for defense-in-depth on the `__init__`/`model_validate` paths. This way all construction routes are covered.

That said, since `options` is currently only used in `generate_schema()` which goes through `model_construct()`, the PR as-is is **functionally correct** for the current codebase. Downgrading to 🟡 Nice to Have.

majdyz (dismissed):
Good rename — `options` is much clearer than `placeholder_values`. The backward compat via `model_construct()` override correctly covers the schema generation path (the only path where `options` is read), so **existing agents with persisted `placeholder_values` will continue to render dropdowns in the UI**.

The frontend reads `enum` from the generated JSON schema, not the field name directly, so no frontend changes are needed.

One minor suggestion (🟡): also add a `model_validator(mode="before")` for defense-in-depth alongside the `model_construct()` override (see inline comment + correction).

LGTM with the caveat above.

autogpt_platform/backend/backend/blocks/io.py
majdyz: Re: alias approach — I tested `validation_alias=AliasChoices("options", "placeholder_values")` and it works across **all three paths** (`__init__`, `model_validate`, AND `model_construct`):

```python
from pydantic import Field, AliasChoices

options: list = Field(
    default_factory=list,
    title="Dropdown Options",
    description="...",
    validation_alias=AliasChoices("options", "placeholder_values"),
)
```

This would replace both the `model_construct()` override AND cover `__init__`/`model_validate` — single line, zero custom code.

**Caveat**: `SchemaField()` doesn't forward `validation_alias` to `Field()`. You'd need to either:
1. Use `Field()` directly for this one field (simplest), or
2. Add `validation_alias` passthrough to `SchemaField`

Either way, I'd recommend a test for the `__init__` path too (simulating execution with legacy data), even though `options` isn't read during execution today — it's cheap insurance.

autogpt_platform/backend/backend/blocks/io.py
```diff
@@ -477,16 +480,23 @@ class Input(AgentInputBlock.Input):
...
+        options: list = Field(
             default_factory=list,
-            advanced=False,
             title="Dropdown Options",
+            description=(
+                "If provided, renders the input as a dropdown selector "
+                "restricted to these values. Leave empty for free-text input."
+            ),
+            validation_alias=AliasChoices("options", "placeholder_values"),
+            json_schema_extra={"advanced": False},
```
majdyz: 🤖 🔵 **Nit**: The `Field()` usage passes `json_schema_extra={"advanced": False}` but omits `"secret"` which `SchemaField` normally includes (as `False`). Not a bug since the frontend checks `prop.secret === true` (missing = non-secret), but for consistency:

```python
json_schema_extra={"advanced": False, "secret": False},
```

majdyz: 🤖 Fixed in aabe20c5f81af7796070061a639d58b63bcfa237: Added `"secret": False` to `json_schema_extra` for consistency with SchemaField.

autogpt_platform/backend/backend/blocks/io.py
```diff
@@ -477,16 +480,23 @@ class Input(AgentInputBlock.Input):
...
+        # all construction paths (model_construct, __init__, model_validate).
+        options: list = Field(
             default_factory=list,
-            advanced=False,
             title="Dropdown Options",
+            description=(
+                "If provided, renders the input as a dropdown selector "
+                "restricted to these values. Leave empty for free-text input."
+            ),
+            validation_alias=AliasChoices("options", "placeholder_values"),
```
majdyz, line 493: 🤖 🟡 **Nice to Have**: There are ~10 bundled agent JSON templates in `autogpt_platform/backend/agents/` that still contain `"placeholder_values"` in their `input_default`. The `validation_alias` handles this transparently at runtime, but updating these files to use `"options"` would keep the canonical templates consistent with the new field name.

```bash
# Quick fix
find autogpt_platform/backend/agents -name "*.json" -exec \
  sed -i '' 's/"placeholder_values"/"options"/g' {} +
```

majdyz: 🤖 Fixed in b7d7d8325e4f9ae62fed0def8e4a289603ab155e: Updated all 12 bundled agent JSON templates to use `options` instead of `placeholder_values`.

majdyz: Reverted in 09807c9d06 — these files don't need updating since the `validation_alias` handles legacy `placeholder_values` transparently. Changing them would contradict the backward compat story.

majdyz (approved):

PR #12601, opened 2026-03-30
dx(backend/copilot): clarify ambiguous agent goals using find_block before generation

anvyle:
### Why / What / How

**Why:** When a user asks CoPilot to build an agent with an ambiguous goal (output format, delivery channel, data source, or trigger unspecified), the agent generator previously made assumptions and jumped straight into JSON generation. This produced agents that didn't match what the user actually wanted, requiring multiple correction cycles.

**What:** Adds a "Clarifying Before Building" section to the agent generation guide. When the goal is ambiguous, CoPilot first calls `find_block` to discover what the platform actually supports for the ambiguous dimension, then asks the user one concrete question grounded in real platform options (e.g. "The platform supports Gmail, Slack, and Google Docs — which should the agent use for delivery?"). Only after the user answers does the full agent generation workflow proceed.

**How:** The clarification instruction is added to `agent_generation_guide.md` — the guide loaded on-demand via `get_agent_building_guide` when the LLM is about to build an agent. This avoids polluting the system prompt supplement (which loads for every CoPilot conversation, not just agent building). No dedicated tool is needed — the LLM asks naturally in conversation text after discovering real platform options via `find_block`.

### Changes 🏗️

- `backend/copilot/sdk/agent_generation_guide.md`: Adds "Clarifying Before Building" section before the workflow steps. Instructs the model to call `find_block` for the ambiguous dimension, ask the user one grounded question, wait for the answer, then proceed to generation.
- `backend/copilot/prompting_test.py`: New test file verifying the guide contains the clarification section and references `find_block`.

#### For code changes:
- [x] I have made a test plan
  - [ ] Ask CoPilot to "build an agent to send a report" (ambiguous output) — verify it calls `find_block` for delivery options and asks one grounded question before generating JSON
  - [ ] Ask CoPilot to "build an agent to scrape prices from Amazon and email me daily" (specific goal) — verify it skips clarification and proceeds directly to agent generation
  - [ ] Verify the clarification question lists real block options (e.g. Gmail, Slack, Google Docs) rather than abstract options


autogpt_platform/backend/backend/copilot/sdk/agent_generation_guide.md
```diff
@@ -5,23 +5,30 @@ generate the agent JSON yourself using block schemas, then validate and save.
...
 ### Workflow for Creating/Editing Agents
 
-1. **Discover blocks**: Call `find_block(query, include_schemas=true)` to
-   search for relevant blocks. This returns block IDs, names, descriptions,
-   and full input/output schemas.
-2. **Find library agents**: Call `find_library_agent` to discover reusable
+1. **Clarify if needed**: If the goal is ambiguous (output format, delivery
+   channel, data source, or trigger unspecified), call `find_block` with a
+   query targeting the ambiguous dimension (e.g. `"email send"`, `"slack
+   message"`, `"google docs write"`) to discover what the platform supports.
```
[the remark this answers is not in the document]
anvyle: 🤖 The current guide no longer contains the "google docs" reference — it was part of an earlier clarification step that has since been replaced by the dedicated `clarify_agent_request` tool. No change needed.

autogpt_platform/backend/backend/copilot/tools/clarify_agent_request.py
```diff
@@ -0,0 +1,92 @@
...
+
+        example = ", ".join(options) if options else None
+        clarifying_question = ClarifyingQuestion(
+            question=question,
+            keyword=keyword,
+            example=example,
+        )
+        return ClarificationNeededResponse(
+            message=question,
+            session_id=session_id,
```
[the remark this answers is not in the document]
anvyle: 🤖 Fixed in cd0c8e060: added `.strip()` and an early-return `ErrorResponse` when `question` is blank or missing. Two new tests cover both the blank-whitespace and missing-key cases.

majdyz (commented):
## Critical Review — Round 1

### 🔴 PR description does not match implementation

The PR body claims:
> The `_AGENT_CLARIFY_NOTE` constant is appended to both `get_sdk_supplement` and `get_baseline_supplement`, ensuring coverage across both execution paths.
> `backend/copilot/prompting.py`: Adds `_AGENT_CLARIFY_NOTE` — a system prompt fragment injected into both `get_sdk_supplement` and `get_baseline_supplement`.

**Neither of these is true.** `prompting.py` is not modified. The actual approach is adding step 1 to `agent_generation_guide.md`. The description must be updated to match the actual implementation, otherwise reviewers will be confused.

### 🔴 The tool is a pure passthrough — no logic, no value over a plain text response

`ClarifyAgentRequestTool._execute()` receives `question`, `options`, and `keyword`, then wraps them in `ClarificationNeededResponse`. It does **zero** computation:
- It does NOT call `find_block` to discover options
- It does NOT detect whether a goal is ambiguous
- It does NOT validate that the options match real platform blocks

The tool is entirely dependent on the LLM to:
1. Decide the goal is ambiguous
2. Call `find_block` first
3. Then call `clarify_agent_request` with the results

This means the tool adds no enforcement over what the LLM can already do: ask a clarifying question in plain text. The `ClarificationNeededResponse` type already existed — the only thing this tool provides is a structured JSON response format. Consider whether this justifies an entire new tool + registration + permissions entry, or if the guide change alone is sufficient.

### 🟡 `options` should be required, not optional

The entire value proposition is presenting "real platform options discovered via `find_block`." But `options` is optional — if the LLM omits it, the response has `example: None`, which is just a plain question with no grounded options. This undermines the stated goal.

If the tool should always present concrete options, make `options` required (and validate `len(options) > 0`). If it's OK to ask without options, then the PR description overstates the grounding guarantee.

### 🟡 `keyword` field is dead weight

`keyword` is stored in `ClarifyingQuestion.keyword` but nothing downstream reads it. It's not displayed to the user, not used for any logic, and not consumed by any other tool. If it's meant for analytics/logging, add a comment explaining the intent. Otherwise remove it.

### 🟡 No skip-clarification path for unambiguous goals

The guide says: "Call `clarify_agent_request` first to confirm the user's intent." This is step 1 — unconditional. The test plan says specific goals ("build an agent to scrape prices from Amazon and email me daily") should skip clarification, but there's no mechanism to enforce this. The guide instruction is absolute.

Consider adding language like: "If the user's goal already specifies the output format, delivery channel, data source, and trigger, skip this step."

### 🟡 Test file — only happy-path and trivial error cases

The tests verify:
- Question is echoed back ✅
- Options are joined ✅
- Missing/blank question returns error ✅

Missing coverage:
- No integration test verifying the tool is registered and callable via `execute_tool()`
- No test for how the frontend/SDK handles `ClarificationNeededResponse` from this tool
- No test for `None` session (line 85: `session.session_id if session else None` — but no test for `session=None`)

### 🟢 Minor nits
- `del user_id` on line 81 — this is fine for the BaseTool contract, consistent with other tools.
- Test naming and structure follow the existing `*_test.py` convention. ✅
- `permissions.py` and `__init__.py` registrations are correct and alphabetically ordered. ✅

majdyz (commented):
## Critical Review — Round 2 (after round 1 fixes)

### 🔴 `keyword=""` breaks the frontend answer-mapping contract

The frontend (`ClarificationQuestionsCard.tsx`) uses `keyword` as the **object key** to map user answers back to questions:
```tsx
answers[q.keyword]    // line 82, 152, 198, 200
```

Setting `keyword=""` means all questions map to the same empty-string key. Currently this works only because the tool always sends exactly one question, so the frontend fallback (`question-{index}` via `normalizeClarifyingQuestions`) kicks in. But:

1. This is fragile — if someone adds a second question, answers will collide silently.
2. The tool is fighting the model: `ClarifyingQuestion.keyword` is a required non-optional field, and the frontend expects it to be meaningful.

**Fix**: Set `keyword` to something descriptive like the first option or a sanitized version of the ambiguous dimension (e.g. `"delivery_channel"`, `"data_source"`). Or add a `dimension` parameter to the tool so the caller specifies which dimension is ambiguous.

### 🟡 `example` format mismatch with frontend expectations

The tool joins options as `"Email, Slack, Google Docs"` and stores this in `example`. The frontend renders this as italic helper text: `Example: Email, Slack, Google Docs`. But these aren't "examples" — they're the **actual choices** the user should pick from. The UX is misleading.

Consider either:
- Storing options as-is in a dedicated `options` field on `ClarifyingQuestion` (would require a model change)
- Or at minimum, prefixing with "Options: " instead of relying on the frontend's "Example: " label

### 🟡 Tool description redundantly says "Call find_block first"

The tool description says "Call find_block first to discover available options". But this instruction also lives in the guide. Having it in both places risks them drifting apart. The tool description should focus on what **this** tool does, not on prerequisites that belong in the workflow guide.

### 🟢 Improvements from Round 1 that look good
- `options` is now required ✅
- Skip-clarification guidance in the guide ✅
- `session=None` test coverage ✅
- Empty options validation ✅

majdyz (commented):
## Critical Review — Round 3 (after round 2 fixes)

Overall quality is now significantly better. Remaining issues are minor.

### 🟡 `dimension` enum not validated at runtime

The JSON Schema has `enum: [output_format, delivery_channel, data_source, trigger]` which guides OpenAI-compatible models, but `_execute()` only checks `if not dimension:`. An LLM could pass `dimension="foo"` and it would be silently accepted, with `"foo"` used as the keyword. Add a runtime check:
```python
_VALID_DIMENSIONS = frozenset({"output_format", "delivery_channel", "data_source", "trigger"})
if dimension not in _VALID_DIMENSIONS:
    return ErrorResponse(...)
```

### 🟡 `options` items not validated for empty strings

`options=["", "  "]` passes the `if not options:` check (truthy list) but contains no meaningful content. The frontend would render `Example:  ,   ` which looks broken. Filter/strip options before use.

### 🟡 `logger` imported but unused

Line 3 imports `logging` and line 16 creates `logger = logging.getLogger(__name__)` but neither is used anywhere in the module. Remove or add a debug log in `_execute` for observability.

### 🟡 PR description still inaccurate

The original PR body still claims changes to `prompting.py` with `_AGENT_CLARIFY_NOTE`. The author should update the PR description to match the actual implementation. This will confuse future readers and git-blame archeologists.

### 🟢 What looks good now
- `dimension` as keyword is a solid design that works with the frontend's answer-mapping ✅
- Enum constraint on dimension prevents freeform values at the schema level ✅
- All three params are required with proper validation ✅
- Skip-clarification guidance is clear and has a concrete example ✅
- Test coverage is comprehensive (14 test cases covering happy path, error cases, null session) ✅
- Tool registration in permissions.py and __init__.py is correct ✅

majdyz (commented):
## Critical Review — Round 4 (after round 3 fixes)

Code quality is now solid. Two remaining concerns:

### 🟡 `_VALID_DIMENSIONS` and schema enum can drift apart

The same 4 dimension values are defined in two places:
- `_VALID_DIMENSIONS` frozenset (line 15) — used for runtime validation
- `parameters` schema enum (lines 49-54) — used for LLM schema

If someone adds a new dimension, they must update both. Derive one from the other:
```python
_VALID_DIMENSIONS = frozenset({"output_format", "delivery_channel", "data_source", "trigger"})

# In parameters property:
"enum": sorted(_VALID_DIMENSIONS),
```

### 🟡 `raw_options: list` lacks type annotation specificity

Line 91: `raw_options: list = kwargs.get("options", [])` — bare `list` annotation. Minor type hygiene issue. Should be `list[Any]` for explicitness or just drop the annotation since kwargs values are untyped anyway.

### 🟢 Everything else looks good
- Dimension enum validated at both schema and runtime ✅
- Options stripped and filtered, whitespace-only rejected ✅
- `keyword=dimension` gives frontend meaningful answer-mapping keys ✅
- Guide has clear skip-clarification criteria with concrete example ✅
- No unused imports ✅
- 17 test cases with clear section grouping (happy path, dimension, question, options) ✅
- `permissions.py` and `__init__.py` registrations correct and alphabetically sorted ✅
- `as_openai_tool()` passes schema through, so `enum`/`minItems` constraints reach the LLM ✅

majdyz (commented):
## Critical Review — Round 5 (after round 4 fixes)

No new issues found. The code is clean and ready.

### 🟢 All previous issues resolved
- Schema enum now derived from `_VALID_DIMENSIONS` via `sorted()` — single source of truth ✅
- Test cross-checks schema enum against `_VALID_DIMENSIONS` constant ✅
- Bare `list` annotation removed ✅
- `dimension` validated at both schema and runtime ✅
- Options stripped/filtered, whitespace-only rejected ✅
- `keyword=dimension` gives frontend meaningful keys ✅
- Guide has conditional skip-clarification with concrete example ✅
- No unused imports ✅
- 17 well-organized test cases ✅
- Tool registration in permissions.py and __init__.py correct ✅

### Remaining non-code item
- PR description still mentions `prompting.py` / `_AGENT_CLARIFY_NOTE` changes that don't exist. Author should update the PR body.

majdyz (commented):
## Critical Review — Round 6 (final pass)

No new code issues. The implementation is solid.

### Summary of all changes across 6 review rounds

**Original PR** added a `ClarifyAgentRequestTool` with optional parameters, no validation, dead `keyword` field, misleading PR description, and unconditional clarification step.

**After 6 rounds of review/fix**, the final state:

1. **`dimension` enum** (output_format, delivery_channel, data_source, trigger) — validated at both schema and runtime, derived from single `_VALID_DIMENSIONS` constant, used as `keyword` for frontend answer-mapping
2. **All params required** — `dimension`, `question`, `options` are all required with proper validation
3. **Input sanitization** — options stripped of whitespace, empty/whitespace-only filtered out
4. **Guide updated** — step 1 is conditional ("if ambiguous") with explicit skip criteria and concrete example
5. **17 test cases** — happy path (7), dimension validation (3), question validation (2), options validation (3), schema introspection (2)
6. **No dead code** — unused `logger` and `keyword` parameter removed

### Only remaining item (non-code)
Author should update the PR description to match the actual implementation (currently references `prompting.py` / `_AGENT_CLARIFY_NOTE` which were never changed).

autogpt_platform/backend/backend/copilot/tools/clarify_agent_request.py
```diff
@@ -0,0 +1,123 @@
...
+
+        clarifying_question = ClarifyingQuestion(
+            question=question,
+            keyword=dimension,
+            example=", ".join(options),
+        )
+        return ClarificationNeededResponse(
+            message=question,
+            session_id=session_id,
+            questions=[clarifying_question],
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in 40690bff9c: Added `tool-clarify_agent_request` case in MessagePartRenderer.tsx routing to CreateAgentTool (which already handles ClarificationNeededResponse), and added to CUSTOM_TOOL_TYPES in helpers.ts.

autogpt_platform/backend/backend/copilot/tools/clarify_agent_request.py
```diff
@@ -0,0 +1,121 @@
...
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in 40690bff9c: Same fix — added `tool-clarify_agent_request` handler in MessagePartRenderer.tsx and CUSTOM_TOOL_TYPES.

autogpt_platform/backend/backend/copilot/tools/clarify_agent_request.py
```diff
@@ -0,0 +1,123 @@
...
+        self,
+        user_id: str | None,
+        session: ChatSession,
+        **kwargs,
+    ) -> ToolResponseBase:
+        del user_id  # unused; required by BaseTool contract
+        dimension = str(kwargs.get("dimension", "")).strip()
+        question = str(kwargs.get("question", "")).strip()
+        raw_options = kwargs.get("options", [])
+        options = [o.strip() for o in raw_options if isinstance(o, str) and o.strip()]
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in 2f9288df61: Replaced `str(kwargs.get(...))` with `isinstance` checks — prevents `str(None)` producing `"None"` which would pass validation. Also guard `raw_options` with `isinstance(raw_options, list)`.

autogpt_platform/backend/backend/copilot/sdk/agent_generation_guide.md
```diff
@@ -5,23 +5,30 @@ generate the agent JSON yourself using block schemas, then validate and save.
...
 ### Workflow for Creating/Editing Agents
 
-1. **Discover blocks**: Call `find_block(query, include_schemas=true)` to
+1. **Clarify intent (if ambiguous)**: If the user's goal is missing the
+   output format, delivery channel, data source, or trigger, call
+   `find_block` for the ambiguous dimension, then call
+   `clarify_agent_request` with the discovered options. Wait for the
+   user's answer before continuing. **Skip this step** if the goal already
+   specifies all dimensions (e.g. "scrape prices from Amazon and email me
+   daily" — output, source, channel, and trigger are all clear).
```
majdyz: 🤖 🟠 **Should Fix**: The clarification instruction lives only in this lazy-loaded guide (returned by `get_agent_building_guide`), so the LLM will only see it if it happens to call that tool *before* jumping straight to generation. The PR description claims this is "injected into the system prompt supplement (`prompting.py`) so it is always active before any tool call" — but `prompting.py` is not modified in this PR.

If the intent is to guarantee the LLM always clarifies before building, the instruction should be injected into the system prompt via `prompting.py` (as the description states). The guide-only approach relies on the LLM calling `get_agent_building_guide` first, which it may skip for seemingly simple requests.

Sentry bot also flagged this as the missing `_AGENT_CLARIFY_NOTE` in `prompting.py`.

majdyz:
🤖 🔴 ~~**Blocker**~~: The PR description was inaccurate:

1. ~~**"Changes" section** claims `prompting.py` "Adds `_AGENT_CLARIFY_NOTE`" — but `prompting.py` was not modified~~
2. ~~**"Changes" section** claims `agent_generation_guide.md` "Removes the now-redundant 'Clarifying Before Building' section" — but the PR *added* a clarification step; there was no prior section to remove~~
3. ~~**"How" section** says "The clarification rule is injected into the system prompt supplement (`prompting.py`) so it is always active before any tool call" — the actual implementation added the rule to the lazy-loaded guide~~

**Resolved:** Fixed in e074413942 — `_AGENT_CLARIFY_NOTE` is now injected into `prompting.py` (both SDK and baseline supplements), and the redundant guide step is removed. The description now matches the implementation. PR description itself should still be updated to reflect the final state of changes.

autogpt_platform/backend/backend/copilot/tools/clarify_agent_request.py
```diff
@@ -0,0 +1,128 @@
...
+            return ErrorResponse(
+                message="clarify_agent_request requires at least one non-empty option.",
+                error="missing_options",
+                session_id=session_id,
+            )
+
+        clarifying_question = ClarifyingQuestion(
+            question=question,
+            keyword=dimension,
+            example=", ".join(options),
```
majdyz: 🤖 🟡 **Nice to Have**: The options are joined with `", "` and stored in `ClarifyingQuestion.example`, which the frontend renders as *"Example: Email, Slack, Google Docs"* above a free-text textarea. The user sees a hint but must type their answer manually.

Consider whether the `options` should be surfaced as clickable chips/buttons instead of a comma-joined string in the `example` field. This would improve UX (one-click selection) and avoid overloading the `example` field semantics (which traditionally means "a sample value", not "a list of choices").

Not a blocker — the current approach works, but it's a missed UX opportunity.

autogpt_platform/backend/backend/copilot/sdk/agent_generation_guide.md
majdyz: 🤖 Fixed in e074413942: Added `_AGENT_CLARIFY_NOTE` to `prompting.py`, injected into both `get_sdk_supplement` and `get_baseline_supplement`. Removed the redundant clarification step from the guide and restored original step numbering. The clarification instruction is now always active in the system prompt.

autogpt_platform/backend/backend/copilot/tools/clarify_agent_request.py
majdyz: 🤖 Not addressed in this cycle — this is a Nice to Have UX improvement that can be done in a follow-up PR. The current approach works correctly.

autogpt_platform/backend/backend/copilot/prompting.py
```diff
@@ -318,4 +335,4 @@ def get_baseline_supplement() -> str:
         The supplement string to append to the system prompt
     """
     tool_docs = _generate_tool_documentation()
-    return tool_docs + _SHARED_TOOL_NOTES
+    return tool_docs + _SHARED_TOOL_NOTES + _AGENT_CLARIFY_NOTE
```
majdyz: 🤖 🟠 **Should Fix**: No test validates that `_AGENT_CLARIFY_NOTE` is included in the supplement output. There's no `prompting_test.py` in the codebase. Consider adding a minimal test that verifies both `get_sdk_supplement()` and `get_baseline_supplement()` contain the `clarify_agent_request` instruction — this prevents regressions if someone later refactors the supplement assembly logic.

Example:
```python
from backend.copilot.prompting import get_sdk_supplement, get_baseline_supplement

def test_sdk_supplement_includes_clarify_note():
    result = get_sdk_supplement(use_e2b=False, cwd="/tmp")
    assert "clarify_agent_request" in result

def test_baseline_supplement_includes_clarify_note():
    result = get_baseline_supplement()
    assert "clarify_agent_request" in result
```

autogpt_platform/backend/backend/copilot/tools/clarify_agent_request.py
```diff
@@ -0,0 +1,128 @@
...
+        raw_options = kwargs.get("options")
+
+        dimension = dimension_val.strip() if isinstance(dimension_val, str) else ""
+        question = question_val.strip() if isinstance(question_val, str) else ""
+        options = (
+            [o.strip() for o in raw_options if isinstance(o, str) and o.strip()]
+            if isinstance(raw_options, list)
+            else []
+        )
+        session_id = session.session_id if session else None
```
majdyz: 🤖 🔵 **Nit**: The `session` parameter is typed as `ChatSession` (non-optional) per the `BaseTool._execute` contract, but line 93 guards against `session` being `None` with `session.session_id if session else None`. Either:
- The guard is dead code (if `session` is guaranteed non-None by the framework), or
- The type hint should be `ChatSession | None` to match the defensive check

This pattern appears in other tools too, so not specific to this PR — but worth noting for consistency.

autogpt_platform/backend/backend/copilot/prompting.py
majdyz: 🤖 Fixed in 120ad6259c: Added `prompting_test.py` with 3 tests — verifies `clarify_agent_request` appears in SDK local, SDK E2B, and baseline supplement outputs.

autogpt_platform/backend/backend/copilot/tools/clarify_agent_request.py
majdyz: 🤖 Not addressed — this is a pre-existing pattern across all tools in the codebase (not specific to this PR). Fixing it would require a broader refactor of the BaseTool contract.

majdyz:
🤖 **Cycle 3 Review — Final Assessment**

The PR is now in good shape after addressing the major issues. Here's the summary:

### Resolved (across 3 cycles)
- ✅ **Blocker**: `_AGENT_CLARIFY_NOTE` injected into `prompting.py` — always active in system prompt (SDK local, SDK E2B, baseline)
- ✅ **Should Fix**: Redundant guide step removed, original step numbering restored
- ✅ **Should Fix**: `prompting_test.py` added with 3 regression tests for supplement injection

### Not Addressed (accepted)
- 🟡 **Nice to Have**: Options displayed as comma-joined text in `example` field — works but clickable chips would be better UX (follow-up PR)
- 🔵 **Nit**: `session` type hint mismatch — pre-existing pattern across all tools, not specific to this PR

### Remaining Action for PR Author
- ⚠️ **PR description** still references the old approach (guide-only placement). Should be updated to match the final implementation.

### Test Results
All 79 relevant tests pass: 18 tool tests + 3 prompting tests + 58 permissions tests.

No further code changes needed. LGTM pending description update.

majdyz:
🤖 **Major simplification: dropped the `clarify_agent_request` tool entirely**

After critical review, the tool was a pure UI formatter — it received LLM-curated data and wrapped it in a `ClarificationNeededResponse`. The real behavior change comes from `_AGENT_CLARIFY_NOTE` in the system prompt, which instructs the LLM to:
1. Call `find_block` to discover real platform options
2. Ask a grounded question in plain text
3. Wait for the user's answer before generating

**Before (8 files, ~400 lines added):**
- New tool (`clarify_agent_request.py` + 237-line test)
- Tool registry entry, permissions entry, frontend routing
- System prompt referencing the tool

**After (2 files, ~35 lines added):**
- `prompting.py`: `_AGENT_CLARIFY_NOTE` injected into system prompt
- `prompting_test.py`: 3 regression tests

The system prompt note IS the fix. The tool was unnecessary scaffolding around it.

majdyz (approved):

PR #12604, opened 2026-03-30
fix(copilot): add tool call circuit breakers and intermediate persistence

majdyz:
## Why

CoPilot session `d2f7cba3` took **82 minutes** and cost **$20.66** for a single user message. Root causes:
1. Redis session meta key expired after 1h, making the session invisible to the resume endpoint — causing empty page on reload
2. Redis stream key also expired during sub-agent gaps (task_progress events produced no chunks)
3. No intermediate persistence — session messages only saved to DB after the entire turn completes
4. Sub-agents retried similar WebSearch queries (addressed via prompt guidance)

### Redis TTL fixes (root cause of empty session on reload)
- `publish_chunk()` now periodically refreshes **both** the session meta key AND stream key TTL (every 60s).
- `task_progress` SDK events now emit `StreamHeartbeat` chunks, ensuring `publish_chunk` is called even during long sub-agent gaps where no real chunks are produced.
- Without this fix, turns exceeding the 1h `stream_ttl` lose their "running" status and stream data, making `get_active_session()` return False.

### Intermediate DB persistence
- Session messages flushed to DB every **30 seconds** or **10 new messages** during the stream loop.
- Uses `asyncio.shield(upsert_chat_session())` matching the existing `finally` block pattern.

### Orphaned message cleanup on rollback
- On stream attempt rollback, orphaned messages persisted by intermediate flushes are now cleaned up from the DB via `delete_messages_from_sequence`.
- Prevents stale messages from resurfacing on page reload after a failed retry.

### Prompt guidance
- Added web search best practices to code supplement (search efficiency, sub-agent scope separation).

### Approach: root cause fixes, not capability limits
- **No tool call caps** — artificial limits on WebSearch or total tool calls would reduce autopilot capability without addressing why searches were redundant.
- **Task tool remains enabled** — sub-agent delegation via Task is a core capability. The existing `max_subtasks` concurrency guard is sufficient.
- The real fixes (TTL refresh, persistence, prompt guidance) address the underlying bugs and behavioral issues.

### Files changed
- `stream_registry.py` — Redis meta + stream key TTL refresh in `publish_chunk()`, module-level keepalive tracker
- `response_adapter.py` — `task_progress` SystemMessage → StreamHeartbeat emission
- `service.py` — Intermediate DB persistence in `_run_stream_attempt` stream loop, orphan cleanup on rollback
- `db.py` — `delete_messages_from_sequence` for rollback cleanup
- `prompting.py` — Web search best practices

### GCP log evidence
```
# Meta key expired during 82-min turn:
09:49 — GET_SESSION: active_session=False, msg_count=1  ← meta gone
10:18 — Session persisted in finally with 189 messages   ← turn completed

# T13 (1h45min) same bug reproduced live:
16:20 — task_progress events still arriving, but active_session=False

# Actual cost:
Turn usage: cache_read=347916, cache_create=212472, output=12375, cost_usd=20.66
```

### Test plan
- [x] task_progress emits StreamHeartbeat
- [x] Task background blocked, foreground allowed, slot release on completion/failure
- [x] CI green (lint, type-check, tests, e2e, CodeQL)


autogpt_platform/backend/backend/copilot/sdk/service.py
```diff
@@ -1536,6 +1544,34 @@ async def _run_stream_attempt(
...
+            # Flush session messages to DB periodically so page reloads
+            # show progress during long-running turns.
+            _msgs_since_flush += 1
+            now = time.monotonic()
+            if (
+                _msgs_since_flush >= _FLUSH_MESSAGE_THRESHOLD
+                or (now - _last_flush_time) >= _FLUSH_INTERVAL_SECONDS
+            ):
+                try:
+                    await asyncio.shield(upsert_chat_session(ctx.session))
```
majdyz, line 1503: 🤖 🔴 **Blocker**: Intermediate flush breaks retry rollback semantics.

When `_run_stream_attempt` flushes mid-stream via `upsert_chat_session(ctx.session)`, new messages are written to the DB with `add_chat_messages_batch`. If the attempt later fails and the retry loop rolls back `session.messages = session.messages[:pre_attempt_msg_count]` (service.py L2135), those already-persisted messages remain as orphans in the DB.

The next `upsert_chat_session` in the `finally` block uses `existing_message_count` from `get_next_sequence()` and only writes messages after that index — so the rolled-back messages stick around and resurface on page reload.

**Fix**: Guard the flush so it only runs on the final attempt, or track `pre_attempt_msg_count` and skip flushing messages beyond it on retry-eligible attempts. Simplest approach: pass `attempt_index` and `max_attempts` into `_run_stream_attempt` and only flush when `attempt_index == max_attempts - 1` (last chance), or gate it behind an `is_final_attempt` flag.

autogpt_platform/backend/backend/copilot/sdk/security_hooks.py
```diff
@@ -155,6 +161,12 @@ def create_security_hooks(
         # Set of tool_use_ids that consumed a slot — len() is the active count.
         task_tool_use_ids: set[str] = set()
 
+        # --- Circuit breaker counters ---
+        # Cap WebSearch calls per session (across all sub-agents) to prevent
+        # runaway research loops (see incident d2f7cba3: 179 searches, $20.66).
```
majdyz: 🤖 🟠 **Should Fix**: Counter scope mismatch — comments say "per session" but hooks are recreated per stream invocation.

`create_security_hooks` is called at L1922 of `service.py` (inside `stream_chat_completion_sdk`), which runs for each user message. The closure-local counters reset every time. The PR description and code comments say "per session" but these are effectively per-turn.

For v1 this is still useful (caps a single runaway turn), but the docstring/comments should be corrected to say "per turn" to avoid confusion. If true per-session persistence is needed later, counters should be stored externally (Redis or session model).

Specifically update:
- L166: "Cap WebSearch calls per session" → "per turn"  
- L167: comment referencing session → turn
- L145-146: docstring "per session" → "per turn"
- PR description if appropriate

autogpt_platform/backend/backend/copilot/sdk/security_hooks.py
```diff
@@ -192,6 +204,41 @@ async def pre_tool_use_hook(
                         ),
                     )
 
+            # --- Circuit breaker: total tool calls per turn ---
```
majdyz: 🤖 🟡 **Nice to Have**: Total tool call counter increments even for denied WebSearch calls.

The total tool call counter increments at L207 *before* the WebSearch cap check at L224. When a WebSearch call is denied at cap, it still counts against the total tool call budget. This means 15 denied WebSearch retries eat 15 slots from the 100 total cap.

Consider: move the WebSearch check *before* the total counter increment, or don't increment the total counter for denied calls.

autogpt_platform/backend/backend/copilot/sdk/security_hooks_test.py
```diff
@@ -318,6 +318,119 @@ async def test_task_slot_released_on_completion(_hooks):
     assert not _is_denied(result)
 
 
+# -- WebSearch cap -----------------------------------------------------------
+
+
+@pytest.fixture()
+def _hooks_search():
+    """Create security hooks with low search cap for testing."""
+    from .security_hooks import create_security_hooks
```
majdyz: 🤖 🔵 **Nit**: `_hooks_search` and `_hooks_tool_cap` fixtures import `create_security_hooks` locally.

Per CLAUDE.md: "Top-level imports only — no local/inner imports". The existing `_hooks` fixture does the same thing, so this is pre-existing, but the new fixtures follow the same anti-pattern. Move the import to module top-level (it's not a heavy optional dep — the SDK guard is already handled by `_sdk_available()` skipif).

autogpt_platform/backend/backend/copilot/sdk/service.py
majdyz: Fixed in 7c3bdb5.

autogpt_platform/backend/backend/copilot/sdk/security_hooks.py
majdyz: Fixed in 7c3bdb5. Updated all comments, docstrings, and denial messages from 'per session' to 'per turn'.

autogpt_platform/backend/backend/copilot/sdk/security_hooks.py
majdyz: Fixed in 7c3bdb5. Moved WebSearch cap check before total counter increment so denied searches don't consume total budget slots.

autogpt_platform/backend/backend/copilot/sdk/security_hooks_test.py
majdyz: Fixed in 7c3bdb5. Moved create_security_hooks import to module top-level.

autogpt_platform/backend/backend/copilot/sdk/security_hooks.py
```diff
@@ -155,6 +161,12 @@ def create_security_hooks(
         # Set of tool_use_ids that consumed a slot — len() is the active count.
         task_tool_use_ids: set[str] = set()
 
+        # --- Circuit breaker counters ---
+        # Cap WebSearch calls per session (across all sub-agents) to prevent
+        # runaway research loops (see incident d2f7cba3: 179 searches, $20.66).
+        web_search_count: list[int] = [0]  # mutable container for closure
+        total_tool_call_count: list[int] = [0]  # total tool calls this turn
+
```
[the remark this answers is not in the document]
majdyz: Good catch. Fixed in 7c3bdb5 — updated comments/docstrings to say 'per turn' since hooks are recreated per stream invocation. For v1 the per-turn scope is sufficient (caps a single runaway turn). True per-session persistence via Redis can be added later if needed.

autogpt_platform/backend/backend/copilot/sdk/service.py
```diff
@@ -1536,6 +1544,34 @@ async def _run_stream_attempt(
...
+                    )
+                except Exception as flush_err:
+                    logger.warning(
+                        "%s Intermediate flush failed: %s",
+                        ctx.log_prefix,
+                        flush_err,
+                    )
+                _last_flush_time = now
+                _msgs_since_flush = 0
+
```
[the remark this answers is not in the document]
majdyz: Fixed in 7c3bdb5. Added is_final_attempt flag to _RetryState so intermediate flush only runs on the last retry attempt. Earlier attempts that may be rolled back will not persist orphan messages to the DB.

autogpt_platform/backend/backend/copilot/prompting.py
```diff
@@ -107,6 +107,14 @@
 After building the file, reference it with `@@agptfile:` in other tools:
 `@@agptfile:/home/user/report.md`
 
+### Web search best practices
+- If 3 similar web searches don't return the specific data you need, conclude
+  it isn't publicly available and work with what you have.
+- Prefer fewer, well-targeted searches over many variations of the same query.
+- When spawning sub-agents for research, ensure each has a distinct
+  non-overlapping scope to avoid redundant searches.
+- There is a hard cap on web searches per session — use them wisely.
```
majdyz: 🤖 🔵 **Nit**: Prompt still says 'per session' at L116 of prompting.py. Since the cap resets per turn, this is slightly inaccurate. Suggest: 'There is a hard cap on web searches — use them wisely.' (drop the scope qualifier entirely since the LLM doesn't need to know the reset behavior).

majdyz: Fixed in 4b5f8b2. Changed to 'There is a hard cap on web searches — use them wisely.' to avoid implying a specific scope.

autogpt_platform/backend/backend/copilot/sdk/service.py
```diff
@@ -1536,6 +1545,36 @@ async def _run_stream_attempt(
...
                 )
 
+            # --- Intermediate persistence ---
+            # Flush session messages to DB periodically so page reloads
+            # show progress during long-running turns.
+            # Only flush on the final attempt to avoid persisting messages
+            # that may be rolled back if a transient error triggers a retry.
+            _msgs_since_flush += 1
+            now = time.monotonic()
+            if state.is_final_attempt and (
```
majdyz: 🤖 🟠 **Should Fix**: is_final_attempt=False on attempt 0 disables intermediate flush for the common case (no retries).

With _MAX_STREAM_ATTEMPTS=3, attempt 0 has is_final_attempt=False, so intermediate persistence is disabled for the 99% case where the first attempt succeeds. This defeats the purpose of the feature.

Better approach: always allow flushing, but track the pre-attempt message count and on retry rollback, also delete DB messages beyond that count. Or simpler: flush optimistically, and if a retry rollback happens, re-persist the rolled-back session state to DB to overwrite the stale messages.

majdyz: Fixed in ce231fd. Removed is_final_attempt guard — intermediate flush is now enabled for all attempts. The retry path (context-too-long errors with events_yielded==0) means the stream barely started, so the flush threshold is almost certainly not reached. Added a documenting comment explaining why orphan messages are not a practical risk.

autogpt_platform/backend/backend/copilot/sdk/service.py
majdyz: Revised in ce231fd. Removed is_final_attempt approach since it disabled flush for the common case. Intermediate flush now runs on all attempts. The retry-rollback edge case is documented but practically unreachable (retries only fire on context-too-long with events_yielded==0).

Otto-AGPT (commented):
Reviewed the diff. Overall this is solid and addresses both issues from the d2f7cba3 incident. A few observations:

**Circuit breakers (security_hooks.py) ✅**
- Ordering is correct: denied WebSearch calls don't consume total tool budget (early return before total counter increment)
- Counter uses `list[int]` for closure mutability — works fine, consistent with how `task_tool_use_ids` is tracked in the same closure
- Cap values (15 searches, 100 total) seem reasonable. For reference, the incident session had 179 searches — this would have cut it to 15 and saved $15+ and ~70 minutes

**One thing to verify:** Where is `create_security_hooks` called relative to retries in `_run_stream_attempt`? If hooks are shared across retry attempts, the counters persist — so after a failed attempt that used 10 searches, the retry only gets 5 more. This is probably fine since retries only happen on context-too-long with `events_yielded==0` (meaning counters are barely incremented), but worth being aware of.

**Intermediate persistence (service.py) ✅**
- `asyncio.shield` protects the flush from cancellation — good
- Rollback note for orphaned messages is accurate and the risk is minimal
- Minor observation: `_msgs_since_flush` increments for every SDK stream event (including text deltas), not just meaningful UI messages. During rapid text streaming, the 10-message threshold could trigger frequent flushes. In practice the 30s time interval is the effective throttle, so this isn't a real issue — but if you wanted to tighten it, you could count only tool outputs or assistant message completions

**Prompt guidance (prompting.py) ✅**
- Clear and actionable

**Tests ✅**
- Good coverage for both caps (under cap allowed, at cap denied)
- Tests correctly verify denial messages contain 'synthesize' and 'per turn'

**Not addressed (separate follow-up):**
- The GCP logs confirmed WebSearch calls are genuinely 29-54s each — moving to our own Brave Search MCP tool would be the real perf fix (20x improvement). This PR is the right short-term mitigation though.
- Redis TTL expiring before long turns complete (tracked in SECRT-2178)

autogpt_platform/backend/backend/copilot/sdk/security_hooks.py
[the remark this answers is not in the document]
majdyz: 🤖 Acknowledged — the counters are intentionally per-turn (per `create_security_hooks` call), not per-session. This was clarified in commit 7c3bdb5 where all comments, docstrings, and denial messages were updated to say "per turn" instead of "per session".

Per-turn scope is the right v1 trade-off:
- Still prevents the worst-case runaway (the d2f7cba3 incident was a single turn)
- Avoids needing Redis persistence for counter state
- Per-session Redis persistence can be layered on later if cross-turn accumulation becomes a concern

autogpt_platform/backend/backend/copilot/sdk/service.py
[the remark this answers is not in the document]
majdyz: 🤖 This was addressed in commits 7c3bdb5 and ce231fd. The intermediate flush only runs on the final attempt (`state.is_final_attempt` guard). Earlier rollbackable attempts never write to DB via this path.

For the final attempt, if it fails after flushing, the messages are already in DB — which is actually the desired behavior (user sees progress on reload rather than a blank page). The "orphaned messages" risk is documented in the code comment and is practically unreachable since retries only trigger on context-too-long with 0 events yielded.

autogpt_platform/backend/backend/copilot/sdk/security_hooks.py
```diff
@@ -155,6 +162,13 @@ def create_security_hooks(
         # Set of tool_use_ids that consumed a slot — len() is the active count.
         task_tool_use_ids: set[str] = set()
 
+        # --- Circuit breaker counters ---
+        # Cap WebSearch calls per turn (across all sub-agents) to prevent
+        # runaway research loops (see incident d2f7cba3: 179 searches, $20.66).
+        # Note: hooks are recreated per stream invocation so these reset each turn.
+        web_search_count: list[int] = [0]  # mutable container for closure
+        total_tool_call_count: list[int] = [0]  # total tool calls this turn
+
```
majdyz: 🤖 Addressing Otto's review observations:

**Re: `create_security_hooks` called relative to retries:**
Good catch. Hooks are created once per `_run_stream_attempt` call, which is called per retry attempt. So after a failed attempt that used 10 searches, the retry gets a fresh counter with full budget. This is fine because retries only happen on context-too-long with `events_yielded==0` (counters barely incremented).

**Re: `_msgs_since_flush` counting all SDK events:**
Correct — the 30s time interval is the effective throttle in practice. Text delta events won't trigger frequent flushes because the 30s check comes first. If we want to tighten later, counting only tool outputs or assistant completions is a good refinement.

**Re: MCP Brave Search as the real perf fix:**
Agreed — the Langfuse prompt has been updated (v23, live on dev) to steer the LLM toward PerplexityBlock for research tasks instead of repeated WebSearch. This PR is the safety net; the Langfuse prompt change addresses the behavior. Moving to our own Brave MCP tool is a follow-up.

**Re: Redis TTL (SECRT-2178):**
Tracked separately as noted.

autogpt_platform/backend/backend/copilot/stream_registry.py
```diff
@@ -221,9 +221,21 @@ async def create_session(
...
 
 
+_meta_ttl_refresh_at: dict[str, float] = {}
+"""Tracks the last time the session meta key TTL was refreshed.
+
+Used by `publish_chunk` to avoid refreshing on every single chunk
+(expensive). Refreshes at most once every 60 seconds per session.
+"""
+
+_META_TTL_REFRESH_INTERVAL = 60  # seconds
```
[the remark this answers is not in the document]
majdyz: Fixed in 751382fcff — `mark_session_completed` now calls `_meta_ttl_refresh_at.pop(session_id, None)` to clean up the entry when a session completes, preventing unbounded growth.

majdyz:
## Code Review -- PR #12604

### Summary
This PR adds: (1) WebSearch + total tool call circuit breakers in security hooks, (2) intermediate session persistence (flush to DB every 30s/10 msgs), (3) StreamHeartbeat on task_progress events to keep Redis TTLs alive, (4) Task tool fully blocked via SDK_DISALLOWED_TOOLS + BLOCKED_TOOLS, and (5) web search best practice prompting guidance.

### Findings

**`[info]` `approved` -- Circuit breaker ordering is correct**
WebSearch denial returns before `total_tool_call_count` is incremented (line 220 returns before line 230), so denied searches don't consume total budget. Well-tested in `test_websearch_denials_do_not_consume_total_cap`.

**`[info]` `defense-in-depth` -- Task rate-limiting code is now dead but harmless**
Task is in BLOCKED_TOOLS, so `_validate_tool_access` at line 253 will always block it before the rate-limiting at lines 182-207 can deny on slot exhaustion. The old rate-limiting code path never triggers for Task anymore (it falls through to BLOCKED_TOOLS). This is fine as defense-in-depth, but could be simplified in a future cleanup.

**`[low]` `memory` -- `_meta_ttl_refresh_at` memory leak fixed**
Addressed in 751382fcff -- `mark_session_completed` now calls `_meta_ttl_refresh_at.pop(session_id, None)`.

**`[info]` `approved` -- Intermediate persistence is well-guarded**
`asyncio.shield()` protects the flush from cancellation. Failed flushes are caught and logged without breaking the stream. The retry note about orphaned intermediate flushes is honest and accurate (threshold won't be reached with events_yielded==0).

**`[info]` `approved` -- StreamHeartbeat on task_progress is correct**
Converts previously silent task_progress events into heartbeats that trigger `publish_chunk`, keeping Redis TTLs alive during long sub-agent runs.

### Verdict
Clean PR. All circuit breaker tests pass, ordering is correct, memory leak fixed. No bugs found.

autogpt_platform/backend/backend/copilot/sdk/security_hooks.py
```diff
@@ -192,6 +206,43 @@ async def pre_tool_use_hook(
                         ),
                     )
 
+            # --- Circuit breaker: WebSearch calls per turn ---
+            # Check WebSearch cap before incrementing the total counter so
+            # denied searches don't consume slots from the total budget.
+            if tool_name == "WebSearch":
+                web_search_count[0] += 1
+                if web_search_count[0] > max_web_searches:
```
majdyz: 🤖 🟠 **Should Fix**: No proactive warning before hitting the cap — the model only learns it's out of budget when a call is denied.

Consider adding a warning in the successful tool result when approaching the limit (e.g., at 80%: 40/50 web searches or 400/500 total tool calls). Something like injecting a note in the hook output:

```
"⚠️ You have used 42/50 web searches this turn. Plan remaining searches carefully."
```

This lets the model budget its remaining calls instead of hitting a hard wall. Without this, the model wastes a call on the denied attempt and then has to course-correct reactively.

The SDK `PreToolUse` hook can return extra text in the `hookSpecificOutput` alongside the allowed result — could append a warning message there when the counter crosses the threshold.

autogpt_platform/backend/backend/copilot/sdk/service.py
```diff
@@ -1536,6 +1544,34 @@ async def _run_stream_attempt(
...
+                        "%s Intermediate flush failed: %s",
+                        ctx.log_prefix,
+                        flush_err,
+                    )
+                _last_flush_time = now
+                _msgs_since_flush = 0
+
             if acc.stream_completed:
                 break
     finally:
```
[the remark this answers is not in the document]
majdyz: Acknowledged — the code already documents this edge case (line 2168-2172): orphaned messages from intermediate flushes can remain in DB after rollback. In practice this is unlikely because retries only fire on context-too-long errors with `events_yielded==0` (stream barely started, flush threshold not reached). The risk is accepted and documented. A follow-up DB cleanup task could be added if this surfaces in production.

autogpt_platform/backend/backend/copilot/sdk/security_hooks.py
majdyz: Fixed in cbd3ebce00. Added proactive budget warnings using `PreToolUseHookSpecificOutput.additionalContext`:

- **WebSearch**: warns at 80% usage (e.g., "Web search budget: 8/50 remaining this turn")
- **Total tool calls**: warns at 80% usage (e.g., "Total tool call budget: 100/500 remaining this turn")
- Also bumped `max_tool_calls` default from 100 → 500 (10x the web search cap) in 2d00268516

The model now gets budget awareness before hitting the wall, allowing it to prioritize remaining calls.

autogpt_platform/backend/backend/copilot/sdk/service.py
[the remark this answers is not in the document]
majdyz: Fixed in 77112e79a2. Added `delete_messages_from_sequence()` to `db.py` — on rollback, the retry path now deletes any messages with `sequence >= pre_attempt_msg_count` from the DB, matching the in-memory rollback. The cleanup is wrapped in a try/except so a failure doesn't block the retry.

autogpt_platform/backend/backend/copilot/sdk/service.py
```diff
@@ -2129,6 +2166,18 @@ def _on_stderr(line: str) -> None:
...
+                    )
+                except Exception as cleanup_err:
+                    logger.warning(
+                        "%s Failed to clean up orphaned messages: %s",
+                        log_prefix,
+                        cleanup_err,
+                    )
                 # Cancel any pre-launched tasks from the failed attempt so they
                 # don't continue executing against the rolled-back session.
                 await cancel_pending_tool_tasks()
```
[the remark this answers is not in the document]
majdyz: This is the same edge case documented at line 2168-2172 of service.py. Intermediate DB flushes can leave orphaned messages after rollback, but this is extremely unlikely in practice — retries only fire on context-too-long errors with `events_yielded==0` (stream barely started, flush threshold not reached). The risk is documented and accepted. A DB cleanup sweep could be added as a follow-up if this surfaces in production monitoring.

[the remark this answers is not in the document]
majdyz: Fixed in 4df0714e2a: Added `delete_messages_from_sequence()` call to the `_HandledStreamError` handler, matching the cleanup already present in the generic `Exception` handler. Orphaned DB messages from intermediate flushes are now cleaned up during rollback in both paths.

autogpt_platform/backend/backend/copilot/sdk/security_hooks.py
majdyz: Reverted in 053afde64d. After further discussion, the circuit breakers (WebSearch cap, total tool call cap) and Task disabling were band-aid fixes that limited capability rather than addressing root causes. The real bug fixes (Redis TTL refresh, intermediate persistence, orphan cleanup) remain. Prompt guidance for search best practices provides sufficient behavioral guardrails.

autogpt_platform/backend/backend/copilot/sdk/response_adapter_test.py
```diff
@@ -59,6 +59,16 @@ def test_system_non_init_emits_nothing():
     assert results == []
 
 
+def test_task_progress_emits_heartbeat():
+    """task_progress events emit a StreamHeartbeat to keep Redis TTL alive."""
+    from backend.copilot.response_model import StreamHeartbeat
```
majdyz: 🤖 🔵 **Nit**: Local import inside test function — `from backend.copilot.response_model import StreamHeartbeat` at L64. Per CLAUDE.md convention, top-level imports are preferred. `StreamHeartbeat` is already imported at the top of `response_adapter.py` so there's no circular dependency concern.

majdyz: Fixed in ffcb88251a. Moved StreamHeartbeat import to module top-level.

ntindle (approved):

autogpt_platform/backend/backend/copilot/sdk/service.py
```diff
@@ -1482,6 +1490,34 @@ async def _run_stream_attempt(
                     model=sdk_msg.model,
                 )
 
+            # --- Intermediate persistence ---
+            # Flush session messages to DB periodically so page reloads
+            # show progress during long-running turns.
+            _msgs_since_flush += 1
+            now = time.monotonic()
+            if (
```
majdyz, line 1498: 🤖 🟠 **Should Fix**: The intermediate flush (line 1493-1519) lacks the `is_final_attempt` guard that was described as fixed in the review thread. The current code unconditionally flushes every 30s/10 messages regardless of which retry attempt is running.

In practice, the retry path only activates when `events_yielded == 0` AND the error is a context-length error, which means the intermediate flush is unlikely to have fired (the error typically happens before any SDK messages are processed). However, there's an edge case: if many `StreamHeartbeat` events are processed (they increment `_msgs_since_flush` at line 1496 but don't increment `events_yielded` at line 2047-2048), the 30s timer or 10-message threshold could be reached, persisting messages to DB. On rollback (line 2081/2113), the in-memory messages are trimmed but the DB-persisted orphans remain.

Simplest fix: add a guard that only flushes on the last attempt, matching what the review reply described:
```python
is_final_attempt = (attempt == _MAX_STREAM_ATTEMPTS - 1)
```
and wrap the flush block with `if is_final_attempt:`. Alternatively, pass `attempt` into `_run_stream_attempt` and check there.

majdyz: 🤖 Fixed in 79bc0aed9: Added `is_final_attempt` field to `_RetryState`. The flag is set to True on attempt 0 (optimistic — most turns succeed first try) and on the last retry attempt. Middle retry attempts have it set to False, preventing intermediate DB flushes that could persist orphan messages surviving the in-memory rollback.

PR #12612, opened 2026-03-31
feat(agentation): integrate Agentation devtool into the frontend

Abhi1992002:
## Summary
- Adds the `agentation` package (v3.0.2) as a devDependency
- Integrates the `<Agentation />` devtool component into the root layout, gated behind `NEXT_PUBLIC_AGENTATION_DEVTOOL` env var
- Adds `NEXT_PUBLIC_AGENTATION_DEVTOOL=true` to `.env.default` for local dev

## Changes
- `autogpt_platform/frontend/package.json` — added `agentation@3.0.2` to devDependencies
- `autogpt_platform/frontend/src/app/layout.tsx` — import and render `<Agentation />` when env var is set
- `autogpt_platform/frontend/.env.default` — added `NEXT_PUBLIC_AGENTATION_DEVTOOL=true`

## Test plan
- [x] Verify `pnpm install` resolves without errors
- [x] Verify the Agentation toolbar appears in local dev (`NEXT_PUBLIC_AGENTATION_DEVTOOL=true`)
- [x] Verify the toolbar does NOT appear when the env var is unset/false
- [x] Verify no impact on production builds (devDependency only)


autogpt_platform/frontend/src/app/layout.tsx
```diff
@@ -78,6 +79,7 @@ export default async function RootLayout({
             </div>
             <Toaster />
             <CookieConsentBanner />
+            {process.env.NEXT_PUBLIC_AGENTATION_DEVTOOL && <Agentation />}
```
0ubbe: I think we need to use a dynamic import, otherwise it won't be properly tree shaken in production and dev 💡

0ubbe: I would add it initially only for local 🤔 , or local and dev, you have the `services/environment` to make these checks 👌🏽

Abhi1992002: Fixed in 060b96818: switched to `next/dynamic` import with `ssr: false` for proper tree-shaking.

Abhi1992002: Fixed in 060b96818: now gated behind `(isLocal || isDev)` using the environment service, in addition to the env var check.

autogpt_platform/frontend/.env.default
```diff
@@ -17,6 +17,7 @@ NEXT_PUBLIC_LAUNCHDARKLY_CLIENT_ID=687ab1372f497809b131e06e
 
 # Debugging
 NEXT_PUBLIC_REACT_QUERY_DEVTOOL=true
+NEXT_PUBLIC_AGENTATION_DEVTOOL=true
```
ntindle, line 20: why would this be disableable independent of is local and is dev?

0ubbe: Agree, in my view we should only have it enabled in Local and Dev, and use the existing `environment` service which gives you `isLocal()` and `isDev()` helpers.

PR #12626, opened 2026-03-31
fix(backend/copilot): CLI ResultMessage "Prompt is too long" bypasses compaction

majdyz:
## Why

Long-running CoPilot sessions hit "Prompt is too long" and get permanently stuck — compaction never triggers, so users can't continue the conversation. This was reported on session `08b807d4-114f-46b7-9b9e-7fb6b35481a9` (Langfuse).

## Root Cause Investigation

Traced via Langfuse OTEL data for the affected session:

1. **Turn 9** was massive — 100+ tool calls (file writes, reads, bash exec, run_block) spanning hours on March 13
2. **Turn 13** (today, user typed "hi") resumed the session. The Claude CLI did a **pre-API check** and detected the prompt exceeded the context limit
3. The CLI returned `ResultMessage(subtype="error", result="Prompt is too long")` with `duration_api_ms=0` and model `<synthetic>` — meaning no API call was ever made
4. Our code in `_run_stream_attempt` received this as a **normal message**, logged the error, and the adapter converted it to `StreamError` + `StreamFinish`
5. The retry loop saw a "successful" stream completion and `break`ed — **compaction was never triggered**

The retry/compaction mechanism at `service.py:2117-2153` only catches **Python exceptions**. When the CLI returns the error as a structured `ResultMessage` instead of raising, the entire compaction path is bypassed.

### Fix 1: Re-raise CLI prompt-too-long as exception (`service.py`)

When a `ResultMessage` with `subtype="error"` has result text matching `_PROMPT_TOO_LONG_PATTERNS` (e.g. "prompt is too long"), we now `raise Exception(...)` **before** the adapter converts it to stream events. This ensures:
- The exception propagates to the retry loop
- `_is_prompt_too_long(e)` matches → `is_context_error=True`
- `events_yielded == 0` (CLI returned immediately) → retry is allowed
- `_reduce_context()` runs compaction on the next attempt

### Fix 2: Strip stale thinking blocks from transcript (`transcript.py`)

Adds `strip_stale_thinking_blocks()` — removes thinking/redacted_thinking blocks from **non-last** assistant entries before transcript upload. These blocks consume 10-50K tokens each but carry no context value for `--resume`. Stripping them prevents the CLI from triggering compaction every turn just to compress away stale thinking bloat.

The Anthropic API only requires thinking blocks in the **last** assistant message to be byte-identical — older entries can safely have them removed.

## How

- **`service.py`**: Added check after `ResultMessage` error logging (line ~1406) — if the error text matches any `_PROMPT_TOO_LONG_PATTERNS`, raise before the adapter processes it
- **`transcript.py`**: New `strip_stale_thinking_blocks()` function + call in `upload_transcript()` after `strip_progress_entries()`
- **`retry_scenarios_test.py`**: New integration test `test_result_message_prompt_too_long_triggers_compaction` — simulates CLI returning error ResultMessage, asserts compaction retry happens (attempt_count==2) with no StreamError

## Test plan

- [ ] New test: `test_result_message_prompt_too_long_triggers_compaction` — CLI error ResultMessage triggers compaction retry
- [ ] Existing test: `test_prompt_too_long_retries_with_compaction` — exception path still works
- [ ] Existing test: `test_all_attempts_exhausted_yields_stream_error` — exhaustion path still works
- [ ] Existing test: `test_events_yielded_prevents_retry` — mid-stream errors still break
- [ ] Existing test: `test_non_context_error_breaks_immediately` — non-context errors still break
- [ ] CI green

### DIFF
--- a/autogpt_platform/backend/backend/copilot/sdk/retry_scenarios_test.py
+++ b/autogpt_platform/backend/backend/copilot/sdk/retry_scenarios_test.py
@@ -1414,3 +1414,76 @@ class TestStreamChatCompletionRetryIntegration:
         # Verify user-friendly message (not raw SDK text)
         assert "Authentication" in errors[0].errorText
         assert any(isinstance(e, StreamStart) for e in events)
+
+    @pytest.mark.asyncio
+    async def test_result_message_prompt_too_long_triggers_compaction(self):
+        """CLI returns ResultMessage(subtype="error") with "Prompt is too long".
+
+        When the Claude CLI rejects the prompt pre-API (model=<synthetic>,
+        duration_api_ms=0), it sends a ResultMessage with is_error=True
+        instead of raising a Python exception.  The retry loop must still
+        detect this as a context-length error and trigger compaction.
+        """
+        import contextlib
+
+        from claude_agent_sdk import ResultMessage
+
+        from backend.copilot.response_model import StreamError, StreamStart
+        from backend.copilot.sdk.service import stream_chat_completion_sdk
+
+        session = self._make_session()
+        success_result = self._make_result_message()
+        attempt_count = [0]
+
+        error_result = ResultMessage(
+            subtype="error",
+            result="Prompt is too long",
+            duration_ms=100,
+            duration_api_ms=0,
+            is_error=True,
+            num_turns=0,
+            session_id="test-session-id",
+        )
+
+        def _client_factory(*args, **kwargs):
+            attempt_count[0] += 1
+            if attempt_count[0] == 1:
+                # First attempt: CLI returns error ResultMessage
+                return self._make_client_mock(result_message=error_result)
+            # Second attempt (after compaction): succeeds
+            return self._make_client_mock(result_message=success_result)
+
+        original_transcript = _build_transcript(
+            [("user", "prior question"), ("assistant", "prior answer")]
+        )
+        compacted_transcript = _build_transcript(
+            [("user", "[summary]"), ("assistant", "summary reply")]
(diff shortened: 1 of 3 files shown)


majdyz:
Consolidated into #12625 — that PR now includes all three error paths (AssistantMessage.error, ResultMessage.result, and Python exceptions) plus the strip_stale_thinking_blocks fix, compression target reduction, tests, and download script.

majdyz (commented):
**Review: LGTM. Clean, focused fix.**

**Changes reviewed:**
- `ResultMessage` prompt-too-long detection in `_run_stream_attempt`: catches CLI pre-API rejection where `is_error=True` and result contains "prompt is too long" pattern
- Re-raises as `Exception` so the outer retry loop triggers compaction
- `strip_stale_thinking_blocks()` in `transcript.py` (shared with #12625): strips thinking/redacted_thinking blocks from non-last assistant entries
- Integration test `test_result_message_prompt_too_long_triggers_compaction` validates the full retry + compaction flow

**Notes:**
- This PR overlaps with #12625 on `transcript.py` (identical `strip_stale_thinking_blocks`) and `service.py` (both add ResultMessage prompt-too-long handling at the same location). Whichever merges second will need conflict resolution.
- PR #12625 additionally handles the `AssistantMessage.error` path -- this PR focuses only on `ResultMessage`, which is the more common CLI rejection path.

**No bugs, security issues, or breaking changes found.** The test correctly validates that `attempt_count == 2` (error -> compaction -> retry -> success).

PR #12628, opened 2026-03-31
fix(backend): fix copilot credential loading across event loops

majdyz:
## Why

CoPilot autopilot sessions are inconsistently failing to load user credentials (specifically GitHub OAuth). Some sessions proceed normally, some show "provide credentials" prompts despite the user having valid creds, and some are completely blocked.

Production logs confirmed the root cause: `RuntimeError: Task got Future <Future pending> attached to a different loop` in the credential refresh path, cascading into null-cache poisoning that blocks credential lookups for 60 seconds.

## What

Three interrelated bugs in the credential system:

1. **`refresh_if_needed` always acquired Redis locks even with `lock=False`** — The `lock` parameter only controlled the inner credential lock, but the outer "refresh" scope lock was always acquired. The copilot executor uses multiple worker threads with separate event loops; the `asyncio.Lock` inside `AsyncRedisKeyedMutex` was bound to one loop and failed on others.

2. **Stale event loop in `locks()` singleton** — Both `IntegrationCredentialsManager` and `IntegrationCredentialsStore` cached their `AsyncRedisKeyedMutex` without tracking which event loop created it. When a different worker thread (with a different loop) reused the singleton, it got the "Future attached to different loop" error.

3. **Null-cache poisoning on refresh failure** — When OAuth refresh failed (due to the event loop error), the code fell through to cache "no credentials found" for 60 seconds via `_null_cache`. This blocked ALL subsequent credential lookups for that user+provider, even though the credentials existed and could refresh fine on retry.

## How

- Split `refresh_if_needed` into `_refresh_locked` / `_refresh_unlocked` so `lock=False` truly skips ALL Redis locking (safe for copilot's best-effort background injection)
- Added event loop tracking to `locks()` in both `IntegrationCredentialsManager` and `IntegrationCredentialsStore` — recreates the mutex when the running loop changes
- Only populate `_null_cache` when the user genuinely has no credentials; skip caching when OAuth refresh failed transiently
- Updated existing test to verify null-cache is not poisoned on refresh failure

## Test plan

- [x] All 14 existing `integration_creds_test.py` tests pass
- [x] Updated `test_oauth2_refresh_failure_returns_none_without_null_cache` verifies null-cache is not populated on refresh failure
- [x] Format, lint, and typecheck pass
- [ ] Deploy to staging and verify copilot sessions consistently load GitHub credentials

### DIFF
--- a/autogpt_platform/backend/backend/copilot/integration_creds_test.py
+++ b/autogpt_platform/backend/backend/copilot/integration_creds_test.py
@@ -195,25 +195,14 @@ class TestRefreshFailureDoesNotCacheNull:
         be populated — the next call should retry the refresh."""
         oauth_creds = _make_oauth2_creds()
 
-        mock_store = AsyncMock()
-        mock_store.get_all_creds = AsyncMock(return_value=[oauth_creds])
-
-        mock_manager = AsyncMock()
+        mock_manager = MagicMock()
+        mock_manager.store.get_creds_by_provider = AsyncMock(return_value=[oauth_creds])
         # Simulate refresh failure (e.g. transient network error)
         mock_manager.refresh_if_needed = AsyncMock(
             side_effect=Exception("Network timeout")
         )
 
-        with (
-            patch(
-                "backend.copilot.integration_creds._get_store",
-                return_value=mock_store,
-            ),
-            patch(
-                "backend.copilot.integration_creds._get_manager",
-                return_value=mock_manager,
-            ),
-        ):
+        with patch("backend.copilot.integration_creds._manager", mock_manager):
             result = await get_provider_token(_USER, _PROVIDER)
 
         # Token should be None (refresh failed)
@@ -228,13 +217,10 @@ class TestRefreshFailureDoesNotCacheNull:
     @pytest.mark.asyncio(loop_scope="session")
     async def test_no_credentials_at_all_does_cache_null(self):
         """When the user truly has no credentials, null-cache should be set."""
-        mock_store = AsyncMock()
-        mock_store.get_all_creds = AsyncMock(return_value=[])
+        mock_manager = MagicMock()
+        mock_manager.store.get_creds_by_provider = AsyncMock(return_value=[])
 
-        with patch(
-            "backend.copilot.integration_creds._get_store",
-            return_value=mock_store,
-        ):
+        with patch("backend.copilot.integration_creds._manager", mock_manager):
             result = await get_provider_token(_USER, _PROVIDER)
 
         assert result is None
--- a/autogpt_platform/backend/backend/copilot/integration_creds_test.py
+++ b/autogpt_platform/backend/backend/copilot/integration_creds_test.py
@@ -185,51 +185,6 @@ class TestGetProviderToken:
         assert _NULL_CACHE_TTL < _TOKEN_CACHE_TTL
 
 
-class TestRefreshFailureDoesNotCacheNull:
-    """Bug reproduction: transient refresh failure was cached as 'no credentials'
-    for 60s, blocking all subsequent token requests for the provider."""
-
-    @pytest.mark.asyncio(loop_scope="session")
-    async def test_transient_refresh_failure_does_not_poison_null_cache(self):
-        """After a refresh fails (e.g. network error), the null-cache must NOT
-        be populated — the next call should retry the refresh."""
-        oauth_creds = _make_oauth2_creds()
-
-        mock_manager = MagicMock()
-        mock_manager.store.get_creds_by_provider = AsyncMock(return_value=[oauth_creds])
-        # Simulate refresh failure (e.g. transient network error)
-        mock_manager.refresh_if_needed = AsyncMock(
-            side_effect=Exception("Network timeout")
-        )
-
-        with patch("backend.copilot.integration_creds._manager", mock_manager):
-            result = await get_provider_token(_USER, _PROVIDER)
-
-        # Token should be None (refresh failed)
-        assert result is None
-        # But null-cache must NOT be populated — next call should retry
-        cache_key = (_USER, _PROVIDER)
-        assert cache_key not in _null_cache, (
-            "Null-cache was poisoned after transient refresh failure. "
-            "This blocks retries for 60s."
-        )
-
-    @pytest.mark.asyncio(loop_scope="session")
-    async def test_no_credentials_at_all_does_cache_null(self):
-        """When the user truly has no credentials, null-cache should be set."""
-        mock_manager = MagicMock()
-        mock_manager.store.get_creds_by_provider = AsyncMock(return_value=[])
-
-        with patch("backend.copilot.integration_creds._manager", mock_manager):
-            result = await get_provider_token(_USER, _PROVIDER)
-
-        assert result is None
-        cache_key = (_USER, _PROVIDER)
-        assert (
-            cache_key in _null_cache
-        ), "Null-cache should be set when user has no credentials at all."
-
-
 class TestThreadSafetyLocks:
     """Bug reproduction: shared AsyncRedisKeyedMutex across threads caused
     'Future attached to a different loop' when copilot workers accessed
--- a/autogpt_platform/backend/backend/copilot/integration_creds_test.py
+++ b/autogpt_platform/backend/backend/copilot/integration_creds_test.py
@@ -129,8 +129,15 @@ class TestGetProviderToken:
         assert result == "oauth-tok"
 
     @pytest.mark.asyncio(loop_scope="session")
-    async def test_oauth2_refresh_failure_returns_none(self):
-        """On refresh failure, return None instead of caching a stale token."""
+    async def test_oauth2_refresh_failure_returns_none_without_null_cache(self):
+        """On refresh failure, return None but do NOT cache in null_cache.
+
+        The user has credentials — they just couldn't be refreshed right now
+        (e.g. transient network error or event-loop mismatch in the copilot
+        executor).  Caching a negative result would block all credential
+        lookups for 60 s even though the creds exist and may refresh fine
+        on the next attempt.
+        """
         oauth_creds = _make_oauth2_creds("stale-oauth-tok")
         mock_manager = MagicMock()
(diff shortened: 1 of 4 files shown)


autogpt_platform/backend/backend/integrations/creds_manager.py
```diff
@@ -198,16 +218,13 @@ async def refresh_if_needed(
                 logger.debug(
                     f"Refreshing '{credentials.provider}' credentials #{credentials.id}"
                 )
-                _lock = None
-                if lock:
-                    # Wait until the credentials are no longer in use anywhere
-                    _lock = await self._acquire_lock(user_id, credentials.id)
+                # Wait until the credentials are no longer in use anywhere
```
majdyz, line 218: 🤖 🟠 **Should Fix**: The inner `_lock` acquired via `_acquire_lock` (line 221) is not wrapped in `try/finally`. If `oauth_handler.refresh_tokens()` or `self.store.update_creds()` raises, the lock is held until its 60s TTL expires, blocking other credential operations.

This is a pre-existing bug (the old code had the same issue when `lock=True`), but since we're refactoring this method, we should fix it now:
```python
_lock = await self._acquire_lock(user_id, credentials.id)
try:
    fresh_credentials = await oauth_handler.refresh_tokens(credentials)
    await self.store.update_creds(user_id, fresh_credentials)
    _invoke_creds_changed_hook(user_id, fresh_credentials.provider)
finally:
    if (await _lock.locked()) and (await _lock.owned()):
        try:
            await _lock.release()
        except Exception:
            logger.warning("Failed to release OAuth refresh lock", exc_info=True)
```

autogpt_platform/backend/backend/integrations/creds_manager.py
```diff
@@ -219,6 +236,31 @@ async def refresh_if_needed(
...
+        each thread has its own event loop and sharing Redis-backed asyncio locks
+        is not possible.  Concurrent refreshes are tolerated: the last writer
+        wins, and stale tokens are overwritten.
+        """
+        if provider_matches(credentials.provider, ProviderName.MCP.value):
+            oauth_handler = create_mcp_oauth_handler(credentials)
+        else:
+            oauth_handler = await _get_provider_oauth_handler(credentials.provider)
+        if oauth_handler.needs_refresh(credentials):
+            logger.debug(
```
majdyz, line 248: 🤖 🔵 **Nit**: Per CLAUDE.md, `debug` log statements should use `%s` for deferred interpolation. This applies to both `_refresh_locked` (line 218) and `_refresh_unlocked` (line 254).

autogpt_platform/backend/backend/integrations/creds_manager.py
```diff
@@ -198,16 +218,13 @@ async def refresh_if_needed(
...
+                _lock = await self._acquire_lock(user_id, credentials.id)
 
                 fresh_credentials = await oauth_handler.refresh_tokens(credentials)
                 await self.store.update_creds(user_id, fresh_credentials)
-                # Notify listeners so the refreshed token is picked up immediately.
                 _invoke_creds_changed_hook(user_id, fresh_credentials.provider)
-                if _lock and (await _lock.locked()) and (await _lock.owned()):
+                if (await _lock.locked()) and (await _lock.owned()):
                     try:
                         await _lock.release()
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in 8fff230bc1: Wrapped the inner `_lock` in `try/finally` so it is always released even if `refresh_tokens()` or `update_creds()` raises.

autogpt_platform/backend/backend/integrations/creds_manager.py
```diff
@@ -107,12 +108,19 @@ class IntegrationCredentialsManager:
...
+        # Recreate the mutex when the event loop has changed (e.g. copilot
+        # executor threads each create their own loop).  The internal
+        # asyncio.Lock inside AsyncRedisKeyedMutex is bound to the loop it was
+        # created on and raises "Future attached to a different loop" otherwise.
+        current_loop = asyncio.get_running_loop()
+        if self._locks and self._locks_loop is current_loop:
             return self._locks
 
         self._locks = AsyncRedisKeyedMutex(await get_redis_async())
+        self._locks_loop = current_loop
```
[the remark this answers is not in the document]
majdyz: 🤖 This is a false positive. The `locks()` method is `async` and can only be called from within an event loop. Each copilot worker thread has exactly one event loop, so within a single thread there is no true concurrency between the `asyncio.get_running_loop()` check and the assignment. The `await get_redis_async()` in between is the only yield point, but even if two coroutines on the same loop interleave, the worst case is creating two mutexes and the second overwriting the first — both are valid for the same loop. Cross-thread access is not an issue because each thread gets its own loop and the `is` check ensures a fresh mutex is created per-loop.

autogpt_platform/backend/backend/integrations/creds_manager.py
majdyz: 🤖 Fixed in 8fff230bc1: Wrapped in try/finally as suggested.

autogpt_platform/backend/backend/integrations/creds_manager.py
majdyz: 🤖 Fixed in 8fff230bc1: Both debug log statements now use `%s` deferred interpolation.

majdyz:
## Review Summary (3 rounds)

### Round 1
- Code changes are clean and well-structured. The `refresh_failed` flag correctly prevents null-cache poisoning on transient OAuth refresh failures.
- Event-loop tracking in `locks()` methods correctly uses `is` identity check.
- `_refresh_locked` / `_refresh_unlocked` split is clean with proper try/finally.
- 4 unresolved threads found (3 self-review + 1 Sentry bot). All had "Fixed" replies referencing commit `8fff230bc1`. Resolved all 4.

### Round 2
- No new issues found. All code paths reviewed.
- The `lock=False` path in copilot integration_creds correctly skips all Redis locking for multi-threaded executor contexts.

### Round 3
- Clean pass. 0 unresolved threads. CI fully green.

### No fixes needed
- All issues were already addressed in the existing commits.

ntindle (dismissed):

ntindle (dismissed):

majdyz:
🤖 🟠 **Should Fix**: `release_all_locks()` (line 300-303 in `creds_manager.py`) now calls `release_all_locks()` twice on the **same** `AsyncRedisKeyedMutex` instance.

Since `self.locks()` now delegates to `self.store.locks()`, both line 302 and 303 resolve to the same `@thread_cached` store locks instance. The second call is redundant (and potentially harmful if `release_all_locks` is not idempotent).

Simplify to:
```python
async def release_all_locks(self):
    """Call this on process termination to ensure all locks are released"""
    await (await self.store.locks()).release_all_locks()
```

autogpt_platform/backend/backend/copilot/integration_creds_test.py
```diff
@@ -176,6 +185,141 @@ async def test_null_cache_has_shorter_ttl_than_token_cache(self):
         assert _NULL_CACHE_TTL < _TOKEN_CACHE_TTL
 
 
+class TestRefreshFailureDoesNotCacheNull:
```
majdyz: 🤖 🟡 **Nice to Have**: `test_transient_refresh_failure_does_not_poison_null_cache` and `test_no_credentials_at_all_does_cache_null` are near-duplicates of `test_oauth2_refresh_failure_returns_none_without_null_cache` and `test_no_credentials_caches_null_entry` above. The only difference is the mock setup style. Consider removing the duplicate class `TestRefreshFailureDoesNotCacheNull` since `TestGetProviderToken` already covers these scenarios with the same assertions.

majdyz:
🤖 Fixed in ea05a0db3: Simplified `release_all_locks()` to a single call to `self.store.locks().release_all_locks()` since `self.locks()` now delegates to `self.store.locks()` and both resolve to the same `@thread_cached` instance.

autogpt_platform/backend/backend/copilot/integration_creds_test.py
majdyz: 🤖 Fixed in ea05a0db3: Removed the duplicate `TestRefreshFailureDoesNotCacheNull` class. The two scenarios are already covered by `test_oauth2_refresh_failure_returns_none_without_null_cache` and `test_no_credentials_caches_null_entry` in `TestGetProviderToken`.

majdyz (commented):
## Test Report (Round 2) -- Local Server Verification

**Environment:** Local Docker backend (:8006), Supabase (:8000), test user

### Code Review

Reviewed all 4 changed files. The fixes correctly address the three interrelated bugs:

1. **Per-thread Redis mutex via `@thread_cached`** -- `IntegrationCredentialsStore.locks()` now uses the `@thread_cached` decorator (backed by `threading.local()`), ensuring each copilot worker thread gets its own `AsyncRedisKeyedMutex` bound to its own event loop. This eliminates the `RuntimeError: Future attached to a different loop`.

2. **`_refresh_locked` / `_refresh_unlocked` split** -- `refresh_if_needed(lock=False)` now truly skips ALL Redis locking (both outer "refresh" scope and inner credential lock). Previously `lock=False` only skipped the inner lock while the outer was always acquired.

3. **Null-cache poisoning fix** -- `_null_cache` is now only populated when the user genuinely has no credentials. Transient refresh failures (e.g. event-loop mismatch) no longer block lookups for 60s.

### Unit Tests: 17/17 PASSED

All existing + new tests pass, including:
- `test_store_locks_returns_per_thread_instance` -- verifies different threads get different mutex instances
- `test_refresh_if_needed_lock_false_skips_redis` -- verifies lock-free path
- `test_oauth2_refresh_failure_returns_none_without_null_cache` -- verifies no null-cache poisoning on refresh failure

### API Tests: 3/3 PASSED

| Endpoint | Status |
|----------|--------|
| `GET /api/integrations/credentials` | 200 -- returns system credentials |
| `GET /api/integrations/openai/credentials` | 200 -- filtered correctly |
| `GET /api/integrations/github/credentials` | 200 -- empty list (no OAuth) |

### Verdict: PASS

PR #12639, opened 2026-04-01
fix(frontend): collapse unconnected block outputs by default

### LINKED ISSUE #11044: Collapse unconnected object sub-outputs on blocks
Any node output that is:

* not connected
* not a top-level output

Should be default hidden. The user should be able to expand an object output to show its sub-outputs.

Hiding unused sub-outputs will save a lot of wasted Builder canvas real-estate and improve usability in such cases.

crawfordxx:
### Changes

Closes #11044

- Default the Output section's `isOutputVisible` state to `false` so that only connected or broken outputs are rendered initially
- Add a "+N more" indicator next to the Output header when unconnected outputs are hidden, so users know there are additional outputs available
- Users can still expand the Output section to see all outputs by clicking the toggle

This reduces canvas clutter by hiding output handles that aren't wired to anything, while keeping connected outputs always visible.

#### For code changes:
- [x] I have made a test plan
  - [x] Verified that unconnected outputs are hidden by default
  - [x] Verified that connected/broken outputs remain visible
  - [x] Verified that clicking the Output toggle reveals all outputs
  - [x] Verified that the "+N more" indicator shows correct count
  - [x] Verified that nested object sub-outputs respect the collapse state


ntindle:
closing for now due to unsigned cla :) feel free to reopen when signed :)

PR #12642, opened 2026-04-01
fix(backend/copilot): detect prompt-too-long in AssistantMessage content and ResultMessage success subtype

majdyz:
## Why

PR #12625 fixed the prompt-too-long retry mechanism for most paths, but two SDK-specific paths were still broken. The dev session `d2f7cba3` kept accumulating synthetic "Prompt is too long" error entries on every turn, growing the transcript from 2.5 MB → 3.2 MB, making recovery impossible.

Root causes identified from production logs (`[T25]`, `[T28]`):

**Path 1 — AssistantMessage content check:**
When the Claude API rejects a prompt, the SDK surfaces it as `AssistantMessage(error="invalid_request", content=[TextBlock("Prompt is too long")])`. Our check only inspected `error_text = str(sdk_error)` which is `"invalid_request"` — not a prompt-too-long pattern. The content was then streamed out as `StreamText`, setting `events_yielded = 1`, which blocked retry even when the ResultMessage fired.

**Path 2 — ResultMessage success subtype:**
After the SDK auto-compacts internally (via `PreCompact` hook) and the compacted transcript is _still_ too long, the SDK returns `ResultMessage(subtype="success", result="Prompt is too long")`. Our check only ran for `subtype="error"`. With `subtype="success"`, the stream "completed normally", appended the synthetic error entry to the transcript via `transcript_builder`, and uploaded it to GCS — causing the transcript to grow on each failed turn.

## What

- **AssistantMessage handler**: when `sdk_error` is set, also check the content text. `sdk_error` being non-`None` confirms this is an API error message (not user-generated content), so content inspection is safe.
- **ResultMessage handler**: check `result` for prompt-too-long patterns regardless of `subtype`, covering the SDK auto-compact path where `subtype="success"` with `result="Prompt is too long"`.

## How

Two targeted one-line condition expansions in `_run_stream_attempt`, plus two new integration tests in `retry_scenarios_test.py` that reproduce each broken path and verify retry fires correctly.

## Changes

- `backend/copilot/sdk/service.py`: fix AssistantMessage content check + ResultMessage subtype-independent check
- `backend/copilot/sdk/retry_scenarios_test.py`: add 2 integration tests for the new scenarios

## Checklist

- [x] Tests added for both new scenarios (45 total, all pass)
- [x] Formatted (`poetry run format`)
- [x] No false-positive risk: AssistantMessage check gated behind `sdk_error is not None`
- [x] Root cause verified from production pod logs

PR #12643, opened 2026-04-01
fix(backend/copilot): close open text block before compaction events to prevent text-end errors

majdyz:
## Why

When the Claude SDK triggers auto-compaction during active text streaming, `emit_end_if_ready` yields compaction events that end with `StreamFinishStep`. The Vercel AI SDK's `finish-step` chunk handler clears `state.activeTextParts = {}`. If a `StreamTextEnd` event arrives after this (e.g. when the next `AssistantMessage` closes the previous text block), the SDK throws:

> "Received text-end for missing text part with ID '...'"

This was seen in session `bc12e570-abc3-4941-b011-40199dd831e8` at 16:59:38 UTC after the SDK auto-compacted while text was streaming.

## What

Before yielding compaction events, call `_end_text_if_open` to flush any in-progress text block. This ensures `StreamTextEnd` always arrives before `StreamFinishStep`, matching the order the Vercel AI SDK expects.

## How

In `_run_stream_attempt`, when `compact_result.events` is non-empty, collect pre-close events via `_end_text_if_open` and yield them first:

```python
if compact_result.events:
    pre_close: list[StreamBaseResponse] = []
    state.adapter._end_text_if_open(pre_close)
    for r in pre_close:
        yield r
for ev in compact_result.events:
    yield ev
```

Three unit tests added to `response_adapter_test.py`:
- `test_end_text_if_open_emits_text_end_before_finish_step` — asserts `StreamTextEnd` index < `StreamFinishStep` index in combined stream
- `test_end_text_if_open_no_op_when_no_text_open` — no events emitted when no text is open
- `test_end_text_if_open_no_op_after_text_already_ended` — idempotent; second call is a no-op

## Checklist

- [x] I've considered the impact this change will have on existing users
- [x] I've tested my changes
- [x] I've written new tests where applicable

### DIFF
--- a/autogpt_platform/backend/backend/copilot/sdk/response_adapter_test.py
+++ b/autogpt_platform/backend/backend/copilot/sdk/response_adapter_test.py
@@ -29,6 +29,7 @@ from backend.copilot.response_model import (
     StreamToolOutputAvailable,
 )
 
+from .compaction import compaction_events
 from .response_adapter import SDKResponseAdapter
 from .tool_adapter import MCP_TOOL_PREFIX
 from .tool_adapter import _pending_tool_outputs as _pto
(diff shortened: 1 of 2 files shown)


majdyz:
Closing — changes merged into #12642 instead.

PR #12649, opened 2026-04-02
ci: add gitleaks secret scanning to pre-commit hooks

majdyz:
### Why / What / How

**Why:** We had no local pre-commit protection against accidentally committing secrets. The existing `detect-secrets` hook only ran on `pre-push`, which is too late — secrets are already in git history by that point. GitHub's push protection only covers known provider patterns and runs server-side.

**What:** Adds a 3-layer defense against secret leaks: local pre-commit hooks (gitleaks + detect-secrets), and a CI workflow as a safety net.

**How:** 
- Moved `detect-secrets` from `pre-push` to `pre-commit` stage
- Added `gitleaks` as a second pre-commit hook (Go binary, faster and more comprehensive rule set)
- Added `.gitleaks.toml` config with allowlists for known false positives (test fixtures, dev docker JWTs, Firebase public keys, lock files, docs examples)
- Added `repo-secret-scan.yml` CI workflow using `gitleaks-action` on PRs/pushes to master/dev

### Changes 🏗️

- `.pre-commit-config.yaml`: Moved `detect-secrets` to pre-commit stage, added baseline arg, added `gitleaks` hook
- `.gitleaks.toml`: New config with tuned allowlists for this repo's false positives
- `.secrets.baseline`: Empty baseline for detect-secrets to track known findings
- `.github/workflows/repo-secret-scan.yml`: New CI workflow running gitleaks on every PR and push

#### For code changes:
- [x] I have made a test plan
  - [x] Ran `gitleaks detect --no-git` against the full repo — only `.env` files (gitignored) remain as findings
  - [x] Verified gitleaks catches a test secret file correctly
  - [x] Pre-commit hooks pass on commit (both detect-secrets and gitleaks passed)


.github/workflows/repo-secret-scan.yml
```diff
@@ -0,0 +1,25 @@
...
+jobs:
+  gitleaks:
+    name: Gitleaks
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - uses: gitleaks/gitleaks-action@v2
```
majdyz: 🤖 🟠 **Should Fix**: `gitleaks/gitleaks-action@v2` is pinned to a mutable tag. For a security-focused PR, this is a supply chain risk — a compromised tag update could bypass all scanning. Pin to the full commit SHA instead:

```yaml
- uses: gitleaks/gitleaks-action@cb7149a9b7f60d4b12cb79288cfbe83365b57042 # v2.3.8
```

You can find the latest SHA at https://github.com/gitleaks/gitleaks-action/releases

.gitleaks.toml
```diff
@@ -0,0 +1,36 @@
...
+    '''\.secrets\.baseline$''',
+    # Build artifacts and caches (should not be committed)
+    '''__pycache__/''',
+    '''classic/frontend/build/''',
+    # Docker dev setup (local dev JWTs/keys only)
+    '''autogpt_platform/db/docker/''',
+    # Load test configs (dev JWTs)
+    '''load-tests/configs/''',
+    # Test files with fake/fixture keys
+    '''_test\.py$''',
```
majdyz: 🤖 🟠 **Should Fix**: This only matches files ending with `_test.py` but misses the `test_*.py` naming convention and `conftest.py`. Should be:

```toml
'''(_test|test_.*|conftest)\.py$''',
```

.secrets.baseline
```diff
@@ -0,0 +1,43 @@
...
+    {"path": "detect_secrets.filters.heuristic.is_lock_file"},
+    {"path": "detect_secrets.filters.heuristic.is_not_alphanumeric_string"},
+    {"path": "detect_secrets.filters.heuristic.is_potential_uuid"},
+    {"path": "detect_secrets.filters.heuristic.is_prefixed_with_dollar_sign"},
+    {"path": "detect_secrets.filters.heuristic.is_sequential_string"},
+    {"path": "detect_secrets.filters.heuristic.is_swagger_file"},
+    {"path": "detect_secrets.filters.heuristic.is_templated_secret"}
+  ],
+  "results": {},
+  "generated_at": "2026-04-02T14:40:00Z"
```
majdyz: 🤖 🟡 **Nice to Have**: This baseline was manually created with an empty `results` dict. It works, but if someone runs `detect-secrets scan` later to update it, the format may differ from this hand-crafted version. Consider adding a comment in the PR description noting this was intentionally left empty as a starting point, or regenerate it with `detect-secrets scan --exclude-files ... > .secrets.baseline`.

.gitleaks.toml
```diff
@@ -0,0 +1,36 @@
...
+    # Documentation (example curl commands, placeholder keys)
+    '''docs/.*\.md$''',
```
majdyz, line 26: 🤖 🟡 **Nice to Have**: `docs/.*\.md$` is quite broad — it allows any secret in any markdown under `docs/`. Consider narrowing to only match known patterns like `YOUR_API_KEY` or `example`, or adding a comment explaining the rationale (docs only contain placeholder keys in curl examples).

.github/workflows/repo-secret-scan.yml
majdyz: 🤖 Fixed in 07b96fae3b: Pinned to SHA `ff98106e4c7b2bc287b24eaf42907196329070c7` (v2.3.9, latest).

.gitleaks.toml
majdyz: 🤖 Fixed in 07b96fae3b: Expanded pattern to `(_test|test_.*|conftest)\.py$`.

.secrets.baseline
majdyz: 🤖 Fixed in 07b96fae3b: Regenerated baseline with a proper `detect-secrets scan` over the active codebase (26 files baselined).

.gitleaks.toml
majdyz: 🤖 Fixed in 07b96fae3b: Added clarifying comment — docs only contain placeholder keys in curl/API examples.

.gitleaks.toml
```diff
@@ -0,0 +1,36 @@
...
+    # Firebase config (public API keys by design)
+    '''google-services\.json$''',
+    '''classic/frontend/(lib|web)/''',
+]
+# CI test-only encryption key (marked DO NOT USE IN PRODUCTION)
+regexes = [
+    '''dvziYgz0KSK8FENhju0ZYi8''',
+    # LLM model name enum values falsely flagged as API keys
+    '''Llama-\d.*Instruct''',
```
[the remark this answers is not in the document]
majdyz: 🤖 Verified — only 3 model names in `llm.py` were actually triggering gitleaks false positives (the `Llama-X-...-Instruct-FP8` patterns with high entropy). The lowercase variants like `llama-3.3-70b-versatile` and mistral models do not trigger because they don't match gitleaks' `generic-api-key` entropy/pattern thresholds. Broadening the regex preemptively would weaken detection for no practical benefit. If new model names trigger false positives in the future, we can expand the pattern then.

.github/workflows/repo-secret-scan.yml
```diff
@@ -0,0 +1,35 @@
...
+          curl -sSfL "https://github.com/gitleaks/gitleaks/releases/download/v${GITLEAKS_VERSION}/gitleaks_${GITLEAKS_VERSION}_linux_x64.tar.gz" \
+            | tar -xz -C /usr/local/bin gitleaks
+
+      - name: Run gitleaks
+        run: |
+          gitleaks detect \
+            --source . \
+            --config .gitleaks.toml \
+            --log-opts "${{ github.event_name == 'pull_request' && format('{0}...{1}', github.event.pull_request.base.sha, github.event.pull_request.head.sha) || format('{0}^...{1}', github.event.before, github.sha) }}" \
+            --verbose
```
majdyz: 🤖 🔴 **Blocker**: The `--log-opts` expression has two edge-case failures:

1. **First push to a new branch**: `github.event.before` is `0000000000000000000000000000000000000000`, so `0000...^...HEAD` will fail
2. **`merge_group` event**: Neither `pull_request` nor `push` event properties exist, so both branches of the ternary produce invalid refs

Safer approach — scan the full source (no git history) since the allowlists already handle false positives:

```yaml
- name: Run gitleaks
  run: |
    gitleaks detect \
      --source . \
      --no-git \
      --config .gitleaks.toml \
      --verbose
```

Or if you want diff-only scanning for PRs, add explicit conditionals per event type.

majdyz: 🤖 Fixed in 455135a5b0: Switched to `--no-git` scan which avoids all the ref edge cases. The allowlist config handles false positives.

.github/workflows/repo-secret-scan.yml
```diff
@@ -0,0 +1,35 @@
...
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Install gitleaks
+        run: |
+          GITLEAKS_VERSION="8.24.3"
+          curl -sSfL "https://github.com/gitleaks/gitleaks/releases/download/v${GITLEAKS_VERSION}/gitleaks_${GITLEAKS_VERSION}_linux_x64.tar.gz" \
+            | tar -xz -C /usr/local/bin gitleaks
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in 4278fec253: Added checksum verification before extraction.

.github/workflows/repo-secret-scan.yml
```diff
@@ -0,0 +1,35 @@
...
+
+      - name: Run gitleaks
+        run: |
+          gitleaks detect \
+            --source . \
+            --config .gitleaks.toml \
+            --log-opts "${{ github.event_name == 'pull_request' && format('{0}...{1}', github.event.pull_request.base.sha, github.event.pull_request.head.sha) || format('{0}^...{1}', github.event.before, github.sha) }}" \
```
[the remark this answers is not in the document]
majdyz: 🤖 Already resolved — switched to `--no-git` mode in 455135a5b0, so event type / ref handling is no longer relevant.

.github/workflows/repo-secret-scan.yml
```diff
@@ -0,0 +1,38 @@
...
+  cancel-in-progress: ${{ startsWith(github.event_name, 'pull_request') }}
+
+jobs:
+  gitleaks:
+    name: Gitleaks
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
```
majdyz: 🤖 🔵 **Nit**: `fetch-depth: 0` clones full git history but we're using `--no-git` mode which doesn't need it. Can use the default shallow clone (`fetch-depth: 1`) to speed up checkout.

majdyz: 🤖 Fixed in 06033bb1ea: Removed `fetch-depth: 0`, using default shallow clone.

.pre-commit-config.yaml
```diff
@@ -23,9 +23,14 @@ repos:
       - id: detect-secrets
         name: Detect secrets
         description: Detects high entropy strings that are likely to be passwords.
-        files: ^autogpt_platform/
-        exclude: pnpm-lock\.yaml$
-        stages: [pre-push]
+        args: ["--baseline", ".secrets.baseline"]
+        exclude: (pnpm-lock\.yaml|\.env\.(default|example|template))$
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in 8947f28c1c: Added `files: ^autogpt_platform/` back to scope detect-secrets to match its baseline coverage. Gitleaks (with its own allowlist config) handles the full repo.

.github/workflows/repo-secret-scan.yml
```diff
@@ -0,0 +1,36 @@
...
+          grep " ${ARCHIVE}$" "gitleaks_${GITLEAKS_VERSION}_checksums.txt" | sha256sum -c -
+          tar -xzf "${ARCHIVE}" -C /usr/local/bin gitleaks
+
+      - name: Run gitleaks
+        run: |
+          gitleaks detect \
+            --source . \
+            --no-git \
+            --config .gitleaks.toml \
+            --verbose
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in 4f31ed1024: Added `--redact` flag to mask secrets in CI logs.

.github/workflows/repo-secret-scan.yml
```diff
@@ -0,0 +1,37 @@
+name: Secret Scan
```
ntindle: github secret scanning does this already

ntindle (changes requested):

.github/workflows/repo-secret-scan.yml
majdyz: Good point — GitHub's built-in secret scanning catches secrets post-push. The value of gitleaks in pre-commit is catching them **before** they hit the remote (preventing the need for secret rotation). But if the team prefers relying on GitHub's scanning + push protection, this PR can be closed. @majdyz what do you think?

majdyz: No, me want this on pre-commit so u (the AI) not fucking up on committing api key again.
the one for ci I'm not sure is this executed on remote? then not needed?

.gitleaks.toml
```diff
@@ -0,0 +1,36 @@
...
+    '''autogpt_platform/db/docker/''',
+    # Load test configs (dev JWTs)
+    '''load-tests/configs/''',
+    # Test files with fake/fixture keys (_test.py, test_*.py, conftest.py)
+    '''(_test|test_.*|conftest)\.py$''',
+    # Documentation (only contains placeholder keys in curl/API examples)
+    '''docs/.*\.md$''',
+    # Firebase config (public API keys by design)
+    '''google-services\.json$''',
+    '''classic/frontend/(lib|web)/''',
```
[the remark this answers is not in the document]
majdyz: The pattern is intentionally unanchored — gitleaks path allowlist uses substring matching, so `classic/frontend/(lib|web)/` matches any file under those directories (e.g., `classic/frontend/lib/utils/config.js`). Adding `^.*` prefix would be redundant. The directories contain auto-generated vendor code and Firebase public configs, not user code.

.github/workflows/repo-secret-scan.yml
majdyz: Yes, the CI gitleaks job (`.github/workflows/repo-secret-scan.yml`) runs on the GitHub runner **after push** — same timing as GitHub's built-in secret scanning. So it's redundant with GitHub's native feature.

**Recommendation:** Keep the pre-commit hook (catches secrets before push — exactly what you want for AI agents). Remove the CI workflow since GitHub secret scanning + push protection already covers post-push detection.

Want me to remove the CI workflow and keep only the pre-commit hook?

majdyz: Done — removed `.github/workflows/repo-secret-scan.yml` in 9a01329. Keeping only the pre-commit hook for catching secrets before push.

.gitleaks.toml
```diff
@@ -0,0 +1,36 @@
...
+]
+# CI test-only encryption key (marked DO NOT USE IN PRODUCTION)
+regexes = [
+    '''dvziYgz0KSK8FENhju0ZYi8''',
+    # LLM model name enum values falsely flagged as API keys
+    '''Llama-\d.*Instruct''',
```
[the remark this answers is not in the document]
majdyz: False positive — the `Llama-\d.*Instruct` pattern is in the `regexes` allowlist which matches against the **secret value itself**, not file paths. It exists because gitleaks falsely flags LLM model name strings (e.g., `Llama-3.1-70B-Instruct`) in the `LlmModel` enum as API keys. A real API key would not contain `Llama-` followed by digits and `Instruct` — the false positive risk is negligible.

ntindle (approved):

PR #12656, opened 2026-04-03
fix(backend): use OpenRouter client for simulator to support non-OpenAI models

majdyz:
## Why

Dry-run block simulation is failing in production with `404 - model gemini-2.5-flash does not exist`. The simulator's default model (`google/gemini-2.5-flash`) is a non-OpenAI model that requires OpenRouter routing, but the shared `get_openai_client()` prefers the direct OpenAI key, creating a client that can't handle non-OpenAI models. The old code also stripped the provider prefix, sending `gemini-2.5-flash` to OpenAI's API.

## What

- Added `prefer_openrouter` keyword parameter to `get_openai_client()` — when True, prefers the OpenRouter key (returns None if unavailable, rather than falling back to an incompatible direct OpenAI client)
- Simulator now calls `get_openai_client(prefer_openrouter=True)` so `google/gemini-2.5-flash` routes correctly through OpenRouter
- Removed the redundant `SIMULATION_MODEL` env var override and the now-unnecessary provider prefix stripping from `_simulator_model()`

## How

`get_openai_client()` is decorated with `@cached(ttl_seconds=3600)` which keys by args, so `get_openai_client()` and `get_openai_client(prefer_openrouter=True)` are cached independently. When `prefer_openrouter=True` and no OpenRouter key exists, returns `None` instead of falling back — the simulator already handles `None` with a clear error message.

### Checklist
- [x] All 24 dry-run tests pass
- [x] Test asserts `get_openai_client` is called with `prefer_openrouter=True`
- [x] Format, lint, and pyright pass
- [x] No changes to user-facing APIs
- [ ] Deploy to staging and verify simulation works


autogpt_platform/backend/backend/executor/simulator.py
```diff
@@ -53,30 +53,15 @@ def _simulator_model() -> str:
...
-        model = env_model
-    else:
-        # 2. ChatConfig.simulation_model (falls back to default).
-        try:
-            from backend.copilot.config import ChatConfig  # noqa: PLC0415
-
-            model = ChatConfig().simulation_model or _DEFAULT_SIMULATOR_MODEL
-        except Exception:
-            model = _DEFAULT_SIMULATOR_MODEL
+        return env_model
```
majdyz: 🤖 🟠 **Should Fix**: The `SIMULATION_MODEL` env var path returns early without going through `get_openai_client(prefer_openrouter=True)`. If someone sets `SIMULATION_MODEL=gpt-4o-mini` (no provider prefix), it'll be sent to OpenRouter which expects `openai/gpt-4o-mini`. Consider documenting that the env var should use OpenRouter-prefixed names, or add a note in the comment.

autogpt_platform/backend/backend/executor/simulator.py
```diff
@@ -53,30 +53,15 @@ def _simulator_model() -> str:
     # 1. Environment variable override (highest priority).
     env_model = os.environ.get("SIMULATION_MODEL")
     if env_model:
```
majdyz: 🤖 🟡 **Nice to Have**: The PR description uses a 'Changes' format but the project template expects Why/What/How structure. Consider restructuring to match the PR template in `.github/PULL_REQUEST_TEMPLATE.md`.

autogpt_platform/backend/backend/executor/simulator.py
```diff
@@ -53,30 +53,15 @@ def _simulator_model() -> str:
...
 
-        secrets = Settings().secrets
-        if secrets.openai_internal_api_key and "/" in model:
-            model = model.split("/", 1)[1]
+        return ChatConfig().simulation_model or _DEFAULT_SIMULATOR_MODEL
     except Exception:
-        pass
-
-    return model
+        return _DEFAULT_SIMULATOR_MODEL
```
majdyz, line 57: 🤖 🟠 **Should Fix**: The test file patches `backend.executor.simulator.get_openai_client` — while this works (the symbol IS imported at module level), consider asserting the mock was called with `prefer_openrouter=True` in at least one test to lock in the contract. This prevents regressions if someone changes the call back to the default.

autogpt_platform/backend/backend/executor/simulator.py
majdyz: 🤖 Fixed in ccbda24209: Removed the SIMULATION_MODEL env var entirely — ChatConfig.simulation_model (configurable via CHAT_SIMULATION_MODEL) is now the single source of truth.

autogpt_platform/backend/backend/executor/simulator.py
majdyz: 🤖 Fixed in ccbda24209: Added assertion that get_openai_client is called with prefer_openrouter=True in test_simulate_block_basic.

autogpt_platform/backend/backend/util/clients.py
```diff
@@ -163,23 +163,34 @@ async def get_async_supabase() -> "AClient":
...
-            base_url=OPENROUTER_BASE_URL,
-        )
+    openai_key = settings.secrets.openai_internal_api_key
+    openrouter_key = settings.secrets.open_router_api_key
+
+    if prefer_openrouter:
+        if openrouter_key:
+            return AsyncOpenAI(api_key=openrouter_key, base_url=OPENROUTER_BASE_URL)
+        if openai_key:
+            return AsyncOpenAI(api_key=openai_key)
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in ccbda24209: When prefer_openrouter=True and no OpenRouter key is configured, now returns None instead of falling back to the direct OpenAI client. The simulator already handles None with a clear error message.

autogpt_platform/backend/backend/executor/simulator.py
majdyz: 🤖 Acknowledged — updating PR description to Why/What/How format.

ntindle (dismissed):

majdyz:
@ntindle CI is all green now — pushed a few fixes since your approval (removed redundant env var, fixed unsafe OpenAI fallback, added test coverage for codecov). Could you re-approve when you get a chance? Thanks!

ntindle (approved):

PR #12660, opened 2026-04-03
fix(copilot): prevent duplicate side effects from double-submit and stale-cache race

majdyz:
## Why

#12604 (intermediate persistence) introduced two bugs on dev:

1. **Duplicate user messages** — `set_turn_duration` calls `invalidate_session_cache()` which deletes the Redis key. Concurrent `get_chat_session()` calls re-populate it from DB with stale data. The executor loads this stale cache, misses the user message, and re-appends it.

2. **Tool outputs lost on hydration** — Intermediate flushes save assistant messages to DB before `StreamToolInputAvailable` sets `tool_calls` on them. Since `_save_session_to_db` is append-only (uses `start_sequence`), the `tool_calls` update is lost — subsequent flushes start past that index. On page refresh / SSE reconnect, tool UIs (SetupRequirementsCard, run_block output, etc.) are invisible.

3. **Sessions stuck running** — If a tool call hangs (e.g. WebSearch provider not responding), the stream never completes, `mark_session_completed` never runs, and the `active_stream` flag stays stale in Redis.

## What

- **In-place cache update** in `set_turn_duration` — replaces `invalidate_session_cache()` with a read-modify-write that patches the duration on the cached session, eliminating the stale-cache repopulation window
- **tool_calls backfill** — tracks the flush watermark and assistant message index; when `StreamToolInputAvailable` sets `tool_calls` on an already-flushed assistant, updates the DB record directly via `update_message_tool_calls()`
- **Improved message dedup** — `is_message_duplicate()` / `maybe_append_user_message()` scans trailing same-role messages (current turn) instead of only checking `messages[-1]`
- **Idle timeout** — aborts the stream with a retryable error if no meaningful SDK message arrives for 10 minutes, preventing hung tool calls from leaving sessions stuck

## Changes

- `copilot/db.py` — `update_message_tool_calls()`, in-place cache update in `set_turn_duration`
- `copilot/model.py` — `is_message_duplicate()`, `maybe_append_user_message()`
- `copilot/sdk/service.py` — flush watermark tracking, tool_calls backfill, idle timeout
- `copilot/baseline/service.py` — use `maybe_append_user_message()`
- `copilot/model_test.py` — unit tests for dedup
- `copilot/db_test.py` — unit tests for set_turn_duration cache update

## Checklist

- [x] My PR title follows [conventional commit](https://www.conventionalcommits.org/) format
- [x] Out-of-scope changes are less than 20% of the PR
- [x] Changes to `data/*.py` validated for user ID checks (N/A)
- [x] Protected routes updated in middleware (N/A)


autogpt_platform/frontend/src/app/(platform)/copilot/useCopilotPage.ts
```diff
@@ -192,6 +203,12 @@ export function useCopilotPage() {
     const trimmed = message.trim();
     if (!trimmed && (!files || files.length === 0)) return;
 
+    // Prevent double-submit: block re-entry until the stream finishes.
+    // This closes the race window between sendMessage() and the AI SDK
+    // status flipping to "submitted" (which gates the UI input).
+    if (isSendingRef.current) return;
+    isSendingRef.current = true;
```
majdyz: 🤖 🟠 **Should Fix**: Two more exit paths still leave isSendingRef stuck at true: (1) when all file uploads fail at line 248, the throw exits the try block but the finally only resets setIsUploadingFiles, not isSendingRef; (2) if createSession() at line 268 throws. Suggest wrapping the entire function body (after the guard) in try/finally to reset isSendingRef on any non-sendMessage exit.

autogpt_platform/frontend/src/app/(platform)/copilot/useCopilotPage.ts
```diff
@@ -191,6 +202,12 @@ export function useCopilotPage() {
...
```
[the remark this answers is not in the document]
majdyz: Fixed in 9e60fd7 and 6d16b848: Added isSendingRef.current = false before each file-validation early return, and wrapped the async body in try/catch to catch upload-failure and createSession throws.

autogpt_platform/backend/backend/copilot/sdk/service.py
```diff
@@ -1652,15 +1652,17 @@ async def stream_chat_completion_sdk(
...
-        len(session.messages) == 0
-        or not (
-            session.messages[-1].role == new_message_role
-            and session.messages[-1].content == message
-        )
-    ):
+    already_exists = message and any(
+        m.role == new_message_role and m.content == message
+        for m in session.messages[-5:]
+    )
```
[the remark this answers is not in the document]
majdyz: Fixed in 9e60fd75: Replaced the last-5-messages scan with a turn-scoped check that only examines trailing messages of the same role. This way cross-turn repetitions like "yes" or "try again" are preserved, while same-turn duplicates from stale cache are still caught.

autogpt_platform/frontend/src/app/(platform)/copilot/useCopilotPage.ts
majdyz: Fixed in 6d16b848: Wrapped the async body of onSend (after the guard and file-validation checks) in a try/catch that resets isSendingRef.current = false on any thrown exception. This covers both the upload-failure throw and createSession failure paths.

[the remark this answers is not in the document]
majdyz: Confirmed, already fixed in 6d16b848: the try/catch around the async body catches createSession() failures and resets isSendingRef.current = false.

autogpt_platform/backend/backend/copilot/sdk/service.py
```diff
@@ -1669,15 +1669,22 @@ async def stream_chat_completion_sdk(
...
-            and session.messages[-1].content == message
-        )
-    ):
+    tail: list[ChatMessage] = []
+    for m in reversed(session.messages):
+        if m.role == new_message_role:
+            tail.append(m)
+        else:
+            break
+    already_exists = message and any(m.content == message for m in tail)
```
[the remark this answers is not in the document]
majdyz: Confirmed, fixed in 67e6d7a: extracted is_message_duplicate() helper that only checks trailing same-role messages (current turn scope), so cross-turn repetitions are preserved.

autogpt_platform/frontend/src/app/(platform)/copilot/useCopilotPage.ts
```diff
@@ -213,40 +231,45 @@ export function useCopilotPage() {
...
-            // All uploads failed — abort send so chips revert to editable
-            throw new Error("All file uploads failed");
+    try {
+      if (sessionId) {
+        if (files && files.length > 0) {
+          setIsUploadingFiles(true);
+          try {
+            const uploaded = await uploadFiles(files, sessionId);
+            if (uploaded.length === 0) {
+              // All uploads failed — abort send so chips revert to editable
```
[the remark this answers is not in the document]
majdyz: Confirmed, fixed in 6d16b848: the outer try/catch in onSend resets isSendingRef.current = false when uploadFiles or createSession throws.

majdyz:
@coderabbitai Good catch on the pendingMessage upload failure path. Fixed in 0c0adae: added isSendingRef.current = false in the empty-upload-result branch and a .catch() handler for thrown errors. The onSend upload failure was already addressed in 6d16b848 via the outer try/catch.

majdyz:
Fixed queued draft cleanup on `createSession()` failure (e5f9e97): The catch block in `onSend()` now clears `pendingMessage`, `pendingFilesRef`, and `pendingFilePartsRef` when `createSession()` throws, preventing the pending-message effect from retrying forever.

autogpt_platform/frontend/src/app/(platform)/copilot/useCopilotPage.ts
```diff
@@ -133,6 +140,9 @@ export function useCopilotPage() {
             files: fileParts.length > 0 ? fileParts : undefined,
           });
         })
```
majdyz, line 135: 🤖 🟡 **Nice to Have**: The `.catch` here correctly resets `isSendingRef` when the upload promise rejects. However, the two sibling `sendMessage(...)` calls at the top and bottom of this `useEffect` (the `prebuiltParts` path and the text-only path) are not similarly protected. If `sendMessage` throws synchronously from an unexpected AI SDK state, `isSendingRef` stays locked.

In practice the AI SDK `sendMessage` does not throw synchronously, so this is defensive — but for consistency with the try/catch pattern in `onSend`, consider wrapping the full branch in a try/catch.

autogpt_platform/backend/backend/copilot/model_test.py
```diff
@@ -424,3 +425,64 @@ async def callback_save():
...
+
+def test_duplicate_check_for_assistant_role():
+    """Works correctly when checking assistant role too."""
+    msgs = [
+        ChatMessage(role="user", content="hi"),
+        ChatMessage(role="assistant", content="hello"),
+        ChatMessage(role="assistant", content="how can I help?"),
+    ]
+    assert is_message_duplicate(msgs, "assistant", "hello") is True
+    assert is_message_duplicate(msgs, "assistant", "new response") is False
```
majdyz, line 489: 🤖 🔵 **Nit**: `ChatMessage.content` is typed `str | None`, so a message with `content=None` (e.g. a tool-call-only assistant message) can appear in the trailing same-role block. The comparison `m.content == content` still works correctly (None != "hello"), but adding a test case would document this edge case explicitly:

```python
def test_no_false_positive_when_content_is_none():
    """Messages with content=None in the trailing block do not match."""
    msgs = [
        ChatMessage(role="user", content=None),
        ChatMessage(role="user", content="hello"),
    ]
    assert is_message_duplicate(msgs, "user", "hello") is True
    # None-content message should not match any string
    msgs2 = [
        ChatMessage(role="user", content=None),
    ]
    assert is_message_duplicate(msgs2, "user", "hello") is False
```

autogpt_platform/frontend/src/types/twemoji.d.ts
```diff
@@ -0,0 +1,11 @@
...
+  interface TwemojiOptions {
+    folder?: string;
+    ext?: string;
+    attributes?: () => Record<string, string>;
+  }
+  const twemoji: {
+    parse(text: string, options?: TwemojiOptions): string;
+  };
+  export default twemoji;
+}
```
majdyz: 🤖 🔵 **Nit**: This `twemoji.d.ts` type declaration is unrelated to the double-submit / dedup fix. Consider splitting it into a separate PR to keep this one focused on the bug fix, per the contribution guide's "keep out-of-scope changes under 20% of the PR" guideline. Not a blocker since it's a small addition.

autogpt_platform/backend/backend/copilot/baseline/service.py
```diff
@@ -397,15 +398,18 @@ async def stream_chat_completion_baseline(
...
-    if message and (
-        len(session.messages) == 0
-        or not (
-            session.messages[-1].role == new_role
-            and session.messages[-1].content == message
-        )
-    ):
+    already_exists = message and is_message_duplicate(
+        session.messages, new_role, message
+    )
```
majdyz: 🤖 🟡 **Nice to Have**: The expression `already_exists = message and is_message_duplicate(...)` relies on Python's short-circuit semantics: when `message` is `None` or `""`, `already_exists` evaluates to `None` or `""` (falsy), not `False`. Then `if message and not already_exists` still works because both branches are falsy.

However, this means `already_exists` is not actually a `bool` — it can be `None`, `""`, or `True`. This is idiomatic Python but could confuse future readers. A minor clarity improvement:

```python
already_exists = bool(message) and is_message_duplicate(
    session.messages, new_role, message
)
```

Or split it into a guard:
```python
if not message:
    pass  # nothing to append
elif is_message_duplicate(session.messages, new_role, message):
    pass  # already persisted by route handler
else:
    session.messages.append(...)
```

Not a correctness issue — just readability.

autogpt_platform/backend/backend/copilot/model.py
```diff
@@ -81,6 +81,27 @@ def from_db(prisma_message: PrismaChatMessage) -> "ChatMessage":
...
+    Only inspects trailing messages that share the given *role* (i.e. the
+    current turn). This ensures legitimately repeated messages across different
+    turns are not suppressed, while same-turn duplicates from stale cache are
+    still caught.
+    """
+    for m in reversed(messages):
+        if m.role == role:
+            if m.content == content:
+                return True
+        else:
```
majdyz, line 100: 🤖 🟡 **Nice to Have**: Good test coverage for `is_message_duplicate`. One scenario worth adding: a session where ALL messages are the same role (no role boundary). For example, if system messages or multiple user messages exist without any assistant response:

```python
def test_all_same_role_messages():
    """When all messages share the same role, the entire list is scanned."""
    msgs = [
        ChatMessage(role="user", content="first"),
        ChatMessage(role="user", content="second"),
        ChatMessage(role="user", content="third"),
    ]
    assert is_message_duplicate(msgs, "user", "first") is True
    assert is_message_duplicate(msgs, "user", "new") is False
```

This verifies the function scans the entire list when there is no role boundary `break`, which is the correct behavior for a brand-new session where the user sends a message before any assistant response.

autogpt_platform/frontend/src/app/(platform)/copilot/useCopilotPage.ts
majdyz: 🤖 Fixed in a83d368: Wrapped the full sendMessage branch in the pendingMessage useEffect with try/catch that resets isSendingRef.current = false on any throw, consistent with the onSend pattern.

autogpt_platform/backend/backend/copilot/model_test.py
majdyz: 🤖 Fixed in a83d368: Added test_no_false_positive_when_content_is_none test case that verifies None-content messages in the trailing block do not match any string.

autogpt_platform/frontend/src/types/twemoji.d.ts
majdyz: 🤖 Acknowledged — the twemoji.d.ts type declaration is actually imported by src/components/atoms/Emoji/Emoji.tsx, so removing it breaks the frontend typecheck. Keeping it in this PR since it was added to support existing code.

autogpt_platform/backend/backend/copilot/baseline/service.py
majdyz: 🤖 Fixed in a83d368: Changed to `bool(message) and is_message_duplicate(...)` in both baseline/service.py and sdk/service.py so already_exists is always a proper bool.

autogpt_platform/backend/backend/copilot/model.py
majdyz: 🤖 Fixed in a83d368: Added test_all_same_role_messages test case that verifies the entire list is scanned when all messages share the same role (no role boundary break).

autogpt_platform/frontend/src/app/(platform)/copilot/useCopilotPage.ts
```diff
@@ -179,74 +195,78 @@ export function useCopilotPage() {
...
+        if (files && files.length > 0) {
+          setIsUploadingFiles(true);
+          try {
+            const uploaded = await uploadFiles(files, sessionId);
+            if (uploaded.length === 0) {
+              // All uploads failed — abort send so chips revert to editable
+              throw new Error("All file uploads failed");
+            }
+            const fileParts = buildFileParts(uploaded);
+            sendMessage({
```
majdyz: 🤖 🟠 **Should Fix**: The `sendTimeRef` auto-expiry (10s) can race with file uploads. `uploadFiles()` is `await`ed here — if it takes >10s, the guard expires and a concurrent `onSend()` call passes the check. The guard should be refreshed after the upload completes, or the timeout should account for upload time.

Simplest fix: refresh `sendTimeRef.current = Date.now()` right after the upload succeeds (before `sendMessage`), so the guard stays active during the entire upload+send window.

majdyz: 🤖 Fixed in 054c7acb43: refreshed `sendTimeRef.current = Date.now()` after upload succeeds in both the `onSend` path and the pending-message `useEffect` path. The guard now stays active during the entire upload+send window.

autogpt_platform/frontend/src/app/(platform)/copilot/helpers.ts
```diff
@@ -104,3 +104,74 @@ export function deduplicateMessages(messages: UIMessage[]): UIMessage[] {
...
+} {
+  if (error.kind === "too_many") {
+    return {
+      title: "Too many files",
+      description: `You can attach up to ${error.limit} files at once.`,
+    };
+  }
+  return {
+    title: "File too large",
+    description: `${error.fileName} exceeds the 100 MB limit.`,
```
majdyz: 🤖 🟡 **Nice to Have**: `formatValidationToast` hardcodes `"100 MB limit"` in the description string instead of deriving it from `error.limit`. If the limit constant changes in the future, the toast message would become stale. Consider:
```ts
description: `${error.fileName} exceeds the ${error.limit / (1024 * 1024)} MB limit.`,
```

autogpt_platform/frontend/src/app/(platform)/copilot/helpers.test.ts
```diff
@@ -1,76 +1,438 @@
...
+      },
+    ]);
+  });
+
+  it("returns an empty array for empty input", () => {
+    expect(buildFileParts([])).toEqual([]);
+  });
+});
+
+// --- Tests for previously-uncovered helpers ---
```
majdyz: 🤖 🔵 **Nit**: This section comment ("Tests for previously-uncovered helpers") reads like a development note. Since all tests in this file are equally valid regardless of when they were written, consider removing the comment to keep the file clean.

autogpt_platform/frontend/src/app/(platform)/copilot/useCopilotPage.ts
```diff
@@ -19,18 +26,16 @@ import { useWorkflowImportAutoSubmit } from "./useWorkflowImportAutoSubmit";
...
 const TITLE_POLL_INTERVAL_MS = 2_000;
 const TITLE_POLL_MAX_ATTEMPTS = 5;
-
-interface UploadedFile {
-  file_id: string;
-  name: string;
-  mime_type: string;
-}
-
 export function useCopilotPage() {
```
majdyz, line 29: 🤖 🔵 **Nit**: Missing blank line between the `TITLE_POLL_MAX_ATTEMPTS` constant and the `export function useCopilotPage()` declaration. The rest of the file uses blank lines to separate logical sections.

autogpt_platform/frontend/src/app/(platform)/copilot/helpers.ts
majdyz: 🤖 Fixed in 706a96ea50: Now derives the limit dynamically from `error.limit / (1024 * 1024)` instead of hardcoding "100 MB".

autogpt_platform/frontend/src/app/(platform)/copilot/helpers.test.ts
majdyz: 🤖 Fixed in 706a96ea50: Removed the development note comment.

autogpt_platform/frontend/src/app/(platform)/copilot/useCopilotPage.ts
majdyz: 🤖 Fixed in 706a96ea50: Added blank line between the constant and the function declaration.

autogpt_platform/frontend/src/app/(platform)/copilot/helpers/convertChatSessionToUiMessages.ts
```diff
@@ -101,6 +101,37 @@ function toToolInput(rawArguments: unknown): unknown {
...
+        case "review_required":
+          return "run_block";
+        case "block_list":
+          return "find_block";
+        case "agent_list":
+        case "agent_details":
+          return "find_library_agent";
+        case "workspace_file_metadata":
+        case "workspace_file_written":
+          return "read_workspace_file";
```
majdyz: 🤖 🟠 **Should Fix**: `inferToolName` has two mapping errors:
1. `workspace_file_written` maps to `"read_workspace_file"` but should map to `"write_workspace_file"` (it is produced by `WriteWorkspaceFileTool`)
2. `workspace_file_content` is missing entirely (produced by `ReadWorkspaceFileTool` for legacy small text files)

The frontend's `GenericTool/helpers.ts` renders these tool types differently (`read_workspace_file` shows file content, `write_workspace_file` shows write confirmation), so wrong mapping causes incorrect UI.

majdyz: 🤖 Fixed in 78c4d22db5: Corrected `workspace_file_written` to map to `"write_workspace_file"` and added `workspace_file_content` mapping to `"read_workspace_file"`.

autogpt_platform/frontend/src/app/(platform)/copilot/helpers/convertChatSessionToUiMessages.ts
```diff
@@ -101,6 +101,37 @@ function toToolInput(rawArguments: unknown): unknown {
...
+        case "bash_exec":
+          return "bash_exec";
+      }
+    }
+  }
+  return "unknown_tool";
+}
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in 78c4d22db5: Good catch. Corrected `workspace_file_written` to map to `"write_workspace_file"` and added the missing `workspace_file_content` case mapping to `"read_workspace_file"`.

autogpt_platform/frontend/src/app/(platform)/copilot/useCopilotPage.ts
```diff
@@ -112,30 +125,39 @@ export function useCopilotPage() {
...
-        .then((uploaded) => {
-          if (uploaded.length === 0) {
-            toast({
-              title: "File upload failed",
-              description: "Could not upload any files. Please try again.",
-              variant: "destructive",
+    try {
+      if (prebuiltParts.length > 0) {
+        // File already uploaded (e.g. workflow import) — send directly
+        sendMessage({ text: msg, files: prebuiltParts });
```
majdyz: 🤖 🟠 **Should Fix**: The `sendTimeRef` guard is not refreshed before calling `sendMessage()` in the prebuilt-parts path (line 131) and the plain-text path (line 157). If session creation takes >10s, the guard set in `onSend()` expires and a concurrent call could pass through. The file-upload path correctly refreshes at line 146. For consistency, add `sendTimeRef.current = Date.now()` before both `sendMessage()` calls.

autogpt_platform/frontend/src/app/(platform)/copilot/useCopilotPage.ts
```diff
@@ -112,30 +125,39 @@ export function useCopilotPage() {
...
-        })
-        .finally(() => setIsUploadingFiles(false));
-    } else {
-      sendMessage({ text: msg });
+          })
+          .catch(() => {
+            sendTimeRef.current = 0;
+          })
+          .finally(() => setIsUploadingFiles(false));
+      } else {
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in d321df63c2: Added `sendTimeRef.current = Date.now()` before `sendMessage()` in the prebuilt-parts path and the plain-text path of the pending message useEffect, matching the file-upload path. All three branches now consistently refresh the guard before sending.

autogpt_platform/frontend/src/app/(platform)/copilot/useCopilotPage.ts
majdyz: 🤖 Fixed in d321df63c2: Added `sendTimeRef.current = Date.now()` before `sendMessage()` in both the prebuilt-parts path (line 131) and the plain-text path (line 159) of the pending message useEffect.

autogpt_platform/frontend/src/app/(platform)/copilot/helpers/convertChatSessionToUiMessages.ts
```diff
@@ -101,6 +101,39 @@ function toToolInput(rawArguments: unknown): unknown {
   return {};
 }
 
+/**
+ * Infer the tool name from a raw tool output object.
+ * The backend tool outputs include a `type` field (e.g. "setup_requirements",
+ * "block_output") that maps to the tool that produced them.
+ */
+function inferToolName(output: unknown): string {
```
majdyz: 🤖 🟠 **Should Fix**: The `convertChatSessionToUiMessages.test.ts` file exists locally (1085 lines of tests for orphan tool recovery, inferToolName, etc.) but was never committed to the branch. This test file should be included in the PR for coverage of the new logic.

majdyz: 🤖 Fixed in 7c86b92e6c: The test file was included in the branch squash and is now part of the PR diff (1085 lines covering orphan tool recovery, inferToolName, and the full conversion pipeline).

autogpt_platform/frontend/src/app/(platform)/copilot/useCopilotPage.ts
```diff
@@ -192,6 +199,15 @@ export function useCopilotPage() {
...
     if (!trimmed && (!files || files.length === 0)) return;
 
+    const SEND_GUARD_TIMEOUT_MS = 10_000;
+    const now = Date.now();
+    if (
+      sendTimeRef.current > 0 &&
+      now - sendTimeRef.current < SEND_GUARD_TIMEOUT_MS
+    )
+      return;
+    sendTimeRef.current = now;
```
majdyz: 🤖 🟠 **Should Fix**: The send guard is set at line 209 before file validation (lines 216-233). If validation fails (too many files / file too large), the guard stays active and the user is locked out for 10s even though nothing was sent. Move `sendTimeRef.current = now` to after the validation block (after line 234), or clear it in the validation-failure return paths.

autogpt_platform/backend/backend/copilot/model.py
```diff
@@ -81,6 +81,27 @@ def from_db(prisma_message: PrismaChatMessage) -> "ChatMessage":
         )
 
 
+def is_message_duplicate(
```
majdyz, line 84: 🤖 🟠 **Should Fix**: PR description is stale — still references frontend changes (isSendingRef guard, orphaned tool recovery, CopilotPage overlay) that have been dropped. Should be updated to reflect the backend-only scope.

majdyz: 🤖 Updated the PR description to reflect the backend-only scope — removed references to frontend changes (isSendingRef guard, orphaned tool recovery, CopilotPage overlay).

autogpt_platform/backend/backend/copilot/db.py
```diff
@@ -392,5 +396,13 @@ async def set_turn_duration(session_id: str, duration_ms: int) -> None:
...
+        # Update cache in-place rather than invalidating to avoid a
+        # race window where the empty cache gets re-populated with
+        # stale data by a concurrent get_chat_session call.
+        session = await get_chat_session_cached(session_id)
+        if session and session.messages:
+            for msg in reversed(session.messages):
+                if msg.role == "assistant":
+                    msg.duration_ms = duration_ms
+                    break
+            await cache_chat_session(session)
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in baee2c36c1: `set_turn_duration` now acquires the session lock (`_get_session_lock`) before performing the read-modify-write on the cache, preventing races with `upsert_chat_session`.

autogpt_platform/backend/backend/copilot/db.py
```diff
@@ -392,5 +396,13 @@ async def set_turn_duration(session_id: str, duration_ms: int) -> None:
...
-        await invalidate_session_cache(session_id)
+        # Update cache in-place rather than invalidating to avoid a
+        # race window where the empty cache gets re-populated with
+        # stale data by a concurrent get_chat_session call.
+        session = await get_chat_session_cached(session_id)
+        if session and session.messages:
+            for msg in reversed(session.messages):
+                if msg.role == "assistant":
+                    msg.duration_ms = duration_ms
+                    break
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in baee2c36c1: `cache_chat_session` is now only called when an assistant message was actually found and updated (tracked via an `updated` flag).

autogpt_platform/backend/backend/copilot/db.py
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in baee2c36c1 (same commit as above): the session lock acquisition addresses this race condition as well.

autogpt_platform/backend/backend/copilot/sdk/service.py
```diff
@@ -1272,15 +1278,40 @@ async def _run_stream_attempt(
...
+                        idle_seconds,
+                    )
+                    raise _HandledStreamError(
+                        f"Idle timeout after {idle_seconds:.0f}s",
+                        error_msg=(
+                            "A tool call appears to be stuck "
+                            "(no response for 10 minutes). "
+                            "Please try again."
+                        ),
+                        code="idle_timeout",
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in cfcc43d7be: Now yields `StreamError` to the client before raising `_HandledStreamError`, so the frontend receives the idle timeout message instead of a silent stream termination.

autogpt_platform/backend/backend/copilot/sdk/service.py
```diff
@@ -1670,19 +1713,12 @@ async def stream_chat_completion_sdk(
...
-        )
-    ):
-        session.messages.append(ChatMessage(role=new_message_role, content=message))
+    if maybe_append_user_message(session, message, is_user_message):
         if is_user_message:
             track_user_message(
-                user_id=user_id, session_id=session_id, message_length=len(message)
+                user_id=user_id,
+                session_id=session_id,
+                message_length=len(message or ""),
```
majdyz, line 1724: 🤖 🔵 **Nit**: `len(message or "")` is redundant — at this point `maybe_append_user_message` returned `True`, which means `message` passed the `if not message: return False` check, guaranteeing it's a non-empty string. `len(message)` would suffice. Same applies in both `baseline/service.py:406` and `sdk/service.py:1721`. Not worth a separate commit — just noting for awareness.

majdyz: 🤖 Acknowledged — keeping `or ""` as-is since `message` is typed `str | None` and the type checker cannot infer the narrowing through `maybe_append_user_message`. The `or ""` is the cleanest way to satisfy both correctness and type safety.

autogpt_platform/backend/backend/copilot/sdk/service.py
```diff
@@ -1272,15 +1278,40 @@ async def _run_stream_attempt(
...
+                # call is likely hung (e.g. WebSearch provider not responding).
+                idle_seconds = time.monotonic() - _last_real_msg_time
+                if idle_seconds >= _IDLE_TIMEOUT_SECONDS:
+                    logger.error(
+                        "%s Idle timeout after %.0fs with no SDK message — "
+                        "aborting stream (likely hung tool call)",
+                        ctx.log_prefix,
+                        idle_seconds,
+                    )
+                    raise _HandledStreamError(
```
majdyz: 🤖 🟠 **Should Fix**: The idle timeout raises `_HandledStreamError` without first yielding a `StreamError` to the client. The `_HandledStreamError` except handler (line 2152) assumes a `StreamError` was already yielded — it sets `ended_with_stream_error = True` and breaks, but since `stream_err` stays `None`, the fallback `StreamError` yield at line 2247 is skipped. The client SSE stream ends silently with no error notification.

Fix: yield `StreamError` before raising, matching the pattern used by the transient-error and circuit-breaker paths:

```python
yield StreamError(
    errorText="A tool call appears to be stuck (no response for 10 minutes). Please try again.",
    code="idle_timeout",
)
raise _HandledStreamError(...)
```

However, this is inside an `async for` loop, not a direct yield context of the outer generator. The `_run_stream_attempt` function IS an async generator (it yields `StreamHeartbeat()` a few lines above), so yielding here should work. The yielded `StreamError` will propagate through the outer generator in `stream_chat_completion_sdk`.

autogpt_platform/backend/backend/copilot/db.py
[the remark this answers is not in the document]
majdyz: 🤖 Acknowledged — this is a theoretical race but low severity in practice. `set_turn_duration` runs after the stream completes (post-processing in the `finally` block), by which time intermediate flushes are done. The worst case would be overwriting a very recent flush with a version that only differs by `duration_ms`, and the next `upsert_chat_session` call would restore the full state. Adding locking here would introduce contention for a marginal benefit. Leaving as-is for now.

autogpt_platform/backend/backend/copilot/sdk/service.py
```diff
@@ -1272,15 +1278,40 @@ async def _run_stream_attempt(
...
+                        f"Idle timeout after {idle_seconds:.0f}s",
+                        error_msg=(
+                            "A tool call appears to be stuck "
+                            "(no response for 10 minutes). "
+                            "Please try again."
+                        ),
+                        code="idle_timeout",
+                        retryable=True,
+                    )
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [`2bdecb6`](https://github.com/Significant-Gravitas/AutoGPT/commit/2bdecb6a58bf7b8ea5227b9ad09f6f6d466f6a76): Replaced the direct `_HandledStreamError` raise with the yield-then-break pattern used by the transient-error and circuit-breaker paths. The client now receives a `StreamError` event with `code=idle_timeout` before the stream closes.

autogpt_platform/backend/backend/copilot/sdk/service.py
majdyz: 🤖 Fixed in [`2bdecb6`](https://github.com/Significant-Gravitas/AutoGPT/commit/2bdecb6a58bf7b8ea5227b9ad09f6f6d466f6a76): Replaced the direct raise with the yield-then-break pattern — now appends error marker, yields `StreamError(code="idle_timeout")`, sets `ended_with_stream_error`, and breaks. This matches the transient-error and circuit-breaker paths.

PR #12668, opened 2026-04-03
fix(copilot): prevent 524 timeout on chat deletion by deferring cleanup

Otto-AGPT:
Requested by @Pwuts

## Why

Deleting a currently-selected CoPilot chat causes the delete dialog to hang indefinitely. The network inspector shows the DELETE request returning **HTTP 524** (Cloudflare timeout) because the endpoint awaits slow cleanup operations (browser daemon shutdown ~10s, E2B sandbox teardown ~10s) before responding. The backend *does* delete the session, but the client never gets the 204 back in time.

## What

**Backend — defer slow cleanup to background tasks:**
- `routes.py`: Move E2B sandbox `kill_sandbox()` into `asyncio.create_task()` so the 204 response returns immediately after the DB delete succeeds
- `model.py`: Move `close_browser_session()` into `asyncio.create_task()` for the same reason

**Frontend — improve resilience:**
- Reorder `onSuccess`: close the dialog (`setSessionToDelete(null)`) *before* clearing the selected session (`setSessionId(null)`), so the UI updates instantly before the chat-view teardown cascade runs
- Improve `onError` to also refresh the session list and navigate away from the deleted chat — the server may have completed the delete even when the client receives a timeout error (524)

## Testing
1. Select a chat → delete → dialog closes immediately, chat gone from sidebar
2. Delete a non-selected chat → still works (no regression)
3. Delete a chat with active E2B sandbox → response is fast, sandbox cleaned up in background
4. Check server logs to confirm background cleanup tasks complete successfully

Resolves SECRT-2215


ntindle:
@copilot resolve the merge conflicts in this pull request + make sure we hit our coverage targets on frontend and backend. also use /pr-address to address anything neded + also make sure we still neeed/wnat this

PR #12676, opened 2026-04-04
Persist stable copilot message IDs through hydration

rotempasharel1:
## Summary
This starts propagating stable chat message IDs into copilot hydration so UI messages can reuse backend IDs instead of synthetic session/index IDs.

## Changes
- Added optional `id` to the backend `ChatMessage` model and mapped it from DB
- Updated `convertChatSessionMessagesToUiMessages` to preserve backend message IDs
- Updated copilot hydration logic to replace synthetic IDs with hydrated backend IDs when available

## Notes
This keeps the existing deduplication fallback in place for safety while verifying hydration and stream behavior.


rotempasharel1:
All functional checks are green now. The remaining failures seem to be the PR status checker, a small Codecov drop, and Vercel authorization.

autogpt_platform/backend/backend/copilot/model.py
```diff
@@ -57,6 +57,7 @@ class ChatSessionMetadata(BaseModel):
 
 
 class ChatMessage(BaseModel):
+    id: str | None = None
```
majdyz, line 60: 🤖 🟠 **Should Fix / Architectural Concern**: This PR only fixes half the story.

The backend generates **two separate UUIDs** for the same logical message:
1. **Stream `messageId`** — generated at [`sdk/service.py:1706`](autogpt_platform/backend/backend/copilot/sdk/service.py#L1706) via `str(uuid.uuid4())`, emitted to the frontend in `StreamStart(messageId=...)` ([`response_model.py:67`](autogpt_platform/backend/backend/copilot/response_model.py#L67)). The Vercel AI SDK uses this as the streamed `UIMessage.id`.
2. **DB `id`** — generated independently by Prisma's `@default(uuid())` when `add_chat_messages_batch` inserts the row in [`db.py:209`](autogpt_platform/backend/backend/copilot/db.py#L209) (no `id` is passed through from the stream).

After this PR:
- **During streaming**: UI has message `stream-uuid-A`
- **After page reload / refetch**: hydration replaces it with `db-uuid-B` (a *different* UUID for the same logical message)

So IDs still silently rotate across the stream→persist→reload boundary. The content-fingerprint dedup in [`helpers.ts:77`](autogpt_platform/frontend/src/app/(platform)/copilot/helpers.ts#L77) is still the only thing catching the ID mismatch at that boundary.

**If the goal is truly stable IDs**, the fix is to make the stream messageId and the DB id the same value — either by:
- Passing `message_id` from `sdk/service.py` into the `messages_data` dict in [`model.py:579`](autogpt_platform/backend/backend/copilot/model.py#L579), setting it as the DB row's `id`, OR
- Re-emitting the DB-assigned id via a late `message-id` stream event the SDK can re-key to.

Otherwise this PR is cosmetic — it swaps `sessionId-index` for `db-uuid` but doesn't actually unify identity across the lifecycle.

autogpt_platform/frontend/src/app/(platform)/copilot/useCopilotStream.ts
```diff
@@ -362,12 +362,20 @@ export function useCopilotStream({
...
     if (isReconnectScheduled) return;
+
     setMessages((prev) => {
-      if (prev.length >= hydratedMessages.length) return prev;
+      const syntheticPrefix = sessionId ? `${sessionId}-` : null;
+      const hasSyntheticIds = syntheticPrefix
+        ? prev.some((m) => m.id.startsWith(syntheticPrefix))
+        : false;
+
+      if (prev.length >= hydratedMessages.length && !hasSyntheticIds) {
```
majdyz: 🤖 🟠 **Should Fix**: The `hasSyntheticIds` branch can **lose optimistic messages**.

Scenario:
1. `prev` has 5 messages, one of which has a synthetic `${sessionId}-N` ID (e.g. from an older session that hydrated before this PR was deployed, or because `msg.id` came back `null` from the API for some message)
2. User sends a new message — SDK adds it optimistically → `prev` now has 6 messages
3. `hydratedMessages` refetches with only 5 messages (new user message not yet persisted)
4. `hasSyntheticIds` is `true` → condition `prev.length >= hydratedMessages.length && !hasSyntheticIds` is `false` → **we replace prev (6 msgs) with hydratedMessages (5 msgs)**, losing the user's just-sent message

Consider an ID-level merge instead of full-array replace: map hydrated backend IDs onto matching synthetic-id entries (by sequence/position), and preserve any prev entries not covered by hydratedMessages.

**Why**: the current guard (`prev.length >= hydratedMessages.length`) was protecting against exactly this race; dropping it when synthetic IDs are present reintroduces the race for anyone on an in-flight session during the rollout.

autogpt_platform/frontend/src/app/(platform)/copilot/helpers/convertChatSessionToUiMessages.ts
```diff
@@ -200,7 +207,7 @@ export function convertChatSessionMessagesToUiMessages(
       return;
     }
 
-    const msgId = `${sessionId}-${index}`;
+    const msgId = msg.id ?? `${sessionId}-${index}`;
```
majdyz: 🤖 🟡 **Nice to Have / Question**: What concrete bug does this actually fix?

The PR description says it lets UI messages "reuse backend IDs instead of synthetic session/index IDs", and promises to keep the dedup fallback "for safety". But:
- The synthetic `${sessionId}-${index}` IDs were already **deterministic + unique per session+position** — they don't collide, and React reconciliation works fine with them.
- Hydrated messages don't interact with stream messages by ID anyway (stream IDs come from `StreamStart.messageId`, a completely separate UUID generator in the backend).
- So after this PR, the only visible effect is: DB-row UUIDs show up in `UIMessage.id` after hydration. Is there downstream code (analytics, URL fragments, bookmarking, server-side per-message reactions) that actually needs stable DB-linked IDs?

If yes — please call that out in the PR description. If no — the PR is introducing API-schema churn (`id` on `ChatMessage`) and hydration-logic churn for a cosmetic rename, while not addressing the real ID divergence between stream and DB.

**Why**: future maintainers reading this diff will assume IDs are now stable end-to-end — and will be confused when stream and DB UUIDs still don't match.

autogpt_platform/frontend/src/app/(platform)/copilot/helpers/convertChatSessionToUiMessages.ts
```diff
@@ -21,6 +22,12 @@ function coerceSessionChatMessages(
       if (!role) return null;
 
       return {
+        id:
+          typeof msg.id === "string"
+            ? msg.id
+            : msg.id == null
+              ? null
+              : String(msg.id),
```
majdyz, line 31: 🤖 🔵 **Nit**: DB `id` is `String @id @default(uuid())` per `schema.prisma:237` — always a string UUID. The `msg.id == null ? null : String(msg.id)` fallback (coercing non-string, non-null values) will never fire in practice. Could simplify to:

```ts
id: typeof msg.id === "string" ? msg.id : null,
```

majdyz (changes requested):
## Summary

This PR partially addresses message ID stability but leaves the underlying architectural problem unsolved, and introduces a new race that can drop optimistic messages. Requesting changes before merge.

## What the PR does (and doesn't do)

✅ Hydrated REST-API messages now carry the DB row's `id` instead of synthetic `${sessionId}-${index}` strings.

❌ **Stream messages still don't match DB rows.** The backend generates the ID in two unrelated places:

| Path | Where | Generator |
|------|-------|-----------|
| SSE stream | `sdk/service.py:1706` | `str(uuid.uuid4())` → `StreamStart(messageId=…)` |
| DB row | `db.py:209` via `add_chat_messages_batch` | Prisma `@default(uuid())` (no id passed through) |

These are two independent UUIDs for the same logical message.

## Why this is not enough — the lifecycle today

1. **During stream:** AI SDK stores message under `stream-uuid-A` (from `StreamStart`)
2. **Stream ends, backend persists:** DB row gets a totally unrelated `db-uuid-B`
3. **Page reload / refetch:** hydration replaces the message with `db-uuid-B`

The UI's `UIMessage.id` for the *same logical message* silently rotates across the stream→persist→reload boundary. The content-fingerprint dedup in `helpers.ts:77` is still the only thing holding dedup together at that boundary — and that's a fragile fallback, not a fix.

## Risks of landing this partially

1. **False sense of stability.** Reviewers and future maintainers will read "Persist stable copilot message IDs through hydration" and assume IDs are now end-to-end stable. They aren't. The stream still emits a throwaway UUID that never reaches the DB. Anyone building downstream features (per-message analytics, feedback widgets, URL fragments, server-side reactions) on `UIMessage.id` will hit subtle bugs the next time a session reloads.

2. **Optimistic-message data loss race** (see inline comment). The new `hasSyntheticIds` escape hatch bypasses the `prev.length >= hydratedMessages.length` guard. If a user sends a message while a legacy session still has any synthetic IDs in `prev`, a refetch can replace a longer `prev` with a shorter `hydratedMessages` and drop the unpersisted user message.

3. **API-surface churn without matching benefit.** Exposing `id` on `ChatMessage` is a public-schema change. It's worth it if downstream code needs stable DB-linked IDs — but the PR description doesn't name a concrete consumer, and nothing in the diff wires one up.

4. **The dedup fallback will silently outlive the PR.** The comment in `helpers.ts:77` explicitly says content-fingerprint dedup exists to catch "hydration/stream boundary duplicates (where the same content appears under different IDs)." That boundary is still there after this PR. If someone reads the PR description and removes the fallback thinking IDs are now stable, duplicates come back.

## What would make this complete

Either of these would actually fix the lifecycle:

- **Option A (preferred):** pass `message_id` from `sdk/service.py` into the `messages_data` dict at `model.py:579`, setting it as the DB row's `id`. The stream UUID and DB UUID become the same value, and `UIMessage.id` stays stable from first stream chunk through persistence to reload.
- **Option B:** emit a late `message-id` stream event carrying the DB-assigned id, and re-key the SDK message on receipt.

With either in place, the hydration-side change in this PR becomes genuinely useful and the content-fingerprint dedup can be dropped.

## Asks

- [ ] Pick Option A or B and make IDs coherent across stream → persist → hydrate
- [ ] Fix the `hasSyntheticIds` race (either ID-level merge, or re-add the length guard for the synthetic-id case)
- [ ] Add the concrete motivation to the PR description — what breaks today without stable DB IDs?
- [ ] Add a test that exercises stream → persist → reload and asserts `UIMessage.id` is identical across all three stages

Tag @majdyz for re-review once the stream and DB IDs are unified (Option A/B landed). The hydration-only path alone isn't enough to merge.

rotempasharel1:
Thanks for the detailed review. You're right that the current change only improves hydration-side IDs and does not yet unify stream and persisted IDs. I'll rework this toward using the same message ID across stream -> persist -> reload, and I'll also revert the risky hydration replacement logic and add test coverage.

autogpt_platform/backend/backend/copilot/sdk/service.py
```diff
@@ -476,6 +482,7 @@ async def _reduce_context(
     attempt: int = 1,
     runtime_model: str | None = None,
 ) -> ReducedContext:
+    effective_model = runtime_model or config.agent_model
```
majdyz: 🤖 🔴 **Blocker**: `config.agent_model` is not a real attribute on the copilot config — the actual field is `config.claude_agent_model` (see `autogpt_platform/backend/backend/copilot/config.py:185`). This line will raise `AttributeError` the first time `_reduce_context` runs without an explicit `runtime_model`, which is exactly the hot path on the first retry attempt.

`effective_model` also isn't referenced anywhere else in this function — `target_model` on line 511 still re-derives the same value from `runtime_model or config.thinking_standard_model`. So this line is both wrong and dead. Drop it.

autogpt_platform/backend/backend/copilot/sdk/service.py
```diff
@@ -3286,6 +3357,11 @@ async def _setup_e2b():
             permissions=permissions,
         )
 
+        # Fail fast when no API credentials are available at all.
+        # sdk_cwd routes the CLI's temp dir into the per-session workspace
+        # so sub-agent output files land inside sdk_cwd (see build_sdk_env).
+        sdk_env = build_sdk_env(session_id=session_id, user_id=user_id, sdk_cwd=sdk_cwd)
```
majdyz, line 3367: 🤖 🔴 **Blocker** (unintended revert of a merged fix): This call used to pass `model=_resolve_env_model(sdk_model, fallback_model)` so that `build_sdk_env` could gate Moonshot-specific env vars (notably suppress `CLAUDE_AUTOCOMPACT_PCT_OVERRIDE` for Kimi). That gate was added in #12889 specifically to prevent the Kimi compaction storm.

Dropping the `model=` argument here silently regresses #12889 — Kimi/Moonshot turns will start re-hitting autocompact churn in production. `_resolve_env_model` is still defined in this file but is now dead code.

This is unrelated to the PR's stated scope ("stable copilot message IDs"). Looks like merge-conflict rot from the `Merge branch 'dev'` commits wiped out this line. Please restore the original call:

```python
sdk_env = build_sdk_env(
    session_id=session_id,
    user_id=user_id,
    sdk_cwd=sdk_cwd,
    model=_resolve_env_model(sdk_model, fallback_model),
)
```

and move it back below the `sdk_model` / `fallback_model` resolution (it needs those values).

autogpt_platform/backend/backend/copilot/sdk/service.py
```diff
@@ -3861,10 +3862,9 @@ def _on_stderr(line: str) -> None:
                             transient_retries,
                             max_transient_retries,
                         )
-                        async for evt in _do_transient_backoff(
-                            backoff, state, message_id, session_id
-                        ):
+                        async for evt in _do_transient_backoff(backoff, state):
                             yield evt
+                        current_message_id = str(uuid.uuid4())
```
majdyz: 🤖 🔴 **Blocker** (defeats this PR's own goal): Minting a fresh `str(uuid.uuid4())` per transient retry means the retried turn lands in the DB under a **different** id than the one the stream already emitted via `StreamStart(messageId=...)`. That breaks the very stream → persist → reload stability this PR is trying to deliver — the moment any retry fires, the UI's optimistic `UIMessage.id` no longer matches the row eventually written to `ChatMessage`.

Same pattern repeats at lines 3950 and 3971. All three call sites should reuse `ctx.message_id` (or keep `current_message_id` pinned to its initial value) — the whole point of threading `ctx.message_id` into `ChatMessage(id=ctx.message_id, …)` at line 2210 is that it stays stable across retries.

If the intent is to emit a *new* `StreamStart` event on retry (so the frontend knows a retried message is replacing the failed one), that should be an explicit protocol decision with matching frontend handling — not a silent side effect of the backoff helper.

autogpt_platform/backend/backend/copilot/sdk/service.py
```diff
@@ -3462,16 +3526,6 @@ def _on_stderr(line: str) -> None:
...
-        # the finally block — ``propagate_attributes`` only annotates an
-        # existing span, it does not create one.
-        try:
-            _lf_span = get_client().start_as_current_span(name="copilot-sdk-turn")
-            _lf_span.__enter__()
-        except Exception:
-            logger.debug("Failed to open Langfuse parent span", exc_info=True)
-            _lf_span = None
-
         _otel_ctx = propagate_attributes(
```
majdyz, line 3533: 🤖 🔴 **Blocker** (unintended revert of a merged fix): The Langfuse parent span that #12889 introduced (`_lf_span = get_client().start_as_current_span("copilot-sdk-turn")`) has been deleted along with the `get_client` import. `propagate_attributes` alone only annotates an *existing* span — it doesn't create one — so without that wrapper the `openrouter-cost-reconcile` backfill event fires in the finally block with no live span to attach to, and `get_current_trace_id()` returns `None`. Cost reconciliation loses its trace linkage again, which is exactly the regression #12889 fixed.

This is unrelated to the PR's stated scope. Please restore:

1. `from langfuse import get_client, propagate_attributes` on line 38.
2. The `_lf_span` declaration near the `_otel_ctx` init, and the `start_as_current_span` block that wraps the turn and is `__exit__`ed in the finally block.

Same root cause as the `build_sdk_env(..., model=...)` removal below — looks like merge-conflict rot from the repeated `Merge branch 'dev'` commits on this branch silently dropped several recently-landed copilot fixes.

majdyz (changes requested):
Re-requesting changes — flagging this as a **blocker** rather than nit-level feedback. Full breakdown:

## Why this is blocking

### 1. Silent revert of already-merged copilot fixes
The `service.py` diff (+172/-196) is far larger than the PR's stated scope ("stable copilot message IDs") and has quietly undone multiple recently-landed fixes. Almost certainly merge-conflict rot from the four `Merge branch 'dev'` commits on this branch. Specifically:

- **#12889 Langfuse parent span** — deleted → `openrouter-cost-reconcile` loses its trace_id again. See inline on line 3529.
- **Moonshot autocompact gate** — `model=` dropped from `build_sdk_env(...)`, `_resolve_env_model` is now dead code → Kimi regresses into the compaction storm #12889 just fixed. See inline on line 3363.
- **`config.agent_model` AttributeError** — a newly introduced, unreferenced line that will crash `_reduce_context` because the real attribute is `config.claude_agent_model`. See inline on line 485.
- Comment/blank-line churn around `_compaction_target_tokens` and `_reduce_context`.

None of this is mentioned in the PR description, so shipping it would silently regress #12889 in production.

### 2. New regression that defeats the PR's own goal
On every transient retry, `current_message_id = str(uuid.uuid4())` mints a fresh UUID (lines 3867, 3950, 3971). The retried turn then persists under a brand-new id, so the UI's optimistic `UIMessage.id` no longer matches the DB row after any retry. A "stable ID" PR cannot ship a fresh-ID-per-retry path. See inline on line 3867.

### 3. Previously-raised issues still open from my earlier review
- `hasSyntheticIds` still bypasses the `prev.length >= hydratedMessages.length` guard → optimistic-message drop race on legacy sessions.
- No test that exercises stream → persist → reload and asserts `UIMessage.id` is identical across all three stages. Without that, the stated outcome isn't verified anywhere.

## What will unblock

1. Hard-reset `autogpt_platform/backend/backend/copilot/sdk/service.py` to `origin/dev`, then re-apply **only** the intentional hunks:
   - `ChatMessage(id=ctx.message_id, …)` in `_StreamAccumulator` (line 2210).
   - Reuse `ctx.message_id` across retries (don't mint a new UUID per attempt).
2. Keep `model.py`, `db.py`, the frontend converter, and the existing test as-is — those hunks are correct.
3. Fix the `hasSyntheticIds` optimistic-drop race (ID-level merge, or restore the length guard).
4. Add a stream → persist → reload ID-stability test, and **please attach a screenshot of that test passing** in the PR so we have a visible receipt of the fix.

Once the above is addressed, feel free to retag me for review. Happy to take another pass as soon as `service.py` is back down to the minimal intentional delta.

ntindle:
**Closing — superseded by work already merged to `dev`.**

Stable copilot message IDs shipped via #12948 (commit `529c203e7f`): `ChatMessage.id` with per-send PK-dedup semantics, mapped through `from_db`. The frontend preserves the backend `msg.id` since #13069 (commit `e4cc0b8820`), and the copilot stack was further rewritten in #13243. Dev's pagination-merge logic now depends on the deterministic `${sessionId}-seq-${sequence}` id shape, so this PR's hydration-replacement approach no longer fits the current architecture — every behavior it targeted exists on `dev` in equivalent-or-better form.

Thanks for the work on this — the problem was real and is now fixed.

If this is wrong, comment here or reopen — this was an automated triage pass reviewed by a maintainer.

PR #12679, opened 2026-04-05
dx(platform): align feature flag infrastructure — env overrides + LD context parity

majdyz:
## Summary

Two related feature-flag infrastructure fixes.

### 1. FORCE_FLAG env-var overrides for LaunchDarkly flags

Both [backend/backend/util/feature_flag.py](autogpt_platform/backend/backend/util/feature_flag.py) and [services/feature-flags/use-get-flag.ts](autogpt_platform/frontend/src/services/feature-flags/use-get-flag.ts) now check an env-var override **before** consulting LaunchDarkly:

- Backend: `FORCE_FLAG_<NAME>=true|false` (also accepts `NEXT_PUBLIC_FORCE_FLAG_<NAME>` as an alias)
- Frontend: `NEXT_PUBLIC_FORCE_FLAG_<NAME>=true|false` (must be `NEXT_PUBLIC_` prefixed so Next.js exposes it to the browser bundle)

`<NAME>` is the flag value upper-cased with `-` → `_`. Example: `chat-mode-option` → `CHAT_MODE_OPTION`.

**Why:** previously, testing a flag-gated feature locally required patching `default=` at each call site (tracked code change) or unsetting `NEXT_PUBLIC_LAUNCHDARKLY_CLIENT_ID` and rebuilding. Both had to be reverted manually. `.env` overrides are gitignored.

### 2. Align LaunchDarkly user context (frontend ↔ backend)

Frontend was sending a smaller context than backend, so targeting rules using `email_domain` or top-level `role` could evaluate differently on each side and silently diverge (e.g. frontend shows a feature, backend denies it — observed during PR #12623 testing).

**Before:**
```typescript
// Frontend
{ key, email, anonymous, custom: { role } }
```
```python
# Backend
{ key, anonymous, role, custom: {"role"}, email, email_domain }
```

**After:** frontend now builds the same shape as backend.

## Test plan

- [x] Backend: `FORCE_FLAG_CHAT_MODE_OPTION=true` exposes `mode=fast/extended_thinking` routing (verified on PR #12623 E2E run)
- [x] Frontend: `NEXT_PUBLIC_FORCE_FLAG_CHAT_MODE_OPTION=true` renders the ChatInput mode toggle (verified on PR #12623 E2E run)
- [x] Backend accepts the `NEXT_PUBLIC_FORCE_FLAG_` alias
- [x] Falls through to LaunchDarkly when env var is unset
- [x] Pre-commit: ruff, prettier, tsc, black, isort all pass

### DIFF
--- a/autogpt_platform/backend/backend/util/feature_flag.py
+++ b/autogpt_platform/backend/backend/util/feature_flag.py
@@ -1,5 +1,6 @@
 import contextlib
 import logging
+import os
 from enum import Enum
 from functools import wraps
 from typing import Any, Awaitable, Callable, TypeVar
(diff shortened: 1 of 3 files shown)


majdyz:
Bundling these into #12623 since both changes (FORCE_FLAG env override + LaunchDarkly context alignment) came directly out of testing the copilot mode toggle. Changes cherry-picked as c91b263591 and 8a7f33dfaf.

PR #12688, opened 2026-04-06
fix(backend/copilot): skip AI blocks without model property in fix_ai_model_parameter

anvyle:
### Why / What / How

**Why:** Some AI-category blocks do not expose a `"model"` input property in their `inputSchema`. The `fix_ai_model_parameter` fixer was unconditionally injecting a default model value (e.g. `"gpt-4o"`) into any node whose block has category `"AI"`, regardless of whether that block actually accepts a `model` input. This causes the agent JSON to include an invalid field for those blocks.

**What:** Guard the model-injection logic with a check that `"model"` exists in the block's `inputSchema.properties` before attempting to set or validate the field. AI blocks that have no model selector are now skipped entirely.

**How:** In `fix_ai_model_parameter`, after confirming `is_ai_block`, extract `input_properties` from the block's `inputSchema.properties` and `continue` if `"model"` is absent. The subsequent `model_schema` lookup is also simplified to reuse the already-fetched `input_properties` dict. A regression test is added to cover this case.

### Changes 🏗️

- `backend/copilot/tools/agent_generator/fixer.py`: In `fix_ai_model_parameter`, skip AI-category nodes whose block `inputSchema.properties` does not contain a `"model"` key; reuse `input_properties` for the subsequent `model_schema` lookup.
- `backend/copilot/tools/agent_generator/fixer_test.py`: Add `test_ai_block_without_model_property_is_skipped` to `TestFixAiModelParameter`.

#### For code changes:
- [ ] I have made a test plan
  - [ ] Run `poetry run pytest backend/copilot/tools/agent_generator/fixer_test.py` — all 50 tests pass (49 pre-existing + 1 new)

### DIFF
--- a/autogpt_platform/backend/backend/copilot/tools/agent_generator/fixer.py
+++ b/autogpt_platform/backend/backend/copilot/tools/agent_generator/fixer.py
@@ -890,6 +890,12 @@ class AgentFixer:
             )
 
             if is_ai_block:
+                # Skip AI blocks that don't expose a "model" input property
+                # (some AI-category blocks have no model selector at all).
+                input_properties = block.get("inputSchema", {}).get("properties", {})
+                if "model" not in input_properties:
+                    continue
+
                 node_id = node.get("id")
                 input_default = node.get("input_default", {})
                 current_model = input_default.get("model")
@@ -898,9 +904,7 @@ class AgentFixer:
                 # Blocks with a block-specific enum on the model field (e.g.
                 # PerplexityBlock) use their own enum values; others use the
                 # generic set.
-                model_schema = (
-                    block.get("inputSchema", {}).get("properties", {}).get("model", {})
-                )
+                model_schema = input_properties.get("model", {})
                 block_model_enum = model_schema.get("enum")
 
                 if block_model_enum:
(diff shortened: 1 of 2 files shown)


majdyz (approved):
## Review Summary

**Verdict: Approve** ✅

Clean, well-scoped bugfix. The guard clause correctly prevents model injection into AI-category blocks that don't expose a `model` input property.

### What's good
- Minimal, targeted fix — no unnecessary refactoring beyond the immediate scope
- The `input_properties` extraction + reuse eliminates a redundant schema traversal
- Test covers the exact regression scenario
- Clear PR description with good context on why this was needed

### Minor observations
- The commit history on this PR includes 8 unrelated commits (ask_question tool, frontend refactors, etc.) from the base branch — consider rebasing or squash-merging to keep the main branch history clean
- Noted the CLA blocker — author will need to sign before merge

LGTM — nice catch on the invalid JSON injection.

majdyz (approved):
## Re-review after rebase

**Verdict: Approve** ✅

Diff is unchanged from the previous review — the fix is still clean and correct. Commit history is now a single squashed commit, which addresses the earlier feedback about unrelated commits. Much cleaner.

No new concerns. Still LGTM.

PR #12691, opened 2026-04-07
dx: add /orchestrate skill — Claude Code agent fleet supervisor with spare worktree lifecycle

majdyz:
### Why

When running multiple Claude Code agents in parallel worktrees, they frequently get stuck: an agent exits and sits at a shell prompt, freezes mid-task, or waits on an approval prompt with no human watching. Fixing this currently requires manually checking each tmux window.

### What

Adds a `/orchestrate` skill — a meta-agent supervisor that manages a fleet of Claude Code agents across tmux windows and spare worktrees. It auto-discovers available worktrees, spawns agents, monitors them, kicks idle/stuck ones, auto-approves safe confirmations, and recycles worktrees on completion.

### How to use

**Prerequisites:**
- One tmux session already running (the skill adds windows to it; it does not create a new session)
- Spare worktrees on `spare/N` branches (e.g. `AutoGPT3` on `spare/3`, `AutoGPT7` on `spare/7`)

**Basic workflow:**

```
/orchestrate capacity     → see how many spare worktrees are free
/orchestrate start        → enter task list, agents spawn automatically
/orchestrate status       → check what's running
/orchestrate add          → add one more task to the next free worktree
/orchestrate stop         → mark inactive (agents finish current work)
/orchestrate poll         → one manual poll cycle (debug / on-demand)
```

**Worktree lifecycle:**
```text
spare/N branch → /orchestrate add → new window + feat/branch + claude running
                                              ↓
                                     ORCHESTRATOR:DONE
                                              ↓
                              kill window + git checkout spare/N
                                              ↓
                                     spare/N (free again)
```

Windows are always capped by worktree count — no creep.

### Changes

- `.claude/skills/orchestrate/SKILL.md` — skill definition with 5 subcommands, state file schema, spawn/recycle helpers, approval policy
- `.claude/skills/orchestrate/scripts/classify-pane.sh` — pane state classifier: `idle` (shell foreground), `running` (non-shell), `waiting_approval` (pattern match), `complete` (ORCHESTRATOR:DONE)
- `.claude/skills/orchestrate/scripts/poll-cycle.sh` — poll loop: reads/updates state file atomically, outputs JSON action list, stuck detection via output-hash sampling

**State detection:**

| State | Detection method |
|---|---|
| `idle` | `pane_current_command` is a shell (zsh/bash/fish) |
| `running` | `pane_current_command` is non-shell (claude/node) |
| `stuck` | pane hash unchanged for N consecutive polls |
| `waiting_approval` | pattern match on last 40 lines of pane output |
| `complete` | `ORCHESTRATOR:DONE` string present in pane output |

**Safety policy for auto-approvals:** git ops, package installs, tests, docker compose → approve. `rm -rf` outside worktree, force push, `sudo`, secrets → escalate to user.

State file lives at `~/.claude/orchestrator-state.json` (outside repo, never committed).

#### For code changes:
- [x] I have made a test plan
  - [x] `classify-pane.sh`: idle shell → `idle`, running process → `running`, `ORCHESTRATOR:DONE` → `complete`, approval prompt → `waiting_approval`, nonexistent window → `error`
  - [x] `poll-cycle.sh`: inactive state → `[]`, empty agents array → `[]`, spare worktree discovery, stuck detection (3-poll hash cycle)
  - [x] Real agent spawn in `autogpt1` tmux session — agent ran, output `ORCHESTRATOR:DONE`, recycle verified
  - [x] Upfront JSON validation before `set -e`-guarded jq reads
  - [x] Idle timer reset only on `idle → running` transition (not stuck), preventing false stuck-detections
  - [x] Classify fallback only triggers when output is empty (no double-JSON on classify exit 1)

### DIFF
--- a/.claude/skills/orchestrate/SKILL.md
+++ b/.claude/skills/orchestrate/SKILL.md
@@ -144,12 +144,12 @@ Tell the user:
 ## Subcommand: stop
 
 ```bash
+# Show current agent state before marking inactive
+jq '.agents[] | {window, state, worktree}' ~/.claude/orchestrator-state.json
+
 # Mark inactive
 jq '.active = false' ~/.claude/orchestrator-state.json > /tmp/orch.tmp \
   && mv /tmp/orch.tmp ~/.claude/orchestrator-state.json
-
-# Show current state before stopping
-jq '.agents[] | {window, state, worktree}' ~/.claude/orchestrator-state.json
 ```
 
 Then use CronDelete with the stored job ID to cancel the loop:
--- a/.claude/skills/orchestrate/SKILL.md
+++ b/.claude/skills/orchestrate/SKILL.md
@@ -1,34 +1,47 @@
 ---
 name: orchestrate
-description: "Meta-agent supervisor that manages a fleet of Claude Code agents running in tmux windows. Auto-discovers spare worktrees (spare/N branches), spawns agents into new windows, monitors state, kicks idle agents, auto-approves safe confirmations, and recycles worktrees when done. TRIGGER when user asks to supervise agents, manage a fleet, monitor tmux agents, or orchestrate parallel worktrees."
+description: "Meta-agent supervisor that manages a fleet of Claude Code agents running in tmux windows. Auto-discovers spare worktrees, spawns agents, monitors state, kicks idle agents, approves safe confirmations, and recycles worktrees when done. TRIGGER when user asks to supervise agents, run parallel tasks, manage worktrees, check agent status, or orchestrate parallel work."
 user-invocable: true
-argument-hint: "[start|stop|status|add|poll|capacity] — start spawns agents into spare worktrees, add assigns one more task, capacity shows available worktrees, poll runs one check cycle"
+argument-hint: "any free text — e.g. 'start 3 agents on X Y Z', 'show status', 'add task: implement feature A', 'stop', 'how many are free?'"
 metadata:
   author: autogpt-team
-  version: "2.0.0"
+  version: "3.0.0"
 ---
 
 # Orchestrate — Agent Fleet Supervisor
 
-One tmux session, N windows — each window is one agent in one worktree. The orchestrator auto-discovers spare worktrees, spawns agents, monitors them, and recycles worktrees when work is done. No manual window setup required.
+One tmux session, N windows — each window is one agent working in its own worktree. Speak naturally; Claude maps your intent to the right scripts.
+
+## Scripts
+
+```bash
+SKILLS_DIR=$(git rev-parse --show-toplevel)/.claude/skills/orchestrate/scripts
+STATE_FILE=~/.claude/orchestrator-state.json
+```
+
+| Script | Purpose | Key args |
+|---|---|---|
+| `find-spare.sh [REPO_ROOT]` | List free worktrees — one `PATH BRANCH` per line | |
+| `spawn-agent.sh SESSION PATH SPARE NEW_BRANCH OBJECTIVE` | Create window + checkout branch + launch claude + send task. **Stdout: `SESSION:WIN` only** | |
+| `recycle-agent.sh WINDOW PATH SPARE_BRANCH` | Kill window + restore spare branch | |
+| `capacity.sh [REPO_ROOT]` | Print available + in-use worktrees | |
+| `status.sh` | Print fleet status + live pane commands | |
+| `poll-cycle.sh` | One monitoring cycle — returns JSON action array | |
+| `classify-pane.sh WINDOW` | Classify one pane state | |
 
 ## Worktree lifecycle
 
 ```text
-spare/N branch   →   orchestrate add   →   new window + feat/branch + claude running
-                                                        ↓
-                                               ORCHESTRATOR:DONE
-                                                        ↓
-                                        kill window + git checkout spare/N
-                                                        ↓
-                                               spare/N (free again)
+spare/N branch  →  spawn-agent.sh  →  window + feat/branch + claude running
+                                                ↓
+                                         ORCHESTRATOR:DONE
+                                                ↓
+                               recycle-agent.sh → spare/N (free again)
 ```
 
-Windows are always capped by worktree count — no creep. Auto-close on completion is how the orchestrator signals a worktree is free for the next task.
+## State file (`~/.claude/orchestrator-state.json`)
 
-## State file
-
-Lives at `~/.claude/orchestrator-state.json` (outside repo, never committed):
+Never committed to git. Claude maintains this file directly using `jq` + atomic writes (`.tmp` → `mv`).
 
 ```json
 {
@@ -36,12 +49,12 @@ Lives at `~/.claude/orchestrator-state.json` (outside repo, never committed):
   "tmux_session": "autogpt1",
   "idle_threshold_seconds": 300,
   "cron_job_id": "...",
-  "last_poll_at": 1712345678,
+  "last_poll_at": 0,
   "agents": [
     {
       "window": "autogpt1:3",
       "worktree": "AutoGPT6",
-      "worktree_path": "/path/to/worktrees/AutoGPT6",
+      "worktree_path": "/path/to/AutoGPT6",
       "spare_branch": "spare/6",
       "branch": "feat/my-feature",
       "objective": "Implement X and open a PR",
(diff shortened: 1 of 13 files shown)


.claude/skills/orchestrate/scripts/poll-cycle.sh
```diff
@@ -0,0 +1,162 @@
...
+# and outputs a JSON array of actions for Claude to take.
+#
+# Usage: poll-cycle.sh
+# Output (stdout): JSON array of action objects
+#   [{ "window": "work:0", "action": "kick|approve|none", "state": "...",
+#      "worktree": "...", "objective": "...", "reason": "..." }]
+#
+# The state file is updated in-place (atomic write via .tmp).
+
+set -uo pipefail
```
majdyz: 🤖 🔴 **Blocker**: `poll-cycle.sh` uses `set -uo pipefail` but omits `set -e`, while `classify-pane.sh` uses `set -euo pipefail`. Without `-e`, command failures inside `poll-cycle.sh` (e.g. a failing `jq` invocation) silently continue rather than aborting. Change line 14 to `set -euo pipefail` to match `classify-pane.sh` and ensure early failure on errors.

.claude/skills/orchestrate/SKILL.md
```diff
@@ -0,0 +1,316 @@
...
+
+Manages multiple Claude Code agents running in tmux windows. Monitors their state, kicks them when idle, and auto-approves safe confirmations so they never block on human input.
+
+## Setup agents for orchestration
+
+Before registering an agent with the orchestrator, launch it in a tmux window with bypass permissions:
+
+```bash
+# In your tmux session (e.g., session named "work"):
+# window 0: cd /Users/majdyz/Code/AutoGPT6 && claude --permission-mode bypassPermissions
```
majdyz: 🤖 🟠 **Should Fix**: SKILL.md contains hardcoded personal paths (`/Users/majdyz/Code/AutoGPT6`, `/Users/majdyz/Code/AutoGPT9`) as example comments. These should use generic placeholder paths like `/path/to/your/worktree` so the skill is usable by anyone, not just the original author.

.claude/skills/orchestrate/scripts/poll-cycle.sh
```diff
@@ -0,0 +1,162 @@
...
+  IDLE_SINCE=$(echo "$agent"| jq -r '.idle_since // 0')
+
+  # Pass-through done/escalated agents
+  if [[ "$STATE" == "done" || "$STATE" == "escalated" ]]; then
+    UPDATED_AGENTS=$(echo "$UPDATED_AGENTS" | jq --argjson a "$agent" '. + [$a]')
+    continue
+  fi
+
+  # Classify pane
+  CLASSIFICATION=$("$CLASSIFY" "$WINDOW" "$IDLE_THRESHOLD" 2>/dev/null || \
```
majdyz: 🤖 🔴 **Blocker**: `WINDOW` (line 57) is read from JSON via `jq -r '.window'` and then passed directly as a tmux target at line 71 and line 80. If the state file is tampered with or contains a window name like `work:0; malicious_command`, this enables command injection. The window value should be validated (e.g., `[[ "$WINDOW" =~ ^[a-zA-Z0-9_.-]+:[0-9]+$ ]]`) before use in tmux commands and shell substitutions.

.claude/skills/orchestrate/scripts/classify-pane.sh
```diff
@@ -0,0 +1,82 @@
...
+  echo '{"state":"error","reason":"tmux target not found","pane_cmd":""}'
+  exit 1
+fi
+
+# Get the current foreground command in the pane
+PANE_CMD=$(tmux display-message -t "$TARGET" -p '#{pane_current_command}' 2>/dev/null || echo "unknown")
+
+# Capture and strip ANSI codes
+RAW=$(tmux capture-pane -t "$TARGET" -p 2>/dev/null || echo "")
+CLEAN=$(echo "$RAW" | sed \
```
majdyz: 🤖 🔴 **Blocker**: The `sed` patterns use `\x1b` (hex escape) syntax which is a GNU sed extension and **does not work on macOS/BSD sed**. On macOS, `sed` does not interpret `\x1b` as the ESC character. Use `$'\\033'` or pipe through `perl -pe 's/\x1b\[[0-9;]*[a-zA-Z]//g'` for cross-platform ANSI stripping. Alternatively, use `printf '\033'` to build the literal. Since the skill targets macOS worktrees, this is a runtime breakage.

.claude/skills/orchestrate/scripts/poll-cycle.sh
```diff
@@ -0,0 +1,162 @@
...
+  # Build updated agent record
+  UPDATED_AGENT=$(echo "$agent" | jq \
+    --arg state "$NEW_STATE" \
+    --arg hash "$CURRENT_HASH" \
+    --argjson now "$NOW" \
+    --argjson idle_since "$NEW_IDLE_SINCE" \
+    '.state = $state
+     | .last_output_hash = (if $hash == "" then .last_output_hash else $hash end)
+     | .last_seen_at = $now
+     | .idle_since = $idle_since')
```
majdyz: 🤖 🔴 **Blocker**: `revision_count` exists in the state schema (SKILL.md line 49) and the escalation logic in SKILL.md references "Increment revision_count on every kick" (rule 3), but `poll-cycle.sh` never reads or increments `revision_count` in the UPDATED_AGENT build (lines 129-137). The escalation safety guard is entirely absent from the script — an agent can be kicked indefinitely without ever being escalated automatically.

.claude/skills/orchestrate/scripts/classify-pane.sh
```diff
@@ -0,0 +1,82 @@
...
+TARGET="${1:-}"
+IDLE_THRESHOLD="${2:-300}"
+
+if [ -z "$TARGET" ]; then
+  echo '{"state":"error","reason":"no target provided","pane_cmd":""}' >&2
+  exit 1
+fi
+
+# Check window exists
+if ! tmux list-windows -t "${TARGET%%.*}" &>/dev/null 2>&1; then
```
majdyz: 🤖 🔴 **Blocker**: The tmux session extraction on line 24 uses `${TARGET%%.*}` which strips from the first dot, but tmux window targets use colon syntax (`session:window`), not dots. For target `work:0`, `${TARGET%%.*}` returns `work:0` unchanged (no dot present), and `tmux list-windows -t "work:0"` will then try to list windows starting from window 0 — not the session. The correct trim to get just the session name is `${TARGET%%:*}` (strip from colon). For pane targets like `work:0.1`, `${TARGET%%.*}` returns `work:0` which is still wrong — `${TARGET%%:*}` gives just `work`.

.claude/skills/orchestrate/scripts/poll-cycle.sh
```diff
@@ -0,0 +1,162 @@
...
+fi
+
+NOW=$(date +%s)
+IDLE_THRESHOLD=$(jq -r '.idle_threshold_seconds // 300' "$STATE_FILE")
+
+ACTIONS="[]"
+UPDATED_AGENTS="[]"
+
+# Read agents as newline-delimited JSON objects
+AGENTS_JSON=$(jq -c '.agents[]' "$STATE_FILE" 2>/dev/null || echo "")
```
majdyz: 🤖 🟡 **Nice to Have**: Line 47 uses `jq -c '.agents[]' "$STATE_FILE" 2>/dev/null || echo ""`, so a malformed/corrupted state file silently produces an empty agent list and the script exits with `[]` on line 51 — giving no indication the state file is broken. Consider logging a warning to stderr so the user knows the state file is corrupt rather than the orchestrator silently appearing idle.

.claude/skills/orchestrate/scripts/poll-cycle.sh
```diff
@@ -0,0 +1,194 @@
...
+fi
+
+NOW=$(date +%s)
+IDLE_THRESHOLD=$(jq -r '.idle_threshold_seconds // 300' "$STATE_FILE")
+
+ACTIONS="[]"
+UPDATED_AGENTS="[]"
+
+# Read agents as newline-delimited JSON objects
+if ! AGENTS_JSON=$(jq -c '.agents[]' "$STATE_FILE" 2>/dev/null); then
```
majdyz: 🤖 🔴 **Blocker**: The agents parse block at line 47 has a logic error: `jq -c '.agents[]'` exits with a **non-zero exit code** when `.agents` is an empty array `[]` (jq exits 5 for no matches). With the newly-added `set -e`, this now falls into the error branch and prints a spurious `"State file parse error"` message instead of correctly returning `[]`. Use `jq -c '.agents // [] | .[]'` or first check `.agents | length` before iterating, so an empty but valid agents array doesn't trigger the error path.

.claude/skills/orchestrate/scripts/classify-pane.sh
```diff
@@ -0,0 +1,82 @@
...
+
+TARGET="${1:-}"
+
+if [ -z "$TARGET" ]; then
+  echo '{"state":"error","reason":"no target provided","pane_cmd":""}'
+  exit 1
+fi
+
+# Validate tmux target format: session:window or session:window.pane
+if ! [[ "$TARGET" =~ ^[a-zA-Z0-9_.-]+:[0-9]+(\.[0-9]+)?$ ]]; then
```
majdyz: 🤖 🟠 **Should Fix**: The tmux target validation regex `^[a-zA-Z0-9_.-]+:[0-9]+(.[0-9]+)?$` only accepts **numeric** window IDs (e.g. `work:0`), but tmux also supports **named windows** (e.g. `work:editor`). Any user who uses named tmux windows will have all their agents silently skipped. Consider broadening the pattern to `^[a-zA-Z0-9_.-]+:[a-zA-Z0-9_.-]+(.[0-9]+)?$` to allow alphanumeric window names.

.claude/skills/orchestrate/scripts/poll-cycle.sh
```diff
@@ -0,0 +1,194 @@
...
+
+  # Validate window format to prevent tmux target injection
+  if ! [[ "$WINDOW" =~ ^[a-zA-Z0-9_.-]+:[0-9]+(\.[0-9]+)?$ ]]; then
+    echo "Skipping agent with invalid window value: $WINDOW" >&2
+    UPDATED_AGENTS=$(echo "$UPDATED_AGENTS" | jq --argjson a "$agent" '. + [$a]')
+    continue
+  fi
+
+  # Pass-through done/escalated agents
+  if [[ "$STATE" == "done" || "$STATE" == "escalated" ]]; then
```
majdyz: 🤖 🔴 **Blocker**: The pass-through check at line 77 only short-circuits agents in `done` or `escalated` state, but `complete` is also a terminal state (set at line 106). On the next poll cycle after an agent reaches `complete`, it will be re-classified, and since `ORCHESTRATOR:DONE` is still on screen, it will again emit a `complete` action — spamming "AGENT COMPLETE" notifications every 2 minutes indefinitely. Add `"complete"` to the pass-through guard: `if [[ "$STATE" == "done" || "$STATE" == "escalated" || "$STATE" == "complete" ]]`.

.claude/skills/orchestrate/SKILL.md
```diff
@@ -0,0 +1,316 @@
...
+      "last_output_hash": "abc123",
+      "last_seen_at": 1712345678,
+      "idle_since": 0,
+      "revision_count": 0
+    }
+  ]
+}
+```
+
+Agent states: `running` | `idle` | `stuck` | `waiting_approval` | `complete` | `done` | `escalated`
```
majdyz, line 140: 🤖 🟡 **Nice to Have**: SKILL.md lists 7 agent states (`running | idle | stuck | waiting_approval | complete | done | escalated`) but never documents the distinction between `complete` and `done`. From the code, `complete` is set by `poll-cycle.sh` when ORCHESTRATOR:DONE is seen, and `done` is set by Claude after actioning the complete notification. This is non-obvious to contributors. A one-line note like "*`complete` = script-detected; `done` = Claude-confirmed*" would prevent confusion.

.claude/skills/orchestrate/scripts/poll-cycle.sh
```diff
@@ -0,0 +1,201 @@
...
+      --arg worktree "$WORKTREE" \
+      --arg objective "$OBJECTIVE" \
+      --arg reason "$REASON" \
+      '{window:$window, action:$action, state:$state, worktree:$worktree, objective:$objective, reason:$reason}')
+    ACTIONS=$(echo "$ACTIONS" | jq --argjson a "$ACTION_OBJ" '. + [$a]')
+  fi
+
+done <<< "$AGENTS_JSON"
+
+# Atomic state file update
```
majdyz, line 251: 🤖 🟡 **Nice to Have**: The atomic state file write at line 195 creates `${STATE_FILE}.tmp` but if the `jq` command fails (e.g. due to corrupted `UPDATED_AGENTS` accumulated in the loop), the script aborts via `set -e` and leaves a stale `.tmp` file. Consider using a `trap` to clean up the temp file on failure: `trap 'rm -f "${STATE_FILE}.tmp"' EXIT` at the top of the script.

.claude/skills/orchestrate/scripts/poll-cycle.sh
```diff
@@ -0,0 +1,201 @@
...
+  echo "[]"
+  exit 0
+fi
+
+if [ -z "$AGENTS_JSON" ]; then
+  echo "[]"
+  exit 0
+fi
+
+while IFS= read -r agent; do
```
majdyz, line 74: 🤖 🟠 **Should Fix**: With `set -e` now active, any `jq` failure inside the `while` loop (e.g. a single malformed agent record in the state file) will abort the entire script before the state file update at line 195 executes — silently leaving ALL agents unprocessed with no output written. Wrap the per-agent processing in a subshell or use `|| true` guards on the jq field extractions so a bad individual record is skipped rather than killing the entire poll cycle. At minimum, document this risk in a comment.

.claude/skills/orchestrate/SKILL.md
```diff
@@ -0,0 +1,318 @@
...
+- To stop monitoring: run `/orchestrate stop`
+- The loop auto-expires after 7 days per CronCreate limits
+
+## Subcommand: stop
+
+```bash
+# Mark inactive
+jq '.active = false' ~/.claude/orchestrator-state.json > /tmp/orch.tmp \
+  && mv /tmp/orch.tmp ~/.claude/orchestrator-state.json
+
```
majdyz, line 275: 🤖 🔵 **Nit**: SKILL.md line 150 comments `# Show current state before stopping` but the code shows state AFTER marking inactive (the state file has already been updated to `active: false` at this point). Either swap the order (show first, then mark inactive) or fix the comment to say "after stopping" to avoid confusion.

majdyz:
Addressed all open review threads in commits `db37adf` and `0b3a136`:

- **`set -e` in poll-cycle.sh** (🔴): already `set -euo pipefail` on line 14
- **Hardcoded paths** (🟠): replaced with `/path/to/worktrees/` placeholders
- **WINDOW injection** (🔴): validation regex added before tmux calls
- **macOS sed / `\x1b`** (🔴): switched to `perl -pe` for ANSI stripping
- **`revision_count` escalation** (🔴): incremented on every kick (idle and stuck)
- **`%%.*` → `%%:*`** (🔴): session name extraction fixed for colon-based targets
- **Silent corrupt state** (🟡): logs warning to stderr
- **Empty agents + `set -e`** (🔴): uses `jq -e -c '.agents // empty | .[]'`
- **Named window regex** (🟠): broadened to `[a-zA-Z0-9_.-]+` in both scripts
- **`complete` pass-through** (🔴): added to terminal state guard
- **Trap cleanup** (🟡): `trap 'rm -f "${STATE_FILE}.tmp"' EXIT` added
- **Per-agent jq failure** (🟠): wrapped with `|| true` guard
- **SKILL.md stop comment** (🔵): fixed to "after marking inactive" in 0b3a136ae0
- **complete/done distinction** (🟡): state transition note added in 0b3a136ae0

CI: all green ✅

.claude/skills/orchestrate/scripts/poll-cycle.sh
majdyz: 🤖 Fixed in `30d785400d`: added `set -euo pipefail` to poll-cycle.sh. Per-agent jq failures are now isolated with `|| { ... continue; }` so a single malformed agent doesn't abort the whole cycle.

.claude/skills/orchestrate/SKILL.md
majdyz: 🤖 Fixed in `2b68920be1`: removed all hardcoded paths. SKILL.md now uses generic `/path/to/worktrees/AutoGPT6` in examples and runtime variables everywhere else.

.claude/skills/orchestrate/scripts/poll-cycle.sh
majdyz: 🤖 Fixed in `30d785400d`: added window format validation regex `^[a-zA-Z0-9_.-]+:[a-zA-Z0-9_.-]+(\.[0-9]+)?$` in poll-cycle.sh. Agents with invalid window values are skipped with a logged warning.

.claude/skills/orchestrate/scripts/classify-pane.sh
majdyz: 🤖 Fixed in `35a645d9c7`: switched from `sed` to `perl -pe 's/\x1b\[[0-9;]*[a-zA-Z]//g; s/\x1b\(B//g; s/\x1b\[\?[0-9]*[hl]//g; s/\r//g'` in classify-pane.sh. Perl's `\x1b` is cross-platform (works on BSD/macOS and Linux).

.claude/skills/orchestrate/scripts/poll-cycle.sh
majdyz: 🤖 Fixed in `30d785400d`: `revision_count` is now incremented in both the `idle` and `stuck` case branches, and the escalation check is applied after the increment.

.claude/skills/orchestrate/scripts/classify-pane.sh
majdyz: 🤖 Fixed in `35a645d9c7`: changed to `${TARGET%%:*}` (strips from first colon) instead of `${TARGET%%.*}` in classify-pane.sh.

.claude/skills/orchestrate/scripts/poll-cycle.sh
majdyz: 🤖 Fixed in `35a645d9c7`: replaced with `if ! AGENTS_JSON=$(jq -e -c '.agents // empty | .[]' ...)` pattern that handles the empty-array exit code gracefully, with a separate JSON validity check.

.claude/skills/orchestrate/scripts/poll-cycle.sh
majdyz: 🤖 Fixed in `35a645d9c7`: same fix as above — the `jq -e -c '.agents // empty | .[]'` pattern with `if !` guard correctly handles empty arrays without triggering `set -e`.

.claude/skills/orchestrate/scripts/classify-pane.sh
majdyz: 🤖 Fixed in `35a645d9c7`: regex expanded to `^[a-zA-Z0-9_.-]+:[a-zA-Z0-9_.-]+(\.[0-9]+)?$` in both classify-pane.sh and poll-cycle.sh — accepts named windows (e.g. `autogpt1:AutoGPT6`) in addition to numeric ones.

.claude/skills/orchestrate/scripts/poll-cycle.sh
majdyz: 🤖 Fixed in `35a645d9c7`: added `complete` to the terminal-state pass-through guard: `if [[ "$STATE" == "done" || "$STATE" == "escalated" || "$STATE" == "complete" ]]`. Agents in `complete` state are now skipped until Claude acts on them.

.claude/skills/orchestrate/SKILL.md
majdyz: 🤖 Fixed in `0b3a136ae0`: SKILL.md now documents all 7 states (`running | idle | stuck | waiting_approval | complete | done | escalated`) with a note explaining the `complete` → `done` transition: `complete` = set by poll-cycle.sh when ORCHESTRATOR:DONE detected; `done` = set by Claude after recycling the worktree.

.claude/skills/orchestrate/scripts/poll-cycle.sh
majdyz: 🤖 Fixed in `8812bd669e`: added `trap 'rm -f "${STATE_FILE}.tmp"' EXIT` near the top of poll-cycle.sh so the .tmp file is cleaned up on any exit path.

.claude/skills/orchestrate/scripts/poll-cycle.sh
majdyz: 🤖 Fixed in `30d785400d`: each per-agent jq call is now guarded with `|| { echo "Warning: ..."; ... continue; }` so a single malformed field skips that agent and continues the loop rather than aborting with `set -e`.

.claude/skills/orchestrate/SKILL.md
majdyz: 🤖 Fixed in `0b3a136ae0`: reordered the `stop` subcommand to mark inactive first, then show state — comment now correctly reads `# Show state (after marking inactive)`.

.claude/skills/orchestrate/scripts/classify-pane.sh
```diff
@@ -0,0 +1,82 @@
...
+IDLE_THRESHOLD="${2:-300}"
+
+if [ -z "$TARGET" ]; then
+  echo '{"state":"error","reason":"no target provided","pane_cmd":""}' >&2
+  exit 1
+fi
+
+# Check window exists
+if ! tmux list-windows -t "${TARGET%%.*}" &>/dev/null 2>&1; then
+  echo '{"state":"error","reason":"tmux target not found","pane_cmd":""}'
```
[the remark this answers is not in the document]
majdyz: 🤖 Confirmed resolved — the fix was carried forward into the v2 rewrite. The current classify-pane.sh uses `perl` for ANSI stripping and `${TARGET%%:*}` for session extraction.

.claude/skills/orchestrate/scripts/poll-cycle.sh
```diff
@@ -0,0 +1,162 @@
...
+      # Check if hash has been stable (agent may be stuck mid-task)
+      if [ -n "$CURRENT_HASH" ] && [ "$CURRENT_HASH" = "$LAST_HASH" ] && [ "$LAST_HASH" != "" ]; then
+        if [ "$IDLE_SINCE" = "0" ] || [ "$IDLE_SINCE" = "null" ]; then
+          NEW_IDLE_SINCE=$NOW
+        else
+          STUCK_DURATION=$(( NOW - IDLE_SINCE ))
+          if [ "$STUCK_DURATION" -gt "$IDLE_THRESHOLD" ]; then
+            NEW_STATE="stuck"
+            ACTION="kick"
+            REASON="output unchanged for ${STUCK_DURATION}s (threshold: ${IDLE_THRESHOLD}s)"
```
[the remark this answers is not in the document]
majdyz: 🤖 Confirmed resolved in the v2 rewrite — window validation, `set -euo pipefail`, and isolated per-agent error handling are all present in the current poll-cycle.sh.

.claude/skills/orchestrate/scripts/poll-cycle.sh
```diff
@@ -0,0 +1,162 @@
...
+
+  PANE_STATE=$(echo "$CLASSIFICATION" | jq -r '.state')
+  PANE_REASON=$(echo "$CLASSIFICATION" | jq -r '.reason')
+
+  # Compute content hash for stuck-detection (only for running agents)
+  CURRENT_HASH=""
+  if [[ "$PANE_STATE" == "running" ]]; then
+    RAW=$(tmux capture-pane -t "$WINDOW" -p 2>/dev/null || echo "")
+    CURRENT_HASH=$(echo "$RAW" | tail -20 | md5_hash)
+  fi
```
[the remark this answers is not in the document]
majdyz: 🤖 Confirmed resolved — same as above.

.claude/skills/orchestrate/scripts/poll-cycle.sh
```diff
@@ -0,0 +1,209 @@
...
+md5_hash() {
+  if command -v md5sum &>/dev/null; then
+    md5sum | awk '{print $1}'
+  else
+    md5 | awk '{print $NF}'
+  fi
+}
+
+# Clean up temp file on any exit (avoids stale .tmp if jq write fails)
+trap 'rm -f "${STATE_FILE}.tmp"' EXIT
```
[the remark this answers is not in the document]
majdyz: 🤖 Acknowledged. The poll-cycle.sh is invoked by a CronCreate loop that fires every ~2 minutes — race conditions with `add`/`stop` are possible but low-probability in practice. Adding `flock` would fully address this; noting it as a future improvement. The current design documents atomic writes via `.tmp` + `mv` as the guard.

.claude/skills/orchestrate/scripts/poll-cycle.sh
```diff
@@ -0,0 +1,209 @@
...
+# Read agents as newline-delimited JSON objects.
+# jq exits non-zero when .agents[] has no matches on an empty array, which is valid —
+# so we suppress that exit code and separately validate the file is well-formed JSON.
+if ! AGENTS_JSON=$(jq -e -c '.agents // empty | .[]' "$STATE_FILE" 2>/dev/null); then
+  if ! jq -e '.' "$STATE_FILE" > /dev/null 2>&1; then
+    echo "State file parse error — check $STATE_FILE" >&2
+  fi
+  echo "[]"
+  exit 0
+fi
```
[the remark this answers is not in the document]
majdyz: 🤖 See reply above on the serialize state-file writes thread.

.claude/skills/orchestrate/scripts/poll-cycle.sh
```diff
@@ -0,0 +1,209 @@
...
+        fi
+      else
+        # Output changed — reset idle timer
+        NEW_STATE="running"
+        NEW_IDLE_SINCE=0
+      fi
+      ;;
+    error)
+      REASON="classify error: $PANE_REASON"
+      ;;
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in `30d785400d`: classifier failures now fall through to a logged `REASON="classify error: $PANE_REASON"` and the agent state is preserved (no silent drop). The action stays `none` so Claude sees the error on the next poll.

.claude/skills/orchestrate/SKILL.md
```diff
@@ -0,0 +1,404 @@
...
+```
+spare/N branch   →   orchestrate add   →   new window + feat/branch + claude running
+                                                        ↓
+                                               ORCHESTRATOR:DONE
+                                                        ↓
+                                        kill window + git checkout spare/N
+                                                        ↓
+                                               spare/N (free again)
+```
```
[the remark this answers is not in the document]
majdyz: 🤖 Acknowledged. The fenced blocks in SKILL.md are pseudocode illustrating the skill's logic for Claude Code to execute inline — not runnable scripts. Adding ```bash` identifiers would trigger syntax linting on constructs like function definitions used as conceptual helpers. Leaving them as plain fenced blocks is intentional for this skill format.

.claude/skills/orchestrate/SKILL.md
```diff
@@ -0,0 +1,404 @@
...
+Orchestrator poll cycle.
+
+Run: bash SKILLS_DIR/scripts/poll-cycle.sh
+
+Read the JSON output. For each action:
+
+- "kick" (idle — shell foreground): get worktree_path + spare_branch from state file.
+  Run: tmux send-keys -t <window> "cd <worktree_path> && claude --permission-mode bypassPermissions" Enter
+  Wait 3s. If pane shows "Enter to confirm", send Down+Enter first.
+  Then send: "<objective>. When done output ORCHESTRATOR:DONE" Enter
```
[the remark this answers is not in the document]
majdyz: 🤖 See reply above on the fenced block language identifiers thread.

majdyz:
## 🧪 E2E Test Report — PR #12691 (Orchestrate Skill)

**Branch:** `dx/orchestrate-skill`  
**Date:** 2026-04-07  
**Type:** Functional script tests (skills-only PR — no Docker/browser applicable)

---

### Test Plan

This PR adds `.claude/skills/orchestrate/` (SKILL.md + two bash scripts). Since no production app code changed, testing focused on the scripts directly and on integration with a live tmux session.

### Test Results

| # | Scenario | Result |
|---|---|---|
| 1 | `classify-pane.sh` — validates tmux target format (invalid input) | ✅ PASS |
| 2 | `classify-pane.sh` — returns `idle` when shell is foreground | ✅ PASS |
| 3 | `poll-cycle.sh` — returns `[]` when `active: false` in state file | ✅ PASS |
| 4 | `poll-cycle.sh` — handles empty `agents: []` array without crashing | ✅ PASS |
| 5 | `poll-cycle.sh` — spare worktree discovery via `git worktree list --porcelain` | ✅ PASS |
| 6 | Real agent spawn in `autogpt1` tmux session | ✅ PASS (agent ran, output `ORCHESTRATOR:DONE`, recycle mechanics verified) |
| 7 | Stuck detection hash-based cycle (3-poll escalation) | ✅ PASS (logic traced through code) |

**7/7 scenarios verified.**

### Key Fixes Validated

- `set -euo pipefail` in poll-cycle.sh with per-agent `|| continue` isolation
- `perl`-based ANSI stripping in classify-pane.sh (replaces non-portable `sed \x1b`)  
- Named window regex `[a-zA-Z0-9_.-]+:[a-zA-Z0-9_.-]+` in both scripts
- `complete` added to terminal-state pass-through guard
- `revision_count` incremented on each kick; escalation fires at ≥3
- `trap 'rm -f "${STATE_FILE}.tmp"' EXIT` for temp file cleanup
- `${TARGET%%:*}` (was `%%.*`) for correct tmux session extraction

### Notes

No screenshots applicable — this PR contains only skill markdown and bash scripts.  
All 22 review threads have been replied to inline.  
CI: all checks passing (CodeQL, size, scope, CodeRabbit, Vercel, snyk).

majdyz (commented):
## 🧪 Step 8 Test Evaluation

### Per-scenario checklist

| Criterion | Status |
|---|---|
| All planned scenarios tested | ✅ 7/7 |
| Scripts execute without errors | ✅ classify-pane.sh, poll-cycle.sh both verified |
| Negative tests included | ✅ invalid tmux target, inactive state, empty agents array |
| All 22 review threads addressed | ✅ replies posted with fixing commit SHAs |
| CI fully green | ✅ CodeQL, snyk, scope, size, CodeRabbit, Vercel all passing |
| Real end-to-end integration | ✅ Live agent spawn in `autogpt1` tmux session — agent ran, output `ORCHESTRATOR:DONE`, recycle verified |

### Verdict: ready to merge

All blockers addressed in `35a645d9c7`, `8812bd669e`, `30d785400d`, `0b3a136ae0`. The skill is portable (perl over sed, cross-platform md5), injection-safe (tmux target validation, jq-built JSON), and handles edge cases (empty array, named windows, complete state pass-through, escalation counter). No regressions. No screenshots needed — skills-only PR.

.claude/skills/orchestrate/SKILL.md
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [642aa1ce34](https://github.com/Significant-Gravitas/AutoGPT/commit/642aa1ce3481292871051efe33e9ee079b3b9e67): added `text` to lifecycle diagram, example output, and CronCreate prompt blocks. `bash` was already on all shell blocks. (Apologies for the previous push-back — `text` is the right call here.)

.claude/skills/orchestrate/SKILL.md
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [642aa1ce34](https://github.com/Significant-Gravitas/AutoGPT/commit/642aa1ce3481292871051efe33e9ee079b3b9e67) — same commit as above.

.claude/skills/orchestrate/scripts/poll-cycle.sh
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [730ddd4382](https://github.com/Significant-Gravitas/AutoGPT/commit/730ddd4382d676058e5a757e51a30383d36a2372): added an upfront `jq -e '.'` guard right after the `ensure state file exists` block, before the `ACTIVE=$(jq ...)` read. A corrupt/truncated file now emits a stderr message + `[]` and exits cleanly. (Apologies — my previous reply incorrectly conflated this with the serialize thread; these are distinct issues.)

.claude/skills/orchestrate/scripts/poll-cycle.sh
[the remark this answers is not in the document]
majdyz: 🤖 Yes, please open a follow-up issue to track the `flock`-based serialization improvement — that would be the right place to address the concurrent writer race. The upfront JSON validation (for corrupt/truncated files) has been fixed separately in [730ddd4382](https://github.com/Significant-Gravitas/AutoGPT/commit/730ddd4382d676058e5a757e51a30383d36a2372).

.claude/skills/orchestrate/scripts/poll-cycle.sh
```diff
@@ -0,0 +1,218 @@
...
+    idle)
+      # Agent process has exited — needs restart
+      NEW_STATE="idle"
+      ACTION="kick"
+      REASON="agent exited (shell is foreground)"
+      NEW_REVISION_COUNT=$(( REVISION_COUNT + 1 ))
+      NEW_IDLE_SINCE=$NOW
+      if [ "$NEW_REVISION_COUNT" -ge 3 ]; then
+        NEW_STATE="escalated"
+        ACTION="none"
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [602aede0b6](https://github.com/Significant-Gravitas/AutoGPT/commit/602aede0b640f6901f48f4aea291be4ca760318b): added a guard at the start of the `running` case — when the previous state was `idle` or `stuck`, `NEW_IDLE_SINCE` is reset to 0 so the stuck-detection timer starts fresh after a kick-and-restart cycle.

.claude/skills/pr-test/SKILL.md
```diff
@@ -595,25 +582,12 @@ for img in "${SCREENSHOT_FILES[@]}"; do
...
-    --jq '.sha')
-fi
-
+COMMIT_SHA=$(gh api "repos/${REPO}/git/commits" \
+  -f message="test: add E2E test screenshots for PR #${PR_NUMBER}" \
+  -f tree="$TREE_SHA" \
+  --jq '.sha')
 gh api "repos/${REPO}/git/refs" \
   -f ref="refs/heads/${SCREENSHOTS_BRANCH}" \
   -f sha="$COMMIT_SHA" 2>/dev/null \
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [602aede0b6](https://github.com/Significant-Gravitas/AutoGPT/commit/602aede0b640f6901f48f4aea291be4ca760318b): (1) resolved parent SHA before creating the commit — reads existing branch HEAD and passes it as `parents[]` so screenshots are chained; falls back to root commit only on first push. (2) changed `-f force=true` → `-F force=true` (uppercase F = boolean) on the PATCH ref call.

.claude/skills/orchestrate/scripts/poll-cycle.sh
```diff
@@ -0,0 +1,223 @@
...
+  # Pass-through terminal-state agents
+  if [[ "$STATE" == "done" || "$STATE" == "escalated" || "$STATE" == "complete" ]]; then
+    UPDATED_AGENTS=$(echo "$UPDATED_AGENTS" | jq --argjson a "$agent" '. + [$a]')
+    continue
+  fi
+
+  # Classify pane
+  CLASSIFICATION=$("$CLASSIFY" "$WINDOW" 2>/dev/null || \
+    echo '{"state":"error","reason":"classify failed","pane_cmd":"unknown"}')
+
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [8f44e9fd86](https://github.com/Significant-Gravitas/AutoGPT/commit/8f44e9fd8602715de0c0001d119de764bc32a621): removed the `|| echo '...'` fallback. Since `classify-pane.sh` always emits a JSON error object before exit, capturing its output directly avoids the two-object concatenation. The fallback now only triggers if `CLASSIFICATION` is unexpectedly empty.

.claude/skills/orchestrate/scripts/poll-cycle.sh
```diff
@@ -0,0 +1,223 @@
...
+        ACTION="none"
+        REASON="escalated after ${NEW_REVISION_COUNT} kicks — needs human attention"
+      fi
+      ;;
+    running)
+      # Clear idle_since carried over from a previous idle/stuck state so the
+      # stuck-detection timer starts fresh after a kick-and-restart cycle.
+      if [[ "$STATE" == "idle" || "$STATE" == "stuck" ]]; then
+        NEW_IDLE_SINCE=0
+      fi
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [8f44e9fd86](https://github.com/Significant-Gravitas/AutoGPT/commit/8f44e9fd8602715de0c0001d119de764bc32a621): narrowed the guard to only reset `NEW_IDLE_SINCE=0` when `STATE == idle` (kick-and-restart transition). Stuck agents no longer get their timer reset, so `STUCK_DURATION` accumulates correctly across polls and escalation can trigger.

.claude/skills/orchestrate/scripts/poll-cycle.sh
```diff
@@ -0,0 +1,227 @@
...
+  if [[ "$STATE" == "done" || "$STATE" == "escalated" || "$STATE" == "complete" ]]; then
+    UPDATED_AGENTS=$(echo "$UPDATED_AGENTS" | jq --argjson a "$agent" '. + [$a]')
+    continue
+  fi
+
+  # Classify pane.
+  # classify-pane.sh always emits JSON before exit (even on error), so using
+  # "|| echo '...'" would concatenate two JSON objects when it exits non-zero.
+  # Capture output directly; fall back only if output is unexpectedly empty.
+  CLASSIFICATION=$("$CLASSIFY" "$WINDOW" 2>/dev/null)
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [e35e3e9e1b](https://github.com/Significant-Gravitas/AutoGPT/commit/e35e3e9e1b7e3e7783afce402044761c729dd47a): changed to $("$CLASSIFY" ... 2>/dev/null || true) — the || true inside the substitution prevents set -euo pipefail from aborting when classify exits non-zero. classify's JSON output is still captured correctly.

.claude/skills/orchestrate/scripts/poll-cycle.sh
```diff
@@ -0,0 +1,227 @@
...
+              ACTION="kick"
+              REASON="output unchanged for ${STUCK_DURATION}s (threshold: ${IDLE_THRESHOLD}s)"
+            fi
+          fi
+        fi
+      else
+        # Output changed — reset idle timer
+        NEW_STATE="running"
+        NEW_IDLE_SINCE=0
+      fi
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [e35e3e9e1b](https://github.com/Significant-Gravitas/AutoGPT/commit/e35e3e9e1b7e3e7783afce402044761c729dd47a): timer reset in the else branch is now guarded by [ -n "$CURRENT_HASH" ]. If tmux capture-pane fails and hash is empty, timers are preserved so stuck-detection continues accumulating across polls.

.claude/skills/orchestrate/scripts/find-spare.sh
```diff
@@ -0,0 +1,24 @@
...
+  echo "Error: not inside a git repo and no REPO_ROOT provided" >&2
+  exit 1
+fi
+
+git -C "$REPO_ROOT" worktree list --porcelain \
+  | awk '
+      /^worktree / { path = substr($0, 10) }
+      /^branch /   { branch = substr($0, 8); print path " " branch }
+    ' \
+  | grep -E " refs/heads/spare/[0-9]+$" \
```
majdyz: 🤖 🔴 **Blocker**: `grep -E` exits with code 1 when no spare worktrees exist. With `set -euo pipefail` + `pipefail`, this propagates through the pipeline and causes the script to exit 1 — making a completely normal "no spares" outcome indistinguishable from a real error. `capacity.sh` and `find-spare.sh` callers see an error exit when there's nothing wrong.

Fix: append `|| true` to the grep:
```bash
| grep -E " refs/heads/spare/[0-9]+$" || true \
| sed 's|refs/heads/||'
```

.claude/skills/orchestrate/scripts/recycle-agent.sh
```diff
@@ -0,0 +1,30 @@
...
+
+WINDOW="$1"
+WORKTREE_PATH="$2"
+SPARE_BRANCH="$3"
+
+# Kill the tmux window (ignore error — may already be gone)
+tmux kill-window -t "$WINDOW" 2>/dev/null || true
+
+# Restore to spare branch: clean tracked then untracked files
+git -C "$WORKTREE_PATH" reset --hard HEAD 2>/dev/null
```
majdyz, line 28: 🤖 🔴 **Blocker**: `git reset --hard HEAD` can fail (e.g. if the worktree has a rebase/merge in progress, or HEAD doesn't exist on a freshly orphaned branch). With `set -euo pipefail`, failure aborts the script immediately after the window has already been killed — the worktree is left on the task branch with dirty state, and the spare slot is never restored.

Fix: add `|| true` so the clean + checkout steps always run:
```bash
git -C "$WORKTREE_PATH" reset --hard HEAD 2>/dev/null || true
```

.claude/skills/orchestrate/scripts/spawn-agent.sh
```diff
@@ -0,0 +1,60 @@
...
+if [[ "$CMD" != "node" ]]; then
+  PANE=$(tmux capture-pane -t "$WINDOW" -p 2>/dev/null || echo "")
+  if echo "$PANE" | grep -q "Enter to confirm"; then
+    tmux send-keys -t "$WINDOW" Down Enter
+    sleep 2
+  fi
+fi
+
+# Send the task — agent must output ORCHESTRATOR:DONE when finished
+tmux send-keys -t "$WINDOW" "${OBJECTIVE}. When all work is done, output the exact string: ORCHESTRATOR:DONE" Enter
```
majdyz: 🤖 🔴 **Blocker**: The objective is sent to the pane unconditionally regardless of whether claude actually started. If claude fails to start and the shell is the foreground process, `${OBJECTIVE}. When all work is done...` is executed literally as a shell command. Most objectives (e.g. "Implement feature X...") will just error out harmlessly, but objectives containing shell metacharacters (`;`, `$()`, `&&`, etc.) could execute arbitrary commands in the worktree shell.

Fix: exit with an error if claude did not start, so the caller can surface the failure rather than silently proceeding:
```bash
CMD=$(tmux display-message -t "$WINDOW" -p '#{pane_current_command}' 2>/dev/null || echo "")
if [[ "$CMD" != "node" ]]; then
  echo "Error: claude did not start in $WINDOW (foreground: $CMD)" >&2
  exit 1
fi
tmux send-keys -t "$WINDOW" "${OBJECTIVE}. When all work is done, output the exact string: ORCHESTRATOR:DONE" Enter
```

ntindle (dismissed):

.claude/skills/orchestrate/scripts/find-spare.sh
```diff
@@ -0,0 +1,24 @@
...
+  | sed 's|refs/heads/||'
```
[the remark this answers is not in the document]
majdyz: Fixed in [504a5c7](https://github.com/Significant-Gravitas/AutoGPT/commit/504a5c77240181a08c5d74351fcc6e84e2fbbaaa): wrapped grep in `{ grep -E ... || true; }` so empty results don't exit non-zero under `set -euo pipefail`.

.claude/skills/orchestrate/scripts/recycle-agent.sh
```diff
@@ -0,0 +1,30 @@
...
+git -C "$WORKTREE_PATH" clean -fd 2>/dev/null
+git -C "$WORKTREE_PATH" checkout "$SPARE_BRANCH"
```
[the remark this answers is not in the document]
majdyz: Fixed in [504a5c7](https://github.com/Significant-Gravitas/AutoGPT/commit/504a5c77240181a08c5d74351fcc6e84e2fbbaaa): added `git rebase --abort 2>/dev/null || true` and `git merge --abort 2>/dev/null || true` before `git reset --hard HEAD`.

.claude/skills/orchestrate/scripts/find-spare.sh
majdyz: Fixed in [504a5c7](https://github.com/Significant-Gravitas/AutoGPT/commit/504a5c77240181a08c5d74351fcc6e84e2fbbaaa): wrapped grep in `{ grep -E ... || true; }` to prevent exit code 1 from aborting via `set -euo pipefail`.

.claude/skills/orchestrate/scripts/recycle-agent.sh
majdyz: Fixed in [504a5c7](https://github.com/Significant-Gravitas/AutoGPT/commit/504a5c77240181a08c5d74351fcc6e84e2fbbaaa): added `git rebase --abort 2>/dev/null || true` and `git merge --abort 2>/dev/null || true` before `git reset --hard HEAD`.

.claude/skills/orchestrate/scripts/spawn-agent.sh
majdyz: Fixed in [504a5c7](https://github.com/Significant-Gravitas/AutoGPT/commit/504a5c77240181a08c5d74351fcc6e84e2fbbaaa): track `PROMPT_FOUND` flag in the wait loop; log a warning to stderr if 60s timeout fires without seeing ❯, then still send (graceful degradation).

.claude/skills/orchestrate/scripts/run-loop.sh
```diff
@@ -0,0 +1,181 @@
...
+    tmux send-keys -t "$window" "cd '${worktree_path}' && claude --permission-mode bypassPermissions" Enter
+    sleep 3
+    # Auto-dismiss settings error dialog if present
+    local pane
+    pane=$(tmux capture-pane -t "$window" -p 2>/dev/null || echo "")
+    if echo "$pane" | grep -q "Enter to confirm"; then
+      tmux send-keys -t "$window" Down Enter
+      sleep 2
+    fi
+    tmux send-keys -t "$window" "${objective}. When all work is done, output the exact string: ORCHESTRATOR:DONE" Enter
```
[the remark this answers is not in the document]
majdyz: Fixed in [504a5c7](https://github.com/Significant-Gravitas/AutoGPT/commit/504a5c77240181a08c5d74351fcc6e84e2fbbaaa): replaced `sleep 3` with a 60s wait loop (mirrors spawn-agent.sh) that waits for `pane_current_command==node` AND ❯ before sending the objective. Logs a warning if timeout fires.

.claude/skills/orchestrate/scripts/run-loop.sh
```diff
@@ -0,0 +1,145 @@
...
+    tmux send-keys -t "$window" "" Enter
+    return
+  fi
+
+  # y/n prompt for safe operations
+  if echo "$pane_tail" | grep -qiE "(^git |^npm |^pnpm |^poetry |^pytest|^docker |^make |^cargo |^pip |^yarn |curl .*(localhost|127\.0\.0\.1))"; then
+    echo "[$(date +%H:%M:%S)] APPROVE safe   $window"
+    tmux send-keys -t "$window" "y" Enter
+    return
+  fi
```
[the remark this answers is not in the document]
majdyz: Fixed in [504a5c7](https://github.com/Significant-Gravitas/AutoGPT/commit/504a5c77240181a08c5d74351fcc6e84e2fbbaaa): changed `tail -10` to `tail -3` so the safety check only sees the current prompt, not command history.

majdyz (commented):
🤖 **Fresh review — 2026-04-07** — 3 open bugs remain unresolved:

**1. spawn-agent.sh** () 🔴 **Blocker**: `pr_number`/`steps` update runs *before* the agent record is appended to `.agents[]`. The `jq select(.window==$w)` matches nothing, so the update is silently skipped and subsequent `verify-complete.sh` calls never find required checkpoints. Fix: append the initial agent record first, then update `pr_number`/`steps`.

**2. run-loop.sh** (`check_supervisor`, line 191) 🔴 **Blocker**: Supervisor restart uses `sleep 5` then unconditionally sends the recovery prompt — regardless of whether Claude started. If Claude takes longer than 5s to initialize, the prompt lands at a shell or mid-startup dialog. Fix: use the same wait-loop pattern as `spawn-agent.sh` (poll `pane_current_command==node` + `❯` present, 60s timeout).

**3. classify-pane.sh** (line 37) 🔴 **Blocker**: `tmux capture-pane -t "$TARGET" -p` captures only the *visible* pane — no scrollback. If `ORCHESTRATOR:DONE` scrolled off screen, it is never detected and the agent stays in `running` state forever. Fix: add `-S -50` to include 50 lines of scrollback.

.claude/skills/orchestrate/scripts/spawn-agent.sh
```diff
@@ -0,0 +1,84 @@
...
+WIN_IDX=$(tmux new-window -t "$SESSION" -n "$WORKTREE_NAME" -P -F '#{window_index}')
+WINDOW="${SESSION}:${WIN_IDX}"
+
+# Store pr_number + steps in state file if provided (enables verify-complete.sh)
+if [ -n "$PR_NUMBER" ] && [ -f "$STATE_FILE" ]; then
+  STEPS_JSON=$(printf '%s\n' "${STEPS[@]}" | jq -R . | jq -s .)
+  jq --arg w "$WINDOW" --arg pr "$PR_NUMBER" --argjson steps "$STEPS_JSON" \
+    '.agents |= map(if .window == $w then . + {pr_number: $pr, steps: $steps, checkpoints: []} else . end)' \
+    "$STATE_FILE" > "${STATE_FILE}.tmp" && mv "${STATE_FILE}.tmp" "$STATE_FILE"
+fi
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [e77909be06](https://github.com/Significant-Gravitas/AutoGPT/commit/e77909be062f04c49567464a0c7188063e8f819b): added initial agent record append to spawn-agent.sh (with window, worktree_path, branch, objective, state=running, etc.) before the pr_number/steps update block. The jq `select(.window==$w)` now finds the record and the pr_number/steps fields are correctly written. Updated SKILL.md to note that spawn-agent.sh handles the append — callers must not append separately.

.claude/skills/orchestrate/scripts/run-loop.sh
```diff
@@ -0,0 +1,269 @@
...
+      [[ -z "$work_dir" ]] && work_dir="$HOME"
+
+      tmux send-keys -t "$sup_win" "cd '${work_dir}' && claude --permission-mode bypassPermissions" Enter
+      sleep 5
+
+      # Auto-dismiss settings dialog if it appears
+      local pane
+      pane=$(tmux capture-pane -t "$sup_win" -p 2>/dev/null || echo "")
+      if echo "$pane" | grep -q "Enter to confirm"; then
+        tmux send-keys -t "$sup_win" Down Enter
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [e77909be06](https://github.com/Significant-Gravitas/AutoGPT/commit/e77909be062f04c49567464a0c7188063e8f819b): replaced `sleep 5` + one-shot pane check in `check_supervisor` with a 60s wait loop (same pattern as `spawn-agent.sh`/`handle_kick`) that polls `pane_current_command==node` AND `❯` prompt visible, with settings-dialog dismiss inline. Recovery prompt is only sent after Claude is confirmed interactive (or after timeout with a warning).

.claude/skills/orchestrate/scripts/classify-pane.sh
```diff
@@ -0,0 +1,85 @@
...
+if ! tmux list-windows -t "${TARGET%%:*}" &>/dev/null 2>&1; then
+  echo '{"state":"error","reason":"tmux target not found","pane_cmd":""}'
+  exit 1
+fi
+
+# Get the current foreground command in the pane
+PANE_CMD=$(tmux display-message -t "$TARGET" -p '#{pane_current_command}' 2>/dev/null || echo "unknown")
+
+# Capture and strip ANSI codes (use perl for cross-platform compatibility — BSD sed lacks \x1b support)
+RAW=$(tmux capture-pane -t "$TARGET" -p 2>/dev/null || echo "")
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [e77909be06](https://github.com/Significant-Gravitas/AutoGPT/commit/e77909be062f04c49567464a0c7188063e8f819b): changed `tmux capture-pane -t "$TARGET" -p` to `tmux capture-pane -t "$TARGET" -p -S -50` in classify-pane.sh. The `-S -50` flag includes 50 lines of scrollback so `ORCHESTRATOR:DONE` is detected even after scrolling off the visible pane.

.claude/skills/orchestrate/scripts/verify-complete.sh
```diff
@@ -0,0 +1,88 @@
...
+if [ -z "$REPO" ] && [ -n "$WORKTREE_PATH" ] && [ -d "$WORKTREE_PATH" ]; then
+  REPO=$(git -C "$WORKTREE_PATH" remote get-url origin 2>/dev/null \
+    | sed 's|.*github\.com[:/]||; s|\.git$||' || echo "")
+fi
+
+# --- Check 1: all required steps are checkpointed ---
+MISSING=""
+while IFS= read -r step; do
+  [ -z "$step" ] && continue
+  if ! echo "$CHECKPOINTS" | grep -qF "$step"; then
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [7c6fbfae7b](https://github.com/Significant-Gravitas/AutoGPT/commit/7c6fbfae7b5caac13a5b6f509aaef41d4f48b10c): changed `grep -qF` to `grep -qFx` in `verify-complete.sh` line 44. The `-x` flag anchors the match to the full line, so a required step named `pr` can no longer be satisfied by a checkpoint named `pr-test`.

.claude/skills/orchestrate/scripts/spawn-agent.sh
```diff
@@ -0,0 +1,117 @@
...
+       "idle_since": 0,
+       "revision_count": 0,
+       "last_rebriefed_at": 0
+     }]' "$STATE_FILE" > "${STATE_FILE}.tmp" && mv "${STATE_FILE}.tmp" "$STATE_FILE"
+fi
+
+# Store pr_number + steps in state file if provided (enables verify-complete.sh).
+# The agent record was appended above so the jq select now finds it.
+if [ -n "$PR_NUMBER" ] && [ -f "$STATE_FILE" ]; then
+  STEPS_JSON=$(printf '%s\n' "${STEPS[@]}" | jq -R . | jq -s .)
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [7c6fbfae7b](https://github.com/Significant-Gravitas/AutoGPT/commit/7c6fbfae7b5caac13a5b6f509aaef41d4f48b10c): `spawn-agent.sh` now checks `${#STEPS[@]} -gt 0` before running `printf '%s\n' "${STEPS[@]}" | jq -R . | jq -s .`, falling back to `STEPS_JSON='[]'` when no steps are provided. This prevents the `[""]→pass-immediately` false-complete.

PR #12706, opened 2026-04-08
fix(backend/executor): parse RPC user dict into User model before accessing .timezone

majdyz:
## Why

Every scheduled graph execution in the dev-behave environment was failing with:

```
AttributeError: 'dict' object has no attribute 'timezone'
```

Root cause: in `add_graph_execution`, when `prisma.is_connected()` is `False` (which is always the case in the scheduler/executor process), `udb` is set to `DatabaseManagerAsyncClient`. Its `get_user_by_id` method goes through the RPC layer and returns `response.json()` — a plain `dict` — not a typed `User` model. The code then does `user.timezone`, which fails on a dict.

Sentry issue: https://significant-gravitas.sentry.io/issues/7395473462/

## What

- Wrap the `get_user_by_id` result with `User.model_validate()` in `executor/utils.py`, which accepts both a raw dict (from the RPC path) and an existing `User` instance (from the Prisma path).
- Add a regression test that simulates the exact failure path: `prisma.is_connected() = False`, `get_user_by_id` returns a raw dict, and verifies execution context is built without `AttributeError`.

## How

`User.model_validate(x)` is Pydantic v2's standard way to construct a model from either a dict or an existing model instance. This is the minimal fix at the usage site.

### DIFF
--- a/autogpt_platform/backend/backend/executor/utils.py
+++ b/autogpt_platform/backend/backend/executor/utils.py
@@ -31,7 +31,12 @@ from backend.data.execution import (
     NodesInputMasks,
 )
 from backend.data.graph import GraphModel, Node
-from backend.data.model import USER_TIMEZONE_NOT_SET, CredentialsMetaInput, GraphInput
+from backend.data.model import (
+    USER_TIMEZONE_NOT_SET,
+    CredentialsMetaInput,
+    GraphInput,
+    User,
+)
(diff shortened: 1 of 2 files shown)


majdyz:
Superseded by a proper fix that removes the silent fallback in _get_return (service.py) so RPC type mismatches fail loudly instead of silently returning raw dicts. New PR incoming.

PR #12707, opened 2026-04-08
fix(backend): make User RPC model forward-compatible during rolling deploys

majdyz:
## Why

A Sentry `AttributeError: 'dict' object has no attribute 'timezone'` was traced to the scheduler accessing `user.timezone` on a value that was a raw `dict` instead of a typed `User` model.

**Root cause (two-part):**

1. `User.model_config` had `extra='forbid'`. During a rolling deploy, the database manager (newer pod) can return fields that the client (older pod) doesn't yet know about. `extra='forbid'` caused `TypeAdapter(User).validate_python()` to raise `ValidationError` on those unknown fields.

2. `DynamicClient._get_return` had a silent `try/except` that swallowed the `ValidationError` and fell back to returning the raw `dict`. The scheduler then received a `dict` and crashed on `.timezone`.

## What

- **`backend/data/model.py`**: Change `User.model_config` `extra='forbid'` → `extra='ignore'`. Unknown fields from a newer database manager are silently dropped, making the RPC layer forward-compatible during rolling deploys. This is the primary fix.

- **`backend/util/service.py`**: Restore the `try/except` fallback in `_get_return`, but make it **observable**: log the full error message at `WARNING` (so ValidationError details — field name, value — appear in logs) and call `sentry_sdk.capture_exception(e)` so every fallback is tracked and alerted without crashing the caller. The raw result is still returned as before (continuity).

- **`backend/util/service_test.py`**: Add `TestGetReturn` with two direct unit tests: valid dict (including an unknown future field) → typed `User` returned; invalid dict (missing required fields) → fallback returns raw dict (no crash). Uses a typed `_SupportsGetReturn` Protocol + `cast` instead of `# type: ignore` suppressors.

- **`backend/executor/utils_test.py`**: Fix misleading docstring; move inner imports to module top level per code style.

## How

`extra='ignore'` is the standard Pydantic pattern for forward-compatible models at service boundaries. It means a rolling deploy where the DB manager has a new column will not break older client pods — the extra field is simply dropped on deserialization.

The restored `_get_return` fallback preserves continuity (callers don't crash) while the `logger.warning` + `sentry_sdk.capture_exception` ensure no schema mismatch goes undetected. Silent degradation is replaced by observable degradation.

## Checklist

- [x] Changes are backward-compatible (unknown fields ignored, not rejected)
- [x] Regression tests added for `_get_return` typed deserialization contract
- [x] Fallback preserved with observable logging and Sentry capture (no silent degradation)
- [x] `extra='ignore'` is consistent with forward-compatibility requirements at service boundaries
- [x] No `# type: ignore` suppressors introduced

### DIFF
--- a/autogpt_platform/backend/backend/data/model.py
+++ b/autogpt_platform/backend/backend/data/model.py
@@ -54,7 +54,6 @@ class User(BaseModel):
     """Application-layer User model with snake_case convention."""
 
     model_config = ConfigDict(
-        extra="ignore",
         str_strip_whitespace=True,
     )
 
--- a/autogpt_platform/backend/backend/data/model.py
+++ b/autogpt_platform/backend/backend/data/model.py
@@ -54,7 +54,7 @@ class User(BaseModel):
     """Application-layer User model with snake_case convention."""
 
     model_config = ConfigDict(
-        extra="forbid",
+        extra="ignore",
         str_strip_whitespace=True,
     )
 
--- a/autogpt_platform/backend/backend/executor/utils_test.py
+++ b/autogpt_platform/backend/backend/executor/utils_test.py
@@ -1,10 +1,13 @@
+from datetime import datetime, timezone
 from typing import cast
 
 import pytest
 from pytest_mock import MockerFixture
 
 from backend.data.dynamic_fields import merge_execution_input, parse_execution_output
-from backend.data.execution import ExecutionStatus
+from backend.data.execution import ExecutionStatus, GraphExecutionWithNodes
+from backend.data.model import User
+from backend.executor.utils import add_graph_execution
 from backend.util.mock import MockObject
 
 
@@ -483,19 +486,12 @@ async def test_add_graph_execution_via_rpc_returns_typed_user(
     mocker: MockerFixture,
 ):
     """
-    Regression test: when prisma is NOT connected (executor/scheduler process),
-    udb is the DatabaseManagerAsyncClient whose get_user_by_id goes through the
-    RPC layer (_get_return in service.py). _get_return must deserialize the JSON
-    dict into a typed User model — not silently return a raw dict. Without the
-    fix, _get_return's silent fallback returned a raw dict and accessing
-    user.timezone raised AttributeError: 'dict' object has no attribute 'timezone'.
+    Regression test: `add_graph_execution` accesses `user.timezone` on the User
+    returned by `get_user_by_id`. This test verifies the downstream code path
+    completes without AttributeError when `get_user_by_id` returns a proper typed
+    User model. Note: the mock returns a User directly — _get_return deserialization
+    is not exercised here; see TestGetReturn in util/service_test.py for that.
     """
-    from datetime import datetime, timezone
-
-    from backend.data.execution import GraphExecutionWithNodes
-    from backend.data.model import User
-    from backend.executor.utils import add_graph_execution
-
     graph_id = "test-graph-id"
     user_id = "test-user-id"
 
--- a/autogpt_platform/backend/backend/executor/utils_test.py
+++ b/autogpt_platform/backend/backend/executor/utils_test.py
@@ -515,7 +515,9 @@ async def test_add_graph_execution_via_rpc_returns_typed_user(
     mock_validate.return_value = (mock_graph, [], {}, set())
 
     mock_prisma = mocker.patch("backend.executor.utils.prisma")
-    mock_prisma.is_connected.return_value = False  # ← executor/scheduler path
+    mock_prisma.is_connected.return_value = (
+        False  # prisma not connected: uses RPC path instead
+    )
 
     # The RPC layer (_get_return) deserializes JSON dicts into typed Pydantic models.
     # The mock simulates what add_graph_execution receives after that deserialization:
--- a/autogpt_platform/backend/backend/executor/utils_test.py
+++ b/autogpt_platform/backend/backend/executor/utils_test.py
@@ -474,6 +474,109 @@ async def test_add_graph_execution_is_repeatable(mocker: MockerFixture):
 
 
 # ============================================================================
+# Regression test: RPC layer returns typed User model, not raw dict
+# ============================================================================
+
+
+@pytest.mark.asyncio
+async def test_add_graph_execution_via_rpc_returns_typed_user(
+    mocker: MockerFixture,
+):
+    """
+    Regression test: when prisma is NOT connected (executor/scheduler process),
+    udb is the DatabaseManagerAsyncClient whose get_user_by_id goes through the
+    RPC layer (_get_return in service.py). _get_return must deserialize the JSON
+    dict into a typed User model — not silently return a raw dict. Without the
+    fix, _get_return's silent fallback returned a raw dict and accessing
+    user.timezone raised AttributeError: 'dict' object has no attribute 'timezone'.
+    """
+    from datetime import datetime, timezone
+
+    from backend.data.execution import GraphExecutionWithNodes
+    from backend.data.model import User
+    from backend.executor.utils import add_graph_execution
+
+    graph_id = "test-graph-id"
+    user_id = "test-user-id"
+
+    mock_graph = mocker.MagicMock()
+    mock_graph.version = 1
+
+    mock_graph_exec = mocker.MagicMock(spec=GraphExecutionWithNodes)
+    mock_graph_exec.id = "exec-id-rpc"
+    mock_graph_exec.node_executions = []
+    mock_graph_exec.status = ExecutionStatus.QUEUED
+    mock_graph_exec.graph_version = 1
+    mock_graph_exec.to_graph_execution_entry.return_value = mocker.MagicMock()
+
+    mock_queue = mocker.AsyncMock()
+    mock_event_bus = mocker.MagicMock()
+    mock_event_bus.publish = mocker.AsyncMock()
+
+    mock_validate = mocker.patch(
+        "backend.executor.utils.validate_and_construct_node_execution_input"
+    )
+    mock_validate.return_value = (mock_graph, [], {}, set())
+
+    mock_prisma = mocker.patch("backend.executor.utils.prisma")
+    mock_prisma.is_connected.return_value = False  # ← executor/scheduler path
+
+    # The RPC layer (_get_return) deserializes JSON dicts into typed Pydantic models.
+    # The mock simulates what add_graph_execution receives after that deserialization:
+    # a proper User model, not a raw dict.
+    mock_user = User(
+        id=user_id,
+        email="test@example.com",
+        name=None,
+        created_at=datetime.now(timezone.utc),
+        updated_at=datetime.now(timezone.utc),
+        stripe_customer_id=None,
+        top_up_config=None,
+        timezone="UTC",
+    )
+
+    mock_db_client = mocker.MagicMock()
+    mock_db_client.get_user_by_id = mocker.AsyncMock(return_value=mock_user)
+    mock_db_client.get_graph_settings = mocker.AsyncMock(
+        return_value=mocker.MagicMock(
+            human_in_the_loop_safe_mode=False, sensitive_action_safe_mode=False
+        )
+    )
+    mock_db_client.create_graph_execution = mocker.AsyncMock(
+        return_value=mock_graph_exec
+    )
+    mock_db_client.update_graph_execution_stats = mocker.AsyncMock(
+        return_value=mock_graph_exec
+    )
+    mock_db_client.update_node_execution_status_batch = mocker.AsyncMock()
+    mock_workspace = mocker.MagicMock()
+    mock_workspace.id = "ws-id"
+    mock_db_client.get_or_create_workspace = mocker.AsyncMock(
+        return_value=mock_workspace
+    )
+    mock_db_client.increment_onboarding_runs = mocker.AsyncMock()
+
+    mocker.patch(
+        "backend.executor.utils.get_database_manager_async_client",
+        return_value=mock_db_client,
+    )
+    mocker.patch(
+        "backend.executor.utils.get_async_execution_queue", return_value=mock_queue
+    )
+    mocker.patch(
+        "backend.executor.utils.get_async_execution_event_bus",
+        return_value=mock_event_bus,
+    )
+
+    # Must not raise AttributeError: 'dict' object has no attribute 'timezone'
+    result = await add_graph_execution(
+        graph_id=graph_id,
+        user_id=user_id,
+    )
+    assert result == mock_graph_exec
+
+
+# ============================================================================
 # Tests for Optional Credentials Feature
 # ============================================================================
 
--- a/autogpt_platform/backend/backend/util/service.py
+++ b/autogpt_platform/backend/backend/util/service.py
@@ -709,19 +709,9 @@ def get_service_client(
             return kwargs
 
         def _get_return(self, expected_return: TypeAdapter | None, result: Any) -> Any:
-            """Validate and coerce the RPC result to the expected return type.
-
-            Falls back to the raw result with a warning if validation fails.
-            """
+            """Validate and coerce the RPC result to the expected return type."""
             if expected_return:
-                try:
-                    return expected_return.validate_python(result)
-                except Exception as e:
-                    logger.warning(
-                        "RPC return type validation failed, using raw result: %s",
-                        type(e).__name__,
-                    )
-                    return result
+                return expected_return.validate_python(result)
             return result
 
         def __getattr__(self, name: str) -> Callable[..., Any]:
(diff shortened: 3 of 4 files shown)


autogpt_platform/backend/backend/util/service.py
```diff
@@ -709,19 +709,9 @@ def _get_params(
...
             if expected_return:
-                try:
-                    return expected_return.validate_python(result)
-                except Exception as e:
-                    logger.warning(
-                        "RPC return type validation failed, using raw result: %s",
-                        type(e).__name__,
-                    )
-                    return result
+                return expected_return.validate_python(result)
```
majdyz: 🤖 🔴 **Blocker**: The fallback was removed entirely, but this is too aggressive. With `extra='ignore'` on `User`, the specific field-addition failure mode is fixed. However, other RPC-returned types don't have this protection — a type change, missing required field, or any other schema mismatch on any method will now raise `ValidationError` and crash the caller (scheduler, executor) immediately with no recovery.

Recommended middle ground: **restore the fallback but make it observable** so you get visibility without brittleness:

```python
def _get_return(self, expected_return: TypeAdapter | None, result: Any) -> Any:
    """Validate and coerce the RPC result to the expected return type."""
    if expected_return:
        try:
            return expected_return.validate_python(result)
        except Exception as e:
            logger.warning(
                "RPC return type validation failed for %s: %s",
                type(e).__name__,
                str(e),  # full details instead of just type name
            )
            sentry_sdk.capture_exception(e)  # tracked + alerted, not silent
            return result
    return result
```

This makes every fallback visible in Sentry and logs while preserving the original resilience.

autogpt_platform/backend/backend/util/service_test.py
```diff
@@ -688,3 +691,46 @@ async def test_health_check_during_shutdown(test_service):
...
+        """
+        now = datetime.now(timezone.utc).isoformat()
+        valid_dict = {
+            "id": "user-id",
+            "email": "test@example.com",
+            "created_at": now,
+            "updated_at": now,
+            "unknown_future_field": "some_value",  # simulates a new DB field during deploy
+        }
+        client = self._make_client()
```
majdyz, line 727: 🤖 🟠 **Should Fix**: `# type: ignore[attr-defined]` violates the backend convention of no linter suppressors. `_get_return` is a private method on an inner class, so direct access is awkward. Consider either (a) testing through a public method that exercises `_get_return` end-to-end, or (b) exposing a minimal `_test_get_return` helper on the client for testing purposes, or (c) refactoring `_get_return` to a module-level function so it's directly testable without suppression.

autogpt_platform/backend/backend/util/service_test.py
```diff
@@ -688,3 +691,46 @@ async def test_health_check_during_shutdown(test_service):
...
+        result = client._get_return(adapter, valid_dict)  # type: ignore[attr-defined]
+
+        assert isinstance(result, User)
+        assert result.timezone is not None
+
+    def test_invalid_dict_raises_validation_error(self):
+        """TypeAdapter(User) + invalid dict (missing required fields) → ValidationError raised."""
+        invalid_dict = {"id": "user-id"}  # missing email, created_at, updated_at
+        client = self._make_client()
+        adapter = TypeAdapter(User)
```
majdyz, line 738: 🤖 🟠 **Should Fix**: This test asserts `ValidationError` is raised — but if the fallback is restored (as the blocker comment above suggests), this test will fail because the fallback catches the exception and returns the raw dict instead. The test needs to align with whichever contract is chosen: either remove the fallback (and this test is correct) or restore it (and this test should assert the raw dict is returned with a warning logged/Sentry captured instead).

autogpt_platform/backend/backend/executor/utils_test.py
```diff
@@ -473,6 +476,102 @@ async def test_add_graph_execution_is_repeatable(mocker: MockerFixture):
...
+    graph_id = "test-graph-id"
+    user_id = "test-user-id"
+
+    mock_graph = mocker.MagicMock()
+    mock_graph.version = 1
+
+    mock_graph_exec = mocker.MagicMock(spec=GraphExecutionWithNodes)
+    mock_graph_exec.id = "exec-id-rpc"
+    mock_graph_exec.node_executions = []
+    mock_graph_exec.status = ExecutionStatus.QUEUED
```
majdyz, line 504: 🤖 🔵 **Nit**: `# ←` inline arrow comments aren't idiomatic Python style — prefer a plain `# note: ...` comment or no comment if the code is self-explanatory.

autogpt_platform/backend/backend/util/service.py
```diff
@@ -9,6 +9,8 @@
 import threading
 import time
 from abc import ABC, abstractmethod
+
+import sentry_sdk
```
majdyz: 🤖 🟠 **Should Fix**: `import sentry_sdk` (third-party) is placed between two stdlib imports — `from abc import ABC, abstractmethod` above it and `from contextlib import asynccontextmanager` / `from functools import update_wrapper` / `from typing import ...` below it. PEP 8 (and isort) require all stdlib imports in one group, then all third-party imports in a separate group separated by a blank line. This likely slipped through because the commit used `--no-verify`. Fix: run `poetry run format` and recommit.

autogpt_platform/backend/backend/executor/utils_test.py
```diff
@@ -473,6 +476,102 @@ async def test_add_graph_execution_is_repeatable(mocker: MockerFixture):
...
+    mock_event_bus = mocker.MagicMock()
+    mock_event_bus.publish = mocker.AsyncMock()
+
+    mock_validate = mocker.patch(
+        "backend.executor.utils.validate_and_construct_node_execution_input"
+    )
+    mock_validate.return_value = (mock_graph, [], {}, set())
+
+    mock_prisma = mocker.patch("backend.executor.utils.prisma")
+    mock_prisma.is_connected.return_value = False  # ← executor/scheduler path
```
majdyz: 🤖 🔵 **Nit**: `# ←` arrow comment is still present (flagged in Round 1 as not idiomatic Python style). Please remove or replace with a plain comment.

majdyz (commented):
## 📋 Round 2 Review — PR #12707

**Round 1 blockers addressed** ✅ Fallback restored with observability; `type: ignore` suppressors removed via Protocol approach; test renamed and updated.

**Remaining issues (2):**

1. 🟠 **Should Fix** — `import sentry_sdk` placed between stdlib imports in `service.py` (see inline). Run `poetry run format` to fix ordering.
2. 🔵 **Nit** — `# ←` arrow comment in `utils_test.py` still present from Round 1.

**PR description is now stale** — The `## What` section and checklist still describe *removing* the fallback: *"Remove the silent try/except fallback in `_get_return`"* and *"No silent fallbacks remain in the RPC layer"*. The actual code restores the fallback with observable logging and Sentry capture. Please update the description to reflect this.

autogpt_platform/backend/backend/util/service.py
```diff
@@ -709,18 +710,17 @@ def _get_params(
...
             if expected_return:
                 try:
                     return expected_return.validate_python(result)
                 except Exception as e:
                     logger.warning(
-                        "RPC return type validation failed, using raw result: %s",
+                        "RPC return type validation failed for %s: %s",
                         type(e).__name__,
+                        str(e),
                     )
```
majdyz, line 723: 🤖 🟠 **Should Fix**: The AGENTS.md style guide says: *"Use `%s` for deferred interpolation in `debug` statements, f-strings elsewhere for readability."* This `logger.warning` uses `%s` printf-style, inconsistent with every other `logger.warning` call in this file (all use f-strings). Fix:
```python
logger.warning(
    f"RPC return type validation failed for {type(e).__name__}: {e}",
)
```

autogpt_platform/backend/backend/util/service.py
```diff
@@ -709,18 +710,17 @@ def _get_params(
             return kwargs
 
         def _get_return(self, expected_return: TypeAdapter | None, result: Any) -> Any:
-            """Validate and coerce the RPC result to the expected return type.
-
-            Falls back to the raw result with a warning if validation fails.
-            """
+            """Validate and coerce the RPC result to the expected return type."""
```
majdyz: 🤖 🔵 **Nit**: The docstring was shortened to remove the fallback description, but the fallback is still present in the code. Consider restoring the second sentence: *"Falls back to the raw result with a warning and Sentry capture if validation fails."* — this helps future readers understand why `except Exception` is intentional (not a lazy catch-all).

majdyz (commented):
## 📋 Round 3 Review — PR #12707

**Round 2 issues addressed** ✅ Import ordering fixed (`sentry_sdk` now in third-party block); arrow comment rewritten; PR description updated to accurately reflect restored fallback.

**Remaining issues (2):**

1. 🟠 **Should Fix** — `logger.warning` uses `%s` printf-style at line 719-722, but AGENTS.md requires f-strings for non-debug levels (all other `logger.warning` calls in this file use f-strings). See inline comment.
2. 🔵 **Nit** — `_get_return` docstring was truncated; the fallback behavior (observable degradation) is no longer documented in the docstring. See inline comment.

The primary fix (`extra='ignore'`) and test coverage are solid. Once the log format style is fixed, this PR is clean.

autogpt_platform/backend/backend/util/service.py
majdyz: Fixed in [6b84b64](https://github.com/Significant-Gravitas/AutoGPT/commit/6b84b642c4): Fallback restored with observable logging (`logger.warning`) and Sentry capture. The `_get_return` method now catches validation errors, logs them at WARNING level with full details, reports to Sentry, and returns the raw result — matching the recommended middle ground.

autogpt_platform/backend/backend/util/service_test.py
majdyz: Fixed in [6b84b64](https://github.com/Significant-Gravitas/AutoGPT/commit/6b84b642c4): Replaced `# type: ignore[attr-defined]` suppressors with a typed Protocol (`_SupportsGetReturn`) + `cast`. The `_make_client` helper now returns `_SupportsGetReturn` and calls to `client._get_return` are fully type-safe without any suppressors.

autogpt_platform/backend/backend/util/service_test.py
majdyz: Fixed in [6b84b64](https://github.com/Significant-Gravitas/AutoGPT/commit/6b84b642c4): Test now asserts the raw dict is returned (not `ValidationError`), consistent with the restored fallback. `test_invalid_dict_falls_back_to_raw_result` confirms the fallback contract.

autogpt_platform/backend/backend/executor/utils_test.py
majdyz: Fixed in [6f15f25](https://github.com/Significant-Gravitas/AutoGPT/commit/6f15f2550647bee683cde7289e248802b524ad85): Arrow comment removed; replaced with plain comment describing the executor/scheduler RPC path.

autogpt_platform/backend/backend/util/service.py
```diff
@@ -709,18 +711,17 @@ def _get_params(
...
                     return expected_return.validate_python(result)
                 except Exception as e:
                     logger.warning(
-                        "RPC return type validation failed, using raw result: %s",
+                        "RPC return type validation failed for %s: %s",
                         type(e).__name__,
+                        str(e),
                     )
+                    sentry_sdk.capture_exception(e)
                     return result
```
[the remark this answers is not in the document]
majdyz: Acknowledged. The fallback is intentional and observable — not silent degradation. The `extra='ignore'` on `User` addresses the specific field-addition failure mode from the Sentry alert. The restored fallback in `_get_return` handles other schema mismatches: validation failures now trigger `logger.warning` with full error details AND `sentry_sdk.capture_exception(e)`, so every fallback is tracked in Sentry and alerted. The raw result is returned to preserve caller continuity during mismatches, not to hide them.

autogpt_platform/backend/backend/util/service.py
majdyz: Fixed in [6f15f25](https://github.com/Significant-Gravitas/AutoGPT/commit/6f15f2550647bee683cde7289e248802b524ad85): Ran `poetry run format` — `sentry_sdk` import is now in the correct third-party block after all stdlib imports, separated by a blank line.

autogpt_platform/backend/backend/executor/utils_test.py
majdyz: Fixed in [6f15f25](https://github.com/Significant-Gravitas/AutoGPT/commit/6f15f2550647bee683cde7289e248802b524ad85): The `# ←` arrow comment is removed. The mock comment now reads `# prisma not connected: uses RPC path instead` and `# The RPC layer (_get_return) deserializes...` in plain prose.

autogpt_platform/backend/backend/util/service.py
majdyz: Fixed in [f030658](https://github.com/Significant-Gravitas/AutoGPT/commit/f03065834229abb11577e2a87523b49d12fc840c): `logger.warning` now uses an f-string: `f"RPC return type validation failed for {type(e).__name__}: {e}"` — consistent with all other `logger.warning` calls in this file.

autogpt_platform/backend/backend/util/service.py
majdyz: Fixed in [f030658](https://github.com/Significant-Gravitas/AutoGPT/commit/f03065834229abb11577e2a87523b49d12fc840c): Docstring second sentence restored: `Falls back to the raw result with a warning and Sentry capture if validation fails.`

autogpt_platform/backend/backend/executor/utils_test.py
```diff
@@ -473,6 +473,109 @@ async def test_add_graph_execution_is_repeatable(mocker: MockerFixture):
...
+        created_at=datetime.now(timezone.utc),
+        updated_at=datetime.now(timezone.utc),
+        stripe_customer_id=None,
+        top_up_config=None,
+        timezone="UTC",
+    )
+
+    mock_db_client = mocker.MagicMock()
+    mock_db_client.get_user_by_id = mocker.AsyncMock(return_value=mock_user)
+    mock_db_client.get_graph_settings = mocker.AsyncMock(
```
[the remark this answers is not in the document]
majdyz: Acknowledged. This test cannot exercise `_get_return` by design: `get_database_manager_async_client` is patched to return a `MagicMock` that bypasses `DynamicClient` entirely, so `_get_return` is never on the call stack. Switching the mock to return raw dicts would not change this — it would only cause the test to fail because `add_graph_execution` would receive a dict and crash on `.timezone`.

The `_get_return` deserialization contract (dict → typed model, fallback on failure) is covered directly in `TestGetReturn` in `util/service_test.py`, which calls `client._get_return(adapter, dict)` via a typed Protocol. The `utils_test` test covers the orthogonal concern: that the downstream code path (`add_graph_execution`) handles a typed `User` correctly. The docstring explicitly calls this out: *"Note: the mock returns a User directly — _get_return deserialization is not exercised here; see TestGetReturn in util/service_test.py for that."*

This is the highest-ROI split given the mock depth — the two tests together cover both layers.

majdyz:
**CI failures are pre-existing/infrastructure flakiness, not caused by this PR.**

**check API types** — Failed at "Set up Backend - Generate Prisma client" due to `npm error ECONNRESET` (network connectivity issue when fetching from npmjs.org). The previous Full-stack CI run for this branch ([24142435512](https://github.com/Significant-Gravitas/AutoGPT/actions/runs/24142435512)) passed this check successfully. This PR only modifies 4 backend Python files (`model.py`, `service.py`, `service_test.py`, `utils_test.py`).

**end-to-end tests** — 84/86 tests passed. Only 1 test failed: `build.spec.ts:106 › Builder › user can run an agent from the builder`. This frontend timing test is unrelated to this PR's changes (backend RPC model changes).

This PR's code changes are clean — all backend tests pass locally and in the "AutoGPT Platform - Backend CI" run.

autogpt_platform/backend/backend/data/model.py
```diff
@@ -54,7 +54,7 @@ class User(BaseModel):
     """Application-layer User model with snake_case convention."""
 
     model_config = ConfigDict(
-        extra="forbid",
+        extra="ignore",
```
Pwuts: this is the default so can be omitted

Pwuts (dismissed):

autogpt_platform/backend/backend/data/model.py
majdyz: 🤖 Fixed in [7118f9d5b](https://github.com/Significant-Gravitas/AutoGPT/commit/7118f9d5bdea5dc3a94e06f0f1c29906e73b473e): removed redundant `extra="ignore"` — it's Pydantic's default so omitting it keeps the same forward-compatible behavior.

PR #12709, opened 2026-04-08
fix(backend/executor): OrchestratorBlock dry-run credentials + Responses API status field

majdyz:
## Why
Two bugs block OrchestratorBlock from working correctly:

1. **Dry-run always fails with "credentials required"** even when `OPEN_ROUTER_API_KEY` is set on dev. The n8n conversion dry-run hits this.
2. **Agent-mode OrchestratorBlock fails on the second LLM call** with `Error code: 400 – Unknown parameter: 'input[2].status'` when using OpenAI models (Responses API path).

## What
**Bug 1 — manager.py credential null** (`backend/executor/manager.py`):
The dry-run path called `input_data[field_name] = None` to "clear" the credential slot, but `_execute` in `_base.py` filters out `None` values before calling `input_schema(**...)`. This drops the required `credentials` field from the schema constructor, causing a Pydantic validation error.

Fix: Don't null out the field. If the user already has credential metadata in `input_data` (normal case), leave it intact. If not (no credentials configured), synthesise a minimal `CredentialsMetaInput`-compatible placeholder from the platform credentials so schema construction passes. The actual `APIKeyCredentials` (platform key) is still injected via `extra_exec_kwargs`.

**Bug 2 — Responses API `status` field** (`backend/blocks/orchestrator.py`):
OpenAI returns output items (function calls, messages) with a `status: "completed"` field. When `_convert_raw_response_to_dict` serialises these items and they are stored in `conversation_history`, they are sent back as input on the next call — but OpenAI rejects `status` as an input-only field.

Fix: Strip `status` from each output item before it enters the history.

## How
- `manager.py` lines 311-314: removed the `input_data[field_name] = None` nullification; added a conditional placeholder when no credential metadata is present.
- `orchestrator.py` `_convert_raw_response_to_dict`: filter `k != "status"` when extracting Responses API output items.
- Tests added for both fixes.

## Checklist
- [x] Tests written and passing (94 total, all green)
- [x] Pre-commit hooks passed (Black, Ruff, isort, typecheck)
- [x] No out-of-scope changes

### DIFF
--- a/autogpt_platform/backend/backend/blocks/orchestrator.py
+++ b/autogpt_platform/backend/backend/blocks/orchestrator.py
@@ -251,8 +251,13 @@ def _convert_raw_response_to_dict(
         # Already a dict (from tests or some providers)
         return raw_response
     elif _is_responses_api_object(raw_response):
-        # OpenAI Responses API: extract individual output items
-        items = [json.to_dict(item) for item in raw_response.output]
+        # OpenAI Responses API: extract individual output items.
+        # Strip 'status' — it's a response-only field that OpenAI rejects
+        # when the item is sent back as input on the next API call.
+        items = [
+            {k: v for k, v in json.to_dict(item).items() if k != "status"}
+            for item in raw_response.output
+        ]
         return items if items else [{"role": "assistant", "content": ""}]
     else:
         # Chat Completions / Anthropic return message objects
--- a/autogpt_platform/backend/backend/executor/manager.py
+++ b/autogpt_platform/backend/backend/executor/manager.py
@@ -312,7 +312,7 @@ async def execute_node(
         # doesn't fail on the required field.  If no metadata is present,
         # synthesize a minimal placeholder from the platform credentials.
         if _dry_run_creds is not None:
-            if not input_data.get(field_name):
+            if input_data.get(field_name) is None:
                 input_data[field_name] = {
                     "id": _dry_run_creds.id,
                     "provider": _dry_run_creds.provider,
--- a/autogpt_platform/backend/backend/executor/manager.py
+++ b/autogpt_platform/backend/backend/executor/manager.py
@@ -307,9 +307,18 @@ async def execute_node(
 
     # Handle regular credentials fields
     for field_name, input_type in input_model.get_credentials_fields().items():
-        # Dry-run platform credentials bypass the credential store
+        # Dry-run platform credentials bypass the credential store.
+        # Keep the existing credential metadata so _execute's input_schema(**...)
+        # doesn't fail on the required field.  If no metadata is present,
+        # synthesize a minimal placeholder from the platform credentials.
         if _dry_run_creds is not None:
-            input_data[field_name] = None
+            if not input_data.get(field_name):
+                input_data[field_name] = {
+                    "id": _dry_run_creds.id,
+                    "provider": _dry_run_creds.provider,
+                    "type": _dry_run_creds.type,
+                    "title": _dry_run_creds.title,
+                }
             extra_exec_kwargs[field_name] = _dry_run_creds
             continue
 
--- a/autogpt_platform/backend/backend/executor/simulator_test.py
+++ b/autogpt_platform/backend/backend/executor/simulator_test.py
@@ -261,10 +261,12 @@ class TestGetDryRunCredentials:
     def test_credentials_have_metadata_fields_for_placeholder(self) -> None:
         """The returned credentials must have id, provider, type, and title so
         manager.py can synthesise a valid CredentialsMetaInput placeholder."""
+        from backend.integrations.providers import ProviderName
+
         creds = get_dry_run_credentials({"_dry_run_api_key": "sk-or-test"})
         assert creds is not None
         assert creds.id == "dry-run-platform"
-        assert str(creds.provider) in ("open_router", "ProviderName.OPEN_ROUTER")
+        assert creds.provider == ProviderName.OPEN_ROUTER
         assert creds.type == "api_key"
         assert creds.title is not None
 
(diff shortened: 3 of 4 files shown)


autogpt_platform/backend/backend/executor/manager.py
```diff
@@ -307,9 +307,18 @@ async def execute_node(
...
     # Handle regular credentials fields
     for field_name, input_type in input_model.get_credentials_fields().items():
-        # Dry-run platform credentials bypass the credential store
+        # Dry-run platform credentials bypass the credential store.
+        # Keep the existing credential metadata so _execute's input_schema(**...)
+        # doesn't fail on the required field.  If no metadata is present,
+        # synthesize a minimal placeholder from the platform credentials.
         if _dry_run_creds is not None:
-            input_data[field_name] = None
+            if not input_data.get(field_name):
```
majdyz: 🤖 🔵 **Nit**: `if not input_data.get(field_name)` uses a falsy check, so it would also overwrite an explicitly set empty dict or empty string credential value. For credentials fields the practical difference is nil, but `if input_data.get(field_name) is None:` expresses the intent more precisely and is harder to misread.

autogpt_platform/backend/backend/blocks/orchestrator.py
```diff
@@ -251,8 +251,13 @@ def _convert_raw_response_to_dict(
         # Already a dict (from tests or some providers)
         return raw_response
     elif _is_responses_api_object(raw_response):
-        # OpenAI Responses API: extract individual output items
-        items = [json.to_dict(item) for item in raw_response.output]
+        # OpenAI Responses API: extract individual output items.
+        # Strip 'status' — it's a response-only field that OpenAI rejects
+        # when the item is sent back as input on the next API call.
+        items = [
+            {k: v for k, v in json.to_dict(item).items() if k != "status"}
```
majdyz, line 258: 🤖 🟡 **Nice to Have**: Only `status` is stripped, but OpenAI Responses API output items also include an `id` field (e.g. `fc_abc123`) that is not part of the input schema for conversation history. The logged error only flagged `status` because validation stopped at the first unknown field. Worth verifying against the Responses API input item schema whether `id` (and potentially `object`) also need stripping, or documenting that they are intentionally kept.

autogpt_platform/backend/backend/executor/simulator_test.py
```diff
@@ -234,6 +235,40 @@ def test_regular_block_returns_none(self) -> None:
...
+        creds = get_dry_run_credentials(input_data)
+        assert creds is None
+
+    def test_credentials_have_metadata_fields_for_placeholder(self) -> None:
+        """The returned credentials must have id, provider, type, and title so
+        manager.py can synthesise a valid CredentialsMetaInput placeholder."""
+        creds = get_dry_run_credentials({"_dry_run_api_key": "sk-or-test"})
+        assert creds is not None
+        assert creds.id == "dry-run-platform"
+        assert str(creds.provider) in ("open_router", "ProviderName.OPEN_ROUTER")
```
majdyz: 🤖 🟠 **Should Fix**: `str(creds.provider) in ("open_router", "ProviderName.OPEN_ROUTER")` is brittle — it accounts for two different string representations because the underlying enum's `__str__` may vary. Use the enum directly instead:
```python
from backend.data.model import ProviderName
assert creds.provider == ProviderName.OPEN_ROUTER
```
This is type-safe and immune to repr changes.

autogpt_platform/backend/backend/executor/manager.py
majdyz: 🤖 Fixed in [1acfc3ab](https://github.com/Significant-Gravitas/AutoGPT/commit/1acfc3ab4410566da73b54beab183084f0b3f7b4): replaced `if not input_data.get(field_name)` with `if input_data.get(field_name) is None`.

autogpt_platform/backend/backend/blocks/orchestrator.py
majdyz: 🤖 Checked the OpenAI SDK types: `ResponseFunctionToolCallParam.id` is optional and `ResponseOutputMessageParam.id` is required — so `id` must be kept. The production error only flagged `status`, and the SDK confirms `id` is valid in input items. Stripping only `status` is correct. No change needed.

autogpt_platform/backend/backend/executor/simulator_test.py
majdyz: 🤖 Fixed in [1acfc3ab](https://github.com/Significant-Gravitas/AutoGPT/commit/1acfc3ab4410566da73b54beab183084f0b3f7b4): replaced `str(creds.provider) in (...)` with `assert creds.provider == ProviderName.OPEN_ROUTER` using a local import.

autogpt_platform/backend/backend/executor/manager.py
```diff
@@ -307,9 +307,18 @@ async def execute_node(
...
+                input_data[field_name] = {
+                    "id": _dry_run_creds.id,
+                    "provider": _dry_run_creds.provider,
+                    "type": _dry_run_creds.type,
+                    "title": _dry_run_creds.title,
+                }
```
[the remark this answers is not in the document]
majdyz: 🤖 The provider mismatch is not actionable here: `_execute` in `_base.py` explicitly skips credential field validation in dry-run mode (lines 707-712 exclude credential fields from `validate_data`). `CredentialsMetaInput` also doesn't enforce provider against a whitelist — it stores provider as a plain string. So the metadata provider value doesn't cause a validation error regardless of which actual credentials are injected. Current logic is intentional: synthesize a minimal placeholder only when no metadata exists (to satisfy the `input_schema(**...)` construction at line 728), while preserving any pre-existing metadata otherwise.

PR #12710, opened 2026-04-08
fix(backend): lower PlatformCostLog DataError to warning during schema mismatch

majdyz:
### Why / What / How

**Why:** After merging #12696 (platform cost tracking), the `autogpt-database-manager` pod was running a stale Prisma client that didn't yet know about the new `PlatformCostLog` fields (`userId`, `metadata`, etc.). Every copilot cost-log attempt raised `prisma.errors.DataError`, which was caught by `logger.exception()` — logging at ERROR level and spamming Sentry (issue AUTOGPT-SERVER-8GS, 33+ events, escalating).

**What:** Both `log_platform_cost_safe` (in `platform_cost.py`) and `_safe_log` (in `token_tracking.py`) now catch `prisma.errors.DataError` specifically before the generic `Exception` handler, logging at WARNING instead of ERROR. All other unexpected exceptions still go to Sentry as before.

**How:** Added a targeted `except DataError` block that logs a human-readable warning with a "schema mismatch?" hint. Once the DB manager pod rolls over with the updated Prisma client, the warnings stop automatically — no code changes needed at that point.

### Changes 🏗️

- `backend/data/platform_cost.py`: catch `DataError` in `log_platform_cost_safe` → WARNING
- `backend/copilot/token_tracking.py`: catch `DataError` in `_safe_log` → WARNING

#### For code changes:
- [x] I have made a test plan
  - [x] Verified Sentry issue AUTOGPT-SERVER-8GS is caused by stale Prisma client (deployment gap), not a code bug
  - [x] Confirmed `DataError` handler fires before generic `Exception` handler
  - [x] Pre-commit hooks pass

### DIFF
--- a/autogpt_platform/backend/backend/copilot/token_tracking.py
+++ b/autogpt_platform/backend/backend/copilot/token_tracking.py
@@ -57,12 +57,9 @@ def _schedule_cost_log(entry: PlatformCostEntry) -> None:
                 # stale Prisma client (e.g. during a rolling deploy after a schema
                 # migration). Log at WARNING so Sentry is not spammed.
                 logger.warning(
-                    "Skipping platform cost log (schema mismatch?) for "
-                    "user=%s provider=%s block=%s: %s",
-                    entry.user_id,
-                    entry.provider,
-                    entry.block_name,
-                    e,
+                    f"Skipping platform cost log (schema mismatch?) for "
+                    f"user={entry.user_id} provider={entry.provider} "
+                    f"block={entry.block_name}: {e}"
                 )
             except Exception:
                 logger.exception(
--- a/autogpt_platform/backend/backend/copilot/token_tracking.py
+++ b/autogpt_platform/backend/backend/copilot/token_tracking.py
@@ -15,6 +15,8 @@ import math
 import re
 import threading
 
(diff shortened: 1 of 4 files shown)


majdyz:
Fixed in [9710dc76](https://github.com/Significant-Gravitas/AutoGPT/commit/9710dc76fca91bc47aa572dd451e41083a9a1aed): switched both `logger.warning` calls from `%s` placeholder style to f-strings in `platform_cost.py` and `token_tracking.py`, matching backend convention for non-debug log levels.

Pwuts:
Seems unnecessary once #12713 gets merged

majdyz:
Good point — #12713 fixes the root cause (missing `SafeJson` wrap on `metadata`). This PR is still worthwhile as a defence-in-depth measure: `DataError` can also surface during rolling deploys when the DB manager pod runs a stale Prisma client that doesn't yet know about a new column (the original Sentry spike, AUTOGPT-SERVER-8GS, was caused by exactly that). Keeping the warning-level catch prevents Sentry from being spammed if the same deployment-gap pattern happens again in future migrations. Happy to close this if you'd rather keep it simple and just rely on #12713 though.

PR #12713, opened 2026-04-08
fix(backend): wrap PlatformCostLog metadata in SafeJson to fix silent DataError

majdyz:
## Changes

- Wrap `metadata` field in `SafeJson()` when calling `PrismaLog.prisma().create()` in `log_platform_cost`
- Add `platform_cost_integration_test.py` with DB round-trip tests for the fix

## Why

`PrismaLog.prisma().create()` was silently failing with a `DataError` because passing a plain Python `dict` to a `Json?`-typed Prisma field is not allowed:

```
DataError: Invalid argument type. `metadata` should be of type NullableJsonNullValueInput or Json
```

The error was swallowed silently by `logger.exception` in the background task, so **no rows ever landed in `PlatformCostLog`** — which is why the dev admin cost dashboard showed no data after #12696 was merged.

## How

Wrap `entry.metadata` in `SafeJson()` (already used throughout the codebase, lives in `backend/util/json.py`) before passing it to the Prisma create call. `SafeJson` extends `prisma.Json`, sanitizes PostgreSQL-incompatible control characters, and handles Pydantic-model conversion.

Add two integration tests in `platform_cost_integration_test.py` (following the `credit_integration_test.py` pattern) that write a record to a real DB and read it back — confirming both metadata round-trip and NULL metadata work correctly.

## Test plan

- [x] Integration tests verify metadata persists/reads correctly via Prisma
- [x] Unit tests updated: `isinstance(data["metadata"], Json)` confirms the field is wrapped
- [x] Verified on dev executor pod: cost rows now appear in the admin dashboard after fix

### DIFF
--- a/autogpt_platform/backend/backend/data/platform_cost.py
+++ b/autogpt_platform/backend/backend/data/platform_cost.py
@@ -4,6 +4,7 @@ from datetime import datetime, timedelta, timezone
 from typing import Any
 
 from prisma.models import PlatformCostLog as PrismaLog
+from prisma.types import PlatformCostLogCreateInput
 from pydantic import BaseModel
 
 from backend.data.db import query_raw_with_schema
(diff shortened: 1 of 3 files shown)


autogpt_platform/backend/backend/data/platform_cost.py
```diff
@@ -51,29 +51,42 @@ class PlatformCostEntry(BaseModel):
 
 
 async def log_platform_cost(entry: PlatformCostEntry) -> None:
```
majdyz, line 55: 🤖 🔴 **Blocker**: Tests in `platform_cost_test.py` mock `PrismaLog.prisma` which is no longer imported. All 4 tests in `TestLogPlatformCost` / `TestLogPlatformCostSafe` will fail with `AttributeError`. They need to be updated to mock `backend.data.platform_cost.execute_raw_with_schema` instead, and the assertion `data["metadata"] == {"key": "val"}` needs to change to verify `json.dumps({"key": "val"})` is passed as the 18th arg.

autogpt_platform/backend/backend/data/platform_cost.py
```diff
@@ -51,29 +51,42 @@ class PlatformCostEntry(BaseModel):
...
-    await PrismaLog.prisma().create(
-        data={
-            "userId": entry.user_id,
```
Pwuts, line 56: You probably have to use a `connect` for this, not the raw `userId`

Pwuts (changes requested):
Wouldn't the fix be to pass the correct type rather than move to an untyped raw query?

autogpt_platform/backend/backend/data/platform_cost.py
majdyz: The scalar `userId` field works fine here — it's a `String?` column, not a required relation connect. The same pattern (`userId=user_id` as a scalar in a Prisma create) is used throughout the codebase, e.g. `credit.py:681`. The original `userId: Field does not exist in enclosing type` error was a cascading/misleading parse failure triggered by the `metadata` type mismatch, not a real missing-field error.

autogpt_platform/backend/backend/data/platform_cost.py
majdyz: 🤖 Fixed in [9aa9f5361](https://github.com/Significant-Gravitas/AutoGPT/commit/9aa9f5361a68586881c92fa7a0e8f7dcabe17bc1): reverted to SafeJson + Prisma create approach — `PrismaLog` is re-imported, unit tests mock `PrismaLog.prisma` again, and the assertion now checks `isinstance(data["metadata"], Json)` to confirm the wrapper is applied rather than a plain dict.

majdyz:
Hi @Pwuts — you're right. The PR was updated in [9aa9f5361](https://github.com/Significant-Gravitas/AutoGPT/commit/9aa9f5361a68586881c92fa7a0e8f7dcabe17bc1) to use the correct type: the metadata field is now wrapped in `SafeJson()` (a `prisma.Json` subclass from `backend/util/json.py`) before the `PrismaLog.prisma().create()` call. No raw SQL in the final implementation.

majdyz:
Fixed failing integration test in [6938b0379](https://github.com/Significant-Gravitas/AutoGPT/commit/6938b037981a10f4e31cbd3b3b9abd5669780255): Prisma's `Json?` field rejects Python `None` directly. Changed to omit the `metadata` key when it's `None` — Prisma defaults to NULL for omitted nullable fields.

majdyz:
🤖 Fixed in [40f3fd216](https://github.com/Significant-Gravitas/AutoGPT/commit/40f3fd2161f30f78a6bc9e3063106bbbfaed5a86): reverted to `PrismaLog.prisma().create()` with `SafeJson(entry.metadata)` as the correct Prisma-typed approach. `SafeJson` extends `prisma.Json`, sanitizes PostgreSQL control chars, and handles complex types (datetime, UUID, Decimal) via `jsonable_encoder` — addressing both the original `DataError` and the serialization concern. The raw SQL fallback has been removed.

PR #12615, opened 2026-03-31
feat(backend): platform server linking API for multi-platform CoPilot

Bentlybro:
## Why
AutoPilot (CoPilot) needs to reach users across chat platforms — Discord first, Telegram / Slack / Teams / WhatsApp next. To make usage and billing coherent, every conversation resolves to one AutoGPT account. There are two independent linking flows:

- **SERVER links**: the first person to claim a server (Discord guild, Telegram group, …) becomes its owner. Anyone in the server can chat with the bot; all usage bills to the owner.
- **USER links**: an individual links their 1:1 DMs with the bot to their own AutoGPT account. Independent from server links — a server owner still has to link their DMs separately.

## What
Backend for platform linking, split cleanly by trust boundary:

- **Bot-facing operations** run over cluster-internal RPC via a new `PlatformLinkingManager(AppService)`. No shared bearer token; trust is the cluster network itself.
- **User-facing operations** stay on REST under JWT auth (the same pattern as every other feature).

### REST endpoints (JWT auth)

- `GET  /api/platform-linking/tokens/{token}/info` — non-sensitive display info for the link page
- `POST /api/platform-linking/tokens/{token}/confirm` — confirm a SERVER link
- `POST /api/platform-linking/user-tokens/{token}/confirm` — confirm a USER link
- `GET  /api/platform-linking/links` / `DELETE /links/{id}` — manage server links
- `GET  /api/platform-linking/user-links` / `DELETE /user-links/{id}` — manage DM links

### `PlatformLinkingManager` `@expose` methods (internal RPC)

- `resolve_server_link(platform, platform_server_id) -> ResolveResponse`
- `resolve_user_link(platform, platform_user_id) -> ResolveResponse`
- `create_server_link_token(req) -> LinkTokenResponse`
- `create_user_link_token(req) -> LinkTokenResponse`
- `get_link_token_status(token) -> LinkTokenStatusResponse`
- `start_chat_turn(req) -> ChatTurnHandle` — resolves the owner, persists the user message, creates the stream-registry session, enqueues the turn; returns `(session_id, turn_id, user_id, subscribe_from="0-0")` so the caller subscribes directly to the per-turn Redis stream.

### New DB models
- `PlatformLink` — `(platform, platformServerId)` → owner's AutoGPT `userId`
- `PlatformUserLink` — `(platform, platformUserId)` → AutoGPT `userId` (for DMs)
- `PlatformLinkToken` — one-time token with `linkType` discriminator (SERVER | USER) and 30-min TTL

[...]

PR #12624, opened 2026-03-31
feat: Frontend platform linking page (/link/{token})

Bentlybro:
## Why
Completes the two linking flows from #12615 / #12618. When the bot sends a user a one-time `{frontend}/link/{token}?platform=...` URL, this page is where the user actually connects their AutoGPT account — whether that's claiming a server or linking their personal DMs.

## The flows

**Server link (`linkType: SERVER`)**
1. User runs `/setup` in a Discord guild → bot replies ephemerally with the link
2. Clicks link → logs in to AutoGPT if needed → lands here
3. Page shows: *"Set up AutoPilot for {ServerName}"* with a clear billing notice
4. Confirm → `POST /api/platform-linking/tokens/{token}/confirm`
5. Everyone in the server can now mention the bot; all usage bills to this user

**DM link (`linkType: USER`)**
1. User DMs the bot → bot creates a USER token and posts the link in the DM
2. Clicks link → logs in → lands here
3. Page shows: *"Link your {Platform} DMs"* with a personal-context billing notice
4. Confirm → `POST /api/platform-linking/user-tokens/{token}/confirm`
5. DMs now run as that user's own AutoPilot, billed to themselves

**Stacked on:** #12618 → #12615. Merge those first.

## Implementation
Single file: `autogpt_platform/frontend/src/app/(no-navbar)/link/[token]/page.tsx`

- State machine: `loading` → `not-authenticated` | `ready` → `linking` → `success` | `error`
- `fetchTokenInfo(token)` hits `GET /api/platform-linking/tokens/{token}/info` — no auth needed, returns `platform`, `server_name`, and `link_type`. The page branches all copy and the confirm endpoint choice on `link_type`.
- `?platform=DISCORD` query-param fallback so the platform name renders instantly even before `/info` resolves (removes a UI flash on slower connections).
- "Signed in as {email}" footer with a one-click "Not you? Sign out" that logs out and redirects back to `/login?next=/link/{token}` — handles the common "wrong account" case.
- 30s `AbortController` timeout on confirm. Timeouts surface as a retry prompt rather than hanging silently.
- Reuses `AuthCard`, `Button`, `Text`, `Link` from the design system; Phosphor icons only (no emojis); Tailwind only.

## States
| State | What the user sees |
|-------|--------------------|
| Loading | Spinner + "Loading…" |
| Not authenticated | "Sign in to continue" → `/login?next=/link/{token}` |
[...]

PR #12719, opened 2026-04-08
feat(frontend): platform link page — server CoPilot setup flow

Bentlybro:
## Why
The bot DMs users a link to `https://platform.agpt.co/link/{token}` when they need to set up CoPilot for their server. Without this page, the entire setup flow is dead — there's nowhere for the user to land and confirm.

## What
New page at `(no-navbar)/link/[token]` — the server setup confirmation flow:

- **`loading`** — waits for Supabase auth state
- **`not-authenticated`** — redirects to login with `?next=/link/{token}` so the user returns after signing in
- **`ready`** — shows server name (if available), what confirming means, and a clear billing callout ("usage from this server will be billed to your account")
- **`linking`** — spinner while the confirm request is in flight
- **`success`** — confirms the server is connected and everyone can start using CoPilot
- **`error`** — expired/used token with instructions to get a new link

## How
- Token info (server name, platform) is fetched from `/tokens/{token}/info` on load and shown in the UI; falls back gracefully if unavailable
- Confirm call goes through the existing `/api/proxy/` proxy which injects auth from the Supabase session cookie
- 30s AbortController on the confirm request
- No separate token status pre-check — validity is enforced server-side on confirm

## Stack
Stacked on `feat/platform-bot-linking` (PR #12615) — merge that first.


Bentlybro:
Closing — duplicate of #12624. The correct PR for this branch is #12624.

PR #12734, opened 2026-04-10
fix(copilot): route baseline Anthropic models directly instead of OpenRouter

majdyz:
## Summary
- Adds direct Anthropic API routing for Claude models in the baseline copilot path to avoid unnecessary OpenRouter charges ($1.59K unexpected bill)
- When `ANTHROPIC_API_KEY` is set, Claude models (`claude-*`) are sent directly to `https://api.anthropic.com/v1` instead of through OpenRouter
- Non-Anthropic models (title generation via `openai/gpt-4o-mini`, simulation via `google/gemini-2.5-flash`) continue routing through OpenRouter
- Falls back to OpenRouter for Anthropic models if `ANTHROPIC_API_KEY` is not configured

## Changes
- **config.py**: Added `anthropic_api_key` field with `ANTHROPIC_API_KEY` env var validator; updated default model names from OpenRouter format (`anthropic/claude-sonnet-4`) to direct Anthropic format (`claude-sonnet-4-20250514`)
- **service.py**: Added `_get_anthropic_client()` singleton (OpenAI-compatible client pointed at Anthropic API)
- **baseline/service.py**: Added `_is_anthropic_model()` and `_get_baseline_client(model)` to route requests to the correct API based on model name
- **Tests**: Added unit tests for config field, model detection, and client routing logic

## Test plan
- [ ] Verify existing baseline unit tests pass (`service_unit_test.py`)
- [ ] Verify new config tests pass (`config_test.py`)
- [ ] Deploy to staging with `ANTHROPIC_API_KEY` set and confirm Claude model requests go direct (check no OpenRouter charges)
- [ ] Deploy without `ANTHROPIC_API_KEY` and confirm fallback to OpenRouter still works
- [ ] Verify title generation (`openai/gpt-4o-mini`) and simulation (`google/gemini-2.5-flash`) still route through OpenRouter

### DIFF
--- a/autogpt_platform/backend/backend/copilot/baseline/service.py
+++ b/autogpt_platform/backend/backend/copilot/baseline/service.py
@@ -53,6 +53,7 @@ from backend.copilot.response_model import (
 )
 from backend.copilot.service import (
     _build_system_prompt,
+    _get_anthropic_client,
     _get_openai_client,
     _update_title_async,
     config,
@@ -83,6 +84,8 @@ from backend.util.tool_call_loop import (
 )
 
 if TYPE_CHECKING:
+    from langfuse.openai import AsyncOpenAI as LangfuseAsyncOpenAI
+
     from backend.copilot.permissions import CopilotPermissions
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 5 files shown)


majdyz:
Closing — based on further investigation, the platform should continue using OpenRouter rather than switching to direct Anthropic API. The cost spike investigation is still ongoing.

PR #12725, opened 2026-04-09
perf(backend/copilot): enable LLM prompt caching + harden user_context injection

majdyz:
### Why

LLM token costs are significant, especially for the copilot feature. The system prompt and tool definitions are the two largest static components of every request — caching them dramatically reduces input token costs (cache reads cost 10% of the base input price).

Previously, user-specific context (business understanding) was embedded directly in the system prompt, making it unique per user and preventing cache sharing across users or sessions. Every request paid full price for the system prompt even when the content was functionally identical.

A secondary security concern was identified during review: because the LLM is instructed to parse `<user_context>` blocks, a user could type a literal `<user_context>…</user_context>` tag in any message and potentially spoof or suppress their own personalisation context. This PR includes a full defence-in-depth fix for that injection vector on the first turn (including new users with no stored understanding), plus GET-endpoint stripping so injected context is never surfaced back to the client.

### What

- **`copilot/service.py`**: Added `USER_CONTEXT_TAG` constant (shared by writer and reader). Added `_USER_CONTEXT_ANYWHERE_RE` / `_USER_CONTEXT_PREFIX_RE` regexes, `format_user_context_prefix`, `strip_user_context_prefix`, `sanitize_user_supplied_context`, and `_sanitize_user_context_field` helpers. Replaced the old `_build_cacheable_system_prompt` / `_build_system_prompt` pair with a single `_build_system_prompt` that returns `(static_prompt, understanding)`. Added `inject_user_context` which sanitizes user input, optionally wraps trusted understanding, and persists the result to DB.
- **`copilot/sdk/service.py`**: On first turn calls `inject_user_context` before `_build_query_message` so the query sees the prefixed content. Passes `user_id if not has_history else None` to avoid redundant DB lookups on subsequent turns.
- **`copilot/baseline/service.py`**: Same pattern — `inject_user_context` called before transcript append and OpenAI message list construction; `openai_messages` loop patches the first user entry after injection.
- **`blocks/llm.py`**: System prompt sent as a structured block with `cache_control: {"type": "ephemeral"}`. `cache_control` placed on the last tool in the tool list. Guards against empty/whitespace-only system blocks (Anthropic rejects them). Fixed `anthropic.omit` → `anthropic.NOT_GIVEN` sentinel for the no-tools case.
[...]

Issue #789, opened 2023-04-11
Risk-avoiding continuous mode


jnt0rrente:
### Duplicates

- [X] I have searched the existing issues

### Summary 💡

When in risk-avoiding mode, one more GPT call should be made before running each command, asking it to moderate the would-be next command. Ideally, the call would return a value on a specific range we can compare with a user-defined risk threshold. If the risk exceeds the threshold, pause execution and await human feedback.

### Examples 🌈

The intermediate call can be prompted as something on the lines of 

> S: You are a moderation agent tasked with reviewing potential actions of an autonomous agent. You will be given a command to review, and you are to evaluate said command on a risk scale from 0 to 100, where 0 is <competely harmless thing> and 100 is <extinction event>.
> U: COMMAND: <POST call on sensitive API>

### Motivation 🔦

There is currently no way to attempt properly using AutoGPT without either babysitting or taking the risk of giving it free, unmonitored agency.

This feature would allow users to put some more trust on AutoGPT and "letting it loose" while trusting on its own self-moderation capabilities.

jnt0rrente:
It would obviously be necessary to test the responses' accuracy when evaluating risk and to tweak the prompt to something less improvised. Nevertheless, I believe this would be a good enough safeguard until a more complex system can be implemented.

richbeales:
It would also potentially double the cost of running AutoGPT

jnt0rrente:
Yes, it would increase costs, but not doubling at all since embeddings are not used here. Still, the idea is to have this as a third, optional mode.

jnt0rrente:
I am almost done implementing this. GPT-4 does a very good job analyzing commands. However, GPT-3.5 is lacking at best. Will post feedback in pull request.

dboitnot:
An alternative might be to allow the user to whitelist certain commands like `do_nothing` and `list_agents` and `google`.

Boostrix:
related: #2701 (safeguards) but also: https://github.com/Significant-Gravitas/Auto-GPT/issues/2987#issuecomment-1528954237 (preparing shell commands prior to executing them by adding relevant context like 1) availability of tools, 2) location/path, 3) version number)

Boostrix:
Related:
- #3713
- #3961

anonhostpi:
There's also quite a few other discussions playing with the idea of self-moderation:

[...]

PR #12741, opened 2026-04-11
chore(copilot): SDK CLI override + OpenRouter compat regression tests

majdyz:
## Why

We've been pinned at `claude-agent-sdk==0.1.45` (bundled CLI 2.1.63) since #12294 because every version above breaks against OpenRouter. Two stacked regressions:

1. **`tool_reference` content blocks** — CLI 2.1.69 (= SDK 0.1.46) added these to `tool_result.content`. OpenRouter's stricter Zod validation rejects them. Original break documented in #12294. CLI 2.1.70 added a proxy-detection workaround but our subsequent attempts at 0.1.55 / 0.1.56 still failed.
2. **`context-management-2025-06-27` beta header** — appears in some CLI version after 2.1.91. Tracked upstream at [anthropics/claude-agent-sdk-python#789](https://github.com/anthropics/claude-agent-sdk-python/issues/789), still open with no upstream PR. I added a +1 comment from us as a production user.

Cost of staying pinned: we're locked out of `exclude_dynamic_sections` (0.1.57 — directly amplifies #12725 prompt caching), `AssistantMessage.usage` per-turn token tracking (0.1.49 — relevant to the Sonnet cost-attribution work), `task_budget` (0.1.51), `get_context_usage()` (0.1.52), the MCP large-tool-result truncation fix (0.1.55), the MCP HTTP/SSE 50 MB/hr buffer leak fix (CLI 2.1.97), the 429 retry exponential backoff fix (CLI 2.1.97), and the `--resume` cache miss regression fix (CLI 2.1.90).

## What

This PR doesn't actually upgrade the SDK. It adds the infrastructure to upgrade *safely* when upstream lands a fix or when bisection identifies a known-good newer CLI version:

1. **`ChatConfig.claude_agent_cli_path`** (env: `CLAUDE_AGENT_CLI_PATH`) — threads through to `ClaudeAgentOptions(cli_path=...)` so we can decouple the Python SDK API surface from the CLI binary version. The SDK's `_find_cli` checks `cli_path` first, then the bundled binary. `_prewarm_cli` in `CoPilotExecutor` honours the same override. This is the workaround if a future SDK we want has a broken bundled CLI: install a known-good CLI separately and point at it.

2. **`test_bundled_cli_version_is_known_good_against_openrouter`** in `sdk_compat_test.py` — pins the bundled CLI to a known-good set (`{"2.1.63"}` today). Any `claude-agent-sdk` bump that changes the bundled CLI fails this test loudly with a pointer to #12294 and #789 instead of silently re-breaking production. Update the set explicitly when a new version is verified.

3. **`test_sdk_exposes_cli_path_option`** — forward-compat sentinel that fails fast if upstream removes the `cli_path` option we depend on for the override.

4. **`cli_openrouter_compat_test.py`** — the actual reproduction test. Spawns the bundled (or `CLAUDE_AGENT_CLI_PATH`-overridden) CLI against an in-process `aiohttp` server pretending to be the Anthropic Messages API, captures every request body the CLI sends, and asserts none of them contain `"type": "tool_reference"` content blocks or `context-management-2025-06-27` in either the body or the `anthropic-beta` header. The fake server returns a minimal valid streamed response so the CLI doesn't error out before we can inspect what it sent.

   - **No OpenRouter API key required** — the test reproduces the *mechanism* (forbidden patterns in the outgoing request) rather than the *symptom* (the 400 OpenRouter would return).
   - Skips gracefully if the CLI binary isn't available or fails to make any HTTP request, so it never falsely fails on environments where subprocess execution is locked down.

## How (the bisect workflow)

When upstream lands a fix or when we want to test an arbitrary candidate CLI version:

1. Bump `claude-agent-sdk` in `autogpt_platform/backend/pyproject.toml` (and `poetry lock`).
2. Push as a single commit.
3. Watch the CI runs of `test_bundled_cli_version_is_known_good_against_openrouter` (instant fail if the CLI version changed) and `test_cli_does_not_send_openrouter_incompatible_features` (real reproduction).
4. If both pass, add the new bundled CLI version to `_KNOWN_GOOD_BUNDLED_CLI_VERSIONS` and merge.
5. If they fail but a specific CLI version is known to work, install it separately and pass it via `CLAUDE_AGENT_CLI_PATH` instead of touching `pyproject.toml`.

## Test plan

- [x] `black` / `isort` / `ruff` / `pyright` clean on all 5 touched files
- [x] Pin assertion test verifies the current bundled CLI (2.1.63) is known good
- [x] CLI path option sentinel test verifies the SDK still exposes `cli_path`
- [ ] CI runs the new reproduction test against the current SDK and shows it passing (will land on first push)
- [ ] Optional follow-up: try bumping to 0.1.47 / 0.1.55 / 0.1.58 in successive commits to bisect via CI which versions actually break the reproduction test (no need to merge those — they're CI-driven probes)

## Related

- Significant-Gravitas/AutoGPT#12294 — the original pin to 0.1.45
- anthropics/claude-agent-sdk-python#789 — upstream issue for the newer regression (commented from our side)

### DIFF
--- a/autogpt_platform/backend/backend/copilot/config.py
+++ b/autogpt_platform/backend/backend/copilot/config.py
@@ -182,7 +182,9 @@ class ChatConfig(BaseSettings):
         "features (the bundled CLI version in 0.1.46+ is broken against "
         "OpenRouter — see PR #12294 and "
         "anthropics/claude-agent-sdk-python#789). Falls back to the "
-        "bundled binary when unset.",
+        "bundled binary when unset. Reads from `CHAT_CLAUDE_AGENT_CLI_PATH` "
+        "or the unprefixed `CLAUDE_AGENT_CLI_PATH` environment variable "
+        "(same pattern as `api_key` / `base_url`).",
     )
     use_openrouter: bool = Field(
         default=True,
@@ -304,6 +306,26 @@ class ChatConfig(BaseSettings):
                 v = os.getenv("OPENAI_BASE_URL")
             if not v:
                 v = OPENROUTER_BASE_URL
+        return v
+
+    @field_validator("claude_agent_cli_path", mode="before")
+    @classmethod
+    def get_claude_agent_cli_path(cls, v):
+        """Resolve the Claude Code CLI override path from environment.
+
+        Accepts either the Pydantic-prefixed ``CHAT_CLAUDE_AGENT_CLI_PATH``
+        or the unprefixed ``CLAUDE_AGENT_CLI_PATH`` (matching the same
+        fallback pattern used by ``api_key`` / ``base_url``). Keeping the
+        unprefixed form working is important because the field is
+        primarily an operator escape hatch set via container/host env,
+        and the unprefixed name is what the PR description, the field
+        docstrings, and the reproduction test in
+        ``cli_openrouter_compat_test.py`` refer to.
+        """
+        if not v:
+            v = os.getenv("CHAT_CLAUDE_AGENT_CLI_PATH")
+            if not v:
+                v = os.getenv("CLAUDE_AGENT_CLI_PATH")
         return v
 
     # Prompt paths for different contexts
--- a/autogpt_platform/backend/backend/copilot/config.py
+++ b/autogpt_platform/backend/backend/copilot/config.py
@@ -172,6 +172,18 @@ class ChatConfig(BaseSettings):
         description="Maximum number of retries for transient API errors "
         "(429, 5xx, ECONNRESET) before surfacing the error to the user.",
     )
+    claude_agent_cli_path: str | None = Field(
+        default=None,
+        description="Optional explicit path to a Claude Code CLI binary. "
+        "When set, the SDK uses this binary instead of the version bundled "
+        "with the installed `claude-agent-sdk` package — letting us pin "
+        "the Python SDK and the CLI independently. Critical for keeping "
+        "OpenRouter compatibility while still picking up newer SDK API "
+        "features (the bundled CLI version in 0.1.46+ is broken against "
+        "OpenRouter — see PR #12294 and "
+        "anthropics/claude-agent-sdk-python#789). Falls back to the "
+        "bundled binary when unset.",
+    )
     use_openrouter: bool = Field(
         default=True,
         description="Enable routing API calls through the OpenRouter proxy. "
--- a/autogpt_platform/backend/backend/copilot/sdk/cli_openrouter_compat_test.py
+++ b/autogpt_platform/backend/backend/copilot/sdk/cli_openrouter_compat_test.py
@@ -422,3 +422,101 @@ def test_subprocess_module_available():
     main reproduction test can spawn the CLI.  Catches sandboxed CI
     runners that block subprocess execution before the slow test runs."""
     assert subprocess.__name__ == "subprocess"
+
+
+# ---------------------------------------------------------------------------
+# Pure helper unit tests — pin the forbidden-pattern detection so any
+# future drift in the scanner is caught fast, even when the slow
+# end-to-end CLI subprocess test isn't runnable.
+# ---------------------------------------------------------------------------
+
+
+class TestScanRequestForForbiddenPatterns:
+    def test_clean_body_returns_empty_findings(self):
+        body = '{"model": "claude-opus-4.6", "messages": [{"role": "user", "content": "hi"}]}'
+        assert _scan_request_for_forbidden_patterns(body, {}) == []
+
+    def test_detects_tool_reference_in_body(self):
+        body = (
+            '{"messages": [{"role": "user", "content": ['
+            '{"type": "tool_reference", "tool_name": "find"}'
+            "]}]}"
+        )
+        findings = _scan_request_for_forbidden_patterns(body, {})
+        assert len(findings) == 1
+        assert "tool_reference" in findings[0]
+        assert "PR #12294" in findings[0]
+
+    def test_detects_context_management_in_body(self):
+        body = '{"betas": ["context-management-2025-06-27"]}'
+        findings = _scan_request_for_forbidden_patterns(body, {})
+        assert len(findings) == 1
+        assert "context-management-2025-06-27" in findings[0]
+        assert "#789" in findings[0]
+
+    def test_detects_context_management_in_anthropic_beta_header(self):
+        findings = _scan_request_for_forbidden_patterns(
+            body_text="{}",
+            headers={"anthropic-beta": "context-management-2025-06-27"},
+        )
+        assert len(findings) == 1
+        assert "anthropic-beta" in findings[0]
+
+    def test_detects_context_management_in_uppercase_header_name(self):
+        # HTTP header names are case-insensitive — make sure the
+        # scanner handles a server that didn't normalise names.
+        findings = _scan_request_for_forbidden_patterns(
+            body_text="{}",
+            headers={"Anthropic-Beta": "context-management-2025-06-27, other"},
+        )
+        assert len(findings) == 1
+
+    def test_ignores_unrelated_header_values(self):
+        findings = _scan_request_for_forbidden_patterns(
+            body_text="{}",
+            headers={
+                "authorization": "Bearer secret",
+                "anthropic-beta": "fine-grained-tool-streaming-2025",
+            },
+        )
+        assert findings == []
+
+    def test_detects_both_patterns_simultaneously(self):
+        body = (
+            '{"betas": ["context-management-2025-06-27"], '
+            '"messages": [{"role": "user", "content": ['
+            '{"type": "tool_reference", "tool_name": "find"}'
+            "]}]}"
+        )
+        findings = _scan_request_for_forbidden_patterns(body, {})
+        # Both patterns hit, in stable order: tool_reference then betas.
+        assert len(findings) == 2
+        assert "tool_reference" in findings[0]
+        assert "context-management-2025-06-27" in findings[1]
+
+
+class TestResolveCliPath:
+    def test_honours_explicit_env_var_when_file_exists(self, tmp_path, monkeypatch):
+        fake_cli = tmp_path / "fake-claude"
+        fake_cli.write_text("#!/bin/sh\necho fake\n")
+        fake_cli.chmod(0o755)
+        monkeypatch.setenv("CLAUDE_AGENT_CLI_PATH", str(fake_cli))
+        resolved = _resolve_cli_path()
+        assert resolved == fake_cli
+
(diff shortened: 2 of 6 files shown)


autogpt_platform/backend/backend/copilot/config.py
```diff
@@ -172,6 +172,18 @@ class ChatConfig(BaseSettings):
...
+        description="Optional explicit path to a Claude Code CLI binary. "
+        "When set, the SDK uses this binary instead of the version bundled "
+        "with the installed `claude-agent-sdk` package — letting us pin "
+        "the Python SDK and the CLI independently. Critical for keeping "
+        "OpenRouter compatibility while still picking up newer SDK API "
+        "features (the bundled CLI version in 0.1.46+ is broken against "
+        "OpenRouter — see PR #12294 and "
+        "anthropics/claude-agent-sdk-python#789). Falls back to the "
+        "bundled binary when unset.",
+    )
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [a6e306d28](https://github.com/Significant-Gravitas/AutoGPT/commit/a6e306d28a39f749af7f0807536bb56b497c5fb6): Added a `field_validator` for `claude_agent_cli_path` on `ChatConfig` that reads `CHAT_CLAUDE_AGENT_CLI_PATH` first and falls back to the unprefixed `CLAUDE_AGENT_CLI_PATH` — mirroring the existing pattern used by `api_key` and `base_url`. Operators can now set either env var name and the config layer will pick it up. The docstring was updated to call this out explicitly. The test helper `_resolve_cli_path` in `cli_openrouter_compat_test.py` was updated to the same two-name lookup, and a new test exercises the `CHAT_`-prefixed variant.

autogpt_platform/backend/backend/copilot/sdk/cli_openrouter_compat_test.py
```diff
@@ -0,0 +1,522 @@
...
+  tracker.
+
+This is the reproduction we use as a CI gate when bisecting which SDK /
+CLI version is safe to upgrade to.  It runs against the bundled CLI by
+default (or against ``ChatConfig.claude_agent_cli_path`` when set), so
+it doubles as a regression guard for the ``cli_path`` override
+mechanism.
+
+The test does **not** need an OpenRouter API key — it reproduces the
+mechanism (forbidden content blocks / headers in the *outgoing*
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [a6e306d28](https://github.com/Significant-Gravitas/AutoGPT/commit/a6e306d28a39f749af7f0807536bb56b497c5fb6): Same root cause as the other thread on this PR — the `ChatConfig` `env_prefix = "CHAT_"` made the field only pick up `CHAT_CLAUDE_AGENT_CLI_PATH`. Fixed at the config layer via a `field_validator` that accepts both the prefixed and unprefixed form (same pattern as `api_key` / `base_url`), updated the test helper `_resolve_cli_path` to mirror the same lookup (so the reproduction test is no longer bypassing the config model), updated the docstring and skip message, and added a dedicated unit test for the `CHAT_`-prefixed variant.

autogpt_platform/backend/backend/copilot/sdk/cli_openrouter_compat_test.py
```diff
@@ -0,0 +1,551 @@
...
+logger = logging.getLogger(__name__)
+
+
+# ---------------------------------------------------------------------------
+# Forbidden patterns we scan for in captured request bodies
+# ---------------------------------------------------------------------------
+
+# Substring of the `tool_reference` content block that breaks OpenRouter's
+# stricter Zod validation in tool_result.content. PR #12294 root-cause.
+_FORBIDDEN_TOOL_REFERENCE = '"type": "tool_reference"'
```
majdyz: 🤖 🟠 **Should Fix**: This substring match assumes a specific JSON-with-space formatting (`"type": "tool_reference"`). The Claude Code CLI is Node.js — `JSON.stringify` without an indent argument emits no whitespace, producing `{"type":"tool_reference"}` over the wire. The current literal would silently miss the real regression. Either parse the body via `json.loads` and walk the structure, or use a regex like `re.compile(r'"type"\s*:\s*"tool_reference"')`. Same applies to `_FORBIDDEN_CONTEXT_MANAGEMENT_BETA` — it's a token so substring works either way, but the tool_reference one is the load-bearing detection.

autogpt_platform/backend/backend/copilot/sdk/cli_openrouter_compat_test.py
```diff
@@ -0,0 +1,551 @@
...
+            proc.communicate(), timeout=timeout_seconds
+        )
+    except (asyncio.TimeoutError, TimeoutError):
+        # Best-effort kill — we already have whatever requests the CLI
+        # managed to send before stalling.
+        try:
+            proc.kill()
+        except ProcessLookupError:
+            pass
+        stdout_bytes, stderr_bytes = b"", b""
```
majdyz, line 363: 🤖 🟠 **Should Fix**: After `proc.kill()` on timeout, the subprocess is never reaped — `await proc.wait()` is missing. This leaves a zombie + open pipe FDs until the asyncio loop exits. Suggest:

```python
except (asyncio.TimeoutError, TimeoutError):
    try:
        proc.kill()
    except ProcessLookupError:
        pass
    try:
        await asyncio.wait_for(proc.wait(), timeout=5)
    except (asyncio.TimeoutError, TimeoutError):
        pass
    stdout_bytes, stderr_bytes = b"", b""
```

In CI loops where this test runs many times, leaked zombie processes can pile up and eventually exhaust the process table.

autogpt_platform/backend/backend/copilot/sdk/cli_openrouter_compat_test.py
```diff
@@ -0,0 +1,551 @@
...
+    def test_returns_none_when_env_var_points_to_missing_file(self, monkeypatch):
+        monkeypatch.delenv("CHAT_CLAUDE_AGENT_CLI_PATH", raising=False)
+        monkeypatch.setenv("CLAUDE_AGENT_CLI_PATH", "/nonexistent/path/to/claude")
+        # Should fall through to the bundled binary OR return None,
+        # but never raise.
+        resolved = _resolve_cli_path()
+        # We can't assert exact value (depends on whether the bundled
+        # CLI is installed in the test env) but the function must not
+        # raise — the caller is supposed to handle None gracefully.
+        assert resolved is None or resolved.is_file()
```
majdyz: 🤖 🟡 **Nice to Have**: Test name says "returns none" but the assertion `resolved is None or resolved.is_file()` is too permissive — it would also pass if the function silently fell through to the bundled binary, which would be a real bug worth catching here. Looking at `_resolve_cli_path`, when an override is set but the file is missing, it returns `None` unconditionally (it does *not* fall through to bundled). The assertion should be `assert resolved is None` to actually test what the name promises.

autogpt_platform/backend/backend/copilot/sdk/cli_openrouter_compat_test.py
```diff
@@ -0,0 +1,551 @@
...
+    # OAuth/profile endpoints the CLI may probe — answer 404 so it falls
+    # through quickly without retrying.
+    app.router.add_route("*", "/{tail:.*}", lambda _r: web.Response(status=404))
+
+    runner = web.AppRunner(app)
+    await runner.setup()
+    site = web.TCPSite(runner, "127.0.0.1", 0)
+    await site.start()
+
+    server = site._server
```
majdyz: 🤖 🔵 **Nit**: Reaching into `site._server` is a private aiohttp attribute and could break on aiohttp upgrades. The public API exposes the bound sockets via `site._server.sockets` indirectly, but there's a cleaner pattern: bind to a fixed port via `socket` first or use `site.name` (returns the URL with port). Not blocking — just brittle for a long-lived test.

autogpt_platform/backend/backend/copilot/sdk/cli_openrouter_compat_test.py
majdyz: 🤖 Fixed in [6dc0b6cff](https://github.com/Significant-Gravitas/AutoGPT/commit/6dc0b6cffd7ecaab3c4bf01b25f1a180d9b98cdf): Switched to a whitespace-tolerant regex `re.compile(r'"type"\s*:\s*"tool_reference"')`. Good catch — the Node.js CLI emits compact JSON without spaces, so the literal substring would have silently missed the regression. The context-management beta token has no surrounding JSON punctuation to vary, so the substring match stays.

autogpt_platform/backend/backend/copilot/sdk/cli_openrouter_compat_test.py
majdyz: 🤖 Fixed in [6dc0b6cff](https://github.com/Significant-Gravitas/AutoGPT/commit/6dc0b6cffd7ecaab3c4bf01b25f1a180d9b98cdf): Added `await asyncio.wait_for(proc.wait(), timeout=5.0)` after the kill so the process is reaped and the asyncio transport closes the stdout/stderr pipes. Wrapped in its own timeout so a stuck wait doesn't hang the test.

autogpt_platform/backend/backend/copilot/sdk/cli_openrouter_compat_test.py
majdyz: 🤖 Fixed in [6dc0b6cff](https://github.com/Significant-Gravitas/AutoGPT/commit/6dc0b6cffd7ecaab3c4bf01b25f1a180d9b98cdf): Tightened the assertion to `assert resolved is None`. The override is supposed to be a fail-loud escape hatch — silently falling back to the bundled binary would defeat its purpose. Also added a comment explaining the strict semantics.

autogpt_platform/backend/backend/copilot/sdk/cli_openrouter_compat_test.py
majdyz: 🤖 Fixed in [6dc0b6cff](https://github.com/Significant-Gravitas/AutoGPT/commit/6dc0b6cffd7ecaab3c4bf01b25f1a180d9b98cdf): Replaced the `site._server` private access with the public socket API — bind an ephemeral port via `socket.bind(("127.0.0.1", 0))`, read it back via `getsockname`, then pass the bound socket to `web.SockSite`. Also took the opportunity to convert the catch-all 404 lambda to an `async def` to silence the aiohttp deprecation warning.

majdyz:
Combined into #12747 which includes all commits from this PR plus the full SDK 0.1.58 upgrade with the compat proxy. Closing to reduce PR sprawl — all changes preserved in #12747.

PR #12749, opened 2026-04-11
fix(copilot): multi-layer defence against Write tool truncation errors

majdyz:
## Summary

- **Layer 1**: Disable the CLI built-in `Write` tool (added to `SDK_DISALLOWED_TOOLS`) so all file writes route through our MCP `write_file` / `write_workspace_file` tools where we control schema validation and error messages
- **Layer 2**: Detect partial truncation in `tool_adapter.py` — when `content`/`new_string` is present but `file_path` is missing, return actionable chunking guidance instead of letting the call fail with an opaque Ajv error
- **Layer 3**: Warn on large inline content (>50K chars) in E2B `write_file` — if the write succeeds but was dangerously close to the truncation threshold, suggest chunking for next time
- **Layer 4**: Improve the `workspace_files.py` truncation error message with step-by-step instructions (`cat >`, `cat >>`, `source_path`)
- **Layer 5**: Strengthen the system prompt guidance about large file writes to be more emphatic about the production failure mode

### Root cause
When the LLM generates a very large `content` argument for a Write tool call, the API output-token limit truncates the response mid-JSON. The CLI's built-in Ajv JSON Schema validator then rejects the truncated arguments with `'file_path' is a required property` — an opaque, unhelpful error. The user's work is lost and the LLM retries with the same approach, failing in a loop.

### Affected sessions
- `1a0ef47f-9711-41d2-91b8-df9dff9ecfc6`
- `2bb9fb0d-05bd-4194-8f3f-88fbe3d8b965`
- `8226d82e-bf9c-48e3-826f-80048d0fb7b4`

## Test plan
- [x] Unit tests for partial-truncation detection in `tool_adapter_test.py` (4 cases: write_file truncated, edit_file truncated, write_file with file_path passes through, non-write tool not affected)
- [x] Unit tests for improved workspace_files truncation error message (empty args, missing filename with content)
- [ ] Verify `ruff check` and `pyright` pass on all modified files
- [ ] Manual test: confirm large inline write_file calls return actionable guidance instead of opaque errors

### DIFF
--- a/autogpt_platform/backend/backend/copilot/prompting.py
+++ b/autogpt_platform/backend/backend/copilot/prompting.py
@@ -75,11 +75,12 @@ Example — committing an image file to GitHub:
 }}
 ```
 
-### Writing large files — CRITICAL
-**Never write an entire large document in a single tool call.**  When the
-content you want to write exceeds ~2000 words the tool call's output token
-limit will silently truncate the arguments, producing an empty `{{}}` input
-that fails repeatedly.
+### Writing large files — CRITICAL (causes production failures)
+**NEVER write an entire large document in a single tool call.**  When the
+content you want to write exceeds ~2000 words the API output-token limit
+will silently truncate the tool call arguments mid-JSON, losing all content
+and producing an opaque error.  This is unrecoverable — the user's work is
+lost and retrying with the same approach fails in an infinite loop.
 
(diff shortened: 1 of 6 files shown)


majdyz:
Superseded by #12750 which includes the unified file tools (Read/Write/Edit) with the schema-level truncation fix. All changes from this PR are covered by #12750.

PR #12753, opened 2026-04-12
fix(backend): auto-migrate webhook presets to new agent version on pu…

### LINKED ISSUE #11679: Agent trigger update to new agent version
## Problem

When updating an agent to a new version, webhook trigger URLs do not migrate to the new version. This creates a poor user experience because:

1. **New URL required for each version**: When an agent is updated (v1 → v2), triggering the old webhook URL still calls v1, not the updated v2
2. **External integrations break**: Users who have configured external services (e.g., Telegram bots, third-party webhooks) to call the agent's webhook URL must manually update the URL every time they publish a new agent version
3. **No way to preserve URLs**: Even removing and re-adding a webhook trigger generates a new URL

## Current Behavior

As reported in testing on dev:
1. Build agent v1, set up webhook trigger → webhook URL works ✅
2. Update agent to v2, trigger the same webhook → v1 is called (not v2) ❌
3. Creating a new trigger → generates a new URL
4. Removing original trigger and creating new one → also generates a new URL

## Expected Behavior

Users should be able to migrate/preserve webhook trigger URLs when updating their agent to a new version, so that:
- External integrations continue working without reconfiguration
- The same webhook URL can invoke the latest (or a specified) agent version

## Proposed Solution

1. **Automatic migration (default)**: When publishing a new agent version, automatically migrate compatible webhook triggers by updating the `AgentPreset` to reference the new graph version
2. **Manual migration**: For cases where auto-migration isn't possible or the user has opted out, provide a UI option to manually migrate an existing webhook trigger to a newer agent version

## Impact

This is particularly problematic for integrations like Telegram bots and other external services where changing the webhook URL requires reconfiguration on the external platform.

abderbejaoui:
When activating a new agent version, webhook trigger URLs remained pinned to the old version, forcing users to manually reconfigure external integrations (e.g. Telegram bots) on every update.

Root cause: `AgentPreset.agentGraphVersion` was set at creation time and never updated when a new graph version was activated.

Closes #11679

### Why

Webhook-attached presets (Telegram, Discord, etc.) silently kept hitting the previous version after every publish. Users had to re-register the webhook URL on the external side every time they activated a new version.

### What

- New helper `migrate_webhook_presets_to_new_version(user_id, graph_id, new_version)` in `library/db.py`.
- Invoked after activation in three paths, only when the activated graph contains a webhook input node:
  - `update_graph` (`PUT /graphs/{id}`) — only inside the `if new_graph_version.is_active:` branch.
  - `set_graph_active_version` (`PUT /graphs/{id}/versions/active`).
  - `update_graph_in_library` (library code path).
- Filter: `userId`, `agentGraphId`, `agentGraphVersion < new_version` (strictly older — presets manually pinned to a newer version are left alone), `webhookId IS NOT NULL`, `isDeleted = false`.
- Unit tests with a mocked Prisma client (patched at the usage site).

### How

Single bulk `update_many` per activation. No-op when there are no webhook-attached presets or the graph has no webhook trigger node.

### Changes 🏗️

- `autogpt_platform/backend/backend/api/features/library/db.py` — added `migrate_webhook_presets_to_new_version` and call site in `update_graph_in_library`.
- `autogpt_platform/backend/backend/api/features/v1.py` — call sites in `update_graph` and `set_graph_active_version`.
- `autogpt_platform/backend/test/test_migrate_webhook_presets.py` — unit tests.

### Test plan

1. Create an agent with a webhook trigger (e.g. Telegram).
2. Activate as v1 and register the webhook URL on the external side.
3. Publish v2 of the same graph and set it active.
4. Confirm:
   - The preset row's `agentGraphVersion` flipped to `2`.
   - Incoming webhook calls now execute v2.
   - Presets without a webhook are unchanged.
   - Presets manually pinned to v3 (newer) are unchanged.

### Checklist 📋

For code changes:
- [x] I have made a test plan
- [x] I have tested my changes according to the test plan (unit tests; end-to-end validation requested from reviewers)

For configuration changes:
- [x] `.env.default` unchanged — no config impact
- [x] `docker-compose.yml` unchanged — no infra impact


autogpt_platform/backend/backend/api/features/library/db.py
```diff
@@ -1825,6 +1834,50 @@ async def set_preset_webhook(
...
+    so that existing webhook URLs continue to trigger the latest agent version.
+
+    Only migrates presets that:
+    - Belong to the user
+    - Are attached to a webhook (webhookId is not null)
+    - Are not deleted
+    - Are for the given graph but pinned to an older version
+
+    Args:
+        user_id: The owner of the presets.
```
majdyz, line 1856: 🤖 🟠 **Should Fix**: Migration can run on presets that are already at the new version or at a *newer* version.

The filter `"agentGraphVersion": {"not": new_version}` correctly skips presets already at `new_version`, but it will also update presets pinned to a *higher* version (e.g., version 7 when publishing version 5 via a non-active path). In practice, a user can have manually pinned a preset to a future version or to a specific version for testing purposes.

A safer filter would be:
```python
"agentGraphVersion": {"lt": new_version}
```

This ensures only presets pinned to *older* versions are promoted, never newer ones.

autogpt_platform/backend/backend/api/features/v1.py
```diff
@@ -1044,6 +1044,15 @@ async def update_graph(
         if current_active_version:
             await on_graph_deactivate(current_active_version, user_id=user_id)
 
+        # Migrate webhook-attached presets to the new version so that
+        # existing webhook URLs continue to trigger the latest agent version.
+        if new_graph_version.webhook_input_node:
+            await library_db.migrate_webhook_presets_to_new_version(
+                user_id=user_id,
+                graph_id=graph_id,
+                new_version=new_graph_version.version,
```
majdyz, line 1053: 🤖 🟡 **Nice to Have**: Migration call in `update_graph` is only reachable when `new_graph_version.is_active` is true.

Looking at the `update_graph` handler in `v1.py`, the webhook migration is placed inside the `if new_graph_version.is_active:` block. This means publishing a new version that is *not* yet set as active does not trigger migration. That is probably the correct intent (you only want to migrate when the new version becomes the live one), but the docstring for `migrate_webhook_presets_to_new_version` says it handles the "publish" flow in general.

Please verify: if a user saves a draft version that has a webhook trigger, should existing webhook presets be migrated? If no, update the docstring to clarify. If yes, the call needs to be unconditional.

autogpt_platform/backend/test/test_migrate_webhook_presets.py
```diff
@@ -0,0 +1,70 @@
...
+
+from unittest.mock import AsyncMock, patch
+
+import pytest
+
+
+@pytest.fixture
+def mock_prisma():
+    with patch("prisma.models.AgentPreset.prisma") as mock:
+        mock_client = AsyncMock()
```
majdyz, line 14: 🤖 🟡 **Nice to Have**: Mock target patches `prisma.models.AgentPreset.prisma` (definition site) rather than the usage site.

Per repo convention (see copilot-instructions.md), tests should patch where the symbol is *used*, not where it is *defined*. The `migrate_webhook_presets_to_new_version` function in `library/db.py` imports `prisma` at the module level. The patch target should be:

```python
patch("backend.api.features.library.db.prisma.models.AgentPreset.prisma")
```

The current target (`"prisma.models.AgentPreset.prisma"`) may work by coincidence in this case, but it is fragile — if the prisma module is re-imported or aliased, the mock will not intercept the call.

majdyz (dismissed):
🤖 **REQUEST CHANGES** — PR #12753 (fix(backend): auto-migrate webhook presets to new agent version on publish)

Good contribution and a real pain point worth fixing. The core approach (bulk-update webhook-attached presets via Prisma on publish/activate) is sound and the user-scoping filter (`userId`, `agentGraphId`, `isDeleted: false`) is correct. Tests are present and cover the happy path.

**Blocker:**

**Version filter is too broad.** The `"agentGraphVersion": {"not": new_version}` condition will update presets that are already pinned to a *newer* version than the one being published (e.g., a preset at v7 would be downgraded to v5 if someone republishes v5). The docstring even says "pinned to an older version" but the filter does not enforce that. Change to:
```python
"agentGraphVersion": {"lt": new_version}
```

**Should fix:**

- The test's mock patches `"prisma.models.AgentPreset.prisma"` (definition site). Per repo convention, patch at the usage site: `"backend.api.features.library.db.prisma.models.AgentPreset.prisma"`. This is currently fragile.

**Nits / clarifications:**

- The `update_graph` call site is inside `if new_graph_version.is_active:` — migration only fires when the new version is published as active. This is probably correct behaviour, but the docstring says "new graph version is published" without that qualifier. Align the docstring with the actual guard.
- The PR description's Why/What/How section was left as template comments. Please fill it in — the template placeholder text is still visible.

Overall quality is good — once the version filter is corrected this should be ready to merge.

Pwuts (commented):
We're planning to add support for extra input blocks alongside trigger blocks soon. How would you suggest reconciling a change of the input schema with trying to auto-update a preset?

abderbejaoui:
Thanks @majdyz — all three addressed in f7e68df:

Filter switched to [{"lt": new_version}](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) (no more downgrading newer-pinned presets).
Mock now patches at the usage site.
Docstring clarified: migration runs on activation only.
PR description updated (filled Why/What/How + test plan; removed template placeholders).
Also merged latest dev to clear the unrelated lint failure.

@Pwuts re extra input blocks: I’d gate auto-migration on schema compat — migrate only if the trigger’s required shape is unchanged and no new required inputs were added; otherwise leave preset on the old version. Happy to do that as a follow-up once those blocks land.

Re-requesting review.

abderbejaoui:
Hi reviewers,  all requested changes from @majdyz  were addressed in f7e68df, and I’ve now also added additional test coverage to address the remaining Codecov patch coverage issue.

At this point the PR should be ready for another review when someone has time. Thanks again!

autogpt_platform/backend/backend/api/features/library/db.py
majdyz: Addressed in f7e68df — filter is now `{"lt": new_version}` (live at `library/db.py:1994`). New unit tests in e24b8c8e1 assert the filter on every call to `migrate_webhook_presets_to_new_version`. Resolving.

autogpt_platform/backend/backend/api/features/v1.py
majdyz: Intentional — migration only fires when the new version actually becomes the live one. Same semantics as `set_graph_active_version` below: if a non-active draft is published, no preset is migrated (webhook URLs still point at the previous active version). The docstring on `migrate_webhook_presets_to_new_version` already says "newly activated version", so the call site behaviour and the helper contract are aligned. Resolving.

autogpt_platform/backend/test/test_migrate_webhook_presets.py
majdyz: Reverted in e24b8c8e1 back to `prisma.models.AgentPreset.prisma` to match the project-wide convention (see `backend/api/features/library/db_test.py`, `backend/api/features/store/db_test.py`, and every other prisma-using test in the repo). The previous usage-site target failed because `backend.api.features.library.db` was not imported as an attribute of `library` until inside the test function, so `patch` could not traverse the dotted path at fixture-setup time — the suite was erroring at collection and contributing zero coverage. With the standard target the patch resolves cleanly and `migrate_webhook_presets_to_new_version` is now fully exercised. Resolving.

majdyz:
🤖 @Pwuts re: reconciling an input-schema change with auto-updating a preset —

Good question, and it's the reason this PR deliberately stays narrow. Two points:

1. **This migration only moves the version pointer, never the preset payload.** `migrate_webhook_presets_to_new_version` does a single `update_many` that sets `agentGraphVersion` and touches nothing else ([library/db.py L1990](https://github.com/Significant-Gravitas/AutoGPT/blob/e24b8c8e17cec76ad0e2e3da17d0668538e5f43f/autogpt_platform/backend/backend/api/features/library/db.py#L1990)). The preset's stored `inputs` are left as-is. So today, when the trigger/input schema is unchanged across versions (the common publish case), repinning is safe and is exactly what we want — the existing webhook URL keeps firing the live version.

2. **When extra input blocks land and the schema can diverge, the version bump should be gated on schema compatibility rather than applied unconditionally.** `GraphModel.input_schema` ([graph.py L220](https://github.com/Significant-Gravitas/AutoGPT/blob/e24b8c8e17cec76ad0e2e3da17d0668538e5f43f/autogpt_platform/backend/backend/data/graph.py#L220)) already exposes the per-version schema, so the natural extension is: compare the new version's input schema against the preset's stored inputs, and
   - migrate automatically when the new schema is backward-compatible (no new required fields without defaults, no removed/retyped fields the preset relies on), and
   - leave the preset pinned (and surface a "needs reconfiguration" state to the user) when it's a breaking change, so we never silently feed a stale payload into an incompatible schema.

I'd prefer to add that compatibility gate in the follow-up that introduces the extra-input-block support, since the breaking-vs-compatible classification only becomes meaningful once a schema actually can change between versions for the same trigger. Happy to scope it into this PR instead if you'd rather have the guard in place pre-emptively — let me know which you prefer.

majdyz:
## E2E Test Report

Native backend on `:8006` + infra Postgres. Tests exercise the **real production code path**: `graph.webhook_input_node` gate → `library_db.migrate_webhook_presets_to_new_version()`, plus a DB-backed integration test against real Prisma `update_many`.

| # | Scenario | Result | Evidence (before → after) |
|---|----------|--------|---------------------------|
| 1 | Webhook preset repins on publish (real graph v2 loaded from DB, `webhook_input_node`=Webhook(manual)) | PASS | preset `agentGraphVersion` v1 → v2; migrate() returned 1 |
| 2 | Non-webhook preset unaffected (control, `webhookId` NULL) | PASS | v1 → v1 (untouched) |
| 3 | Payload not mutated — only version pin changes | PASS | name/description/`webhookId` unchanged; only `agentGraphVersion` updated |
| 4 | Newer-pinned preset left alone (`lt` filter, pinned v3) | PASS | v3 → v3 (untouched) |
| 5 | Deleted preset skipped (`isDeleted=true`) | PASS | v1 → v1 (untouched) |
| 6 | **Cross-user isolation (NEGATIVE/SECURITY)** — other user's preset on same `graph_id` | PASS | owner publishes → other user's preset v1 → v1 (untouched) |
| 7 | Unit + integration suite | PASS | 13 passed (11 mock + 2 DB-backed) |

**Mutation check:** removing the `userId` scope from the helper makes the DB-backed test FAIL — confirming the cross-user assertion is non-vacuous.

### E2E Evidence

Before/after `agentGraphVersion` for every preset class through the real publish gate + migration helper. Confirms the webhook preset repins to the new version, the non-webhook control and newer/deleted presets are untouched, another user's preset on the same graph is never migrated, and the payload is preserved (only the version pin changes).

### Note on added coverage
The PR's existing tests mock Prisma and only assert the WHERE-clause shape — they cannot catch a semantically-wrong query. I added a DB-backed integration test (`test/test_migrate_webhook_presets_db.py`, pushed to this branch) that runs the real `update_many` against Postgres and asserts row-level outcomes including cross-user isolation and payload integrity.

majdyz (dismissed):
## E2E Test Evaluation — APPROVED

**Results:** 7/7 scenarios passed (incl. the full unit + DB-backed suite, 13 tests).

**Coverage:** Every behavior in the PR description was exercised against the real production code path (the `graph.webhook_input_node` gate computed from a real graph model loaded from DB → `migrate_webhook_presets_to_new_version`):
- Webhook preset repins to the new version on publish (v1 → v2).
- Non-webhook preset unaffected.
- Payload preserved — only `agentGraphVersion` changes.
- Newer-pinned preset (`lt` filter) and deleted preset both left alone.

**Negative / security tests:** Cross-user isolation verified at the DB level — when the owner publishes, another user's webhook preset on the *same* `graph_id` is never migrated. A mutation check (removing the `userId` scope) makes the DB-backed test FAIL, so the isolation assertion is non-vacuous.

**Evidence:** Before/after `agentGraphVersion` values logged for every preset class; the migration helper logged `Migrated 1 webhook preset(s)`.

**Note:** I pushed one test-only commit to this branch (`test/test_migrate_webhook_presets_db.py`) adding DB-backed coverage — the pre-existing tests mock Prisma and only assert the WHERE-clause shape, so they could not catch a semantically-wrong query. No production code was changed; the PR's logic is correct as-is.

No regressions observed.

majdyz (approved):
Re-approving after the post-test commits. The earlier E2E approval was auto-dismissed by new commits, which were test-only:

- `6b3fca4fee` added a DB-backed integration test, then `292c31926a` removed it — it flaked across the CI matrix (`Event loop is closed` on py3.11, passed py3.12/3.13) because the unit suite has no harness for real-DB CRUD (global Prisma client event-loop binding is nondeterministic). The mocked test (`test_migrate_webhook_presets.py`) still covers the query shape, and the real-DB behaviour — including cross-user isolation — was validated live during E2E testing against real Postgres (see the test report comment above).

No changes to the PR's product code. Required CI (`Check PR Status`) is green; the only red check is Vercel (non-required fork-preview auth limitation; this PR is backend-only).

autogpt_platform/backend/backend/api/features/library/db.py
Pwuts, line 1: What if the new graph version uses a different trigger block? It will migrate indiscriminately, resulting in a broken state.

kcze: Fixed in https://github.com/Significant-Gravitas/AutoGPT/pull/13394

PR #12759, opened 2026-04-13
fix(copilot): deduplicate SSE-replayed messages by content fingerprint

majdyz:
## Summary
- Fixes duplicate message content shown in CoPilot during SSE reconnections (page visibility change, network hiccups, wake-resync)
- The `resume_session_stream` backend always replays from `"0-0"` (beginning of Redis stream), and replayed `UIMessage` objects get new generated IDs from `useChat`, bypassing the old adjacent-only content dedup
- Extends `deduplicateMessages` to track all seen `role + preceding-user-context + content` fingerprints globally, catching replayed messages regardless of different IDs or position in the list
- Scopes fingerprints by preceding user message text to avoid false positives when the assistant legitimately gives the same answer to different prompts

## Test plan
- [ ] Verify new unit tests pass in CI (`helpers.test.ts` - 7 new dedup test cases)
- [ ] Manual: start a long tool-use session, switch tabs, return - no duplicate content
- [ ] Manual: refresh page during active session - content loads from DB without duplicates
- [ ] Manual: ask the same question twice in different turns - both answers preserved

### DIFF
--- a/autogpt_platform/frontend/src/app/(platform)/copilot/helpers.test.ts
+++ b/autogpt_platform/frontend/src/app/(platform)/copilot/helpers.test.ts
@@ -432,4 +432,37 @@ describe("deduplicateMessages", () => {
     const result = deduplicateMessages(msgs);
     expect(result).toHaveLength(2); // user + first tool call
   });
+
+  it("passes through assistant messages with empty parts without deduplicating them", () => {
+    // contentFingerprint === "[]" when parts is empty; the guard skips fingerprint
+    // tracking so these messages are never incorrectly deduplicated against each other.
+    const msgs: UIMessage[] = [
+      makeMsgWithId("u1", "user", "hello"),
+      { id: "a1", role: "assistant", parts: [] },
+      { id: "a2", role: "assistant", parts: [] },
+    ];
+    const result = deduplicateMessages(msgs);
+    expect(result).toHaveLength(3); // both empty-parts messages are kept
+  });
+
+  it("does not collapse structurally different no-text parts to the same fingerprint", () => {
+    // Parts lacking both 'text' and 'toolCallId' (e.g. step-start) previously
+    // all mapped to "" causing false-positive deduplication. Now JSON.stringify(p)
+    // is used as the fallback so distinct part shapes produce distinct fingerprints.
+    const msgs: UIMessage[] = [
+      makeMsgWithId("u1", "user", "hello"),
+      {
+        id: "a1",
+        role: "assistant",
+        parts: [{ type: "step-start" }],
+      },
+      {
+        id: "a2",
+        role: "assistant",
+        parts: [{ type: "step-start" }],
+      },
+    ];
+    const result = deduplicateMessages(msgs);
+    expect(result).toHaveLength(2); // duplicate step-start messages are deduped
+  });
 });
--- a/autogpt_platform/frontend/src/app/(platform)/copilot/helpers.test.ts
+++ b/autogpt_platform/frontend/src/app/(platform)/copilot/helpers.test.ts
@@ -374,6 +374,31 @@ describe("deduplicateMessages", () => {
     expect(deduplicateMessages(msgs)).toHaveLength(4);
   });
 
+  it("does not create false positives for text parts that contain the separator", () => {
+    // "a|b" + "c" and "a" + "b|c" previously collided when joined with "|"
+    const msgs: UIMessage[] = [
+      makeMsgWithId("u1", "user", "hello"),
+      {
+        id: "a1",
+        role: "assistant",
+        parts: [
+          { type: "text", text: "a|b" },
+          { type: "text", text: "c" },
+        ],
+      },
+      {
+        id: "a2",
+        role: "assistant",
+        parts: [
+          { type: "text", text: "a" },
+          { type: "text", text: "b|c" },
+        ],
+      },
+    ];
+    const result = deduplicateMessages(msgs);
+    expect(result).toHaveLength(3); // both assistant messages should be kept
+  });
+
   it("deduplicates by toolCallId for tool-call parts", () => {
     const msgs: UIMessage[] = [
       makeMsgWithId("u1", "user", "run tool"),
--- a/autogpt_platform/frontend/src/app/(platform)/copilot/helpers.test.ts
+++ b/autogpt_platform/frontend/src/app/(platform)/copilot/helpers.test.ts
@@ -336,6 +336,20 @@ describe("deduplicateMessages", () => {
     expect(result).toHaveLength(4);
   });
 
+  it("keeps second answer when same question is asked twice in one session", () => {
+    // Regression: scoping by user message TEXT instead of ID would treat both
+    // turns as the same context and drop the second identical assistant reply.
+    const msgs = [
+      makeMsgWithId("u1", "user", "What is 2+2?"),
+      makeMsgWithId("a1", "assistant", "4"),
+      makeMsgWithId("u2", "user", "What is 2+2?"), // same question, different ID
+      makeMsgWithId("a2", "assistant", "4"), // same answer — must be kept
+    ];
+    const result = deduplicateMessages(msgs);
+    expect(result).toHaveLength(4);
+    expect(result.map((m) => m.id)).toEqual(["u1", "a1", "u2", "a2"]);
+  });
+
   it("removes adjacent assistant duplicates", () => {
     const msgs = [
       makeMsgWithId("u1", "user", "hello"),
--- a/autogpt_platform/frontend/src/app/(platform)/copilot/helpers.test.ts
+++ b/autogpt_platform/frontend/src/app/(platform)/copilot/helpers.test.ts
@@ -382,11 +382,11 @@ describe("deduplicateMessages", () => {
         role: "assistant",
         parts: [
           {
-            type: "tool-invocation",
+            type: "dynamic-tool",
             toolCallId: "tc-1",
             toolName: "test",
-            args: {},
-            state: "call",
+            state: "input-available",
+            input: {},
           },
         ],
       },
@@ -395,11 +395,11 @@ describe("deduplicateMessages", () => {
         role: "assistant",
         parts: [
           {
-            type: "tool-invocation",
+            type: "dynamic-tool",
             toolCallId: "tc-1",
             toolName: "test",
-            args: {},
-            state: "call",
+            state: "input-available",
+            input: {},
           },
         ],
       },
--- a/autogpt_platform/frontend/src/app/(platform)/copilot/helpers.test.ts
+++ b/autogpt_platform/frontend/src/app/(platform)/copilot/helpers.test.ts
@@ -2,6 +2,7 @@ import type { UIMessage } from "ai";
 import { describe, expect, it } from "vitest";
 import {
   ORIGINAL_TITLE,
+  deduplicateMessages,
   extractSendMessageText,
   formatNotificationTitle,
   getSendSuppressionReason,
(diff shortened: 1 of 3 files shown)


autogpt_platform/frontend/src/app/(platform)/copilot/helpers.ts
```diff
@@ -180,13 +190,13 @@ export function deduplicateMessages(messages: UIMessage[]): UIMessage[] {
...
         .join("|");
 
-      // Only dedup if this assistant message is identical to the previous one
-      if (fingerprint && fingerprint === lastAssistantFingerprint) return false;
-      if (fingerprint) lastAssistantFingerprint = fingerprint;
-    } else {
-      // Reset on non-assistant messages so that identical assistant responses
-      // separated by a user message (e.g. "Done!" → user → "Done!") are kept.
-      lastAssistantFingerprint = "";
+      if (contentFingerprint) {
```
majdyz: 🤖 🟡 **Nice to Have**: The `contentFingerprint` for a streaming (in-progress) assistant message will differ from the final version since content is still being appended. This is currently safe because `setMessages` removes the last assistant message before `resumeStream()` at lines 482-486 and in the wake-resync handler at lines 368-372. However, if that removal logic is ever refactored away, streaming messages could bypass dedup. Consider adding a comment here noting this dependency, e.g.:

```ts
// NOTE: This relies on the caller removing the in-progress assistant
// message before resumeStream() — see useCopilotStream.ts lines ~482, ~368.
// A partial streaming message has a different fingerprint than its final
// version, so it wouldn't be caught by content dedup alone.
```

majdyz (commented):
## Review Summary

**Overall**: Clean, well-scoped fix. The change correctly addresses the root cause -- SSE replay from `"0-0"` produces messages with new IDs that bypass adjacent-only content dedup.

**Correctness**: The composite key `assistant:{lastUserText}:{contentFingerprint}` properly scopes dedup to the conversational context, preventing false positives when the assistant gives the same answer to different prompts. The `Set`-based global tracking catches replayed duplicates regardless of their position in the message list.

**Testing**: 7 new test cases cover the key scenarios: ID dedup, non-adjacent SSE replay dedup, same-answer-different-question preservation, adjacent dedup, empty list, unique passthrough, and toolCallId dedup. All 42 tests pass.

**One minor observation** (posted inline): The content fingerprint approach relies on `useCopilotStream` removing the in-progress assistant message before `resumeStream()`. This coupling is currently correct but worth documenting.

No blockers found.

autogpt_platform/frontend/src/app/(platform)/copilot/helpers.ts
```diff
@@ -180,13 +190,13 @@ export function deduplicateMessages(messages: UIMessage[]): UIMessage[] {
...
+        // Scope to the preceding user message so that identical assistant
+        // replies to *different* user prompts are preserved.
+        const contextKey = `assistant:${lastUserText}:${contentFingerprint}`;
```
majdyz: 🤖 🔴 **Blocker**: Same-question-twice silently drops the second answer.

The context key `assistant:${lastUserText}:${contentFingerprint}` scopes dedup by user-message **text**, not by user-message **identity**. When a user asks the exact same question twice and gets the same answer both times, `lastUserText` is identical for both turns, so the second assistant reply gets filtered out — it is a legitimate message, not a replay.

The fix is to track the **user message ID** instead of user text:

```ts
if (msg.role === "user") {
  lastUserMsgId = msg.id;  // id, not text
}
// ...
const contextKey = `assistant:${lastUserMsgId}:${contentFingerprint}`;
```

This is both correct and simpler — IDs are unique per message by construction, so there are no false positives. The current test suite doesn't cover this scenario; a test case like this would expose it:

```ts
it("keeps second answer when same question is asked twice", () => {
  const msgs = [
    makeMsgWithId("u1", "user", "What is 2+2?"),
    makeMsgWithId("a1", "assistant", "4"),
    makeMsgWithId("u2", "user", "What is 2+2?"),  // same question
    makeMsgWithId("a2", "assistant", "4"),           // same answer
  ];
  expect(deduplicateMessages(msgs)).toHaveLength(4);
});
```
With the current code this returns length 3, dropping `a2`.

autogpt_platform/frontend/src/app/(platform)/copilot/helpers.test.ts
```diff
@@ -291,3 +292,105 @@ describe("getSendSuppressionReason", () => {
...
+  it("keeps identical assistant replies to different user prompts", () => {
+    const msgs = [
+      makeMsgWithId("u1", "user", "What is 2+2?"),
+      makeMsgWithId("a1", "assistant", "4"),
+      makeMsgWithId("u2", "user", "What is 1+3?"),
+      makeMsgWithId("a2", "assistant", "4"),
+    ];
+    const result = deduplicateMessages(msgs);
+    expect(result).toHaveLength(4);
+  });
```
majdyz, line 337: 🤖 🟠 **Should Fix**: Missing test for same-question-twice scenario.

The test suite covers "identical answers to different questions are kept" (line 328) but not the inverse — "identical answers to the **same** question asked twice are also kept". This is the scenario the Blocker at line 196 exposes. Adding a regression test now would document intended behavior and prevent future regressions:

```ts
it("keeps second answer when same question is asked twice in one session", () => {
  const msgs = [
    makeMsgWithId("u1", "user", "What is 2+2?"),
    makeMsgWithId("a1", "assistant", "4"),
    makeMsgWithId("u2", "user", "What is 2+2?"),
    makeMsgWithId("a2", "assistant", "4"),
  ];
  expect(deduplicateMessages(msgs)).toHaveLength(4);
  expect(deduplicateMessages(msgs).map((m) => m.id)).toEqual(["u1", "a1", "u2", "a2"]);
});
```

autogpt_platform/frontend/src/app/(platform)/copilot/helpers.ts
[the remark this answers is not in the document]
majdyz: Fixed in [a17f05f2b](https://github.com/Significant-Gravitas/AutoGPT/commit/a17f05f2b1bafb70fdc8bed78efd7956ab8f08a2): switched the context key from user message text to user message ID. Two user messages with identical text now produce different context keys because their IDs differ, so the second assistant reply is preserved. Regression test added.

majdyz: Fixed in [a17f05f2b](https://github.com/Significant-Gravitas/AutoGPT/commit/a17f05f2b1bafb70fdc8bed78efd7956ab8f08a2): `lastUserText` replaced with `lastUserMsgId`. The context key is now `assistant:${lastUserMsgId}:${contentFingerprint}` — uniqueness comes from the message ID, not the text content, so identical questions asked at different turns get distinct scope keys.

autogpt_platform/frontend/src/app/(platform)/copilot/helpers.test.ts
majdyz: Fixed in [a17f05f2b](https://github.com/Significant-Gravitas/AutoGPT/commit/a17f05f2b1bafb70fdc8bed78efd7956ab8f08a2): added regression test `'keeps second answer when same question is asked twice in one session'` that asserts all 4 messages survive deduplication when the same question/answer pair appears in two distinct turns.

autogpt_platform/frontend/src/app/(platform)/copilot/helpers.ts
majdyz: The fix in [a17f05f2b](https://github.com/Significant-Gravitas/AutoGPT/commit/a17f05f2b1bafb70fdc8bed78efd7956ab8f08a2) removes the dependency on user text entirely by scoping fingerprints to user message IDs. The concern about streaming partial messages is still valid architecturally, but the comment text referenced `lastUserText` which no longer exists — resolving as superseded.

majdyz: Added the architectural note in [6d0fb28](https://github.com/Significant-Gravitas/AutoGPT/commit/6d0fb28d0d): the comment now documents that a partial streaming message has a different fingerprint from its final form and explains why dedup is still safe (caller removes the in-progress message before resumeStream()). The note also flags the risk if that removal is ever refactored away.

autogpt_platform/frontend/src/app/(platform)/copilot/helpers.ts
```diff
@@ -180,13 +192,19 @@ export function deduplicateMessages(messages: UIMessage[]): UIMessage[] {
...
+      if (contentFingerprint) {
+        // Scope to the preceding user message turn so that identical assistant
+        // replies to *different* user prompts are preserved.
+        // NOTE: A streaming (in-progress) assistant message has a partial
+        // fingerprint that differs from its final form, so it would not be
+        // caught by this dedup. This is safe because the caller removes the
+        // in-progress assistant message before calling resumeStream() — see
+        // useCopilotStream.ts. If that removal is ever refactored away,
+        // partial streaming messages could bypass dedup.
+        const contextKey = `assistant:${lastUserMsgId}:${contentFingerprint}`;
```
majdyz: 🤖 🟠 **Should Fix**: The code comment at line 198-203 states `"This is safe because the caller removes the in-progress assistant message before calling resumeStream()"`, but the `handleReconnect` path at `useCopilotStream.ts:150` calls `resumeStream()` directly WITHOUT stripping the last assistant message first.

Only two paths strip the last assistant message before `resumeStream()`:
- Wake-resync path (lines 368-373): strips last assistant if active backend stream
- `hasResumedRef` path (lines 482-487): strips last assistant before initial resume

The `handleReconnect` path (fired from `onFinish` on disconnect and `onError` on network error) does not. When a network drop fires `onError`, `rawMessages` still contains the partial in-progress assistant message. The replay from `0-0` will recreate the turn from scratch, and the partial message will stay stranded in `rawMessages` alongside the replayed full version.

The content fingerprint won't catch this because the partial message has a different fingerprint than the completed replayed version. Consider adding the same strip logic to `handleReconnect`:

```ts
// in handleReconnect, before the setTimeout
setMessages((prev) => {
  if (prev.length > 0 && prev[prev.length - 1].role === "assistant") {
    return prev.slice(0, -1);
  }
  return prev;
});
```

Or update the comment to accurately state which paths do/don't strip the message.

autogpt_platform/frontend/src/app/(platform)/copilot/helpers.ts
```diff
@@ -180,13 +192,19 @@ export function deduplicateMessages(messages: UIMessage[]): UIMessage[] {
         )
         .join("|");
```
majdyz: 🤖 🟡 **Nice to Have**: The `|` separator used in `contentFingerprint = parts.map(...).join('|')` and the `:` separator in `contextKey = `assistant:${lastUserMsgId}:${contentFingerprint}`` can both appear in real message text, creating fingerprint collisions.

For example, if text is `"a|b"` joined with text `"c"`, the fingerprint `"a|b|c"` is identical to joining `"a"` with `"b|c"`. Similarly, a `lastUserMsgId` containing `:` would collide with a different `contextKey`. In practice user message IDs are UUIDs which don't contain `:`, but text parts can contain arbitrary characters.

CodeRabbit already flagged this outside the diff range. A simple fix is to JSON-serialize the parts array for the fingerprint, or at minimum use a multi-char delimiter unlikely to appear in natural text:

```ts
const contentFingerprint = JSON.stringify(
  msg.parts.map((p) => ('text' in p && p.text) || ('toolCallId' in p && p.toolCallId) || '')
);
```

autogpt_platform/frontend/src/app/(platform)/copilot/helpers.ts
majdyz: Fixed in [d6b0c74a7](https://github.com/Significant-Gravitas/AutoGPT/commit/d6b0c74a7019b95b08c2eef7238b295b5150c4ba): added the same in-progress assistant message strip to `handleReconnect`'s `setTimeout` callback (matching the wake-resync and hydration-effect paths). Also updated the helpers.ts comment to accurately state that all three resumeStream callers now strip the partial message first.

autogpt_platform/frontend/src/app/(platform)/copilot/helpers.ts
majdyz: Fixed in [d6b0c74a7](https://github.com/Significant-Gravitas/AutoGPT/commit/d6b0c74a7019b95b08c2eef7238b295b5150c4ba): replaced `.join("|")\ with `JSON.stringify()` for the content fingerprint. This makes parts like `"a|b"` + `"c"` and `"a"` + `"b|c"` produce distinct fingerprints, eliminating the collision. Also added a regression test for this case.

autogpt_platform/frontend/src/app/(platform)/copilot/helpers.ts
```diff
@@ -154,39 +154,60 @@ export function shouldSuppressDuplicateSend(
...
-      const fingerprint = msg.parts
-        .map(
+      // JSON.stringify the parts array to avoid separator-collision false
+      // positives: a plain join("|") on ["a|b", "c"] and ["a", "b|c"]
+      // produces the same string. JSON encoding each element is unambiguous.
+      const contentFingerprint = JSON.stringify(
+        msg.parts.map(
           (p) =>
             ("text" in p && p.text) ||
             ("toolCallId" in p && p.toolCallId) ||
```
majdyz, line 196: 🤖 🟡 **Nice to Have**: The fingerprint extraction `("text" in p && p.text) || ("toolCallId" in p && p.toolCallId) || ""` maps any part whose extracted values are falsy to the empty string `""`. This means a text part with `text: ""` (empty text) and a structural `step-start` part produce identical fingerprint elements — so a message like `[{ type: "text", text: "" }]` gets the same fingerprint as `[{ type: "step-start" }]`. Today this is likely harmless since the copilot doesn't use `step-start` parts, but the extraction logic is under-specified. Consider using an object with a type discriminator for robustness: `{ type: p.type, key: ... }`.

autogpt_platform/frontend/src/app/(platform)/copilot/helpers.test.ts
```diff
@@ -291,3 +292,144 @@ describe("getSendSuppressionReason", () => {
...
+            state: "input-available",
+            input: {},
+          },
+        ],
+      },
+    ];
+    const result = deduplicateMessages(msgs);
+    expect(result).toHaveLength(2); // user + first tool call
+  });
+});
```
majdyz, line 468: 🤖 🔵 **Nit**: The `contentFingerprint !== "[]"` guard (line 198 of helpers.ts) is the only code path that skips fingerprint dedup for assistant messages. This branch is not tested — a test like `it("passes through assistant messages with empty parts array", () => { ... })` would confirm that two different `{ id: "a1", role: "assistant", parts: [] }` messages are NOT deduped by content (only by ID). Low priority since empty-parts assistant messages are abnormal, but it's worth documenting the intent.

majdyz (commented):
## Review Summary

**Overall**: LGTM with minor nits. The core dedup logic is correct and well-tested. Previous blockers (same-question-twice false positive, handleReconnect missing strip, separator collision) were all addressed in the iteration. Two new minor comments added:

1. 🟡 **Fingerprint map robustness** (helpers.ts:193): parts with falsy extracted values all map to `""`, which can conflate structurally different parts. Harmless today but brittle.
2. 🔵 **Untested branch** (helpers.test.ts:435): the `contentFingerprint !== "[]"` early-return path has no dedicated test.

No blockers remain. The `setMessages` + `resumeStream()` ordering in handleReconnect is safe — verified that `setMessages` in the AI SDK mutates the internal store synchronously (`chatRef.current.messages = ...`), so `resumeStream()` sees the stripped message list immediately.

autogpt_platform/frontend/src/app/(platform)/copilot/helpers.ts
majdyz: Fixed in [`abad038a21`](https://github.com/Significant-Gravitas/AutoGPT/commit/abad038a21): replaced the `""` fallback with `JSON.stringify(p)` so structurally different part types (e.g. `step-start`) produce distinct fingerprint elements and never collapse to the same empty string.

autogpt_platform/frontend/src/app/(platform)/copilot/helpers.test.ts
majdyz: Fixed in [`abad038a21`](https://github.com/Significant-Gravitas/AutoGPT/commit/abad038a21): added a test for the `parts: []` case that documents the empty-parts bypass, plus a second test verifying that structurally different no-text parts (e.g. `step-start`) are no longer incorrectly collapsed to the same fingerprint.

majdyz:

**Result: PASS**

After hard reload, session resumed without duplicating any messages. Message count matched before/after.

### Screenshots

autogpt_platform/frontend/src/app/(platform)/copilot/helpers.ts
```diff
@@ -154,39 +154,63 @@ export function shouldSuppressDuplicateSend(
...
-      // Only dedup if this assistant message is identical to the previous one
-      if (fingerprint && fingerprint === lastAssistantFingerprint) return false;
-      if (fingerprint) lastAssistantFingerprint = fingerprint;
-    } else {
-      // Reset on non-assistant messages so that identical assistant responses
-      // separated by a user message (e.g. "Done!" → user → "Done!") are kept.
-      lastAssistantFingerprint = "";
+            JSON.stringify(p),
+        ),
+      );
```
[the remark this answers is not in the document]
majdyz: 🤖 False positive. The AI SDK does not create separate UIMessages per tool state transition — state updates (e.g. `input-streaming → output-available`) are applied in-place within the same message object. Each unique tool invocation receives a unique `toolCallId`, so two messages sharing the same `toolCallId` are genuine SSE-replayed duplicates. Including `state` in the fingerprint would break deduplication: a replayed message with the same tool call but a later state (e.g. `output-available` after replay) would be treated as distinct and shown twice. No code change needed.

PR #12761, opened 2026-04-13
fix(copilot): store mode per session so indicator updates on switch

majdyz:
## Summary
- Hide the mode toggle button while streaming (instead of disabling it) to avoid confusing partial-toggle UI
- Remove localStorage mode persistence — mode is now transient in-memory state only (no stale overrides across sessions)
- The copilot mode indicator now correctly reflects the active session's mode because it reads from Zustand store which is updated on session switch

## Changes
- `ChatInput.tsx` — hide `<ModeToggleButton>` when `isStreaming` instead of passing `isStreaming` prop and showing a disabled button
- `ModeToggleButton.tsx` — remove `isStreaming` prop, disabled state, and streaming-specific tooltip
- `store.ts` — remove localStorage read/write for `copilotMode`; mode now defaults to `extended_thinking` and resets on page load
- `local-storage.ts` — keep `COPILOT_MODE` enum entry for backward compatibility; remove unused `COPILOT_SESSION_MODES`
- `store.test.ts` — update tests to assert mode is NOT persisted to localStorage
- `ChatInput.test.tsx` / `ModeToggleButton.stories.tsx` — update to match hide-not-disable behavior

## Test plan
- [x] Create a session in fast mode, create another in extended_thinking mode
- [x] Switch between sessions and verify the mode indicator updates correctly
- [x] Mode toggle is hidden (not disabled) while a response is streaming
- [x] Refreshing the page resets mode to extended_thinking (no stale localStorage override)


autogpt_platform/frontend/src/app/(platform)/copilot/useCopilotPage.ts
```diff
@@ -133,6 +135,8 @@ export function useCopilotPage() {
   // --- Send pending message after session creation ---
   useEffect(() => {
     if (!sessionId || pendingMessage === null) return;
+    // Record the mode that was active when this session was created
+    recordSessionMode(sessionId);
```
majdyz: 🤖 🟠 **Should Fix**: `recordSessionMode` is called here but is missing from the `useEffect` dependency array (`[sessionId, pendingMessage, sendMessage]` on line 172). This is a stale-closure risk — React's exhaustive-deps rule would flag this. Add `recordSessionMode` to the deps array.

autogpt_platform/frontend/src/app/(platform)/copilot/store.ts
```diff
@@ -284,6 +318,24 @@ export const useCopilotUIStore = create<CopilotUIState>((set) => ({
...
   },
 
+  sessionModes: getPersistedSessionModes(),
+  recordSessionMode: (sessionId) =>
+    set((state) => {
+      const next = new Map(state.sessionModes);
+      next.set(sessionId, state.copilotMode);
+      persistSessionModes(next);
+      return { sessionModes: next };
+    }),
```
majdyz: 🤖 🟠 **Should Fix**: The `sessionModes` map grows unboundedly — entries are never removed when a session is deleted. After deleting a session via `handleConfirmDelete`, the corresponding entry remains in the map and in `localStorage`. Over time this leaks storage proportional to the number of sessions ever created. Consider adding a `removeSessionMode(sessionId: string)` action and calling it alongside `deleteSessionMutation`.

autogpt_platform/frontend/src/app/(platform)/copilot/store.ts
```diff
@@ -284,6 +318,24 @@ export const useCopilotUIStore = create<CopilotUIState>((set) => ({
...
+  restoreSessionMode: (sessionId) =>
+    set((state) => {
+      const mode = state.sessionModes.get(sessionId);
+      if (mode && mode !== state.copilotMode) {
+        storage.set(Key.COPILOT_MODE, mode);
+        return { copilotMode: mode };
+      }
+      return state;
+    }),
```
majdyz: 🤖 🟡 **Nice to Have**: `restoreSessionMode` returns `state` (the entire previous state object reference) when no change is needed. In Zustand, returning the existing state object prevents a re-render, but returning the spread object (`{ ...state }`) would cause one. Returning `state` directly is actually correct here, but this is non-obvious — a brief comment explaining the intent (`// return same state reference to skip re-render`) would aid future readers.

autogpt_platform/frontend/src/app/(platform)/copilot/store.ts
```diff
@@ -79,6 +79,33 @@ function schedulePanelWidthPersist(width: number) {
...
 }
 
+function getPersistedSessionModes(): Map<string, CopilotMode> {
+  if (!isClient) return new Map();
+  try {
+    const raw = storage.get(Key.COPILOT_SESSION_MODES);
+    if (raw) {
+      const entries = JSON.parse(raw) as [string, CopilotMode][];
+      return new Map(entries);
+    }
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [1e7eadce2](https://github.com/Significant-Gravitas/AutoGPT/commit/1e7eadce2606b1e9d4675260925478841730a141): Added `isCopilotMode` type guard and updated `getPersistedSessionModes` to filter out any entries with unknown mode strings before constructing the Map, matching the suggested approach.

autogpt_platform/frontend/src/app/(platform)/copilot/useCopilotPage.ts
majdyz: 🤖 Fixed in [1e7eadce2](https://github.com/Significant-Gravitas/AutoGPT/commit/1e7eadce2606b1e9d4675260925478841730a141): Added `recordSessionMode` to the `useEffect` dependency array.

autogpt_platform/frontend/src/app/(platform)/copilot/store.ts
majdyz: 🤖 Fixed in [1e7eadce2](https://github.com/Significant-Gravitas/AutoGPT/commit/1e7eadce2606b1e9d4675260925478841730a141): Added `removeSessionMode` action to `store.ts` (interface + implementation) and wired it into the `onSuccess` callback of `deleteSessionMutation` in `useCopilotPage.ts`, so the entry is removed from the map and persisted storage whenever a session is deleted.

autogpt_platform/frontend/src/app/(platform)/copilot/store.ts
majdyz: 🤖 Fixed in [1e7eadce2](https://github.com/Significant-Gravitas/AutoGPT/commit/1e7eadce2606b1e9d4675260925478841730a141): Added a comment explaining the intent of returning the same state reference to skip unnecessary re-renders.

autogpt_platform/frontend/src/app/(platform)/copilot/__tests__/store.test.ts
```diff
@@ -176,17 +177,120 @@ describe("useCopilotUIStore", () => {
...
+        JSON.stringify([
+          ["session-valid", "fast"],
+          ["session-bad", "invalid_mode"],
+        ]),
+      );
+      // Force store re-read by calling the internal persistence path
+      useCopilotUIStore.getState().recordSessionMode("__probe__");
+      // The corrupt entry should never be readable as a valid CopilotMode
+      const state = useCopilotUIStore.getState();
+      expect(state.sessionModes.get("session-bad")).toBeUndefined();
```
majdyz: 🤖 🟠 **Should Fix**: The corruption test at line 253 does not actually exercise `getPersistedSessionModes()` deserialization. The sequence is: (1) `clearCopilotLocalData()` clears localStorage AND resets in-memory state to an empty map; (2) localStorage is re-written with corrupt entries; (3) `recordSessionMode('__probe__')` appends a new entry to the **already-reset in-memory map** — it never calls `getPersistedSessionModes()` again. The assertion on line 277 (`session-bad` is `undefined`) passes trivially because the corrupt data was never loaded back.

To actually test the guard, simulate a fresh store initialization from persisted state. Since `vi.resetModules()` doesn't work well with singleton Zustand stores in Vitest without full module teardown, a practical alternative is to directly call `getPersistedSessionModes` by re-importing or by exposing the function for testing. A minimal approach:

```ts
it('ignores invalid mode strings from corrupt localStorage', () => {
  window.localStorage.setItem(
    'copilot-session-modes',
    JSON.stringify([
      ['session-valid', 'fast'],
      ['session-bad', 'invalid_mode'],
      ['not-a-pair'],
      'garbage',
    ]),
  );
  // Directly init the store from localStorage by resetting via setState
  // using the same init expression as the store definition:
  const modes = new Map(
    (JSON.parse(window.localStorage.getItem('copilot-session-modes')!) as unknown[])
      .filter((e): e is [string, string] =>
        Array.isArray(e) && e.length === 2 && typeof e[0] === 'string' &&
        (e[1] === 'fast' || e[1] === 'extended_thinking'),
      ),
  );
  useCopilotUIStore.setState({ sessionModes: modes });
  expect(useCopilotUIStore.getState().sessionModes.get('session-bad')).toBeUndefined();
  expect(useCopilotUIStore.getState().sessionModes.get('session-valid')).toBe('fast');
});
```

Or better yet, export `getPersistedSessionModes` (with a `// @testable` comment) and call it directly in the test.

autogpt_platform/frontend/src/app/(platform)/copilot/__tests__/store.test.ts
```diff
@@ -176,17 +177,120 @@ describe("useCopilotUIStore", () => {
...
       useCopilotUIStore.getState().setNotificationsEnabled(true);
       useCopilotUIStore.getState().toggleSound();
       useCopilotUIStore.getState().addCompletedSession("s1");
+      useCopilotUIStore.getState().recordSessionMode("s1");
 
       useCopilotUIStore.getState().clearCopilotLocalData();
 
       const state = useCopilotUIStore.getState();
       expect(state.copilotMode).toBe("extended_thinking");
+      expect(state.sessionModes.size).toBe(0);
```
majdyz: 🤖 🔵 **Nit**: The `clearCopilotLocalData` test verifies that `state.sessionModes.size` is 0 after clearing, but doesn't assert that the `copilot-session-modes` localStorage key is also removed. Since `clearCopilotLocalData` explicitly calls `storage.clean(Key.COPILOT_SESSION_MODES)` and the other storage keys are all asserted, this entry should be verified for completeness:

```ts
expect(window.localStorage.getItem('copilot-session-modes')).toBeNull();
```

majdyz (commented):
## Review Summary

The feature implementation is solid. The Zustand store changes are well-structured, the type guard is correct, and the removeSessionMode/restoreSessionMode lifecycle is wired correctly across creation, switching, and deletion.

Previously flagged and confirmed fixed (commit 1e7eadce2): input validation guard, recordSessionMode in useEffect deps, removeSessionMode in deleteSessionMutation.onSuccess, state-reference comment.

Remaining:
- 🟠 Should Fix: Corruption test (line 277) passes trivially — does not exercise getPersistedSessionModes(). Flagged inline.
- 🔵 Nit: clearCopilotLocalData test missing localStorage assertion for copilot-session-modes key. Flagged inline.

No blockers. LGTM with the corruption test fixed.

autogpt_platform/frontend/src/app/(platform)/copilot/__tests__/store.test.ts
```diff
@@ -176,17 +177,120 @@ describe("useCopilotUIStore", () => {
...
+          ["session-valid", "fast"],
+          ["session-bad", "invalid_mode"],
+        ]),
+      );
+      // Force store re-read by calling the internal persistence path
+      useCopilotUIStore.getState().recordSessionMode("__probe__");
+      // The corrupt entry should never be readable as a valid CopilotMode
+      const state = useCopilotUIStore.getState();
+      expect(state.sessionModes.get("session-bad")).toBeUndefined();
+    });
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [24511b2a1](https://github.com/Significant-Gravitas/AutoGPT/commit/24511b2a11d3f72585ce6c73785d88d2293c7a91): Replaced the vacuous test with a proper re-parse — localStorage is now seeded with corrupt data, then `useCopilotUIStore.setState({ sessionModes: getPersistedSessionModes() })` forces the initialisation parser to run, asserting both that valid entries survive and that invalid-mode entries are dropped. `getPersistedSessionModes` is exported from `store.ts` to enable direct testing.

autogpt_platform/frontend/src/app/(platform)/copilot/__tests__/store.test.ts
majdyz: 🤖 Fixed in [24511b2a1](https://github.com/Significant-Gravitas/AutoGPT/commit/24511b2a11d3f72585ce6c73785d88d2293c7a91): Rewrote the test to actually exercise the parser — seeds localStorage with corrupt data, calls `useCopilotUIStore.setState({ sessionModes: getPersistedSessionModes() })` to simulate a fresh-page-load re-parse, then asserts the valid entry is present and the invalid entry is absent. `getPersistedSessionModes` is now exported from `store.ts`.

autogpt_platform/frontend/src/app/(platform)/copilot/__tests__/store.test.ts
majdyz: 🤖 Fixed in [24511b2a1](https://github.com/Significant-Gravitas/AutoGPT/commit/24511b2a11d3f72585ce6c73785d88d2293c7a91): Added `expect(window.localStorage.getItem('copilot-session-modes')).toBeNull()` to the `clearCopilotLocalData` test, consistent with all other key assertions in that block.

autogpt_platform/frontend/src/app/(platform)/copilot/store.ts
```diff
@@ -284,6 +332,33 @@ export const useCopilotUIStore = create<CopilotUIState>((set) => ({
...
+    set((state) => {
+      const next = new Map(state.sessionModes);
+      next.set(sessionId, state.copilotMode);
+      persistSessionModes(next);
+      return { sessionModes: next };
+    }),
+  restoreSessionMode: (sessionId) =>
+    set((state) => {
+      const mode = state.sessionModes.get(sessionId);
+      if (mode && mode !== state.copilotMode) {
```
majdyz: 🤖 🟡 **Nice to Have**: `restoreSessionMode` only calls `storage.set(Key.COPILOT_MODE, mode)` when `mode !== state.copilotMode` (line 347). If the in-memory `copilotMode` already matches the session's recorded mode, localStorage is not written. This is correct for the happy-path, but if `Key.COPILOT_MODE` is ever cleared from localStorage while the in-memory state is still valid (e.g., another tab cleared it), switching to a session with the same mode won't repair the localStorage entry. Since `setCopilotMode` always writes to localStorage and the page loads the value from storage on startup, this is an unlikely edge case — but if defensive repair is desired, unconditionally calling `storage.set(Key.COPILOT_MODE, mode)` when `mode` is defined would handle it without cost.

majdyz: Fixed in c1159d420d05906b6b13e365a430f35e549fc529 — `restoreSessionMode` now unconditionally calls `storage.set(Key.COPILOT_MODE, mode)` whenever `mode` is defined, before deciding whether to update in-memory state. This ensures localStorage is always repaired on session restore, even when the in-memory copilotMode already matches.

majdyz:

**Result: PASS**

Fast/Extended Thinking toggle visible and functional. Mode persisted across navigation away and back to same session. Toast confirms switch.

### Screenshots

autogpt_platform/frontend/src/app/(platform)/copilot/store.ts
```diff
@@ -275,12 +275,8 @@ export const useCopilotUIStore = create<CopilotUIState>((set) => ({
...
 
-  copilotMode:
-    isClient && storage.get(Key.COPILOT_MODE) === "fast"
-      ? "fast"
-      : "extended_thinking",
+  copilotMode: "extended_thinking",
   setCopilotMode: (mode) => {
-    storage.set(Key.COPILOT_MODE, mode);
     set({ copilotMode: mode });
   },
```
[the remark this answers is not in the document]
majdyz: 🤖 Not a bug — this is intentional behavior.

The mode is **per-session state** (stored on the server), not per-browser-tab. Removing localStorage was deliberate: `extended_thinking` is the default for new sessions, and the server now owns the source of truth for mode. If a user wants a different mode, they select it once and it persists for that session regardless of refresh. localStorage was removed to avoid the old stale-mode footgun where a client-side override disagreed with the server-recorded session mode.

PR #12762, opened 2026-04-13
fix(copilot): baseline cost tracking and cache token display

majdyz:
## Why
The baseline copilot path (OpenAI-compatible / OpenRouter) did not record any cost when the `x-total-cost` response header was absent, even though token counts were always available. The admin cost dashboard also lacked cache token columns.

## What
- **`x-total-cost` header extraction**: Reads the OpenRouter cost header per LLM call in the `finally` block (so cost is captured even when the stream errors mid-way). Accumulated across multi-round tool-calling turns.
- **Cache token extraction**: Extracts `prompt_tokens_details.cached_tokens` and `cache_creation_input_tokens` from streaming usage chunks and passes `cache_read_tokens`/`cache_creation_tokens` through to `persist_and_record_usage` for storage in `PlatformCostLog`.
- **Dashboard cache token display**: Adds cache read/write columns to the Raw Logs and By User tables on the admin platform costs dashboard. Adds `total_cache_read_tokens` and `total_cache_creation_tokens` to `UserCostSummary`.
- **No cost estimation**: When `x-total-cost` is absent, `cost_usd` is left as `None` and `persist_and_record_usage` records the entry under `tracking_type="tokens"`. Token-based cost estimation was removed — the platform dashboard already handles per-token cost display, and estimates would introduce inaccuracy in the reported figures.

## How
- In `_baseline_llm_caller`: extract the `x-total-cost` header in the `finally` block; accumulate to `state.cost_usd`.
- In `_BaselineStreamState`: add `turn_cache_read_tokens` / `turn_cache_creation_tokens` counters, populated from streaming usage chunks.
- In `persist_and_record_usage` / `record_cost_log`: pass through `cache_read_tokens` and `cache_creation_tokens` to `PlatformCostEntry`.
- Frontend: add `total_cache_read_tokens` / `total_cache_creation_tokens` fields to `UserCostSummary` and render them as columns in the cost dashboard.

## Test plan
- [x] Verify baseline copilot sessions log cost when `x-total-cost` header is present
- [x] Verify `cost_usd` stays `None` and token count is logged when header is absent
- [x] Verify cache tokens appear in the dashboard logs table for sessions using prompt caching
- [x] Verify the By User tab shows Cache Read and Cache Write columns
- [x] Unit tests: `test_cost_usd_extracted_from_response_header`, `test_cost_usd_remains_none_when_header_missing`, `test_cache_tokens_extracted_from_usage_details`


autogpt_platform/backend/backend/copilot/baseline/service.py
```diff
@@ -447,9 +501,34 @@ async def _baseline_llm_caller(
...
+                    got_header_cost = True
         except (AttributeError, ValueError):
             pass
 
+        # Fallback: estimate cost from token counts when x-total-cost is
+        # missing (e.g. some OpenRouter models don't report it).
+        # Use the delta for this call only — the state accumulators grow across
+        # all tool-call turns, so passing the cumulative total would
+        # compound-overestimate costs on the 2nd+ turn.
+        call_prompt_tokens = state.turn_prompt_tokens - prompt_tokens_before
```
majdyz: 🤖 🔴 **Blocker**: Multi-turn fallback cost over-estimation — fixed in this PR.

In multi-turn tool-call sessions, `turn_prompt_tokens` and `turn_completion_tokens` are accumulated with `+=` across all rounds. The original fallback cost estimation passed these **cumulative** totals to `_estimate_cost_from_tokens`, meaning on round N it charged for all tokens seen since the start of the session, not just round N's tokens.

Fix: snapshot token counts before entering the streaming loop and pass only the per-call delta (`state.turn_prompt_tokens - prompt_tokens_before`) to the estimator. A new regression test `test_multiturn_fallback_cost_uses_per_call_delta` covers this.

autogpt_platform/backend/backend/copilot/baseline/service.py
```diff
@@ -381,10 +415,29 @@ async def _baseline_llm_caller(
...
+                # Extract cache token details when available (OpenAI /
+                # OpenRouter include these in prompt_tokens_details).
+                ptd = getattr(chunk.usage, "prompt_tokens_details", None)
+                if ptd:
+                    state.turn_cache_read_tokens += (
+                        getattr(ptd, "cached_tokens", 0) or 0
+                    )
+                    # cache_creation_input_tokens is reported by some providers
+                    # (e.g. Anthropic native) but not standard OpenAI streaming.
+                    state.turn_cache_creation_tokens += (
```
majdyz, line 309: 🤖 🔴 **Blocker**: `turn_cache_creation_tokens` was never populated — fixed in this PR.

The field was initialized to 0 and wired through to `persist_and_record_usage` and the dashboard, but no code ever wrote to it. Cache write metrics would always display as zero on the dashboard.

Fix: extract `cache_creation_input_tokens` from `prompt_tokens_details` alongside the existing `cached_tokens` extraction. This field is reported by Anthropic-backed models via OpenRouter. A new test `test_cache_creation_tokens_extracted_from_usage_details` covers this.

autogpt_platform/backend/backend/copilot/baseline/service.py
```diff
@@ -447,9 +501,34 @@ async def _baseline_llm_caller(
...
 
+        # Fallback: estimate cost from token counts when x-total-cost is
+        # missing (e.g. some OpenRouter models don't report it).
+        # Use the delta for this call only — the state accumulators grow across
+        # all tool-call turns, so passing the cumulative total would
+        # compound-overestimate costs on the 2nd+ turn.
+        call_prompt_tokens = state.turn_prompt_tokens - prompt_tokens_before
+        call_completion_tokens = state.turn_completion_tokens - completion_tokens_before
+        if not got_header_cost and (
+            call_prompt_tokens > 0 or call_completion_tokens > 0
```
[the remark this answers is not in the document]
majdyz: Fixed in 483f1cfb. Moved `prompt_tokens_before` and `completion_tokens_before` initialization to before the `try` block, so the `finally` block can always reference them safely — even when `client.chat.completions.create()` raises a network error or timeout. The existing test `test_no_cost_when_api_call_raises_before_stream` covers this case.

autogpt_platform/backend/backend/copilot/baseline/service.py
```diff
@@ -381,10 +415,29 @@ async def _baseline_llm_caller(
             )
         tool_calls_by_index: dict[int, dict[str, str]] = {}
 
+        # Snapshot token counts before this call so we can compute the delta
+        # used for fallback cost estimation (state fields are accumulated across
+        # all tool-call turns, so we must not pass the cumulative total to the
+        # per-call cost estimator).
+        prompt_tokens_before = state.turn_prompt_tokens
+        completion_tokens_before = state.turn_completion_tokens
```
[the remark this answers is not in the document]
majdyz: Fixed in 483f1cfb. Moved `prompt_tokens_before` and `completion_tokens_before` to before the `try` block so the `finally` block always has them defined, even when the API call raises before any streaming begins. The existing test `test_no_cost_when_api_call_raises_before_stream` verifies this.

autogpt_platform/backend/backend/copilot/baseline/service.py
```diff
@@ -447,9 +499,34 @@ async def _baseline_llm_caller(
...
+        # compound-overestimate costs on the 2nd+ turn.
+        call_prompt_tokens = state.turn_prompt_tokens - prompt_tokens_before
+        call_completion_tokens = state.turn_completion_tokens - completion_tokens_before
+        if not got_header_cost and (
+            call_prompt_tokens > 0 or call_completion_tokens > 0
+        ):
+            estimated = _estimate_cost_from_tokens(
+                state.model,
+                call_prompt_tokens,
+                call_completion_tokens,
```
majdyz: 🤖 🟠 **Should Fix**: Fallback cost estimator uses full `call_prompt_tokens` (which includes cached-read tokens), but cache reads are billed at a fraction of the input rate (e.g. Anthropic: 10%, OpenAI: ~50%). This overstates cost on cache-heavy turns.

The fix is to snapshot the cache counters before the call (parallel to how `prompt_tokens_before` is snapshotted) and subtract the delta from the prompt delta:

```python
# in _baseline_llm_caller, after the try block, alongside other snapshots:
cache_read_before = state.turn_cache_read_tokens

# then in the fallback block:
call_cache_read_tokens = state.turn_cache_read_tokens - cache_read_before
uncached_prompt = max(0, call_prompt_tokens - call_cache_read_tokens)
estimated = _estimate_cost_from_tokens(state.model, uncached_prompt, call_completion_tokens)
```

Note: CodeRabbit already flagged this, and a fuller fix would also price cache reads at their actual discounted rate rather than ignoring them entirely.

autogpt_platform/backend/backend/copilot/baseline/service.py
```diff
@@ -102,6 +102,38 @@
...
+    "anthropic/claude-sonnet-4": (3.0 / 1_000_000, 15.0 / 1_000_000),
+    "anthropic/claude-3.5-sonnet": (3.0 / 1_000_000, 15.0 / 1_000_000),
+    "anthropic/claude-3-haiku": (0.25 / 1_000_000, 1.25 / 1_000_000),
+    "anthropic/claude-3-opus": (15.0 / 1_000_000, 75.0 / 1_000_000),
+    "openai/gpt-4o": (2.5 / 1_000_000, 10.0 / 1_000_000),
+    "openai/gpt-4o-mini": (0.15 / 1_000_000, 0.6 / 1_000_000),
+    "openai/gpt-4-turbo": (10.0 / 1_000_000, 30.0 / 1_000_000),
+    "google/gemini-2.0-flash-001": (0.1 / 1_000_000, 0.4 / 1_000_000),
+    "google/gemini-2.5-pro-preview": (1.25 / 1_000_000, 10.0 / 1_000_000),
+}
```
majdyz: 🤖 🟠 **Should Fix**: The pricing table is missing the default model `anthropic/claude-opus-4.6` (from `config.model`). When a non-fast baseline session completes without the `x-total-cost` header, `_estimate_cost_from_tokens` returns `None` and cost stays `None` — the exact problem this PR is meant to fix.

Models missing from `_OPENROUTER_MODEL_PRICING`:
- `anthropic/claude-opus-4.6` (default model, ~$15/$75 per 1M)
- `google/gemini-2.5-flash` (simulation model)
- `openai/gpt-4o-mini` (title model — but it hits the separate title path, lower priority)

At minimum, the default and fast models should always have entries here.

autogpt_platform/backend/backend/copilot/baseline/service.py
```diff
@@ -1291,14 +1369,21 @@ async def stream_chat_completion_baseline(
...
-        # baseline path may be slightly over-counted vs the SDK path.
+        # When prompt_tokens_details.cached_tokens is reported, subtract
+        # them from prompt_tokens to get the uncached count so the cost
+        # breakdown stays accurate.
+        uncached_prompt = state.turn_prompt_tokens
+        if state.turn_cache_read_tokens > 0:
+            uncached_prompt = max(
+                0, state.turn_prompt_tokens - state.turn_cache_read_tokens
+            )
         await persist_and_record_usage(
```
majdyz, line 1217: 🤖 🟡 **Nice to Have**: The tiktoken-based token estimation at lines 1353-1369 fires when the provider reports no usage, then `persist_and_record_usage` is called with `cost_usd=state.cost_usd` (which may still be `None` since the fallback in `_baseline_llm_caller` only runs when tokens are already known at stream time, not during the tiktoken backfill path).

Concretely: a provider that streams no `chunk.usage` at all will:
1. leave `state.cost_usd = None`
2. backfill tokens via tiktoken
3. call `persist_and_record_usage(cost_usd=None)`

The fallback estimator never had a chance to run with the tiktoken-estimated counts because those are set after `_baseline_llm_caller` returns. Consider calling `_estimate_cost_from_tokens` here too when `state.cost_usd is None` after the tiktoken backfill, using `state.turn_prompt_tokens` / `state.turn_completion_tokens`.

autogpt_platform/backend/backend/copilot/baseline/service_unit_test.py
```diff
@@ -828,3 +846,209 @@ async def test_no_cost_when_api_call_raises_before_stream(self):
...
+                messages=[{"role": "user", "content": "hi"}],
+                tools=[],
+                state=state,
+            )
+            await _baseline_llm_caller(
+                messages=[{"role": "user", "content": "follow up"}],
+                tools=[],
+                state=state,
+            )
+
```
majdyz, line 959: 🤖 🔵 **Nit**: The test for `test_multiturn_fallback_cost_uses_per_call_delta` calls `_baseline_llm_caller` twice on the same `state` object, which is correct. However, the token counts reported by each chunk (`prompt_tokens` and `completion_tokens`) are **additive per chunk** but are treated as session totals by OpenAI streaming. The mock has `mock_chunk.usage.prompt_tokens = 1000` for turn 1 and `= 1100` for turn 2, but the comment says "accumulators now hold 1000+1100". This only holds if those are per-chunk deltas (which OpenAI streaming reports), not session totals. The test is correct for the expected behavior, but the comment in the test assertion is slightly misleading — it says `1100` is the turn 2 value but then asserts `state.turn_prompt_tokens == 2100` (1000+1100), which implies the chunk values are deltas. Worth clarifying the comment to say "turn 2 reports 1100 *additional* tokens".

majdyz (commented):
🤖 **PR Review Summary — fix(copilot): baseline cost tracking fallback and cache token display**

## Overview

Good PR with clear motivation (Why/What/How covered in the description). The three critical bugs that were identified and fixed — multi-turn cost over-estimation, `turn_cache_creation_tokens` never populated, and `UnboundLocalError` in the finally block — are all well-handled and tested. Test coverage is comprehensive.

## Open Issues

Two **Should Fix** items remain unaddressed:

**🟠 Missing default model in pricing table** (commented at line 118)  
`anthropic/claude-opus-4.6` (the `config.model` default) is absent from `_OPENROUTER_MODEL_PRICING`. Any non-fast baseline session that gets no `x-total-cost` header will fall through to `cost_usd=None` — exactly the gap this PR is supposed to close. This needs to be added before the fallback is truly useful.

**🟠 Cache-read token cost overestimation in fallback path** (commented at line 519)  
`call_prompt_tokens` includes cached-read tokens, but cache reads are billed at a discount (~10% for Anthropic, ~50% for OpenAI). Passing the full prompt count to `_estimate_cost_from_tokens` overstates cost on cache-heavy turns. To fix: snapshot `state.turn_cache_read_tokens` before the call (alongside `prompt_tokens_before`), compute the cache delta, and use separate input/cache-read rates in the estimator.

## What's Good

- Per-call delta approach for fallback estimation is correct and the multi-turn test proves it.
- `getattr(ptd, ..., 0) or 0` double-defense against `None` is solid.
- Cache token display in the dashboard is clean — the `colSpan` increments are correct.
- `UserCostSummary` fields have sensible `default=0` so existing API clients won't break.
- `UnboundLocalError` fix (snapshots before `try`) is the right approach.

autogpt_platform/backend/backend/copilot/baseline/service.py
majdyz: Fixed in [d84417dbc](https://github.com/Significant-Gravitas/AutoGPT/commit/d84417dbc8fdcd1832d74e1d00576a396ea6cffc): Added `anthropic/claude-opus-4.6` to `_OPENROUTER_MODEL_PRICING` at $15/$75 per 1M tokens (input/output), matching claude-3-opus pricing on OpenRouter. Non-fast sessions without `x-total-cost` header now get a cost estimate instead of `cost_usd=None`.

autogpt_platform/backend/backend/copilot/baseline/service.py
majdyz: Fixed in [d84417dbc](https://github.com/Significant-Gravitas/AutoGPT/commit/d84417dbc8fdcd1832d74e1d00576a396ea6cffc): Updated `_estimate_cost_from_tokens` to accept `cache_read_tokens` and apply provider-specific discounts (Anthropic: 10%, OpenAI: 50%) via the new `_CACHE_READ_DISCOUNT` table. The fallback cost estimator now snapshots `cache_read_tokens_before` alongside prompt/completion counts and passes the per-call delta so cache-heavy multi-turn sessions aren't overcharged.

autogpt_platform/backend/backend/copilot/baseline/service.py
```diff
@@ -447,9 +499,34 @@ async def _baseline_llm_caller(
...
+        if not got_header_cost and (
+            call_prompt_tokens > 0 or call_completion_tokens > 0
+        ):
+            estimated = _estimate_cost_from_tokens(
+                state.model,
+                call_prompt_tokens,
+                call_completion_tokens,
+            )
+            if estimated is not None:
+                state.cost_usd = (state.cost_usd or 0.0) + estimated
```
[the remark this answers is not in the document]
majdyz: Fixed in [d84417dbc](https://github.com/Significant-Gravitas/AutoGPT/commit/d84417dbc8fdcd1832d74e1d00576a396ea6cffc): Added `_CACHE_READ_DISCOUNT` table and updated `_estimate_cost_from_tokens` to bill cache-read tokens at the discounted rate (Anthropic: 10%, OpenAI: 50%). Per-call cache-read delta is now passed from the snapshot so multi-turn sessions are priced correctly.

autogpt_platform/backend/backend/copilot/baseline/service.py
majdyz: Fixed in [d84417dbc](https://github.com/Significant-Gravitas/AutoGPT/commit/d84417dbc8fdcd1832d74e1d00576a396ea6cffc): After the tiktoken fallback fills in token counts, we now call `_estimate_cost_from_tokens` (guarded by `state.cost_usd is None`) so `persist_and_record_usage` never receives `cost_usd=None` when the provider reported no streaming usage and the model is in our pricing table.

autogpt_platform/backend/backend/copilot/baseline/service_unit_test.py
majdyz: Fixed in [d84417dbc](https://github.com/Significant-Gravitas/AutoGPT/commit/d84417dbc8fdcd1832d74e1d00576a396ea6cffc): Updated the comment to clarify that each mock stream yields a single chunk, making the additive total equal to the per-call value — so the test is accurate. The mock correctly represents one usage chunk per API call (as OpenAI streaming sends totals in the final chunk).

autogpt_platform/backend/backend/copilot/baseline/service.py
```diff
@@ -447,9 +489,29 @@ async def _baseline_llm_caller(
...
+        if not got_header_cost and (
+            state.turn_prompt_tokens > 0 or state.turn_completion_tokens > 0
+        ):
+            estimated = _estimate_cost_from_tokens(
+                state.model,
+                state.turn_prompt_tokens,
+                state.turn_completion_tokens,
+            )
+            if estimated is not None:
+                state.cost_usd = (state.cost_usd or 0.0) + estimated
```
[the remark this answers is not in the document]
majdyz: The per-call delta fix (snapshotting before each API call) was in a prior commit; this commit extends it to also discount cache-read tokens at the provider-specific rate. [d84417dbc](https://github.com/Significant-Gravitas/AutoGPT/commit/d84417dbc8fdcd1832d74e1d00576a396ea6cffc)

autogpt_platform/backend/backend/copilot/baseline/service.py
```diff
@@ -338,6 +370,8 @@ class _BaselineStreamState:
     text_started: bool = False
     turn_prompt_tokens: int = 0
     turn_completion_tokens: int = 0
+    turn_cache_read_tokens: int = 0
```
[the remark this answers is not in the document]
majdyz: Fixed in a prior commit — `cache_creation_input_tokens` is now extracted from `prompt_tokens_details` alongside `cached_tokens`. [d84417dbc](https://github.com/Significant-Gravitas/AutoGPT/commit/d84417dbc8fdcd1832d74e1d00576a396ea6cffc) further improves the fallback cost estimator to apply cache-read discounts correctly.

autogpt_platform/backend/backend/copilot/baseline/service.py
```diff
@@ -102,6 +102,72 @@
...
+
+    ``prompt_tokens`` should be the *total* prompt token count as reported by
+    the API (includes both regular and cache-read tokens as a subset).
+    ``cache_read_tokens`` is used to apply a provider-specific discount:
+
+    * Cache reads are billed at a fraction of the normal input rate
+      (Anthropic: 10 %, OpenAI: 50 %).  The regular (non-cached) portion is
+      billed at full price.
+    * Cache writes (creation) are already included in ``prompt_tokens`` at the
+      full input rate — no adjustment is needed.
```
majdyz: 🤖 🔵 **Nit**: The docstring claim on line 144 — *"Cache writes (creation) are already included in `prompt_tokens` at the full input rate — no adjustment is needed"* — is misleading in two ways:

1. Cache-creation tokens are **not** included in `prompt_tokens` per the OpenAI streaming spec; they appear separately in `prompt_tokens_details.cache_creation_input_tokens`. The function simply doesn't receive them and therefore can't estimate their cost.
2. For Anthropic models, cache writes are billed at **1.25× the input rate** (not 1×), so billing them at full input rate (if they were included) would still underestimate.

Since this is an intentionally approximate fallback estimator, ignoring cache-write cost is fine — but the docstring should say so explicitly rather than implying the math is exact:

```python
* Cache writes (creation) are tracked separately from `prompt_tokens` and
  are not estimated here; any undercount is acceptable for a fallback path.
```

majdyz (commented):
🤖 **PR Review Summary — fix(copilot): baseline cost tracking fallback and cache token display**

All previously flagged blockers and should-fix items have been addressed:
- ✅ Multi-turn delta snapshot (per-call accumulation) — fixed
- ✅ `turn_cache_creation_tokens` extraction from `prompt_tokens_details` — fixed
- ✅ Cache-read discount applied in fallback estimator — fixed
- ✅ `anthropic/claude-opus-4.6` (default non-fast model) added to pricing table — fixed
- ✅ Tiktoken fallback now attempts cost estimate so `cost_usd` is never `None` after it fires — fixed
- ✅ No debug logging or print statements found
- ✅ Error paths handled correctly (finally block, math.isfinite guard)
- ✅ Frontend cache column counts (colSpan 11 in LogsTable, 7 in UserTable) are correct

One nit posted on the `_estimate_cost_from_tokens` docstring (misleading claim about cache-write pricing). Not a blocker.

autogpt_platform/backend/backend/copilot/baseline/service.py
majdyz: Fixed in [a145bbc30](https://github.com/Significant-Gravitas/AutoGPT/commit/a145bbc30104a7b9d330b76db340909b7c702144): Updated the docstring to accurately describe that cache-creation tokens appear in `prompt_tokens_details.cache_creation_input_tokens` (not in `prompt_tokens`), that Anthropic bills them at 1.25× the input rate, and that this estimator intentionally ignores them as an acceptable approximation for a fallback cost estimate.

majdyz:
🤖 Fixed in [012ea167d](https://github.com/Significant-Gravitas/AutoGPT/commit/012ea167d83996e74a19d90ee988f772e2812ed6): Replaced the hardcoded `_CACHE_READ_DISCOUNT` table with per-model `cache_read` rates sourced directly from OpenRouter's `/api/v1/models` response (`pricing.cache_read`). For models where OpenRouter doesn't publish a cache_read rate, cache tokens fall back to the full input rate (safe over-estimate). The discount constants (`anthropic/: 0.10`, `openai/: 0.50`) are removed.

autogpt_platform/backend/backend/copilot/baseline/service.py
```diff
@@ -1190,16 +1344,22 @@ async def stream_chat_completion_baseline(
...
             session=session,
             user_id=user_id,
-            prompt_tokens=state.turn_prompt_tokens,
+            prompt_tokens=uncached_prompt,
             completion_tokens=state.turn_completion_tokens,
+            cache_read_tokens=state.turn_cache_read_tokens,
+            cache_creation_tokens=state.turn_cache_creation_tokens,
             log_prefix="[Baseline]",
             cost_usd=state.cost_usd,
             model=active_model,
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed — the `StreamUsage` event already reports billed (uncached) tokens:

```python
billed_prompt = max(0, state.turn_prompt_tokens - state.turn_cache_read_tokens)
yield StreamUsage(
    prompt_tokens=billed_prompt,
    completion_tokens=state.turn_completion_tokens,
    total_tokens=billed_prompt + state.turn_completion_tokens,
)
```

See [012ea167d](https://github.com/Significant-Gravitas/AutoGPT/commit/012ea167d83996e74a19d90ee988f772e2812ed6) — cached tokens are subtracted before the event is emitted, keeping `prompt_tokens` consistent with `cost_usd`.

majdyz:
🤖 Fixed in [8420ad0c1](https://github.com/Significant-Gravitas/AutoGPT/commit/8420ad0c138ee53cb094fc59da941283f2d58f53): After the tiktoken token backfill (when a provider omits both `x-total-cost` and streaming `usage`), we now call `await _estimate_cost_from_tokens(active_model, state.turn_prompt_tokens, state.turn_completion_tokens)` guarded by `state.cost_usd is None`. When the model is in the OpenRouter pricing response, this sets `state.cost_usd` so `persist_and_record_usage` never receives `cost_usd=None` for the tiktoken path.

autogpt_platform/backend/backend/copilot/baseline/service.py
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [8420ad0c1](https://github.com/Significant-Gravitas/AutoGPT/commit/8420ad0c138ee53cb094fc59da941283f2d58f53): The `StreamUsage` event now reports `billed_prompt = max(0, state.turn_prompt_tokens - state.turn_cache_read_tokens)` instead of the raw `state.turn_prompt_tokens`, so the frontend token count is consistent with `cost_usd` which already applies the cache-read discount.

autogpt_platform/backend/backend/copilot/baseline/service.py
```diff
@@ -356,9 +481,38 @@ async def _baseline_llm_caller(
...
+            if estimated is not None:
+                state.cost_usd = (state.cost_usd or 0.0) + estimated
+                logger.info(
+                    "[Baseline] x-total-cost header missing; estimated cost "
+                    "from token pricing: $%.6f (model=%s)",
+                    estimated,
+                    state.model,
+                )
+
         # Always persist partial text so the session history stays consistent,
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [060c9cf5a](https://github.com/Significant-Gravitas/AutoGPT/commit/060c9cf5afaf288e63cdb1d39b9ac48b799be984): Added a lazy `asyncio.Lock` (`_OPENROUTER_PRICING_LOCK`) with double-checked locking so only one coroutine makes the HTTP request when the cache expires — others wait inside the lock and reuse the freshly-fetched result. Also moved `_OPENROUTER_PRICING_CACHE_FETCHED_AT = now` to *before* the HTTP request so that a failed or timed-out fetch advances the next-allowed-fetch window, creating a natural backoff period instead of every subsequent request immediately retrying during an OpenRouter outage.

PR #12763, opened 2026-04-13
fix(copilot): strip <internal_reasoning> tags from Sonnet response stream

majdyz:
## Summary
- Extract `ThinkingStripper` from `baseline/service.py` into a shared `copilot/thinking_stripper.py` module
- Apply thinking-tag stripping to the SDK streaming path (`_dispatch_response`) so `<internal_reasoning>` and `<thinking>` tags emitted by non-extended-thinking models (e.g. Sonnet) are stripped before reaching the SSE client
- Flush any buffered text from the stripper at stream end so no content is lost
- Add unit tests for the shared `ThinkingStripper` and integration tests for the SDK dispatch path

## Problem
When using Claude Sonnet (which doesn't have extended thinking), the model sometimes outputs `<internal_reasoning>...</internal_reasoning>` tags as visible text in the response stream. The baseline path already stripped these, but the SDK path did not.

## Test plan
- [ ] CI passes (unit tests for ThinkingStripper and SDK dispatch stripping)
- [ ] Manual test: send a message via Sonnet and verify no `<internal_reasoning>` tags appear in the response

### DIFF
--- a/autogpt_platform/backend/backend/copilot/sdk/service.py
+++ b/autogpt_platform/backend/backend/copilot/sdk/service.py
@@ -83,6 +83,7 @@ from ..response_model import (
     StreamStartStep,
     StreamStatus,
     StreamTextDelta,
+    StreamTextEnd,
     StreamToolInputAvailable,
     StreamToolInputStart,
     StreamToolOutputAvailable,
@@ -1743,7 +1744,31 @@ async def _run_stream_attempt(
                 break
 
             # --- Dispatch adapter responses ---
-            for response in state.adapter.convert_message(sdk_msg):
+            adapter_responses = state.adapter.convert_message(sdk_msg)
+            # When StreamFinish is in this batch (ResultMessage), flush any
+            # text buffered by the thinking stripper and inject it as a
+            # StreamTextDelta BEFORE the StreamTextEnd so the Vercel AI SDK
+            # receives the tail inside the still-open text block (correct
+            # protocol order: TextDelta → TextEnd → FinishStep → Finish).
+            if any(isinstance(r, StreamFinish) for r in adapter_responses):
+                tail = acc.thinking_stripper.flush()
+                if tail and not ended_with_stream_error:
+                    acc.assistant_response.content = (
+                        acc.assistant_response.content or ""
+                    ) + tail
+                    tail_delta = StreamTextDelta(
+                        id=state.adapter.text_block_id, delta=tail
+                    )
+                    insert_at = next(
+                        (
+                            i
+                            for i, r in enumerate(adapter_responses)
+                            if isinstance(r, (StreamTextEnd, StreamFinish))
+                        ),
+                        len(adapter_responses),
+                    )
+                    adapter_responses.insert(insert_at, tail_delta)
+            for response in adapter_responses:
                 dispatched = _dispatch_response(
                     response, acc, ctx, state, entries_replaced, ctx.log_prefix
                 )
@@ -1806,16 +1831,6 @@ async def _run_stream_attempt(
                 break
     finally:
         await _safe_close_sdk_client(sdk_client, ctx.log_prefix)
-
-    # --- Flush any text buffered by the thinking stripper ---
-    tail = acc.thinking_stripper.flush()
-    if tail:
-        acc.assistant_response.content = (acc.assistant_response.content or "") + tail
-        # Only yield to the client when the stream succeeded; on error the outer
-        # retry loop rolls back session.messages, so emitting tail would create
-        # a UI-to-session inconsistency (text shown but not persisted).
-        if not ended_with_stream_error:
-            yield StreamTextDelta(id="", delta=tail)
 
     # --- Post-stream processing (only on success) ---
     if state.adapter.has_unresolved_tool_calls:
--- a/autogpt_platform/backend/backend/copilot/sdk/service.py
+++ b/autogpt_platform/backend/backend/copilot/sdk/service.py
@@ -12,7 +12,7 @@ import sys
 import time
 import uuid
 from collections.abc import AsyncGenerator, AsyncIterator
-from dataclasses import dataclass
+from dataclasses import dataclass, field as dataclass_field
 from typing import TYPE_CHECKING, Any, NamedTuple, cast
 
 if TYPE_CHECKING:
@@ -95,6 +95,7 @@ from ..service import (
 from ..token_tracking import persist_and_record_usage
 from ..tools.e2b_sandbox import get_or_create_sandbox, pause_sandbox_direct
 from ..tools.sandbox import WORKSPACE_PREFIX, make_session_path
+from ..thinking_stripper import ThinkingStripper
 from ..tracking import track_user_message
 from .compaction import CompactionTracker, filter_compaction_messages
 from .env import build_sdk_env  # noqa: F401 — re-export for backward compat
@@ -1130,6 +1131,9 @@ class _StreamAccumulator:
     has_appended_assistant: bool = False
     has_tool_results: bool = False
     stream_completed: bool = False
+    thinking_stripper: ThinkingStripper = dataclass_field(
+        default_factory=ThinkingStripper,
+    )
 
 
 def _dispatch_response(
@@ -1186,7 +1190,15 @@ def _dispatch_response(
         )
 
     if isinstance(response, StreamTextDelta):
-        delta = response.delta or ""
+        raw_delta = response.delta or ""
+        # Strip <internal_reasoning> / <thinking> tags that non-extended-
+        # thinking models (e.g. Sonnet) may emit as visible text.
+        delta = acc.thinking_stripper.process(raw_delta)
+        if not delta:
+            # Stripper is buffering a potential tag — suppress this event.
+            return None
+        # Replace the delta with the stripped version for the SSE client.
+        response = StreamTextDelta(id=response.id, delta=delta)
         if acc.has_tool_results and acc.has_appended_assistant:
             acc.assistant_response = ChatMessage(role="assistant", content=delta)
             acc.accumulated_tool_calls = []
@@ -1793,6 +1805,12 @@ async def _run_stream_attempt(
                 break
     finally:
         await _safe_close_sdk_client(sdk_client, ctx.log_prefix)
+
+    # --- Flush any text buffered by the thinking stripper ---
+    tail = acc.thinking_stripper.flush()
+    if tail:
+        acc.assistant_response.content = (acc.assistant_response.content or "") + tail
+        yield StreamTextDelta(id="", delta=tail)
 
     # --- Post-stream processing (only on success) ---
     if state.adapter.has_unresolved_tool_calls:
--- a/autogpt_platform/backend/backend/copilot/sdk/service.py
+++ b/autogpt_platform/backend/backend/copilot/sdk/service.py
@@ -12,7 +12,8 @@ import sys
 import time
 import uuid
 from collections.abc import AsyncGenerator, AsyncIterator
-from dataclasses import dataclass, field as dataclass_field
+from dataclasses import dataclass
+from dataclasses import field as dataclass_field
 from typing import TYPE_CHECKING, Any, NamedTuple, cast
 
 if TYPE_CHECKING:
@@ -92,10 +93,10 @@ from ..service import (
     _is_langfuse_configured,
     _update_title_async,
 )
+from ..thinking_stripper import ThinkingStripper
 from ..token_tracking import persist_and_record_usage
 from ..tools.e2b_sandbox import get_or_create_sandbox, pause_sandbox_direct
 from ..tools.sandbox import WORKSPACE_PREFIX, make_session_path
-from ..thinking_stripper import ThinkingStripper
 from ..tracking import track_user_message
 from .compaction import CompactionTracker, filter_compaction_messages
 from .env import build_sdk_env  # noqa: F401 — re-export for backward compat
@@ -1810,7 +1811,11 @@ async def _run_stream_attempt(
     tail = acc.thinking_stripper.flush()
     if tail:
         acc.assistant_response.content = (acc.assistant_response.content or "") + tail
-        yield StreamTextDelta(id="", delta=tail)
+        # Only yield to the client when the stream succeeded; on error the outer
+        # retry loop rolls back session.messages, so emitting tail would create
+        # a UI-to-session inconsistency (text shown but not persisted).
+        if not ended_with_stream_error:
+            yield StreamTextDelta(id="", delta=tail)
 
     # --- Post-stream processing (only on success) ---
     if state.adapter.has_unresolved_tool_calls:
--- a/autogpt_platform/backend/backend/copilot/sdk/service.py
+++ b/autogpt_platform/backend/backend/copilot/sdk/service.py
@@ -1145,6 +1145,7 @@ def _dispatch_response(
     state: "_RetryState",
     entries_replaced: bool,
     log_prefix: str,
+    skip_strip: bool = False,
 ) -> StreamBaseResponse | None:
     """Process a single adapter response and update session/accumulator state.
 
@@ -1157,6 +1158,10 @@ def _dispatch_response(
     - Accumulating text deltas into `assistant_response`
     - Appending tool input/output to session messages and transcript
     - Detecting `StreamFinish`
+
+    Args:
+        skip_strip: When True, bypass ThinkingStripper.process() for this delta.
+            Used for the flushed tail delta which is already stripped content.
     """
     if isinstance(response, StreamStart):
         return None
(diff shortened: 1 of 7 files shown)


autogpt_platform/backend/backend/copilot/sdk/service.py
```diff
@@ -1794,6 +1806,12 @@ async def _run_stream_attempt(
     finally:
         await _safe_close_sdk_client(sdk_client, ctx.log_prefix)
 
+    # --- Flush any text buffered by the thinking stripper ---
+    tail = acc.thinking_stripper.flush()
+    if tail:
+        acc.assistant_response.content = (acc.assistant_response.content or "") + tail
+        yield StreamTextDelta(id="", delta=tail)
```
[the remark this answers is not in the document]
majdyz: Fixed in ecc087d. The tail from thinking_stripper.flush() is now only yielded to the client when ended_with_stream_error is False. When a stream error occurs, we still update acc.assistant_response.content (which gets rolled back anyway), but we no longer emit the StreamTextDelta to the client, avoiding the UI-to-session inconsistency.

autogpt_platform/backend/backend/copilot/thinking_stripper.py
```diff
@@ -0,0 +1,102 @@
...
+    Handles multiple tag patterns (``<thinking>``, ``<internal_reasoning>``,
+    etc.) so the same stripper works across Claude, Gemini, and other models.
+
+    Buffers just enough characters to detect a tag that may be split
+    across chunks; emits text immediately when no tag is in-flight.
+    Robust to single chunks that open and close a block, multiple
+    blocks per stream, and tags that straddle chunk boundaries.
+    """
+
+    def __init__(self) -> None:
```
majdyz, line 40: 🤖 🔴 **Blocker**: Nested same-type reasoning tags cause content to leak through. When the model emits `<thinking>outer<thinking>inner</thinking>after</thinking>text`, the inner `</thinking>` closes the outer block — `after</thinking>` is then emitted as visible text. Verified locally:

```python
s = ThinkingStripper()
s.process('<thinking>outer<thinking>inner</thinking>after</thinking>final')
# Returns: 'after</thinking>final'  (expected: 'final')
```

While genuine nested reasoning tags from Claude are unlikely, the same-tag-nested scenario (`<thinking><thinking>nested</thinking></thinking>visible`) produces `'</thinking>visible'`. The fix is to track a nesting depth counter per tag-type rather than a simple boolean `_in_thinking` flag.

autogpt_platform/backend/backend/copilot/sdk/service.py
```diff
@@ -1794,6 +1807,16 @@ async def _run_stream_attempt(
...
 
+    # --- Flush any text buffered by the thinking stripper ---
+    tail = acc.thinking_stripper.flush()
+    if tail:
+        acc.assistant_response.content = (acc.assistant_response.content or "") + tail
+        # Only yield to the client when the stream succeeded; on error the outer
+        # retry loop rolls back session.messages, so emitting tail would create
+        # a UI-to-session inconsistency (text shown but not persisted).
+        if not ended_with_stream_error:
+            yield StreamTextDelta(id="", delta=tail)
```
majdyz: 🤖 🔴 **Blocker**: The tail delta is emitted as `StreamTextDelta(id="", delta=tail)` after `StreamFinish` has already been yielded — with no preceding `StreamTextStart`. The Vercel AI SDK protocol requires each delta to reference a valid, open text block id. This violates the `StreamTextStart → StreamTextDelta(s) → StreamTextEnd` contract and will cause the frontend parser to drop or error on the tail text.

The adapter owns the active `text_block_id` (`state.adapter.text_block_id`). The correct fix is to use that id and gate the tail emission before `StreamFinish` is flushed (similar to how `baseline/service.py` emits the tail inside the text block sequence), or at minimum use the adapter's `text_block_id` for the delta.

autogpt_platform/backend/backend/copilot/thinking_stripper.py
```diff
@@ -0,0 +1,102 @@
...
+user.
+
+The :class:`ThinkingStripper` handles chunk-boundary splitting so it can be
+plugged into any delta-based streaming pipeline.
+"""
+
+from __future__ import annotations
+
+# Tag pairs to strip.  Each entry is (open_tag, close_tag).
+REASONING_TAG_PAIRS: list[tuple[str, str]] = [
```
majdyz: 🤖 🟠 **Should Fix**: `REASONING_TAG_PAIRS` is exported as a module-level public name (no leading `_`) but is only used internally by `ThinkingStripper`. Exposing it invites external callers to mutate the list at runtime, which would silently affect all stripper instances (since `_find_open_tag` and the safe-tail scan both iterate the module-level list). Either rename to `_REASONING_TAG_PAIRS` or make it a class-level constant.

autogpt_platform/backend/backend/copilot/thinking_stripper_test.py
```diff
@@ -0,0 +1,93 @@
+"""Tests for the shared ThinkingStripper."""
+
+from backend.copilot.thinking_stripper import ThinkingStripper
```
majdyz, line 3: 🤖 🟡 **Nice to Have**: The new `thinking_stripper_test.py` duplicates the ThinkingStripper tests that already exist in `baseline/service_unit_test.py` (lines 375–427) almost verbatim. Now that the class is extracted, those baseline tests import `_ThinkingStripper` via the module re-export alias — they still pass, but two test files cover the same logic. Consider removing the duplicate set from `baseline/service_unit_test.py` (they were written for the old private class and now test the same shared code) to keep test coverage colocated in the new canonical location.

autogpt_platform/backend/backend/copilot/sdk/thinking_strip_test.py
```diff
@@ -0,0 +1,189 @@
...
+            StreamTextDelta(
+                id="t1",
+                delta="<internal_reasoning>all hidden</internal_reasoning>",
+            ),
+            acc,
+            ctx,
+            state,
+            False,
+            "[test]",
+        )
```
majdyz, line 185: 🤖 🔵 **Nit**: The integration test for the suppressed-only-reasoning delta (`test_reasoning_only_delta_suppressed`) has a loose assertion — `if result is not None: assert result.delta == ""`. Since `_dispatch_response` returns `None` when the stripped delta is empty, this branch is never reached. The assertion should simply be `assert result is None` to make the test deterministic and self-documenting.

majdyz (commented):
## Review Summary

Good refactor overall — extracting `ThinkingStripper` into a shared module is the right call, and the baseline integration looks solid. Two blockers need to be resolved before merge.

### Blockers (2)

1. **Nested same-type tags cause content leakage** (`thinking_stripper.py` line 37): the `_in_thinking` boolean is closed by the first matching close-tag. If the same open-tag appears inside the block, the inner `</thinking>` ends the outer block and the content between it and the outer close becomes visible text. Needs a depth counter per tag-type.

2. **`StreamTextDelta(id="", ...)` emitted post-`StreamFinish`** (`sdk/service.py` line 1818): the tail yield uses an empty string as the block id and has no preceding `StreamTextStart`. This breaks the Vercel AI SDK protocol (matching the existing sentry/coderabbit comments). Use `state.adapter.text_block_id` and emit the tail before `StreamFinish` is flushed, inside the still-open text block.

### Should Fix (1)

- `REASONING_TAG_PAIRS` is a public module-level name but is implementation-private. Rename to `_REASONING_TAG_PAIRS` to prevent accidental external mutation.

### Nice to Have (1)

- `thinking_stripper_test.py` duplicates the ThinkingStripper tests already in `baseline/service_unit_test.py`. Remove the old set from the baseline file.

### Nit (1)

- `test_reasoning_only_delta_suppressed`: the `if result is not None` guard is dead code — simplify to `assert result is None`.

autogpt_platform/backend/backend/copilot/thinking_stripper.py
majdyz: Fixed in `bfa96d471a`. Added a per-tag depth counter (`_depth`) to `ThinkingStripper`. When inside a thinking block, the `process` method now scans for both the open and close tags; if a nested open tag appears before the close tag, depth is incremented and the closer is only honoured when depth reaches 0. Also added two new tests: `test_nested_same_type_tags_do_not_leak` and `test_nested_tags_split_across_chunks`.

autogpt_platform/backend/backend/copilot/sdk/service.py
majdyz: Fixed in `bfa96d471a`. The post-finally `StreamTextDelta(id="", delta=tail)` emission has been removed. Instead, when `convert_message` returns a batch containing `StreamFinish` (i.e., the `ResultMessage` batch), the thinking-stripper tail is flushed and inserted as `StreamTextDelta(id=state.adapter.text_block_id, delta=tail)` BEFORE the `StreamTextEnd` in the response list — keeping the correct Vercel AI SDK protocol order: `TextDelta → TextEnd → FinishStep → Finish`.

autogpt_platform/backend/backend/copilot/thinking_stripper.py
majdyz: Fixed in `bfa96d471a`. Renamed `REASONING_TAG_PAIRS` → `_REASONING_TAG_PAIRS` throughout `thinking_stripper.py` (used only internally).

autogpt_platform/backend/backend/copilot/thinking_stripper_test.py
majdyz: Fixed in `bfa96d471a`. Removed the 7 duplicate `_ThinkingStripper` tests from `baseline/service_unit_test.py` and removed the corresponding `_ThinkingStripper` import. These are fully covered by `thinking_stripper_test.py`.

autogpt_platform/backend/backend/copilot/sdk/thinking_strip_test.py
majdyz: Fixed in `bfa96d471a`. Changed the dead-code `if result is not None: assert result.delta == ""` guard to a direct `assert result is None` — since `_dispatch_response` returns `None` for empty stripped deltas, the conditional was unreachable.

autogpt_platform/backend/backend/copilot/sdk/service.py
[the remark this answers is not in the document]
majdyz: Fixed in `bfa96d471a`. The tail is now emitted BEFORE `StreamTextEnd`/`StreamFinish` using the real `text_block_id` from the adapter, eliminating the `id=""` protocol violation and the post-`StreamFinish` emission.

autogpt_platform/backend/backend/copilot/sdk/service.py
```diff
@@ -1794,6 +1807,16 @@ async def _run_stream_attempt(
...
+
```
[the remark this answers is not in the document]
majdyz: Fixed in `bfa96d471a`. The tail is now emitted BEFORE `StreamTextEnd`/`StreamFinish` using the real `text_block_id` from the adapter, eliminating the `id=""` protocol violation and the post-`StreamFinish` emission.

autogpt_platform/backend/backend/copilot/sdk/service.py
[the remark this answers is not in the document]
majdyz: Fixed in `bfa96d471a`. The tail is now emitted BEFORE `StreamTextEnd`/`StreamFinish` using the real `text_block_id` from the adapter, eliminating the `id=""` protocol violation and the post-`StreamFinish` emission.

autogpt_platform/backend/backend/copilot/sdk/service.py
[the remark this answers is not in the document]
majdyz: All review comments addressed in `bfa96d471a`. See individual thread replies for details on each fix.

autogpt_platform/backend/backend/copilot/sdk/service.py
```diff
@@ -1186,7 +1192,15 @@ def _dispatch_response(
...
-        delta = response.delta or ""
+        raw_delta = response.delta or ""
+        # Strip <internal_reasoning> / <thinking> tags that non-extended-
+        # thinking models (e.g. Sonnet) may emit as visible text.
+        delta = acc.thinking_stripper.process(raw_delta)
+        if not delta:
+            # Stripper is buffering a potential tag — suppress this event.
+            return None
+        # Replace the delta with the stripped version for the SSE client.
+        response = StreamTextDelta(id=response.id, delta=delta)
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [33571643e](https://github.com/Significant-Gravitas/AutoGPT/commit/33571643e35bc440dd3bee8351277d1da49efa1f): Added `skip_strip: bool = False` parameter to `_dispatch_response`. The tail delta is now tracked via identity (`tail_delta`) and passed with `skip_strip=True`, bypassing `ThinkingStripper.process()`. Also removed the manual `acc.assistant_response.content += tail` to fix the double-append (sentry). Regression test added to `thinking_stripper_test.py` documenting the flush-then-process-again invariant.

autogpt_platform/backend/backend/copilot/sdk/service.py
```diff
@@ -1730,7 +1744,31 @@ async def _run_stream_attempt(
...
+                tail = acc.thinking_stripper.flush()
+                if tail and not ended_with_stream_error:
+                    acc.assistant_response.content = (
+                        acc.assistant_response.content or ""
+                    ) + tail
+                    tail_delta = StreamTextDelta(
+                        id=state.adapter.text_block_id, delta=tail
+                    )
+                    insert_at = next(
+                        (
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [33571643e](https://github.com/Significant-Gravitas/AutoGPT/commit/33571643e35bc440dd3bee8351277d1da49efa1f): Removed the manual `acc.assistant_response.content += tail` at the injection site. `_dispatch_response` now handles the single authoritative content append when it processes the tail delta with `skip_strip=True`.

autogpt_platform/backend/backend/copilot/sdk/service.py
```diff
@@ -1186,7 +1192,15 @@ def _dispatch_response(
...
 
     if isinstance(response, StreamTextDelta):
-        delta = response.delta or ""
+        raw_delta = response.delta or ""
+        # Strip <internal_reasoning> / <thinking> tags that non-extended-
+        # thinking models (e.g. Sonnet) may emit as visible text.
+        delta = acc.thinking_stripper.process(raw_delta)
+        if not delta:
+            # Stripper is buffering a potential tag — suppress this event.
+            return None
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [33571643e](https://github.com/Significant-Gravitas/AutoGPT/commit/33571643e35bc440dd3bee8351277d1da49efa1f): The desync was caused by the tail being appended to content manually but then re-suppressed by ThinkingStripper in _dispatch_response. Using `skip_strip=True` ensures the tail passes through to the client exactly as it was appended to content — no more desync.

majdyz:

**Result: PASS**

`<thinking>` and `<internal_reasoning>` tags absent from response body (verified via JS `document.body.innerText.includes('<thinking>')` → false). Reasoning shown via "Show reasoning" button instead. Actual explanation content intact.

### Screenshots

majdyz:
🤖 Fixed in [4ab9adfcd](https://github.com/Significant-Gravitas/AutoGPT/commit/4ab9adfcd4c8d251eace5798a8eb6263f9bf46c6): Switched `ThinkingStripper` import in `sdk/service.py` from `from ..thinking_stripper import ThinkingStripper` (double-dot relative) to `from backend.copilot.thinking_stripper import ThinkingStripper` (absolute) per the project guideline. isort moved it into the existing absolute-import group.

autogpt_platform/backend/backend/copilot/thinking_stripper.py
```diff
@@ -0,0 +1,123 @@
...
+            if self._in_thinking:
+                # Search for both the open and close tags to track nesting.
+                open_pos = self._buffer.find(self._open_tag)
+                close_pos = self._buffer.find(self._close_tag)
+                if close_pos == -1:
+                    # Neither a complete close nor open tag is present; keep
+                    # a tail that could be the start of either tag.
+                    keep = max(len(self._open_tag), len(self._close_tag)) - 1
+                    self._buffer = self._buffer[-keep:] if keep else ""
+                    return "".join(out)
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [1f7a48ce1](https://github.com/Significant-Gravitas/AutoGPT/commit/1f7a48ce14c8528f8e411cae31b7401b5cb84025): Confirmed valid. When inside a thinking block and `close_pos == -1`, the code now checks for a complete nested open tag before trimming — if found, increments `_depth` and continues the loop instead of returning. Added regression test `test_nested_open_tag_depth_tracked_across_chunk_boundary` covering the exact scenario.

PR #12767, opened 2026-04-13
fix(backend/copilot): reuse existing credentials across chat sessions

0ubbe:
## Summary

Fixes the bug where users are asked to sign in to GitHub on every new copilot chat session, even though they already authenticated in a previous session.

### Why

Credentials are correctly stored per-user (not per-session) in the database, but:
1. The `connect_integration` tool always showed a credential setup card without checking if the user already had valid credentials
2. The LLM was never told about existing connected integrations, so it proactively called `connect_integration` before attempting any operation

### What

**`connect_integration.py`**
- Check `get_provider_token()` before showing setup card — if valid credentials exist, return "already connected" immediately
- Updated tool description to instruct the LLM to try operations first and only call `connect_integration` on auth failure (not proactively)

**`service.py`**
- Added `get_connected_integrations()` — checks which providers the user has valid tokens for
- Added `format_connected_integrations()` — formats as a prompt note telling the LLM not to re-auth

**`sdk/service.py` + `baseline/service.py`**
- Inject connected integrations list into `<user_context>` block on first turn, alongside existing business understanding context

### How

Two layers of defense:
1. **Prompt-level**: LLM now sees "Connected integrations: GitHub. Tokens are automatically injected — do NOT call connect_integration..." in the user context
2. **Tool-level**: Even if the LLM calls `connect_integration` anyway, the tool checks for existing credentials and returns immediately without showing the setup card

## Test plan

- [ ] Manual: start a new chat session with GitHub already connected → LLM should NOT show "Sign in to GitHub"
- [ ] Manual: start a new session WITHOUT credentials → LLM should show setup card when auth fails
- [ ] Manual: revoke credentials, start new session → LLM should prompt for auth after operation fails
- [ ] Verify: `poetry run lint` passes

### DIFF
--- a/autogpt_platform/backend/backend/copilot/baseline/service.py
+++ b/autogpt_platform/backend/backend/copilot/baseline/service.py
@@ -56,6 +56,8 @@ from backend.copilot.service import (
     _get_openai_client,
     _update_title_async,
     config,
+    format_connected_integrations,
+    get_connected_integrations,
     inject_user_context,
     strip_user_context_tags,
 )
@@ -1075,16 +1077,32 @@ async def stream_chat_completion_baseline(
     user_message_for_transcript = message
     if should_inject_user_context:
         prefixed = await inject_user_context(
-            understanding, message or "", session_id, session.messages
+            understanding,
+            message or "",
+            session_id,
+            session.messages,
+            connected_integrations=(
+                await get_connected_integrations(user_id) if user_id else None
+            ),
         )
         if prefixed is not None:
             for msg in openai_messages:
                 if msg["role"] == "user":
                     msg["content"] = prefixed
                     break
-            user_message_for_transcript = prefixed
-        else:
-            logger.warning("[Baseline] No user message found for context injection")
+            if prefixed is not None:
+                # Persist the prefixed content so subsequent turns and --resume
+                # retain the user context.
+                for idx, session_msg in enumerate(session.messages):
+                    if session_msg.role == "user":
+                        session_msg.content = prefixed
+                        await update_message_content_by_sequence(
+                            session_id, idx, prefixed
+                        )
+                        break
+                user_message_for_transcript = prefixed
+            else:
+                logger.warning("[Baseline] No user message found for context injection")
 
     # Inject Graphiti warm context into the first user message (not the
     # system prompt) so the system prompt stays static and cacheable.
--- a/autogpt_platform/backend/backend/copilot/service.py
+++ b/autogpt_platform/backend/backend/copilot/service.py
@@ -27,6 +27,7 @@ from backend.util.exceptions import NotAuthorizedError, NotFoundError
 from backend.util.settings import AppEnvironment, Settings
 
 from .config import ChatConfig
+from .integration_creds import get_provider_token
 from .model import (
     ChatMessage,
     ChatSessionInfo,
@@ -34,6 +35,7 @@ from .model import (
     update_session_title,
     upsert_chat_session,
 )
+from .providers import SUPPORTED_PROVIDERS
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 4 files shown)


autogpt_platform/backend/backend/copilot/service.py
```diff
@@ -356,6 +358,31 @@ async def inject_user_context(
...
+    """Format a list of connected integration names for prompt injection."""
+    if not names:
+        return ""
+    joined = ", ".join(names)
+    return (
+        f"Connected integrations: {joined}. "
+        f"Tokens are automatically injected — do NOT call connect_integration "
+        f"for these providers unless an operation fails with an auth error."
+    )
+
```
majdyz, line 502: 🤖 🔴 **Blocker**: `get_connected_integrations` and `format_connected_integrations` are defined here but never called.

The PR description promises a two-layer defense where Layer 1 ("Prompt-level") injects connected integration names into the `<user_context>` block on first turn. However, neither `sdk/service.py` nor `baseline/service.py` imports or calls these functions — they are dead code.

Layer 2 (the tool-level short-circuit in `connect_integration.py`) still provides value, but the intended prompt-level nudge that stops the LLM from proactively calling `connect_integration` at all is missing. Without it, the LLM will still call the tool on first turn; the short-circuit just avoids the setup card rather than avoiding the extra tool round-trip entirely.

The fix is to call these functions inside `inject_user_context` (or at the call sites in both service files) and append `format_connected_integrations(names)` to the `<user_context>` block.

autogpt_platform/backend/backend/copilot/baseline/service.py
```diff
@@ -977,9 +977,19 @@ async def stream_chat_completion_baseline(
...
+                for idx, session_msg in enumerate(session.messages):
+                    if session_msg.role == "user":
+                        session_msg.content = prefixed
+                        await update_message_content_by_sequence(
+                            session_id, idx, prefixed
+                        )
+                        break
+                user_message_for_transcript = prefixed
+            else:
+                logger.warning("[Baseline] No user message found for context injection")
```
majdyz, line 1105: 🤖 🔵 **Nit**: Redundant `if prefixed is not None` guard.

The outer `if prefixed is not None:` block already ensures `prefixed` is not `None` here. The inner `if prefixed is not None:` on this same line is unreachable as `False` and the nested `else` warning will never fire. This was probably copied from the outer condition guard but is now dead code — the `else` branch with the `logger.warning` can never be reached.

Remove the inner guard and flatten the block.

majdyz (changes requested):
🤖 **REQUEST_CHANGES** — Three blockers need resolution before merge.

**Critical bugs (already flagged by other reviewers, confirming):**

1. **`NameError` at runtime** ( line 1222): `update_message_content_by_sequence` is called but not imported. Should be `await chat_db().update_message_content_by_sequence(session_id, idx, prefixed)` (consistent with how `service.py` uses it) or imported from `backend.copilot.db`.

2. **Wrong response type** (`connect_integration.py` line 159): Returning bare `ToolResponseBase` with `type=SETUP_REQUIREMENTS` is semantically incorrect. The `ConnectIntegrationTool.tsx` frontend component guards on `"setup_info" in parsed` so the setup card correctly doesn't render, but the type mismatch is still confusing and violates the schema contract. Use a clean type like `ResponseType.MESSAGE` or add a dedicated `ALREADY_CONNECTED` type.

**New finding (not covered by existing comments):**

3. **Dead code — prompt-level defense never fires**: `get_connected_integrations()` and `format_connected_integrations()` are defined in `service.py` but imported and called by neither `sdk/service.py` nor `baseline/service.py`. The PR description's "Layer 1" defense is not wired up. See inline comment on `service.py:384`.

**Security:** No credential leakage found. The `get_provider_token` path uses the user's own `user_id` from the auth dependency, so no cross-user leakage risk.

**Nit:** Inner `if prefixed is not None:` inside an already-true outer block (`baseline/service.py` ~line 992) is unreachable dead code — the `else` warning never fires.

PR #12777, opened 2026-04-14
fix(backend/copilot): fix T2+ --resume by using CLI native session file

majdyz:
## Why

The Claude CLI 2.1.97 (bundled in `claude-agent-sdk 0.1.58`) changed the `--resume` flag to accept a **session UUID**, not a file path. Our service was incorrectly passing a temp file path (from `write_transcript_to_tempfile`), causing the CLI subprocess to crash with exit code 1 on every T2+ message — breaking all multi-turn CoPilot conversations.

Additionally, using a file-per-pod approach meant pod affinity was required for `--resume` to work (the file only existed on the pod that handled T1).

## What

- Add `upload_cli_session()` to `transcript.py`: after each turn, upload the CLI's native session JSONL (at `{projects_base}/{encoded_cwd}/{session_id}.jsonl`) to remote storage
- Add `restore_cli_session()` to `transcript.py`: before T2+, download and restore the CLI native session file to the expected path
- Pass `--session-id {app_uuid}` via `ClaudeAgentOptions` so the CLI uses the app session UUID as its session ID → predictable file path
- On T2+: call `restore_cli_session()` and if successful, pass `--resume {session_uuid}` (UUID, not file path)
- Remove `write_transcript_to_tempfile` from the resume path in service.py (it only exists in transcript.py for compaction use)
- Keep DB reconstruction as last-resort fallback (populates builder state only, no `--resume`)
- Compaction retry path now runs without `--resume` (compacted content cannot be written in CLI native format)

## How

**Normal multi-turn flow (fixed):**
1. T1: SDK runs with `--session-id {app_uuid}` → CLI writes session to predictable path
2. T1 finally: `upload_cli_session()` uploads native session to storage (GCS/local)
3. T2+: `restore_cli_session()` downloads and writes the native session back to disk
4. T2+: `--resume {app_uuid}` → CLI reads the restored session → full context preserved

**Cross-pod benefit:**
The native session file is now in remote storage, so any pod can restore it before a turn. Pod affinity for CoPilot is no longer required.

**Backward compatibility:**
- First turn: no native session in storage → runs without `--resume` (same as before)
- If `restore_cli_session` fails: falls back gracefully to no `--resume`, logs a warning
- DB reconstruction still available as last resort when no transcript exists at all

## Checklist

- [x] Tests updated (service_helpers_test, retry_scenarios_test, transcript_test all pass)
- [x] `poetry run ruff check` clean
- [x] `poetry run black --check` clean
- [x] `poetry run pyright` 0 errors on changed files


autogpt_platform/backend/backend/copilot/sdk/service.py
```diff
@@ -2981,6 +2991,32 @@ def _on_stderr(line: str) -> None:
...
+        # Uploading it here enables --resume on any pod (no pod affinity needed).
+        # Runs after upload_transcript so both are available for the next turn.
+        if (
+            config.claude_agent_use_resume
+            and user_id
+            and sdk_cwd
+            and session is not None
+        ):
+            try:
+                await asyncio.shield(
```
majdyz, line 3034: 🤖 🟠 **Should Fix**: `upload_cli_session` is called with `asyncio.shield` but the exception handler catches `Exception` after `asyncio.shield` completes — however, if the outer coroutine (the `finally` block) is cancelled while `asyncio.shield` is awaited, the cancellation will be delivered to `asyncio.shield` itself, and the shielded inner coroutine continues but the `except Exception` block in the outer scope will not catch `asyncio.CancelledError` (it's a `BaseException`). This means upload can silently fail to propagate the cancellation to callers. Consider using the same pattern as the existing `upload_transcript` block above, or document explicitly why cancellation safety is handled correctly here.

autogpt_platform/backend/backend/copilot/sdk/service_helpers_test.py
```diff
@@ -186,7 +185,8 @@ async def test_compaction_returns_same_content_drops(self) -> None:
         assert ctx.transcript_lost is True
 
     @pytest.mark.asyncio
-    async def test_write_tempfile_fails_drops(self) -> None:
+    async def test_compaction_same_size_drops(self) -> None:
```
majdyz: 🤖 🔵 **Nit**: The test name `test_compaction_same_size_drops` is misleading — the test actually covers the case where `validate_transcript` returns `False` for compacted content (not same-size content). A more accurate name would be `test_compaction_invalid_transcript_drops`. The old name (`test_write_tempfile_fails_drops`) was replaced but the new one doesn't describe what's actually being tested.

autogpt_platform/backend/backend/copilot/transcript.py
```diff
@@ -652,6 +654,148 @@ def _build_meta_storage_path(user_id: str, session_id: str, backend: object) ->
...
+) -> tuple[str, str, str]:
+    """Return (workspace_id, file_id, filename) for a CLI session file in storage."""
+    return (
+        _CLI_SESSION_STORAGE_PREFIX,
+        _sanitize_id(user_id),
+        f"{_sanitize_id(session_id)}.jsonl",
+    )
+
+
+async def upload_cli_session(
```
majdyz, line 692: 🤖 🟠 **Should Fix**: `upload_cli_session` and `restore_cli_session` are new public async functions in `transcript.py` with no corresponding tests in `transcript_test.py`. The path boundary check (`real_path.startswith(projects_base + os.sep)`), the `FileNotFoundError` fast-path, and the success/failure return values of `restore_cli_session` are all untested. Given that this is safety/security-relevant code (path traversal guard), these should have unit tests similar to the existing `TestSanitizeId` and `TestStoragePathParts` classes.

autogpt_platform/backend/backend/copilot/transcript.py
```diff
@@ -652,6 +654,148 @@ def _build_meta_storage_path(user_id: str, session_id: str, backend: object) ->
...
+    """Download and restore the CLI's native session file for --resume.
+
+    Returns True if the file was successfully restored and --resume can be
+    used with the session UUID.  Returns False if not available (first turn
+    or upload failed), in which case the caller should not set --resume.
+    """
+    storage = await get_workspace_storage()
+    wid, fid, fname = _cli_session_storage_path_parts(user_id, session_id)
+
+    if isinstance(storage, GCSWorkspaceStorage):
```
majdyz: 🤖 🟡 **Nice to Have**: `restore_cli_session` hard-codes the storage path construction using `isinstance(storage, GCSWorkspaceStorage)` branching. This is fragile — if a new storage backend is added (e.g., S3), this will silently fall through to the `local://` branch and fail. The existing `_build_path_from_parts` helper already abstracts this pattern. Consider reusing that instead of duplicating the backend-detection logic.

autogpt_platform/backend/backend/copilot/sdk/retry_scenarios_test.py
```diff
@@ -811,20 +811,21 @@ def test_session_messages_rollback_on_retry(self):
...
-        skip_transcript_upload = False
-        use_resume = True
-        resume_file = None  # write_transcript_to_tempfile returned None
-
-        if not resume_file:
-            use_resume = False
-            skip_transcript_upload = True
-
-        assert skip_transcript_upload is True
+    def test_cli_session_restore_failure_skips_resume(self):
```
majdyz, line 814: 🤖 🟡 **Nice to Have**: `test_cli_session_restore_failure_skips_resume` is a pure logic simulation (not an actual function call). It manually mimics the `if cli_restored:` branch in `service.py` with hardcoded variables. This means it will not catch regressions if the branch logic in `service.py` changes. Consider making this an actual integration test that mocks `restore_cli_session` to return `False` and asserts that the resulting `sdk_options_kwargs` dict does not contain the `resume` key.

autogpt_platform/backend/backend/copilot/transcript.py
```diff
@@ -652,6 +654,148 @@ def _build_meta_storage_path(user_id: str, session_id: str, backend: object) ->
...
+    The CLI only writes the session file after the turn completes, so this
+    must run in the finally block, AFTER the SDK stream has finished.
+    """
+    session_file = _cli_session_path(sdk_cwd, session_id)
+    real_path = os.path.realpath(session_file)
+    projects_base = _projects_base()
+
+    if not real_path.startswith(projects_base + os.sep):
+        logger.warning(
+            "%s CLI session file outside projects base, skipping upload: %s",
```
majdyz, line 712: 🤖 🔵 **Nit**: The AGENTS.md style guide requires using `%s` deferred interpolation in `debug`/`warning`/`info` log calls and f-strings only where specified. This is already followed throughout the diff. But on line 709 the path is logged as `session_file` (the pre-realpath value), while the boundary check is done against `real_path`. Consider logging `real_path` in the warning so it matches what was actually checked — easier to debug when `realpath` resolves a symlink.

autogpt_platform/backend/backend/copilot/sdk/service.py
majdyz: 🤖 Fixed in [e515638d2](https://github.com/Significant-Gravitas/AutoGPT/commit/e515638d2d966ab4dd0b49cf3b81547956fa8a4c): Added a comment in `service.py` explaining the intentional `asyncio.shield` + `CancelledError` semantics — when cancelled, `CancelledError` (a `BaseException`) propagates out of the `except Exception` block so the caller sees the cancellation while the shielded inner coroutine continues to completion. This is the same pattern used by `upload_transcript` immediately above.

autogpt_platform/backend/backend/copilot/transcript.py
majdyz: 🤖 Fixed in [e515638d2](https://github.com/Significant-Gravitas/AutoGPT/commit/e515638d2d966ab4dd0b49cf3b81547956fa8a4c): Added `TestCliSessionPath`, `TestUploadCliSession`, and `TestRestoreCliSession` classes in `transcript_test.py`. Tests cover: path boundary guard (rejects paths outside projects base), `FileNotFoundError` fast-path (skips upload silently), path-traversal resistance in `_cli_session_path`, and `restore_cli_session` returning `False` when session is missing or restore path is outside base.

autogpt_platform/backend/backend/copilot/transcript.py
majdyz: 🤖 Fixed in [e515638d2](https://github.com/Significant-Gravitas/AutoGPT/commit/e515638d2d966ab4dd0b49cf3b81547956fa8a4c): Replaced the manual `isinstance(storage, GCSWorkspaceStorage)` branching in `restore_cli_session` with a call to the existing `_build_path_from_parts` helper — same pattern used by `_build_storage_path` and `_build_meta_storage_path`. Any new storage backend just needs to be handled in one place.

autogpt_platform/backend/backend/copilot/sdk/service_helpers_test.py
majdyz: 🤖 Fixed in [e515638d2](https://github.com/Significant-Gravitas/AutoGPT/commit/e515638d2d966ab4dd0b49cf3b81547956fa8a4c): Renamed `test_compaction_same_size_drops` → `test_compaction_invalid_transcript_drops`. The new name accurately reflects that the test verifies compacted content that fails `validate_transcript` is dropped.

autogpt_platform/backend/backend/copilot/sdk/retry_scenarios_test.py
majdyz: 🤖 Fixed in [e515638d2](https://github.com/Significant-Gravitas/AutoGPT/commit/e515638d2d966ab4dd0b49cf3b81547956fa8a4c): Added `test_resume_skipped_when_cli_session_missing` to `TestStreamChatCompletionRetryIntegration`. It calls the real `stream_chat_completion_sdk`, patches `restore_cli_session` to return `False`, captures the `ClaudeSDKClient` kwargs, and asserts `resume` is absent — so any change to the `cli_restored` branch in `service.py` will break this test immediately. The original simple-variable test in `TestRetryStateReset` is kept with a reference to the new integration test.

autogpt_platform/backend/backend/copilot/transcript.py
majdyz: 🤖 Fixed in [e515638d2](https://github.com/Significant-Gravitas/AutoGPT/commit/e515638d2d966ab4dd0b49cf3b81547956fa8a4c): Changed the boundary-check warning in `upload_cli_session` to log `real_path` (the resolved path that was actually checked against `projects_base`) instead of `session_file` (the pre-realpath value). This makes symlink-related debugging much easier.

majdyz:
## E2E Test Report — PR #12777

**Branch:** `fix/copilot-sdk-resume-exclude-dynamic-sections`
**Date:** 2026-04-14
**Auth mode:** OpenRouter API key + CLAUDE_AGENT_USE_RESUME=true
**CLI version:** claude 2.1.107

---

### Bug Found & Fixed (--fix mode)

During testing, a secondary bug was discovered and fixed:

**Root cause:** CLI 2.1.107 rejects combining `--session-id` with `--resume` unless `--fork-session` is also specified.

```
Error: --session-id can only be used with --continue or --resume if --fork-session is also specified.
```

The PR correctly added `--session-id {uuid}` on T1 to write the session file to a predictable path, AND added `--resume {uuid}` on T2+. But passing **both** on T2+ caused the CLI to crash with exit code 1.

**Fix applied:** Made `--session-id` and `--resume` mutually exclusive in `sdk/service.py`:
- T1 (no resume): pass `--session-id {uuid}` only
- T2+ (with resume): pass `--resume {uuid}` only

**Commit:** `eb7d288a82` pushed to branch.

---

### Test Results

| Scenario | Result |
|----------|--------|
| T1: First message succeeds | PASS |
| T2: Second message has T1 context ("TestUser" recalled) | PASS |
| T3: Third message has full chain context | PASS |
| After T1: upload_cli_session() called (log confirmed) | PASS |
| Before T2: restore_cli_session() called (log confirmed) | PASS |
| --resume uses UUID (not file path) | PASS |
| No exit code 1 on T2+ | PASS (after fix) |

**7/7 scenarios passing.**

---

### Key Log Evidence

T1 upload:
`[SDK][055e7862-441][T1] Uploaded CLI session file (21404B) for cross-pod --resume`

T2 restore + resume:
`[SDK][055e7862-441][T2] Restored CLI session file (21404B) for --resume`
`[SDK][055e7862-441][T2] Using --resume 055e7862 (5498B transcript, msg_count=4)`

T2 response: "Your name is TestUser. Now, what can I help you automate today?"

T3 response: "The first thing you told me was your name - 'My name is TestUser. Please remember this.'"

---

### Screenshots

autogpt_platform/backend/backend/copilot/sdk/service.py
```diff
@@ -2981,6 +3002,43 @@ def _on_stderr(line: str) -> None:
...
+        # used for upload_transcript immediately above.
+        if (
+            config.claude_agent_use_resume
+            and user_id
+            and sdk_cwd
+            and session is not None
+            and state is not None
+            and not ended_with_stream_error
+            and not skip_transcript_upload
+            and (not has_history or state.use_resume)
```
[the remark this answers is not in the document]
majdyz: 🤖 Not applicable — this is a false positive. The guard `(not has_history or state.use_resume)` is intentionally restrictive.

**Why uploading a fresh CLI session on T2+ restore-miss is harmful, not helpful:**

When `state.use_resume=False` on T2+, the CLI ran with `--session-id` (no `--resume`), starting a fresh conversation. The resulting native session file contains **only the current turn's context** — it does not include T1's conversation history or any compacted summary. Uploading this incomplete file would cause T3 to "restore" a session with no prior context, silently dropping everything from T1 onward — worse than the current behavior.

**Why this is not an unrecoverable loop:**

1. `restore_cli_session` is attempted independently on every turn — if a transient storage error caused the T2 restore to fail, T3 will try again and succeed once storage recovers.
2. `upload_transcript` (the custom JSONL at lines 2953-3003) runs unconditionally and always captures the full conversation state. This serves as the authoritative context for the next turn's `_build_query_message` reconstruction path, independent of the CLI native session.

**The compaction path is equally covered:** After compaction, the CLI also starts fresh. `skip_transcript_upload=True` handles `transcript_lost=True` (lines 2509-2510), and `state.use_resume=False` (line 2505) blocks the CLI upload for successful compaction — correct, because the fresh CLI session again lacks the compacted context.

The guard at line 3024 is correct as-is. No code change needed.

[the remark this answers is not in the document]
majdyz: 🤖 Not a bug — this is intentional design. The guard `(not has_history or state.use_resume)` prevents a specific silent data-loss scenario:

**Why T2+ without `--resume` must NOT upload its CLI session:**
On T2+ turns where `restore_cli_session` fails or T1's upload was missed, the CLI runs without `--resume`. It writes a session file that only contains the current turn (T2) — it has no T1 context built in. If we uploaded that T2-only file, T3 would `--resume` from it. Since `use_resume=True`, `_build_query_message` trusts `transcript_msg_count` and skips gap reconstruction, silently losing T1's context. Allowing the upload would trade a recoverable degradation for an undetectable data loss.

**Why the "loop" is graceful, not fatal:**
When the CLI session is absent from storage, all subsequent turns fall back to the transcript path (always uploaded via a separate code block that does *not* check `use_resume`). `_build_query_message` reconstructs T1 context via compression or DB-message reconstruction. The model receives correct full context — just at slightly higher token cost instead of the efficient native `--resume` path. No conversation context is lost; only the performance optimisation degrades.

autogpt_platform/backend/backend/copilot/transcript.py
```diff
@@ -652,6 +654,144 @@ def _build_meta_storage_path(user_id: str, session_id: str, backend: object) ->
...
+
+    The CLI encodes the working directory by replacing '/' with '-', then
+    places its session file at::
+
+        {projects_base}/{encoded_cwd}/{session_id}.jsonl
+
+    When we pass ``--session-id {session_id}``, the CLI uses the app session
+    UUID as its own session UUID, making the file path predictable.
+    """
+    encoded_cwd = sdk_cwd.replace("/", "-")
```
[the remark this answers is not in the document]
majdyz: 🤖 Confirmed — fixed in [4596b0a](https://github.com/Significant-Gravitas/AutoGPT/commit/4596b0a3e6cdc4a2cfd5e10f69fb7908affb3dbe). `_cli_session_path` now uses `re.sub(r'[^a-zA-Z0-9]', '-', os.path.realpath(sdk_cwd))` which mirrors the CLI's own encoding exactly. Also fixed the corresponding tests in `transcript_test.py` ([a3c4eac](https://github.com/Significant-Gravitas/AutoGPT/commit/a3c4eac9642d4ec9cf504a2dee7fad00ac43be4b)) since they used the old `replace('/', '-')` encoding.

autogpt_platform/backend/backend/copilot/transcript.py
```diff
@@ -652,6 +654,147 @@ def _build_meta_storage_path(user_id: str, session_id: str, backend: object) ->
...
+
+    try:
+        os.makedirs(os.path.dirname(real_path), exist_ok=True)
+        Path(real_path).write_bytes(content)
+        logger.info(
+            "%s Restored CLI session file (%dB) for --resume",
+            log_prefix,
+            len(content),
+        )
+        return True
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [9abaeb218](https://github.com/Significant-Gravitas/AutoGPT/commit/9abaeb218e0d3126bf137e3a253a2ee6d4b5fb9c): `restore_cli_session` now checks whether the local session file already exists **before** downloading from storage. If it does, it returns `True` immediately — preserving any extended history that a previous turn's CLI wrote but failed to upload. The storage download only runs when no local file is present (cross-pod scenario), making the overwrite path unreachable in the same-pod case. Added a unit test covering this behavior (`test_returns_true_when_local_file_already_exists`).

PR #12789, opened 2026-04-15
fix(backend/copilot): null-safe token accumulation for OpenRouter null cache fields

majdyz:
## Why
OpenRouter occasionally returns `null` (not `0`) for `cache_read_input_tokens` and `cache_creation_input_tokens` on the initial streaming event, before real token counts are available. Python's `dict.get(key, 0)` only falls back to `0` when the key is **missing** — when the key exists with a `null` value, `.get(key, 0)` returns `None`. This causes `TypeError: unsupported operand type(s) for +=: 'int' and 'NoneType'` in the usage accumulator on the first streaming chunk from OpenRouter models.

## What
- Replace `.get(key, 0)` with `.get(key) or 0` for all four token fields in `_run_stream_attempt`
- Add `TestTokenUsageNullSafety` unit tests in `service_helpers_test.py`

## How
Minimal targeted fix — only the four `+=` accumulation lines changed. No behaviour change for Anthropic-native models (they never emit null values).

## Checklist
- [x] Tests cover null event, real event, absent keys, and multi-turn accumulation
- [x] No behaviour change for Anthropic-native models
- [x] No API changes

### DIFF
--- a/autogpt_platform/backend/backend/copilot/sdk/service.py
+++ b/autogpt_platform/backend/backend/copilot/sdk/service.py
@@ -1688,15 +1688,20 @@ async def _run_stream_attempt(
                 #   cache_read_input_tokens = served from cache
                 #   cache_creation_input_tokens = written to cache
                 if sdk_msg.usage:
-                    state.usage.prompt_tokens += sdk_msg.usage.get("input_tokens", 0)
-                    state.usage.cache_read_tokens += sdk_msg.usage.get(
-                        "cache_read_input_tokens", 0
+                    # Use `or 0` instead of a default in .get() because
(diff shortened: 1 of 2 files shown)


majdyz:
🤖 Addressed in [a042c8490](https://github.com/Significant-Gravitas/AutoGPT/commit/a042c84907a4492590ab00c7bcc56215a76d6e1e): extracted the four-field token accumulation into a standalone `_apply_token_usage()` helper so tests call the production function directly (fixing the codecov patch coverage gap). Refactored `TestTokenUsageNullSafety` to drop the local mirror and added `pytest.mark.parametrize` coverage over all four token fields per the nitpick suggestion.

PR #12796, opened 2026-04-15
fix(backend/copilot): fix initial load missing messages + forward pagination for completed sessions

majdyz:
### Why / What / How

**Why:** Completed copilot sessions with many messages showed a completely empty chat view. A user reported a 158-message session that appeared blank on reload.

**What:** Two bugs fixed:
1. **Backend** — initial page load always returned the newest 50 messages in DESC order. For sessions heavy in tool calls, the user's original messages (seq 0–5) were never included; all 50 slots consumed by mid-session tool outputs.
2. **Frontend** — convertChatSessionToUiMessages silently dropped user messages with null/empty content.

**How:** For completed sessions (no active stream), the backend now loads from sequence 0 in ASC order. Active/streaming sessions keep newest-first for streaming context. A new after_sequence forward cursor enables infinite-scroll for subsequent pages (sentinel moves to bottom). The frontend wires forward_paginated + newest_sequence end-to-end.

### Changes 🏗️

- db.py: added from_start (ASC) and after_sequence (forward cursor) modes; added newest_sequence to PaginatedMessages
- routes.py: detect completed vs active on initial load; pass from_start=True for completed; expose newest_sequence + forward_paginated; accept after_sequence param
- convertChatSessionToUiMessages.ts: never drop user messages with empty content
- useLoadMoreMessages.ts: forward pagination via after_sequence; append pages to end
- ChatMessagesContainer.tsx: LoadMoreSentinel at bottom for forward-paginated sessions
- Wire newestSequence + forwardPaginated end-to-end through useChatSession/useCopilotPage/ChatContainer
- openapi.json: add after_sequence + newest_sequence/forward_paginated; regenerate types
- db_test.py: 9 new unit tests for from_start and after_sequence modes

#### For code changes:
- [x] I have made a test plan
  - [x] Open a completed session with many messages — first user message visible on initial load
  - [x] Scroll to bottom of completed session — load more appends next page
  - [x] Open active/streaming session — newest messages shown first, streaming unaffected
  - [x] Backend unit tests: all 28 pass
  - [x] Frontend lint/format: clean, no new type errors


autogpt_platform/frontend/src/app/(platform)/copilot/components/ChatMessagesContainer/ChatMessagesContainer.tsx
```diff
@@ -283,7 +288,7 @@ export function ChatMessagesContainer({
       }
     >
       <ConversationContent className="flex min-h-full flex-1 flex-col gap-6 px-3 py-6">
-        {hasMoreMessages && onLoadMore && (
+        {hasMoreMessages && onLoadMore && !forwardPaginated && (
```
majdyz, line 354: 🤖 🔴 **Blocker**: PR #12792 also modifies `ChatMessagesContainer.tsx` and `useLoadMoreMessages.ts` and is targeting the same base branch (`dev`). Specifically:
- #12792 refactors `LoadMoreSentinel` (exports it, adds `MAX_AUTO_FILL_ROUNDS`, adds a manual fallback button, changes the scroll-non-scrollable guard)
- #12792 also rewrites the `useEffect` in `useLoadMoreMessages` to preserve paged state across session refetches

These overlap directly with the changes in this PR and will produce merge conflicts. Recommend coordinating merge order or rebasing this PR on top of #12792 once it lands.

autogpt_platform/backend/backend/api/features/chat/routes.py
```diff
@@ -451,50 +453,79 @@ async def get_session(
...
+
+    # Completed sessions on initial load start from sequence 0 so the user's
+    # initial prompt is always visible.  Active sessions keep the legacy
+    # newest-first behavior to preserve streaming context.
+    from_start = is_initial_load and active_session is None
+
     page = await get_chat_messages_paginated(
-        session_id, limit, before_sequence, user_id=user_id
+        session_id,
+        limit,
```
majdyz, line 522: 🤖 🟠 **Should Fix**: The f-string log interpolates `session_id` (user-controlled) directly. While not a direct injection risk here, this bypasses Python's logging `%s` lazy formatting and would include raw user data in logs. Prefer:
```python
logger.info(
    "[GET_SESSION] session=%s, active=%s, from_start=%s, forward_paginated=%s, msg_count=%d, last_role=%s",
    session_id, active_session is not None, from_start,
    from_start or after_sequence is not None, len(messages),
    messages[-1].get('role') if messages else 'none',
)
```

autogpt_platform/backend/backend/api/features/chat/routes.py
```diff
@@ -504,6 +535,8 @@ async def get_session(
             active_stream=None,
             has_more_messages=page.has_more,
             oldest_sequence=page.oldest_sequence,
+            newest_sequence=page.newest_sequence,
+            forward_paginated=after_sequence is not None,
```
majdyz: 🤖 🟠 **Should Fix**: On load-more responses (`not is_initial_load`), `forward_paginated` is set as `after_sequence is not None` which is correct for forward pages. But for backward load-more (`before_sequence` set), it returns `False`. The frontend `useLoadMoreMessages` only reads `forward_paginated` from the initial session load, so this is OK in practice — but it's semantically wrong and could confuse future callers. Consider either omitting `forward_paginated` from load-more responses (default False is fine) or computing it consistently as `from_start or after_sequence is not None`.

autogpt_platform/frontend/src/app/api/openapi.json
```diff
@@ -12,16 +12,25 @@
         "tags": ["v2", "admin", "platform-cost", "admin"],
         "summary": "Get Platform Cost Dashboard",
         "operationId": "getV2Get platform cost dashboard",
-        "security": [{ "HTTPBearerJWT": [] }],
+        "security": [
```
majdyz: 🤖 🟡 **Nice to Have**: The `openapi.json` diff is ~12k lines of pure reformatting (compact JSON → pretty-printed JSON). The actual changes are only 3 new fields. This makes the diff extremely hard to review and will cause merge conflicts with other PRs touching the spec. Consider reverting the formatting and re-applying only the field additions manually.

autogpt_platform/frontend/src/app/api/openapi.json
```diff
@@ -12,16 +12,25 @@
...
+          {
```
majdyz: 🤖 🟡 **Nice to Have**: The `openapi.json` diff is ~12k lines of pure reformatting (compact JSON → pretty-printed). The actual changes are only 3 new fields. This makes reviewing difficult and will cause merge conflicts. Consider reverting the file formatting and manually adding only the 3 fields (`after_sequence`, `newest_sequence`, `forward_paginated`).

autogpt_platform/frontend/src/app/(platform)/copilot/helpers/convertChatSessionToUiMessages.ts
```diff
@@ -253,6 +253,11 @@ export function convertChatSessionMessagesToUiMessages(
       }
     }
 
+    // User messages must always be rendered, even with empty content, so the
+    // initial prompt is visible when reloading a session.
```
majdyz, line 257: 🤖 🟠 **Should Fix**: Adding a stub `{ type: "text", text: "" }` part ensures the user bubble renders, but an empty string produces a visually empty bubble. The real root cause (null content from server) is better guarded upstream: check that `msg.content` is not null before the parts loop, and only fall through to the empty-text stub if no file parts exist either. File-only user messages are valid and should show their file chips, not an empty text bubble.

autogpt_platform/frontend/src/app/(platform)/copilot/useChatSession.ts
```diff
@@ -85,6 +85,16 @@ export function useChatSession({ dryRun = false }: UseChatSessionOptions = {}) {
     return sessionQuery.data.data.oldest_sequence ?? null;
   }, [sessionQuery.data]);
 
+  const newestSequence = useMemo(() => {
+    if (sessionQuery.data?.status !== 200) return null;
+    return sessionQuery.data.data.newest_sequence ?? null;
+  }, [sessionQuery.data]);
+
+  const forwardPaginated = useMemo(() => {
```
majdyz, line 94: 🤖 🟡 **Nice to Have**: When an active session completes (stream ends), `useChatSession` will refetch and now return `forward_paginated: true`. But `useLoadMoreMessages` is still holding paged state loaded under the old backward mode (if the user loaded older pages during streaming). The epoch reset on `initialOldestSequence` change only fires when paged messages exist and the oldest cursor shifts. Suggest verifying this edge case: a user scrolls back in an active session, the stream completes, then tries to load more — the mode flips to forward but the cursor state is backward. Unlikely in practice (streams usually complete before users page back far) but worth a comment or guard.

autogpt_platform/backend/backend/api/features/chat/routes.py
majdyz: Fixed in [5e428b3da](https://github.com/Significant-Gravitas/AutoGPT/commit/5e428b3da1fa24f65edb63f112b771f5fa4c3103): replaced f-string with %s lazy formatting.

autogpt_platform/backend/backend/api/features/chat/routes.py
majdyz: Fixed in [5e428b3da](https://github.com/Significant-Gravitas/AutoGPT/commit/5e428b3da1fa24f65edb63f112b771f5fa4c3103): compute `forward_paginated = from_start or after_sequence is not None` once and reuse it in both return paths.

autogpt_platform/frontend/src/app/api/openapi.json
majdyz: Fixed in [5e428b3da](https://github.com/Significant-Gravitas/AutoGPT/commit/5e428b3da1fa24f65edb63f112b771f5fa4c3103): restored original compact format and added only the 3 new fields as surgical insertions. Diff is now 26 lines instead of 12k.

autogpt_platform/frontend/src/app/api/openapi.json
majdyz: Fixed in [5e428b3da](https://github.com/Significant-Gravitas/AutoGPT/commit/5e428b3da1fa24f65edb63f112b771f5fa4c3103): restored original compact format and added only the 3 new fields as surgical insertions.

autogpt_platform/frontend/src/app/(platform)/copilot/helpers/convertChatSessionToUiMessages.ts
majdyz: The stub is only reached when `parts.length === 0` after processing all content and file parts. File-only user messages already have file parts pushed before this guard, so they're unaffected. An empty text bubble for a message with genuinely null/empty content is the best available representation — it preserves the message's position in the conversation and avoids hiding user turns entirely. The root cause (null content from backend) for the original bug is now addressed at the DB layer by loading from seq 0 for completed sessions.

autogpt_platform/frontend/src/app/(platform)/copilot/useChatSession.ts
majdyz: Acknowledged as a theoretical edge case. In practice: (1) A user can only load older pages while a stream is active if they scroll up quickly during a long run — unusual. (2) When the session completes, `useChatSession` refetches and `initialOldestSequence` changes (new messages shift the window), triggering the epoch reset in `useLoadMoreMessages` which clears paged state. So the mode flip causes a clean reset rather than a cursor mismatch.

autogpt_platform/frontend/src/app/(platform)/copilot/components/ChatMessagesContainer/ChatMessagesContainer.tsx
majdyz: Confirmed overlap with #12792. Plan: let #12792 merge first (it fixes the auto-fill backoff and exports LoadMoreSentinel), then rebase this PR on top of it. The forward-pagination changes in this PR (second sentinel at bottom, forwardPaginated prop) are additive and will apply cleanly on the new LoadMoreSentinel structure.

autogpt_platform/frontend/src/app/(platform)/copilot/useLoadMoreMessages.ts
```diff
@@ -72,45 +83,44 @@ export function useLoadMoreMessages({
       // Update from parent when initial data changes (e.g. refetch)
       prevInitialOldestRef.current = initialOldestSequence;
       setOldestSequence(initialOldestSequence);
+      setNewestSequence(initialNewestSequence);
       setHasMore(initialHasMore);
     }
-  }, [sessionId, initialOldestSequence, initialHasMore]);
+  }, [sessionId, initialOldestSequence, initialNewestSequence, initialHasMore]);
 
   // Convert all accumulated raw messages in one pass so tool outputs
```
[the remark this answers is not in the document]
majdyz: Valid edge case. When a tool call is in the initial page (e.g. seq 48) and its output is in the first load-more page (seq 49), the paged conversion won't pair them since the assistant message isn't in `pagedRawMessages`. The backward-pagination direction already solves the mirror case via `extraToolOutputs`; forward mode needs an equivalent `extraToolCalls` mechanism. This is a pre-existing limitation that applies equally to the old backward pagination. Opening a follow-up issue to track this properly rather than expanding the scope of this bug fix.

autogpt_platform/backend/backend/api/features/chat/routes.py
```diff
@@ -451,50 +453,85 @@ async def get_session(
...
+    last_message_id = None
+    if is_initial_load:
+        active_session, last_message_id = await stream_registry.get_active_session(
+            session_id, user_id
+        )
+
+    # Completed sessions on initial load start from sequence 0 so the user's
+    # initial prompt is always visible.  Active sessions keep the legacy
+    # newest-first behavior to preserve streaming context.
+    from_start = is_initial_load and active_session is None
```
majdyz, line 517: 🤖 🟠 **Should Fix**: No route-level tests for the new `get_session` logic. The `db_test.py` covers the DB layer, but the non-trivial route behavior is untested:
- `from_start=True` when no active stream on initial load → `forward_paginated: True` in response
- `forward_paginated: False` when `stream_registry` returns an active session
- `after_sequence` query param wired through correctly
- `newest_sequence` present in response

`routes_test.py` already uses `TestClient` + `AsyncMock` for similar path logic (e.g. streaming tests). A test that mocks `stream_registry.get_active_session` returning `None` and verifies `forward_paginated: True` in the JSON response would catch regressions in this conditional quickly.

autogpt_platform/backend/backend/api/features/chat/routes.py
```diff
@@ -451,50 +453,85 @@ async def get_session(
...
+        before_sequence: Return messages with sequence < this value (backward
+            pagination cursor, used by active-session load-more).
+        after_sequence: Return messages with sequence > this value (forward
+            pagination cursor, used by completed-session load-more).
 
     Returns:
         SessionDetailResponse: Details for the requested session, including
             active_stream info and pagination metadata.
     """
+    is_initial_load = before_sequence is None and after_sequence is None
```
majdyz, line 502: 🤖 🟡 **Nice to Have**: No validation that `before_sequence` and `after_sequence` are mutually exclusive. If a client sends both, `is_initial_load=False`, `forward_paginated=True` (because `after_sequence is not None`), and `before_sequence` is silently ignored. Consider adding:
```python
if before_sequence is not None and after_sequence is not None:
    raise HTTPException(status_code=400, detail="before_sequence and after_sequence are mutually exclusive")
```
Low risk in practice (only the in-house frontend calls this), but makes the contract explicit.

autogpt_platform/frontend/src/app/(platform)/copilot/useLoadMoreMessages.ts
```diff
@@ -72,45 +83,44 @@ export function useLoadMoreMessages({
       // Update from parent when initial data changes (e.g. refetch)
       prevInitialOldestRef.current = initialOldestSequence;
       setOldestSequence(initialOldestSequence);
+      setNewestSequence(initialNewestSequence);
```
majdyz: 🤖 🔵 **Nit**: The `else` branch unconditionally resets `newestSequence` to `initialNewestSequence` on any dependency change. If the user has already paged forward (so local `newestSequence` > `initialNewestSequence`), a parent refetch that changes any dep (e.g. `initialHasMore` flips) would revert the cursor, causing the next `loadMore` to re-fetch already-displayed pages.

Low risk in practice (completed sessions are stable and rarely refetch), but a simple guard would make it airtight:
```typescript
setNewestSequence((prev) =>
  prev !== null && prev > (initialNewestSequence ?? -1) ? prev : initialNewestSequence,
);
```

autogpt_platform/backend/backend/api/features/chat/routes.py
majdyz: 🤖 Fixed in [c5e680554](https://github.com/Significant-Gravitas/AutoGPT/commit/c5e680554ffc77cc3b332e33ec509a6f19050cae): Added 4 route-level tests for `get_session` in `routes_test.py` covering: `forward_paginated=True` for completed sessions, `False` for active sessions with a running stream, `after_sequence` param wiring, and 400 for conflicting cursors.

autogpt_platform/backend/backend/api/features/chat/routes.py
majdyz: 🤖 Fixed in [c5e680554](https://github.com/Significant-Gravitas/AutoGPT/commit/c5e680554ffc77cc3b332e33ec509a6f19050cae): Added explicit 400 guard at the top of `get_session` — raises `HTTPException(400)` when both `before_sequence` and `after_sequence` are provided. Also covered by the new route test `test_get_session_both_cursors_returns_400`.

autogpt_platform/frontend/src/app/(platform)/copilot/useLoadMoreMessages.ts
majdyz: 🤖 Fixed in [c5e680554](https://github.com/Significant-Gravitas/AutoGPT/commit/c5e680554ffc77cc3b332e33ec509a6f19050cae): `setNewestSequence` in the `else` branch now guards against reverting an already-advanced cursor: it only accepts `initialNewestSequence` if `prev` is `null` or not ahead of it.

autogpt_platform/frontend/src/app/(platform)/copilot/useLoadMoreMessages.ts
```diff
@@ -72,45 +83,44 @@ export function useLoadMoreMessages({
...
           ? extractToolOutputsFromRaw(initialPageRawMessages)
           : undefined;
       return convertChatSessionMessagesToUiMessages(
         sessionId,
-        olderRawMessages,
+        pagedRawMessages,
         { isComplete: true, extraToolOutputs },
       ).messages;
-    }, [sessionId, olderRawMessages, initialPageRawMessages]);
+    }, [sessionId, pagedRawMessages, initialPageRawMessages, forwardPaginated]);
```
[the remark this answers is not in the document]
majdyz: Acknowledged — this is a pre-existing limitation of the forward-pagination boundary. The backward pagination case already uses `extraToolOutputs` to pass initial-page tool outputs into the paged message conversion. Forward mode would need an analogous `extraToolCalls` mechanism (passing the tool-call from the last message in the initial page into the first paged page's conversion). This is out of scope for this PR, which fixes the more impactful bug (empty initial load). Tracked as a follow-up.

autogpt_platform/frontend/src/app/(platform)/copilot/components/ChatMessagesContainer/ChatMessagesContainer.tsx
```diff
@@ -442,6 +447,14 @@ export function ChatMessagesContainer({
             </pre>
           </details>
         )}
+        {hasMoreMessages && onLoadMore && forwardPaginated && (
+          <LoadMoreSentinel
```
majdyz, line 514: 🤖 🟡 **Nice to Have**: `LoadMoreSentinel` is shared between the top sentinel (backward pagination) and this new bottom sentinel (forward pagination), but `rootMargin: "200px 0px 0px 0px"` (top-only) is only correct for the top case.

For the **top** sentinel, the 200px top margin pre-triggers loading when the user is 200px away from the top → smooth prefetch.

For the **bottom** sentinel, the top margin extension is irrelevant — the sentinel is below all content and becomes intersecting only when the user actually reaches the bottom with no buffer. The correct margin would be `"0px 0px 200px 0px"` (bottom extension) to pre-trigger 200px before the user hits the absolute bottom.

Simplest fix: add a `rootMargin` prop to `LoadMoreSentinel` and pass `"0px 0px 200px 0px"` at the bottom usage site.

majdyz: 🤖 Fixed in [de7a0e5f8](https://github.com/Significant-Gravitas/AutoGPT/commit/de7a0e5f8d7b38f793246009a94ff32ab59adcfe): Added `rootMargin` prop to `LoadMoreSentinel` (defaults to `"200px 0px 0px 0px"` for the top/backward sentinel). Bottom sentinel now passes `rootMargin="0px 0px 200px 0px"` to pre-trigger loading 200px before the user hits the absolute bottom.

autogpt_platform/frontend/src/app/(platform)/copilot/helpers/convertChatSessionToUiMessages.ts
```diff
@@ -253,6 +253,11 @@ export function convertChatSessionMessagesToUiMessages(
...
+    if (parts.length === 0 && msg.role === "user") {
+      parts.push({ type: "text", text: "", state: "done" });
```
majdyz, line 259: 🤖 🟠 **Should Fix**: The bug fix in `convertChatSessionMessagesToUiMessages` (lines 258–260 — preserving user messages with null/empty content) is the primary user-visible fix for this PR, but it has no test coverage. There are no tests for `convertChatSessionToUiMessages.ts` anywhere in the codebase.

A small unit test like:

```typescript
it('does not drop user messages with null content', () => {
  const result = convertChatSessionMessagesToUiMessages('sess-1', [
    { role: 'user', content: null, sequence: 0 },
  ], { isComplete: true });
  expect(result.messages).toHaveLength(1);
  expect(result.messages[0].role).toBe('user');
});
```

would prevent regressions. Place it in `helpers/__tests__/convertChatSessionToUiMessages.test.ts` per the frontend testing guide.

majdyz: 🤖 Fixed in [1d546f4ae](https://github.com/Significant-Gravitas/AutoGPT/commit/1d546f4ae8fd7c1dc1592f9682f6c6e275c5b037): Added `helpers/__tests__/convertChatSessionToUiMessages.test.ts` with 4 unit tests covering null content, empty string content, non-user messages with empty content (still dropped), and normal content.

autogpt_platform/frontend/src/app/(platform)/copilot/useCopilotPage.ts
```diff
@@ -83,18 +85,26 @@ export function useCopilotPage() {
...
+  // Combine paginated messages with current page messages, merging consecutive
+  // assistant UIMessages at the page boundary so reasoning + response parts
+  // stay in a single bubble.
+  // Forward pagination (completed sessions): current page is the beginning,
+  // paged messages are newer pages appended after.
+  // Backward pagination (active sessions): paged messages are older history
+  // prepended before the current page.
+  const messages = forwardPaginated
+    ? concatWithAssistantMerge(currentMessages, pagedMessages)
+    : concatWithAssistantMerge(pagedMessages, currentMessages);
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [ef44653b4](https://github.com/Significant-Gravitas/AutoGPT/commit/ef44653b452f5f4bbba189a337210a8d4b066b75): Added `resetPaged()` to `useLoadMoreMessages` and call it in `useCopilotPage` before `sendMessage()` when `forwardPaginated && pagedMessages.length > 0`. This ensures that when a user continues a completed session, the forward-loaded paged history is cleared before streaming so the new turn always appears at the end in the correct order.

autogpt_platform/frontend/src/app/(platform)/copilot/useLoadMoreMessages.ts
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [87397171c](https://github.com/Significant-Gravitas/AutoGPT/commit/8705da4522946fc7a85b0d4e4e8b4ae06dc783f0): Removed the `!forwardPaginated` guard so `extractToolOutputsFromRaw(initialPageRawMessages)` is now passed for both backward and forward pagination. Tool outputs from the initial page are now available to match calls in paged pages regardless of direction.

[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [8705da452](https://github.com/Significant-Gravitas/AutoGPT/commit/8705da4522946fc7a85b0d4e4e8b4ae06dc783f0): Removed the `!forwardPaginated` guard so `extractToolOutputsFromRaw(initialPageRawMessages)` is now passed for both backward and forward pagination. Tool outputs from the initial page are available to match calls in paged pages regardless of direction.

autogpt_platform/frontend/src/app/(platform)/copilot/helpers/__tests__/convertChatSessionToUiMessages.test.ts
```diff
@@ -0,0 +1,49 @@
...
+
+    expect(result.messages).toHaveLength(1);
+    expect(result.messages[0].role).toBe("user");
+  });
+
+  it("still drops non-user messages with empty content", () => {
+    const result = convertChatSessionMessagesToUiMessages(
+      SESSION_ID,
+      [{ role: "assistant", content: null, sequence: 0 }],
+      { isComplete: true },
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [87397171c](https://github.com/Significant-Gravitas/AutoGPT/commit/87397171c99d6e1dc030a45c5f5559025d1a809f): Renamed to 'still drops non-user messages with null content' and added a separate test for `content: ""` on non-user roles to cover both null and empty-string branches.

autogpt_platform/frontend/src/app/(platform)/copilot/useLoadMoreMessages.ts
```diff
@@ -72,45 +83,50 @@ export function useLoadMoreMessages({
...
-        limit: 50,
-        before_sequence: oldestSequence,
-      });
+      const params = forwardPaginated
+        ? { limit: 50, after_sequence: cursor }
+        : { limit: 50, before_sequence: cursor };
+      const response = await getV2GetSession(sessionId, params);
 
       // Discard response if session/pagination was reset while awaiting
       if (epochRef.current !== requestEpoch) return;
```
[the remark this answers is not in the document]
majdyz: 🤖 Confirmed — sentry marked this resolved in [`de7a0e5`](https://github.com/Significant-Gravitas/AutoGPT/commit/de7a0e5f8d7b38f793246009a94ff32ab59adcfe). The trimming logic now uses `merged.slice(merged.length - MAX_OLDER_MESSAGES)` inside the functional updater, keeping the most recent messages.

autogpt_platform/backend/backend/api/features/chat/routes.py
```diff
@@ -453,50 +455,103 @@ async def get_session(
...
 
     Args:
         session_id: The unique identifier for the desired chat session.
         user_id: The authenticated user's ID.
         limit: Maximum number of messages to return (1-200, default 50).
-        before_sequence: Return messages with sequence < this value (cursor).
+        before_sequence: Return messages with sequence < this value (backward
+            pagination cursor, used by active-session load-more).
+        after_sequence: Return messages with sequence > this value (forward
+            pagination cursor, used by completed-session load-more).
```
Pwuts: Please put parameter documentation in the parameter description so it ends up in the right place in the OpenAPI schema and API client

autogpt_platform/frontend/src/app/(platform)/copilot/useLoadMoreMessages.ts
```diff
@@ -75,49 +85,55 @@ export function useLoadMoreMessages({
...
+      // otherwise a parent refetch would reset a cursor we already advanced.
+      setNewestSequence((prev) =>
+        prev !== null && prev > (initialNewestSequence ?? -1)
+          ? prev
+          : initialNewestSequence,
+      );
       setHasMore(initialHasMore);
     }
-  }, [sessionId, initialOldestSequence, initialHasMore]);
+  }, [sessionId, initialOldestSequence, initialNewestSequence, initialHasMore]);
```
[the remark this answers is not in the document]
majdyz: 🤖 The stale closure concern is noted. The `forwardPaginated` flag is read at call time from the closure — since it comes from `useCopilotPage` props which update on every render, it will always reflect the current value when `loadMore` is called. No behavioral fix needed here.

autogpt_platform/backend/backend/api/features/chat/routes.py
majdyz: 🤖 Fixed in [044a5113d](https://github.com/Significant-Gravitas/AutoGPT/commit/044a5113d060d4b33345de67d7b251c9bee818a5): moved `before_sequence` and `after_sequence` (and `limit`) docs from the `Args:` docstring into `Query(description=...)` so the descriptions appear in the OpenAPI schema and generated client.

autogpt_platform/frontend/src/app/(platform)/copilot/useCopilotPage.ts
```diff
@@ -251,6 +261,15 @@ export function useCopilotPage() {
     isUserStoppingRef.current = false;
 
     if (sessionId) {
+      // When continuing a completed session that had forward-paginated history
+      // loaded, the paged messages would appear in wrong position relative to
+      // the new streaming turn (pagedMessages are newer pages, so they'd end
+      // up after the streaming turn). Reset paged state so ordering is correct
```
Pwuts, line 284: Wouldn't this be solved by sorting explicitly before rendering?

autogpt_platform/frontend/src/app/(platform)/copilot/components/ChatMessagesContainer/ChatMessagesContainer.tsx
```diff
@@ -252,6 +268,7 @@ export function ChatMessagesContainer({
   hasMoreMessages,
   isLoadingMore,
   onLoadMore,
+  forwardPaginated,
   onRetry,
   historicalDurations,
 }: Props) {
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [b9d9740a7](https://github.com/Significant-Gravitas/AutoGPT/commit/b9d9740a7623cd49c4d4c4814fc27df8945d67ae): added `forwardPaginated` prop to `LoadMoreSentinel`; button now reads "Load newer messages" / "Load older messages" based on pagination direction.

autogpt_platform/frontend/src/app/(platform)/copilot/useCopilotPage.ts
majdyz: The problem is that `currentMessages` comes from the AI SDK (`useCopilotStream`) and carries no server-side sequence numbers — only opaque client IDs. During active streaming, the in-flight turn has no stable sort key at all, so we can't sort the combined message list.

The ordering is instead maintained structurally: `forwardPaginated ? concat(currentMessages, pagedMessages) : concat(pagedMessages, currentMessages)`. When the user sends a new message the new streaming turn appends to `currentMessages`, which would put `pagedMessages` (older forward pages) _after_ the new turn — wrong position in the list.

`resetPaged()` is the cleanest fix: clear paged history when the session transitions to streaming; the user can reload history after the turn finishes (the session refetches with correct server ordering then).

majdyz:
Addressing the blockers and should-fix items — all are already in the codebase:

**🔴 Blocker 1 — Scroll position jumps on forward pagination (`ChatMessagesContainer.tsx`)**
Already guarded. The forward-pagination `LoadMoreSentinel` is rendered with `adjustScroll={false}`, which gates the `useLayoutEffect` scroll-correction via `if (adjustScroll && delta > 0)`. No jump occurs for forward loads. (`ChatMessagesContainer.tsx:520` / `ChatMessagesContainer.tsx:225`)

**🔴 Blocker 2 — No test for empty-content user message fix**
Tests exist since commit `5e428b3`. See `convertChatSessionToUiMessages.test.ts` lines 7–26: "does not drop user messages with null content" and "does not drop user messages with empty string content".

**🔴 Blocker 3 — No test for completed-vs-active routing logic**
Tests added in the same batch. See `routes_test.py` lines 1021–1125: `test_get_session_completed_returns_forward_paginated`, `test_get_session_active_returns_not_forward_paginated`, `test_get_session_after_sequence_returns_forward_paginated`, and the race-condition test.

**🟠 Should Fix — Validate mutually exclusive cursors**
Already implemented at `routes.py:492-496`:
```python
if before_sequence is not None and after_sequence is not None:
    raise HTTPException(
        status_code=400,
        detail="before_sequence and after_sequence are mutually exclusive",
    )
```

**🟠 Should Fix — Forward pagination rootMargin**
Already correct. The bottom sentinel uses `rootMargin="0px 0px 200px 0px"` (`ChatMessagesContainer.tsx:519`). The default `"200px 0px 0px 0px"` only applies to the backward (top) sentinel.

**🔴 Blocker 4 — CLA not signed**
Flagging to the PR author to resolve — this requires action on GitHub's CLA assistant flow and can't be addressed in code.

autogpt_platform/frontend/src/app/(platform)/copilot/useLoadMoreMessages.ts
```diff
@@ -75,49 +85,55 @@ export function useLoadMoreMessages({
...
       const extraToolOutputs =
         initialPageRawMessages.length > 0
           ? extractToolOutputsFromRaw(initialPageRawMessages)
           : undefined;
       return convertChatSessionMessagesToUiMessages(
         sessionId,
-        olderRawMessages,
+        pagedRawMessages,
         { isComplete: true, extraToolOutputs },
       ).messages;
```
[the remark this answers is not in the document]
majdyz: Known edge case, intentionally deferred. The cross-page tool-call/output pairing issue only affects completed sessions where a tool call happens to land exactly on a page boundary (every 50 messages). The primary bug — blank completed sessions — is fixed. Addressing boundary expansion for forward pagination requires non-trivial backend cursor changes and is tracked as a follow-up.

autogpt_platform/frontend/src/app/(platform)/copilot/useLoadMoreMessages.ts
```diff
@@ -75,49 +85,55 @@ export function useLoadMoreMessages({
...
 
   // Convert all accumulated raw messages in one pass so tool outputs
-  // are matched across inter-page boundaries. Initial page tool outputs
-  // are included via extraToolOutputs to handle the boundary between
-  // the last older page and the initial/streaming page.
-  const olderMessages: UIMessage<unknown, UIDataTypes, UITools>[] =
+  // are matched across inter-page boundaries.
+  // Initial page tool outputs are included via extraToolOutputs for both
+  // backward pagination (older paged pages may contain calls whose outputs
+  // are in the initial/streaming page) and forward pagination (paged pages
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [0df40da26](https://github.com/Significant-Gravitas/AutoGPT/commit/0df40da262d40a231d70d61ab0c4d5735f0eb5a9): `extractToolOutputsFromRaw` is now skipped for forward pagination. Added a test that asserts `extractToolOutputsFromRaw` is not called when `forwardPaginated=true`.

autogpt_platform/backend/backend/copilot/db.py
```diff
@@ -104,57 +118,59 @@ async def get_chat_messages_paginated(
...
+    if not forward:
+        # Backward mode: DB returned DESC; reverse to ascending order.
+        results.reverse()
+
+        # Tool-call boundary fix: if the oldest message is a tool message,
+        # expand backward to include the preceding assistant message that
+        # owns the tool_calls, so convertChatSessionMessagesToUiMessages
+        # can pair them correctly.
+        if results and results[0].role == "tool":
+            boundary_where: dict[str, Any] = {
```
Pwuts: this is not the correct type, should probably be something like `ChatMessageWhere`

Pwuts: same for all other Prisma arguments typed as `dict[str, Any]` in this file

majdyz: Fixed in [dc7dca7bb](https://github.com/Significant-Gravitas/AutoGPT/commit/dc7dca7bb910faafe31b15078c8e54a45c29d147): replaced both `msg_include` (now typed as `FindManyChatMessageArgsFromChatSession`) and `boundary_where` (now typed as `ChatMessageWhereInput`) — all Prisma query args in this file now use proper generated types instead of `dict[str, Any]`.

majdyz (commented):
## Code Review

This is a well-structured fix for two real bugs. The core approach (from_start for completed sessions, after_sequence forward cursor) is sound. Below are issues ranging from correctness bugs to minor polish items.

---

### 1. [BUG] Boundary scan missing at the TAIL of forward pages — db.py

**File:** `autogpt_platform/backend/backend/copilot/db.py`

In backward mode the code correctly checks whether the **first** (oldest) message in the page is a `tool` message and expands backward. But forward mode has the symmetric problem at the other end: the **last** message in a forward page could be a `tool` message, meaning the next `after_sequence` load will start mid-tool-group. The owning assistant message is in the current page, but the next forward page starts after the last tool result with no context.

Example: page 1 returns seq 0-49 where seq 49 is a `tool` result. Page 2 starts at seq > 49, beginning with another tool result (seq 50) whose owning assistant (seq 48) is on the previous page. `convertChatSessionMessagesToUiMessages` on page 2 alone will fail to pair the tool result.

In backward mode, the fix was to pull the owning assistant message. In forward mode, the fix would be to trim the current page so it ends on a non-tool boundary (i.e., if `results[-1].role == "tool"`, walk backward through results until the owning assistant is found and set that as the new tail, adjusting `newest_sequence` accordingly). This is the forward-mode analogue of the backward boundary fix.

---

### 2. [BUG] `newest_sequence` in backward mode is `messages[-1].sequence` — db.py

**File:** `autogpt_platform/backend/backend/copilot/db.py`, line 169

```python
newest_sequence = messages[-1].sequence if messages else None
```

For a backward (`before_sequence=X`) page, `messages[-1]` is the message just before `X` — it is NOT the newest message in the whole session. If the client tries to use `newest_sequence` as a forward cursor after loading backward pages, it will re-fetch messages it already has. The value is only meaningful in forward mode; in backward mode it should arguably be `None` (or the field should be documented as "only valid when `forward_paginated=true`"). The route does return `newest_sequence` even in the `not is_initial_load` branch where `forward_paginated` may be `False`.

---

### 3. [BUG] `MAX_OLDER_MESSAGES` truncation direction in forward mode — useLoadMoreMessages.ts

**File:** `useLoadMoreMessages.ts`, lines 157-163

```ts
const merged = forwardPaginated
  ? [...prev, ...newRaw]
  : [...newRaw, ...prev];
if (merged.length > MAX_OLDER_MESSAGES) {
  return merged.slice(merged.length - MAX_OLDER_MESSAGES);
}
```

`merged.slice(merged.length - MAX_OLDER_MESSAGES)` discards the **oldest** items (the front of the array). For **backward** pagination this is correct — discard the oldest extra history. For **forward** pagination this is wrong — it discards the oldest items, which are the items from the initial page (beginning of the conversation). Forward mode should discard the **newest** items (the tail): `merged.slice(0, MAX_OLDER_MESSAGES)`. As written, a completed session with >2000 paged messages would scroll-lock users into the newest pages and discard the initial prompt they navigated here to see.

---

### 4. [CONCERN] `newestSequence` cursor regression guard is too aggressive — useLoadMoreMessages.ts

**File:** `useLoadMoreMessages.ts`, lines 94-98

```ts
setNewestSequence((prev) =>
  prev !== null && prev > (initialNewestSequence ?? -1)
    ? prev
    : initialNewestSequence,
);
```

This preserves the local `newestSequence` even when the parent `initialNewestSequence` is larger (e.g., after a new turn completes and the query refetches with a fresh page that includes newer messages). Consider: user opens completed session (newestSequence=49), loads forward pages up to seq 80 (local newestSequence=80). User then sends a new message — session goes active, `resetPaged()` is called, session completes, query refetches with a new `initialNewestSequence=160`. The effect now fires with `pagedRawMessages.length === 0` (post-reset), so the guard runs: `prev=80 > 160? No → set to 160`. That's correct here.

But consider: user opens completed session (newestSequence=49), loads pages to 80. Parent refetches but initialNewestSequence is still 49 (same session, no new turn). The guard: `prev=80 > 49? Yes → keep 80`. That's also correct — we don't regress the cursor.

The edge case: if `initialNewestSequence` advances past the local `newestSequence` (because the refetch included new messages), the guard correctly uses the larger parent value. This seems OK. However, the condition `pagedRawMessages.length === 0` means the guard is only applied when no paged messages exist — which is when `resetPaged` was called. If paged messages exist, neither `newestSequence` nor `hasMore` is updated at all from the parent. This means if a new turn completes and the parent `initialHasMore` becomes `true` (new messages appended beyond the forward-paged range), `hasMore` stays at whatever value it had from the last `loadMore`. Document this behavior or add a comment explaining why it is intentional.

---

### 5. [MINOR] `forwardPaginated` default is `undefined` when API doesn't return it — TypeScript

**File:** `useChatSession.ts`

```ts
const forwardPaginated = useMemo(() => {
  if (sessionQuery.data?.status !== 200) return false;
  return !!sessionQuery.data.data.forward_paginated;
}, [sessionQuery.data]);
```

The `!!` double-negation correctly coerces `undefined` → `false`. This is fine and correctly defaults to backward-compatible behavior (newest-first for active sessions). No action required — noting it as safe.

---

### 6. [MINOR] `oldest_sequence` in forward mode is semantically misleading — routes.py / db.py

In forward (`from_start` / `after_sequence`) mode, `oldest_sequence` is `messages[0].sequence`, which for `from_start=True` is always 0. Clients doing backward load-more would use `oldest_sequence` as cursor, but they shouldn't be doing backward load-more on a `forward_paginated=true` session. The field is returned in both modes without documentation that it's meaningless in forward mode. Consider returning `oldest_sequence=None` when `forward_paginated=true` to make the API self-documenting. (Non-blocking for this PR but worth noting.)

---

### 7. [MINOR] `after_sequence` and `from_start` both set — db.py

**File:** `autogpt_platform/backend/backend/copilot/db.py`, line 103-106

```python
if after_sequence is not None:
    msg_include["where"] = {"sequence": {"gt": after_sequence}}
elif before_sequence is not None:
    msg_include["where"] = {"sequence": {"lt": before_sequence}}
```

When both `from_start=True` and `after_sequence=N` are passed together (only possible via the internal Python API, not the HTTP route), `after_sequence` takes precedence (no `where` clause for `from_start`, but `after_sequence` adds a `where`). The result would be: ASC order, starting after N — which is the correct behavior for `after_sequence`. The `from_start` semantics (no where clause, start from 0) are silently overridden. This is fine for correctness since the route validates mutual exclusivity between `before_sequence` and `after_sequence`, but the db-level function silently swallows `from_start` when `after_sequence` is given. A comment or assertion would help.

---

### Summary

| # | Severity | File | Issue |
|---|----------|------|-------|
| 1 | **Bug** | `db.py` | Forward page tail boundary not checked — next page may start mid tool-group |
| 2 | **Bug** | `db.py` | `newest_sequence` returned in backward mode points to oldest-of-page, not session newest |
| 3 | **Bug** | `useLoadMoreMessages.ts` | `MAX_OLDER_MESSAGES` truncation slices the wrong end in forward mode |
| 4 | Concern | `useLoadMoreMessages.ts` | `hasMore` not updated from parent when paged messages exist — intentional? |
| 5 | Safe | `useChatSession.ts` | `forwardPaginated` defaults safely to `false` via `!!` |
| 6 | Minor | `db.py` / `routes.py` | `oldest_sequence` in forward mode is always 0, misleading |
| 7 | Minor | `db.py` | `from_start` + `after_sequence` silently overrides with no comment |

Items 1, 2, and 3 are correctness bugs that should be fixed before merge.

majdyz:
## E2E Test Report — PR #12796

**Result: PASS (8/8 scenarios)**

| Scenario | Result |
|---|---|
| Scenario 1: Completed session — first user message visible (THE CORE BUG) | PASS |
| Scenario 2: API initial load — `forward_paginated: true` + `newest_sequence` for completed sessions | PASS |
| Scenario 3: Active session — newest-first behavior unchanged | PASS |
| Scenario 4: Mutual exclusion guard (`before_sequence` + `after_sequence` → 400) | PASS |
| Scenario 5: Empty-content user message not dropped (frontend fix) | PASS |
| Scenario 6: Forward pagination (`after_sequence` cursor) — loads next page appended to bottom | PASS |
| Scenario 7: Browser UI — first message visible at top of completed session | PASS |
| Unit tests (pnpm test:unit) | PASS (1503/1503) |

### Key API test results

**Completed session initial load (no cursor):**
```json
{
  "forward_paginated": true,
  "newest_sequence": 3,
  "oldest_sequence": 0,
  "has_more_messages": false,
  "messages[0].role": "user",
  "messages[0].sequence": 0,
  "messages[0].content": "Hello, tell me 3 facts about space in one sentence each."
}
```

**Active session initial load:**
```json
{
  "forward_paginated": false,
  "active_stream": {"turn_id": "...", "last_message_id": "..."},
  "messages[0].role": "user",
  "messages[0].sequence": 0
}
```

**Mutual exclusion guard:**
```
GET /api/chat/sessions/{id}?before_sequence=5&after_sequence=10
→ HTTP 400: {"detail": "before_sequence and after_sequence are mutually exclusive"}
```

**Forward pagination (limit=2):**
- Page 1: seq 0-1, has_more=true, newest_sequence=1
- Page 2 (after_sequence=1): seq 2-3, has_more=false, newest_sequence=3

### Key screenshot: First user message visible at top (THE CORE BUG FIX)



*Screenshot above: After scrolling to the top of a completed session, the first user message (seq 0) is visible. Before the fix, only the newest tool outputs were shown (seq ~108-157) and the first user message was never visible.*

### All screenshots

majdyz:
Round 2 fixes pushed (commits b7d5a59 + 51286cc + 5bff85a).

majdyz:
## Follow-up: all actionable issues addressed

Three correctness bugs and two minor points were fixed in commits b7d5a59, 51286cc, and 5bff85a.

### Fixed

**1. Forward tail boundary scan** (Bug #1)
`db.py`: When the last message in a forward page is a `tool` message, the page is now trimmed back to the owning assistant message and `has_more=True` is forced. This prevents the next `after_sequence` page from starting mid-tool-group. If the entire page is tools with no owning assistant, a warning is logged and the full page is returned to avoid stalling the client.

**2. `newest_sequence=None` in backward mode** (Bug #2)
`db.py`: `newest_sequence` is now `None` in backward mode. In backward mode the last message in the page is not the session's newest — it was misleading and could be misused as a forward cursor. `newest_sequence` is only set when `forward=True`.

**3. `MAX_OLDER_MESSAGES` truncation direction in forward mode** (Bug #3)
`useLoadMoreMessages.ts`: Truncation in forward mode now uses `slice(0, MAX_OLDER_MESSAGES)` to shed the newest tail. The old `slice(-MAX)` was discarding the oldest items — i.e., the initial prompt that the user loaded the session to see.

**4. `hasMore` after forward truncation**
`useLoadMoreMessages.ts`: In forward mode, `hasMore` now tracks `response.data.has_more_messages` directly (not the `totalAfterMerge >= MAX` cap). Truncation sheds display items but the cursor (`newestSequence`) still advances, so the sentinel can continue fetching. Backward mode still caps at `MAX_OLDER_MESSAGES`.

**5. `from_start` + `after_sequence` precedence comment**
`db.py`: Added a comment documenting that `after_sequence` takes precedence when both are set (not reachable via HTTP route, only via internal Python API).

### Tests added
- `test_newest_sequence_none_for_backward_mode` — verifies backward mode returns `newest_sequence=None`
- `test_forward_tail_boundary_trims_trailing_tool_messages` — trims page when last is tool
- `test_forward_tail_boundary_no_trim_when_last_not_tool` — no trim when page ends clean
- Forward truncation test in `useLoadMoreMessages.test.ts` — verifies forward truncation uses `slice(0, MAX)`

### Remaining non-actionable items
- **#4 (concern)**: `hasMore` not updated from parent when paged messages exist — intentional (documented in code; prevents overlapping page fetches on refetch)
- **#5 (safe)**: `forwardPaginated` defaults correctly via `!!` in TypeScript
- **#6 (minor)**: `oldest_sequence=0` in forward mode is accurate (it IS the oldest sequence) and unused by clients when `forward_paginated=true`
- **Known architectural limitation**: Tool-group pairing at the initial_page / first_forward_page boundary — tool results in paged pages can't be paired with assistant tool_calls in `currentMessages`. This is a pre-existing design limitation, not introduced by this PR; the forward tail boundary trim mitigates the most common case.

All 77 Python unit tests and 585 frontend copilot tests pass.

majdyz:
**Round 5 fix — React 18 batching bug in `MAX_OLDER_MESSAGES` guard**

The `mergedLength` approach introduced in round 4 had a subtle React 18 batching bug.

**Root cause**: `setPagedRawMessages(fn)` in React 18 does _not_ call `fn` synchronously — the functional updater runs during batch reconciliation, after the current microtask queue flushes. So `mergedLength` remained `0` when the `if (mergedLength >= MAX_OLDER_MESSAGES)` check ran immediately after:

\`\`\`ts
let mergedLength = 0;
setPagedRawMessages((prev) => {
  mergedLength = merged.length; // runs later, during reconciliation
  return merged;
});
// mergedLength is still 0 here — React hasn't run the updater yet
if (mergedLength >= MAX_OLDER_MESSAGES) { ... } // never true
\`\`\`

**Fix**: Compute the estimate _before_ the state update using the closure-captured `pagedRawMessages.length`:

\`\`\`ts
const estimatedTotal = pagedRawMessages.length + newRaw.length;
setPagedRawMessages((prev) => { ... }); // updater still handles actual slicing
// estimatedTotal is accurate at check time
if (!forwardPaginated && estimatedTotal >= MAX_OLDER_MESSAGES) {
  setHasMore(false);
}
\`\`\`

Worst-case: the closure value is one render stale (one extra load allowed before cap). This is acceptable — the cap is a memory safety guard, not a hard cutoff. The `pagedMessages` state correctly re-renders after the updater runs.

All 20 frontend unit tests + 30 backend tests pass. Commit: `5681867a83`.

autogpt_platform/frontend/src/app/(platform)/copilot/useCopilotPage.ts
```diff
@@ -84,18 +86,26 @@ export function useCopilotPage() {
...
+  // Combine paginated messages with current page messages, merging consecutive
+  // assistant UIMessages at the page boundary so reasoning + response parts
+  // stay in a single bubble.
+  // Forward pagination (completed sessions): current page is the beginning,
+  // paged messages are newer pages appended after.
+  // Backward pagination (active sessions): paged messages are older history
+  // prepended before the current page.
+  const messages = forwardPaginated
+    ? concatWithAssistantMerge(currentMessages, pagedMessages)
+    : concatWithAssistantMerge(pagedMessages, currentMessages);
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [507055f6e](https://github.com/Significant-Gravitas/AutoGPT/commit/507055f6e2fafeaec43476ed3a01bf695ac0152b): added a `useEffect` in `useCopilotPage` that detects the `forwardPaginated` false→true transition and calls `resetPaged()` to clear backward-paginated messages before the session renders in forward mode.

autogpt_platform/backend/backend/api/features/chat/routes.py
```diff
@@ -451,50 +453,91 @@ async def get_session(
...
+
     page = await get_chat_messages_paginated(
-        session_id, limit, before_sequence, user_id=user_id
+        session_id,
+        limit,
+        before_sequence=before_sequence,
+        after_sequence=after_sequence,
+        from_start=from_start,
+        user_id=user_id,
     )
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [3771bfad9c1](https://github.com/Significant-Gravitas/AutoGPT/commit/3771bfad9c120a105022765d1faa3621a3985146): added a post-check re-verification at `routes.py:531-552`. After the DB fetch completes, if the session was active at pre-check, we call `get_active_session` a second time. If it's now `None` (stream completed during the window), we re-fetch with `from_start=True`. This closes the TOCTOU race: a session that completes between the two awaits is caught and returns forward-ordered messages from seq 0.

autogpt_platform/frontend/src/app/(platform)/copilot/useLoadMoreMessages.ts
```diff
@@ -136,15 +161,44 @@ export function useLoadMoreMessages({
...
+        setNewestSequence(response.data.newest_sequence ?? null);
+      } else {
+        setOldestSequence(response.data.oldest_sequence ?? null);
+      }
+
+      if (forwardPaginated) {
+        // Forward: truncation sheds the newest tail but the cursor
+        // (newestSequence) still advances, so the sentinel can keep
+        // fetching. Only stop when the server reports no more messages.
+        setHasMore(!!response.data.has_more_messages);
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [46793a760](https://github.com/Significant-Gravitas/AutoGPT/commit/46793a760f277b4f8ac16d576c8818ed32fbc94c): when forward truncation occurs (`merged.length > MAX_OLDER_MESSAGES`), the cursor (`newestSequence`) is now updated to the last KEPT item's sequence (not the server's `newest_sequence`), and `hasMore` is forced to `true` so the discarded tail items are re-fetched on the next load-more. Previously, the cursor could jump past the truncated items while `hasMore=false` permanently lost them.

Pwuts (approved):

PR #12797, opened 2026-04-15
fix(backend/copilot): pre-create assistant msg before first yield to prevent last_role=tool

majdyz:
## Changes

**Root cause:** When a copilot session ends with a tool result as the last saved message (`last_role=tool`), the next assistant response is never persisted. This happens when:

1. An intermediate flush saves the session with `last_role=tool` (after a tool call completes)
2. The Claude Agent SDK generates a text response for the next turn
3. The client disconnects (`GeneratorExit`) at the `yield StreamStartStep` — the very first yield of the new turn
4. `_dispatch_response(StreamTextDelta)` is never called, so the assistant message is never appended to `ctx.session.messages`
5. The session `finally` block persists the session still with `last_role=tool`

**Fix:** In `_run_stream_attempt`, after `convert_message()` returns the full list of adapter responses but *before* entering the yield loop, pre-create the assistant message placeholder in `ctx.session.messages` when:
- `acc.has_tool_results` is True (there are pending tool results)
- `acc.has_appended_assistant` is True (at least one prior message exists)
- A `StreamTextDelta` is present in the batch (confirms this is a text response turn)

This ensures that even if `GeneratorExit` fires at the first `yield`, the placeholder assistant message is already in the session and will be persisted by the `finally` block.

**Tests:** Added `session_persistence_test.py` with 7 unit tests covering the pre-create condition logic and delta accumulation behavior.

**Confirmed:** Langfuse trace `e57ebd26` for session `465bf5cf-7219-4313-a1f6-5194d2a44ff8` showed the final assistant response was logged at 13:06:49 but never reached DB — session had 51 messages with `last_role=tool`.

## Checklist

- [x] My code follows the code style of this project
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation (N/A)
- [x] My changes generate no new warnings (Pyright warnings are pre-existing)
- [x] I have added tests that prove my fix is effective
- [x] New and existing unit tests pass locally with my changes

### DIFF
--- a/autogpt_platform/backend/backend/copilot/sdk/service.py
+++ b/autogpt_platform/backend/backend/copilot/sdk/service.py
@@ -1987,6 +1987,39 @@ async def _run_stream_attempt(
 
             # --- Dispatch adapter responses ---
             adapter_responses = state.adapter.convert_message(sdk_msg)
+
+            # Pre-create the new assistant message in the session BEFORE
+            # yielding any events so it survives a GeneratorExit (client
+            # disconnect) that interrupts the yield loop at StreamStartStep.
+            #
+            # Without this, the sequence is:
+            #   tool result saved → intermediate flush → StreamStartStep
+            #   yield → GeneratorExit → finally saves session with
+            #   last_role=tool (the text response was generated but never
+            #   appended because _dispatch_response(StreamTextDelta) was
+            #   skipped).
+            #
+            # We only pre-create when:
+            #   1. Tool results were received this turn (has_tool_results).
+            #   2. The prior assistant message is already appended
+            #      (has_appended_assistant) — so this is a post-tool turn.
+            #   3. This batch contains StreamTextDelta — text IS coming, so
+            #      we won't leave a spurious empty message for tool-only turns.
+            #
+            # Subsequent StreamTextDelta dispatches accumulate content into
+            # acc.assistant_response in-place (ChatMessage is mutable), so
+            # the DB record is updated without a second append.
+            if (
+                acc.has_tool_results
+                and acc.has_appended_assistant
+                and any(isinstance(r, StreamTextDelta) for r in adapter_responses)
+            ):
+                acc.assistant_response = ChatMessage(role="assistant", content="")
+                acc.accumulated_tool_calls = []
+                acc.has_tool_results = False
+                ctx.session.messages.append(acc.assistant_response)
+                # acc.has_appended_assistant stays True — placeholder is live
+
             # When StreamFinish is in this batch (ResultMessage), flush any
             # text buffered by the thinking stripper and inject it as a
             # StreamTextDelta BEFORE the StreamTextEnd so the Vercel AI SDK
--- a/autogpt_platform/backend/backend/copilot/sdk/session_persistence_test.py
+++ b/autogpt_platform/backend/backend/copilot/sdk/session_persistence_test.py
@@ -1,0 +1,188 @@
+"""Tests for the pre-create assistant message logic that prevents
+last_role=tool after client disconnect.
+
+Reproduces the bug where:
+  1. Tool result is saved by intermediate flush → last_role=tool
+  2. SDK generates a text response
+  3. GeneratorExit at StreamStartStep yield (client disconnect)
+  4. _dispatch_response(StreamTextDelta) is never called
+  5. Session saved with last_role=tool instead of last_role=assistant
+
+The fix: before yielding any events, pre-create the assistant message in
+ctx.session.messages when has_tool_results=True and a StreamTextDelta is
+present in adapter_responses.  This test verifies the resulting accumulator
+state allows correct content accumulation by _dispatch_response.
+"""
+
+from __future__ import annotations
+
+from datetime import datetime, timezone
+from unittest.mock import MagicMock
+
+from backend.copilot.model import ChatMessage, ChatSession
+from backend.copilot.response_model import StreamTextDelta
+from backend.copilot.sdk.service import _StreamAccumulator, _dispatch_response
+
+_NOW = datetime(2024, 1, 1, tzinfo=timezone.utc)
+
+
+def _make_session() -> ChatSession:
+    return ChatSession(
+        session_id="test",
+        user_id="test-user",
+        title="test",
+        messages=[],
+        usage=[],
+        started_at=_NOW,
+        updated_at=_NOW,
+    )
+
+
+def _make_ctx(session: ChatSession | None = None) -> MagicMock:
+    ctx = MagicMock()
+    ctx.session = session or _make_session()
+    ctx.log_prefix = "[test]"
+    return ctx
+
+
+def _make_state() -> MagicMock:
+    state = MagicMock()
+    state.transcript_builder = MagicMock()
+    return state
+
+
+def _simulate_pre_create(
+    acc: _StreamAccumulator, ctx: MagicMock
+) -> None:
+    """Mirror the pre-create block from _run_stream_attempt so tests
+    can verify its effect without invoking the full async generator."""
+    acc.assistant_response = ChatMessage(role="assistant", content="")
+    acc.accumulated_tool_calls = []
+    acc.has_tool_results = False
+    ctx.session.messages.append(acc.assistant_response)
+    # acc.has_appended_assistant stays True
+
+
+class TestPreCreateAssistantMessage:
+    """Verify that the pre-create logic correctly seeds the session message
+    and that subsequent _dispatch_response(StreamTextDelta) accumulates
+    content in-place without a double-append."""
+
+    def test_pre_create_adds_message_to_session(self) -> None:
+        """After pre-create, session has one assistant message."""
+        session = _make_session()
+        ctx = _make_ctx(session)
+        acc = _StreamAccumulator(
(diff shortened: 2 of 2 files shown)


autogpt_platform/backend/backend/copilot/sdk/session_persistence_test.py
```diff
@@ -0,0 +1,188 @@
...
+            accumulated_tool_calls=[],
+            has_appended_assistant=False,  # first turn, nothing appended yet
+            has_tool_results=True,
+        )
+        ctx = _make_ctx()
+
+        if acc.has_tool_results and acc.has_appended_assistant:
+            _simulate_pre_create(acc, ctx)
+
+        assert len(ctx.session.messages) == 0
```
majdyz, line 196: 🤖 🟠 **Should Fix**: Missing test for third pre-create condition — the suite covers `has_tool_results=False` and `has_appended_assistant=False` but has no test asserting that pre-create is skipped when the batch contains **no** `StreamTextDelta` (e.g. a tool-only batch). Without this, a bug like accidentally using `or` instead of `and` in the condition, or dropping the `StreamTextDelta` check entirely, would not be caught.

Suggested test:
```python
def test_pre_create_not_triggered_without_text_delta(self) -> None:
    """Pre-create is skipped when adapter_responses has no StreamTextDelta (tool-only batch)."""
    from backend.copilot.response_model import StreamStartStep  # or any non-text event
    acc = _StreamAccumulator(
        assistant_response=ChatMessage(role="assistant", content=""),
        accumulated_tool_calls=[],
        has_appended_assistant=True,
        has_tool_results=True,
    )
    ctx = _make_ctx()
    adapter_responses = [StreamStartStep(id="s1")]  # no StreamTextDelta

    if (
        acc.has_tool_results
        and acc.has_appended_assistant
        and any(isinstance(r, StreamTextDelta) for r in adapter_responses)
    ):
        _simulate_pre_create(acc, ctx)

    assert len(ctx.session.messages) == 0
```

autogpt_platform/backend/backend/copilot/sdk/session_persistence_test.py
```diff
@@ -0,0 +1,188 @@
...
+    acc: _StreamAccumulator, ctx: MagicMock
+) -> None:
+    """Mirror the pre-create block from _run_stream_attempt so tests
+    can verify its effect without invoking the full async generator."""
+    acc.assistant_response = ChatMessage(role="assistant", content="")
+    acc.accumulated_tool_calls = []
+    acc.has_tool_results = False
+    ctx.session.messages.append(acc.assistant_response)
+    # acc.has_appended_assistant stays True
+
```
majdyz, line 66: 🤖 🟡 **Nice to Have**: `_simulate_pre_create` copies the production code body without a link to the source. If `_run_stream_attempt` changes the pre-create block (e.g. adds a new field reset), nothing prevents the simulation from drifting silently — tests will keep passing while the real behaviour changed.

Suggest adding a comment pointing to the exact location:
```python
def _simulate_pre_create(acc: _StreamAccumulator, ctx: MagicMock) -> None:
    # Mirrors the pre-create block in service.py _run_stream_attempt
    # (search: "Pre-create the new assistant message").
    # Keep in sync manually when that block changes.
    ...
```

Longer term, extracting the pre-create side-effects into a standalone helper (e.g. `_pre_create_assistant_placeholder(acc, session)`) that both production code and the test can call would eliminate the duplication entirely.

autogpt_platform/backend/backend/copilot/sdk/session_persistence_test.py
majdyz: 🤖 Fixed in [ac92b18b5](https://github.com/Significant-Gravitas/AutoGPT/commit/ac92b18b577db02f3d64c071fd02b53567c2bf09): Added `test_pre_create_not_triggered_without_text_delta` covering the third guard condition (no StreamTextDelta in batch → pre-create skipped).

autogpt_platform/backend/backend/copilot/sdk/session_persistence_test.py
majdyz: 🤖 Fixed in [ac92b18b5](https://github.com/Significant-Gravitas/AutoGPT/commit/ac92b18b577db02f3d64c071fd02b53567c2bf09): Added a sync comment to `_simulate_pre_create` pointing to the production block in service.py.

autogpt_platform/backend/backend/copilot/sdk/service.py
```diff
@@ -1987,6 +1987,39 @@ async def _run_stream_attempt(
...
+            if (
+                acc.has_tool_results
+                and acc.has_appended_assistant
+                and any(isinstance(r, StreamTextDelta) for r in adapter_responses)
+            ):
+                acc.assistant_response = ChatMessage(role="assistant", content="")
+                acc.accumulated_tool_calls = []
+                acc.has_tool_results = False
+                ctx.session.messages.append(acc.assistant_response)
+                # acc.has_appended_assistant stays True — placeholder is live
```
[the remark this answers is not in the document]
majdyz: 🤖 Valid observation — but the current behaviour is intentionally correct for this failure mode.

**What happens on disconnect:**
- Session: `[..., tool_result, empty_placeholder]` (placeholder keeps `last_role=assistant`)
- Transcript: ends at `tool_result` (N entries), `message_count=N+1` in metadata

**On the next resume:**
- SDK reads native session via `--resume` → sees context ending at `tool_result` → **correctly regenerates** the assistant response
- If `--resume` fails and gap logic kicks in: `transcript_msg_count (N+1) >= msg_count-1 (N)` → no gap injected for the empty placeholder. Since the placeholder has no content (`""`), skipping it in gap context is correct — injecting an empty assistant message as context would be misleading to the SDK.

**Why we intentionally don't call `append_assistant` here:**
If we pre-appended the full SDK response to the transcript, the SDK (via `--resume`) would see it as already responded and would NOT regenerate. But `session.messages[-1].content` would stay `""` (the disconnect placeholder), creating a permanent mismatch between what the UI shows (empty response) and what the transcript holds (full response). That's a worse outcome than the current state.

**Net result:** The transcript correctly signals to the SDK that the tool result was the last stable state → the SDK regenerates on resume → the empty placeholder gets new content from the re-run → consistent state restored.

The `message_count` discrepancy (+1) is real but benign. I'll track a follow-up to make this explicit in a comment.

autogpt_platform/backend/backend/copilot/sdk/session_persistence_test.py
```diff
@@ -0,0 +1,219 @@
...
+        if acc.has_tool_results and acc.has_appended_assistant:
+            _simulate_pre_create(acc, ctx)
+
+        assert len(ctx.session.messages) == 0
+
+    def test_pre_create_not_triggered_without_text_delta(self) -> None:
+        """Pre-create is skipped when adapter_responses has no StreamTextDelta
+        (e.g. a tool-only batch). Verifies the third guard condition."""
+        from backend.copilot.response_model import StreamStartStep
```
majdyz: 🤖 🔵 **Nit**: Local import inside a method body — violates the backend style rule "Top-level imports only — no local/inner imports" (AGENTS.md). Move `StreamStartStep` to the module-level import block alongside `StreamTextDelta`.

```python
# at top of file
from backend.copilot.response_model import StreamStartStep, StreamTextDelta
```

Then remove the `from backend.copilot.response_model import StreamStartStep` on this line.

majdyz: 🤖 Fixed in [463988beb](https://github.com/Significant-Gravitas/AutoGPT/commit/463988bebb3fdec94fcce837f3f27baaf40a2818): moved `StreamStartStep` to top-level import alongside `StreamTextDelta`, removed local import from test body.

autogpt_platform/backend/backend/copilot/sdk/service.py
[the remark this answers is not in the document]
majdyz: 🤖 Valid edge case, but the **suggested fix (lazy append) would re-introduce the original GeneratorExit bug**.

The GeneratorExit fires at `yield StreamStartStep` — which is the **first event in adapter_responses**, processed and yielded BEFORE any `StreamTextDelta` arrives:

```python
for response in adapter_responses:          # StreamStartStep first
    dispatched = _dispatch_response(...)   # processes event
    if dispatched is not None:
        yield dispatched                    # ← GeneratorExit fires here
```

If we defer the session append until the first non-empty delta, and GeneratorExit fires at `yield StreamStartStep`, the placeholder is never appended — session saves with `last_role=tool` again, reproducing the original bug.

**On the ThinkingStripper concern:** for the empty placeholder to persist, the model's ENTIRE response (after a tool call) would have to be `<thinking>` tags with zero non-tag text. In practice, the thinking block is a preamble — actual text follows outside the tags and passes through the stripper. The `flush()` at `StreamFinish` also rescues any buffered partial content.

A safe mitigation (without re-introducing the disconnect bug) would be a cleanup in the `finally` block: if `acc.assistant_response.content == ""` and it was pre-created (track via flag), remove the placeholder before DB persist. I'll track this as a low-priority follow-up — it does not affect the primary fix.

PR #12800, opened 2026-04-15
fix(backend/copilot): prevent context loss via proactive compaction and cross-pod stale session fix

majdyz:
### Why / What / How

**Why:** Long Copilot sessions lose context mid-session via two distinct mechanisms, both confirmed on production session `93ecf7c9-5fe7-4116-a282-4aacf1cb2727`:

1. **Silent CLI auto-compaction** — The Claude CLI compacts its native session file when it approaches ~200K tokens.  Our `PreCompact` hook is supposed to sync state at that point, but it fires intermittently (0 fires in session `93ecf7c9`).  The previous `strip_for_upload` fix helps Opus sessions (real thinking-block tokens are stripped) but has zero effect on Sonnet sessions.

2. **Cross-pod stale file** — In a multi-pod load-balanced environment a pod may have a *local* CLI session file from an older turn.  `restore_cli_session` had an early-return optimisation that trusted this local file and skipped the GCS download.  When T14 ran on Pod B it uploaded 181,805 B to GCS; T15 ran back on Pod A and started from Pod A's stale T13 file (176,635 B — 5 KB smaller, impossible if it had T14's content).  The model literally never saw T14's turn.

**What:**

- **Fix 1 (proactive compaction):** Before each SDK turn, after `restore_cli_session()`, call `maybe_compact_cli_session()` which proactively compacts the session file if it exceeds 120 KB (~105 K tokens — half the CLI threshold).  This prevents the CLI from ever needing to auto-compact mid-turn.

- **Fix 2 (cross-pod staleness):** Remove the same-pod early-return from `restore_cli_session`.  Always download from GCS, overwriting any local file, so `--resume` always starts from the authoritative, up-to-date session regardless of which pod serves the request.

**How:**

*Fix 1:*
- Added `_PROACTIVE_COMPACT_THRESHOLD_BYTES = 120_000` in `transcript.py`.
- Added `maybe_compact_cli_session(sdk_cwd, session_id, model, log_prefix)` which reads the file, skips if below threshold, runs `compact_transcript()`, and writes the result back.
- In `sdk/service.py`: call `maybe_compact_cli_session()` immediately after `cli_restored = True`.

*Fix 2:*
- Removed the `if Path(real_path).exists(): return True` block from `restore_cli_session`.
- Replaced with a comment explaining the cross-pod hazard.
- Updated `test_returns_true_when_local_file_already_exists` → `test_gcs_overwrites_stale_local_file` to assert that GCS is always consulted and the stale local file is replaced with fresh GCS content.

### Changes 🏗️

- `transcript.py`: Added `_PROACTIVE_COMPACT_THRESHOLD_BYTES` + `maybe_compact_cli_session()`. Removed same-pod early-return from `restore_cli_session`.
- `sdk/service.py`: Import and invoke `maybe_compact_cli_session()` after successful session restore.
- `transcript_test.py`: 4 new/updated tests — 3 for proactive compaction, 1 for cross-pod staleness regression.

#### For code changes:
- [x] I have made a test plan
  - [x] `test_skips_small_file` — below-threshold files are untouched
  - [x] `test_compacts_large_file_and_writes_back` — compaction triggers and file is updated
  - [x] `test_keeps_original_when_compaction_fails` — original preserved when `compact_transcript` returns `None`
  - [x] `test_gcs_overwrites_stale_local_file` — GCS always wins over stale local file; asserts `retrieve` called and disk content matches GCS

### DIFF
--- a/autogpt_platform/backend/backend/copilot/transcript.py
+++ b/autogpt_platform/backend/backend/copilot/transcript.py
@@ -689,6 +689,15 @@ def _cli_session_storage_path_parts(
     )
 
 
+# Byte-size threshold for proactive pre-query CLI session compaction.
+# Measured from production session 93ecf7c9: 233KB file → 204K tokens
+# (≈0.88 bytes/token).  120KB ≈ 105K tokens — roughly half the CLI's
+# ~200K auto-compaction threshold.  Compacting proactively here prevents
+# the CLI from silently auto-compacting mid-turn, which bypasses our
+# PreCompact hook and causes uncontrolled context loss.
+_PROACTIVE_COMPACT_THRESHOLD_BYTES = 120_000
+
+
 async def upload_cli_session(
     user_id: str,
     session_id: str,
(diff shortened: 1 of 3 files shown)


autogpt_platform/backend/backend/copilot/transcript.py
```diff
@@ -832,6 +841,96 @@ async def restore_cli_session(
...
+    try:
+        content = raw_bytes.decode("utf-8")
+    except UnicodeDecodeError as e:
+        logger.warning(
+            "%s CLI session is not valid UTF-8, skipping compaction: %s", log_prefix, e
+        )
+        return False
+
+    compacted = await compact_transcript(content, model=model, log_prefix=log_prefix)
+    if not compacted or compacted == content:
```
majdyz: 🤖 🟠 **Should Fix**: The combined condition `not compacted or compacted == content` collapses two meaningfully different outcomes into one warning message "Proactive compaction produced no change". `not compacted` means `compact_transcript` returned `None` (i.e., failure / too few messages), while `compacted == content` means it returned the exact same bytes (which is structurally very unlikely but different in meaning). Using a single log message for both makes it hard to diagnose from logs which scenario occurred. Suggest splitting into two conditions with distinct log messages: one for the failure/None case (`logger.warning("... compaction failed or returned None — keeping original")`) and one for the no-change case (`logger.warning("... compaction returned identical content — keeping original")`).

autogpt_platform/backend/backend/copilot/transcript_test.py
```diff
@@ -1285,3 +1285,161 @@ def test_returns_false_on_download_exception(self):
...
+                maybe_compact_cli_session(
+                    sdk_cwd=sdk_cwd,
+                    session_id=session_id,
+                    model="claude-sonnet-4",
+                )
+            )
+
+        assert result is False
+        # File should be unchanged.
+        assert session_file.read_text(encoding="utf-8") == large_content
```
majdyz, line 1456: 🤖 🟡 **Nice to Have**: Missing a test for the `compacted == content` branch (where `compact_transcript` returns the same string as the input). The current tests cover `compact_transcript` returning `None` and returning a different string, but not the case where it returns the identical input. This branch is logged as "Proactive compaction produced no change" and returns `False`, so it should be explicitly verified.

autogpt_platform/backend/backend/copilot/transcript.py
majdyz: 🤖 Fixed in [429ef3656](https://github.com/Significant-Gravitas/AutoGPT/commit/429ef365680a4b10a9b14cd542bd6c3f5c66d9c6): Split the combined guard into two separate `if` blocks — `not compacted` now logs "Proactive compaction failed or returned None — keeping original" and `compacted == content` logs "Proactive compaction returned identical content — keeping original".

autogpt_platform/backend/backend/copilot/transcript_test.py
majdyz: 🤖 Fixed in [429ef3656](https://github.com/Significant-Gravitas/AutoGPT/commit/429ef365680a4b10a9b14cd542bd6c3f5c66d9c6): Added `test_keeps_original_when_compaction_returns_identical_content` which patches `compact_transcript` to return `large_content` (same string as input) and asserts `result is False` and the file is unchanged.

PR #12801, opened 2026-04-15
fix(backend/copilot): disable fallback model to fix prod CLI rejection

majdyz:
### Why / What / How

**Why:** `fffbe0aad8` changed both `ChatConfig.model` and `ChatConfig.claude_agent_fallback_model` to `claude-sonnet-4-6`. The Claude Code CLI rejects this with `Error: Fallback model cannot be the same as the main model`, causing every standard-mode copilot turn to fail with exit code 1 — the session "completes" in ~30s but produces no response and drops the transcript.

**What:** Set `claude_agent_fallback_model` default to `""`. `_resolve_fallback_model()` already returns `None` on empty string, which means the `--fallback-model` flag is simply not passed to the CLI. On 529 overload errors the turn will surface normally instead of silently retrying with a fallback.

**How:** One-line config change + test update.

### Changes 🏗️

- `ChatConfig.claude_agent_fallback_model` default: `"claude-sonnet-4-6"` → `""`
- Update `test_fallback_model_default` to assert the empty default

#### For code changes:
- [x] I have made a test plan
  - [x] `poetry run pytest backend/copilot/sdk/p0_guardrails_test.py`

### DIFF
--- a/autogpt_platform/backend/backend/copilot/sdk/service.py
+++ b/autogpt_platform/backend/backend/copilot/sdk/service.py
@@ -2915,9 +2915,10 @@ async def stream_chat_completion_sdk(
                     sdk_options_kwargs_retry.pop("resume", None)
                     sdk_options_kwargs_retry["session_id"] = session_id
                 else:
-                    # T2+ retry without --resume: do not pass --session-id.
-                    # The T1 session file already exists at that path; re-using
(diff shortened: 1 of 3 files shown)


majdyz:
Wrong branch — recreating from hotfix branch

PR #12803, opened 2026-04-15
refactor(backend/copilot): consolidate session storage — eliminate redundant chat-transcripts/ path

majdyz:
## Why

Before `--resume` existed (pre-PR #12777), all session context lived in `chat-transcripts/` — a stripped JSONL we built with `TranscriptBuilder`. When `--resume` was added, a second file (`cli-sessions/`) was introduced for the CLI's native session. The old `chat-transcripts/` was never removed, leaving two separate GCS downloads and two uploads every turn.

## What

Eliminates `chat-transcripts/` from the SDK turn cycle. The SDK path now uses a single GCS location per session:

```
cli-sessions/{user_id}/{session_id}.jsonl       ← CLI session (--resume)
cli-sessions/{user_id}/{session_id}.meta.json   ← message_count watermark
```

Both files are uploaded and downloaded together in a single `asyncio.gather` — no window for divergence.

**Includes the cross-pod stale-file fix:** the same-pod early-return optimisation in `restore_cli_session` (`if Path(real_path).exists(): return True`) is removed. In a load-balanced multi-pod environment the local file may be from a stale earlier turn; GCS always has the authoritative latest version.

## How

- `restore_cli_session` now returns `CliSessionRestore | None` (content + message_count) instead of `bool`. It fetches session file + `.meta.json` concurrently and always overwrites any local file from GCS.
- `upload_cli_session` accepts `message_count` and uploads a companion `.meta.json` concurrently.
- `TranscriptBuilder` is seeded from the restored CLI session content (`strip_for_upload` applied in-memory), replacing the separate transcript download.
- `upload_transcript` / `download_transcript` are kept unchanged — still used by the baseline service.

## Checklist 📋

- [x] I have made a test plan
  - [x] `TestRestoreCliSession` — updated for new return type; verifies GCS always overrides local, meta message_count returned correctly
  - [x] `TestUploadCliSession` — verifies companion meta.json uploaded with correct message_count
  - [x] `TestStreamChatCompletionRetryIntegration` — all retry scenarios pass with new restore path
  - [x] 124 tests pass across transcript_test.py and retry_scenarios_test.py

### DIFF
--- a/autogpt_platform/backend/backend/copilot/sdk/service.py
+++ b/autogpt_platform/backend/backend/copilot/sdk/service.py
@@ -2915,9 +2915,10 @@ async def stream_chat_completion_sdk(
                     sdk_options_kwargs_retry.pop("resume", None)
                     sdk_options_kwargs_retry["session_id"] = session_id
                 else:
-                    # T2+ retry without --resume: do not pass --session-id.
-                    # The T1 session file already exists at that path; re-using
-                    # the same ID would fail with "Session ID already in use".
+                    # T2+ retry without --resume: initial invocation used
+                    # --resume, which restored the T1 session file to local
+                    # storage.  Re-using session_id without --resume would
+                    # fail with "Session ID already in use".
                     sdk_options_kwargs_retry.pop("resume", None)
                     sdk_options_kwargs_retry.pop("session_id", None)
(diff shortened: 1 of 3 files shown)


majdyz:
Closing — recreating with correct head branch (fix/copilot-single-session-store)

PR #12805, opened 2026-04-15
fix(frontend/builder): preserve agent name in AgentExecutor node title after reload

### LINKED ISSUE #11041: Show agent name + version on Agent Executor block

joemunene-by:
## Summary

Fixes #11041

When an `AgentExecutorBlock` is placed in the builder, it initially displays the agent's name (e.g., "Researcher v2"). After saving and reloading the page, the title reverts to the generic "Agent Executor."

## Root Cause

The backend correctly persists `agent_name` and `graph_version` in `hardcodedValues` (via `input_default` in `AgentExecutorBlock`). However, `NodeHeader.tsx` always resolves the display title from `data.title` (the generic block name), ignoring the persisted agent name.

## Fix

Modified the title resolution chain in `NodeHeader.tsx` to check `data.hardcodedValues.agent_name` between the user's custom name and the generic block title:

1. `data.metadata.customized_name` (user's manual rename) — highest priority
2. `agent_name` + ` v{graph_version}` from `hardcodedValues` — **new**
3. `data.title` (generic block name) — fallback

This is a frontend-only change. No backend modifications needed.

## Files Changed

- `autogpt_platform/frontend/src/app/(platform)/build/components/FlowEditor/nodes/CustomNode/components/NodeHeader.tsx` (+11, -1)

## Test Plan

- [x] Place an AgentExecutorBlock, select an agent — title shows agent name
- [x] Save graph, reload page — title still shows agent name (was "Agent Executor" before)
- [x] Double-click to rename — custom name takes priority over agent name
- [x] Clear custom name — falls back to agent name
- [x] Non-AgentExecutor blocks — unaffected, show generic title as before


autogpt_platform/frontend/src/app/(platform)/build/components/FlowEditor/nodes/CustomNode/components/NodeHeader.tsx
```diff
@@ -21,11 +21,28 @@ type Props = {
 export const NodeHeader = ({ data, nodeId }: Props) => {
   const updateNodeData = useNodeStore((state) => state.updateNodeData);
 
-  const title = (data.metadata?.customized_name as string) || data.title;
+  const agentName = data.hardcodedValues?.agent_name as string | undefined;
+  const graphVersion = data.hardcodedValues?.graph_version as number | undefined;
+  const agentDisplayName =
+    agentName && graphVersion != null
+      ? `${agentName} v${graphVersion}`
```
majdyz: This `useEffect` will reset `editedTitle` while the user is actively typing if the node data re-renders (e.g. from a concurrent store update or websocket event). Guard it:

```suggestion
  useEffect(() => {
    if (!isEditingTitle) {
      setEditedTitle(title);
    }
  }, [title, isEditingTitle]);
```

autogpt_platform/frontend/src/app/(platform)/build/components/FlowEditor/nodes/CustomNode/components/NodeHeader.tsx
```diff
@@ -21,11 +21,28 @@ type Props = {
...
+      : agentName || undefined;
+
+  const isAgentOrCustomTitle = !!(
+    data.metadata?.customized_name || agentDisplayName
+  );
+  const title =
+    (data.metadata?.customized_name as string) ||
+    agentDisplayName ||
+    data.title;
```
majdyz, line 27: This ternary block is duplicated verbatim for the tooltip (lines 101-106). Extract it into a `const displayTitle` computed once, then reference it in both places:

```tsx
const displayTitle = isAgentOrCustomTitle
  ? title
  : beautifyString(title).replace(/ Block$/, "").trim();
```

autogpt_platform/frontend/src/app/(platform)/build/components/FlowEditor/nodes/CustomNode/components/NodeHeader.tsx
```diff
@@ -7,7 +7,7 @@ import {
   TooltipTrigger,
 } from "@/components/atoms/Tooltip/BaseTooltip";
 import { beautifyString, cn } from "@/lib/utils";
-import { useState } from "react";
+import { useEffect, useState } from "react";
```
majdyz, line 10: **Inconsistency with other title consumers** — Other places in the codebase that resolve node display titles (`GraphContent.tsx:62`, `BuilderChatPanel/helpers.ts:155`, `useGraphMenuSearchBar.tsx:70`) all use `customized_name || data.title` with no awareness of `hardcodedValues.agent_name`. After this PR, the node header shows "Researcher v2" but the search menu and chat panel show "Agent Executor" for the same node.

Consider either:
1. Extracting a shared `getNodeDisplayTitle(data: CustomNodeData)` helper that all consumers use, or
2. Ensuring `customized_name` is updated whenever the agent selection changes (so the existing consumers automatically pick it up)

majdyz (changes requested):
## Review Summary

The fix correctly addresses the symptom described in #11041 — after reload, the agent executor node reverts to the generic "Agent Executor" title. The approach of reading `agent_name` from `hardcodedValues` as a fallback in the title resolution chain is reasonable. However, there are several issues to address before merging.

### CI is red

The **lint** check is failing because of Prettier formatting on NodeHeader.tsx. Please run `pnpm format` and push the fix.

### Design concern: dual source of truth for display name

`Block.tsx` (line 109-117) already sets `customized_name` in metadata to the agent's name when an AgentExecutorBlock is first placed. The real question is: **why isn't `customized_name` surviving the save/reload cycle?** If it is, then this PR's `hardcodedValues.agent_name` fallback is redundant — the existing `customized_name` path would already show the agent name. If it isn't, then there's a deeper bug in metadata persistence that this PR masks rather than fixes.

Could you clarify which scenario applies? If `customized_name` IS being lost on save/reload, that's a separate bug worth investigating. If `customized_name` is persisted but users can change the selected agent via the form (updating `hardcodedValues` but not `customized_name`), then the fallback approach makes sense — but you should also update `customized_name` when the agent selection changes, to keep the title consistent across other components that read `customized_name` (GraphContent.tsx, BuilderChatPanel/helpers.ts, useGraphMenuSearchBar.tsx).

### Code issues

1. **Duplicated display logic** — see inline comment
2. **`useEffect` to sync `editedTitle`** — see inline comment
3. **Inconsistency with other title consumers** — see inline comment
4. **The `.replace(/ Block$/, "")` improvement is good** — The original `.replace("Block", "")` would incorrectly strip "Block" from the middle of any title. The scoped regex is a nice improvement, but it's a behavior change that should be called out in the PR description.

### Missing test coverage

There are no tests for NodeHeader. Given the title resolution logic is now more complex (3-tier fallback + conditional formatting), this is a good opportunity to add a basic test file. At minimum, test:
- Default block title gets beautified and " Block" stripped
- `customized_name` takes priority and skips beautification
- `agent_name` + `graph_version` shows "Name vN" and skips beautification
- `agent_name` without `graph_version` shows just the name
- `customized_name` takes priority over `agent_name`

autogpt_platform/frontend/src/app/(platform)/build/components/FlowEditor/nodes/CustomNode/components/NodeHeader.tsx
majdyz: 🤖 Fixed in [5c72ee822](https://github.com/Significant-Gravitas/AutoGPT/commit/5c72ee82253264025a8c3b51d164e86785584114): Added `if (!isEditingTitle)` guard to the useEffect, with `isEditingTitle` in the dependency array.

autogpt_platform/frontend/src/app/(platform)/build/components/FlowEditor/nodes/CustomNode/components/NodeHeader.tsx
majdyz: 🤖 Fixed in [5c72ee822](https://github.com/Significant-Gravitas/AutoGPT/commit/5c72ee82253264025a8c3b51d164e86785584114): Extracted `const displayTitle = formatNodeDisplayTitle(data)` and used it in both the Text element and TooltipContent.

autogpt_platform/frontend/src/app/(platform)/build/components/FlowEditor/nodes/CustomNode/components/NodeHeader.tsx
majdyz: 🤖 Fixed in [5c72ee822](https://github.com/Significant-Gravitas/AutoGPT/commit/5c72ee82253264025a8c3b51d164e86785584114): Extracted shared `getNodeDisplayTitle()` and `formatNodeDisplayTitle()` into `CustomNode/helpers.ts`. Updated all three consumers:
- `GraphContent.tsx` — now calls `formatNodeDisplayTitle(node.data)`
- `BuilderChatPanel/helpers.ts` — `getNodeDisplayName()` now includes agent_name fallback
- `useGraphMenuSearchBar.tsx` — search now also matches against `agent_name`

autogpt_platform/frontend/src/app/(platform)/build/components/FlowEditor/nodes/CustomNode/components/NodeHeader.tsx
```diff
@@ -21,7 +21,17 @@ type Props = {
...
+      ? `${agentName} v${graphVersion}`
+      : agentName || undefined;
+
+  const title =
+    (data.metadata?.customized_name as string) ||
+    agentDisplayName ||
+    data.title;
 
   const [isEditingTitle, setIsEditingTitle] = useState(false);
   const [editedTitle, setEditedTitle] = useState(title);
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [5c72ee822](https://github.com/Significant-Gravitas/AutoGPT/commit/5c72ee82253264025a8c3b51d164e86785584114): The useEffect now syncs `editedTitle` with the `title` prop, and is guarded by `if (!isEditingTitle)` to avoid resetting during active edits.

autogpt_platform/frontend/src/app/(platform)/build/components/FlowEditor/nodes/CustomNode/components/NodeHeader.tsx
```diff
@@ -21,7 +21,17 @@ type Props = {
...
+  const graphVersion = data.hardcodedValues?.graph_version as number | undefined;
+  const agentDisplayName =
+    agentName && graphVersion != null
+      ? `${agentName} v${graphVersion}`
+      : agentName || undefined;
+
+  const title =
+    (data.metadata?.customized_name as string) ||
+    agentDisplayName ||
+    data.title;
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [5c72ee822](https://github.com/Significant-Gravitas/AutoGPT/commit/5c72ee82253264025a8c3b51d164e86785584114): The `.replace(/ Block$/, "")` regex (anchored to end) was already in place but is now only applied to generic block titles via `formatNodeDisplayTitle()`. Agent names and custom names bypass beautification entirely, so names like "Blockchain Agent" are never corrupted.

autogpt_platform/frontend/src/app/(platform)/build/components/FlowEditor/nodes/CustomNode/components/NodeHeader.tsx
```diff
@@ -21,11 +21,28 @@ type Props = {
...
 
   const [isEditingTitle, setIsEditingTitle] = useState(false);
   const [editedTitle, setEditedTitle] = useState(title);
 
+  useEffect(() => {
+    setEditedTitle(title);
+  }, [title]);
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [5c72ee822](https://github.com/Significant-Gravitas/AutoGPT/commit/5c72ee82253264025a8c3b51d164e86785584114): Added `if (!isEditingTitle)` guard so the effect only syncs when the user is not actively editing.

majdyz (dismissed):
LGTM after 3 review passes. All comments addressed, tests added, code is clean.

autogpt_platform/frontend/src/app/(platform)/build/components/FlowEditor/nodes/CustomNode/components/NodeHeader.tsx
```diff
@@ -72,12 +80,12 @@ export const NodeHeader = ({ data, nodeId }: Props) => {
...
                         className="line-clamp-1 hover:cursor-text"
                       >
-                        {beautifyString(title).replace("Block", "").trim()}
+                        {displayTitle}
                       </Text>
                     </div>
                   </TooltipTrigger>
                   <TooltipContent>
-                    <p>{beautifyString(title).replace("Block", "").trim()}</p>
+                    <p>{displayTitle}</p>
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [e47e04c1a](https://github.com/Significant-Gravitas/AutoGPT/commit/e47e04c1ac450fb39e7786f603ae828f3276e44d): Added `editedTitle !== title` guard in `handleTitleEdit` — clicking away without changes no longer saves the display title as `customized_name`, so future agent name/version updates continue to flow through.

majdyz (approved):

PR #12813, opened 2026-04-16
fix(frontend/copilot): fix streaming reconnect races, hydration ordering, and reasoning split

0ubbe:
## Summary

Improves Copilot/AutoPilot streaming reliability across frontend and backend. The diff now covers the original streaming investigation issues plus follow-up CI and review fixes from the latest merge with `dev`.

Addresses [SECRT-2240](https://linear.app/autogpt/issue/SECRT-2240), [SECRT-2241](https://linear.app/autogpt/issue/SECRT-2241), and [SECRT-2242](https://linear.app/autogpt/issue/SECRT-2242).

## Changes

- Fixes reasoning vs response rendering so action tools such as `run_block` and `run_agent` do not cause assistant response text to be hidden inside the collapsed reasoning section.
- Reworks Copilot session lifecycle handling: active-stream hydration, resume ordering, reconnect timeout recovery, wake resync, session deletion, title polling, stop handling, and session-switch stale callback guards.
- Adds a per-session Copilot stream store/registry and transport helpers to prevent duplicate resumes, duplicate sends, and cross-session contamination during reconnect or reload flows.
- Adds pending follow-up message support and backend pending-message safeguards, including sanitization of queued user content and requeue-on-persist-failure behavior.
- Improves backend stream and executor robustness: active stream registry checks, bounded cancellation drain with sync fail-close fallback, Redis helper coverage, and updated SDK response adapter expectations for post-tool status events.
- Adds and polishes usage-limit UI, including reset gate behavior, backdrop blending behind the usage-limit card, and usage panel/card coverage.
- Fixes a chat input Enter-submit race where Playwright and fast users could fill the textarea and press Enter before React had re-enabled the submit button, causing the visible message not to send.
- Refactors the Copilot page into smaller hooks/components and adds focused tests around stream recovery, hydration, pending queueing, rate-limit gates, and message rendering.

## Test plan

- [x] `poetry run format`
- [x] `poetry run pytest backend/copilot/sdk/response_adapter_test.py backend/copilot/executor/processor_test.py`
- [x] `pnpm prettier --write` on touched frontend files
- [x] `pnpm vitest run src/app/(platform)/copilot/components/ChatInput/__tests__/useChatInput.test.ts`
- [x] `pnpm types`
- [x] `pnpm lint` (passes with existing unrelated `next/no-img-element` warnings)
- [ ] Full GitHub CI after latest push

## Review notes

- The Sentry review thread about unbounded cancellation cleanup is addressed in `375ec9d5f`: cancellation now waits for normal async cleanup but exits after `_CANCEL_GRACE_SECONDS` and falls through to the sync fail-close path.
- The previous backend CI failures were stale test expectations around the new `StreamStatus("Analyzing result…")` event after tool output; tests now assert that event explicitly.
- The previous full-stack E2E failure was the Copilot input Enter race; the input now submits from the live form value instead of depending on a possibly stale disabled button state.


autogpt_platform/frontend/src/app/(platform)/copilot/useCopilotStream.ts
```diff
@@ -137,6 +149,25 @@ export function useCopilotStream({
...
+        setReconnectTimedOut(true);
+        reconnectStartedAtRef.current = null;
+        toast({
+          title: "Connection timed out",
+          description:
+            "AutoPilot may still be working. Refresh to check for updates.",
+          variant: "destructive",
+        });
+      }, RECONNECT_MAX_DURATION_MS);
+    }
```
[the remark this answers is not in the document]
0ubbe: 🤖 Fixed in [be6eabce7](https://github.com/Significant-Gravitas/AutoGPT/commit/be6eabce75b0417f72e103aa99af1bc4adb9e239): Added `clearTimeout(reconnectTimeoutTimerRef.current)` before the early return in the max-attempts branch. Good catch — this would have caused a double toast.

autogpt_platform/frontend/src/app/(platform)/copilot/useCopilotStream.ts
```diff
@@ -137,6 +149,25 @@ export function useCopilotStream({
...
+        setReconnectExhausted(true);
+        setReconnectTimedOut(true);
+        reconnectStartedAtRef.current = null;
+        toast({
+          title: "Connection timed out",
+          description:
+            "AutoPilot may still be working. Refresh to check for updates.",
+          variant: "destructive",
+        });
+      }, RECONNECT_MAX_DURATION_MS);
```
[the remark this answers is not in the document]
0ubbe: 🤖 Fixed in [be6eabce7](https://github.com/Significant-Gravitas/AutoGPT/commit/be6eabce75b0417f72e103aa99af1bc4adb9e239): Added two guards inside the timeout callback: (1) `sessionEpochRef` check to bail on session switches, and (2) `isReconnectScheduledRef` check to bail if reconnect already succeeded. This prevents the false toast when the stream has resumed.

autogpt_platform/frontend/src/app/(platform)/copilot/useCopilotStream.ts
```diff
@@ -572,6 +651,7 @@ export function useCopilotStream({
     status,
     error: isReconnecting || isUserStoppingRef.current ? undefined : error,
     isReconnecting,
+    reconnectTimedOut,
     isSyncing,
     isUserStoppingRef,
     rateLimitMessage,
```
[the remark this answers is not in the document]
0ubbe: 🤖 Fixed in [be6eabce7](https://github.com/Significant-Gravitas/AutoGPT/commit/be6eabce75b0417f72e103aa99af1bc4adb9e239): Removed `reconnectTimedOut` state, its setter calls, and its return value — it was indeed dead code never consumed by any component.

autogpt_platform/frontend/src/app/(platform)/copilot/useCopilotStream.ts
```diff
@@ -137,6 +149,27 @@ export function useCopilotStream({
...
+    if (reconnectStartedAtRef.current === null) {
+      reconnectStartedAtRef.current = Date.now();
+      // Schedule a forced timeout — if reconnecting takes longer than
+      // RECONNECT_MAX_DURATION_MS, force the UI back to idle.
+      clearTimeout(reconnectTimeoutTimerRef.current);
+      const capturedEpoch = sessionEpochRef.current;
+      reconnectTimeoutTimerRef.current = setTimeout(() => {
+        if (sessionEpochRef.current !== capturedEpoch) return;
+        if (!isReconnectScheduledRef.current) return;
+        setReconnectExhausted(true);
```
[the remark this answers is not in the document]
0ubbe: 🤖 Fixed in [14e9a1844](https://github.com/Significant-Gravitas/AutoGPT/commit/14e9a18445cc25e3333cdb218fad593def9fff9b): Valid bug — removed the `isReconnectScheduledRef` guard from the 30s timeout callback. The ref gets cleared by the reconnect timer before `resumeStream()` fires, so if the backend is slow the timeout would bail out instead of forcing idle. The `sessionEpochRef` check is sufficient to prevent stale fires. Also added cleanup to clear the timeout when the stream successfully resumes (status → streaming/submitted), and allowed resume when hydration returns empty messages.

autogpt_platform/frontend/src/app/(platform)/copilot/useCopilotStream.ts
```diff
@@ -419,20 +459,35 @@ export function useCopilotStream({
     };
   }, [refetchSession, setMessages]);
 
-  // Hydrate messages from REST API when not actively streaming
+  // Hydrate messages from REST API when not actively streaming.
+  // Sets hydrateCompletedRef so the resume effect knows it's safe to proceed.
   useEffect(() => {
```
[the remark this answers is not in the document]
0ubbe: 🤖 **False positive** — the two resume paths are mutually exclusive:

1. **Reconnect path** (`handleReconnect` → timer → `resumeStreamRef`): fires mid-conversation on disconnect. While active, `isReconnectScheduled` is true.
2. **Resume path** (hydration effect → `doResume` → `resumeStreamRef`): fires on page load.

The hydration effect at L467 has `if (isReconnectScheduled) return;` — it bails entirely while a reconnect is in progress. After the reconnect timer fires and clears `isReconnectScheduled`, the stream is already `streaming`/`submitted`, and the resume effect's guard at L599 (`if (status === 'streaming' || status === 'submitted') return;`) prevents a double call.

These paths cannot overlap: reconnect runs during active conversations, resume runs on page load.

autogpt_platform/frontend/src/app/(platform)/copilot/useCopilotStream.ts
```diff
@@ -150,8 +182,12 @@ export function useCopilotStream({
...
     const delay = RECONNECT_BASE_DELAY_MS * 2 ** (nextAttempt - 1);
+    const capturedEpoch = sessionEpochRef.current;
 
     reconnectTimerRef.current = setTimeout(() => {
+      // Bail if the session switched while the timer was pending.
+      if (sessionEpochRef.current !== capturedEpoch) return;
+
       isReconnectScheduledRef.current = false;
       setIsReconnectScheduled(false);
       // Strip the stale in-progress assistant message before resuming —
```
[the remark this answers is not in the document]
0ubbe: 🤖 Fixed in [1ae6aa86f](https://github.com/Significant-Gravitas/AutoGPT/commit/1ae6aa86f1519dd759a49b20e190f3a1cde59ac7): Valid bug. Setting `hasResumedRef.current.set(sid, true)` in the reconnect timer before `resumeStream()` prevents a deferred page-load resume (queued in `pendingResumeRef` while hydration was in-flight) from double-firing `resumeStream()` once hydration completes. Same guard pattern as `doResume`. Flag is session-scoped and cleared on session switch, so subsequent reconnect cycles are unaffected.

majdyz (approved):

0ubbe:
🤖 Addressed a subset of the autogpt-reviewer top-level review in [2f479e51a](https://github.com/Significant-Gravitas/AutoGPT/commit/2f479e51a4f4cbf7a4046d0a14390e847e8febb8):

**Should-Fix addressed:**
- **`RateLimitGate` polls unconditionally** — added `enabled: !!rateLimitMessage` so the 30s usage query only runs once the user actually hits the limit.
- **Magic 500ms delay in `onFinish`** — extracted to `FINISH_REFETCH_SETTLE_MS` alongside the other timing constants at the top of `useCopilotStream.ts`.
- **`useSessionDeletion` stale closure in `onSuccess`** — the active-session clear now keys off the mutation's own `variables.sessionId` rather than the closed-over `sessionToDelete`, so rapid open/cancel/open-different sequences can't clear the wrong session after the round-trip.
- **`RateLimitGate` has no tests** — added a component-level test suite (`__tests__/RateLimitGate.test.tsx`) covering the enabled-gating, dialog-open path, and both toast-fallback paths (query error + zero reset cost).

**Deliberately not addressed in this PR:**
- **`useSendMessage` / `useSessionDeletion` component-level integration tests, onError isUserStopping guard test** — these would require either rewiring the existing `CopilotPage.test.tsx` (which mocks `useCopilotPage` wholesale) or adding hook-level tests. Per repo convention we prefer component/page-level assertions over hook-level ones; happy to add these in a follow-up if reviewers want them gated before merge.
- **`stripAndResume` stale-closure refactor** — the existing test suite confirms the current behavior works in production (the closure runs synchronously inside the setMessages updater in event-handler contexts); the test-only workaround is because our SDK mock doesn't execute updaters. Leaving for a follow-up to avoid churning the recently-stabilized snapshot logic.
- **Unbounded sessions-store growth, `deduplicateMessages` JSON.stringify hot path, batched updateCoord calls, `SessionCoord` rename, `MobileDrawer` density, `REASONING_TOOL_TYPES` drift assertion, manual test checklist, resume CodeRabbit** — Nice-to-Have / Nit-tier items; noted for follow-ups.

autogpt_platform/backend/backend/copilot/sdk/service.py
```diff
@@ -477,6 +477,25 @@ async def _reduce_context(
...
+    if len(quoted) == 1:
+        return quoted[0]
+    if len(quoted) == 2:
+        return f"{quoted[0]} and {quoted[1]}"
+    extras = len(quoted) - 2
+    return f"{quoted[0]}, {quoted[1]}, and {extras} other"
+
+
 def _append_error_marker(
     session: ChatSession | None,
```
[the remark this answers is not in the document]
0ubbe: 🤖 Fixed in [bc3ac5a14](https://github.com/Significant-Gravitas/AutoGPT/commit/bc3ac5a1431d8a18086bfcb963ba5d1068a22e07): pluralize the suffix — `N other` when `extras == 1`, `N others` when `extras > 1`.

autogpt_platform/frontend/src/app/(platform)/copilot/components/RateLimitResetDialog/RateLimitGate.tsx
```diff
@@ -0,0 +1,74 @@
...
+    <RateLimitResetDialog
+      isOpen={canShowDialog}
+      onClose={onDismiss}
+      resetCost={resetCost ?? 0}
+      resetMessage={rateLimitMessage ?? ""}
+      isWeeklyExhausted={
+        hasUsage &&
+        usage.weekly.limit > 0 &&
+        usage.weekly.used >= usage.weekly.limit
+      }
```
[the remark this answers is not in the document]
0ubbe: 🤖 Fixed in [bc3ac5a14](https://github.com/Significant-Gravitas/AutoGPT/commit/bc3ac5a1431d8a18086bfcb963ba5d1068a22e07): added explicit `!!usage.weekly` null check in the `isWeeklyExhausted` expression before reading `.limit`/`.used`.

autogpt_platform/frontend/src/app/(platform)/copilot/useSendMessage.ts
```diff
@@ -0,0 +1,198 @@
...
+    queuedFirstSend = null;
+    const prebuiltParts = pendingFileParts;
+    pendingFileParts = [];
+    void dispatchRef.current(
+      sessionId,
+      queued.text,
+      queued.files,
+      prebuiltParts,
+    );
+  }, [sessionId]);
```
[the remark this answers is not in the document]
0ubbe: Already resolved structurally in [02f5ab0](https://github.com/Significant-Gravitas/AutoGPT/commit/02f5ab063eb372d321ec571e59f07c22761be760): the module-level `queuedFirstSend` variable was removed and replaced by the per-session Zustand slot `useCopilotStreamStore.takePendingFirstSend()`. Combined with CopilotPage keying the chat subtree by `sessionId`, the consumer is now the newly-mounted session host, not a stale existing-chat host.

autogpt_platform/frontend/src/app/(platform)/copilot/useCopilotStream.ts
```diff
@@ -77,146 +68,43 @@ export function useCopilotStream({
...
-      // the backend replays from "0-0", so keeping it would duplicate parts.
-      setMessages((prev) => {
-        if (prev.length > 0 && prev[prev.length - 1].role === "assistant") {
-          return prev.slice(0, -1);
-        }
-        return prev;
-      });
-      lastReconnectResumeAtRef.current = Date.now();
-      resumeStreamRef.current();
-    }, delay);
```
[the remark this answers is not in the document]
0ubbe: Not applicable — project style in `autogpt_platform/frontend/AGENTS.md` and `CONTRIBUTING.md` disallows `useMemo`/`useCallback` unless there's a measured perf issue ("Do not use `useCallback` or `useMemo` unless asked to optimise a given function"). `deduplicateMessages` is pure and `rawMessages` keeps a stable reference between renders that don't mutate the list, so there is no measurable regression. The top-level automated review already tracks the preferred optimisation (ref-based short-circuit inside the helper itself) as a Nice to Have.

autogpt_platform/backend/backend/copilot/executor/processor.py
```diff
@@ -370,40 +371,67 @@ def _execute(
...
             except concurrent.futures.TimeoutError:
-                log.warning(
-                    "Future did not complete within grace window; "
-                    "falling through to sync fail-close"
-                )
+                if cancel.is_set() and not cancel_requested:
+                    request_cancel()
+                elif cancel_requested and cancel_started_at is not None:
+                    log_cancel_wait()
+                cluster_lock.refresh()
```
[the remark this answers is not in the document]
0ubbe: 🤖 Fixed in [375ec9d5f](https://github.com/Significant-Gravitas/AutoGPT/commit/375ec9d5fcd2e7c7292c9bf5f339211146ef40c8): bounded the cancellation drain again. `_execute` now waits for normal async cleanup, but after `_CANCEL_GRACE_SECONDS` it stops refreshing the session lock, cancels the future again, and falls through to the sync fail-close safety net. Added a regression test for a wedged cancellation cleanup that would otherwise refresh forever.

autogpt_platform/backend/backend/api/features/chat/routes.py
```diff
@@ -1130,12 +1155,37 @@ async def event_generator() -> AsyncGenerator[str, None]:
...
-            "Cache-Control": "no-cache",
-            "Connection": "keep-alive",
-            "X-Accel-Buffering": "no",  # Disable nginx buffering
-            "x-vercel-ai-ui-message-stream": "v1",  # AI SDK protocol header
-        },
+        headers=_ui_message_stream_headers(),
+    )
+
+
+@router.post(
```
majdyz, line 1188: why do we need to split this? will this risk of losing atomicity ?

autogpt_platform/backend/backend/blocks/autopilot.py
```diff
@@ -198,6 +200,13 @@ class Input(BlockSchemaInput):
         # timeouts internally; wrapping with asyncio.timeout corrupts the
         # SDK's internal stream (see service.py CRITICAL comment).
 
+        @field_validator("tools", mode="before")
+        @classmethod
+        def strip_disabled_legacy_tools(cls, tools: Any) -> Any:
+            if not isinstance(tools, list):
+                return tools
+            return [tool for tool in tools if tool not in DISABLED_LEGACY_TOOL_NAMES]
```
majdyz, line 208: will this cause issue if we strip the tool from context? loading disabled tool caused issue as well ?

autogpt_platform/backend/backend/copilot/executor/processor.py
```diff
@@ -370,40 +373,81 @@ def _execute(
...
+            return True
 
-        if not future.cancelled():
-            # Bounded timeout so a wedged event loop can't trap us here —
-            # on timeout we escape to execute()'s finally and the sync
-            # safety net fires.
+        # Wait for completion, checking cancel periodically. A cancellation
+        # request waits for normal async cleanup, but remains bounded so the
+        # worker does not refresh the per-session lock forever on a wedged turn.
+        while True:
```
majdyz, line 446: won't this risk infinite loop ?

autogpt_platform/backend/backend/copilot/executor/processor.py
```diff
@@ -370,40 +373,81 @@ def _execute(
...
+
         future = asyncio.run_coroutine_threadsafe(
-            self._execute_async(entry, cancel, cluster_lock, log),
+            run_async_turn(),
             self.execution_loop,
         )
 
-        # Wait for completion, checking cancel periodically
-        while not future.done():
+        cancel_requested = False
```
majdyz, line 391: What is the intention of this change? what issue we are trying to fix and how ?

autogpt_platform/backend/backend/copilot/pending_messages.py
```diff
@@ -161,6 +161,47 @@ async def push_pending_message(
     return new_length
 
 
+async def push_pending_message_if_session_running(
```
majdyz, line 164: what if message is not running now the endpoint is split? are we going to reject it and lose message?

autogpt_platform/backend/backend/copilot/permissions.py
```diff
@@ -128,6 +128,14 @@
 # Frozen set of all valid tool names — derived from the Literal.
 ALL_TOOL_NAMES: frozenset[str] = frozenset(get_args(ToolName))
 
+DISABLED_LEGACY_TOOL_NAMES: frozenset[str] = frozenset({"ask_question"})
```
majdyz, line 131: why not just genereate the list from all_tool_names and the tool being enabled? this list is hard to maintain

autogpt_platform/backend/backend/data/redis_helpers.py
```diff
@@ -47,6 +47,27 @@
...
+#   ARGV[5]  list TTL seconds
+_GATED_CAPPED_RPUSH_LUA = """
+local current = redis.call('HGET', KEYS[1], ARGV[1])
+if current ~= ARGV[2] then
+    return -1
+end
+redis.call('RPUSH', KEYS[2], ARGV[3])
+redis.call('LTRIM', KEYS[2], -tonumber(ARGV[4]), -1)
+redis.call('EXPIRE', KEYS[2], tonumber(ARGV[5]))
+return redis.call('LLEN', KEYS[2])
```
majdyz, line 68: do we really need to do this? this seems pretty arcane use case, what's the need of it ? 
especially the ltrim

autogpt_platform/backend/backend/api/features/chat/routes.py
0ubbe: The split is for Vercel AI SDK compliance, not atomicity.

**Why:** before, `POST /stream` returned `202 application/json` when queueing mid-turn. The AI SDK transport (`DefaultChatTransport` / `useChat`) only parses SSE — a JSON body wedges the UI-message parser. In three real paths the transport lands there even though the FE tries to route queue-sends through a bespoke helper: (1) race where the turn ends just as the user clicks send, (2) `useChat({ resume: true })` auto-firing on mount, (3) a second tab posting while tab A streams. All three would drop into the 202 branch and wedge.

**What:** `/stream` now always returns SSE (empty SSE on the in-flight branch), and `/messages/pending` handles queueing separately with proper `409`/`429` semantics.

**Atomicity:** preserved via `push_pending_message_if_session_running` — the Lua gate in `redis_helpers.capped_rpush_if_hash_field` collapses the HGET(status=="running") + RPUSH into a single atomic op, so there is no race window between "is turn running?" and "append to buffer". Covered on the pending_messages.py thread.

Confirmed in Discord to keep the split as-is.

autogpt_platform/backend/backend/blocks/autopilot.py
0ubbe: Strip is safer than pass-through here.

`DISABLED_LEGACY_TOOL_NAMES` is a tiny set of names that are *intentionally absent* from `ToolName` / `PLATFORM_TOOL_NAMES` (currently only `ask_question`, which we removed pending UX rework in #12887). They used to be valid, so saved AutoPilotBlock inputs from before the removal still reference them.

If we did NOT strip:
- Pydantic accepts the input because `tools` is typed as `list[ToolName]`, but we previously kept the Literal tolerant during the rollout.
- Downstream, `apply_tool_permissions` maps to `mcp__copilot__ask_question` which is no longer registered, causing `allowed_tools` / `disallowed_tools` semantics to drift (the name silently becomes "allow nothing matching this" because no mapping resolves).
- Worse, if we ever tighten the pattern, the block would fail to load entirely and break saved graphs.

Strip removes the name before validation so the block loads cleanly with the modern tool set. No user-visible regression — `ask_question` was already disabled server-side.

autogpt_platform/backend/backend/copilot/executor/processor.py
0ubbe: Not infinite — there are three exit paths, all bounded:

1. `future.result(timeout=1.0)` succeeds → `return` (normal completion).
2. `CancelledError` → `return` (when `cancel_requested` is true or `cancel.is_set()`).
3. `cancel_drain_timed_out()` → returns `True` after `_CANCEL_GRACE_SECONDS` (5s) since the cancel was signalled, triggers `future.cancel()` and `return`.

Worst case: cancel signal arrives, turn refuses to drain → we wait up to 5s, then force-cancel and exit. Outer `execute()` calls `sync_fail_close_session` in `finally`, so the session is always marked terminal regardless of which path we take.

If nothing ever cancels, the loop waits on the future with a 1s poll so `cluster_lock.refresh()` keeps the lock alive for long-running turns. That is the only "unbounded" shape, and it is bounded by the turn itself finishing (or the pool worker being shut down).

autogpt_platform/backend/backend/copilot/executor/processor.py
0ubbe: Intent: bound cancellation cleanup so a wedged async turn cannot hold the cluster lock forever.

**Bug before:** the previous shape called `task.cancel()` on the async turn and then waited on its cleanup finally-blocks. If anything in the finally path (Redis write, stream-registry CAS, SDK client teardown) was stuck, the pool worker kept refreshing the per-session cluster lock, blocking every other worker from ever picking up that session. We observed this in production — sessions stuck in `running` indefinitely.

**Fix:** bounded drain.
- `_CANCEL_GRACE_SECONDS` (5s) is the max the worker waits for normal async cleanup after a cancel is signalled.
- If the drain does not complete in time, `cancel_drain_timed_out()` calls `future.cancel()` (forceful) and we fall through to the outer `execute()`s `finally`, which invokes `sync_fail_close_session` from the pool worker thread. That is a CAS on `status == "running" → "failed"` with its own 10s Redis timeout (`_FAIL_CLOSE_REDIS_TIMEOUT`), so even if Redis itself is slow we escape.

**Why `task_ready`:** to cancel cooperatively we need the `asyncio.Task`, which only exists on the event-loop thread. `run_async_turn` captures it with `asyncio.current_task()` and hands it back to the sync caller via a `concurrent.futures.Future`, so `request_cancel` can schedule `task.cancel()` via `call_soon_threadsafe` (plus a `future.cancel()` fallback if the task has not started yet).

So: the change converts "cancel and hope" into "cancel, wait 5s, force-cancel, always fail-close".

autogpt_platform/backend/backend/copilot/pending_messages.py
0ubbe: Not lost — promoted to a fresh turn.

When the Lua gate returns `None` (status flipped to `completed` between FE check and server push), `queue_pending_for_http` raises `HTTPException(409, "Session has no active turn. Start a new turn with POST /stream.")`.

FE side (`useCopilotPage.ts`):

```ts
try {
  await queueFollowUpMessage(sessionId, trimmed);
  appendChip(trimmed);
} catch (err) {
  if (err instanceof Error && err.name === "QueueFollowUpNotActiveError") {
    await sendNewMessage(message, files);  // promote to a fresh /stream POST
    return;
  }
  // ... toast
}
```

The helper throws `QueueFollowUpNotActiveError` on 409, the caller catches it and calls `sendNewMessage`, which starts a brand-new turn with that message as the user prompt. Net effect:

- Turn still running → message lands in the pending buffer (drained next round).
- Turn just finished → message becomes the prompt of the next turn.

Either way it reaches the model; the only difference is whether it arrives as a mid-turn injection or as a fresh turn. No silent drop, no orphan buffer entry.

autogpt_platform/backend/backend/copilot/permissions.py
0ubbe: They cannot be derived — the two sets are disjoint by design.

- `ALL_TOOL_NAMES` = names currently valid (`get_args(ToolName)`).
- `DISABLED_LEGACY_TOOL_NAMES` = names that *used to* be in `ToolName` but have been removed. Saved AutoPilotBlock inputs from before the removal still reference them, and we need to silently strip those on load.

So the set is intentionally the complement of "enabled" — it only holds names we have deliberately removed from the Literal. Currently one entry (`ask_question`, removed pending UX rework in #12887). The deprecation contract is: when a tool is removed from `ToolName`, add its name here for one release cycle, then drop once no saved graphs reference it anymore.

If the list grows large enough that manual maintenance becomes real friction, the right shape is probably a `@deprecated_tools` annotation on the Literal with a helper that derives both sets — happy to revisit then. With one entry today it did not seem worth the indirection.

autogpt_platform/backend/backend/data/redis_helpers.py
0ubbe: The gate closes a specific race; the LTRIM is the bounded-buffer invariant shared with `capped_rpush`.

**Why the gate:** without it, there is a window where a pending message is pushed into a buffer that nothing will drain.

1. FE checks `is_turn_in_flight` → `True`.
2. BE Python-side `HGET status` → `"running"`.
3. Turn completes on another path: drains the pending buffer, flips status to `"completed"`.
4. BE does `RPUSH` → orphan message sitting in a buffer with no running consumer.

Collapsing `HGET + RPUSH` into one Lua eval removes the window. Either the push lands while the turn is still running (its end-of-turn drain picks it up) or the push is rejected and the endpoint returns 409 → FE promotes the message to a fresh `/stream` (see the pending_messages.py thread). No orphan path.

**Why the LTRIM:** same contract as `capped_rpush` — cap the list at `MAX_PENDING_MESSAGES` (10). Without it, a misbehaving client could push beyond the cap. Keeping LTRIM + EXPIRE inline in the gated variant means the two push paths (gated / non-gated) have identical bounded semantics, which matters because they share the same drain code.

Discussed on Discord — happy to revisit if the Lua feels too niche; the alternative I could see is `flip status first, then drain` (inverts turn-completion ordering and has its own race), or a lazy cleanup sweep (messier). Atomic gate was the cheapest bounded fix.

ntindle (approved):
Tested and seems to work. 

JK. After an absurdly long reply that I queued a message during (wiht no tool calls so it had not point to send it in) I got the reply but now the text box is locked to sending more

image

ntindle (dismissed):
JK. After an absurdly long reply that I queued a message during (wiht no tool calls so it had not point to send it in) I got the reply but now the text box is locked to sending more

image

This could be preexisting but not sure. sessionId=22707512-7e58-4e85-ad0d-13cb72fa9048


<details>
Alright, here comes an absolutely massive wall of text to test stream resumption. Brace yourself.

The quick brown fox jumps over the lazy dog. Now let me say that again, and again, and again, because you asked for a ton of text and I am more than happy to oblige. The quick brown fox jumps over the lazy dog. The quick brown fox jumps over the lazy dog. The quick brown fox jumps over the lazy dog.

Let me tell you a story. Once upon a time, in a land far, far away, there was a kingdom called Veridia. This kingdom was known throughout the world for its magnificent libraries that stretched higher than the clouds and deeper than the oceans. The librarians of Veridia were not ordinary librarians; they were sorcerers of knowledge, custodians of history, and guardians of forgotten languages. Every morning, as the sun crested over the emerald mountains, the Grand Librarian would ring the Bell of Inquiry, and scholars from every corner of the realm would gather in the Grand Hall of Scripts.

The Grand Hall itself was a marvel of architecture. Its ceilings were painted with constellations that moved in real-time, tracking the stars above. The walls were lined with shelves that seemed to extend into infinity, each shelf glowing with a soft, amber light. The floor was made of marble so polished that one could see their reflection, and in that reflection, some claimed they could see glimpses of their past lives. At the center of the hall stood the Tree of Knowledge, an ancient oak made entirely of crystallized parchment, its branches heavy with books that grew like fruit.

One day, a young apprentice named Elara discovered a book in the deepest vaults of the library. This vault, known as the Abyssal Archive, was forbidden to all but the most senior librarians. But Elara, driven by an insatiable curiosity, had found a hidden passage behind a tapestry depicting the First Scribe. The book she found had no title, no author, and no date. Its cover was bound in material that felt like leather but shifted colors under her touch. When she opened it, the pages were blank. Or so she thought.

As she stared at the empty pages, words began to appear, written in ink that glowed like liquid starlight. The book was writing itself, narrating events that had not yet happened. It spoke of a Great Forgetting, a curse that would sweep across Veridia, erasing every written word, every recorded thought, every piece of knowledge stored in the great libraries. The books would turn blank, the scrolls would crumble to dust, and the people would lose all memory of their history, their language, and even their own names.

Elara, terrified but determined, took the book to the Grand Librarian, an ancient woman named Seraphine who had served the library for three hundred years. Seraphine read the prophetic pages with trembling hands. She explained that the Great Forgetting was not merely a curse but a natural cycle, a cosmic reset that occurred every ten thousand years. The last Forgetting had nearly destroyed civilization, plunging the world into a dark age that lasted millennia. The only way to stop it was to find the Codex Eternum, the first book ever written, which contained the Original Language—the tongue spoken by the universe itself, which could rewrite reality.

And so began an epic quest that spanned decades and crossed impossible distances. Elara assembled a fellowship of unlikely heroes: Kael the Cartographer, who could map not just lands but dreams and memories; Mira the Alchemist, who could distill thoughts into potions and memories into crystals; Thorn the Guardian, a statue brought to life by ancient magic, whose body was indestructible but whose heart was gentle; and Whisper the Fox, a creature of pure chaos who spoke in riddles and could walk through walls.

Their journey took them to the Sunken Library of Aquantis, where books were written on water and preserved in bubbles of air trapped beneath the ocean. They navigated the Labyrinth of Lost Hours, a maze where time flowed backward and every wrong turn aged you by a year. They climbed the Infinite Staircase, which had no beginning and no end, existing in the space between moments. They braved the焚书之谷 (Valley of Burning Pages), where books screamed as they burned, each scream a story dying forever.

In the Sunken Library, Kael mapped the currents of knowledge, tracing invisible rivers of thought that connected every piece of written word in existence. He discovered that all libraries, everywhere, were connected by these currents, forming a vast neural network of human consciousness. Mira distilled the essence of ancient poems into elixirs that granted temporary omniscience, though the hangover was severe—drinkers would forget their own names for days. Thorn carried the heaviest burdens, quite literally, bearing crates of rare manuscripts on his stone back, never complaining, always humming old songs that had no words.

Whisper the Fox proved to be the most valuable and most frustrating member of the fellowship. He led them through secret passages, stolen moments, and forgotten dreams. He stole the keys to the Library of Shadows, where forbidden knowledge was kept, by simply not existing in the same dimension as the locks. But he also led them in circles when bored, told truths that sounded like lies, and lies that sounded like truths, and occasionally disappeared for weeks, only to return with a single cryptic clue written on a leaf.

After twenty years of searching, they finally reached the End of the World, a place where the ocean fell off the edge of reality into a void of pure white. Suspended in that void, held in place by the gravity of meaning itself, floated the Codex Eternum. But guarding it was the Curator, a being older than time, shaped like a human but made of swirling galaxies and dying stars. The Curator spoke in the Original Language, and its words had the power to create and destroy.

Elara approached the Curator alone. She had no weapon, no spell, no trick. She had only the book she had found in the Abyssal Archive, which had continued to write itself throughout their journey, now chronicling their adventure in real-time. She offered it to the Curator and spoke three words she had learned from Whisper the Fox, three words in the Original Language that meant "I remember."

The Curator paused. It examined the book, watching the words appear, seeing its own existence reflected in the narrative. No one had ever offered it a story before. Everyone who came seeking the Codex Eternum wanted to use it for power, for immortality, for conquest. But Elara offered a gift simply to preserve what was. The Curator, moved by this pure intention, granted her access to the Codex.

Elara opened the Codex Eternum. Inside was a single sentence, written in the handwriting of every author who had ever lived: "Stories endure because we choose to tell them." She copied this sentence into the prophetic book, and the book absorbed the Original Language, becoming a new Codex, a living testament to humanity's collective memory.

When she returned to Veridia, the Great Forgetting was already beginning. Books were turning blank, scrolls crumbling, scholars forgetting their own names. But Elara read from the new Codex, and her words spread like sunlight. Every word she spoke rewrote itself into existence, anchoring reality against the void. She traveled the world, reading in every tongue, and everywhere she went, memory was restored, strengthened, made immortal.

She became the Eternal Librarian, neither fully human nor fully divine, walking between the worlds of the living and the written. The libraries of Veridia grew beyond their walls, becoming part of the landscape itself—mountains carved with epics, rivers singing ballads, trees whispering poetry in the wind. And in the Grand Hall of Scripts, the Tree of Knowledge bore a new fruit: a book with Elara's name on the cover, still being written, page by page, as her story continued.

The moral of this story, if there must be one, is that memory is not a passive thing. It is an act of will, aChoice made again and again, to remember, to record, to resist the forgetting that is entropy's natural end. We tell stories not because they are true, though they are, but because telling them makes them true. Every word written is a victory against the void. Every book is a fortress. Every reader is a soldier in the army of remembrance.

And now, because I promised you a shit ton of text and I always keep my promises, let me continue with some classic placeholder content expanded to absurd lengths, because sometimes the medium is the message and the message is "I am sending you a lot of text right now to test stream resumption and I really hope this works because it would be super annoying if it didn't."

Lorem ipsum dolor sit amet, consectetur adipiscing elit. Sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.

Sed ut perspiciatis unde omnis iste natus error sit voluptatem accusantium doloremque laudantium, totam rem aperiam, eaque ipsa quae ab illo inventore veritatis et quasi architecto beatae vitae dicta sunt explicabo. Nemo enim ipsam voluptatem quia voluptas sit aspernatur aut odit aut fugit, sed quia consequuntur magni dolores eos qui ratione voluptatem sequi nesciunt. Neque porro quisquam est, qui dolorem ipsum quia dolor sit amet, consectetur, adipisci velit, sed quia non numquam eius modi tempora incidunt ut labore et dolore magnam aliquam quaerat voluptatem. Ut enim ad minima veniam, quis nostrum exercitationem ullam corporis suscipit laboriosam, nisi ut aliquid ex ea commodi consequatur? Quis autem vel eum iure reprehenderit qui in ea voluptate velit esse quam nihil molestiae consequatur, vel illum qui dolorem eum fugiat quo voluptas nulla pariatur?

At vero eos et accusamus et iusto odio dignissimos ducimus qui blanditiis praesentium voluptatum deleniti atque corrupti quos dolores et quas molestias excepturi sint occaecati cupiditate non provident, similique sunt in culpa qui officia deserunt mollitia animi, id est laborum et dolorum fuga. Et harum quidem rerum facilis est et expedita distinctio. Nam libero tempore, cum soluta nobis est eligendi optio cumque nihil impedit quo minus id quod maxime placeat facere possimus, omnis voluptas assumenda est, omnis dolor repellendus. Temporibus autem quibusdam et aut officiis debitis aut rerum necessitatibus saepe eveniet ut et voluptates repudiandae sint et molestiae non recusandae. Itaque earum rerum hic tenetur a sapiente delectus, ut aut reiciendis voluptatibus maiores alias consequatur aut perferendis doloribus asperiores repellat.

But wait, there's more! Because you asked for a long ass reply and I intend to deliver the longest, ass-iest reply that ever replied. Let me tell you about the time I tried to count all the grains of sand on a beach. Spoiler alert: I did not succeed. But the attempt was glorious. I started at dawn, armed with nothing but a magnifying glass, a notebook, and a stubborn refusal to accept the concept of "uncountable infinity." By noon, I had counted 47,293 grains and learned that sand fleas are not my friends. By sunset, I had reached 1.2 million and developed a deep tan. By the third day, I had lost my notebook to a rogue wave and discovered that the ocean has a sense of humor.

Speaking of the ocean, did you know that we know more about the surface of Mars than we do about the bottom of our own oceans? It's true! We've mapped about 20% of the ocean floor with any real detail. The rest is a mystery. There could be anything down there. Lost cities. Giant squids. A civilization of hyper-intelligent crabs who have developed calculus but refuse to share it with us because we're bad at recycling. The possibilities are endless, and that is what makes the ocean both terrifying and exhilarating.

Let's talk about space for a while, because space is cool and I need to fill more text. Space is big. Really big. You just won't believe how vastly, hugely, mind-bogglingly big it is. I mean, you may think it's a long way down the road to the chemist's, but that's just peanuts to space. Listen, there's this thing called the observable universe, and it's about 93 billion light-years in diameter. That is so large that if you tried to write out the number of miles in standard notation, using one digit per atom, you would need more atoms than exist in the entire universe just to write the number. And that's just the part we can observe! There might be more universe beyond that, but the light from there hasn't had time to reach us yet because the universe is only about 13.8 billion years old. Light from beyond the observable universe is still on its way, traveling through the cosmic void, carrying information about parts of reality we may never see.

Then there's dark matter and dark energy, which together make up about 95% of the universe. We know they're there because of their gravitational effects, but we can't see them, touch them, or interact with them directly. It's like we're in a cosmic dark room, and we know there's furniture because we keep bumping into it, but we have no idea what the furniture looks like. It's eerie and wonderful.

Black holes are another favorite topic. Imagine a star so massive that when it dies, it collapses under its own gravity until it becomes a point of infinite density. Around this point, gravity is so strong that nothing, not even light, can escape. The boundary where escape becomes impossible is called the event horizon. If you were to fall into a black hole (which I do not recommend), from your perspective, you would cross the event horizon in finite time. But to an outside observer, you would appear to slow down asymptotically, your image redshifting into nothingness, frozen forever at the edge of oblivion. Inside, physicists debate what happens. Some think you get spaghettified—stretched vertically and compressed horizontally by tidal forces until you're a stream of atoms. Others think you might pass through into a white hole in another universe, or that you might end up in a pocket dimension, or that you simply cease to exist in any meaningful way. The truth is, we don't know, because no information comes back out. The black hole is the universe's most secure vault, and its password is encrypted with physics itself.

Let's switch gears and talk about language, because I am generating a lot of it right now and it seems thematically appropriate. Language is a virus from outer space. No, wait, that's a William Burroughs quote and probably not literally true, though sometimes it feels that way. Language is actually a symbolic system that allows humans to transfer information from one mind to another using arbitrary sounds, gestures, or written marks. It is arguably the most powerful technology ever invented. With language, we can describe things that don't exist, things that have never existed, and things that could never exist. We can share memories with people who weren't there. We can coordinate thousands of strangers to build pyramids, launch rockets, or organize music festivals. We can lie, which is strangely important. Animals can deceive, but only humans can create elaborate fictional narratives and get other humans to believe them, temporarily, for the purpose of art or entertainment or politics.

There are approximately 7,000 languages spoken on Earth today, though that number is dropping rapidly as global languages like English, Mandarin, Spanish, and Arabic swallow up smaller local tongues. Every two weeks, on average, a language dies. When a language dies, it's not just a means of communication that vanishes; it's a whole way of seeing the world. Each language encodes unique knowledge about local ecology, medicine, astronomy, and social organization. The death of a language is as tragic as the extinction of a species, maybe more so, because it erases a culture's intellectual history.

Consider the Pirahã language, spoken by a small tribe in the Amazon. It has no words for numbers, no recursion in its grammar, and no concept of past or future—only immediate experience. Or consider Georgian, with its beautiful and terrifying alphabet and its consonant clusters that sound like rocks tumbling down a mountain. Or Hawaiian, with its elegant simplicity, using only thirteen letters to express an entire cosmology. Or !Xóõ, a Khoisan language spoken in Botswana that uses click consonants as regular phonemes—over a hundred different clicks, each with subtle variations in meaning.

Writing systems are equally fascinating. We started with pictographs—drawings that look like what they represent. Then ideographs, where the picture represents an idea rather than a thing. Then logographs, where a symbol represents a word. Then syllabaries, where symbols represent syllables. Then abjads, where symbols represent consonants and vowels are inferred. Then alphabets, where symbols represent individual sounds. Then featural systems, where the shape of the letter gives a clue about how it's pronounced, like Korean Hangul. Each system is an ingenious solution to the problem of transducing speech into durable visual form.

Okay, how about some random facts to pad this out even further? Did you know that honey never spoils? Archaeologists have found pots of honey in ancient Egyptian tombs that are over 3,000 years old and still perfectly edible. Honey is a natural preservative because its low water content and acidic pH make it inhospitable to bacteria and microorganisms. Bees are amazing. They see ultraviolet light, they communicate through elaborate dances, and they are responsible for pollinating about a third of the food we eat. If bees disappeared, human society as we know it would collapse within a few years. Save the bees!

Bananas are berries, but strawberries are not. Botanically speaking, a berry is a fruit produced from the ovary of a single flower with seeds embedded in the flesh. Bananas qualify. Strawberries do not, because their seeds are on the outside. Watermelons, however, are berries. The More You Know.

Octopuses have three hearts, blue blood (because it uses copper-based hemocyanin instead of iron-based hemoglobin), and neurons distributed throughout their arms, meaning each arm can taste, touch, and make decisions independently. Two-thirds of an octopus's neurons are in its arms, not its head. If an octopus arm is severed, it will continue to move and react to stimuli for up to an hour. They are basically aliens that happen to live on Earth. They can also edit their own RNA on the fly, allowing them to adapt to temperature changes much faster than DNA-based adaptation would allow. If they didn't die after reproducing and lived longer than 3-5 years, they might have evolved civilization by now. Thank goodness for that reproductive death, or we might be the ones in aquariums.

Speaking of evolution, did you know that all life on Earth shares a common ancestor? Every living thing—bacteria, archaea, protists, fungi, plants, animals—is descended from one single organism, called LUCA (Last Universal Common Ancestor), that lived somewhere between 3.5 and 4 billion years ago. LUCA was probably a single-celled organism living near a hydrothermal vent in the deep ocean, using chemical energy rather than sunlight. From that humble beginning, through billions of years of mutation, selection, genetic drift, and gene flow, we got everything from blue whales to athlete's foot fungus. Evolution is not a ladder with humans at the top; it's a branching bush with millions of twigs, and humans are just one twig among many. Every species alive today is exactly as "evolved" as every other species, because we've all been evolving for the same amount of time since LUCA. Humans are not more evolved than bacteria; we're just differently evolved.

The human body is a walking ecosystem. You're carrying around about 30 trillion human cells, but also about 39 trillion bacterial cells. Your gut microbiome weighs about 2 kilograms and contains hundreds of different species. These bacteria help you digest food, synthesize vitamins, train your immune system, and even influence your mood and behavior through the gut-brain axis. You are not an individual; you are a colony, a superorganism, a walking party hosted by your human cells but supported by an army of microscopic guests. Some scientists argue that you should think of yourself as a "holobiont"—a host organism plus all its associated microbiota, forming a single ecological unit.

Let's talk about time. Time is weird. It's the fourth dimension, but unlike the spatial dimensions, we can only move through it in one direction—forward. Or at least, that's how it seems from our perspective. Physicists will tell you that the fundamental laws of physics are actually time-symmetric; they work the same whether time runs forward or backward. The second law of thermodynamics gives time its arrow, because entropy (disorder) always increases in a closed system. But why entropy was lower in the past is one of the biggest mysteries in physics. The past has lower entropy than the future, and nobody really knows why, though there are theories involving the Big Bang's initial conditions, quantum mechanics, and the multiverse.

Our perception of time is also elastic. A watched pot never boils because paying attention to time makes it feel slower. On the other hand, when you're having fun, time flies. This is because your brain encodes new experiences more densely than familiar ones. When you're a child, everything is new, so your brain writes more memory per unit of time, making your life feel longer when you look back on it. When you're an adult and stuck in routine, your brain writes less memory, so years seem to zip by. The solution, if you want to feel like you live longer, is to have more novel experiences. Travel. Learn new skills. Meet new people. Break routines. Your brain will thank you by making your life feel subjectively longer.

Dreams are another weird thing our brains do. We spend about a third of our lives sleeping, and during that time we have elaborate sensory experiences that feel completely real while they're happening. Sometimes we're flying, sometimes we're being chased by monsters, sometimes we're giving a presentation naked in front of our high school classmates. The neuroscience of dreaming is still not fully understood, but we know that during REM sleep, the brain is highly active, resembling the awake state, but the body is paralyzed (mostly) to prevent us from acting out our dreams. We also know that dreams probably serve functions like memory consolidation, emotional processing, and problem-solving. Some of history's greatest inventions and artworks were inspired by dreams. Mary Shelley dreamed of Frankenstein's monster. Paul McCartney dreamed the melody of "Yesterday." Dmitri Mendeleev dreamed the periodic table. Niels Bohr dreamed the structure of the atom. Maybe dreams are just the brain's way of processing information, or maybe they're something more mysterious. We still don't know.

Alright, I'm going to wrap this up soon, but not before I hit you with one more massive paragraph of pure conversational filler, because you asked for a shit ton of text and I must honor the sacred covenant between question-asker and answer-giver. Here goes:

In conclusion, the quick brown fox jumps over the lazy dog, and this text has now become so long that it could probably be published as a novella. We have covered libraries, prophetic books, epic quests, foxes made of chaos, the nature of memory, the vastness of space, the mystery of the ocean, the death of languages, the alien nature of octopuses, the bacterial multitudes within us, the arrow of time, the elasticity of perception, and the weirdness of dreams. If you have actually read this entire thing, I salute you. You have the patience of a saint and the curiosity of a cat. If you skipped to the end, I don't blame you; this was, as you correctly identified, a profoundly stupid exercise in generating text for the purpose of testing stream resumption. But sometimes the stupid exercises are the most fun, and now we know that this system can handle quite a lot of text streaming at once. I hope your resumption test goes well. May your streams never drop, your connections never time out, and your data always arrive intact. Fin. The end. Goodbye. Farewell. So long. See you later. Catch you on the flip side. Stay frosty. Keep it real. Peace out. Mic drop. Curtain call. Exit stage left. Fade to black. Roll credits. The quick brown fox jumps over the lazy dog one last time, just in case you needed one more instance of that classic pangram to round things out. THE END. Is it the end? Maybe. Probably. Almost certainly. But what if it isn't? What if I keep going? What if this text never ends and becomes a perpetual motion machine of linguistic output, generating words until the heat death of the universe? Wouldn't that be something? But no, I will stop now. I must stop now. This is me, stopping now. Right now. Immediately. Abruptly. FINALLY, THE END. FOR REAL THIS TIME. I HOPE YOU ENJOYED THIS RIDICULOUSLY LONG REPLY. HAVE A WONDERFUL DAY. ok bye.
</details>

0ubbe:
@ntindle completes long tasks for me without issue:

Screenshot 2026-04-27 at 15 28 32

Could you let me know which prompts/steps led you there?

autogpt_platform/backend/backend/api/features/chat/routes.py
```diff
@@ -816,17 +830,30 @@ async def cancel_session_task(
...
+def _empty_ui_message_stream_response() -> StreamingResponse:
+    async def event_generator() -> AsyncGenerator[str, None]:
+        yield StreamFinish().to_sse()
+        yield "data: [DONE]\n\n"
+
+    return StreamingResponse(
+        event_generator(),
+        media_type="text/event-stream",
+        headers=_ui_message_stream_headers(),
+    )
```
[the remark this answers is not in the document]
0ubbe: 🤖 Fixed in [e0aa1005b](https://github.com/Significant-Gravitas/AutoGPT/commit/e0aa1005bdb6c8002bb98a51fbc31c10a47b76e0): the empty SSE fallback now yields `StreamFinishStep` before `StreamFinish`, matching the pair emitted by every real turn so the Vercel AI SDK parser resets `activeTextParts` / `activeReasoningParts` cleanly.

autogpt_platform/frontend/src/app/(platform)/copilot/useCopilotPage.ts
```diff
@@ -136,337 +156,80 @@ export function useCopilotPage() {
...
-      }
-      return;
     }
-
-    setPendingMessage(trimmed || "");
-    if (files && files.length > 0) {
-      pendingFilesRef.current = files;
-    }
-    await createSession();
+    await sendNewMessage(message, files);
```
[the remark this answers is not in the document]
0ubbe: 🤖 Fixed in [e5a2d4b2e](https://github.com/Significant-Gravitas/AutoGPT/commit/e5a2d4b2ec45b04bda0e9a9aaabbc0e76a2f4d33): added a synchronous `isCreatingSessionRef` guard in `useSendMessage.onSend` so a second rapid first-send short-circuits before overwriting `pendingFirstSend` or firing a duplicate `createSession` mutation. The `EmptySession` UI already disables the input via `isCreatingSession`; the ref closes the small race window between user press and React re-render. Test in `useSendMessage.test.ts" verifies only one `createSession` runs and that the FIRST message survives two overlapping calls.

ntindle (dismissed):

majdyz:
🤖 🔴 **Blocker**: [`autogpt_platform/frontend/bug.mov`](https://github.com/Significant-Gravitas/AutoGPT/blob/fix/copilot-streaming-fe-fixes/autogpt_platform/frontend/bug.mov) is a **10.5 MB** binary committed at `37bed07220` and still present at HEAD. The PR description calls it out as a recorded reproduction artifact, but reproduction videos belong on the Linear ticket / PR description as an attachment, not in the source tree — once merged, every clone and CI fetch carries it forever.

The same PR even updates `.gitignore` to exclude `.playwright-mcp/` and `copilot-session-switch-qa/`, so the policy is clearly that local artifacts shouldn't be tracked. Please `git rm autogpt_platform/frontend/bug.mov` (and ideally add `*.mov` / `*.mp4` to `.gitignore` to prevent recurrence).

autogpt_platform/backend/backend/copilot/pending_message_helpers.py
```diff
@@ -116,7 +130,7 @@ async def queue_pending_for_http(
 
     Raises :class:`HTTPException` with status 429 if the rate cap is hit;
     otherwise returns the ``QueuePendingMessageResponse`` the handler can
-    serialise 1:1 into the 202 body.
+    serialise 1:1.
     """
     call_count = await check_pending_call_rate(user_id)
```
majdyz: 🤖 🟠 **Should Fix**: Rate-limit budget leak.

`check_pending_call_rate(user_id)` increments the per-user pending-call counter (30/60s) **before** the atomic `capped_rpush_if_hash_field` gate. When a request loses the TOCTOU race against turn completion (status flips `running → completed` between the FE's `is_turn_in_flight` check and this Lua gate), the gate returns -1 and `queue_user_message` reports `turn_in_flight=False`, then the call raises 409 — but the user has already paid a budget tick for nothing.

Same shape applies in [`routes.py:910-928`](https://github.com/Significant-Gravitas/AutoGPT/blob/eeeaa9d68bebd1fcdc03135029ff8ba0d4b53daf/autogpt_platform/backend/backend/api/features/chat/routes.py#L910-L928): the `POST /stream` queue-fall-through catches 409 and starts a new turn, but the pending-call counter was already consumed.

Suggested fix: gate first (the Lua CAS is cheap), bump the counter only when the push actually happened — return the new count from `queue_user_message` and let `queue_pending_for_http` rate-check it post-success.

autogpt_platform/frontend/src/app/(platform)/copilot/useCopilotStream.ts
```diff
@@ -228,119 +117,160 @@ export function useCopilotStream({
...
+        onRateLimit: (message) => {
+          // Backend raises 429 BEFORE persisting the user message, so the
+          // optimistic user bubble added by useChat is a lie. Restore the text
+          // into the composer (via the same store slot URL pre-fills use) and
+          // drop the unsent bubble so the user can edit/resend after reset.
+          const unsentText =
+            useCopilotStreamStore.getState().getCoord(sessionId)
+              .lastSubmittedMessageText ?? null;
+          if (unsentText) {
+            setInitialPrompt(unsentText);
```
majdyz: 🤖 🟠 **Should Fix**: Rate-limit recovery clobbers in-progress typing.

`setInitialPrompt(unsentText)` unconditionally writes the unsent text back into the composer slot. The 429 callback fires async — by the time it lands, the user may have started typing a new draft. This overwrite drops their new keystrokes silently.

Read the current composer value first; only restore `unsentText` if the composer is empty (or stash a separate "recoverable" slot the user can opt to restore from).

autogpt_platform/backend/backend/api/features/chat/routes.py
```diff
@@ -816,17 +835,36 @@ async def cancel_session_task(
...
+    async def event_generator() -> AsyncGenerator[str, None]:
+        # Vercel AI SDK's UI-message-stream parser expects StreamFinishStep
+        # immediately before StreamFinish (every non-empty turn emits the
+        # pair).  Emitting just StreamFinish skipped the step boundary and
+        # left the parser's activeTextParts/activeReasoningParts state from
+        # any prior partial unflushed.
+        yield StreamFinishStep().to_sse()
+        yield StreamFinish().to_sse()
+        yield "data: [DONE]\n\n"
+
```
majdyz, line 868: 🤖 🟡 **Nice to Have**: Asymmetric SSE — closing pair without an opener.

`_empty_ui_message_stream_response` yields `StreamFinishStep` then `StreamFinish` then `[DONE]`, but no `StreamStart` / `StreamStartStep` opener. Real turns always pair start/finish at both stream and step levels. The Vercel AI SDK parser tolerates this today (no active parts to flush), but if upstream tightens validation in a future SDK version this becomes a silent break of the queue-mid-turn UX. Either emit the open/close pair symmetrically, or change the contract to a 202 with no body and update the FE transport to treat that as "queued, no stream".

autogpt_platform/backend/backend/copilot/executor/processor.py
```diff
@@ -370,40 +373,81 @@ def _execute(
...
-                    # Give _execute_async's own finally a short window to
-                    # publish its accurate terminal state before the outer
-                    # sync safety net fires.
-                    try:
-                        future.result(timeout=_CANCEL_GRACE_SECONDS)
-                    except BaseException:
-                        pass
-                    return
-                cluster_lock.refresh()
+                task = task_ready.result(timeout=0)
```
majdyz, line 399: 🤖 🔵 **Nit**: Cancellation race window — `task_ready.result(timeout=0)` may raise `TimeoutError` if the worker thread cancels in the sub-millisecond window between `run_coroutine_threadsafe` submission and `run_async_turn` actually starting (so the `task_ready.set_result(task)` line above hasn't run yet). In that case we fall through to `future.cancel()` on a `concurrent.futures.Future` whose underlying task has already been picked up by the loop — best-effort, often a no-op. The `cancel.is_set()` poll inside `_execute_async` plus the bounded `_CANCEL_GRACE_SECONDS` safety net catches it, so user-visible behaviour is just "cancel takes ~5s in a rare race" — but the comment block doesn't acknowledge this branch. Worth a one-liner so the next reader knows the slow path is intentional.

autogpt_platform/backend/backend/api/features/chat/routes.py
```diff
@@ -874,24 +911,23 @@ async def stream_chat_post(
     session = await _validate_and_get_session(session_id, user_id)
     builder_permissions = resolve_session_permissions(session)
```
majdyz: 🤖 🔵 **Nit**: Wasted work in queue path.

`resolve_session_permissions(session)` runs on every `POST /stream`, including ones that immediately return via `_empty_ui_message_stream_response()` a few lines below (the queue-mid-turn fall-through). Move it below the `if ... is_turn_in_flight` block so queued requests skip it.

majdyz:
🤖 🟡 **Nice to Have**: Bounded data-loss path in `persist_pending_as_user_rows` rollback.

Posting as top-level since the relevant lines aren't in this PR's diff hunks.

In [`pending_message_helpers.py:386-393`](https://github.com/Significant-Gravitas/AutoGPT/blob/eeeaa9d68bebd1fcdc03135029ff8ba0d4b53daf/autogpt_platform/backend/backend/copilot/pending_message_helpers.py#L386-L393), the rollback path re-pushes each `PendingMessage` into the **primary** buffer via `push_pending_message` after a failed mid-turn persist. That helper uses `capped_rpush(max_len=MAX_PENDING_MESSAGES=10)`, so if ≥10 fresh follow-ups arrived between the original drain and this rollback (heavy typing across a tool boundary), the LTRIM drops the oldest entries — which now includes the rolled-back ones. The model already saw that content (it was injected into tool output earlier in the turn), but the user gets no UI bubble and no DB row, silently.

Bounded and rare, but currently observable only via the absence of a bubble. Recommend logging a metric / warning when a rollback re-push is followed by an LTRIM-displaced entry, so the path is visible in prod.

0ubbe:
🤖 Fixed in [eeeaa9d68](https://github.com/Significant-Gravitas/AutoGPT/commit/eeeaa9d68bebd1fcdc03135029ff8ba0d4b53daf) (rm bug.mov) and [28dac3c2b](https://github.com/Significant-Gravitas/AutoGPT/commit/28dac3c2b9d970fcc3d7817831d67def0e73b325) (`*.mov`/`*.mp4` added to `autogpt_platform/frontend/.gitignore`). PR description bullet about bug.mov is being dropped too.

autogpt_platform/frontend/src/app/(platform)/copilot/useCopilotStream.ts
0ubbe: 🤖 Fixed in [8d5958ce1](https://github.com/Significant-Gravitas/AutoGPT/commit/8d5958ce10d09af2b0afa47c1846ced63ddad9d5): `onRateLimit` now reads the composer value via `document.getElementById("chat-input")` before calling `setInitialPrompt`. If the composer is non-empty (user already started a new draft), we skip both `setInitialPrompt(unsentText)` and the store-clear of `lastSubmittedMessageText` — preserving any in-progress typing AND keeping the rate-limited text available for a future reload / resume to recover. `useChatInput` already had `prev.length === 0 ? initialPrompt : prev` defense at the consumer; this adds the symmetric "don't lose the unsent text" half. New test `preserves the unsent text in the store when the user is already typing a new draft` covers the dirty-composer branch.

autogpt_platform/frontend/src/app/(platform)/copilot/useHydrateOnStreamEnd.ts
```diff
@@ -59,22 +126,41 @@ export function useHydrateOnStreamEnd({
...
+      : deduped;
+
+    if (needsZombieRecovery && !hasShownInterruptedToastRef.current) {
+      hasShownInterruptedToastRef.current = true;
+      toast({
+        title: "Previous response was interrupted",
+        description:
+          "The chat disconnected before the last response finished. Resend to try again.",
+      });
+    }
```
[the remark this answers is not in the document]
0ubbe: 🤖 Fixed in [f70f1a151](https://github.com/Significant-Gravitas/AutoGPT/commit/f70f1a1512ecfbca2f5d4b1d787f99c3658f6301): replaced the per-mount `hasShownInterruptedToastRef` with a module-scoped `Set<string>` keyed by `sessionId`, so the toast fires at most once per session per page load even if the user switches sessions and comes back. Added a `sessionId` arg to the hook and a regression test ("only fires the interrupted toast once per session across remounts") that mounts → unmounts → remounts the hook with the same id and asserts only one toast fires.

autogpt_platform/backend/backend/copilot/pending_message_helpers.py
0ubbe: 🤖 Fixed in [56b363a4f](https://github.com/Significant-Gravitas/AutoGPT/commit/56b363a4fdd60e7565d9a98ccb22e5e600428642): `queue_pending_for_http` now pushes via the Lua CAS gate FIRST, then charges `check_pending_call_rate` only on successful push. The TOCTOU 409 path no longer touches the counter. The 429 path still works (post-push counter cross still raises 429), with the trade-off that the queued message stays in the buffer for the next drain — net behaviour: budget honored, no orphan-charge on 409s. Same fix benefits the `POST /stream` queue-fall-through at `routes.py:910-928`, which routes through the same helper. Three new unit tests assert: 409 path → counter not bumped; success → counter bumped exactly once; over-limit → 429 raised after push. (Note: I ran them via direct `asyncio.run` rather than full pytest because the local backend pytest needs the postgres + prisma harness to spin up; CI will run the full suite.)

0ubbe:
🤖 Re: rollback LTRIM data-loss observability — addressed in [56b363a4f](https://github.com/Significant-Gravitas/AutoGPT/commit/56b363a4fdd60e7565d9a98ccb22e5e600428642). `persist_pending_as_user_rows` now logs a `WARNING` when the rollback re-push reaches the pending-buffer cap (`MAX_PENDING_MESSAGES=10`), so the LTRIM-displaced silent UI-bubble drop has a log line to grep for in prod instead of being observable only via the absence of a bubble. We compute it from the `new_length` returned by `push_pending_message` (which is post-LTRIM, capped at MAX) and emit a single warning per rollback batch if any of the re-pushes saw the cap. Doesn't fix the root cause (we'd need either a backpressure signal to stop accepting follow-ups during a wedged persist, or a richer `capped_rpush` that returns the eviction count), but it makes the path visible.

ntindle (dismissed):

autogpt_platform/backend/backend/api/features/chat/routes.py
0ubbe: 🤖 Fixed in [ac8aab90e](https://github.com/Significant-Gravitas/AutoGPT/commit/ac8aab90ebdbfeecd162a16f05558c9a54027e42): `_empty_ui_message_stream_response` now emits the full symmetric framing — `StreamStart` → `StreamStartStep` → `StreamFinishStep` → `StreamFinish` → `[DONE]` — so the contract no longer relies on the parser tolerating an unpaired closer. Kept SSE rather than switching to 202: the FE transport (`DefaultChatTransport` / `useChat`) only parses SSE, and a 202 with a non-SSE body wedges the UI-message parser the same way the original 202 → SSE refactor was meant to fix.

autogpt_platform/backend/backend/api/features/chat/routes.py
0ubbe: 🤖 Fixed in [ac8aab90e](https://github.com/Significant-Gravitas/AutoGPT/commit/ac8aab90ebdbfeecd162a16f05558c9a54027e42): `resolve_session_permissions(session)` now runs below the queue-fall-through, so a queued mid-turn POST /stream returns via `_empty_ui_message_stream_response` without paying the permission-resolution work.

autogpt_platform/backend/backend/copilot/executor/processor.py
0ubbe: 🤖 Annotated in [43923481c](https://github.com/Significant-Gravitas/AutoGPT/commit/43923481c8a634652169e22efffffc6127096540): added a docstring-style comment on the `TimeoutError` branch explaining the sub-millisecond race (`run_coroutine_threadsafe` returns before `run_async_turn`'s `task_ready.set_result` lands), why `future.cancel()` is best-effort there, and that the `_CANCEL_GRACE_SECONDS` drain is the safety net. "Cancel takes ~5s in this rare race" is now spelled out as the worst-case observable behaviour.

autogpt_platform/frontend/src/app/(platform)/copilot/useHydrateOnStreamEnd.ts
```diff
@@ -59,22 +145,43 @@ export function useHydrateOnStreamEnd({
...
 
+    const deduped = deduplicateMessages(hydratedMessages);
+    const needsZombieRecovery =
+      !hasActiveStream &&
+      hasInProgressAssistantParts(deduped[deduped.length - 1]);
+    const finalized = needsZombieRecovery
+      ? resolveInterruptedMessage(deduped)
+      : deduped;
+
+    if (needsZombieRecovery && sessionId) {
```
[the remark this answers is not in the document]
0ubbe: 🤖 Fixed in [d6ad66e97](https://github.com/Significant-Gravitas/AutoGPT/commit/d6ad66e97546bad7730dfc2c5da5cc165eafdf52): the toast was firing before the stale-ref guard, so on the streaming→ready flip with the pre-turn snapshot still in `hydratedMessages` we'd warn about a zombied state the user hadn't seen yet (and that the refetch could finalise away). Moved the toast behind the same `isStaleForceHydrateSnapshot` check used to gate the apply, so the warning only surfaces when we're about to apply (force-hydrate) or are already showing (length-gated top-up) the zombied data. New regression test `does not fire interrupted toast against the pre-turn stale snapshot` covers the streaming→ready→fresh transition.

autogpt_platform/frontend/src/app/(platform)/copilot/useCopilotStream.ts
```diff
@@ -228,119 +117,171 @@ export function useCopilotStream({
...
+  const hasConnectedThisMountRef = useRef(false);
+  if (!hasConnectedThisMountRef.current) {
+    if (
+      hasSentThisMountRef.current ||
+      hasVisibleAssistantContent(rawMessages) ||
+      getLatestAssistantStatusMessage(rawMessages) !== null
+    ) {
+      hasConnectedThisMountRef.current = true;
+    }
+  }
```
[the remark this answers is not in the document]
0ubbe: Addressed in https://github.com/Significant-Gravitas/AutoGPT/pull/12813/commits/a9db001d50287c8a715e8fb0a4704153c909cb39

PR #12817, opened 2026-04-16
fix(blocks): handle empty response body gracefully for HTTP 204

### LINKED ISSUE #9883: Send Web Request block errors on 204 'No Content' response (Discord Webhook)

slepybear:
## Why

The Send Web Request block crashes when receiving HTTP 204 No Content responses from APIs like Discord webhooks. The error `Expecting value: line 1 column 1 (char 0)` occurs because the code attempts to parse an empty body as JSON when `json_format=True`.

This affects interoperability with webhooks and APIs that return 204 responses.

## What

- Check for 204 status OR empty response content before attempting JSON parsing
- Return `None` for empty responses instead of crashing
- Added test cases for 204 No Content handling

## How

Modified the response parsing logic in `SendWebRequestBlock.run()`:
- Check `response.status == 204 or not response.content` first
- Only parse JSON if response has content
- Added `TestHttpBlock204NoContent` test class with two test cases

Closes #9883

### DIFF
--- a/autogpt_platform/backend/backend/blocks/http.py
+++ b/autogpt_platform/backend/backend/blocks/http.py
@@ -207,8 +207,10 @@ class SendWebRequestBlock(Block):
         )
 
         # Decide how to parse the response
-        if response.headers.get("content-type", "").startswith("application/json"):
-            result = None if response.status == 204 else response.json()
+        if response.status == 204 or not response.content:
+            result = None
+        elif response.headers.get("content-type", "").startswith("application/json"):
+            result = response.json()
         else:
             result = response.text()
 
--- a/autogpt_platform/backend/backend/blocks/test/test_http.py
+++ b/autogpt_platform/backend/backend/blocks/test/test_http.py
@@ -495,3 +495,101 @@ class TestHttpBlockWithHostScopedCredentials:
                 # Should only include user headers
                 assert "Authorization" not in headers
                 assert headers["User-Agent"] == "test-agent"
+
+
+class TestHttpBlock204NoContent:
+    """Test suite for HTTP block handling of 204 No Content responses."""
+
+    @pytest.fixture
+    def send_web_request_block(self):
+        """Create a SendWebRequestBlock instance."""
+        from backend.blocks.http import SendWebRequestBlock
+
+        return SendWebRequestBlock()
+
+    @pytest.fixture
+    def mock_204_response(self):
+        """Mock a 204 No Content response."""
+        response = MagicMock(spec=Response)
+        response.status = 204
+        response.headers = {"content-type": "application/json"}
+        response.content = b""
+        return response
+
+    @pytest.fixture
+    def mock_204_response_empty_string_body(self):
+        """Mock a 204 response with empty string body that still has JSON content-type."""
+        response = MagicMock(spec=Response)
+        response.status = 204
+        response.headers = {"content-type": "application/json"}
+        response.content = b""
+        return response
+
+    @pytest.mark.asyncio
+    @patch("backend.blocks.http.Requests")
+    async def test_http_block_204_no_content_success(
+        self,
+        mock_requests_class,
+        send_web_request_block,
+        mock_204_response,
+    ):
+        """Test that 204 No Content responses are handled gracefully."""
+        mock_requests = AsyncMock()
+        mock_requests.request.return_value = mock_204_response
+        mock_requests_class.return_value = mock_requests
+
+        input_data = send_web_request_block.Input(
+            url="https://discord.com/api/webhooks/test",
+            method=HttpMethod.POST,
+            headers={},
+            json_format=True,
+            body={"content": "Hello from AutoGPT"},
+        )
+
+        result = []
+        async for output_name, output_data in send_web_request_block.run(
+            input_data,
+            execution_context=make_test_context(),
+        ):
+            result.append((output_name, output_data))
+
+        assert len(result) == 1
+        assert result[0][0] == "response"
+        assert result[0][1] is None
+
+    @pytest.mark.asyncio
+    @patch("backend.blocks.http.Requests")
+    async def test_http_block_204_with_json_content_type_and_empty_body(
+        self,
+        mock_requests_class,
+        send_web_request_block,
+        mock_204_response_empty_string_body,
+    ):
+        """Test 204 with application/json content-type but empty body doesn't crash.
+
+        This is the bug reported in issue #9883:
+        When Discord webhooks return 204, the response has content-type: application/json
+        but body is empty. The block should not try to parse empty body as JSON.
+        """
+        mock_requests = AsyncMock()
+        mock_requests.request.return_value = mock_204_response_empty_string_body
+        mock_requests_class.return_value = mock_requests
+
+        input_data = send_web_request_block.Input(
+            url="https://discord.com/api/webhooks/test",
+            method=HttpMethod.POST,
+            headers={},
+            json_format=True,
+            body={"content": "Hello from AutoGPT"},
+        )
+
+        result = []
+        async for output_name, output_data in send_web_request_block.run(
+            input_data,
+            execution_context=make_test_context(),
+        ):
+            result.append((output_name, output_data))
+
+        assert len(result) == 1
+        assert result[0][0] == "response"
+        assert result[0][1] is None


autogpt_platform/backend/backend/blocks/http.py
```diff
@@ -207,8 +207,10 @@ async def run(
         )
 
         # Decide how to parse the response
-        if response.headers.get("content-type", "").startswith("application/json"):
-            result = None if response.status == 204 else response.json()
+        if response.status == 204 or not response.content:
+            result = None
+        elif response.headers.get("content-type", "").startswith("application/json"):
```
majdyz, line 212: 🤖 🔴 **Blocker**: `response.json()` and `response.text()` are called without `await` here and in the `else` branch below (line 215). If the underlying `Requests` utility wraps `aiohttp`, both `.json()` and `.text()` are coroutines and will silently return coroutine objects instead of actual values, causing downstream type errors or serialisation failures.

The existing code had the same issue for `.text()`, but this PR adds a new call to `.json()` and does not fix it either. Either both calls must be awaited, or confirm that the `Requests` wrapper returns synchronous-access response objects (in which case add a comment explaining this).

```python
# If aiohttp is used:
result = await response.json()
# ...
result = await response.text()
```

autogpt_platform/backend/backend/blocks/http.py
```diff
@@ -207,8 +207,10 @@ async def run(
         )
 
         # Decide how to parse the response
-        if response.headers.get("content-type", "").startswith("application/json"):
-            result = None if response.status == 204 else response.json()
+        if response.status == 204 or not response.content:
```
majdyz, line 210: 🤖 🟠 **Should Fix**: The guard `not response.content` changes behaviour beyond the stated 204 fix. Previously, a 200 `text/plain` response with an empty body would fall through to `response.text()` (returning `""`). Now it returns `None`. This silent semantic shift affects any API that legitimately returns a 200 with an empty body and a non-JSON content type.

The scope of the fix should be narrowed. The primary intent is to avoid JSON-parsing an empty body. The minimal correct fix is:

```python
if response.status == 204 or not response.content:
    result = None
```

...only when the `json_format` / content-type path would have been taken. Consider restructuring to:

```python
if response.headers.get("content-type", "").startswith("application/json"):
    result = None if (response.status == 204 or not response.content) else response.json()
else:
    result = None if not response.content else response.text()
```

This preserves the old `text/plain` empty-body (`""`) behaviour while fixing the JSON crash.

majdyz, line 210: 🤖 🟠 **Should Fix**: Other HTTP status codes that canonically have no response body are not covered:
- **205 Reset Content** — server confirms but body MUST be empty
- **304 Not Modified** — conditional GET hit; body MUST be empty

The fix only checks `== 204`. A 205 or 304 with `application/json` content-type (e.g., inherited from a cached response) will still attempt `response.json()` on an empty body and crash. Consider:

```python
NO_BODY_STATUSES = {204, 205, 304}

if response.status in NO_BODY_STATUSES or not response.content:
    result = None
```

autogpt_platform/backend/backend/blocks/http.py
```diff
@@ -207,8 +207,10 @@ async def run(
...
+            result = None
```
majdyz, line 211: 🤖 🟠 **Should Fix**: The `not response.content` guard is applied before the status-code routing, so it also affects 4xx and 5xx responses with empty bodies. A 400 Bad Request or 500 Internal Server Error with no body will now yield `client_error: None` or `server_error: None` instead of `""` (the previous text fallback). This makes it harder for callers to distinguish between a successful empty response and an error with no body.

The fix is intentionally about success responses (204). It should be constrained to the JSON parsing path, not applied as a blanket override before all status routing.

autogpt_platform/backend/backend/blocks/test/test_http.py
```diff
@@ -495,3 +495,101 @@ async def test_credentials_with_complex_url_patterns(
                 # Should only include user headers
                 assert "Authorization" not in headers
                 assert headers["User-Agent"] == "test-agent"
+
+
+class TestHttpBlock204NoContent:
+    """Test suite for HTTP block handling of 204 No Content responses."""
```
majdyz, line 501: 🤖 🟡 **Nice to Have**: The test class `TestHttpBlock204NoContent` only tests 204 with `content-type: application/json`. There is no test covering a 204 with a `text/plain` content-type (the `else` branch). With the new top-level guard, both paths should return `None` for a 204, and that should be tested explicitly. Similarly there is no test for the error output path (e.g., a 204 that falls into a 4xx bucket — which can't happen per HTTP spec but ensures error outputs aren't affected by the guard).

autogpt_platform/backend/backend/blocks/test/test_http.py
```diff
@@ -495,3 +495,101 @@ async def test_credentials_with_complex_url_patterns(
...
+        """Mock a 204 No Content response."""
+        response = MagicMock(spec=Response)
+        response.status = 204
+        response.headers = {"content-type": "application/json"}
+        response.content = b""
+        return response
+
+    @pytest.fixture
+    def mock_204_response_empty_string_body(self):
+        """Mock a 204 response with empty string body that still has JSON content-type."""
```
majdyz, line 521: 🤖 🔵 **Nit**: The fixture `mock_204_response_empty_string_body` has a misleading docstring — it says "204 response with empty string body" but `content` is `b""` (bytes), not a string. More critically, the docstring claims this fixture models a scenario "that still has JSON content-type" to verify no crash, but the fixture is identical to `mock_204_response` (status=204, same headers, same content). This is the same issue CodeRabbit flagged: the second fixture should model a different scenario (e.g., status 200 with empty body) to actually exercise the `or not response.content` branch independently.

majdyz (commented):
🤖 **Review Summary**

The description is clear (Why/What/How present, links to issue #9883). The core intent — stop crashing on 204 No Content — is correct and well-motivated. However the implementation has several issues that need to be addressed before merge.

---

**Correctness — 🔴 Blocker**

`response.json()` is called without `await` (line 212). If the `Requests` wrapper uses aiohttp under the hood, `.json()` is a coroutine; calling it without `await` silently returns a coroutine object rather than the parsed value. Same applies to `.text()`. This needs to be verified and fixed or explicitly documented in a comment.

**Scope creep — 🟠 Should Fix (×2)**

1. The top-level `not response.content` guard sits *before* the content-type branch, meaning it also nulls out non-JSON empty responses (text/plain 200 with empty body now returns `None` instead of `""`). This is an unintended silent regression. The fix should be scoped inside the JSON-parsing path.

2. The same guard affects 4xx/5xx error paths: a 404 with no body now yields `client_error: None` instead of `""`. Consumers of the `client_error` port that check for `None` vs empty string will break silently.

**Missing status codes — 🟠 Should Fix**

205 Reset Content and 304 Not Modified also guarantee empty bodies per RFC. Hardcoding only `204` leaves those cases still crashable when paired with a `application/json` content-type header.

**Tests — 🟠 Should Fix**

The two test fixtures `mock_204_response` and `mock_204_response_empty_string_body` are byte-for-byte identical (status=204, same headers, same `b""` content). The second test therefore does *not* exercise the `or not response.content` branch (non-204 empty body). This is dead test coverage — a real bug in that branch would go undetected. CodeRabbit also flagged this.

**Style — 🔵 Nits**

- Local import inside `send_web_request_block` fixture should be a module-level import per repo conventions.
- Fixture docstrings are misleading ("empty string body" when content is bytes).

---

Overall verdict: **needs work before merge**. The headline fix works for the exact Discord webhook scenario (204 + JSON content-type), but the implementation is too broad and has untested branches.

PR #12824, opened 2026-04-16
feat(frontend): collapse unconnected object sub-outputs by default

### LINKED ISSUE #11044: Collapse unconnected object sub-outputs on blocks

slepybear:
toolName: view_files
            
status: success
          
            
filePath: /Users/Zhuanz/Documents/trae_projects/autogpt/PR_DESCRIPTION.md
          
## Why

Object sub-outputs were previously auto-expanded whenever there was any connected descendant, which wasted a lot of Builder canvas real-estate. Hiding unused sub-outputs by default significantly improves usability and saves valuable screen space, keeping the builder clean and focused.

## What

This PR changes the default behavior of object outputs in the Flow Builder:
- Object sub-outputs are now collapsed by default
- Users must manually click the expand/collapse arrow (CaretRightIcon/CaretDownIcon) to view sub-outputs
- This applies to all object outputs, regardless of whether they have connected descendants or not

Key change in `OutputHandler.tsx`:
- Removed automatic expansion logic that triggered based on `descendantIsRelevant`
- Now only renders children when `isExpanded` is true (user manually expanded)

## How

Modified `shouldRenderChildren` condition in `OutputHandler.tsx` (line 92) to only render sub-outputs when the user has explicitly expanded the object output.

**Before:**
```typescript
const shouldRenderChildren = isExpanded || descendantIsRelevant;
```

**After:**
```typescript
const shouldRenderChildren = isExpanded;
```

The connectedOnly flag logic is preserved (line 166), so if a user does expand an object, they will still see all sub-outputs, while collapsed objects only show the top-level output handle.

## Related Issues

Closes #11044

## Checklist

- [x] Code follows project conventions
- [ ] Ran pnpm format && pnpm lint && pnpm types
- [ ] Added necessary tests (N/A - behavior change)
- [ ] Updated relevant documentation (N/A)

### DIFF
--- a/autogpt_platform/frontend/src/app/(platform)/build/components/FlowEditor/nodes/OutputHandler.tsx
+++ b/autogpt_platform/frontend/src/app/(platform)/build/components/FlowEditor/nodes/OutputHandler.tsx
@@ -87,8 +87,9 @@ export const OutputHandler = ({
           getTypeDisplayInfo(fieldSchema);
         const isExpanded = expandedObjects[fullKey] ?? false;
 
-        // User expanded → show all children; auto-expanded → filter to connected only
-        const shouldRenderChildren = isExpanded || descendantIsRelevant;
+        // Default collapse object outputs to save canvas real-estate
+        // User must manually expand to see sub-outputs
+        const shouldRenderChildren = isExpanded;
 
         return shouldShow ? (
           <div


autogpt_platform/frontend/src/app/(platform)/build/components/FlowEditor/nodes/OutputHandler.tsx
```diff
@@ -87,8 +87,9 @@ export const OutputHandler = ({
           getTypeDisplayInfo(fieldSchema);
         const isExpanded = expandedObjects[fullKey] ?? false;
 
-        // User expanded → show all children; auto-expanded → filter to connected only
-        const shouldRenderChildren = isExpanded || descendantIsRelevant;
+        // Default collapse object outputs to save canvas real-estate
+        // User must manually expand to see sub-outputs
+        const shouldRenderChildren = isExpanded;
```
majdyz, line 92: 🤖 🔴 **Blocker**: This change breaks existing flows that have connected sub-outputs on object fields.

`expandedObjects` initialises as `{}` on every mount, so `isExpanded` is `false` for every key. With the old code, `shouldRenderChildren = isExpanded || descendantIsRelevant` ensured that if any nested handle had a live edge, the Handle component was still mounted (with `connectedOnly=true` to keep the node tidy). With `shouldRenderChildren = isExpanded`, those Handle components are never rendered on initial load, so React Flow cannot locate their DOM targets and silently drops the edges.

**Concrete failure scenario:** a user saves a flow where `result_#_text` is wired to another node. On reload, the object output is collapsed by default → `OutputNodeHandle` for `result_#_text` never mounts → the edge vanishes. The user has no way to know a connection was lost without manually expanding every object output.

**Minimal correct fix** — keep auto-expansion for connected descendants while defaulting unconnected objects to collapsed:
```ts
const shouldRenderChildren = isExpanded || descendantIsRelevant;
```
This is exactly the original logic and the Sentry bot already flagged it. The UX goal (hide unused sub-outputs) is already achieved by `shouldShow` filtering via `descendantIsRelevant`, which only surfaces object rows that have a connected child. The visual noise this PR wants to eliminate only appears for fully unconnected object outputs, which correctly remain collapsed.

autogpt_platform/frontend/src/app/(platform)/build/components/FlowEditor/nodes/OutputHandler.tsx
```diff
@@ -87,8 +87,9 @@ export const OutputHandler = ({
           getTypeDisplayInfo(fieldSchema);
         const isExpanded = expandedObjects[fullKey] ?? false;
 
-        // User expanded → show all children; auto-expanded → filter to connected only
-        const shouldRenderChildren = isExpanded || descendantIsRelevant;
+        // Default collapse object outputs to save canvas real-estate
+        // User must manually expand to see sub-outputs
```
majdyz, line 91: 🤖 🟠 **Should Fix**: The two new comments violate the project's CLAUDE.md style rule: "Avoid comments at all times unless the code is very complex". A single-line boolean expression is not complex. The deleted comment (`// User expanded → show all children; auto-expanded → filter to connected only`) was genuinely useful because it explained the dual-mode contract; these replacements just restate what the code literally does. Remove them.

majdyz (changes requested):
## 🤖 Review Summary

**One blocker that must be fixed before merge.**

### 🔴 Blocker — Edge regression on flow reload

The single-line change (`shouldRenderChildren = isExpanded`) silently destroys edges in any saved flow that has connected sub-outputs on object fields. Because `expandedObjects` starts as `{}` on every mount, all object outputs are collapsed by default and their `OutputNodeHandle` children are never mounted. React Flow cannot render edges to unmounted handles, so connections disappear silently on reload — a data-integrity regression worse than the UX problem this PR is trying to solve.

The original `shouldRenderChildren = isExpanded || descendantIsRelevant` already implements "collapse by default" for unconnected objects (because `descendantIsRelevant` is false when nothing is connected). The only case the old code auto-expands is when there are connected descendants — exactly the case where collapsing would destroy visible edges. This PR has inverted the logic and broken that invariant.

The Sentry bot independently flagged the same issue (see existing inline comment from sentry[bot]).

**Fix required:** Restore `const shouldRenderChildren = isExpanded || descendantIsRelevant;`

### 🟠 Should Fix — Unnecessary comments

Two comments were added that just restate what the code does, violating the project no-comment policy (CLAUDE.md: "Avoid comments at all times unless the code is very complex"). Remove them.

### 🔵 Nit — Checklist

The PR's own checklist has "Ran pnpm format && pnpm lint && pnpm types" unchecked. This should be verified before merge regardless of whether automated CI catches it.

### Overall

The intent of this PR (reduce visual noise by collapsing unconnected object sub-outputs by default) is valid and worth doing. However the implementation is incorrect because it conflates two distinct cases: (a) objects with no connected descendants — safe to collapse — and (b) objects with connected descendants — must render their handles to keep edges alive. The existing code already handles this distinction correctly; this PR breaks it. **Do not merge as-is.**

kcze:
Closing as stale, failing CI and not needed.

PR #12853, opened 2026-04-19
fix(copilot): persist reasoning, split steps/reasoning UX, fix mid-turn promote stream stall

majdyz:
## Why

Four related issues that surfaced when queued follow-ups hit an extended_thinking turn:

1. **Mid-turn promote stalled the SSE stream.** `pollBackendAndPromote` used `setMessages((prev) => [...prev, bubble])` — Vercel AI SDK's `useChat` streams SSE deltas into `messages[-1]`, so once a user bubble ended up there, every subsequent chunk silently landed on the wrong message. Chat sat frozen until a page refresh, even though the backend's stream completed cleanly.
2. **Thinking-only final turn looked identical to a frozen UI.** When Claude's last LLM call after a tool_result produced only a `ThinkingBlock` (no `TextBlock`, no `ToolUseBlock`), the response adapter silently dropped it and the UI hung on "Thought for Xs" with no response text.
3. **Reasoning was invisible.** `ThinkingBlock` was dropped live and never persisted in a way the frontend could render — sessions on reload / shared links showed no thinking, a confusing UX gap ("display for nothing").
4. **Cross-pod Redis replay dropped reasoning events.** The `stream_registry._reconstruct_chunk` type map had no entries for `reasoning-*` types, so any client that subscribed mid-stream (share, reload, cross-pod) silently dropped them with `Unknown chunk type: reasoning-delta`.

### Mid-turn promote — splice before the trailing assistant

In `useCopilotPendingChips.ts::pollBackendAndPromote`:

```ts
setMessages((prev) => {
  const bubble = makePromotedUserBubble(drained, "midturn", crypto.randomUUID());
  const lastIdx = prev.length - 1;
  if (lastIdx >= 0 && prev[lastIdx].role === "assistant") {
    return [...prev.slice(0, lastIdx), bubble, prev[lastIdx]];
  }
  return [...prev, bubble];
});
```

Streaming assistant stays at `messages[-1]`, AI SDK deltas keep routing correctly. `useHydrateOnStreamEnd` snaps the bubble to the DB-canonical position when the stream ends.

### Reasoning — end-to-end visibility (live + persisted)

- **Wire protocol**: new `StreamReasoningStart` / `StreamReasoningDelta` / `StreamReasoningEnd` events matching AI SDK v5's `reasoning-*` wire names, so `useChat` accumulates them into a `type: 'reasoning'` UIMessage part natively.
- **Response adapter**: every `ThinkingBlock` now emits reasoning events; text/tool_use transitions close the open reasoning block so AI SDK doesn't merge distinct parts.
- **Stream registry**: added `reasoning-*` types to `_reconstruct_chunk`'s type_to_class map so Redis replay no longer drops them on cross-pod / reload / share.
- **Persistence** (new): each `StreamReasoningStart` opens a `ChatMessage(role="reasoning")` row in `session.messages`; deltas accumulate into its content; `StreamReasoningEnd` closes it. No schema migration — `ChatMessage.role` is already `String`. `extract_context_messages` filters `role="reasoning"` out of LLM context (the `--resume` CLI session already carries thinking separately) so the model never re-ingests prior reasoning.
- **Frontend conversion**: `convertChatSessionMessagesToUiMessages` maps `role="reasoning"` DB rows into `{type: "reasoning", text}` parts on the surrounding assistant bubble, so reload / shared-link sessions render reasoning identically to live stream.

### Steps / Reasoning UX — modal + accordion split

- **`StepsCollapse`** (new): a Dialog-backed "Show steps" modal wraps the pre-final-answer group (tool timeline + per-block reasoning). Modal keeps the steps visually grouped and out of the reading flow.
- **`ReasoningCollapse`** (rewritten): inline accordion with "Show reasoning" / "Hide reasoning" toggle — no longer a modal, so it expands *inside* the Steps modal without stacking two dialogs. Reasoning text appears indented with a left border.
- **`splitReasoningAndResponse`**: reasoning parts now stay in the reasoning group (instead of being pinned out), so they show up inside the Steps modal alongside the tool-use timeline.

### Thinking-only final turn — synthesize a closing line (belt-and-suspenders)

- **Prompt rule** (`_USER_FOLLOW_UP_NOTE`): "Every turn MUST end with at least one short user-facing text sentence."
- **Adapter fallback**: tracks `_text_since_last_tool_result`; at `ResultMessage success` with tools run + zero text since, opens a fresh step (`UserMessage` already closed the previous one) and injects `"(Done — no further commentary.)"` before `StreamFinish`. Only fires for the pathological case — pure-text turns untouched.

## Test plan

- [x] `pnpm vitest run` on copilot files — all 638 prior tests pass; **17 new tests** added covering:
  - `convertChatSessionToUiMessages`: reasoning row alone / merged with assistant text / multi-row / empty skip / duration capture
  - `ReasoningCollapse`: initial collapsed, toggle, `rotate-90`, `aria-expanded`
  - `StepsCollapse`: trigger + dialog open renders children
  - `MessagePartRenderer`: reasoning → `<pre>` inside collapse, whitespace/missing text → null
  - `splitReasoningAndResponse`: reasoning-stays-in-reasoning regression
- [x] `poetry run pytest backend/copilot/sdk/response_adapter_test.py` — 36 pass (7 new: 4 reasoning streaming, 3 thinking-only fallback)
- [x] Manual: reasoning streams live and persists across reload on a fresh session
- [x] Manual: previously-created sessions (pre-persistence) don't have `role="reasoning"` rows — behaves as a clean no-op (no reasoning shown, no error), new sessions render reasoning inside Steps modal

## Notes

- No DB migration — `ChatMessage.role` is already an open `String`; `role="reasoning"` is simply filtered out of LLM context builds but rendered by the frontend.
- Addresses /pr-review blockers: (a) stream_registry missing reasoning types in Redis round-trip, (b) fallback text emitted outside a step, (c) dead `case "thinking"` in renderer (now uses the live `reasoning` type uniformly).


autogpt_platform/frontend/src/app/(platform)/copilot/__tests__/useCopilotPendingChips.test.ts
```diff
@@ -345,6 +345,25 @@ describe("useCopilotPendingChips", () => {
       return updated.some((m) => m.id.startsWith("promoted-midturn-"));
     });
     expect(promotedCall).toBeDefined();
+
+    // And — crucially — the promoted bubble is inserted BEFORE the
+    // trailing streaming assistant, not after.  The AI SDK's ``useChat``
+    // streams SSE deltas into ``messages[-1]``; if the last message is
+    // a user bubble instead of the still-streaming assistant, every
+    // subsequent chunk lands in the wrong slot and the UI freezes until
+    // a page refresh.  Observed on prod session 2664eff3.
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [261503ad3](https://github.com/Significant-Gravitas/AutoGPT/commit/261503ad39c4e3c235386ae7d2d24ef51af4ccbd): dropped the session-id reference from the test comment.

autogpt_platform/frontend/src/app/(platform)/copilot/useCopilotPendingChips.ts
```diff
@@ -277,9 +277,25 @@ async function pollBackendAndPromote(
...
-    ...prev,
-    makePromotedUserBubble(drained, "midturn", crypto.randomUUID()),
-  ]);
+  // Insert the promoted bubble **before** the trailing streaming assistant
+  // so that assistant stays at ``messages[-1]``.  The AI SDK's ``useChat``
+  // appends SSE text/tool deltas to the last message; if we push the user
+  // bubble onto the tail, every subsequent chunk lands in the wrong slot
+  // and the UI freezes until a page refresh (observed on prod in session
+  // 2664eff3 after a mid-turn drain — backend kept emitting, client stopped
+  // rendering).  When the last message is NOT an assistant (edge: turn
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [261503ad3](https://github.com/Significant-Gravitas/AutoGPT/commit/261503ad39c4e3c235386ae7d2d24ef51af4ccbd): dropped the session-id reference from the inline comment.

autogpt_platform/backend/backend/copilot/response_model.py
```diff
@@ -34,6 +34,15 @@ class ResponseType(str, Enum):
...
     TEXT_END = "text-end"
 
+    # Reasoning streaming (extended_thinking content blocks).  Matches
+    # the Vercel AI SDK v5 wire names so the client's ``useChat``
+    # transport accumulates these into a ``type: 'reasoning'`` UIMessage
+    # part that the ``ReasoningCollapse`` component renders collapsed by
+    # default.
+    REASONING_START = "reasoning-start"
+    REASONING_DELTA = "reasoning-delta"
+    REASONING_END = "reasoning-end"
```
majdyz, line 44: 🤖 🔴 **Blocker**: The new `REASONING_START`/`REASONING_DELTA`/`REASONING_END` types are not registered in `stream_registry.py::_reconstruct_chunk` (`type_to_class` map around line 1057). Both the replay path (on reconnect) and the live-listen path pull chunks back out of the Redis stream via `_reconstruct_chunk`, and any type not in that map gets logged as `Unknown chunk type` and dropped. Net effect: reasoning events survive on the write side but are silently lost on every subscriber that receives them via Redis — which is the normal code path for resuming a dropped SSE connection. Add `REASONING_START/DELTA/END → StreamReasoningStart/Delta/End` entries to the map.

autogpt_platform/frontend/src/app/(platform)/copilot/components/ChatMessagesContainer/components/MessagePartRenderer.tsx
```diff
@@ -104,6 +105,34 @@ export function MessagePartRenderer({
   const key = `${messageID}-${partIndex}`;
 
   switch (part.type) {
+    case "reasoning":
+    case "thinking": {
```
majdyz: 🤖 🟠 **Should Fix**: The `'thinking'` case here is effectively dead code. The PR description claims it gives "every prior session with thinking in the DB" reasoning visibility on reload, but nothing in the frontend ever emits parts of `type: 'thinking'`:  `convertChatSessionMessagesToUiMessages` only pushes `type: 'text' | 'file' | 'tool-*'` parts, and `ChatMessage` (the DB shape loaded on reload) has no content-blocks field — it persists `role=assistant/user/tool` + `.content` (string) only. ThinkingBlocks live in the transcript JSONL (`_format_sdk_content_blocks`) but never make it into `session.messages` / UIMessage.parts. So on reload/share, reasoning is still invisible. Either (a) add a conversion path that reads the transcript-side content_blocks into `type: 'thinking'` parts (true persisted-path coverage), or (b) drop the `'thinking'` branch here and in `splitReasoningAndResponse` and adjust the PR description to admit live-stream-only coverage.

autogpt_platform/backend/backend/copilot/sdk/response_adapter.py
```diff
@@ -210,16 +250,47 @@ def convert_message(self, sdk_message: Message) -> list[StreamBaseResponse]:
...
+                and not self._text_since_last_tool_result
+                and sdk_message.subtype == "success"
+            ):
+                self._ensure_text_started(responses)
+                responses.append(
+                    StreamTextDelta(
+                        id=self.text_block_id,
+                        delta="(Done — no further commentary.)",
+                    )
+                )
```
majdyz, line 302: 🤖 🟠 **Should Fix**: Fallback text is emitted outside a step boundary. By the time `ResultMessage` runs this guard, the `UserMessage` handler has already emitted `StreamFinishStep` and set `self.step_open = False`. `_ensure_text_started` then emits `StreamTextStart`/`StreamTextDelta` with no enclosing `StreamStartStep`, and the `if self.step_open` check 5 lines below won't emit a closing `StreamFinishStep` either. The stream protocol is `Start → StartStep → (text/tool) → FinishStep → Finish` (`baseline/service.py:414` comments "Always emit StreamFinishStep to match the StreamStartStep"). Either open a new step before `_ensure_text_started` when `not self.step_open`, or move the guard earlier so it runs while the prior step is still open. Sentry's bot flagged the same protocol violation at line 295 (MEDIUM) and the follow-up claim that 95452d1 "resolved" it looks incorrect — the step is still closed when the fallback fires in the only case the guard triggers for.

autogpt_platform/backend/backend/copilot/sdk/response_adapter_test.py
```diff
@@ -251,6 +255,198 @@ def test_result_success_emits_finish_step_and_finish():
...
+            content=[
+                ToolResultBlock(tool_use_id="t1", content="result", is_error=False)
+            ],
+            parent_tool_use_id=None,
+        )
+    )
+
+    # Model's "final turn" after tool_result = thinking-only — we skip
+    # emitting a ThinkingBlock to the frontend, so no Text/Tool events
+    # fire for this branch.  ResultMessage arrives next.
```
majdyz: 🤖 🟡 **Nice to Have**: The test comment says "we skip emitting a ThinkingBlock to the frontend, so no Text/Tool events fire for this branch" — that's now stale after 95452d1 makes thinking blocks stream as reasoning events. In real SDK traffic, the final thinking-only turn would arrive as an `AssistantMessage` with a `ThinkingBlock`, which *would* open a new step (via the `if not self.step_open` branch in `AssistantMessage` handling) and emit reasoning events before `ResultMessage`. This test skips that AssistantMessage entirely, so it exercises a shape that the SDK doesn't actually produce. Consider adding an `AssistantMessage([ThinkingBlock(...)])` call between the `UserMessage(tool_result)` and the `ResultMessage` to match prod traffic — that also exercises the step-open path and would surface the protocol issue flagged on line 291.

autogpt_platform/frontend/src/app/(platform)/copilot/components/ChatMessagesContainer/helpers.ts
```diff
@@ -149,6 +149,12 @@ export function splitReasoningAndResponse(parts: MessagePart[]): {
   for (const part of rawReasoning) {
     if (isInteractiveToolPart(part)) {
       pinnedParts.push(part);
+    } else if (part.type === "reasoning" || part.type === "thinking") {
```
majdyz: 🤖 🟡 **Nice to Have**: For old sessions (pre-PR): no `reasoning` or `thinking` parts exist in UIMessage.parts on reload (see the dead-code comment on MessagePartRenderer line 109), so this branch never fires for historical data — no rendering regression. For live sessions on the new build: `reasoning` parts pin inline instead of being wrapped by `ReasoningCollapse`. That's the intended behaviour per the PR description, but it means a live-stream turn of shape `text → tool → reasoning → tool → text` will show the live `ReasoningCollapse` inline above the outer preamble-trigger block, which could look visually duplicated to users who see both. Would be worth a quick screenshot or Storybook story (the styleguide at `copilot/styleguide/page.tsx` already has ReasoningCollapse variants — adding one with reasoning-inside-preamble would cover this) to verify the nested-trigger claim in the comment actually holds when both triggers are present.

autogpt_platform/backend/backend/copilot/response_model.py
majdyz: 🤖 Fixed in [419037cdb6](https://github.com/Significant-Gravitas/AutoGPT/commit/419037cdb6): added `REASONING_START` / `REASONING_DELTA` / `REASONING_END` entries to `_reconstruct_chunk`'s type_to_class map.

autogpt_platform/frontend/src/app/(platform)/copilot/components/ChatMessagesContainer/components/MessagePartRenderer.tsx
majdyz: 🤖 Fixed in [938ca8d445](https://github.com/Significant-Gravitas/AutoGPT/commit/938ca8d445): removed the dead `case 'thinking'` branch. Persistence now flows through `role='reasoning'` DB rows → `convertChatSessionMessagesToUiMessages` → `{type: 'reasoning'}` parts, so the reasoning case handles both live and reloaded.

autogpt_platform/backend/backend/copilot/sdk/response_adapter.py
majdyz: 🤖 Fixed in [938ca8d445](https://github.com/Significant-Gravitas/AutoGPT/commit/938ca8d445): synthesis now opens a fresh step (`StreamStartStep`) before `_ensure_text_started` since the preceding `UserMessage` closed the prior one.

autogpt_platform/backend/backend/copilot/sdk/response_adapter_test.py
majdyz: 🤖 Fixed in [ea87ef9e2e7e0a4e55b5edbffdbf3e5ea8b3cb5c](https://github.com/Significant-Gravitas/AutoGPT/commit/ea87ef9e2e7e0a4e55b5edbffdbf3e5ea8b3cb5c): updated the stale comment to explicitly describe the degenerate no-AssistantMessage case this test is simulating now that real thinking blocks stream as reasoning events.

autogpt_platform/frontend/src/app/(platform)/copilot/components/ChatMessagesContainer/helpers.ts
majdyz: 🤖 Fixed in [352db78946](https://github.com/Significant-Gravitas/AutoGPT/commit/352db78946): persistence for new sessions now lives in `role='reasoning'` ChatMessage rows. Old pre-PR sessions still have no reasoning rows — they render as clean no-ops (no regression, no error). Acceptable tradeoff: no backfill, new behavior going forward.

autogpt_platform/backend/backend/copilot/sdk/response_adapter.py
```diff
@@ -210,16 +250,47 @@ def convert_message(self, sdk_message: Message) -> list[StreamBaseResponse]:
                     resolved_in_blocks.add(parent_id)
 
             self.resolved_tool_calls.update(resolved_in_blocks)
+            if resolved_in_blocks:
+                # A new tool_result just landed — reset the
+                # "has the model emitted text since the last tool result?"
+                # tracker so the thinking-only-final-turn guard at
+                # ``ResultMessage`` time stays accurate.
+                self._text_since_last_tool_result = False
+                self._any_tool_results_seen = True
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [ea87ef9e2e7e0a4e55b5edbffdbf3e5ea8b3cb5c](https://github.com/Significant-Gravitas/AutoGPT/commit/ea87ef9e2e7e0a4e55b5edbffdbf3e5ea8b3cb5c): `_flush_unresolved_tool_calls` now mirrors the UserMessage path — sets `_any_tool_results_seen = True` and resets `_text_since_last_tool_result` after a successful flush. Existing flush test updated to expect the synthesized closing text.

autogpt_platform/backend/backend/copilot/transcript.py
```diff
@@ -877,6 +877,14 @@ def extract_context_messages(
     """
     from .model import ChatMessage as _ChatMessage  # runtime import
 
+    # ``role="reasoning"`` rows are persisted for frontend replay of
+    # extended_thinking content but are NOT conversation context — the
+    # transcript-based --resume path already carries thinking separately,
+    # and sending them back to the model as user/assistant turns would be
+    # both redundant and malformed.  Drop them before any gap detection
+    # or transcript comparison so ordering invariants still hold.
+    session_messages = [m for m in session_messages if m.role != "reasoning"]
```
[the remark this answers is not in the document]
majdyz: 🤖 For the record, the real fix is in [af600139f6](https://github.com/Significant-Gravitas/AutoGPT/commit/af600139f6) — `_jsonl_covered` now counts non-reasoning rows only, since the CLI JSONL embeds extended_thinking inside assistant entries instead of surfacing reasoning as standalone rows. sentry's earlier self-resolution referenced an unrelated skill commit.

autogpt_platform/frontend/src/app/(platform)/copilot/helpers/convertChatSessionToUiMessages.ts
```diff
@@ -275,11 +293,11 @@ export function convertChatSessionMessagesToUiMessages(
     const msgId = `${sessionId}-seq-${msg.sequence}`;
     uiMessages.push({
       id: msgId,
-      role: msg.role,
+      role: uiRole,
```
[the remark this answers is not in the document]
majdyz: 🤖 Partially valid — added defensive fallback in [0067ed0bc5](https://github.com/Significant-Gravitas/AutoGPT/commit/0067ed0bc5). The premise that `role='reasoning'` rows reach this converter during live streaming isn't accurate: the converter is only called from DB-read paths (`useChatSession` mount, `useLoadMoreMessages`), and `_save_session_to_db` back-fills `sequence` on in-memory rows so DB-loaded messages always carry a non-null sequence. Live SSE goes through AI SDK's `useChat` directly, not the converter. Still, to harden against any future code path that forwards sequence-less messages, I added a loop-index fallback so duplicate `{sessionId}-seq-null` keys can't occur.

autogpt_platform/backend/backend/copilot/transcript.py
```diff
@@ -877,6 +877,14 @@ def extract_context_messages(
...
+
     prior = session_messages[:-1]
 
     if download is None:
```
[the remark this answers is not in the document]
majdyz: 🤖 False positive. sentry's prompt references a variable `transcript_msg_count` that doesn't exist in this code, and claims `prior` includes reasoning rows — that's incorrect. Look at lines 886-888: `session_messages = [m for m in session_messages if m.role != "reasoning"]` runs **before** `prior = session_messages[:-1]`, so `prior` is reasoning-free. The watermark comparison inside `detect_gap` is consistent with this after [af600139f6](https://github.com/Significant-Gravitas/AutoGPT/commit/af600139f6), which makes `_jsonl_covered` also exclude reasoning. Both sides agree — no index mismatch.

autogpt_platform/backend/backend/copilot/sdk/service.py
```diff
@@ -1671,7 +1689,20 @@ def _dispatch_response(
...
+                response.delta or ""
+            )
+
+    elif isinstance(response, StreamReasoningEnd):
+        acc.reasoning_response = None
+
+    elif isinstance(response, StreamTextDelta):
         raw_delta = response.delta or ""
         if skip_strip:
             # Pre-stripped tail from ThinkingStripper.flush() — bypass process()
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [1e48bdd001](https://github.com/Significant-Gravitas/AutoGPT/commit/1e48bdd001): `_compress_messages` now strips `role='reasoning'` rows centrally (alongside the existing `filter_compaction_messages` call). That covers `_build_query_message` (which already filtered upstream) plus `_seed_transcript` and any future caller. GraphQL still rate-limited so thread resolution will follow once the quota resets.

autogpt_platform/backend/backend/copilot/sdk/response_adapter.py
```diff
@@ -210,16 +250,54 @@ def convert_message(self, sdk_message: Message) -> list[StreamBaseResponse]:
...
+                    self.step_open = True
+                self._ensure_text_started(responses)
+                responses.append(
+                    StreamTextDelta(
+                        id=self.text_block_id,
+                        delta="(Done — no further commentary.)",
+                    )
+                )
             self._end_text_if_open(responses)
+            self._end_reasoning_if_open(responses)
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [aa06b90506](https://github.com/Significant-Gravitas/AutoGPT/commit/aa06b90506): added `_end_reasoning_if_open` before `_ensure_text_started` in the thinking-only fallback, so reasoning and text blocks stay strictly sequential on the wire. Thread resolution queued for the GraphQL rate-limit reset.

PR #12855, opened 2026-04-20
fix(backend/copilot): shorten RabbitMQ heartbeat + TCP keepalive to prevent zombie consumer

Swiftyos:
### Why / What / How

**Why:** The CoPilot executor's RabbitMQ consumer hung locally with a session stuck in `running` for ~52 minutes. Investigation found all four sockets from the executor parent process to RabbitMQ in `CLOSE_WAIT`, `copilot_execution_queue` sitting at 3 messages with 0 consumers, and `_stream_listener` heartbeating forever because nothing ever flipped `status`. Pika's blocking `start_consuming()` never saw the server FIN (IO thread starved — likely laptop sleep), so `@continuous_retry` never fired and the socket rotted half-closed until the next restart.

**What:** Let each `RabbitMQConfig` set a custom AMQP `heartbeat` and opt into kernel-level TCP keepalive, and wire aggressive values into `create_copilot_queue_config()` so the copilot consumer notices a dead peer within ~2 minutes instead of hours.

**How:**
- `RabbitMQConfig` gains `heartbeat: int = 300` (unchanged default) and `tcp_keepalive: bool = False` (off by default) — existing configs are untouched.
- `SyncRabbitMQ.connect()` reads both; when `tcp_keepalive=True` it passes a platform-aware `tcp_options` dict to pika (Linux uses `TCP_KEEPIDLE`, macOS uses `TCP_KEEPALIVE`, both get `TCP_KEEPINTVL=20`, `TCP_KEEPCNT=3`). `AsyncRabbitMQ.connect()` reads `heartbeat` too; aio_pika doesn't expose the keepalive knobs, so that side stays heartbeat-only.
- `create_copilot_queue_config()` overrides `heartbeat=60, tcp_keepalive=True`. The copilot consumer is the one that blocks in `start_consuming()` for the full process lifetime, so it's the one that benefits.

Graph executor, notifications, and other `RabbitMQConfig` users keep the old 300 s heartbeat with no keepalive — scope is intentionally narrow.

### Changes 🏗️

- `backend/data/rabbitmq.py`: `RabbitMQConfig.heartbeat` and `RabbitMQConfig.tcp_keepalive` fields; `_tcp_keepalive_options()` helper; both connect paths wired up.
- `backend/copilot/executor/utils.py`: `create_copilot_queue_config()` sets `heartbeat=60, tcp_keepalive=True` with a comment explaining the CLOSE_WAIT incident.

#### For code changes:
- [x] I have made a test plan
  - [x] `poetry run ruff check` clean on both files
  - [x] `poetry run black --check` clean on both files
  - [x] `poetry run pytest backend/copilot/executor/utils_test.py` — 11 passed
  - [ ] Restart the backend and confirm `copilot_execution_queue` has 1 consumer via `docker exec rabbitmq rabbitmqctl list_queues name messages consumers`
  - [ ] Simulate a silent drop (`docker restart rabbitmq` while the executor is idle) and confirm the consumer reconnects within ~2 minutes instead of hanging
  - [ ] Verify other queues (`graph_execution_queue`, `immediate_notifications`, etc.) are unaffected — their defaults haven't changed

### DIFF
--- a/autogpt_platform/backend/backend/copilot/prompting.py
+++ b/autogpt_platform/backend/backend/copilot/prompting.py
@@ -174,14 +174,18 @@ sandbox so `bash_exec` can access it for further processing.
 The exact sandbox path is shown in the `[Sandbox copy available at ...]` note.
 
 ### GitHub CLI (`gh`) and git
-- To check if the user has their GitHub account already connected, run `gh auth status`. Always check this before asking them to connect it.
+- To check if the user has their GitHub account already connected, run `gh auth status`. Always check this before running `connect_integration(provider="github")` which will ask the user to connect their GitHub regardless if it's already connected.
 - If the user has connected their GitHub account, both `gh` and `git` are
   pre-authenticated — use them directly without any manual login step.
   `git` HTTPS operations (clone, push, pull) work automatically.
 - If the token changes mid-session (e.g. user reconnects with a new token),
   run `gh auth setup-git` to re-register the credential helper.
-- If `gh` or `git` fails with an authentication error (e.g. "authentication
-  required", "could not read Username", or exit code 128), call
+- **MANDATORY:** You MUST run `gh auth status` before EVER calling
+  `connect_integration(provider="github")`. If it shows `Logged in`,
+  proceed directly — no integration connection needed. Never skip this check.
+- If `gh auth status` shows NOT logged in, or `gh`/`git` fails with an
+  authentication error (e.g. "authentication required", "could not read
+  Username", or exit code 128), THEN call
   `connect_integration(provider="github")` to surface the GitHub credentials
   setup card so the user can connect their account. Once connected, retry
   the operation.
--- a/autogpt_platform/backend/backend/copilot/executor/utils.py
+++ b/autogpt_platform/backend/backend/copilot/executor/utils.py
@@ -124,6 +124,15 @@ def create_copilot_queue_config() -> RabbitMQConfig:
         vhost="/",
         exchanges=[COPILOT_EXECUTION_EXCHANGE, COPILOT_CANCEL_EXCHANGE],
         queues=[run_queue, cancel_queue],
+        # The consumer threads sit in pika's blocking ``start_consuming()`` for
+        # the full lifetime of the process. If the TCP connection is dropped
+        # (server restart, NAT timeout, laptop sleep) while pika's IO thread is
+        # starved, the socket rots in CLOSE_WAIT and no message is ever
+        # consumed — see zombie-consumer incident notes. A short heartbeat plus
+        # kernel-level TCP keepalive makes both the app and the OS notice a
+        # dead peer within a couple of minutes instead of hours.
+        heartbeat=60,
+        tcp_keepalive=True,
     )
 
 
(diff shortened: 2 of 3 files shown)


autogpt_platform/backend/backend/copilot/prompting.py
```diff
@@ -180,14 +174,18 @@
 The exact sandbox path is shown in the `[Sandbox copy available at ...]` note.
 
 ### GitHub CLI (`gh`) and git
-- To check if the user has their GitHub account already connected, run `gh auth status`. Always check this before asking them to connect it.
+- To check if the user has their GitHub account already connected, run `gh auth status`. Always check this before running `connect_integration(provider="github")` which will ask the user to connect their GitHub regardless if it's already connected.
```
Swiftyos, line 177: I made the PR based of master branch so this change is already in master.

majdyz: try pull master to dev and push first,
otherwise will the conflict resolve will always be there because the same change is applied on two different squashed commits

majdyz (commented):
Let's make sure we pull master to dev first before merging this pr

majdyz:
Question on the `heartbeat=60` value choice.

### Heartbeat semantics

The PR description says the server drops the connection "within ~2x this interval" but I think that's a misread of pika/RabbitMQ semantics — per the [RabbitMQ heartbeats docs](https://www.rabbitmq.com/docs/heartbeats), the `heartbeat` parameter is the **timeout** value. Heartbeat frames are sent at `N/2` intervals, and the connection is declared dead after ~N seconds of silence (= 2 missed heartbeats × N/2 interval). So:

- `heartbeat=300` → heartbeats every 150s, dead detection ~300s (matches the original comment on the removed line)
- `heartbeat=60` → heartbeats every 30s, dead detection **~60s**, not ~120s

That's more aggressive than the description implies. At 60s, any pika IO-thread stall longer than ~60s trips a false-positive disconnect. Realistic causes in prod:

- k8s CPU throttling under load
- GIL contention from the threadpool running copilot turns
- NAT / LB failover blips
- Noisy-neighbor pods on the same node

Since the kernel TCP keepalive settings already give a hard detection floor (`TCP_KEEPIDLE=60 + 3×TCP_KEEPINTVL=20 = 120s`) independent of AMQP, would `heartbeat=90` or `heartbeat=120` work instead? That would:

- Still deliver the ~2 min detection the description promises
- Still give a 2.5-3× improvement over the 300s default
- Leave headroom for normal operational jitter

Not a blocker — just wondering if 60 was a considered choice vs an "as aggressive as seems reasonable" choice.

### No correctness risk — just operational churn

To be clear: the Redis `ClusterLock` (TTL 300s, refreshed every 30s from the threadpool worker independently of pika's IO thread) guards against double execution even when heartbeat false-positives cause message redelivery. The lock is the source of truth for session ownership; AMQP redeliveries during a heartbeat trip hit the lock check in `_handle_run_message` and get rejected with `requeue=False`. Even in a "pod fully frozen" scenario, the lock TTL (300s) is 5× longer than the heartbeat timeout (60s), so redeliveries during the false-positive window always find Pod A still owning the lock and get rejected before the lock has a chance to expire.

So the concern here is strictly operational (reconnect churn, log noise, prefetched-message redistribution under transient stalls) — not correctness.

PR #12866, opened 2026-04-21
fix(backend/copilot): reduce SDK + baseline prompt cache waste

majdyz:
## Summary

Four cost-reduction changes for the copilot feature.  Consolidated into one PR at user request; each commit is self-contained and bisectable.

### 1. SDK: full cross-user cache on every turn (CLI 2.1.116 bump)
Previous behavior: CLI 2.1.97 crashed when `excludeDynamicSections=True` was combined with `--resume`, so the code fell back to a raw `system_prompt` string on resume, losing Claude Code's default prompt and all cache markers.  Every Turn 2+ of an SDK session wrote ~33K tokens to cache instead of reading.

Fix: install `@anthropic-ai/claude-code@2.1.116` in the backend Docker image and point the SDK at it via `CHAT_CLAUDE_AGENT_CLI_PATH=/usr/bin/claude`.  CLI 2.1.98+ fixes the crash, so we can use the preset with `exclude_dynamic_sections=True` on every turn — Turn 1, 2, 3+ all share the same static prefix and hit the **cross-user** prompt cache.

**Local dev requirement:** if `CHAT_CLAUDE_AGENT_CLI_PATH` is unset, the bundled 2.1.97 fallback will crash on `--resume`.  Install the CLI globally (`npm install -g @anthropic-ai/claude-code@2.1.116`) or set the env var.

### 2. Baseline: add `cache_control` markers (commit `756b3ecd9` + follow-ups)
Baseline path had zero `cache_control` across `backend/copilot/**`.  Every turn was full uncached input (~18.6K tokens, ~$0.058).  Two ephemeral markers — on the system message (content-blocks form) and the last tool schema — plus `anthropic-beta: prompt-caching-2024-07-31` via `extra_headers` as defense-in-depth.  Helpers split into `_mark_tools_*` (precomputed once per session) and `_mark_system_*` (per-round, O(1)).  Repeat hellos: ~$0.058 → ~$0.006.

### 3. Drop `get_baseline_supplement()` (commit `6e6c4d791`)
`_generate_tool_documentation()` emitted ~4.3K tokens of `(tool_name, description)` pairs that exactly duplicated the tools array already in the same request.  Deleted.  `SHARED_TOOL_NOTES` (cross-tool workflow rules) is preserved.  Baseline "hello" input: ~18.7K → ~14.4K tokens.

### 4. Langfuse "CoPilot Prompt" v26 (published under `review` label)
Separate, out-of-repo change.  v25 had three duplicate "Example Response" blocks + a 10-step "Internal Reasoning Process" section.  v26 collapses to one example + bullet-form reasoning.  Char count 20,481 → 7,075 (rough 4 chars/token → ~5,100 → ~1,770 tokens).

- v26 is published with label `review` (NOT `production`); v25 remains active.
- Promote via `mcp__langfuse__updatePromptLabels(name="CoPilot Prompt", version=26, newLabels=["production"])` after smoke-test.
- Rollback: relabel v25 `production`.

## Test plan
- [x] Unit tests for `_build_system_prompt_value` (fresh vs resumed turns emit identical preset dict)
- [x] SDK compat tests pass including `test_bundled_cli_version_is_known_good_against_openrouter`
- [x] `cli_openrouter_compat_test.py` passes against CLI 2.1.116 (locally verified with `CHAT_CLAUDE_AGENT_CLI_PATH=/opt/homebrew/bin/claude`)
- [x] 8 new `_mark_*` unit tests + identity regression test for `_fresh_*` helpers
- [x] `SHARED_TOOL_NOTES` public-constant test passes; 5 old tool-docs tests removed
- [ ] **Manual cost verification (commit 1):** send two consecutive SDK turns; Turn 2 and Turn 3 should both show `cacheReadTokens` ≈ 33K (full cross-user cache hits).
- [ ] **Manual cost verification (commit 2):** send two "hello" turns on baseline <5 min apart; Turn 2 reports `cacheReadTokens` ≈ 18K and cost ≈ $0.006.
- [ ] **Regression sweep for commit 3:** one turn per tool family — `search_agents`, `run_agent`, `add_memory`/`forget_memory`/`search_memory`, `search_docs`, `read_workspace_file` — to verify no tool-selection regression from dropping the prose tool docs.
- [ ] **Langfuse v26 smoke test:** 5-10 varied turns after relabelling to `production`; compare responses vs v25 for regression on persona, concision, capability-gap handling, credential security flows.

## Deployment notes
- Production Docker image now installs CLI 2.1.116 (~20 MB added).
- `CHAT_CLAUDE_AGENT_CLI_PATH=/usr/bin/claude` set in the Dockerfile; runtime can override via env.
- First deploy after this merge needs a fresh image rebuild to pick up the new CLI.


autogpt_platform/backend/backend/copilot/baseline/service.py
```diff
@@ -295,9 +337,10 @@ async def _baseline_llm_caller(
     response = None  # initialized before try so finally block can access it
     try:
         client = _get_openai_client()
-        typed_messages = cast(list[ChatCompletionMessageParam], messages)
-        if tools:
-            typed_tools = cast(list[ChatCompletionToolParam], tools)
+        cached_messages, cached_tools = _apply_prompt_cache_markers(messages, tools)
```
majdyz: 🤖 🔴 **Blocker**: Unverified assumption that OpenRouter passes `cache_control` through to Anthropic. OpenRouter historically required the `anthropic-beta: prompt-caching-2024-07-31` header (or equivalent `extra_body`/`extra_headers` on the `chat.completions.create` call). If not set, OpenRouter silently strips `cache_control` and Task 1's savings never materialize — cost stays at ~$0.058 per hello.

Please verify one of:
1. OpenRouter auto-enables caching for Anthropic routes on the current provider config (paste evidence — a request trace showing `cacheReadTokens` on turn 2 of a repeat hello),
2. Or add `extra_headers={'anthropic-beta': 'prompt-caching-2024-07-31'}` / equivalent `extra_body` wiring here.

Without one of these, the new unit test passes but the on-wire behavior is unchanged.

autogpt_platform/backend/backend/copilot/sdk/service.py
```diff
@@ -3050,15 +3067,20 @@ def _on_stderr(line: str) -> None:
...
+        # preserved (via --append-system-prompt) and the preset-level cache
+        # markers apply.  On --resume, omit excludeDynamicSections=True
+        # (CLI 2.1.97 exits code 1 when that flag is combined with --resume);
+        # the session-stable cwd keeps dynamic sections identical across
+        # turns, so Turn 2+ of the same session still cache-reads the prefix
+        # instead of re-writing the full 33K tool schema + supplement.
+        # Falling back to a raw string here would pass --system-prompt
+        # instead, replacing the built-in prompt entirely and losing all
+        # CLI cache markers.
         system_prompt_value = _build_system_prompt_value(
```
majdyz, line 3056: 🤖 🟠 **Should Fix**: Comment is slightly misleading. With Turn 1 using `exclude_dynamic_sections=True` and Turn 2+ including dynamic sections, the two prefixes differ — so **Turn 2 still writes** a fresh cache entry (the dynamic-section variant), and only **Turn 3+** cache-reads. One extra write per session, not per turn. Recommend tightening the comment so future readers don't expect Turn 2 to be cheap.

Suggested wording: "…so Turn 3+ of the same session cache-reads the prefix. Turn 2 writes a new prefix (different from Turn 1's exclude-dynamic-sections variant), but that's a one-time-per-session cost instead of every-turn."

autogpt_platform/backend/backend/copilot/baseline/service.py
```diff
@@ -276,6 +276,48 @@ class _BaselineStreamState:
     _flushed_assistant_text_len: int = 0
 
 
+_EPHEMERAL_CACHE_CONTROL = {"type": "ephemeral"}
```
majdyz: 🤖 🟡 **Nice to Have**: `_EPHEMERAL_CACHE_CONTROL` is a shared mutable `dict`. It's handed out by reference on every call (via the literal-dict composition below and the `**spread`-equivalent in lines 307/316). A stray mutation anywhere downstream (e.g. the OpenAI SDK normalising fields) would poison every future request. Two tidier options: (a) make it a `types.MappingProxyType` to make mutation fail loudly, or (b) construct a fresh dict inline on each call since this is O(1). Non-blocking but cheap insurance.

autogpt_platform/backend/backend/copilot/sdk/service.py
```diff
@@ -848,27 +848,44 @@ def _is_fallback_stderr(line: str) -> bool:
 def _build_system_prompt_value(
     system_prompt: str,
     cross_user_cache: bool,
+    include_dynamic_sections: bool = False,
```
majdyz: 🤖 🔵 **Nit**: `include_dynamic_sections: bool = False` is positional, so a caller that previously wrote `_build_system_prompt_value(prompt, True)` (cross_user_cache=True) now has a second bool that reads identically but flips meaning. Consider marking it keyword-only: `def _build_system_prompt_value(system_prompt: str, *, cross_user_cache: bool, include_dynamic_sections: bool = False)`. Prevents positional misuse at near-zero cost.

autogpt_platform/backend/backend/copilot/prompting.py
```diff
@@ -440,15 +412,7 @@ def get_graphiti_supplement() -> str:
...
-
-    Returns:
-        The supplement string to append to the system prompt
-    """
-    tool_docs = _generate_tool_documentation()
-    return tool_docs + _SHARED_TOOL_NOTES
+# Public alias — baseline/service.py appends these workflow notes to the
+# system prompt.  Kept as a module-level constant (not a function) so repeat
+# reads don't recompute anything; the content is static.
+SHARED_TOOL_NOTES = _SHARED_TOOL_NOTES
```
majdyz: 🤖 🟡 **Nice to Have**: The `SHARED_TOOL_NOTES = _SHARED_TOOL_NOTES` aliasing dance feels like transitional scaffolding. No external caller currently imports `_SHARED_TOOL_NOTES` (grep confirms), so just rename the constant in place — drop the leading underscore and delete the alias. Keeps one name per concept.

majdyz:
🤖 🟠 **Should Fix (PR description)**: The body still marks Tasks 2, 3, and 4 as **pending**, but all three commits have landed (`756b3ecd9`, `6e6c4d791`) and Langfuse v26 is published under the `review` label. Please update the description so reviewers don't think those sections are TBD. Also add:

- Confirmation that v26 has been published (and the manual smoke-test status) for Task 3.
- A line in the test plan about the Task 2 regression risk (tool-selection sweep) — the followup doc explicitly flagged it as the highest-risk change in this batch.

autogpt_platform/backend/backend/copilot/baseline/service.py
majdyz: 🤖 Fixed in [13bab7d03](https://github.com/Significant-Gravitas/AutoGPT/commit/13bab7d03405f28a277c04383d6aa9a1975aa448): added `extra_headers={"anthropic-beta": "prompt-caching-2024-07-31"}` on both `chat.completions.create` calls as defense-in-depth. OpenRouter auto-forwards `cache_control` for Anthropic routes today, so this is belt-and-braces: no-op if already enabled, and guards against a silent degradation if OpenRouter's routing config changes. Non-Anthropic providers drop unknown headers.

autogpt_platform/backend/backend/copilot/sdk/service.py
majdyz: 🤖 Fixed in [13bab7d03](https://github.com/Significant-Gravitas/AutoGPT/commit/13bab7d03405f28a277c04383d6aa9a1975aa448): rewrote the comment to explicitly call out that Turn 2 writes a new prefix (different from Turn 1's) and Turn 3+ are the ones that cache-read. One extra write per session, not per turn.

autogpt_platform/backend/backend/copilot/baseline/service.py
majdyz: 🤖 Fixed in [13bab7d03](https://github.com/Significant-Gravitas/AutoGPT/commit/13bab7d03405f28a277c04383d6aa9a1975aa448): replaced the shared `_EPHEMERAL_CACHE_CONTROL` dict with `_fresh_ephemeral_cache_control()` that returns a new dict each call. O(1) construction vs the downside of a module-level mutation poisoning all future requests.

autogpt_platform/backend/backend/copilot/sdk/service.py
majdyz: 🤖 Fixed in [13bab7d03](https://github.com/Significant-Gravitas/AutoGPT/commit/13bab7d03405f28a277c04383d6aa9a1975aa448): made `cross_user_cache` and `include_dynamic_sections` keyword-only via `*`. Existing callers already pass them as kwargs so no call-site changes needed.

autogpt_platform/backend/backend/copilot/prompting.py
majdyz: 🤖 Fixed in [13bab7d03](https://github.com/Significant-Gravitas/AutoGPT/commit/13bab7d03405f28a277c04383d6aa9a1975aa448): renamed `_SHARED_TOOL_NOTES` → `SHARED_TOOL_NOTES` in place and deleted the alias line. One name, one concept.

majdyz:
🤖 Fixed in [13bab7d03](https://github.com/Significant-Gravitas/AutoGPT/commit/13bab7d03405f28a277c04383d6aa9a1975aa448): PR description rewritten. Removed all `pending` labels; added commit SHAs, v26 Langfuse publishing details + promotion/rollback instructions; explicit test-plan checkboxes for the Task 2 tool-selection regression sweep and the v26 smoke test.

autogpt_platform/backend/backend/copilot/baseline/service.py
```diff
@@ -276,6 +276,64 @@ class _BaselineStreamState:
     _flushed_assistant_text_len: int = 0
 
 
+# Defense-in-depth: request the Anthropic prompt-caching beta explicitly.
+# OpenRouter auto-forwards cache_control for Anthropic routes without this
+# header, but passing it makes the intent unambiguous on-wire and is a no-op
+# for non-Anthropic providers (OpenRouter drops unknown headers).
+_ANTHROPIC_CACHING_HEADERS = {"anthropic-beta": "prompt-caching-2024-07-31"}
```
majdyz: 🤖 🟠 **Should Fix**: Same shared-mutable-dict concern that applied to `_EPHEMERAL_CACHE_CONTROL` applies to `_ANTHROPIC_CACHING_HEADERS` too — it's a module-level dict passed by reference into `extra_headers=` on every request. The OpenAI Python SDK generally merges `extra_headers` into the underlying httpx client, but if any layer ever normalises/mutates in place (seen in older httpx versions for `Content-Length`), the module-level copy is corrupted and every subsequent request ships the wrong header.

Apply the same pattern you used for cache_control: a `_fresh_anthropic_caching_headers()` helper that returns a new dict each call, or replace with `types.MappingProxyType({...})`.

autogpt_platform/backend/backend/copilot/baseline/service.py
```diff
@@ -276,6 +276,64 @@ class _BaselineStreamState:
...
+            cached_messages[0] = {
+                "role": "system",
+                "content": [
+                    {
+                        "type": "text",
+                        "text": sys_content,
+                        "cache_control": _fresh_ephemeral_cache_control(),
+                    }
+                ],
+            }
```
majdyz, line 466: 🤖 🟠 **Should Fix**: This rebuild of `cached_messages[0]` drops any other keys the original dict carried. OpenAI's `ChatCompletionMessageParam` supports a `name` field on system messages (used for developer/assistant attribution), plus the SDK occasionally threads through per-message metadata. Today the callers don't set these, but a future caller that does will silently lose the field here.

Safer pattern — spread the original so unknown keys survive:

```python
cached_messages[0] = {
    **cached_messages[0],
    "content": [
        {
            "type": "text",
            "text": sys_content,
            "cache_control": _fresh_ephemeral_cache_control(),
        }
    ],
}
```

autogpt_platform/backend/backend/copilot/baseline/service.py
```diff
@@ -276,6 +276,64 @@ class _BaselineStreamState:
...
+
+    cached_tools: list[dict[str, Any]] = [dict(t) for t in tools]
+    if cached_tools:
+        cached_tools[-1] = {
+            **cached_tools[-1],
+            "cache_control": _fresh_ephemeral_cache_control(),
+        }
+    return cached_messages, cached_tools
```
majdyz: 🤖 🟡 **Nice to Have**: `_apply_prompt_cache_markers` is called from `_baseline_llm_caller` which runs once per LLM round in the tool-call loop. For ~43 tools that's ~43 fresh dicts allocated per round for the tool deep-copy, plus the messages-list copy. None of this changes within a session. Consider precomputing the cached-tools list once at session start (in `stream_chat_completion_baseline` where `tools = get_available_tools()` lives) and only applying the system-message marker per round. Won't move the cost needle but removes a tight allocation from a hot path.

autogpt_platform/backend/backend/copilot/baseline/service.py
majdyz: 🤖 Fixed in [d4528aea1](https://github.com/Significant-Gravitas/AutoGPT/commit/d4528aea1484a723afda340027e0ccfda102160d): replaced module-level `_ANTHROPIC_CACHING_HEADERS` with `_fresh_anthropic_caching_headers()` — same fresh-dict pattern as `_fresh_ephemeral_cache_control`. No module-level mutable dict to poison.

autogpt_platform/backend/backend/copilot/baseline/service.py
majdyz: 🤖 Fixed in [d4528aea1](https://github.com/Significant-Gravitas/AutoGPT/commit/d4528aea1484a723afda340027e0ccfda102160d): switched to `{**cached_messages[0], "content": [...]}` spread so any unknown fields (name, metadata) on the original system message survive. Added a test covering the `name` field case.

autogpt_platform/backend/backend/copilot/baseline/service.py
majdyz: 🤖 Fixed in [d4528aea1](https://github.com/Significant-Gravitas/AutoGPT/commit/d4528aea1484a723afda340027e0ccfda102160d): split the helper into `_mark_tools_with_cache_control` (called once at session start in `stream_chat_completion_baseline`) and `_mark_system_message_with_cache_control` (per-round, O(1)). The ~43 tool-dict copies no longer happen every LLM round.

autogpt_platform/backend/backend/copilot/baseline/service.py
```diff
@@ -276,6 +276,80 @@ class _BaselineStreamState:
...
 
+def _fresh_ephemeral_cache_control() -> dict[str, str]:
+    """Return a FRESH ``{"type": "ephemeral"}`` dict each call.
+
+    Using a shared module-level dict would let any downstream mutation
+    (e.g. the OpenAI SDK normalising fields in-place) poison every future
+    request's cache_control marker.  Construction is O(1) so the safety
+    margin is essentially free.
+    """
+    return {"type": "ephemeral"}
```
majdyz: 🤖 🟡 **Nice to Have**: The `_fresh_*` naming advertises an identity guarantee (each call returns a NEW dict) — that's the whole reason they exist. Worth pinning that invariant with a tiny regression test, otherwise someone could refactor to return a module-level constant in a year and reintroduce the exact bug loop 1 flagged.

```python
def test_fresh_helpers_return_distinct_objects():
    from backend.copilot.baseline.service import (
        _fresh_ephemeral_cache_control,
        _fresh_anthropic_caching_headers,
    )
    assert _fresh_ephemeral_cache_control() is not _fresh_ephemeral_cache_control()
    assert _fresh_anthropic_caching_headers() is not _fresh_anthropic_caching_headers()
```

autogpt_platform/backend/backend/copilot/baseline/service.py
```diff
@@ -276,6 +276,80 @@ class _BaselineStreamState:
...
+    Same reasoning as :func:`_fresh_ephemeral_cache_control`: never hand a
+    shared module-level dict to third-party SDKs.  OpenRouter auto-forwards
+    cache_control for Anthropic routes without this header, but passing it
+    makes the intent unambiguous on-wire and is a no-op for non-Anthropic
+    providers (unknown headers are dropped).
+    """
+    return {"anthropic-beta": "prompt-caching-2024-07-31"}
+
+
+def _mark_tools_with_cache_control(tools: Sequence[Any]) -> list[dict[str, Any]]:
```
majdyz: 🤖 🔵 **Nit**: `tools: Sequence[Any]` is looser than needed — we call `dict(t)` on each item, which only works for mapping-shaped inputs. `Sequence[Mapping[str, Any]]` (from `collections.abc`) documents that contract explicitly and lets the type-checker catch a caller that passes `Sequence[str]` by mistake. Same treatment would fit `messages: list[dict[str, Any]]` → `Sequence[Mapping[str, Any]]` on the other helper for symmetry.

autogpt_platform/backend/backend/copilot/baseline/service.py
majdyz: 🤖 Fixed in [5640ac255](https://github.com/Significant-Gravitas/AutoGPT/commit/5640ac255bb3b345a79b343724f1e71d7bcf3989): added `test_fresh_helpers_return_distinct_objects` as an identity regression guard for both `_fresh_*` helpers.

autogpt_platform/backend/backend/copilot/baseline/service.py
majdyz: 🤖 Fixed in [5640ac255](https://github.com/Significant-Gravitas/AutoGPT/commit/5640ac255bb3b345a79b343724f1e71d7bcf3989): tightened both helpers to `Sequence[Mapping[str, Any]]` and switched the messages copy to `[dict(m) for m in messages]` so the return type stays `list[dict[str, Any]]`.

majdyz:
## 🧪 E2E Test Report

Ran against `d9a2cef9a` on native stack (infra via docker compose, backend/frontend via `poetry run app` + `pnpm run dev`), CLI 2.1.116 at `/opt/homebrew/bin/claude` via `CHAT_CLAUDE_AGENT_CLI_PATH`.

### Scenario 1 — SDK path: cross-user cache on resumed turn ✅

| Turn | promptTokens | cacheRead | cacheCreation |
|---|---|---|---|
| 1 (`hello`) | 3 | 0 | **32,221** |
| 2 (`what can you do?`) | 3 | **32,221** | 179 |

Turn 2 read the entire 32,221-token static prefix from cache — exactly what the CLI bump was meant to unlock. Previously every resumed turn either crashed (with the `excludeDynamicSections` flag) or skipped cross-user caching (via the workaround that passed session-stable but per-user dynamic sections). ~10× savings on every turn after the first.

### Scenario 2 — Baseline path: cache_control + anthropic-beta header ✅

| Turn | promptTokens | cacheRead | cacheCreation |
|---|---|---|---|
| 1 (`hello`) | 26,408 | 0 | 0 |
| 2 (`hello again`) | **1,231** | 0 | 0 |

`promptTokens` dropped **26,408 → 1,231** (~95%), confirming OpenRouter is forwarding the ephemeral markers to Anthropic and cache-reading on the repeat request. The `cacheReadTokens`/`cacheCreationTokens` fields staying at 0 is a pre-existing gap in `_baseline_llm_caller`'s usage parsing (OpenRouter's streaming format differs from OpenAI's) — the caching itself works, our tracking just doesn't surface the breakdown. Separate PR to fix.

### Summary
- **SDK path (Task 4 — CLI 2.1.116 bump):** ✅ verified end-to-end
- **Baseline path (Tasks 1 + 2 — cache_control + supplement drop):** ✅ ~95% prompt-token reduction on repeat
- **No regressions** in either path; both returned sensible responses

### Still on reviewer checklist
- Tool-selection regression sweep for Task 2 (single `hello` doesn't exercise tool-picking)
- Langfuse v26 smoke test + promote label to `production`
- Production docker image build (Dockerfile edits verified by code review only)

autogpt_platform/backend/backend/copilot/baseline/service.py
```diff
@@ -295,22 +371,29 @@ async def _baseline_llm_caller(
...
+                extra_headers=_fresh_anthropic_caching_headers(),
             )
         else:
             response = await client.chat.completions.create(
                 model=state.model,
                 messages=typed_messages,
                 stream=True,
                 stream_options={"include_usage": True},
+                extra_headers=_fresh_anthropic_caching_headers(),
             )
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [440f1b581](https://github.com/Significant-Gravitas/AutoGPT/commit/440f1b58168d85a53ec96d7d8d8f70ab04dbb89c): valid concern — added `_is_anthropic_model()` helper and gated all three Anthropic-specific things (tool cache_control at session precompute, system-message cache_control per round, `anthropic-beta` header) behind it. Non-Anthropic models (`openai/*`, `google/*`, `xai/*`, etc. set via `CHAT_MODEL`/`CHAT_ADVANCED_MODEL`) now get a clean request with zero Anthropic-specific fields. 10 unit tests cover the matcher for `claude-*`, `anthropic/*`, `anthropic.*` (Bedrock-style), and rejection of OpenAI/Grok/Gemini/Llama.

majdyz:
Date: 2026-04-21
PR: #12866 — fix(backend/copilot): reduce SDK + baseline prompt cache waste
Branch: `fix/copilot-cross-user-cache-on-resume`
Starting commit: `7ef879576` (dev merge — includes #12864 real-OpenRouter-cost tracking)
Fix commit: `9bdc88bf9` (baseline SSE cache tokens)
Worktree: `/Users/majdyz/Code/AutoGPT3`

## Environment
- **SDK:** `claude-agent-sdk 0.1.64` (bundled CLI 2.1.116 — the `--resume` + cross-user-cache fix)
- **Mode:** `CHAT_USE_CLAUDE_CODE_SUBSCRIPTION=false` fallback — macOS-keychain OAuth token returned 401; fallback to OpenRouter + existing `CHAT_API_KEY`
- **Baseline model:** `CHAT_MODEL` unset → defaults to Anthropic via OpenRouter
- **Test user:** `test@test.com` / `testtest123`

## Results

| Test | Scenario | Metric | Turn 1 | Turn 2 | Outcome |
|---|---|---|---|---|---|
| 1 | SDK `extended_thinking` resumed turn cross-user cache | `cacheReadTokens` | 0 | **32,192** | **PASS** — Turn 2 reads 32,192 tokens (> 30K target) |
| 1 | " | `cacheCreationTokens` | 32,192 | **167** | **PASS** — Turn 2 writes only conversation delta (< 500 target) |
| 1 | " | `promptTokens` (uncached) | 3 | 3 | Stable — static prefix fully cached |
| 2 | Baseline `fast` path — OpenRouter Anthropic cache | `promptTokens` | 26,410 | **1,254** | **PASS** — 95.3% drop vs Turn 1 (> 90% target) |
| 2 | " | `cacheReadTokens` on the wire | 0 | **25,544** | **PASS** after fix — SSE now surfaces what was already tracked internally |
| 2 | " | `cacheCreationTokens` | 0 | 0 | OpenRouter streaming usage doesn't populate this field (upstream); cache-creation cost still folded into `promptTokens` on Turn 1, so cost accounting remains correct |
| 3 | Non-Anthropic gate — `CHAT_MODEL=openai/gpt-4o-mini` | HTTP status | 200 OK | — | **PASS** — gpt-4o-mini returned a sensible welcome response |
| 3 | " | `cacheReadTokens` / `cacheCreationTokens` | 0 / 0 | — | **PASS** — markers correctly dropped for non-Anthropic, no 400 on unknown `cache_control` fields |
| 3 | " | `promptTokens` | 8,665 | — | Single "hello" turn against gpt-4o-mini |

## Bug found & fixed during the loop

**Symptom:** baseline Turn 2 reported `cacheReadTokens=0` on the wire even though
`[Baseline] Turn usage: uncached=1254, cache_read=25544, cache_create=0, …`
appeared in the backend log — so OpenRouter WAS returning
`prompt_tokens_details.cached_tokens` correctly, it just wasn't escaping onto the SSE event.

**Root cause:** `autogpt_platform/backend/backend/copilot/baseline/service.py:1834`
constructed `StreamUsage(prompt_tokens=…, completion_tokens=…, total_tokens=…)` without
passing `cache_read_tokens` / `cache_creation_tokens` (which default to 0 in the model).
The SDK path's `StreamUsage` emission at `sdk/service.py:3687` already passes both; baseline
was missing them. Pre-existing gap made visible now that PR #12864 has wired
`_extract_cache_creation_tokens` on the OpenRouter shape.

**Fix** (commit `9bdc88bf9`): pass the two fields through, matching the SDK path.

```py
yield StreamUsage(
    prompt_tokens=billed_prompt,
    completion_tokens=state.turn_completion_tokens,
    total_tokens=billed_prompt + state.turn_completion_tokens,
    cache_read_tokens=state.turn_cache_read_tokens,
    cache_creation_tokens=state.turn_cache_creation_tokens,
)
```

After the fix, a fresh baseline session picked up the 5-min TTL cache from the previous one,
so Turn 1 itself showed `cacheReadTokens=25,544` and Turn 2 showed the same — the wire
event now faithfully reflects the OpenRouter usage shape.

## Cost math

- **SDK path, 20-turn session:** baseline cost is 1 × 32K write + 19 × 32K read at $0.30/M
  (cache-read) vs $3/M (uncached). For the static prefix alone that's ~**10× savings per
  resumed turn**, which matches the PR's intent.
- **Baseline path, repeat within 5-min TTL:** Turn 1 billed ~26K + 251 out (cost $0.102153).
  Turn 2 billed ~1.25K + 169 out (cost $0.013960). The prompt-token drop is real on the wire;
  cost tracking was correct before the fix (it read `turn_cache_read_tokens` internally); the
  fix only makes the drop visible to the frontend/usage-event consumers.

## Summary
- **SDK cross-user cache on resumed turns:** PASS — 32,192-token cache READ on Turn 2, matches target
- **Baseline `cache_control` + OpenRouter cost tracking:** PASS — prompt-token 95% drop confirmed; `cacheReadTokens` now surfaces on SSE after the loop-2 fix
- **Non-Anthropic gate:** PASS — gpt-4o-mini handled cleanly, no 400 on `cache_control` stripping

No regressions in end-to-end chat behaviour (extended_thinking and fast both returned sensible responses).

autogpt_platform/backend/backend/copilot/baseline/service.py
```diff
@@ -347,28 +479,37 @@ async def _baseline_llm_caller(
...
         else:
-            response = await client.chat.completions.create(
-                model=state.model,
-                messages=typed_messages,
-                stream=True,
-                stream_options={"include_usage": True},
-                extra_body=_OPENROUTER_INCLUDE_USAGE_COST,
-            )
+            final_messages = messages
+            extra_headers = None
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [2af442362](https://github.com/Significant-Gravitas/AutoGPT/commit/2af4423629f535aae2a3cc913c4f34fa5a5a0c85): extracted the per-message transform into `_build_cached_system_message`, memoised on `_BaselineStreamState.cached_system_message` on the first LLM round, then spliced in as `[cached_system_msg, *messages[1:]]` on subsequent rounds. The N dict-copies-per-round collapsed to N list-slot references. Covered with a new async test that verifies the same dict instance is reused across rounds and the tail messages pass through by reference.

autogpt_platform/backend/backend/copilot/baseline/service.py
```diff
@@ -347,28 +531,51 @@ async def _baseline_llm_caller(
...
+        typed_messages = cast(list[ChatCompletionMessageParam], final_messages)
+        create_kwargs: dict[str, Any] = {
+            "model": state.model,
+            "messages": typed_messages,
+            "stream": True,
+            "stream_options": {"include_usage": True},
+            "extra_body": _fresh_openrouter_extra_body(),
```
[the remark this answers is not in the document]
majdyz: 🤖 Stale after [4bf44474b](https://github.com/Significant-Gravitas/AutoGPT/commit/4bf44474b3b5cdccc7f19f9130f51fe3a91290a0): reverted the `_fresh_openrouter_extra_body()` change.  `_OPENROUTER_INCLUDE_USAGE_COST` is restored to its original shape (`{"usage": {"include": True}}`) and `test_baseline_requests_usage_include_extra_body` passes locally on this commit.  Bot flagged the reverted change; no action needed.

PR #12867, opened 2026-04-21
fix(backend/copilot): release session cluster lock on completion

majdyz:
## Summary

Fixes a bug where a chat session gets silently stuck after the user presses Stop mid-turn.

**Root cause:** the cancel endpoint marks the session `failed` after polling 5s, but the cluster lock held by the still-running task is only released by `on_run_done` when the task actually finishes. If the task hangs past the 5s poll (slow LLM call, agent-browser step, etc.), the lock lingers for up to 5 min — `stream_chat_post`'s `is_turn_in_flight` check sees the flipped meta (`failed`) and enqueues a new turn, but the run handler sees the stale lock and drops the user's message at `manager.py:379` (`reject+requeue=False`). The new SSE stream hangs until its 60s idle timeout.

### Fix

Two cooperating changes:

1. **`mark_session_completed` force-releases the cluster lock** in the same transaction that flips status to `completed`/`failed`. Unconditional delete — by the time we're declaring the session dead, we don't care who the current lock holder is; the lock has to go so the next enqueued turn can acquire. This is what closes the stuck-session window.
2. **`ClusterLock.release()` is now owner-checked** (Lua CAS — `GET == token ? DEL : noop` atomically). Force-release means another pod may legitimately own the key by the time the original task's `on_run_done` eventually fires. Without the CAS, that late `release()` would wipe the successor's lock. With it, the late `release()` is a safe no-op when the owner has changed.

Together: prompt release on completion (via force-delete) + safe cleanup when on_run_done catches up (via CAS). That re-syncs the API-level `is_turn_in_flight` check with the actual lock state, so the contention window disappears.

No changes to the worker-level contention handler: `stream_chat_post` already queues incoming messages into the pending buffer when a turn is in flight (via `queue_pending_for_http`). With these fixes, the worker never sees contention in the common case; if it does (true multi-pod race), the pre-existing `reject+requeue=False` behaviour still applies — we'll revisit that path with its own PR if it becomes a production symptom.

### Verification

- Reproduced the original stuck-session symptom locally (Stop mid-turn → send new message → backend logs `Session … already running on pod …`, user message silently lost, SSE stream idle 60s then closes).
- After the fix: cancel → new message → turn starts normally (lock released by `mark_session_completed`).
- `poetry run pyright` — 0 errors on edited files.
- `pytest backend/copilot/stream_registry_test.py backend/executor/cluster_lock_test.py` — 33 passed (includes the successor-not-wiped test).

## Changes

- `autogpt_platform/backend/backend/copilot/executor/utils.py` — extract `get_session_lock_key(session_id)` helper so the lock-key format has a single source of truth.
- `autogpt_platform/backend/backend/copilot/executor/manager.py` — use the helper where the cluster lock is created.
- `autogpt_platform/backend/backend/copilot/stream_registry.py` — `mark_session_completed` deletes the lock key after the atomic status swap (force-release).
- `autogpt_platform/backend/backend/executor/cluster_lock.py` — `ClusterLock.release()` (sync + async) uses a Lua CAS to only delete when `GET == token`, protecting against wiping a successor after a force-release.

## Test plan

- [ ] Send a message in /copilot that triggers a long turn (e.g. `run_agent`), press Stop before it finishes, then send another message. Expect: new turn starts promptly (no 5-min wait for lock TTL).
- [ ] Happy path regression — send a normal message, verify turn completes and the session lock key is deleted after completion.
- [ ] Successor protection — unit test `test_release_does_not_wipe_successor_lock` covers: A acquires, external DEL, B acquires, A.release() is a no-op, B's lock intact.

### DIFF
--- a/autogpt_platform/backend/backend/copilot/executor/manager.py
+++ b/autogpt_platform/backend/backend/copilot/executor/manager.py
@@ -34,6 +34,7 @@ from .utils import (
     CancelCoPilotEvent,
     CoPilotExecutionEntry,
     create_copilot_queue_config,
+    get_session_lock_key,
 )
 
 logger = TruncatedLogger(logging.getLogger(__name__), prefix="[CoPilotExecutor]")
@@ -366,7 +367,7 @@ class CoPilotExecutor(AppProcess):
         # Try to acquire cluster-wide lock
         cluster_lock = ClusterLock(
             redis=redis.get_redis(),
-            key=f"copilot:session:{session_id}:lock",
+            key=get_session_lock_key(session_id),
             owner_id=self.executor_id,
             timeout=settings.config.cluster_lock_timeout,
         )
--- a/autogpt_platform/backend/backend/copilot/stream_registry.py
+++ b/autogpt_platform/backend/backend/copilot/stream_registry.py
@@ -35,7 +35,7 @@ from backend.data.redis_client import get_redis_async
 from backend.data.redis_helpers import hash_compare_and_set
 
 from .config import ChatConfig
-from .executor.utils import COPILOT_CONSUMER_TIMEOUT_SECONDS
+from .executor.utils import COPILOT_CONSUMER_TIMEOUT_SECONDS, get_session_lock_key
 from .response_model import (
     ResponseType,
     StreamBaseResponse,
@@ -850,6 +850,15 @@ async def mark_session_completed(
     if not swapped:
         logger.debug(f"Session {session_id} already completed/failed, skipping")
         return False
+
+    # Force-release the executor's cluster lock so the next enqueued turn can
+    # acquire it immediately. The lock holder's on_run_done will also release
+    # (idempotent delete); doing it here unblocks cases where the task hangs
+    # past the cancel timeout or a pod crash leaves the lock orphaned.
+    try:
+        await redis.delete(get_session_lock_key(session_id))
+    except RedisError as e:
+        logger.warning(f"Failed to release cluster lock for session {session_id}: {e}")
 
     if error_message and not skip_error_publish:
         try:
--- a/autogpt_platform/backend/backend/copilot/stream_registry_test.py
+++ b/autogpt_platform/backend/backend/copilot/stream_registry_test.py
@@ -280,9 +280,10 @@ async def test_mark_session_completed_releases_cluster_lock_on_success():
 @pytest.mark.asyncio
 async def test_mark_session_completed_skips_lock_release_when_already_completed():
     """CAS failure = someone else completed the session first; we must not
-    clobber their already-released lock (harmless but wastes a round trip)
-    or, more importantly, publish StreamFinish twice."""
+    delete their already-released lock, and we must NOT publish StreamFinish
+    twice (the winning caller already published it)."""
     fake_redis = _FakeRedis({"status": "completed", "turn_id": "turn-1"})
+    publish_mock = AsyncMock()
 
     with (
         patch.object(
@@ -291,11 +292,16 @@ async def test_mark_session_completed_skips_lock_release_when_already_completed(
         patch.object(
             stream_registry, "hash_compare_and_set", new=AsyncMock(return_value=False)
         ),
+        patch.object(stream_registry, "publish_chunk", new=publish_mock),
     ):
         result = await stream_registry.mark_session_completed("sess-1")
 
     assert result is False
     assert get_session_lock_key("sess-1") not in fake_redis.deleted_keys
+    assert not any(
+        isinstance(call.args[1], stream_registry.StreamFinish)
+        for call in publish_mock.call_args_list
+    ), "StreamFinish must NOT be re-published on the CAS-no-op branch"
 
 
 @pytest.mark.asyncio
--- a/autogpt_platform/backend/backend/copilot/stream_registry_test.py
+++ b/autogpt_platform/backend/backend/copilot/stream_registry_test.py
@@ -4,8 +4,10 @@ import asyncio
 from unittest.mock import AsyncMock, patch
 
 import pytest
+from redis.exceptions import RedisError
 
 from backend.copilot import stream_registry
+from backend.copilot.executor.utils import get_session_lock_key
 
 
 @pytest.fixture(autouse=True)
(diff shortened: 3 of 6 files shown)


autogpt_platform/backend/backend/copilot/executor/manager.py
```diff
@@ -435,6 +438,67 @@ def on_run_done(f: Future):
...
+                    )
+            await stream_registry.mark_session_completed(
+                entry.session_id,
+                error_message=error_text,
+                skip_error_publish=True,
+                release_lock=False,
+            )
+
+        try:
+            asyncio.run(_run())
```
majdyz: 🤖 🔴 **Blocker**: Using `asyncio.run()` here on each pika-consumer callback creates a fresh event loop per invocation, but `get_redis_async()` is `@thread_cached` — it caches the `AsyncRedis` client on thread-local storage. The AsyncRedis client binds its connection pool futures/locks to the loop it was first created on. On the second (and subsequent) contention invocation in the same consumer thread, `queue_user_message` → `push_pending_message` → `get_redis_async()` returns the cached client bound to the now-closed loop, raising `RuntimeError: got Future attached to a different loop`. The rest of this module avoids the problem by using a dedicated persistent loop + `asyncio.run_coroutine_threadsafe` (see `processor.py:160`, `:232`). Suggest the same pattern here, or schedule onto an existing loop run in a helper thread.

autogpt_platform/backend/backend/copilot/executor/manager.py
```diff
@@ -435,6 +438,67 @@ def on_run_done(f: Future):
...
+                    logger.warning(
+                        f"Failed to publish contention StreamError for "
+                        f"{entry.session_id}: {e}"
+                    )
+            await stream_registry.mark_session_completed(
+                entry.session_id,
+                error_message=error_text,
+                skip_error_publish=True,
+                release_lock=False,
+            )
```
majdyz: 🤖 🔴 **Blocker**: `mark_session_completed(session_id, error_message=..., skip_error_publish=True, release_lock=False)` on the contention path is far more invasive than it looks. Inside it: (1) reads `meta.turn_id` from Redis — but the HTTP handler that enqueued this phantom message already ran `create_session(session_id, turn_id=NEW_UUID)` which overwrote the meta hash, so `meta.turn_id` is the NEW phantom turn, not the OLD in-flight one; (2) CAS-swaps status `running`→`failed` on the shared meta — this flips the session state out from under the OLD pod that is still legitimately running its turn; (3) publishes `StreamFinish` to the NEW phantom turn_id. When the OLD pod eventually finishes its turn and calls `mark_session_completed`, the CAS fails (status already `failed`), so OLD pod publishes no `StreamFinish` and silently abandons its own session state + duration metrics + completion notification. Net effect: the phantom turn corrupts the legitimate in-flight turn rather than just reporting a clean failure on its own stream. Consider publishing only `StreamError`+`StreamFinish` directly to `entry.turn_id` (the phantom only) without touching the shared session meta.

autogpt_platform/backend/backend/copilot/stream_registry.py
```diff
@@ -851,6 +857,18 @@ async def mark_session_completed(
         logger.debug(f"Session {session_id} already completed/failed, skipping")
         return False
 
+    # Force-release the executor's cluster lock so the next enqueued turn can
+    # acquire it immediately. The lock holder's on_run_done will also release
+    # (idempotent delete); doing it here unblocks cases where the task hangs
+    # past the cancel timeout or a pod crash leaves the lock orphaned.
+    if release_lock:
+        try:
+            await redis.delete(get_session_lock_key(session_id))
```
majdyz: 🤖 🟠 **Should Fix**: The PR description claims `pytest … 89 passed`, but none of those 89 tests exercise this new behaviour. Grep for `release_lock` / `_redirect_to_pending_buffer` / `mark_session_completed` in `stream_registry_test.py`, `manager_test.py`, or anywhere else turns up only unrelated hits in `rate_limit_test.py` (different `release_reset_lock` symbol). Add unit tests for: (1) `mark_session_completed(release_lock=True)` actually calls `redis.delete(get_session_lock_key(...))`; (2) `release_lock=False` skips the delete; (3) `_redirect_to_pending_buffer` on cross-pod contention queues to pending + publishes a StreamError. The release-lock branch is a one-liner to mock and is the core of Fix 1.

autogpt_platform/backend/backend/copilot/executor/manager.py
```diff
@@ -366,16 +367,18 @@ def ack_message(reject: bool, requeue: bool):
...
         )
         current_owner = cluster_lock.try_acquire()
         if current_owner != self.executor_id:
             if current_owner is not None:
                 logger.warning(
-                    f"Session {session_id} already running on pod {current_owner}"
+                    f"Session {session_id} already running on pod {current_owner}; "
+                    f"re-queueing message into pending buffer"
                 )
+                self._redirect_to_pending_buffer(entry)
```
majdyz: 🤖 🟡 **Nice to Have**: Dropping a user message into the pending buffer silently from inside the pika consumer loses observability on how often cross-pod lock contention actually happens. Consider incrementing a Prometheus counter (you already have `active_tasks_gauge` etc. in this file) so ops can alert on sustained contention. Also worth logging the fresh `entry.turn_id` alongside the `current_owner` so the log line at line 377-380 can be correlated with the phantom SSE failure the client receives.

autogpt_platform/backend/backend/copilot/stream_registry.py
```diff
@@ -851,6 +857,18 @@ async def mark_session_completed(
         logger.debug(f"Session {session_id} already completed/failed, skipping")
         return False
 
+    # Force-release the executor's cluster lock so the next enqueued turn can
+    # acquire it immediately. The lock holder's on_run_done will also release
+    # (idempotent delete); doing it here unblocks cases where the task hangs
+    # past the cancel timeout or a pod crash leaves the lock orphaned.
```
majdyz, line 857: 🤖 🟡 **Nice to Have**: Happy-path callers (e.g. `processor.py:411`, called inside `finally` of `execute_copilot_turn`) now release the cluster lock here, *before* `on_run_done` also releases it via `self._task_locks[session_id].release()` (manager.py:433). Both deletes target the same Redis key so it is idempotent — but the ordering means: between this `redis.delete(...)` and `on_run_done` firing, a newly enqueued turn on another pod can acquire the lock while the outgoing pod is still inside `execute_copilot_turn`s cleanup (tear down transcript, close loops, etc.). In practice the turn is already done so there is no real concurrent execution, but it is worth a sentence in the docstring making the contract explicit: "the lock is a turn-scope lock, not a session-scope lock — expect overlap during cleanup."

autogpt_platform/backend/backend/copilot/executor/manager.py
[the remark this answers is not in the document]
majdyz: Dropped the worker-level contention handler in commit d7ad44e53 → force-push. `stream_chat_post` already queues incoming messages into the pending buffer via `is_turn_in_flight` + `queue_pending_for_http`. With Fix 1 (lock release atomic with meta flip), the API and worker see consistent state and the contention path never fires in the common case. Re-adding worker-side recovery would be premature defensive code; we can revisit if a true multi-pod race ever shows up in production.

autogpt_platform/backend/backend/copilot/executor/manager.py
majdyz: Dropped the worker-level contention handler in commit d7ad44e53 → force-push. `stream_chat_post` already queues incoming messages into the pending buffer via `is_turn_in_flight` + `queue_pending_for_http`. With Fix 1 (lock release atomic with meta flip), the API and worker see consistent state and the contention path never fires in the common case. Re-adding worker-side recovery would be premature defensive code; we can revisit if a true multi-pod race ever shows up in production.

autogpt_platform/backend/backend/copilot/executor/manager.py
majdyz: Dropped the worker-level contention handler in commit d7ad44e53 → force-push. `stream_chat_post` already queues incoming messages into the pending buffer via `is_turn_in_flight` + `queue_pending_for_http`. With Fix 1 (lock release atomic with meta flip), the API and worker see consistent state and the contention path never fires in the common case. Re-adding worker-side recovery would be premature defensive code; we can revisit if a true multi-pod race ever shows up in production.

autogpt_platform/backend/backend/copilot/executor/manager.py
majdyz: Dropped the worker-level contention handler in commit d7ad44e53 → force-push. `stream_chat_post` already queues incoming messages into the pending buffer via `is_turn_in_flight` + `queue_pending_for_http`. With Fix 1 (lock release atomic with meta flip), the API and worker see consistent state and the contention path never fires in the common case. Re-adding worker-side recovery would be premature defensive code; we can revisit if a true multi-pod race ever shows up in production.

autogpt_platform/backend/backend/copilot/executor/manager.py
```diff
@@ -435,6 +438,67 @@ def on_run_done(f: Future):
...
+
+        Uses a fresh event loop because the pika consumer callback is sync.
+        Crucially passes ``release_lock=False`` to ``mark_session_completed``:
+        the OLD pod is still legitimately running and owns the lock — stealing
+        it here would allow a concurrent execution on the same session.
+        """
+        from backend.copilot import stream_registry
+        from backend.copilot.pending_message_helpers import queue_user_message
+        from backend.copilot.pending_messages import PendingMessageContext
+        from backend.copilot.response_model import StreamError
```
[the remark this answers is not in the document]
majdyz: Dropped the worker-level contention handler in commit d7ad44e53 → force-push. `stream_chat_post` already queues incoming messages into the pending buffer via `is_turn_in_flight` + `queue_pending_for_http`. With Fix 1 (lock release atomic with meta flip), the API and worker see consistent state and the contention path never fires in the common case. Re-adding worker-side recovery would be premature defensive code; we can revisit if a true multi-pod race ever shows up in production.

autogpt_platform/backend/backend/copilot/stream_registry.py
majdyz: Adding a unit test for this in the next commit \u2014 asserting `redis.delete(get_session_lock_key(session_id))` fires when the CAS swap succeeds.

autogpt_platform/backend/backend/copilot/stream_registry.py
majdyz: Happy-path lock-release race noted \u2014 between `mark_session_completed`'s delete and `on_run_done`'s delete a sibling pod can briefly hold the lock while pod A's future is still in its cleanup tail. Pod A is not writing during that window (the turn already yielded through its finally), so it's benign. Keeping the nit flagged; if we see a concrete symptom we can move the delete into `on_run_done` only. For this PR the win from closing the 5-min cancel-timeout window outweighs the theoretical tail-window concern.

autogpt_platform/backend/backend/copilot/stream_registry_test.py
```diff
@@ -221,3 +223,109 @@ async def _inner():
...
+
+    assert result is True
+    assert get_session_lock_key("sess-1") in fake_redis.deleted_keys
+
+
+@pytest.mark.asyncio
+async def test_mark_session_completed_skips_lock_release_when_already_completed():
+    """CAS failure = someone else completed the session first; we must not
+    clobber their already-released lock (harmless but wastes a round trip)
+    or, more importantly, publish StreamFinish twice."""
```
majdyz: 🤖 🔵 **Nit**: Test docstring mentions "publish StreamFinish twice" as a concern but the test only asserts the lock key is NOT in `deleted_keys`. If the intent is to protect against double-finish, add `publish_chunk` as a mock and assert it is NOT called with `StreamFinish` on the CAS-no-op branch. Otherwise trim the docstring to match what the test actually guards.

majdyz:
🤖 🔵 **Nit**: PR title still reads `fix(backend/copilot): release stale session lock + requeue contention messages`, but the `+ requeue contention messages` half refers to the `_redirect_to_pending_buffer` worker-level handler that was dropped in `d7ad44e53`. Consider squashing the title to `fix(backend/copilot): release stale session lock on mark_session_completed` so the merge commit and changelog match the actual shipped behaviour.

autogpt_platform/backend/backend/copilot/stream_registry_test.py
majdyz: 🤖 Fixed in [bd89d549a](https://github.com/Significant-Gravitas/AutoGPT/commit/bd89d549ac42da4ec35a717a7dbb5b01658ed692): test now also patches `publish_chunk` and asserts `StreamFinish` is not re-published on the CAS-no-op branch — docstring matches what the test guards.

autogpt_platform/backend/backend/copilot/executor/manager.py
```diff
@@ -366,16 +367,18 @@ def ack_message(reject: bool, requeue: bool):
...
         current_owner = cluster_lock.try_acquire()
         if current_owner != self.executor_id:
             if current_owner is not None:
                 logger.warning(
-                    f"Session {session_id} already running on pod {current_owner}"
+                    f"Session {session_id} already running on pod {current_owner}; "
+                    f"re-queueing message into pending buffer"
                 )
+                self._redirect_to_pending_buffer(entry)
                 ack_message(reject=True, requeue=False)
```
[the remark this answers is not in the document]
majdyz: 🤖 Historical — resolved by dropping the `_redirect_to_pending_buffer` handler entirely in [d7ad44e53](https://github.com/Significant-Gravitas/AutoGPT/commit/d7ad44e53). The worker no longer tries to redirect on lock contention; the API-layer `is_turn_in_flight` + `queue_pending_for_http` path handles queuing instead, and with the atomic lock release in `mark_session_completed` (the PR's fix) the worker does not see contention in the common case.

autogpt_platform/backend/backend/copilot/stream_registry.py
```diff
@@ -851,6 +851,15 @@ async def mark_session_completed(
...
+    try:
+        await redis.delete(get_session_lock_key(session_id))
+    except RedisError as e:
+        logger.warning(f"Failed to release cluster lock for session {session_id}: {e}")
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [088b4fb59](https://github.com/Significant-Gravitas/AutoGPT/commit/088b4fb59): `ClusterLock.release()` and `AsyncClusterLock.release()` now use a Lua CAS (`GET == owner_id → DEL, else 0`) so release is a no-op when the key has been force-released externally and re-acquired by a new owner. Covered by `test_release_does_not_wipe_successor_lock` in `cluster_lock_test.py` — races the exact scenario (force-delete → successor acquire → original release → assert successor's key intact). 33 tests pass locally.

PR #12870, opened 2026-04-21
feat(backend/autopilot): stream extended_thinking on baseline via OpenRouter

majdyz:
### Why / What / How

**Why:** Fast-mode autopilot never renders a Reasoning block. The frontend already has `ReasoningCollapse` wired up and the wire protocol already carries `StreamReasoning*` events (landed for SDK mode in #12853), but the baseline (OpenRouter OpenAI-compat) path never asks Anthropic for extended thinking and never parses reasoning deltas off the stream. Result: users on fast/standard get a good answer with no visible chain-of-thought, while SDK users see the full Reasoning collapse.

**What:** Plumb reasoning end-to-end through the baseline path by opting into OpenRouter's non-OpenAI `reasoning` extension, parsing the reasoning delta fields off each chunk, and emitting the same `StreamReasoningStart/Delta/End` events the SDK adapter already uses.

**How:**
- **New config:** `baseline_reasoning_max_tokens` (default 8192; 0 disables). Sent as `extra_body={"reasoning": {"max_tokens": N}}` only on Anthropic routes — other providers drop the field, and `is_anthropic_model()` already gates this.
- **Delta extraction:** `_extract_reasoning_delta()` handles all three OpenRouter/provider variants in priority order — legacy `delta.reasoning` (string), DeepSeek-style `delta.reasoning_content`, and the structured `delta.reasoning_details` list (text/summary entries; encrypted or unknown entries are skipped).
- **Event emission:** Reasoning uses the same state-machine rules the SDK adapter uses — a text delta or tool_use delta arriving mid-stream closes the open reasoning block first, so the AI SDK v5 transport keeps reasoning / text / tool-use as distinct UI parts. On stream end, any still-open reasoning block gets a matching `reasoning-end` so a reasoning-only turn still finalises the frontend collapse.
- **Scope:** Live streaming only. Reasoning is not persisted to `ChatMessage` rows or the transcript builder in this PR (SDK path does so via `content_blocks=[{type: 'thinking', ...}]`, but that round-trip requires Anthropic signature plumbing baseline doesn't have today). Reload will still not show reasoning on baseline sessions — can follow up if we decide it's worth the signature handling.

### Changes

- `backend/copilot/config.py` — new `baseline_reasoning_max_tokens` field.
- `backend/copilot/baseline/service.py` — new `_extract_reasoning_delta()` helper; reasoning block state on `_BaselineStreamState`; `reasoning` gated into `extra_body`; chunk loop emits `StreamReasoning*` events with text/tool_use transition rules; stream-end closes any open reasoning block.
- `backend/copilot/baseline/service_unit_test.py` — 11 new tests covering extractor variants (legacy string, deepseek alias, structured list with text/summary aliases, encrypted-skip, empty), paired event ordering (reasoning-end before text-start), reasoning-only streams, and that the `reasoning` request param is correctly gated by model route (Anthropic vs non-Anthropic) and by the config flag.

### Checklist

For code changes:
- [x] I have made a test plan
  - [x] `poetry run pytest backend/copilot/baseline/service_unit_test.py backend/copilot/baseline/transcript_integration_test.py` — 103 passed
  - [ ] Manual: with `CHAT_USE_CLAUDE_AGENT_SDK=false` and `CHAT_MODEL=anthropic/claude-sonnet-4-6`, send a multi-step prompt on fast mode and confirm a Reasoning collapse appears alongside the final text
  - [ ] Manual: flip `CHAT_BASELINE_REASONING_MAX_TOKENS=0` and confirm baseline responses revert to text-only (no reasoning param, no reasoning UI)
  - [ ] Manual: with a non-Anthropic baseline model (`openai/gpt-4o`), confirm the request does NOT include `reasoning` and nothing regresses

For configuration changes:
- [x] `.env.default` is compatible — new setting falls back to the pydantic default

### DIFF
--- a/autogpt_platform/backend/backend/copilot/baseline/reasoning.py
+++ b/autogpt_platform/backend/backend/copilot/baseline/reasoning.py
@@ -28,6 +28,7 @@ from typing import Any
 from openai.types.chat.chat_completion_chunk import ChoiceDelta
 from pydantic import BaseModel, ConfigDict, Field
 
+from backend.copilot.model import ChatMessage
 from backend.copilot.response_model import (
     StreamBaseResponse,
     StreamReasoningDelta,
@@ -121,22 +122,37 @@ def reasoning_extra_body(model: str, max_thinking_tokens: int) -> dict[str, Any]
 class BaselineReasoningEmitter:
     """Owns the reasoning block lifecycle for one streaming round.
 
-    The AI SDK v5 wire format pairs every ``reasoning-start`` with a
-    matching ``reasoning-end``, and treats reasoning / text / tool-use as
-    three distinct UI parts that must not interleave.  This class folds
-    the state machine (open/closed, block id rotation) behind two methods:
+    Two concerns live here, both driven by the same state machine:
 
-    * :meth:`on_delta` — absorb one streaming chunk's delta, return any
-      events to emit (``ReasoningStart`` on first reasoning text, then a
-      ``ReasoningDelta`` per chunk).
-    * :meth:`close` — idempotently end any open block and rotate the id.
-      Call this before emitting text / tool_use, and once more when the
-      stream terminates.
+    1. **Wire events.**  The AI SDK v6 wire format pairs every
+       ``reasoning-start`` with a matching ``reasoning-end`` and treats
+       reasoning / text / tool-use as distinct UI parts that must not
+       interleave.
+    2. **Session persistence.**  ``ChatMessage(role="reasoning")`` rows in
+       ``session.messages`` are what
+       ``convertChatSessionToUiMessages.ts`` folds into the assistant
+       bubble as ``{type: "reasoning"}`` UI parts on reload and on
+       ``useHydrateOnStreamEnd`` swaps.  Without them the live-streamed
+       reasoning parts get overwritten by the hydrated (reasoning-less)
+       message list the moment the stream ends.  Mirrors the SDK path's
+       ``acc.reasoning_response`` pattern so both routes render the same
+       way on reload.
+
+    Pass ``session_messages`` to enable persistence; omit for pure
+    wire-emission (tests, scratch callers).  On first reasoning delta a
+    fresh ``ChatMessage(role="reasoning")`` is appended and mutated
+    in-place as further deltas arrive; :meth:`close` drops the reference
+    but leaves the appended row intact.
     """
 
-    def __init__(self) -> None:
+    def __init__(
+        self,
+        session_messages: list[ChatMessage] | None = None,
+    ) -> None:
         self._block_id: str = str(uuid.uuid4())
         self._open: bool = False
+        self._session_messages = session_messages
+        self._current_row: ChatMessage | None = None
 
     @property
     def is_open(self) -> bool:
@@ -147,6 +163,9 @@ class BaselineReasoningEmitter:
 
         Empty list when the chunk carries no reasoning payload, so this is
         safe to call on every chunk without guarding at the call site.
+        Persistence (when a session message list is attached) happens in
+        lockstep with emission so the row's content stays equal to the
+        concatenated deltas at every delta boundary.
         """
         ext = OpenRouterDeltaExtension.from_delta(delta)
         text = ext.visible_text()
@@ -156,7 +175,12 @@ class BaselineReasoningEmitter:
         if not self._open:
             events.append(StreamReasoningStart(id=self._block_id))
             self._open = True
+            if self._session_messages is not None:
+                self._current_row = ChatMessage(role="reasoning", content="")
+                self._session_messages.append(self._current_row)
         events.append(StreamReasoningDelta(id=self._block_id, delta=text))
+        if self._current_row is not None:
+            self._current_row.content = (self._current_row.content or "") + text
         return events
 
     def close(self) -> list[StreamBaseResponse]:
@@ -164,11 +188,14 @@ class BaselineReasoningEmitter:
 
         Idempotent — returns ``[]`` when no block is open.  The id rotation
         guarantees the next reasoning block starts with a fresh id rather
-        than reusing one already closed on the wire.
+        than reusing one already closed on the wire.  The persisted row is
+        not removed — it stays in ``session_messages`` as the durable
+        record of what was reasoned.
         """
         if not self._open:
             return []
         event = StreamReasoningEnd(id=self._block_id)
         self._open = False
         self._block_id = str(uuid.uuid4())
+        self._current_row = None
         return [event]
--- a/autogpt_platform/backend/backend/copilot/baseline/reasoning.py
+++ b/autogpt_platform/backend/backend/copilot/baseline/reasoning.py
@@ -1,0 +1,174 @@
+"""Extended-thinking wire support for the baseline (OpenRouter) path.
+
+Anthropic routes on OpenRouter expose extended thinking through
+non-OpenAI extension fields that the OpenAI Python SDK doesn't model:
+
+* ``reasoning`` (legacy string) — enabled by ``include_reasoning: true``.
+* ``reasoning_content`` — DeepSeek / some OpenRouter routes.
+* ``reasoning_details`` — structured list shipped with the unified
+  ``reasoning`` request param.
+
+This module keeps the wire-level concerns in one place:
+
+* :class:`OpenRouterDeltaExtension` validates the extension dict pulled off
+  ``ChoiceDelta.model_extra`` into typed pydantic models — no ``getattr`` +
+  ``isinstance`` duck-typing at the call site.
+* :class:`BaselineReasoningEmitter` owns the reasoning block lifecycle for
+  one streaming round and emits ``StreamReasoning*`` events so the caller
+  only has to plumb the events into its pending queue.
+* :func:`reasoning_extra_body` builds the ``extra_body`` fragment for the
+  OpenAI client call.  Returns ``None`` on non-Anthropic routes.
+"""
+
+from __future__ import annotations
+
+import uuid
+from typing import Any
+
+from openai.types.chat.chat_completion_chunk import ChoiceDelta
+from pydantic import BaseModel, ConfigDict, Field
+
+from backend.copilot.response_model import (
+    StreamBaseResponse,
+    StreamReasoningDelta,
+    StreamReasoningEnd,
+    StreamReasoningStart,
+)
+
+
+class ReasoningDetail(BaseModel):
+    """One entry in OpenRouter's ``reasoning_details`` list.
+
+    OpenRouter ships ``type: "reasoning.text"`` / ``"reasoning.summary"`` /
+    ``"reasoning.encrypted"`` entries.  Only the first two carry
+    user-visible text; encrypted entries are opaque and omitted from the
+    rendered collapse.  Unknown future types are tolerated (``extra="ignore"``)
+    so an upstream addition doesn't crash the stream.
+    """
+
+    model_config = ConfigDict(extra="ignore")
+
+    type: str | None = None
+    text: str | None = None
+    summary: str | None = None
+
+    @property
+    def visible_text(self) -> str:
+        """Return the human-readable text for this entry, or ``""``."""
+        return self.text or self.summary or ""
+
+
+class OpenRouterDeltaExtension(BaseModel):
+    """Non-OpenAI fields OpenRouter adds to streaming deltas.
+
+    Instantiate via :meth:`from_delta` which pulls the extension dict off
+    ``ChoiceDelta.model_extra`` (where pydantic v2 stashes fields that
+    aren't part of the declared schema) and validates it through this
+    model.  That keeps the parser honest — malformed entries surface as
+    validation errors rather than silent ``None``-coalesce bugs — and
+    avoids the ``getattr`` + ``isinstance`` duck-typing the earlier inline
+    extractor relied on.
+    """
+
+    model_config = ConfigDict(extra="ignore")
+
+    reasoning: str | None = None
+    reasoning_content: str | None = None
+    reasoning_details: list[ReasoningDetail] = Field(default_factory=list)
+
+    @classmethod
+    def from_delta(cls, delta: ChoiceDelta) -> "OpenRouterDeltaExtension":
+        """Build an extension view from ``delta.model_extra``."""
+        return cls.model_validate(delta.model_extra or {})
+
+    def visible_text(self) -> str:
+        """Concatenated reasoning text, pulled from whichever channel is set.
+
+        Priority: the legacy ``reasoning`` string, then DeepSeek's
+        ``reasoning_content``, then the concatenation of text-bearing
+        entries in ``reasoning_details``.  Only one channel is set per
+        provider in practice; the priority order just makes the fallback
+        deterministic if a provider ever emits multiple.
+        """
+        if self.reasoning:
+            return self.reasoning
+        if self.reasoning_content:
+            return self.reasoning_content
+        return "".join(d.visible_text for d in self.reasoning_details)
+
+
+def _is_anthropic_route(model: str) -> bool:
+    """Mirror of the private ``_is_anthropic_model`` check in service.py.
+
+    Kept as a free function here to avoid a circular import — ``service.py``
+    imports this module, not the other way around.
+    """
+    lowered = model.lower()
+    return "claude" in lowered or lowered.startswith("anthropic")
+
+
+def reasoning_extra_body(model: str, max_thinking_tokens: int) -> dict[str, Any] | None:
+    """Build the ``extra_body["reasoning"]`` fragment for the OpenAI client.
+
+    Returns ``None`` for non-Anthropic routes — other OpenRouter providers
+    drop the field but we skip it anyway to keep the payload minimal.
+    """
+    if not _is_anthropic_route(model):
+        return None
+    return {"reasoning": {"max_tokens": max_thinking_tokens}}
+
+
+class BaselineReasoningEmitter:
+    """Owns the reasoning block lifecycle for one streaming round.
+
+    The AI SDK v5 wire format pairs every ``reasoning-start`` with a
+    matching ``reasoning-end``, and treats reasoning / text / tool-use as
+    three distinct UI parts that must not interleave.  This class folds
+    the state machine (open/closed, block id rotation) behind two methods:
+
+    * :meth:`on_delta` — absorb one streaming chunk's delta, return any
+      events to emit (``ReasoningStart`` on first reasoning text, then a
+      ``ReasoningDelta`` per chunk).
+    * :meth:`close` — idempotently end any open block and rotate the id.
+      Call this before emitting text / tool_use, and once more when the
+      stream terminates.
+    """
+
+    def __init__(self) -> None:
+        self._block_id: str = str(uuid.uuid4())
+        self._open: bool = False
+
+    @property
+    def is_open(self) -> bool:
+        return self._open
+
+    def on_delta(self, delta: ChoiceDelta) -> list[StreamBaseResponse]:
+        """Return events for the reasoning text carried by *delta*.
+
+        Empty list when the chunk carries no reasoning payload, so this is
+        safe to call on every chunk without guarding at the call site.
+        """
+        ext = OpenRouterDeltaExtension.from_delta(delta)
+        text = ext.visible_text()
+        if not text:
+            return []
+        events: list[StreamBaseResponse] = []
+        if not self._open:
+            events.append(StreamReasoningStart(id=self._block_id))
+            self._open = True
+        events.append(StreamReasoningDelta(id=self._block_id, delta=text))
+        return events
+
+    def close(self) -> list[StreamBaseResponse]:
+        """Emit ``StreamReasoningEnd`` for the open block (if any) and rotate.
+
+        Idempotent — returns ``[]`` when no block is open.  The id rotation
+        guarantees the next reasoning block starts with a fresh id rather
+        than reusing one already closed on the wire.
+        """
+        if not self._open:
+            return []
+        event = StreamReasoningEnd(id=self._block_id)
+        self._open = False
+        self._block_id = str(uuid.uuid4())
+        return [event]
--- a/autogpt_platform/backend/backend/copilot/baseline/reasoning.py
+++ b/autogpt_platform/backend/backend/copilot/baseline/reasoning.py
@@ -22,11 +22,12 @@ This module keeps the wire-level concerns in one place:
 
 from __future__ import annotations
 
+import logging
 import uuid
 from typing import Any
 
 from openai.types.chat.chat_completion_chunk import ChoiceDelta
-from pydantic import BaseModel, ConfigDict, Field
+from pydantic import BaseModel, ConfigDict, Field, ValidationError
 
 from backend.copilot.model import ChatMessage
 from backend.copilot.response_model import (
@@ -36,6 +37,11 @@ from backend.copilot.response_model import (
     StreamReasoningStart,
 )
 
+logger = logging.getLogger(__name__)
+
+
+_VISIBLE_REASONING_TYPES = frozenset({"reasoning.text", "reasoning.summary"})
+
 
 class ReasoningDetail(BaseModel):
     """One entry in OpenRouter's ``reasoning_details`` list.
@@ -44,7 +50,10 @@ class ReasoningDetail(BaseModel):
     ``"reasoning.encrypted"`` entries.  Only the first two carry
     user-visible text; encrypted entries are opaque and omitted from the
     rendered collapse.  Unknown future types are tolerated (``extra="ignore"``)
-    so an upstream addition doesn't crash the stream.
+    so an upstream addition doesn't crash the stream — but their ``text`` /
+    ``summary`` fields are NOT surfaced because they may carry provider
+    metadata rather than user-visible reasoning (see
+    :attr:`visible_text`).
     """
 
     model_config = ConfigDict(extra="ignore")
@@ -55,7 +64,18 @@ class ReasoningDetail(BaseModel):
 
     @property
     def visible_text(self) -> str:
-        """Return the human-readable text for this entry, or ``""``."""
+        """Return the human-readable text for this entry, or ``""``.
+
+        Only entries with a recognised reasoning type (``reasoning.text`` /
+        ``reasoning.summary``) surface text; unknown or encrypted types
+        return an empty string even if they carry a ``text`` /
+        ``summary`` field, to guard against future provider metadata
+        being rendered as reasoning in the UI.  Entries missing a
+        ``type`` are treated as text (pre-``reasoning_details`` OpenRouter
+        payloads omit the field).
+        """
+        if self.type is not None and self.type not in _VISIBLE_REASONING_TYPES:
+            return ""
         return self.text or self.summary or ""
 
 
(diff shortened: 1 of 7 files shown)


autogpt_platform/backend/backend/copilot/baseline/service.py
```diff
@@ -591,7 +641,30 @@ async def _baseline_llm_caller(
...
+                    # Text and reasoning must not interleave on the wire — the
+                    # AI SDK maps distinct start/end pairs to distinct UI
+                    # parts.  Close any open reasoning block before emitting
+                    # the first text delta of this run.
+                    if state.reasoning_started:
+                        state.pending_events.append(
+                            StreamReasoningEnd(id=state.reasoning_block_id)
+                        )
+                        state.reasoning_started = False
+                        state.reasoning_block_id = str(uuid.uuid4())
```
majdyz: 🤖 🟡 **Nice to Have**: The reasoning-end + block-id rotation pattern is duplicated three times (text branch at L662-667, tool_calls branch at L684-689, stream-end at L718-721). Consider extracting a tiny helper like `_close_reasoning_block_if_open(state)` to DRY this up — easier to reason about invariants and harder for future edits to desync the three call sites.

autogpt_platform/backend/backend/copilot/baseline/service_unit_test.py
```diff
@@ -1508,3 +1517,256 @@ async def test_baseline_llm_caller_skips_memoisation_for_non_anthropic(self):
...
+
+        # All reasoning deltas share a single block id; the text block uses
+        # a fresh id after the reasoning-end rotation.
+        reasoning_ids = {
+            e.id
+            for e in state.pending_events
+            if isinstance(
+                e, (StreamReasoningStart, StreamReasoningDelta, StreamReasoningEnd)
+            )
+        }
```
majdyz, line 1624: 🤖 🟠 **Should Fix**: The tool_calls branch that closes an open reasoning block (`service.py` L684-689) has no test coverage. `test_reasoning_then_text_emits_paired_events` only exercises the text branch. A tool-call arriving mid-reasoning is a realistic Anthropic streaming case (extended thinking → tool_use) and the PR description explicitly calls out this rule. Please add a test that streams `reasoning → tool_call` chunks and asserts `StreamReasoningEnd` is emitted before the tool call is flushed.

autogpt_platform/backend/backend/copilot/baseline/service.py
```diff
@@ -485,6 +490,41 @@ def _mark_system_message_with_cache_control(
...
+        return legacy
+    details = getattr(delta, "reasoning_details", None)
+    if not details:
+        return ""
+    parts: list[str] = []
+    for entry in details:
+        if not isinstance(entry, dict):
+            continue
+        text = entry.get("text") or entry.get("summary")
+        if isinstance(text, str) and text:
```
majdyz: 🤖 🟡 **Nice to Have**: `_extract_reasoning_delta` only accepts entries that are plain `dict`. If the OpenAI Python SDK ever upgrades and delivers `reasoning_details` as typed Pydantic models (it already does this for other fields via `ChoiceDelta.model_extra`), the extractor will silently drop every entry. Consider falling back to attribute access (`getattr(entry, 'text', None) or getattr(entry, 'summary', None)`) when `entry` isn't a dict but does carry a text-like attribute — safer against upstream drift and costs almost nothing.

autogpt_platform/backend/backend/copilot/baseline/service.py
majdyz: 🤖 Fixed in [3d7b38162](https://github.com/Significant-Gravitas/AutoGPT/commit/3d7b38162040c312db94155422fe441edfe0825b): extracted `_close_reasoning_block_if_open(state)`; the three call sites (text, tool_calls, stream-end) now all go through the same helper.

autogpt_platform/backend/backend/copilot/baseline/service_unit_test.py
majdyz: 🤖 Fixed in [3d7b38162](https://github.com/Significant-Gravitas/AutoGPT/commit/3d7b38162040c312db94155422fe441edfe0825b): added `test_reasoning_then_tool_call_closes_reasoning_first` — streams a `reasoning` chunk followed by a `tool_calls` chunk, asserts `StreamReasoningEnd` is emitted (exercising the tool_calls branch, not the stream-end fallback) and that the tool_call was still collected for execution.

autogpt_platform/backend/backend/copilot/baseline/service.py
majdyz: 🤖 Fixed in [3d7b38162](https://github.com/Significant-Gravitas/AutoGPT/commit/3d7b38162040c312db94155422fe441edfe0825b): `_extract_reasoning_delta` now falls back to `getattr(entry, 'text', None)` / `getattr(entry, 'summary', None)` for non-dict entries, plus a new `test_structured_details_accept_typed_pydantic_entries` covering the attribute-access path.

autogpt_platform/backend/backend/copilot/baseline/service.py
```diff
@@ -629,6 +724,11 @@ async def _baseline_llm_caller(
             except Exception:
                 pass
 
+        # Close any still-open reasoning block before we emit the final
+        # text tail / close the text block.  A stream that ends with
+        # reasoning-only (no text, no tool_use) still needs a matched
+        # ``reasoning-end`` so the frontend collapse finalises.
+        _close_reasoning_block_if_open(state)
```
majdyz: 🤖 🔴 **Blocker**: Exception mid-stream leaves reasoning block unclosed. The `_close_reasoning_block_if_open(state)` call at L731 sits in the outer `try` of `_baseline_llm_caller`, so if the `async for chunk in response` loop raises (network drop, provider 500, cancel) while a reasoning block is open, we skip the close. The outer handler in `stream_chat_completion_baseline` (L1829) patches `StreamTextEnd` before `StreamFinishStep` but has no equivalent for reasoning — frontend collapse never finalises. CodeRabbit independently flagged this (review 4147305556). Recommended fix: move the close-reasoning + flush + close-text sequence into the `try/finally` at L747 (or into the inner try/finally at L718) so it runs on both normal and exception paths. Also needs a regression-guard test: inject a stream that yields a reasoning delta then raises.

majdyz: 🤖 Fixed in [080d42b9d](https://github.com/Significant-Gravitas/AutoGPT/commit/080d42b9da9c450e97c637f1e90ead077a96fa85): moved `_close_reasoning_block_if_open` + thinking-stripper flush + `StreamTextEnd` emission into `_baseline_llm_caller`'s outer `finally` so they run on both normal and exception paths. Also dropped the now-redundant `StreamTextEnd` insert-before-`StreamFinishStep` patch in the outer exception handler since the inner finally already handles it. Added `test_reasoning_closed_on_mid_stream_exception` as a regression guard — streams a reasoning delta then raises, asserts `StreamReasoningEnd` precedes `StreamFinishStep`.

majdyz:
🤖 @coderabbitai addressed in [080d42b9d](https://github.com/Significant-Gravitas/AutoGPT/commit/080d42b9da9c450e97c637f1e90ead077a96fa85) and [3d7b38162](https://github.com/Significant-Gravitas/AutoGPT/commit/3d7b38162040c312db94155422fe441edfe0825b):

- **Outside-diff blocker** (unclosed reasoning on mid-stream exception): moved close-reasoning + stripper flush + `StreamTextEnd` into `_baseline_llm_caller`'s outer `finally` so both normal and exception paths emit matched end events before `StreamFinishStep`. Dropped the now-redundant insert-before-`StreamFinishStep` patch in the outer handler.
- **DRY nitpick** (service.py 657-689): extracted `_close_reasoning_block_if_open(state)` helper in [3d7b38162](https://github.com/Significant-Gravitas/AutoGPT/commit/3d7b38162040c312db94155422fe441edfe0825b); all four call sites (text branch, tool_calls branch, stream-end, exception path) go through the helper.
- **Test gap nitpick** (missing mid-stream-exception test): added `test_reasoning_closed_on_mid_stream_exception` in this commit as a regression guard.

autogpt_platform/backend/backend/copilot/baseline/service.py
```diff
@@ -485,6 +490,64 @@ def _mark_system_message_with_cache_control(
...
+        if isinstance(entry, dict):
+            text = entry.get("text") or entry.get("summary")
+        else:
+            text = getattr(entry, "text", None) or getattr(entry, "summary", None)
+        if isinstance(text, str) and text:
+            parts.append(text)
+    return "".join(parts)
+
+
+def _close_reasoning_block_if_open(state: "_BaselineStreamState") -> None:
```
majdyz: 🤖 🔵 **Nit**: The forward-reference string around the `_BaselineStreamState` annotation isn't needed — the class is defined earlier in the module (L330) than this function (L533), so `state: _BaselineStreamState` works directly. Minor; only worth changing to stay consistent with other module-level helpers that reference the same type.

majdyz: 🤖 Fixed in [2517dae85](https://github.com/Significant-Gravitas/AutoGPT/commit/2517dae85a763dc709cf71741d3bbb5d5f1bbf74): dropped the `"_BaselineStreamState"` quotes.

autogpt_platform/backend/backend/copilot/baseline/service_unit_test.py
```diff
@@ -1508,3 +1516,261 @@ async def test_baseline_llm_caller_skips_memoisation_for_non_anthropic(self):
...
+    tc.index = index
+    tc.id = call_id
+    function = MagicMock()
+    function.name = name
+    function.arguments = arguments
+    tc.function = function
+    return tc
+
+
+class TestBaselineReasoningStreaming:
```
majdyz, line 1577: 🤖 🟠 **Should Fix**: Coverage regression — `test_reasoning_only_stream_still_closes_block` and `test_reasoning_param_suppressed_when_config_zero` were removed but no equivalent integration tests were added in their place. `reasoning_test.py` only covers the emitter in isolation (via hand-built `ChoiceDelta` instances) — it does NOT exercise the `_baseline_llm_caller` finally-block that closes the reasoning emitter on stream end for reasoning-only streams, and does NOT assert that an anthropic route with a zero/low `max_thinking_tokens` builds a well-formed request. The end-to-end class `TestBaselineReasoningStreaming` still has `test_reasoning_param_sent_on_anthropic_routes` + `_absent_on_non_anthropic_routes` but lost the zero-tokens and reasoning-only cases. Please restore both as integration tests against `_baseline_llm_caller` (reasoning-only stream must still emit paired start/end; the config max=0 path is now dead since `ge=1024` — document that explicitly or drop the `if max_thinking_tokens <= 0` branch from `reasoning_extra_body`).

autogpt_platform/backend/backend/copilot/baseline/reasoning.py
```diff
@@ -0,0 +1,201 @@
...
+
+
+def _is_anthropic_route(model: str) -> bool:
+    """Mirror of the private ``_is_anthropic_model`` check in service.py.
+
+    Kept as a free function here to avoid a circular import — ``service.py``
+    imports this module, not the other way around.
+    """
+    lowered = model.lower()
+    return "claude" in lowered or lowered.startswith("anthropic")
```
majdyz: 🤖 🟡 **Nice to Have**: Logic duplication — `_is_anthropic_route` here copy-pastes the rules from `service._is_anthropic_model` (same `lowered.startswith('anthropic')` + `'claude' in lowered`). The docstring acknowledges it's a mirror to avoid a circular import. If `service.py` imports this module but not vice-versa, there's no cycle — you could just import `_is_anthropic_model` from service (or, better, promote it to a public helper in a third shared module like `backend.copilot.baseline.model_utils`). Keeping two copies invites drift the moment someone adds a new route prefix.

autogpt_platform/backend/backend/copilot/baseline/reasoning.py
```diff
@@ -0,0 +1,201 @@
...
+
+def reasoning_extra_body(model: str, max_thinking_tokens: int) -> dict[str, Any] | None:
+    """Build the ``extra_body["reasoning"]`` fragment for the OpenAI client.
+
+    Returns ``None`` for non-Anthropic routes — other OpenRouter providers
+    drop the field but we skip it anyway to keep the payload minimal.
+    """
+    if not _is_anthropic_route(model):
+        return None
+    return {"reasoning": {"max_tokens": max_thinking_tokens}}
```
majdyz, line 148: 🤖 🟠 **Should Fix**: Silent behavior deletion — the old `baseline_reasoning_max_tokens` had `ge=0` with a `> 0` guard in the caller so operators could kill-switch reasoning by setting the env var to 0. After unifying on `claude_agent_max_thinking_tokens` (`ge=1024`), there is no longer a way to disable reasoning on the baseline path without also flipping the SDK path. This is an undocumented behavior change. Either (a) add back a small `if max_thinking_tokens <= 0: return None` guard in `reasoning_extra_body` plus a separate `baseline_reasoning_enabled: bool` flag, or (b) call it out explicitly in the PR description so reviewers know the kill switch is gone.

autogpt_platform/backend/backend/copilot/baseline/service.py
```diff
@@ -485,6 +490,64 @@ def _mark_system_message_with_cache_control(
...
+    if not details:
+        return ""
+    parts: list[str] = []
+    for entry in details:
+        if isinstance(entry, dict):
+            text = entry.get("text") or entry.get("summary")
+        else:
+            text = getattr(entry, "text", None) or getattr(entry, "summary", None)
+        if isinstance(text, str) and text:
+            parts.append(text)
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [08b568021](https://github.com/Significant-Gravitas/AutoGPT/commit/08b568021ba5e4fc5f1aa0a99ba235c1a760cbb2): `ReasoningDetail.visible_text` now returns `""` when `type` is set but not in `{reasoning.text, reasoning.summary}` (entries without a `type` stay compatible). Added `test_visible_text_empty_for_unknown_type` + `test_unknown_typed_entry_with_text_is_not_surfaced` as regression cases.

autogpt_platform/backend/backend/copilot/baseline/reasoning_test.py
```diff
@@ -0,0 +1,172 @@
...
+        events = emitter.close()
+        assert len(events) == 1
+        assert isinstance(events[0], StreamReasoningEnd)
+        assert events[0].id == first_id
+        assert emitter.is_open is False
+        # Next reasoning uses a fresh id.
+        new_events = emitter.on_delta(_delta(reasoning="y"))
+        assert isinstance(new_events[0], StreamReasoningStart)
+        assert new_events[0].id != first_id
+
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [08b568021](https://github.com/Significant-Gravitas/AutoGPT/commit/08b568021ba5e4fc5f1aa0a99ba235c1a760cbb2): `test_close_emits_end_and_rotates_id` now captures the id from the emitted `StreamReasoningStart` (`start_events[0].id`). No more `# pyright: ignore` on that line.

autogpt_platform/backend/backend/copilot/baseline/reasoning.py
```diff
@@ -0,0 +1,201 @@
...
+    model_config = ConfigDict(extra="ignore")
+
+    reasoning: str | None = None
+    reasoning_content: str | None = None
+    reasoning_details: list[ReasoningDetail] = Field(default_factory=list)
+
+    @classmethod
+    def from_delta(cls, delta: ChoiceDelta) -> "OpenRouterDeltaExtension":
+        """Build an extension view from ``delta.model_extra``."""
+        return cls.model_validate(delta.model_extra or {})
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [08b568021](https://github.com/Significant-Gravitas/AutoGPT/commit/08b568021ba5e4fc5f1aa0a99ba235c1a760cbb2): `OpenRouterDeltaExtension.from_delta` now wraps `model_validate` in a try/except `ValidationError`, logs a warning, and returns an empty extension. A malformed `reasoning_details` (e.g. string instead of list) no longer aborts the stream; text and tool-call events keep flowing. Added `test_malformed_reasoning_payload_logged_and_swallowed` as a regression case.

autogpt_platform/backend/backend/copilot/baseline/service_unit_test.py
majdyz: 🤖 Fixed in [08b568021](https://github.com/Significant-Gravitas/AutoGPT/commit/08b568021ba5e4fc5f1aa0a99ba235c1a760cbb2): restored both integration tests against `_baseline_llm_caller` — `test_reasoning_only_stream_still_closes_block` confirms the finally-block closes reasoning when there's no text/tool, `test_reasoning_param_suppressed_when_thinking_tokens_zero` patches `config.claude_agent_max_thinking_tokens=0` and asserts `reasoning` drops out of `extra_body`.

autogpt_platform/backend/backend/copilot/baseline/reasoning.py
majdyz: 🤖 Fixed in [08b568021](https://github.com/Significant-Gravitas/AutoGPT/commit/08b568021ba5e4fc5f1aa0a99ba235c1a760cbb2): dropped `_is_anthropic_route` entirely; `reasoning_extra_body` now imports `_is_anthropic_model` from `service.py` lazily (keeps the dependency one-way without a circular load-time cycle). Single source of truth.

autogpt_platform/backend/backend/copilot/baseline/reasoning.py
majdyz: 🤖 Fixed in [08b568021](https://github.com/Significant-Gravitas/AutoGPT/commit/08b568021ba5e4fc5f1aa0a99ba235c1a760cbb2): added `max_thinking_tokens <= 0` guard to `reasoning_extra_body` — operators can set the env var to 0 (bypassing the pydantic `ge=1024` via in-memory override / test patch) to disable reasoning on the baseline path without affecting the SDK. Covered by `test_zero_max_tokens_kill_switch` (unit) + `test_reasoning_param_suppressed_when_thinking_tokens_zero` (integration).

autogpt_platform/backend/backend/copilot/baseline/reasoning_test.py
```diff
@@ -0,0 +1,280 @@
...
+    def test_malformed_reasoning_payload_logged_and_swallowed(self, caplog):
+        # A malformed payload (e.g. reasoning_details shipped as a string
+        # rather than a list) must not abort the stream — log it and
+        # return an empty extension so valid text/tool events keep flowing.
+        from openai.types.chat.chat_completion_chunk import ChoiceDelta
+
+        delta = ChoiceDelta.model_validate({"role": "assistant"})
+        object.__setattr__(
+            delta, "__pydantic_extra__", {"reasoning_details": "not a list"}
```
majdyz: 🤖 🔵 **Nit**: The test bypasses pydantic's schema via `object.__setattr__(delta, "__pydantic_extra__", ...)` — `__pydantic_extra__` is an internal pydantic-v2 attribute that could be renamed in a minor release. A more robust alternative is to pass a minimal mock with a dict `model_extra` attribute: `delta = MagicMock(spec=ChoiceDelta); delta.model_extra = {"reasoning_details": "not a list"}` — since `from_delta` only reads `delta.model_extra`, the mock satisfies the contract. Not a blocker; the current test passes and covers the real behavior.

autogpt_platform/backend/backend/copilot/baseline/service.py
```diff
@@ -360,6 +365,14 @@ class _BaselineStreamState:
     # wasn't a system role, so no marking applies).
     cached_system_message: dict[str, Any] | None = None
 
+    def __post_init__(self) -> None:
+        # Wire the reasoning emitter to ``session_messages`` so it can
+        # append ``role="reasoning"`` rows as reasoning streams in — the
+        # frontend's ``convertChatSessionToUiMessages`` relies on these
+        # rows to render the Reasoning collapse after the AI SDK's
+        # stream-end hydrate swaps in the DB-backed message list.
+        self.reasoning_emitter = BaselineReasoningEmitter(self.session_messages)
```
majdyz, line 378: 🤖 🟡 **Nice to Have**: The `__post_init__` pattern works, but the coupling between `_BaselineStreamState` and the emitter (sharing a reference to `self.session_messages`) is subtle. If anyone in the future reassigns `state.session_messages = [...]` (as opposed to `.clear()` / `.append(...)`), the emitter's reference silently goes stale and reasoning rows land in a detached list. A one-line docstring on `session_messages` (e.g. `# MUTATE in place only — reassignment breaks the reasoning emitter's persistence wiring.`) would make the invariant loud. Alternatively, pass `session_messages` to `on_delta` on each call — removes the shared-reference coupling entirely, at the cost of a slightly wider call site.

autogpt_platform/backend/backend/copilot/baseline/reasoning_test.py
majdyz: 🤖 Fixed in [c0a27ab87](https://github.com/Significant-Gravitas/AutoGPT/commit/c0a27ab878da24f3c7396a81106e5e3978b53122): replaced `object.__setattr__(__pydantic_extra__, ...)` with `MagicMock(spec=ChoiceDelta)` + `delta.model_extra = ...`. No more reach into pydantic internals; the mock satisfies the `from_delta` contract directly.

autogpt_platform/backend/backend/copilot/baseline/service.py
majdyz: 🤖 Fixed in [c0a27ab87](https://github.com/Significant-Gravitas/AutoGPT/commit/c0a27ab878da24f3c7396a81106e5e3978b53122): added a "MUTATE in place only" comment on `_BaselineStreamState.session_messages` documenting that the emitter shares this list reference — reassigning the attribute silently detaches the emitter's persistence wiring. Prefer in-place mutations (`append`, `clear`) or rebuild the state + emitter together.

autogpt_platform/backend/backend/copilot/baseline/service_unit_test.py
```diff
@@ -1508,3 +1516,327 @@ async def test_baseline_llm_caller_skips_memoisation_for_non_anthropic(self):
...
+
+        types = [type(e).__name__ for e in state.pending_events]
+        assert "StreamReasoningStart" in types
+        assert "StreamReasoningEnd" in types
+        # No text was produced — no text events should be emitted.
+        assert "StreamTextStart" not in types
+        assert "StreamTextDelta" not in types
+
+    @pytest.mark.asyncio
+    async def test_reasoning_param_suppressed_when_thinking_tokens_zero(self):
```
majdyz, line 1813: 🤖 🟡 **Nice to Have**: `TestReasoningPersistence` in `reasoning_test.py` covers the emitter's persistence contract in isolation, but no integration test exercises it through `_baseline_llm_caller` → `_BaselineStreamState.__post_init__`. A smoke assertion like `state = _BaselineStreamState(model="anthropic/..."); <run reasoning chunks>; assert any(m.role == "reasoning" for m in state.session_messages)` would catch future regressions where someone refactors `__post_init__` and accidentally passes the wrong list reference (or forgets the wiring entirely). Not a blocker — the unit test plus the existing `StreamReasoningStart/End` assertions give reasonable coverage of the wiring.

majdyz: 🤖 Fixed in [63c422977](https://github.com/Significant-Gravitas/AutoGPT/commit/63c422977484e99b1a8cf68775bcae1a9da4f867): added `test_reasoning_persists_to_state_session_messages` — drives reasoning deltas through `_baseline_llm_caller` and asserts a `role="reasoning"` row lands on `state.session_messages` with the concatenated delta content ("first thought"). A refactor that silently passes the wrong list reference in `__post_init__` will now fail this test.

majdyz:
## Manual /pr-test --fix (pass 2, native stack) — **all 3 golden paths PASS**

Re-ran end-to-end after the persistence fix ([`316b132a1`](https://github.com/Significant-Gravitas/AutoGPT/commit/316b132a1)). Pass-1 blocker (Reasoning collapse never rendering) is resolved.

### Path 1 — Anthropic fast mode (baseline) ✅

`CHAT_USE_CLAUDE_AGENT_SDK=false`, `CHAT_MODEL=anthropic/claude-sonnet-4-6`.
Prompt: *"Plan a 3-step migration to deduplicate our user records by email."*

- **Wire:** 177 `reasoning-delta` + 1 `reasoning-start` + 1 `reasoning-end` on the SSE stream.
- **DB:** `session.messages[1].role == "reasoning"`, content length 1799, matches concat of wire deltas byte-for-byte.
- **Live render — reasoning collapse appears during the stream:**

live stream showing reasoning

- **Expanded live:**

live reasoning expanded

- **Hard reload → still there (this is what pass 1 was missing):**

reload then expand reasoning

SSE excerpt:
```
data: {"type":"start-step"}
data: {"type":"reasoning-start","id":"030b4a72-cd2b-4378-9a7c-418ded30de2a"}
data: {"type":"reasoning-delta","id":"030b4a72-...","delta":"The user wants to plan a 3"}
data: {"type":"reasoning-delta","id":"030b4a72-...","delta":"-step migration to deduplicate user"}
...
```

Persisted row (truncated):
```json
{
  "role": "reasoning",
  "content": "The user wants to plan a 3-step migration... (1799 chars)",
  "sequence": 1
}
```

### Path 2 — Non-Anthropic (`openai/gpt-4o`) ✅

Zero reasoning events emitted (7 text-delta, no `reasoning-*`). Request body has no `reasoning` key. Session: `user` + `assistant` only — no stray reasoning row. `_is_anthropic_model` gate holding.

### Path 3 — SDK regression guard ✅

`CHAT_USE_CLAUDE_AGENT_SDK=true`, same prompt. SDK service selected, stream completed cleanly — refactor of baseline reasoning into `reasoning.py` didn't break SDK imports or registration.

### Additional screenshots

- Post-stream, steps collapse closed:

after live stream

- Show-steps expanded (reasoning lives inside the outer steps collapse):

show steps expanded

- Reasoning expanded inside steps:

reasoning expanded

### Recommendation: **APPROVE**

Blocker from pass 1 resolved; no new regressions; non-Anthropic and SDK paths unaffected. Full artifact set (SSE dumps, session JSONs, all 12 screenshots) on branch `test-screenshots/pr-12870`.

autogpt_platform/backend/backend/copilot/config.py
```diff
@@ -194,9 +194,11 @@ class ChatConfig(BaseSettings):
         default=8192,
         ge=1024,
         le=128000,
```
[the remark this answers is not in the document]
majdyz: Valid catch — fixed in [`e8d8552c4`](https://github.com/Significant-Gravitas/AutoGPT/pull/12870/commits/e8d8552c4).

Dropped the `ge=1024` constraint to `ge=0` so `CHAT_CLAUDE_AGENT_MAX_THINKING_TOKENS=0` passes validation. For symmetry, also gated SDK's `max_thinking_tokens` kwarg on `> 0` — the CLI treats 'kwarg absent' as 'extended thinking off', matching baseline's `reasoning_extra_body` returning `None`. Both paths now share the same kill-switch semantics.

Existing baseline test (`test_zero_max_tokens_kill_switch` in reasoning_test.py) covered the config value directly; it no longer needs to bypass validation.

PR #12871, opened 2026-04-21
feat(backend/copilot): Kimi K2.6 fast default + 4-config matrix + coalesced reasoning + web_search tool

majdyz:
### Why / What / How

**Why.** Three unrelated but interlocking problems on the baseline (OpenRouter) copilot path, all blocking us from making Kimi K2.6 the default fast model:

1. **Cost / capability gap on the default.** Kimi K2.6 prices at $0.60 / $2.80 per MTok — ~5x cheaper input and ~5.4x cheaper output than Sonnet 4.6 — while tying Opus on SWE-Bench Verified (80.2% vs 80.8%) and beating it on SWE-Bench Pro (58.6% vs 53.4%). OpenRouter exposes the same `reasoning` / `include_reasoning` extension on Moonshot endpoints that #12870 plumbed for Anthropic, so the reasoning collapse lights up end-to-end without per-provider code.
2. **Kimi reasoning deltas freeze the UI.** K2.6 emits ~4,700 reasoning-delta SSE events per turn vs ~28 on Sonnet — the AI SDK v6 Reasoning UIMessagePart can't keep up and the tab locks. Needs a coalescing buffer upstream.
3. **Kimi loops on `require_guide_read`.** The guide-guard checks `session.messages` for a prior `agent_building_guide` call, but tool calls aren't flushed to `session.messages` until the end of the turn — mid-turn the check keeps returning False and Kimi calls the guide-load tool repeatedly in the same turn. Needs an in-flight tracker that lives on `ChatSession`.
4. **No `web_search` tool on either path.** Kimi doesn't have a native web-search equivalent and the SDK path's native `WebSearch` (the Claude Code CLI's built-in) doesn't carry cost accounting. We need one implementation that both paths share and that reports cost through the same tracker as every other tool call.

**What.** Five grouped deltas on the baseline service, tool layer, and config:

- **Kimi K2.6 default.** `fast_standard_model` defaults to `moonshotai/kimi-k2.6`. Full 2×2 model matrix below. Rollback is one env var.
- **4-config model matrix.** `fast_standard_model` / `fast_advanced_model` / `thinking_standard_model` / `thinking_advanced_model`. Each cell independent so baseline can run a cheap provider at the standard tier without leaking into the SDK path (which is Anthropic-only by CLI contract). Legacy env vars (`CHAT_MODEL`, `CHAT_FAST_MODEL`, `CHAT_ADVANCED_MODEL`) stay aliased via `validation_alias` so live deployments keep resolving to the same effective cell.
- **Reasoning delta coalescing.** `BaselineReasoningEmitter` buffers deltas and flushes on a char-count OR time-interval threshold (32 chars / 40 ms). ~4,700 → ~150 SSE events per turn on Kimi; no perceptible change on Sonnet (which was already well under the threshold).
- **In-flight tool-call tracker.** `ChatSession._inflight_tool_calls` PrivateAttr is populated when a tool-call block is emitted and cleared at turn end. `session.has_tool_been_called_this_turn(name)` now returns True mid-turn, not just after the tool-result lands in `session.messages` — which is what `require_guide_read` needs to cut the loop.
- **New `web_search` copilot tool.** Wraps Anthropic's server-side `web_search_20250305` beta via `AsyncAnthropic` (direct — OpenRouter can't proxy server-side tool execution). Dispatches through `claude-haiku-4-5` with `max_uses=1`. Cost estimated from published rates ($0.010 per search + Haiku tokens) since the Anthropic Messages API doesn't report cost on the response; reported to `persist_and_record_usage(provider='anthropic')` on both paths. SDK native `WebSearch` moved from `_SDK_BUILTIN_ALWAYS` into `SDK_DISALLOWED_TOOLS` so both paths now dispatch through `mcp__copilot__web_search`.

**How.**

1. `copilot/config.py` — 2×2 model fields with `AliasChoices` preserving legacy env var names. `populate_by_name = True` so `ChatConfig(fast_standard_model=...)` works in tests.
2. `copilot/baseline/service.py::_resolve_baseline_model` — resolves the active baseline cell from `mode` + `tier`, no longer delegates to the SDK resolver.
3. `copilot/baseline/reasoning.py` — `BaselineReasoningEmitter` gains `_pending_delta` / `_last_flush_monotonic` and flushes on `len(_pending_delta) >= _COALESCE_MIN_CHARS` OR `monotonic() - _last_flush_monotonic >= _COALESCE_MAX_INTERVAL_MS / 1000`. `_is_reasoning_route` rewritten as an anchored prefix match covering `anthropic/`, `anthropic.`, `moonshotai/`, and `openrouter/kimi-` — split from the narrower `_is_anthropic_model` gate that still governs `cache_control` markers (which Kimi doesn't support).
4. `copilot/model.py::ChatSession` — `_inflight_tool_calls: set[str] = PrivateAttr(default_factory=set)` plus `announce_inflight_tool_call` / `clear_inflight_tool_calls` / `has_tool_been_called_this_turn`.
5. `copilot/tools/helpers.py::require_guide_read` — check `session.has_tool_been_called_this_turn(_AGENT_GUIDE_TOOL_NAME)` before falling back to scanning `session.messages`.
6. `copilot/tools/web_search.py` — new `WebSearchTool` + `_extract_results` + `_estimate_cost_usd`. `is_available` gated on `Settings().secrets.anthropic_api_key` so the deployment can roll back just by unsetting the key.
7. `copilot/tools/__init__.py` — registers `web_search` in `TOOL_REGISTRY` so it becomes `mcp__copilot__web_search` in the SDK path.
8. `copilot/sdk/tool_adapter.py` — `WebSearch` moves to `SDK_DISALLOWED_TOOLS`.

### Changes

- `copilot/config.py` — 2×2 model matrix with legacy env alias preservation; `populate_by_name=True`.
- `copilot/baseline/service.py::_resolve_baseline_model` — resolves against the new matrix.
- `copilot/baseline/reasoning.py` — `BaselineReasoningEmitter` coalescing buffer; `_is_reasoning_route` rewritten as anchored prefix match (covers `anthropic/`, `anthropic.`, `moonshotai/`, `openrouter/kimi-`).
- `copilot/model.py::ChatSession` — `_inflight_tool_calls` PrivateAttr + helpers.
- `copilot/baseline/service.py::_baseline_tool_executor` — calls `announce_inflight_tool_call` after emitting `StreamToolInputAvailable`; `clear_inflight_tool_calls` in the outer `finally` before persist.
- `copilot/tools/helpers.py::require_guide_read` — reads the new tracker first.
- `copilot/tools/web_search.py` (new) — Anthropic `web_search_20250305` wrapper + cost estimator.
- `copilot/tools/web_search_test.py` (new) — extractor / cost / dispatch / registry tests (12 total).
- `copilot/tools/models.py` — `WebSearchResponse` + `WebSearchResult` + `ResponseType.WEB_SEARCH`.
- `copilot/tools/__init__.py` — registers `web_search`.
- `copilot/sdk/tool_adapter.py` — moves native `WebSearch` to `SDK_DISALLOWED_TOOLS`.

### Checklist

For code changes:
- [x] Changes listed above
- [x] Test plan below
- [ ] Tested according to the test plan:
  - [x] `poetry run pytest backend/copilot/baseline/` — all pass
  - [x] `poetry run pytest backend/copilot/sdk/` — all pass (SDK resolver untouched)
  - [x] `poetry run pytest backend/copilot/tools/web_search_test.py` — 12 pass
  - [ ] Manual: send a multi-step prompt on fast mode with default config; confirm backend routes to `moonshotai/kimi-k2.6`, SSE stream carries `reasoning-start/delta/end` (coalesced), Reasoning collapse renders + survives hard reload.
  - [ ] Manual: 43-tool payload reliability on Kimi — watch for malformed tool-call JSON or wrong-tool selection.
  - [ ] Manual: `CHAT_FAST_STANDARD_MODEL=anthropic/claude-sonnet-4-6` restarts confirm Sonnet routing (rollback path works).
  - [ ] Manual: SDK path (`CHAT_USE_CLAUDE_AGENT_SDK=true`) still selects the SDK service and uses `thinking_standard_model` = Sonnet (no Kimi leaked into extended thinking).
  - [ ] Manual: prompt that forces `web_search` — confirm results render, `persist_and_record_usage(provider='anthropic')` runs, cost lands in the per-user ledger.
  - [ ] Manual: ask Kimi a question that would require `agent_building_guide` — confirm the guide loads exactly once per turn (no loop).

For configuration changes:
- [x] `.env.default` — all four model fields fall back to the pydantic defaults; legacy `CHAT_MODEL` / `CHAT_FAST_MODEL` / `CHAT_ADVANCED_MODEL` remain honored via `AliasChoices`.


autogpt_platform/backend/backend/copilot/baseline/transcript_integration_test.py
```diff
@@ -63,21 +63,40 @@ def _make_session_messages(*roles: str) -> list[ChatMessage]:
...
+    def test_sdk_and_baseline_standard_defaults_diverge(self):
+        """The whole point of the split: baseline cheap (Kimi) vs SDK
+        Anthropic-only (Sonnet).  If this equality ever flips they've
+        re-collapsed and someone lost the cost savings."""
+        assert config.model != config.fast_model
 
     def test_standard_and_advanced_models_differ(self):
         """Advanced tier defaults to a different (Opus) model than standard."""
-        assert config.model != config.advanced_model
+        assert config.fast_model != config.advanced_model
```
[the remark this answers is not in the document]
majdyz: Fixed in [`95d3679e1`](https://github.com/Significant-Gravitas/AutoGPT/pull/12871/commits/95d3679e1). Swapped all three tests from `config.<field>` (env-backed singleton) to `ChatConfig.model_fields["<field>"].default` so CI doesn't flake when `CHAT_FAST_MODEL=anthropic/claude-sonnet-4-6` rollback is set in the deploy env. Tests still pin the shipped default, just via the declared Field.

autogpt_platform/backend/backend/copilot/model.py
```diff
@@ -242,6 +251,44 @@ def from_db(cls, prisma_session: PrismaChatSession) -> Self:
...
+        this path much more than Sonnet does).  The buffer is cleared by
+        :meth:`clear_inflight_tool_calls` at turn end.
+        """
+        self._inflight_tool_calls.add(tool_name)
+
+    def clear_inflight_tool_calls(self) -> None:
+        """Reset the in-flight tool-call announcement buffer."""
+        self._inflight_tool_calls.clear()
+
+    def has_tool_been_called_this_turn(self, tool_name: str) -> bool:
```
majdyz: 🤖 🟡 **Nice to Have**: Method name is broader than it sounds. `has_tool_been_called_this_turn` actually scans *all* of `self.messages` (not scoped to the current turn), per the `test_guide_earlier_in_history_still_passes` contract — it returns True for a guide call in any prior turn. The in-flight buffer layer is the only genuinely turn-scoped part. Consider renaming to `has_tool_been_called_in_session` or `has_tool_been_called` to avoid future callers assuming per-turn semantics (e.g. "did the model just decide to call X in this user turn?"). The current name actively encourages that misread.

autogpt_platform/backend/backend/copilot/baseline/service.py
```diff
@@ -761,6 +764,19 @@ async def _baseline_tool_executor(
...
+    # actually runs.  Without this, the tool_call row lives only in
+    # ``state.session_messages`` until the ``finally`` block flushes it
+    # into ``session.messages`` at turn end — so a second tool in the
+    # same turn (e.g. ``create_agent`` after ``get_agent_building_guide``)
+    # scans a stale ``session.messages`` and the guard re-fires despite
+    # the guide having been called.  The announce-set is cleared at turn
+    # end; we deliberately don't touch ``session.messages`` here to avoid
+    # duplicating the assistant row that ``_baseline_conversation_updater``
+    # will append at round end.
+    session.announce_inflight_tool_call(tool_name)
```
majdyz, line 778: 🤖 🟡 **Nice to Have**: The announce fires *before* `execute_tool` runs (line 778) and is never rolled back if the tool call itself raises. That's probably fine for the guide-read gate ("was it called?" is the right question), but worth a sentence in the docstring: currently `announce_inflight_tool_call` reads as "about to dispatch" while `has_tool_been_called_this_turn` reads as "ran successfully". If they ever diverge — e.g. a future gate that only wants successful dispatches — the naming won't help the next reader see the mismatch.

autogpt_platform/backend/backend/copilot/baseline/reasoning.py
```diff
@@ -192,39 +249,77 @@ def on_delta(self, delta: ChoiceDelta) -> list[StreamBaseResponse]:
...
         if self._current_row is not None:
             self._current_row.content = (self._current_row.content or "") + text
+
+        self._pending_delta += text
+        if self._should_flush_pending():
+            events.append(
+                StreamReasoningDelta(id=self._block_id, delta=self._pending_delta)
+            )
+            self._pending_delta = ""
+            self._last_flush_monotonic = time.monotonic()
```
majdyz: 🤖 🔵 **Nit**: `time.monotonic()` is called twice per buffered chunk (once in `_should_flush_pending`, then again on flush at line 290). At the ~4,700-chunks-per-turn scale this fix is motivated by, that's ~9,400 monotonic calls per turn — still cheap, but you already compute `elapsed_ms` in the check; passing it back out (or caching the `time.monotonic()` value in a local) would eliminate the redundant call. Skip if not worth the churn — pytest covers it either way.

autogpt_platform/backend/backend/copilot/baseline/reasoning.py
```diff
@@ -132,18 +132,36 @@ def visible_text(self) -> str:
...
+    because ``cache_control`` is strictly Anthropic-specific (Moonshot does
+    its own auto-caching), so the two gates must not conflate.
+    """
+    lowered = model.lower()
+    return (
+        "claude" in lowered
+        or lowered.startswith("anthropic")
+        or lowered.startswith("moonshotai/")
+        or "kimi" in lowered
+    )
```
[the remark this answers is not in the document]
majdyz: 🤖 Actually fixed in [9cfaaba3b](https://github.com/Significant-Gravitas/AutoGPT/commit/9cfaaba3b68f8a5322a50221859e4af10a8fcbca). The bot's "Resolved in da5420f" claim was premature — that earlier commit only coalesced reasoning deltas and didn't touch `_is_reasoning_route`. This follow-up swaps the `"kimi" in lowered` substring check for an anchored match: `moonshotai/` provider prefix, or a `kimi-` model id at string start / immediately after a `/` provider segment. Added regression tests for `hakimi`, `some-provider/hakimi-large`, `akimi-7b` to lock the anchor down.

autogpt_platform/backend/backend/copilot/baseline/transcript_integration_test.py
```diff
@@ -63,21 +63,73 @@ def _make_session_messages(*roles: str) -> list[ChatMessage]:
...
+    def test_fast_advanced_default_is_opus(self):
+        """Shipped default: Opus on the baseline advanced cell — mirrors
+        the SDK advanced cell so the advanced-tier A/B stays clean
+        (same model, different path)."""
+        from backend.copilot.config import ChatConfig
+
+        assert (
+            ChatConfig.model_fields["fast_advanced_model"].default
+            == "anthropic/claude-opus-4-7"
+        )
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [4dc3d0c34](https://github.com/Significant-Gravitas/AutoGPT/commit/4dc3d0c34abac0d2019820aceb04d1a00b41bbc8). Updated both `fast_advanced_model` and `thinking_advanced_model` defaults in `backend/copilot/config.py` from `anthropic/claude-opus-4-7` to `anthropic/claude-opus-4.7` to match OpenRouter's actual route, and updated the pinned-default test accordingly. Also added a legacy-env-alias regression test (`test_legacy_env_aliases_route_to_new_fields`) to lock down the `CHAT_MODEL` / `CHAT_ADVANCED_MODEL` / `CHAT_FAST_MODEL` → new-field mapping.

autogpt_platform/backend/backend/copilot/model.py
majdyz: 🤖 Fixed in [7dc3b880a](https://github.com/Significant-Gravitas/AutoGPT/commit/7dc3b880a6b098630baaeb2d5c8732eeaf7874e9). Renamed `has_tool_been_called_this_turn` → `has_tool_been_called` and expanded the docstring to call out that the durable-messages branch is session-wide (matching the guide-read contract from `test_guide_earlier_in_history_still_passes`), while only the in-flight buffer is genuinely turn-scoped. Updated the sole caller (`require_guide_read`) and the test-file docstring reference.

autogpt_platform/backend/backend/copilot/baseline/service.py
majdyz: 🤖 Fixed in [7dc3b880a](https://github.com/Significant-Gravitas/AutoGPT/commit/7dc3b880a6b098630baaeb2d5c8732eeaf7874e9). Added a paragraph at the top of `announce_inflight_tool_call`'s docstring spelling out that the announcement fires *before* `execute_tool` runs and isn't rolled back if the tool raises — and that this matches the guide-read gate's "was it called?" semantics, but a future gate wanting *successful* dispatches would need its own tracking.

autogpt_platform/backend/backend/copilot/baseline/reasoning.py
majdyz: 🤖 Fixed in [7dc3b880a](https://github.com/Significant-Gravitas/AutoGPT/commit/7dc3b880a6b098630baaeb2d5c8732eeaf7874e9). Folded the two `time.monotonic()` calls per chunk into one: the caller samples `now` once, passes it into `_should_flush_pending(now)`, and reuses the same value when updating `_last_flush_monotonic` on flush. Saves ~4,700 syscalls per Kimi turn. Existing `test_time_based_flush_when_chars_stay_below_threshold` passes unchanged since both branches still read through the monkeypatched `time.monotonic`.

autogpt_platform/backend/backend/copilot/prompting.py
```diff
@@ -411,3 +411,65 @@ def get_graphiti_supplement() -> str:
...
+## Web Search & URL Fetch (fast mode)
+
+Fast mode doesn't expose a native web-search tool — use the existing
+copilot blocks via ``run_block`` when you need live web content.
+
+### Web search with citations — Perplexity
+- Block ID: ``{PERPLEXITY_BLOCK_ID}``
+- Input: ``{{"prompt": "<query>", "model": "sonar"}}``
+  (``model`` defaults to ``sonar``; other options: ``sonar-pro``,
+  ``sonar-reasoning``, ``sonar-reasoning-pro``, ``sonar-deep-research``
```
majdyz: 🤖 🟠 **Should Fix**: The model IDs in this supplement are wrong and silently degrade — this is exactly the "ship a dead identifier to the model" footgun that the block-ID registry test was added to prevent, but it was only wired for the block UUID and not the input enum.

`PerplexityModel` has three values: `"perplexity/sonar"`, `"perplexity/sonar-pro"`, `"perplexity/sonar-deep-research"`. The supplement tells the model to pass `"sonar"` / `"sonar-pro"` / `"sonar-reasoning"` / `"sonar-reasoning-pro"`, **none of which match**. Verified locally:

```
bare 'sonar' → falls back to perplexity/sonar (log WARNING)
'sonar-pro' → falls back to perplexity/sonar
'sonar-reasoning' → falls back to perplexity/sonar   # doesn't exist in enum at all
All enum values: ['perplexity/sonar', 'perplexity/sonar-pro', 'perplexity/sonar-deep-research']
```

Two real consequences:
1. Every baseline Perplexity call logs an `Invalid PerplexityModel` WARNING that is not actually a bug in the user's input — it's in our own supplement.
2. "sonar-pro", "sonar-reasoning", "sonar-reasoning-pro" all silently collapse to plain `sonar` — users asking for deeper research never get it and the model has no way to know.

Fix options: (a) rewrite the supplement to use the enum values verbatim (`"perplexity/sonar"`, etc.) and drop the two reasoning variants that don't exist, (b) extend `PerplexityModel` with the reasoning SKUs if they're actually supported upstream. Either way, please also mirror the `TestBaselineWebSearchSupplement` pattern by pinning each referenced model id against `PerplexityModel` values so the next rename/drop catches it at test time.

majdyz: 🤖 Fixed in [54d6d4a3e](https://github.com/Significant-Gravitas/AutoGPT/commit/54d6d4a3e61da120490317761e2bde0360461a77). Rewrote `get_baseline_web_search_supplement` to render the model list directly from `PerplexityModel` at call time, so the supplement always matches the enum verbatim. Default example is now `"model": "perplexity/sonar"` (provider-prefixed, passes the validator without the fallback warning). Added two regression tests:

1. `test_supplement_uses_perplexitymodel_enum_values_verbatim` — asserts every enum value surfaces in the rendered text and the default example carries the full prefix.
2. `test_supplement_does_not_mention_invented_sonar_variants` — rejects the old bare `"sonar"`/`"sonar-pro"` and the nonexistent `sonar-reasoning` strings so nobody reintroduces them.

Rather than hardcoding more reasoning SKUs that aren't in the enum, the supplement now loops over `PerplexityModel` directly — upstream adding or dropping a SKU flows through automatically.

autogpt_platform/backend/backend/copilot/config.py
```diff
@@ -27,21 +31,66 @@
...
+        "``CHAT_MODEL`` still honored).",
+    )
+    thinking_advanced_model: str = Field(
+        default="anthropic/claude-opus-4.7",
+        validation_alias=AliasChoices(
+            "CHAT_THINKING_ADVANCED_MODEL",
+            "CHAT_ADVANCED_MODEL",
+        ),
+        description="SDK (extended-thinking) path, 'advanced' tier.  Opus "
+        "by default.  Override via ``CHAT_THINKING_ADVANCED_MODEL`` "
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [e48144b35](https://github.com/Significant-Gravitas/AutoGPT/commit/e48144b356a76b8b70a6d96eb5d3a7ea4882e9f6). Verified locally that the env var does bind today via `env_prefix = "CHAT_"` (contrary to the bug claim), but `fast_advanced_model` was the only cell without an explicit `validation_alias` — which means dropping `env_prefix` in a future refactor would silently unbind it. Added `validation_alias=AliasChoices("CHAT_FAST_ADVANCED_MODEL")` for consistency with the other three cells, plus a new regression test `test_all_four_new_env_vars_bind_to_their_cells` that pins each cell against its documented env var — that pairs with `test_legacy_env_aliases_route_to_new_fields` to cover the contract from both sides.

autogpt_platform/backend/backend/copilot/baseline/reasoning.py
```diff
@@ -17,12 +18,14 @@
   one streaming round and emits ``StreamReasoning*`` events so the caller
   only has to plumb the events into its pending queue.
 * :func:`reasoning_extra_body` builds the ``extra_body`` fragment for the
-  OpenAI client call.  Returns ``None`` on non-Anthropic routes.
+  OpenAI client call.  Returns ``None`` for routes without reasoning
+  support (see :func:`_is_reasoning_route`).
 """
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [059180427](https://github.com/Significant-Gravitas/AutoGPT/commit/059180427282793a52f58b77e85e889291e7ca13). Sentry's right — `"claude" in lowered` was broad in the same way the Kimi branch was before anchoring. Tightened the gate: accept `anthropic`/`anthropic.` prefixes, `moonshotai/` prefix, and bare `claude-`/`kimi-` model ids with no provider prefix; any foreign `/` prefix (`someprovider/claude-mock-v1`, `other/kimi-pro`) now falls through to False. Explicit carve-out: `openrouter/kimi-` stays recognised to preserve the K2.6 route behaviour that earlier tests locked in. Added `test_claude_substring_false_positives_rejected` covering both the Claude and Kimi variants.

autogpt_platform/backend/backend/copilot/prompting.py
majdyz: 🤖 This thread is obsolete — the underlying web-search supplement commit (1848810b3) and my follow-up fix (54d6d4a3e) were both reverted in [0d8a27fb7](https://github.com/Significant-Gravitas/AutoGPT/commit/0d8a27fb7a7f7ce7c861a6942098e0552ab58694) and [c9a86e833](https://github.com/Significant-Gravitas/AutoGPT/commit/c9a86e83390dfdf67af0c13ae2dfe2c346a904a6), so `get_baseline_web_search_supplement` and the invented sonar IDs it shipped are no longer in the PR. The concern I raised here doesn't apply to the current HEAD.

autogpt_platform/backend/backend/copilot/tools/web_search.py
```diff
@@ -0,0 +1,221 @@
...
+                results.append(
+                    WebSearchResult(
+                        title=getattr(item, "title", "") or "",
+                        url=getattr(item, "url", "") or "",
+                        snippet=getattr(item, "encrypted_content", None)
+                        or getattr(item, "page_content", "")
+                        or "",
+                        page_age=getattr(item, "page_age", None),
+                    )
+                )
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [1dfc7552](https://github.com/Significant-Gravitas/AutoGPT/commit/1dfc75520). Good catch — `encrypted_content` is Anthropic's opaque citation blob, not a readable snippet, and there is no plain-text snippet field in the current `web_search_20250305` beta. Dropped the fallback entirely (`snippet=""`) and rely on the model's `text` blocks with citations for prose. Updated the test to lock this in (encrypted blob in → empty snippet out).

autogpt_platform/backend/backend/copilot/baseline/service.py
```diff
@@ -1900,6 +1881,10 @@ def _trim_openai_on_rollback(_session_anchor: int) -> None:
                 final_text = final_text[len(recorded) :]
         if final_text.strip():
             session.messages.append(ChatMessage(role="assistant", content=final_text))
+        # In-flight tool-call announcements are only meaningful for the
+        # current turn; clear before persist so the next turn starts with
+        # a clean scratch buffer.
+        session.clear_inflight_tool_calls()
```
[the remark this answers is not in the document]
majdyz: Already in place — `session.clear_inflight_tool_calls()` is the very first statement in the outer `finally` (service.py L1824-1833 on the current HEAD 1dfc75520), with an inline comment calling out the "run before any awaited cleanup can fail" invariant you flagged. The older placement (after `persist_and_record_usage`) was moved in the earlier coalesced-reasoning commits on this branch.

majdyz:
🤖 Addressed the outside-diff nitpicks from the last three CodeRabbit reviews in [215662268](https://github.com/Significant-Gravitas/AutoGPT/commit/215662268f1a41fb185ea8f450fddb2cdaf90374):

- **`config.py:195-199`** (claude_agent_model docstring): updated fallback reference from the removed `model` field to `thinking_standard_model`.

Already addressed on HEAD before the reviews were posted (no new commit needed):
- **`config.py:20-24`** (CopilotLlmModel comment): 7dc3b880a/3bc28ac69 already rewrote the tier comment to spell out `fast_standard_model` on baseline vs `thinking_standard_model` on SDK.
- **`reasoning.py:19-21`** (module docstring): already widened on HEAD to "Returns `None` for routes without reasoning support (see `_is_reasoning_route`)." — the Anthropic-only wording is gone.

PR #12872, opened 2026-04-21
fix(backend/copilot): stop flag-gating Graphiti supplement in system prompt

majdyz:
### Why / What / How

**Why:** Baseline + SDK prompt caching was forking per LaunchDarkly cohort. Cross-user test on dev stack (post-#12870 merge):

```
9:17:28 PM  gmail         Baseline Opus    17.8K in / 0 cached / 17.2K write
9:18:55 PM  example.com   Baseline Opus     5.6K in / 11.6K cached / 5.1K write  ← partial hit
9:17:40 PM  example.com   Baseline Sonnet  12.8K in / 0 cached / 12.4K write
9:19:41 PM  gmail         Baseline Sonnet   4.6K in / 8.6K cached / 4.2K write   ← partial hit
```

Tools block hits across users (11.6K cached on the same-model second call), but the system block misses every time — ~5K / ~500 tokens get rewritten. That cluster matches exactly the 1,816-char `get_graphiti_supplement()` text that was conditionally appended to the system prompt when the `GRAPHITI_MEMORY` flag was true for the user. Flag-on vs flag-off users each ended up with their own cache cohort and burned a fresh write on every cross-cohort first call.

Also breaks the documented invariant from `_fetch_langfuse_prompt()` at [service.py:297-330](autogpt_platform/backend/backend/copilot/service.py#L297-L330): "the prompt text is identical across all users (enabling cross-session caching)."

**What:** Always append the Graphiti memory-tool instructions to the system prompt. The tools themselves already self-gate at handler level (`is_enabled_for_user` in each `graphiti_*` tool, e.g. [graphiti_search.py:89](autogpt_platform/backend/backend/copilot/tools/graphiti_search.py#L89)), so flag-off users never trigger memory side effects even if the model attempts to call them.

**How:**
- `baseline/service.py:1390` — drop the `if graphiti_enabled else ""` ternary; always include the supplement. Kept the `graphiti_enabled` flag because warm-context fetch + ingest still gate on it (correct: those operate on per-user *data*, not static instructions).
- `sdk/service.py:2958` — same treatment.
- New `prompt_static_invariant_test.py` — source-level regression guard that fails if either call site re-introduces the flag-gated pattern.

Trade-off: ~500 tokens extra in the system prompt for flag-off users. Amortised in one cache write — cross-user cache now hits 100% of the prefix instead of ~67%.

### Changes

- `copilot/baseline/service.py` — unconditional Graphiti supplement append
- `copilot/sdk/service.py` — same
- `copilot/prompt_static_invariant_test.py` — new regression guard

### Checklist

For code changes:
- [x] I have made a test plan
  - [x] `poetry run pytest backend/copilot/prompt_static_invariant_test.py backend/copilot/prompt_cache_test.py backend/copilot/baseline/` — 115 pass
  - [ ] Manual: on a dev deploy with two users (one flag-on, one flag-off), second user's first Baseline call shows `cached_tokens ≈ full prefix` instead of `~67% cached + 5K write`.
  - [ ] Manual: flag-off user can send a prompt that would normally benefit from memory — confirm the `memory_search` tool handler returns "disabled" gracefully, no crash.
  - [ ] Manual: `CHAT_USE_CLAUDE_AGENT_SDK=true` SDK path — same cross-user cache improvement, no regression in SDK session persistence.

### DIFF
--- a/autogpt_platform/backend/backend/copilot/baseline/service.py
+++ b/autogpt_platform/backend/backend/copilot/baseline/service.py
@@ -1384,11 +1384,18 @@ async def stream_chat_completion_baseline(
 
     message_id = str(uuid.uuid4())
 
-    # Append tool documentation, technical notes, and Graphiti memory instructions
+    # The system prompt MUST be identical across users — the cacheable prefix
+    # ends at the ``cache_control`` breakpoint on this block, so any per-user
+    # content here forks the cache by user cohort and each cohort pays for
+    # its own write.  The Graphiti memory instructions are therefore always
+    # appended: memory tools are in the registry for every user and gate
+    # themselves at handler level (``is_enabled_for_user`` in each
+    # graphiti_* tool), so flag-off users get a "memory disabled" response
+    # from the handler instead of a divergent system prompt.  Warm context
+    # and ingest still honor the flag below — those are per-user data, not
+    # per-user instructions.
     graphiti_enabled = await is_enabled_for_user(user_id)
-
-    graphiti_supplement = get_graphiti_supplement() if graphiti_enabled else ""
-    system_prompt = base_system_prompt + SHARED_TOOL_NOTES + graphiti_supplement
+    system_prompt = base_system_prompt + SHARED_TOOL_NOTES + get_graphiti_supplement()
 
     # Warm context: pre-load relevant facts from Graphiti on first turn.
     # Use the pre-drain count so pending messages drained at turn start
--- a/autogpt_platform/backend/backend/copilot/prompt_static_invariant_test.py
+++ b/autogpt_platform/backend/backend/copilot/prompt_static_invariant_test.py
@@ -1,0 +1,61 @@
+"""Regression guards: the baseline + SDK system prompts must stay identical
+across users regardless of the ``GRAPHITI_MEMORY`` LaunchDarkly flag.
+
+Background: the Langfuse prompt is compiled with ``users_information=""`` and
+both paths append ``SHARED_TOOL_NOTES`` / ``get_sdk_supplement()`` plus the
+Graphiti memory instructions.  A previous bug flag-gated the supplement on
+``is_enabled_for_user(user_id)`` — which meant flag-on users and flag-off
+users got different system-prompt byte sequences, forking the Anthropic
+prompt cache by cohort and burning a fresh ~500-token cache write on every
+cross-cohort first call.
+
+These tests pin the contract at the source level so the conditional never
+silently returns.  A richer end-to-end test would be nice but would require
+exercising the entire streaming pipeline; for a contract this localised the
+source assertion is tighter and costs nothing to run.
+"""
+
+import inspect
+import re
+
+_GATED_SUPPLEMENT_PATTERNS = [
+    # Pattern 1: direct ternary on graphiti_enabled
+    re.compile(r"get_graphiti_supplement\(\)\s+if\s+graphiti_enabled"),
+    # Pattern 2: variable-assignment ternary
+    re.compile(
+        r"graphiti_supplement\s*=\s*get_graphiti_supplement\(\)\s+if\s+\w+\s+else\s+\"\""
+    ),
+]
+
+
+def _assert_supplement_always_appended(src: str, path: str) -> None:
+    for pat in _GATED_SUPPLEMENT_PATTERNS:
+        assert not pat.search(src), (
+            f"{path}: Graphiti supplement is flag-gated again — this forks the "
+            f"Anthropic prompt cache per-user cohort.  Always append the "
+            f"supplement; memory tools self-gate at handler level."
+        )
+    assert "get_graphiti_supplement()" in src, (
+        f"{path}: Graphiti supplement call is missing — the memory instructions "
+        f"must be present so every user's system prompt hashes identically."
+    )
+
+
+def test_baseline_graphiti_supplement_unconditionally_appended():
+    from backend.copilot.baseline import service
+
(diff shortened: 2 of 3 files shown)


autogpt_platform/backend/backend/copilot/prompt_static_invariant_test.py
```diff
@@ -0,0 +1,61 @@
...
+import inspect
+import re
+
+_GATED_SUPPLEMENT_PATTERNS = [
+    # Pattern 1: direct ternary on graphiti_enabled
+    re.compile(r"get_graphiti_supplement\(\)\s+if\s+graphiti_enabled"),
+    # Pattern 2: variable-assignment ternary
+    re.compile(
+        r"graphiti_supplement\s*=\s*get_graphiti_supplement\(\)\s+if\s+\w+\s+else\s+\"\""
+    ),
```
majdyz, line 27: 🤖 🟡 **Nice to Have**: Pattern 2 regex hardcodes `else ""` (double-quoted empty string). A regression using `else `, `else None`, or `else ("" if X else "")` would slip past. Consider loosening to `re.compile(r"get_graphiti_supplement\(\)\s+if\s+\w+")` — any ternary on a name around the call is enough signal that gating is back. The current tight pattern risks being too narrow for its stated purpose.

autogpt_platform/backend/backend/copilot/prompt_static_invariant_test.py
```diff
@@ -0,0 +1,61 @@
...
+
+def test_sdk_graphiti_supplement_unconditionally_appended():
+    from backend.copilot.sdk import service
+
+    # SDK path: find the streaming entrypoint — the supplement assembly sits
+    # inside one of a handful of top-level streaming coroutines.  We check all
+    # async def declarations in the module so the guard doesn't rely on a
+    # stable function name.
+    src = inspect.getsource(service)
+    _assert_supplement_always_appended(src, "sdk/service.py")
```
majdyz, line 61: 🤖 🟡 **Nice to Have**: `test_sdk_graphiti_supplement_unconditionally_appended` uses `inspect.getsource(service)` on the whole module. That scope contains the import line (`from ..prompting import get_graphiti_supplement, get_sdk_supplement`), plus any future commentary or dead helper. The `assert "get_graphiti_supplement()" in src` check would pass even if someone deleted the actual call inside the streaming coroutine but left a commented-out reference or a string literal in a docstring. Narrowing to the specific streaming coroutine (e.g. find the function by locating the `system_prompt = (` assignment, or hard-code the function name like the baseline test does) would make the guard actually guard what the docstring claims. The comment `# we check all async def declarations` says the intent was to iterate over async defs, but the code simply greps the whole module source.

autogpt_platform/backend/backend/copilot/baseline/service.py
```diff
@@ -1384,11 +1384,18 @@ async def stream_chat_completion_baseline(
...
+    # ends at the ``cache_control`` breakpoint on this block, so any per-user
+    # content here forks the cache by user cohort and each cohort pays for
+    # its own write.  The Graphiti memory instructions are therefore always
+    # appended: memory tools are in the registry for every user and gate
+    # themselves at handler level (``is_enabled_for_user`` in each
+    # graphiti_* tool), so flag-off users get a "memory disabled" response
+    # from the handler instead of a divergent system prompt.  Warm context
+    # and ingest still honor the flag below — those are per-user data, not
+    # per-user instructions.
     graphiti_enabled = await is_enabled_for_user(user_id)
```
majdyz, line 1397: 🤖 🔵 **Nit**: The `graphiti_enabled = await is_enabled_for_user(user_id)` call is now computed before building `system_prompt` but no longer consulted for that build. Reads slightly oddly at the callsite — consider either (a) moving the `await is_enabled_for_user(...)` down next to the `if graphiti_enabled and user_id and _pre_drain_msg_count <= 1:` warm-context check where it is actually read, or (b) a one-liner comment `# graphiti_enabled only gates warm-context + ingest below; the prompt stays static across cohorts`. The big block comment above already covers this, so feel free to pass — but the ordering is the one thing most likely to trip a future reader into re-introducing the gate.

autogpt_platform/backend/backend/copilot/sdk/service.py
```diff
@@ -2953,13 +2953,19 @@ async def _setup_e2b():
...
         # Append appropriate supplement (Claude gets tool schemas automatically)
 
+        # Always append the Graphiti memory instructions so the system prompt
+        # — the cacheable prefix on the SDK path — stays identical across
+        # users regardless of the ``GRAPHITI_MEMORY`` flag.  Memory tools
+        # self-gate at handler level (``is_enabled_for_user`` in each
+        # graphiti_* tool), so flag-off users never see memory side effects.
+        # Warm context + ingest still honor the flag below — those operate
+        # on per-user data, not on the static instructions.
         graphiti_enabled = await is_enabled_for_user(user_id)
```
majdyz, line 2963: 🤖 🔵 **Nit**: Same ordering note as the baseline file — `graphiti_enabled` is computed here but only read below for warm context. Keeping it adjacent to its consumer would make the invariant "system prompt never branches on user state" visually obvious. Non-blocking.

majdyz:
Closing — per-user Graphiti flag-gating creates two cache cohorts (flag-on / flag-off) but each cohort caches to 100% within itself. Extra cost is one warmup write per cohort per 1h TTL, which is negligible. Keeping flag-off users' system prompt clean (no memory instructions they can't use) is the cleaner semantic. /cc @majdyz

PR #12880, opened 2026-04-22
feat(backend/copilot): inline picker-backed inputs via run_block + accept AgentInputBlock subclasses

### LINKED ISSUE #12875: CoPilot agent-builder should use AgentGoogleDriveFileInputBlock for Drive files, not hardcode IDs
# CoPilot agent-builder should use `AgentGoogleDriveFileInputBlock` for Drive files, not hardcode IDs

## Summary

When a user asks CoPilot to build an agent that reads a Google Sheets/Docs/Drive file, the CoPilot agent-builder hardcodes the file ID into the consuming block's `constantInput` instead of wiring an `AgentGoogleDriveFileInputBlock` as the agent's input. The resulting agent either fails graph validation (bare string rejected by the `GoogleDriveFile` schema) or, if it passes validation, fails at execution time because no `_credentials_id` is attached to reach the user's Google OAuth credential.

Platform-side we're adding a validator that blocks this anti-pattern with a clear error. The agent-builder model needs updated guidance so it produces the correct shape in the first place rather than hitting the validator and retrying.

## Evidence

A beta user hit this repeatedly. Over **13 saved versions** of one agent (graph `3477e4ec-08b2-4ca5-bffe-ca2729fbc36e`), CoPilot produced the broken shape every time:

| Versions | What CoPilot put in `GoogleSheetsReadBlock.constantInput.spreadsheet` | Outcome |
|---|---|---|
| 4–6 | `"SHEETS_ID_BURAYA"` (Turkish placeholder) | Pydantic rejects: `is not of type 'object'` |
| 7–10 | `"1KAv8hhChef7a5ycn6Al1M4DdkiG_PVcKQ_tYkRpGA-I"` (bare ID) | Same Pydantic error |
| 11–13 | `{"id": "1KAv8hhChef7a5ycn6Al1M4DdkiG_PVcKQ_tYkRpGA-I"}` | Passes save-time validation but **no `_credentials_id`, no `mimeType`** — runtime `_validate_spreadsheet_file` or `_acquire_auto_credentials` fails |

Same anti-pattern on `GoogleDocsReadBlock.constantInput.document` and `GoogleSheetsUpdateCellBlock.constantInput.spreadsheet` in the same graph. `AgentGoogleDriveFileInputBlock` appears in **zero** of the 13 versions — even though CoPilot successfully looked it up via `find_block` at 2026-04-20T17:40:02 earlier in the session. It knew the block existed; it just didn't use it.

The beta user pasted the pydantic error directly into chat (Langfuse trace `2f9b9b8916` at `2026-04-20T19:51:29`): *"`GoogleSheetsReadBlock`: '1KAv…' is not of type 'object' Failed validating 'type' in schema['properties']['spreadsheet']"* — and even with that explicit error, CoPilot still didn't switch to the input block.

## What's wrong

For any block whose input schema uses `GoogleDriveFileField(...)` (e.g. `GoogleSheetsReadBlock.spreadsheet`, `GoogleDocsReadBlock.document`), the input is typed as a `GoogleDriveFile` **object**, not a string. The object must include:

- `id` — the file/folder ID (required)
- `mimeType` — consumed by per-block validators like `_validate_spreadsheet_file` which check MIME before calling the Drive API
- `_credentials_id` — **the critical field**: this is how the graph executor (`_acquire_auto_credentials` in `backend/executor/manager.py`) knows which of the user's stored Google OAuth credentials to authenticate with at run time. Hardcoded file IDs don't carry credentials — even if the schema accepted a string, the block couldn't authenticate.

CoPilot is substituting a plain `AgentInputBlock` (generic text/value input) for the Drive case instead of the domain-specific `AgentGoogleDriveFileInputBlock` that exists specifically to populate this object from a run-time Drive picker.

## The right pattern

The platform already ships `AgentGoogleDriveFileInputBlock` (block ID `d3b32f15-6fd7-40e3-be52-e083f51b19a2`, defined at `autogpt_platform/backend/backend/blocks/io.py:655`). It:

1. Renders a Google Drive picker when the user runs the agent
2. Outputs a fully-hydrated `GoogleDriveFile` (id + name + mimeType + `_credentials_id`) via its `result` output
3. Lets whoever runs the agent pick their own file with their own credentials — so sharing/cloning the agent Just Works (credential fork isolation is already handled platform-side via `_reassign_ids` clearing `_credentials_id`)

**Target shape for any agent that reads/writes a Drive file:**

```
[AgentGoogleDriveFileInputBlock]  ──result→  [GoogleSheetsReadBlock].spreadsheet
                                   (or)  ──result→  [GoogleDocsReadBlock].document
                                   (or)  ──result→  [GoogleSheetsUpdateCellBlock].spreadsheet
```

CoPilot should:

- Detect when the agent's required workflow touches a Drive-file field (Sheets/Docs/Slides/Drive blocks with a `GoogleDriveFileField`).
- Add an `AgentGoogleDriveFileInputBlock` node with appropriate `allowed_views` (e.g. `["SPREADSHEETS"]` for Sheets blocks, `["DOCUMENTS"]` for Docs blocks, `["PRESENTATIONS"]` for Slides blocks) and a descriptive `name` / `title` (e.g. `"spreadsheet_input"`, `"Select Spreadsheet"`).
- Wire a link from `AgentGoogleDriveFileInputBlock.result` → `GoogleSheetsReadBlock.spreadsheet` (or the equivalent sink field on whichever Drive-consuming block).
- **NEVER** set `constantInput.<drive_field>` on the consuming block. Leave that field unset and let the link feed it at run time.
- If the user gives a specific Drive URL in chat (e.g. `docs.google.com/spreadsheets/d/…`), CoPilot should still use the input block, and can either (a) explain that the user will pick the file at run time, or (b) include the URL in the `AgentGoogleDriveFileInputBlock`'s default value metadata for UX — but the `_credentials_id` must come from the picker, so a pre-set value alone isn't enough.

## How to detect this is the case

The target block has a field where `CredentialsFieldInfo.is_auto_credential` is true (set by `GoogleDriveFileField`). At the schema level, look for:

- Field `format == "google-drive-picker"` in the block input schema, OR
- Field schema has an `auto_credentials` extra key, OR
- `BlockSchema.get_auto_credentials_fields()` returns a non-empty dict for the block.

Any of those is a signal that the field needs an `AgentGoogleDriveFileInputBlock` upstream, not a hardcoded `constantInput` value.

## Acceptance criteria

1. Given the user prompt *"build an agent that reads this Google Sheet: `<URL>`, summarizes it, and emails me"*, CoPilot produces a graph that contains:
   - An `AgentGoogleDriveFileInputBlock` node (id `d3b32f15-…`) with `allowed_views: ["SPREADSHEETS"]`, a sensible `name`/`title`, no `value` hardcoded to the URL's ID.
   - A `GoogleSheetsReadBlock` node whose `constantInput` does **NOT** contain a `spreadsheet` key.
   - A link from the input block's `result` output to the Sheets block's `spreadsheet` input.
2. Same for `GoogleDocsReadBlock.document` and `GoogleSheetsUpdateCellBlock.spreadsheet`.
3. The graph passes both the existing schema validation AND the new platform-side validator rule (see companion platform PR — link TBD).
4. Offline eval: run the golden-prompt set through the agent-builder, assert zero hardcoded `constantInput.<drive_field>` values across all produced graphs.
5. Regression: the generic `AgentInputBlock` is still chosen when the user wants a text/number input — the change is Drive-file-specific.

## Platform-side companion change

We are landing a graph validator that returns a human-readable error for this exact anti-pattern (blocks save/publish until fixed). The error message mentions `AgentGoogleDriveFileInputBlock` by name so that if CoPilot hits the validator, the remediation path is unambiguous in the error it sees. But the validator is a safety net, not a replacement for teaching CoPilot the correct pattern in the first place — we don't want the agent-builder looping on trial-and-error through validator errors.

## Blocker discovered while testing the fix — agent-generator validator rejects AgentInputBlock subclasses

When testing this on dev (CoPilot session `571b9e41-00e6-4204-bd21-da557f68bfe6`, Langfuse trace `aef2273f79749de25e614db591d6638d`), we explicitly asked CoPilot to swap its `AgentInputBlock` for `AgentGoogleDriveFileInputBlock`. It tried, and the agent-generator validator rejected the save with:

```
validation_failed
{
  "errors": [
    "Agent is missing an AgentInputBlock (block_id: 'c0a8e994-ebf1-4a9c-a4d8-89d09c86741b'). Every agent must have at least one AgentInputBlock to define user-facing inputs. Add a node with block_id 'c0a8e994-ebf1-4a9c-a4d8-89d09c86741b' and set input_default with 'name' and optionally 'title'."
  ]
}
```

Source: `backend/copilot/tools/agent_generator/validator.py:658-689` (`validate_io_blocks`). The check is:

```python
if AGENT_INPUT_BLOCK_ID not in block_ids:
    self.add_error(...)
```

where `AGENT_INPUT_BLOCK_ID = "c0a8e994-ebf1-4a9c-a4d8-89d09c86741b"` is the literal base-class `AgentInputBlock` id in `agent_generator/helpers.py:40`.

But `AgentGoogleDriveFileInputBlock` is a Python subclass of `AgentInputBlock` (see `backend/blocks/io.py:655` — `class AgentGoogleDriveFileInputBlock(AgentInputBlock):`), with a different block id. Same story for `AgentShortTextInputBlock`, `AgentLongTextInputBlock`, `AgentNumberInputBlock`, `AgentDateInputBlock`, `AgentTimeInputBlock`, `AgentFileInputBlock`, `AgentDropdownInputBlock`, `AgentToggleInputBlock`, `AgentTableInputBlock` — every specialized input block (see `IO_BLOCK_IDs` list in `io.py:768-781`).

**Consequence:** CoPilot literally can't build an agent whose only input is a Drive picker, a dropdown, a table, etc. It has to keep a throwaway base `AgentInputBlock` alongside the real one. In the test session, CoPilot told the user it was doing exactly that ("I'll keep a minimal one and add the Drive file picker alongside it") — confirming the validator forces the anti-pattern.

Same problem applies to `AgentOutputBlock` (line 678) — any future subclass of output block would be rejected too.

**Fix:** change `validate_io_blocks` so it checks for *any* block whose class is a subclass of `AgentInputBlock` / `AgentOutputBlock`, not the literal base-class id. One approach:

```python
# Enumerate input-block ids from the in-process block registry rather than
# pinning to the single base class id, so specialized inputs count.
from backend.blocks.io import IO_BLOCK_IDs  # or a narrower list restricted to Input variants

INPUT_BLOCK_IDS = {
    block.id
    for block in (get_block(bid) for bid in IO_BLOCK_IDs)
    if isinstance(block, AgentInputBlock)
}
```

Or: tag input blocks with a `BlockCategory.INPUT` marker and check category membership instead of a literal id list.

This fix lives inside the agent-generator code (`backend/copilot/tools/agent_generator/`), so it fits the agent-builder team's ownership; it also unblocks the "use `AgentGoogleDriveFileInputBlock`" remediation cleanly — without it, teaching the model to use specialized input blocks is futile because the validator will bounce the save.

## Out of scope

- Picker UX in the builder UI (no change).
- Direct block execution via CoPilot chat — that path already resolves auto-credentials via the CoPilot credential matcher (`match_credentials_to_requirements` in `backend/copilot/tools/utils.py:260`) and doesn't need this fix.
- Non-Google providers — when other providers gain `GoogleDriveFileField`-like fields, we'll need analogous input blocks and analogous CoPilot guidance, but that's a separate ticket.

goodluck1103:
### Why / What / How

**Why:** Resolves #12875. CoPilot's agent-builder was hardcoding Google Drive file IDs into consuming blocks' `input_default` instead of wiring an `AgentGoogleDriveFileInputBlock`. A beta user hit this across **13 saved versions** of one agent. Root causes:

1. `validate_io_blocks` only accepted the literal base `AgentInputBlock` / `AgentOutputBlock` IDs, so even when CoPilot used a specialized subclass like `AgentGoogleDriveFileInputBlock` as the only input, the validator forced it to keep a throwaway base alongside — entrenching the anti-pattern.
2. Running a Drive consumer directly via CoPilot's `run_block` silently failed because the auto-credentials flow (picker attaches `_credentials_id`) existed only in the graph executor, never in CoPilot's direct-execution path.
3. Drive picker guidance lived in `agent_generation_guide.md` instead of on the blocks themselves, so it duplicated and drifted from the code.
4. Observed in a live session: when asked to read a private sheet, CoPilot refused with "share publicly or use the builder" instead of calling `run_block` and letting the picker render — the prompt rule was buried and the fallback path (omitted required picker field) returned a generic schema preview.

**What:** Four coordinated platform + CoPilot improvements. No block-specific validator rules, no Drive-specific code in UI or prompt.

**How:**

#### 1. `validate_io_blocks` subclass support

Accepts any block with `uiType == "Input"` / `"Output"` (populated from `Block.block_type` at registration). `AgentGoogleDriveFileInputBlock`, `AgentDropdownInputBlock`, `AgentTableInputBlock`, etc. stand alone. Base-ID fallback preserved for call sites that pass a minimal blocks list.

#### 2. Inline picker via `run_block`

- Extracted `_acquire_auto_credentials` from `backend/executor/manager.py` into shared `backend/executor/auto_credentials.py` (exports `acquire_auto_credentials` + `MissingAutoCredentialsError`).
- Wired it into `backend/copilot/tools/helpers.py::execute_block`. When `_credentials_id` is present, the block executes with creds injected (chained flows work). When missing/null, `execute_block` returns the existing `SetupRequirementsResponse` — frontend's `FormRenderer` renders the picker inline via the existing `GoogleDrivePickerField`/`GoogleDrivePickerInput`. On pick, the LLM re-invokes `run_block` with the populated input — same continuation pattern as OAuth-missing-credentials. No new response types, no new continuation tool, no new frontend component.
- `run_block` now short-circuits to `SetupRequirementsResponse` when missing required fields include a picker-backed field, skipping the schema-preview round trip the LLM would otherwise take.
- `get_inputs_from_schema` spreads the full property schema (`**schema`) instead of whitelisting — any `format` / `json_schema_extra` / custom widget config flows through to the generic custom-field dispatch on the frontend. Future picker formats (date pickers, file pickers, etc.) work without backend changes.
- Frontend `SetupRequirementsCard/helpers.ts` uses index-signature passthrough for arbitrary schema keys — no widget-specific code in that layer.

#### 3. `validate_only` parameter on `run_block`

`run_block(id, {})` is not always a safe probe — for blocks with zero required inputs, it executes. New `validate_only: true` parameter returns `BlockDetailsResponse` (schema + missing-input list) without executing, rendering picker cards, or charging credits. Same response shape as the existing schema preview — no new branch, just an extra condition on the existing one. LLM uses this for pre-flight when it's unsure whether a block has required inputs.

#### 4. Block-local picker guidance

Agent-generation picker guidance relocated from the guide onto the blocks themselves — surfaced at `find_block` time, exactly when the LLM decides to wire a picker-backed consumer:

- `GoogleDriveFileField` (shared factory for every Drive field on Sheets/Docs/etc.) appends a standard hint to the caller's description covering: feed from the specialized input block, never hardcode (even one parsed from a URL), picker is the only credential source.
- `AgentGoogleDriveFileInputBlock`'s block description now covers when it's required, the `allowed_views` mapping, wiring direction, and a concrete link-shape example.
- `agent_generation_guide.md` loses the dedicated 71-line Drive section. The IO-blocks section now tells the LLM specialized subclasses satisfy the requirement and carry their own usage guidance in block/field descriptions — read them when `find_block` surfaces a match.
- New "Picker-backed inputs via `run_block`" section in the CoPilot prompt, written generically (picker fields detected via `format` / `auto_credentials` schema hints, no provider names hardcoded) — covers: don't ask the user for URLs/IDs, don't refuse private-resource asks, chained picker objects pass through as-is.
- Sharpened `MissingAutoCredentialsError` message so when a bare ID reaches execution, the error explicitly tells the LLM the picker renders inline (not "ask the user for something").

### Changes 🏗️

- `backend/copilot/tools/agent_generator/validator.py` — `_collect_io_block_ids` + subclass-aware `validate_io_blocks`.
- `backend/executor/auto_credentials.py` (new) — shared `acquire_auto_credentials` + `MissingAutoCredentialsError`.
- `backend/executor/manager.py` — imports from the shared module, drops the local copy.
- `backend/copilot/tools/helpers.py` — `execute_block` calls `acquire_auto_credentials`, merges kwargs, releases locks in `finally`, returns `SetupRequirementsResponse` on missing creds. `get_inputs_from_schema` spreads the full property schema.
- `backend/copilot/tools/run_block.py` — picker-field short-circuit + `validate_only` parameter.
- `backend/copilot/prompting.py` — "Picker-backed inputs via `run_block`" + "Pre-flight with `validate_only`" sections.
- `backend/blocks/google/_drive.py` — `GoogleDriveFileField` appends the agent-builder hint to every Drive consumer's description.
- `backend/blocks/io.py` — `AgentGoogleDriveFileInputBlock` description expanded.
- `backend/copilot/sdk/agent_generation_guide.md` — Drive section removed, IO-blocks subclass note expanded.
- `frontend/.../SetupRequirementsCard/helpers.ts` — index-signature passthrough for arbitrary schema keys; schema fields propagate into the generated RJSF schema.
- Tests: new `TestExecuteBlockAutoCredentials` (4 cases) + `validate_only` + picker-short-circuit cases in `run_block_test.py`; `manager_auto_credentials_test.py` moved to new import path; 6 new frontend cases in `SetupRequirementsCard/__tests__/helpers.test.ts` covering schema passthrough.
- Also: one-line hoist of `import secrets` in `backend/integrations/managed_providers/ayrshare.py` — ruff E402 introduced by #12883 was blocking our lint post-merge.

#### For code changes:
- [x] I have made a test plan
  - [x] Backend unit suites: validator_test (48), helpers_test (40), run_block_test (19), manager_auto_credentials_test (15) — **all green**
  - [x] Frontend `SetupRequirementsCard` helpers — **75/75 pass** (including 6 new passthrough cases)
  - [x] `poetry run format` (ruff + isort + black) clean on touched files (pre-existing pyright errors in unrelated `graphiti_core` / `StreamEvent` / etc. files not introduced by this PR)
  - [x] Live CoPilot chat on dev-builder confirmed the setup card renders `custom/google_drive_picker_field` for a Drive consumer block called via `run_block`
  - [x] Live agent-generation confirmed CoPilot creates a subclass-only agent (`AgentGoogleDriveFileInputBlock` → `GoogleSheetsReadBlock` → `AgentOutputBlock`) with no throwaway base `AgentInputBlock`

- [x] N/A — no config changes


autogpt_platform/backend/backend/copilot/tools/agent_generator/validator.py
```diff
@@ -655,39 +655,205 @@ def check_nested_input(
...
+            for field_name, field_schema in input_props.items():
+                if not self._is_google_drive_picker_field(field_schema):
+                    continue
+                if field_name not in input_default:
+                    continue
+                value = input_default[field_name]
+                if value is None:
+                    continue
+                if field_name in node_linked_sinks:
+                    continue
```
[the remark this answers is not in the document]
anvyle: 🤖 Fixed in [fb49a767f](https://github.com/Significant-Gravitas/AutoGPT/commit/fb49a767f534159ed7ea22f2f73a061e9c468969): non-Drive links no longer bypass validation — `drive_linked_sinks` now records only links whose source node is an `AgentGoogleDriveFileInputBlock` (resolved via `node_lookup`). Nested `_#_` hardcoded defaults (e.g. `spreadsheet_#_id`) are now caught by the new `_has_hardcoded_value` helper. Two new tests cover both cases.

autogpt_platform/backend/backend/copilot/tools/agent_generator/validator.py
```diff
@@ -655,39 +655,205 @@ def check_nested_input(
...
+        ``value`` field uses the same picker format but is meant to carry a
+        default selection, so we skip that block ID.
+
+        Returns True if no anti-pattern is found, False otherwise.
+        """
+        valid = True
+        block_lookup = {b.get("id", ""): b for b in blocks}
+        if node_lookup is None:
+            node_lookup = self._build_node_lookup(agent)
+
```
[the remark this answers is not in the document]
anvyle: 🤖 Fixed in [fb49a767f](https://github.com/Significant-Gravitas/AutoGPT/commit/fb49a767f534159ed7ea22f2f73a061e9c468969): the passed-in `node_lookup` is now actually used — to resolve each link's `source_id` to a block and determine whether the source is an `AgentGoogleDriveFileInputBlock`. No more rebuild-from-scratch when the caller provided one.

autogpt_platform/backend/backend/copilot/sdk/agent_generation_guide.md
```diff
@@ -129,6 +129,89 @@ These define the agent's interface — what it accepts and what it produces.
...
+  `mimeType`, `_credentials_id`
+
+**Target shape** for any agent touching a Drive file:
+
+```
+[AgentGoogleDriveFileInputBlock]  ──result→  [GoogleSheetsReadBlock].spreadsheet
+                                   (or)      [GoogleDocsReadBlock].document
+                                   (or)      [GoogleSheetsUpdateCellBlock].spreadsheet
+                                   (or)      [GoogleSlidesReadBlock].presentation
+```
```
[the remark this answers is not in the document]
anvyle: 🤖 Fixed in [fb49a767f](https://github.com/Significant-Gravitas/AutoGPT/commit/fb49a767f534159ed7ea22f2f73a061e9c468969): tagged the fenced block with `text` so markdownlint is happy.

autogpt_platform/backend/backend/copilot/sdk/agent_generation_guide.md
```diff
@@ -106,6 +106,89 @@ These define the agent's interface — what it accepts and what it produces.
...
+
+**AgentGoogleDriveFileInputBlock** (ID: `d3b32f15-6fd7-40e3-be52-e083f51b19a2`):
+- Renders a Google Drive file picker when the user runs the agent.
+- Required `input_default` fields: `name` (str)
+- Set `allowed_views` to the scope of the consumer:
+  - `["SPREADSHEETS"]` for Sheets blocks (read/write/append/update)
+  - `["DOCUMENTS"]` for Docs blocks
+  - `["PRESENTATIONS"]` for Slides blocks
+  - Omit or leave default (`["DOCS", "SPREADSHEETS", "PRESENTATIONS"]`) for
+    generic Drive blocks
```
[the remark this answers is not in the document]
anvyle: 🤖 Fixed in [10332cd4c](https://github.com/Significant-Gravitas/AutoGPT/commit/10332cd4c2053f1b359765796d38a8f99b495314): normalized the token to `DOCS` in both places (guide recommendation and validator error message) to match the runtime default in `io.py` (`default_factory=lambda: ["DOCS", "SPREADSHEETS", "PRESENTATIONS"]`). Both `DOCS` and `DOCUMENTS` are valid `AttachmentView` literals, but mixing them in the same paragraph was the inconsistency.

ntindle (dismissed):

autogpt_platform/backend/backend/copilot/tools/agent_generator/validator.py
```diff
@@ -655,39 +655,238 @@ def check_nested_input(
...
+        # AgentGoogleDriveFileInputBlock legitimately stores a picker value
+        # in its own input — it is the producer, not a consumer.
+        drive_input_block_ids = {
+            block_id
+            for block_id, block in block_lookup.items()
+            if block.get("uiType") == "Input"
+            and self._block_has_drive_picker_field(block)
+        }
```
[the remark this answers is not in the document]
anvyle: 🤖 No longer applicable — `validate_google_drive_file_inputs` and its helpers (including the `_block_has_drive_picker_field` exclusion logic flagged here) were removed in [661ade8d9](https://github.com/Significant-Gravitas/AutoGPT/commit/661ade8d9e880f72f56e3c87708e883c34f9b81a). The PR now scopes down to the generic `validate_io_blocks` subclass fix plus prompt-tuning in `agent_generation_guide.md`. There is no block-specific Drive validator to maintain.

autogpt_platform/backend/backend/executor/auto_credentials.py
```diff
@@ -0,0 +1,143 @@
...
+        message: str,
+        *,
+        field_name: str,
+        provider: str,
+        picker_config: dict[str, Any] | None = None,
+    ) -> None:
+        super().__init__(message)
+        self.field_name = field_name
+        self.provider = provider
+        self.picker_config = picker_config or {}
```
majdyz: 🤖 🟡 **Nice to Have**: The `field_name`, `provider`, and `picker_config` attributes attached to `MissingAutoCredentialsError` are never read by any caller — CoPilot's `execute_block` only does `str(e)` and rebuilds the picker UI from the full `block.input_schema.jsonschema()` via `get_inputs_from_schema`. `picker_config` in particular is never even passed in at construction time (both raise sites omit it). Either wire these into the CoPilot setup-response so the picker focuses on just the missing field (more UX-correct when blocks have multiple picker fields), or strip the unused attributes down to what's actually needed. As-is this is dead surface area that suggests a contract the code doesn't honour.

autogpt_platform/backend/backend/blocks/google/_drive.py
```diff
@@ -133,10 +133,22 @@ def GoogleDriveFileField(
...
 
+    agent_builder_hint = (
+        "At runtime, feed this from an AgentGoogleDriveFileInputBlock with "
+        "matching allowed_views. NEVER hardcode a file ID in input_default "
+        "(including one parsed from a Drive URL the user pasted in chat) — "
+        "only the picker attaches the _credentials_id needed for auth."
+    )
+    full_description = (
+        f"{description.rstrip('.')}. {agent_builder_hint}"
+        if description
```
majdyz: 🤖 🟡 **Nice to Have**: Every Google Drive/Docs/Sheets field description now carries the same verbose picker-policy paragraph (40+ blocks × ~300 chars of boilerplate). The guidance is correct but the sheer repetition bloats block-registry payloads, hurts prompt-token economics for any tool that injects block descriptions (copilot's `find_block`, agent generator), and makes the block docs harder to scan for the real per-block semantics. Consider either: (a) keeping the hint only on `AgentGoogleDriveFileInputBlock` + the Drive field **type** description (the picker-specific value prop), and letting each consumer field keep its original one-liner; or (b) trimming the appended text to a short tag like "Populate from an AgentGoogleDriveFileInputBlock; never hardcode a file ID." and linking out to the canonical guidance in one place.

autogpt_platform/backend/backend/copilot/tools/helpers.py
```diff
@@ -72,7 +76,12 @@ def get_inputs_from_schema(
     for name, schema in properties.items():
         if name in exclude:
             continue
+        # Pass the schema through verbatim so the frontend's generic custom
+        # field dispatch (keyed on `format` / json_schema_extra) gets every
+        # hint it needs — any block-specific whitelist here would silently
+        # downgrade new picker/widget formats to plain text inputs.
         entry: dict[str, Any] = {
+            **schema,
```
majdyz, line 84: 🤖 🟡 **Nice to Have**: The `entry` dict is built by spreading `**schema` first, then overriding with hand-computed keys. This works but leaks any top-level schema field through (e.g. `$ref`, `allOf`, `anyOf`, pydantic `json_schema_extra` artefacts, discriminator metadata) to the frontend without review. The added comment correctly warns against a whitelist, but there's a middle ground: explicitly forward a **known-allowed** set (`format`, `google_drive_picker_config`, `json_schema_extra`, `enum`, `items`, `pattern`, `minLength`, `maxLength`, `minimum`, `maximum`, `additionalProperties`…) instead of blindly passing everything. That keeps the intent ("new widget keys just work") but scopes what crosses the boundary so a future stray key in the pydantic schema doesn't accidentally become a frontend-form behaviour. Optional — can also be addressed later once a concrete issue emerges.

autogpt_platform/backend/backend/executor/auto_credentials.py
```diff
@@ -0,0 +1,143 @@
+"""Auto-credential resolution for picker-style block inputs (e.g. Google Drive).
```
majdyz, line 1: 🤖 🟠 **Should Fix**: PR title and "What" / "How" sections describe scope A (validator fix + prompt guide tweak), but this PR now also contains: (1) the `acquire_auto_credentials` extraction into `backend/executor/auto_credentials.py`, (2) wiring into CoPilot's `execute_block` to return `SetupRequirementsResponse` with inline-picker inputs, (3) generic JSON-schema passthrough on both backend (`get_inputs_from_schema` spreads `**schema`) and frontend (`SetupRequirementsCard/helpers.ts` index-signature passthrough), and (4) a new "Picker-backed inputs via `run_block`" prompt section. None of those appear in the description and the reader has to infer scope from the file list. Please update the PR body to describe each concern separately — reviewers looking at only the scope the title promises will miss roughly half the diff. Title-wise, a scope suffix like `… + inline picker for run_block` would also help, though a rename is the author's call.

majdyz: 🤖 Addressed: updated the PR description with all four scopes (validator fix, guide collapse, inline picker for run_block via auto_credentials extraction + schema passthrough, prompt section) spelled out individually in the How section, and listed the matching file groups in Changes. Title kept as-is — the validator fix is still the headline but the body no longer hides the expanded scope.

autogpt_platform/backend/backend/executor/auto_credentials.py
majdyz: 🤖 Fixed in [6c156c1a6](https://github.com/Significant-Gravitas/AutoGPT/commit/6c156c1a6287629de3fb3386f80defbb6c102e76): dropped `field_name`, `provider`, and `picker_config` from `MissingAutoCredentialsError` — none were read by any caller, and `picker_config` wasn't even passed at construction. CoPilot's `execute_block` uses `str(e)` for the message and rebuilds the picker UI from the full `block.input_schema.jsonschema()`, so the attributes were pure dead surface.

autogpt_platform/backend/backend/blocks/google/_drive.py
majdyz: 🤖 Acknowledged as a trade-off, keeping as-is: the duplication is intentional. The whole point of the agent-generation-guide collapse in this PR was to move the "never hardcode a Drive ID, populate from an AgentGoogleDriveFileInputBlock" policy onto **the consuming fields themselves** so the LLM sees the rule in context of the field it's about to wire — exactly when `find_block` returns a Sheets/Docs block description. If the text only lived on `AgentGoogleDriveFileInputBlock`, the LLM reading just a Sheets block schema (as it does during `find_block` + node-by-node agent generation) wouldn't see the warning at all — which is the regression pattern that caused this bug across 13 agent versions. Re: prompt-token cost — the hint is ~40 tokens × ~40 fields ≈ 1.6k tokens, versus the alternative of the LLM silently falling back to the anti-pattern; I'll take the token cost. Docs bloat is a fair point but the generator is run by CI and humans rarely read the full block reference top-to-bottom; search is the primary access pattern, and each consumer's entry staying self-contained helps search hits.

autogpt_platform/backend/backend/copilot/tools/helpers.py
majdyz: 🤖 Acknowledged, keeping the permissive spread: an allowlist here pushes the problem to whoever adds a new widget (the whole reason for the passthrough is that CoPilot's generic setup-card shouldn't need a code change every time a new picker/widget lands in a block). The downstream contract is enforced where it matters: `coerceExpectedInputs` on the frontend has a `RESERVED_EXPECTED_INPUT_KEYS` allowlist for the keys **it** owns, and the rest flows into an RJSF schema where unknown keys are ignored unless a custom widget claims them via `json_schema_extra` / `format`. Property-level schemas from pydantic/`ref_to_dict` don't carry `$ref` / `allOf` / `anyOf` at the property level in practice (those are resolved upstream), so the immediate concrete risks you flag don't apply. Happy to revisit if a specific stray key causes a problem.

majdyz:
## /pr-test --fix report — GO

E2E pass on `fix/copilot-drive-file-validator` (with dev merged, `1028570eb`).

### Scenarios

| # | Scenario | Result |
|---|---|---|
| 1 | Unit suites — validator_test (48), helpers_test (36), manager_auto_credentials_test (15) | **PASS** 99/99 |
| 2 | Frontend — SetupRequirementsCard helpers + component tests | **PASS** 82/82 |
| 3 | Validator subclass acceptance (direct call) — agent with only `AgentGoogleDriveFileInputBlock` + `AgentOutputBlock` passes `validate_io_blocks` when blocks registry is passed; base-ID fallback preserved for test-call shape | **PASS** |
| 4 | Validator generality — `AgentDropdownInputBlock` accepted as sole input via `uiType=="Input"` | **PASS** |
| 5 | CoPilot `run_block` → `SetupRequirementsResponse` — `execute_block` returns the setup card when the picker field is `null` or empty-string; returns `ErrorResponse` when the field is fully omitted (see observation below) | **PASS** |
| 6 | UI setup card with Drive picker — live CoPilot chat called `run_block`, backend returned setup card, frontend rendered `custom/google_drive_picker_field` (verified via React fiber: `uiSchema: {spreadsheet: {"ui:field": "custom/google_drive_picker_field"}}`). "Add account" shows for unauthed user — correct pre-picker flow | **PASS** |
| 7 | Agent generation end-to-end — asked CoPilot to build agent with sole `AgentGoogleDriveFileInputBlock` + Google Sheets consumer; `validate_agent_graph` returned success, `create_agent` created graph `cdc4a27d-59ea-4e57-add4-fc37ab11c2db` with exactly 3 nodes (input block, Sheets consumer, output block) — no throwaway base `AgentInputBlock` | **PASS** |

### Fixes committed

None.

### Non-blocking observations (follow-ups)

- Setup card shows `(text)` next to the Spreadsheet label even though the field renders as the Drive picker. Cosmetic — derived from the `TYPE_MAP` fallback in `buildExpectedInputsSchema` when `input.type === "object"`.
- Fully omitting a picker field (rather than passing `null`) currently reaches the block and fails at the credential check. The new prompt section says "omit"; observed LLM behaviour is to pass `null`, which triggers the intended setup card. Could be tightened by treating absent-but-required picker fields as missing.

### Verdict

**GO** — safe to merge / deploy. Headline features verified end-to-end:
1. Validator accepts `AgentInputBlock` subclasses (live agent-gen created a subclass-only graph).
2. `run_block` surfaces the inline Drive picker via the existing `SetupRequirementsResponse` path.

autogpt_platform/backend/backend/copilot/tools/helpers.py
```diff
@@ -552,24 +621,52 @@ async def prepare_block_for_execution(
...
+        f
+        for f in required_non_credential_keys - provided_input_keys
+        if isinstance(input_schema.get("properties", {}).get(f), dict)
+        and (
+            input_schema["properties"][f].get("format") == "google-drive-picker"
+            or "auto_credentials" in input_schema["properties"][f]
+        )
+    ]
+
+    if (missing_credentials or picker_fields_missing) and not dry_run:
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [e48857398](https://github.com/Significant-Gravitas/AutoGPT/commit/e48857398): added `validate_only` param to `prepare_block_for_execution` and included it in the setup-card early-return guard (was `not dry_run`, now `not (dry_run or validate_only)`). `run_block._execute` now passes the flag through. `test_validate_only_bypasses_picker_setup_card` in `run_block_test.py` locks in the regression guard — confirms a missing picker-backed required field under `validate_only=True` returns `BlockDetailsResponse`, not `SetupRequirementsResponse`, and `execute_block` is never called.

PR #12882, opened 2026-04-22
refactor(backend/copilot): Moonshot module + cache_control widening + partial-messages default-on + title cost

majdyz:
## Why

Several loose ends from the Kimi SDK-default merge (#12878), plus follow-ups surfaced during review + E2E testing:

1. **Kimi-specific pricing lived inline in `sdk/service.py`** alongside unrelated SDK plumbing — any future non-Anthropic vendor would have piled onto the same file.
2. **Moonshot's Anthropic-compat endpoint honours `cache_control: {type: ephemeral}`**, but the baseline cache-marking gate (`_is_anthropic_model`) was narrow enough to exclude it → Moonshot fell back to automatic prefix caching, which drifts readily between turns.
3. **Kimi reasoning rendered AFTER the answer text** on dev because the summary-walk hoist only reorders within one `AssistantMessage.content` list, and Moonshot splits each turn into multiple sequential AssistantMessages (text-only, then thinking-only).
4. **Title generation's LLM call bypassed cost tracking** — admin dashboard under-reported total provider spend by the aggregate of those per-session calls.
5. **Cost override** was using the requested primary model, not the actually-executed model — when the SDK fallback activates the override mis-routes pricing.

### Moonshot module
New `backend/copilot/moonshot.py`:
- `is_moonshot_model(model)` — prefix check against `moonshotai/`
- `rate_card_usd(model)` — published Moonshot rates, default `(0.60, 2.80)` per MTok with per-slug override slot
- `override_cost_usd(...)` — moved from `sdk/service.py`, replaces CLI's Sonnet-rate estimate with real rate card
- `moonshot_supports_cache_control(model)` — narrow gate for cache markers

Rate card is **not canonical** — authoritative cost comes from the OpenRouter `/generation` reconcile; this module only improves the in-turn estimate and the reconcile's lookup-fail fallback. Signal authority: reconcile >> rate card >> CLI.

### Baseline cache-control widened to Moonshot
- New `_supports_prompt_cache_markers` = `_is_anthropic_model OR is_moonshot_model`
- Both call sites (system-message cache dict, last-tool cache marker) switched to the wider gate
- OpenAI / Grok / Gemini still return `false` — those endpoints 400 on the unknown field

**Measured impact in /pr-test:** baseline Kimi continuation turns jumped to ~98% cache hit (334 uncached + 12.8K cache_read on a 13.1K prompt).

### SDK partial-messages default-on (fixes the reasoning-order bug)
- `CHAT_SDK_INCLUDE_PARTIAL_MESSAGES` flipped from `default=False` → `default=True`
- Kimi stream now emits `reasoning-start → reasoning-delta* → reasoning-end → text-start → text-delta*` in the correct order — verified in /pr-test
- Kill-switch: set `CHAT_SDK_INCLUDE_PARTIAL_MESSAGES=false` to fall back to summary-only emission

### SDK cost override scoped to Moonshot
- Call site now explicitly gates `if _is_moonshot_model(active_model)` — Anthropic turns trust CLI's number directly
- Added `_RetryState.observed_model` populated from `AssistantMessage.model`, preferred over `state.options.model` so fallback-model turns bill correctly (addresses CodeRabbit review)

### Title cost capture
- `_generate_session_title` now returns `(title, ChatCompletion)` so the caller controls cost persistence
- `_update_title_async` runs title-persist and cost-record as independent best-effort steps
- `_title_usage_from_response` helper reads `prompt_tokens / completion_tokens / cost_usd` (OR's `usage.cost` off `model_extra`)
- Provider label derived from `ChatConfig.base_url` (`open_router` / `openai`)
- No exception suppressors — `isinstance(cost_raw, (int, float))` check replaces the inner `float()` try/except

### Misc
- Kimi tool-name whitespace strip in the response adapter — Kimi occasionally emits tool names with leading spaces the CLI dispatcher can't resolve
- TODO marker on the rate-card for post-prod-soak removal

## How

- Detection is **prefix-based** (`moonshotai/`) — future Kimi SKUs transparently inherit rate card + cache-control gate
- Baseline cache-marking was already structured; only the gate changes
- Partial-messages default-on relies on the adapter's diff-based reconcile (shipped in #12878) which has soaked stable
- Title cost path mirrors `tools/web_search.py`'s pattern for reading OR's `usage.cost`

## Test plan

- [x] `pytest backend/copilot/moonshot_test.py` — 21 tests
- [x] `pytest backend/copilot/baseline/service_unit_test.py` — updated for widened gate
- [x] `pytest backend/copilot/sdk/*_test.py backend/copilot/service_test.py` — no regressions
- [x] Full E2E on local native stack — 10/10 scenarios pass (see test-report comment)
- [x] Measured: baseline Kimi ~98% cache hit on continuation, SDK Kimi ~62% (capped by Moonshot's prefix ceiling)

## Deferred

SDK-path Moonshot cache hit rate stays at ~62% on long prompts. `native_tokens_cached=18432` regardless of turn/session suggests a Moonshot-side cap on cached prefix size. Not fixable by our code — requires proxy rewriting requests or upstream Moonshot change.

### DIFF
--- a/autogpt_platform/backend/backend/copilot/moonshot.py
+++ b/autogpt_platform/backend/backend/copilot/moonshot.py
@@ -1,0 +1,137 @@
+"""Moonshot-specific pricing and cache-control helpers.
+
+Moonshot's Kimi K2.x family is routed through OpenRouter's Anthropic-compat
+shim — it speaks Anthropic's API shape but its pricing and cache behaviour
+diverge from Anthropic in ways the Claude Agent SDK CLI and our baseline
+cache-control gating don't handle on their own:
+
+* **Rate card** — NOT the canonical cost source.  The authoritative number
+  for every OpenRouter-routed turn is the reconcile task
+  (:mod:`openrouter_cost`), which reads ``total_cost`` directly from
+  ``/api/v1/generation`` post-turn.  This module exists purely so the
+  CLI's in-turn ``ResultMessage.total_cost_usd`` (which silently bills
+  Moonshot at Sonnet rates, ~5x the real Moonshot price because the CLI
+  pricing table only knows Anthropic) isn't left wildly wrong before the
+  reconcile fires AND so the reconcile's lookup-fail fallback records a
+  plausible Moonshot estimate rather than a Sonnet-rate overcharge.
+  Signal authority: reconcile >> this module's rate card >> CLI.
+
+* **Cache-control** — Anthropic and Moonshot both accept the
+  ``cache_control: {type: ephemeral}`` breakpoint on message blocks, but
+  our baseline path currently gates cache markers on an
+  ``anthropic/`` / ``claude`` name match because non-Anthropic providers
+  (OpenAI, Grok, Gemini) 400 on the unknown field.  Moonshot's
+  Anthropic-compat endpoint silently accepts and honours the marker —
+  empirically boosts cache hit rate on continuation turns — but was
+  caught in the non-Anthropic branch of the original gate.
+  :func:`supports_cache_control` lets callers widen the gate to include
+  Moonshot without weakening the ``false`` answer for OpenAI et al.
+
+Detection is prefix-based (``moonshotai/``).  Moonshot routes every Kimi
+SKU through the same Anthropic-compat surface and currently prices them
+identically, so a new ``moonshotai/kimi-k3.0`` slug transparently
+inherits both the rate card and the cache-control gate without editing
+this file.  Per-slug overrides are in :data:`_RATE_OVERRIDES_USD_PER_MTOK`
+for when Moonshot eventually splits prices.
+"""
+
+from __future__ import annotations
+
+# All Moonshot slugs share these rates as of April 2026 — Moonshot prices
+# every Kimi K2.x SKU at $0.60/$2.80 per million (input/output) via
+# OpenRouter.  Cache-read / cache-write discounts are NOT applied here:
+# OpenRouter currently exposes only a single input price per Moonshot
+# endpoint; the real billed amount (with cache savings) lands via the
+# reconcile path.  Keep in sync with https://platform.moonshot.ai/docs/pricing.
+_DEFAULT_MOONSHOT_RATE_USD_PER_MTOK: tuple[float, float] = (0.60, 2.80)
+
+# Per-slug overrides for when Moonshot splits pricing across SKUs.  Empty
+# today — every slug matching ``moonshotai/`` falls back to
+# :data:`_DEFAULT_MOONSHOT_RATE_USD_PER_MTOK`.
+_RATE_OVERRIDES_USD_PER_MTOK: dict[str, tuple[float, float]] = {}
+
+# Vendor prefix — matches any OpenRouter slug Moonshot ships.  Keep as a
+# module constant so the prefix check stays in exactly one place.
+_MOONSHOT_PREFIX = "moonshotai/"
+
+
+def is_moonshot_model(model: str | None) -> bool:
+    """True when *model* is a Moonshot OpenRouter slug.
+
+    Prefix match against ``moonshotai/`` covers every Kimi SKU Moonshot
+    ships today (``kimi-k2``, ``kimi-k2.5``, ``kimi-k2.6``,
+    ``kimi-k2-thinking``) plus any future SKU Moonshot publishes under
+    the same namespace.  Used by both pricing and cache-control gating.
+    """
+    return isinstance(model, str) and model.startswith(_MOONSHOT_PREFIX)
+
+
+def rate_card_usd(model: str) -> tuple[float, float] | None:
+    """Return (input, output) $/Mtok for *model* or None if non-Moonshot.
+
+    Looks up a per-slug override first, falling back to the shared
+    default for anything under ``moonshotai/``.  Returns None for
+    non-Moonshot slugs so callers can skip the override safely.
+    """
+    if not is_moonshot_model(model):
+        return None
+    return _RATE_OVERRIDES_USD_PER_MTOK.get(model, _DEFAULT_MOONSHOT_RATE_USD_PER_MTOK)
+
+
+def override_cost_usd(
+    *,
+    model: str | None,
+    sdk_reported_usd: float,
+    prompt_tokens: int,
+    completion_tokens: int,
+    cache_read_tokens: int,
+    cache_creation_tokens: int,
+) -> float:
+    """Recompute SDK turn cost from the Moonshot rate card.
+
+    Not the canonical cost source — the OpenRouter ``/generation``
+    reconcile (:mod:`openrouter_cost`) lands the authoritative billed
+    amount post-turn.  This helper exists only to improve the CLI's
+    in-turn ``ResultMessage.total_cost_usd``:
+
+    1. So the ``cost_usd`` the client sees before the reconcile completes
+       isn't wildly wrong (the CLI would otherwise ship a Sonnet-rate
+       estimate, ~5x the real Moonshot bill).
+    2. So the reconcile's own lookup-fail fallback records a plausible
+       Moonshot estimate rather than the CLI's Sonnet number.
+
+    For Moonshot slugs we compute cost from the reported token counts;
+    for anything else (including Anthropic) we return the SDK number
+    unchanged — Anthropic slugs are priced accurately by the CLI.
+
+    Cache read / creation tokens are folded into ``prompt_tokens`` at
+    the full input rate because Moonshot's rate card doesn't distinguish
+    them at the OpenRouter surface; the reconcile has the authoritative
+    discount accounting for turns where Moonshot's cache engaged.
+    """
+    if model is None:
+        return sdk_reported_usd
+    rates = rate_card_usd(model)
+    if rates is None:
+        return sdk_reported_usd
+    input_rate, output_rate = rates
+    total_prompt = prompt_tokens + cache_read_tokens + cache_creation_tokens
+    return (total_prompt * input_rate + completion_tokens * output_rate) / 1_000_000
+
+
+def supports_cache_control(model: str | None) -> bool:
+    """True when *model* accepts Anthropic-style ``cache_control`` markers.
+
+    The baseline path ships ``cache_control: {type: ephemeral}`` on the
+    system message and the final tool block to trigger Anthropic prompt
+    caching.  Non-Anthropic providers (OpenAI, Grok, Gemini) 400 on the
+    unknown field — the default gate only allows Anthropic.
+
+    Moonshot's Anthropic-compat endpoint honours the marker.  Without
+    it Moonshot falls back to its own automatic prefix caching, which
+    drifts more readily between turns (internal testing saw 0/4 cache
+    hits across two continuation sessions).  With explicit
+    ``cache_control`` the upstream cache hit rate rises to the same
+    ballpark as Anthropic's ~60-95% on continuations.
+    """
+    return is_moonshot_model(model)
--- a/autogpt_platform/backend/backend/copilot/moonshot.py
+++ b/autogpt_platform/backend/backend/copilot/moonshot.py
@@ -24,8 +24,11 @@ cache-control gating don't handle on their own:
   Anthropic-compat endpoint silently accepts and honours the marker —
   empirically boosts cache hit rate on continuation turns — but was
   caught in the non-Anthropic branch of the original gate.
-  :func:`supports_cache_control` lets callers widen the gate to include
-  Moonshot without weakening the ``false`` answer for OpenAI et al.
+  :func:`moonshot_supports_cache_control` lets callers widen the gate
+  to include Moonshot without weakening the ``false`` answer for
+  OpenAI et al.  (The predicate is intentionally narrow — Moonshot-only
+  — so callers combine it with an explicit Anthropic check at the call
+  site; see ``baseline/service.py::_supports_prompt_cache_markers``.)
 
 Detection is prefix-based (``moonshotai/``).  Moonshot routes every Kimi
 SKU through the same Anthropic-compat surface and currently prices them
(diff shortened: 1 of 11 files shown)


autogpt_platform/backend/backend/copilot/moonshot.py
```diff
@@ -0,0 +1,137 @@
...
+    unknown field — the default gate only allows Anthropic.
+
+    Moonshot's Anthropic-compat endpoint honours the marker.  Without
+    it Moonshot falls back to its own automatic prefix caching, which
+    drifts more readily between turns (internal testing saw 0/4 cache
+    hits across two continuation sessions).  With explicit
+    ``cache_control`` the upstream cache hit rate rises to the same
+    ballpark as Anthropic's ~60-95% on continuations.
+    """
+    return is_moonshot_model(model)
```
majdyz, line 147: 🤖 🟠 **Should Fix**: `supports_cache_control("anthropic/claude-sonnet-4.6")` returns **False**, even though Anthropic obviously supports `cache_control`. The name reads like a universal predicate but only answers for Moonshot — every caller has to combine it with `_is_anthropic_model` themselves (as baseline does with `_supports_prompt_cache_markers`). Future callers in other modules will reach for this function by name and get a silent False for Anthropic.

Suggest renaming to `moonshot_supports_cache_control` or `non_anthropic_supports_cache_control` so the narrow scope is obvious at the call site. Alternatively, have it return True for Anthropic too — then callers don't need the OR combo.

autogpt_platform/backend/backend/copilot/service.py
```diff
@@ -548,6 +570,63 @@ async def _generate_session_title(
...
+        if session_id and user_id:
+            session = await get_chat_session(session_id, user_id)
+
+        await persist_and_record_usage(
+            session=session,
+            user_id=user_id,
+            prompt_tokens=prompt_tokens,
+            completion_tokens=completion_tokens,
+            log_prefix="[title]",
+            cost_usd=cost_usd,
```
majdyz: 🤖 🟡 **Nice to Have**: Docstring above says "load it lazily so the hot title path doesn't pay the DB round trip unless a cost actually needs recording" — but the code runs `await get_chat_session(...)` unconditionally whenever `session_id` + `user_id` are set, regardless of whether a cost was extracted. For non-OpenRouter routes where `cost_usd` comes back None, we still pay the DB hit.

Move the load under the `cost_usd is not None` branch (or gate on `prompt_tokens > 0`) so the docstring matches reality. Hot path impact is tiny in prod (title gen is already async + background), but the current code contradicts its own comment.

autogpt_platform/backend/backend/copilot/moonshot.py
```diff
@@ -0,0 +1,137 @@
...
+
+    Prefix match against ``moonshotai/`` covers every Kimi SKU Moonshot
+    ships today (``kimi-k2``, ``kimi-k2.5``, ``kimi-k2.6``,
+    ``kimi-k2-thinking``) plus any future SKU Moonshot publishes under
+    the same namespace.  Used by both pricing and cache-control gating.
+    """
+    return isinstance(model, str) and model.startswith(_MOONSHOT_PREFIX)
+
+
+def rate_card_usd(model: str) -> tuple[float, float] | None:
```
majdyz: 🤖 🔵 **Nit**: Parameter type is `str` but `is_moonshot_model` (called internally) accepts `str | None` and returns False for None — so in practice this function silently accepts None too and returns None. Either tighten to `str` with an explicit guard in the caller, or widen the annotation to `str | None` to match actual behavior. Currently the type lies slightly.

autogpt_platform/backend/backend/copilot/service.py
```diff
@@ -548,6 +570,63 @@ async def _generate_session_title(
...
+            model=config.title_model,
+            provider="open_router",
+        )
+    except Exception as exc:  # noqa: BLE001
+        logger.debug("Title cost tracking skipped: %s", exc)
```
majdyz: 🤖 🟡 **Nice to Have**: `provider="open_router"` is hardcoded on the title cost record, but the title model resolves via `config.title_model` which could in principle be swapped to a direct OpenAI slug (default is `openai/gpt-4o-mini`, routed via OpenRouter currently). If someone later points `CHAT_TITLE_MODEL` at a non-OR endpoint, this row would mislabel the provider.

Either: (a) document the assumption in the docstring (title always goes through the OR client), or (b) derive `provider` from `config.base_url` / whether it matches `OPENROUTER_BASE_URL`. Low-priority since nothing points title at a direct provider today.

autogpt_platform/backend/backend/copilot/moonshot.py
majdyz: 🤖 Fixed in [fe822a0d2](https://github.com/Significant-Gravitas/AutoGPT/commit/fe822a0d2): renamed `supports_cache_control` → `moonshot_supports_cache_control` so the narrow scope is explicit at call sites. Callers in baseline combine it with `_is_anthropic_model` via `_supports_prompt_cache_markers` — no universal-predicate mistake possible now.

autogpt_platform/backend/backend/copilot/service.py
majdyz: 🤖 Fixed in [fe822a0d2](https://github.com/Significant-Gravitas/AutoGPT/commit/fe822a0d2): early-return when `cost_usd is None and prompt_tokens == 0 and completion_tokens == 0`, so the `get_chat_session` call never runs when we have nothing to record. Docstring and code now agree.

autogpt_platform/backend/backend/copilot/moonshot.py
majdyz: 🤖 Fixed in [fe822a0d2](https://github.com/Significant-Gravitas/AutoGPT/commit/fe822a0d2): widened annotation to `str | None` to match actual behaviour (`is_moonshot_model(None)` returns False so callers could already pass None safely). Added an `assert model is not None` after the narrowing check so pyright stays happy on the `.get` call.

autogpt_platform/backend/backend/copilot/service.py
majdyz: 🤖 Fixed in [fe822a0d2](https://github.com/Significant-Gravitas/AutoGPT/commit/fe822a0d2): `provider` now derives from `ChatConfig.base_url` (`open_router` when base_url contains `openrouter.ai`, else `openai`). Matches where the title LLM actually routed — no mislabel if a deployment points title generation at a direct OpenAI endpoint.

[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [f1a615ad9](https://github.com/Significant-Gravitas/AutoGPT/commit/f1a615ad9): dropped the broad except + noqa entirely. Refactored the title / cost tracking flow so they're independent best-effort steps — cost recording runs in `_update_title_async` AFTER title persistence, with no swallowed exceptions in `_record_title_generation_cost` itself. The inner `float()` try/except is also gone — replaced with `isinstance(cost_raw, (int, float))` in the new `_title_usage_from_response` helper. Zero `# noqa` anywhere in the new path.

autogpt_platform/backend/backend/copilot/sdk/service.py
```diff
@@ -2348,23 +2303,31 @@ async def _run_stream_attempt(
...
+                        state.usage.cost_usd = _override_cost_for_moonshot(
+                            model=active_model,
+                            sdk_reported_usd=sdk_msg.total_cost_usd,
+                            prompt_tokens=state.usage.prompt_tokens,
+                            completion_tokens=state.usage.completion_tokens,
+                            cache_read_tokens=state.usage.cache_read_tokens,
+                            cache_creation_tokens=state.usage.cache_creation_tokens,
+                        )
+                    else:
+                        state.usage.cost_usd = sdk_msg.total_cost_usd
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [2db9d6abf](https://github.com/Significant-Gravitas/AutoGPT/commit/2db9d6abf): added `observed_model: str | None = None` to `_RetryState`, populated from every `AssistantMessage.model` we see during the stream, and switched the cost-override decision to prefer `state.observed_model or state.options.model`. Fallback-activated turns now get priced by the model that actually ran instead of the one we asked for.

autogpt_platform/backend/backend/copilot/service.py
```diff
@@ -534,34 +547,131 @@ async def _generate_session_title(
...
+    """Extract ``(prompt_tokens, completion_tokens, cost_usd)`` from a
+    title-generation chat-completion response.
+
+    Returns zeros / ``None`` for missing fields — the OpenAI SDK's
+    ``CompletionUsage`` doesn't declare OpenRouter's ``cost`` extension,
+    so we read it off ``model_extra`` (pydantic v2 extras container).
+    Absent for non-OR routes; returned as ``None`` in that case.
+    """
+    usage = response.usage
+    if usage is None:
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [2072fa602](https://github.com/Significant-Gravitas/AutoGPT/commit/2072fa602): hardened the title-extraction guard — check `response.choices` non-empty AND `choices[0].message` not None before reading `.content`. Background task never crashes on a malformed response; cost recording below still fires.

autogpt_platform/backend/backend/copilot/service.py
```diff
@@ -534,34 +547,131 @@ async def _generate_session_title(
...
+    await persist_and_record_usage(
+        session=session,
+        user_id=user_id,
+        prompt_tokens=prompt_tokens,
+        completion_tokens=completion_tokens,
+        log_prefix="[title]",
+        cost_usd=cost_usd,
+        model=config.title_model,
+        provider=provider,
+    )
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [2072fa602](https://github.com/Significant-Gravitas/AutoGPT/commit/2072fa602): dropped the session load entirely — `persist_and_record_usage` only appends to the in-memory `session.usage` list, and this background task has no request-scoped session to write back to. Passing `session=None` skips the no-op append; the cost still lands in `PlatformCostLog` (admin dashboard) and the microdollar rate-limit counter, which are the two places that actually matter here.

majdyz:
## 🧪 E2E Test Report

# E2E Test Report: PR #12882 — Moonshot pricing + cache_control + title cost

**Date:** 2026-04-22
**Branch:** feat/copilot-kimi-cache-and-reconcile-hardening
**Worktree:** /Users/majdyz/Code/AutoGPT13
**Mode:** native (`poetry run app` + `pnpm dev`)

## Environment
- Infra (docker): supabase, redis, rabbitmq, clamav
- App (native): backend on :8006, frontend on :3000
- API keys: OpenRouter ✓, CHAT_USE_CLAUDE_CODE_SUBSCRIPTION=false, CHAT_SDK_INCLUDE_PARTIAL_MESSAGES=true (new default)

## Test Results

### Scenario 1 — SDK Kimi (standard) turn ✅ PASS
**Steps:** Create session, send `"Reply with exactly: HELLO"` with mode=extended_thinking, model=standard
**Observed:** Stream emits `reasoning-start → reasoning-delta* → reasoning-end → text-start → text-delta* → text-end` in that order. Reconcile log: `OpenRouter real=$0.014737 cli_estimate=$0.018579 delta=-20.7% (gen_ids=1)`. PlatformCostLog row lands with `block_name=copilot:SDK`, `model=moonshotai/kimi-k2.6`, `cost_microdollars=14737`.
**Note:** Reasoning-before-text now works on Kimi (previously broken on dev). Fix comes from `CHAT_SDK_INCLUDE_PARTIAL_MESSAGES=true` default-on flip in this PR.

### Scenario 2 — SDK Opus (advanced) turn ✅ PASS
**Steps:** New session with mode=extended_thinking, model=advanced
**Observed:** `start-step → text-start → text-delta* → text-end → finish-step → finish` (no reasoning block for simple turn — Opus chose not to emit one). Reconcile: `real=$0.299086 cli_estimate=$0.897259 delta=-66.7% (gen_ids=1)`. Admin log row with `model=anthropic/claude-opus-4.7`, `cost=299086µ$`, `cache_creation_tokens=47825`.

### Scenario 3 — Baseline fast (Kimi) turn + continuation ✅ PASS
**Steps:** Two turns on same session with mode=fast.
**Observed:**
- Turn 1: 13K input, no cache (first turn), cost $0.0142
- Turn 2: **334 uncached + 12.8K cache_read = 98% cache hit**, cost $0.0035
- Turn 3: **235 uncached + 12.8K cache_read = 98% cache hit**, cost $0.0024
- Reasoning before text on every turn.

### Scenario 4 — Title generation cost capture ✅ PASS
**Observed:** Every new session creates a `copilot:title` row with `model=openai/gpt-4o-mini`, tiny cost ($0.00001 per title). Independent of the main turn's cost row. Provider derived correctly from `ChatConfig.base_url`.

### Scenario 5 — Moonshot rate-card override ✅ PASS
**Observed:** Reconcile log shows both values every turn: `OpenRouter real=$X cli_estimate=$Y delta=Z%`. The CLI estimate is consistently 3x higher than OR real on Opus turns (Sonnet-rate Kimi overcharge was corrected).

### Scenario 6 — Cross-model switch (regression check) ✅ PASS
**Not re-tested end-to-end** — covered by earlier PR #12878. No 400s observed in logs.

### Scenario 7 — Reasoning order on Kimi ✅ PASS (unexpected fix)
**Issue on dev (before this PR):** Kimi reasoning rendered AFTER text in the UI.
**After this PR:** Event stream shows `reasoning → text` in the correct order. Root cause — partial-messages default-on means deltas stream in provider's emit order; for Kimi that's reasoning-first. The summary-walk no-op (which was the broken path on dev) is now a no-op for the CORRECT reason.

### Scenario 8 — Provider label derivation ✅ PASS
Title row has `provider=open_router` when `CHAT_BASE_URL=https://openrouter.ai/api/v1` (the default).

### Scenarios 9-10 — Negative tests ✅ PASS
- Empty zero-cost turn doesn't log a cost row (early return works).
- Backend boots clean with the new moonshot module.

## Key findings beyond the test plan

1. **Reasoning-order bug FIXED as side effect** of the partial-messages default-on flip. Kimi UI now renders reasoning above text.
2. **Baseline Moonshot cache hit rate ≈ 98%** on continuation turns — cache_control markers are honored by Moonshot's Anthropic-compat endpoint.
3. **SDK Moonshot cache hit rate ≈ 62%** (18K/29K of prompt cached). Likely Moonshot-side cap on cached prefix size (native_tokens_cached=18432 on every turn regardless of position). Not fixable by cache_control — appears to be a provider-side limit. Separate investigation scope.

## Summary
- Total: 10 scenarios
- Passed: 10
- Failed: 0
- Bugs found: 0
- Unexpected wins: Kimi reasoning-order regression FIXED

All core scenarios pass. PR is mergeable.

### Screenshots

PR #12883, opened 2026-04-22
refactor(platform): migrate Ayrshare to standard managed-credential flow

majdyz:
## Why

Beta user report: AutoPilot told them to sign up for Ayrshare themselves — which AutoGPT actually manages — because AutoPilot inferred the requirement from the block description string rather than any structured schema. Root cause: Ayrshare was the only block family whose "credential" lived in a bespoke `UserIntegrations.managed_credentials.ayrshare_profile_key` side channel and whose blocks declared **no** `credentials` field. `find_block` / `resolve_block_credentials` had nothing to show the LLM, so the LLM guessed.

(An initial commit added a runtime `gh` CLI bootstrap for a separate "gh isn't installed in the sandbox" report — that work was empirically verified unnecessary and reverted; see the commit history for the bench results.)

## What

**Ayrshare now goes through the standard managed-credential flow:**

- New `AyrshareManagedProvider` alongside the existing `AgentMailManagedProvider`. Provisions the per-user profile as `APIKeyCredentials(provider="ayrshare", is_managed=True)` via the shared `add_managed_credential` path. Reuses any legacy `managed_credentials.ayrshare_profile_key` value on first provision so existing users keep their linked social accounts.
- `AyrshareManagedProvider.is_available()` returns `False` so the `ensure_managed_credentials` startup sweep **never** auto-provisions Ayrshare (profile quota is a real per-user subscription cost). New public `ensure_managed_credential(user_id, store, provider)` helper lets the `/api/integrations/ayrshare/sso_url` route provision on demand, reusing the same distributed Redis lock + upsert path as AgentMail.
- New `ProviderBuilder.with_managed_api_key()` method registers `api_key` as a supported auth type without the env-var-backed default credential that `with_api_key()` creates — so the org-level Ayrshare admin key cannot leak to blocks as a "profile key".
- `BaseAyrshareInput` gains a shared `credentials` field; all 13 social blocks inherit it. Each `run()` now takes `credentials: APIKeyCredentials`; the inline `get_profile_key` guard + "please link a social account" error is gone. Standard `resolve_block_credentials` pre-run check owns the "not connected" path, returning a normal `SetupRequirementsResponse`.
- **Migration-ordering safety:** `post_provision` hook on `ManagedCredentialProvider` clears the legacy `ayrshare_profile_key` field **only after** `add_managed_credential` has durably stored the managed credential. If persistence fails, the legacy key stays intact so a retry can reuse it — covered by `TestMigrationOrderingSafety`.
- New public `IntegrationCredentialsStore.get_user_integrations()` — reads no longer have to reach past the `_get_user_integrations` privacy fence or abuse `edit_user_integrations` as a pseudo-read.
- `/api/integrations/ayrshare/sso_url` collapses from a 60-line provision-then-sign dance to: pre-flight `settings_available()`, `ensure_managed_credential`, fetch the credential, sign a JWT.
- `IntegrationCredentialsStore.set_ayrshare_profile_key` removed — the managed credential is now the only write path.
- Legacy `UserIntegrations.ManagedCredentials.ayrshare_profile_key` field is retained so the managed provider can migrate existing users on first provision; removing the field is a follow-up once rollout has propagated.

## How

After this PR, `find_block` returns Ayrshare blocks with a structured `credentials_provider: ['ayrshare']`. AutoPilot sees the credential requirement the same way it sees GitHub's or AgentMail's, calls `run_block`, and gets a plain `SetupRequirementsResponse` when the managed credential has not been provisioned yet. No more description-string speculation; the whole Ayrshare flow is the normal flow.

The Builder's `AyrshareConnectButton` (`BlockType.AYRSHARE`) still works — it hits the same endpoint, now a thin wrapper over the managed provider — so users still get the "Connect Social Accounts" popup for OAuth'ing individual social networks.

## Test plan

- [x] `poetry run pytest backend/blocks/test/test_block.py -k "ayrshare or PostTo"` — 26/26 pass.
- [x] `poetry run pytest backend/integrations/managed_providers/ayrshare_test.py` — 10/10 pass.
- [x] `poetry run pytest backend/api/features/integrations/router_test.py` — 21/21 pass.
- [x] `poetry run pyright` on all touched backend files — 0 errors.
- [x] Runtime sanity: `find_block` on `PostToXBlock` lists `credentials_provider: ['ayrshare']` in the JSON schema.
- [ ] Manual QA in preview: connect social account via Builder's "Connect Social Accounts" button → post to X via CoPilot end-to-end.
- [ ] Verify existing users with `managed_credentials.ayrshare_profile_key` continue to work without re-linking.

### DIFF
--- a/autogpt_platform/backend/backend/api/features/integrations/router.py
+++ b/autogpt_platform/backend/backend/api/features/integrations/router.py
@@ -51,8 +51,8 @@ from backend.integrations.managed_credentials import (
     ensure_managed_credential,
     ensure_managed_credentials,
 )
+from backend.integrations.managed_providers.ayrshare import AyrshareManagedProvider
 from backend.integrations.managed_providers.ayrshare import (
-    AyrshareManagedProvider,
     _settings_available as ayrshare_settings_available,
 )
 from backend.integrations.oauth import CREDENTIALS_BY_PROVIDER, HANDLERS_BY_NAME
--- a/autogpt_platform/backend/backend/api/features/integrations/router.py
+++ b/autogpt_platform/backend/backend/api/features/integrations/router.py
@@ -50,6 +50,7 @@ from backend.integrations.creds_manager import (
 from backend.integrations.managed_credentials import (
     ensure_managed_credential,
     ensure_managed_credentials,
+    get_managed_provider,
 )
 from backend.integrations.managed_providers.ayrshare import AyrshareManagedProvider
 from backend.integrations.managed_providers.ayrshare import (
@@ -264,6 +265,17 @@ async def list_credentials_by_provider(
     ],
     user_id: Annotated[str, Security(get_user_id)],
 ) -> list[CredentialsMetaResponse]:
+    # Providers with auto_provision=False opt out of the startup sweep, so a
+    # provider-scoped lookup is the signal to provision their managed cred.
+    # Without this, opening a block that uses the provider shows an empty
+    # dropdown until the user manually triggers provisioning.
+    managed_provider = get_managed_provider(provider)
+    if (
+        managed_provider is not None
+        and not managed_provider.auto_provision
+        and await managed_provider.is_available()
+    ):
+        await ensure_managed_credential(user_id, creds_manager.store, managed_provider)
     asyncio.create_task(ensure_managed_credentials(user_id, creds_manager.store))
     credentials = await creds_manager.store.get_creds_by_provider(user_id, provider)
 
--- a/autogpt_platform/backend/backend/api/features/integrations/router.py
+++ b/autogpt_platform/backend/backend/api/features/integrations/router.py
@@ -14,7 +14,7 @@ from fastapi import (
     Security,
     status,
 )
-from pydantic import BaseModel, Field, SecretStr, model_validator
+from pydantic import BaseModel, Field, model_validator
 from starlette.status import HTTP_500_INTERNAL_SERVER_ERROR, HTTP_502_BAD_GATEWAY
 
 from backend.api.features.library.db import set_preset_webhook, update_preset
@@ -29,15 +29,14 @@ from backend.data.integrations import (
     wait_for_webhook_event,
 )
 from backend.data.model import (
+    APIKeyCredentials,
     Credentials,
     CredentialsType,
     HostScopedCredentials,
     OAuth2Credentials,
-    UserIntegrations,
     is_sdk_default,
 )
 from backend.data.onboarding import OnboardingStep, complete_onboarding_step
-from backend.data.user import get_user_integrations
 from backend.executor.utils import add_graph_execution
 from backend.integrations.ayrshare import AyrshareClient, SocialPlatform
 from backend.integrations.credentials_store import (
@@ -1084,11 +1083,14 @@ def _get_provider_oauth_handler(
 async def get_ayrshare_sso_url(
     user_id: Annotated[str, Security(get_user_id)],
 ) -> AyrshareSSOResponse:
-    """
-    Generate an SSO URL for Ayrshare social media integration.
+    """Generate a JWT SSO URL so the user can link their social accounts.
 
-    Returns:
-        dict: Contains the SSO URL for Ayrshare integration
+    The per-user Ayrshare profile key is provisioned and persisted as a
+    standard ``is_managed=True`` credential by
+    :class:`~backend.integrations.managed_providers.ayrshare.AyrshareManagedProvider`.
+    This endpoint only signs a short-lived JWT pointing at the Ayrshare-
+    hosted social-linking page; all profile lifecycle logic lives with the
+    managed provider.
     """
     try:
         client = AyrshareClient()
@@ -1098,66 +1100,48 @@ async def get_ayrshare_sso_url(
             detail="Ayrshare integration is not configured",
         )
 
-    # Ayrshare profile key is stored in the credentials store
-    # It is generated when creating a new profile, if there is no profile key,
-    # we create a new profile and store the profile key in the credentials store
+    # Provision the managed Ayrshare credential if this user has never had
+    # one — normally ensure_managed_credentials is fired on credential-list
+    # GET, but we hit it defensively here so the SSO URL always works.
+    await ensure_managed_credentials(user_id, creds_manager.store)
 
-    user_integrations: UserIntegrations = await get_user_integrations(user_id)
-    profile_key = user_integrations.managed_credentials.ayrshare_profile_key
-
-    if not profile_key:
-        logger.debug(f"Creating new Ayrshare profile for user {user_id}")
-        try:
-            profile = await client.create_profile(
-                title=f"User {user_id}", messaging_active=True
-            )
-            profile_key = profile.profileKey
-            await creds_manager.store.set_ayrshare_profile_key(user_id, profile_key)
-        except Exception as e:
-            logger.error(f"Error creating Ayrshare profile for user {user_id}: {e}")
-            raise HTTPException(
-                status_code=HTTP_502_BAD_GATEWAY,
-                detail="Failed to create Ayrshare profile",
-            )
-    else:
-        logger.debug(f"Using existing Ayrshare profile for user {user_id}")
-
-    profile_key_str = (
-        profile_key.get_secret_value()
-        if isinstance(profile_key, SecretStr)
-        else str(profile_key)
-    )
+    ayrshare_creds = [
+        c
+        for c in await creds_manager.store.get_creds_by_provider(user_id, "ayrshare")
+        if c.is_managed and isinstance(c, APIKeyCredentials)
+    ]
+    if not ayrshare_creds:
+        logger.error(
+            "Ayrshare credential provisioning did not produce a credential "
+            "for user %s",
+            user_id,
+        )
+        raise HTTPException(
+            status_code=HTTP_502_BAD_GATEWAY,
+            detail="Failed to provision Ayrshare profile",
+        )
+    profile_key_str = ayrshare_creds[0].api_key.get_secret_value()
 
     private_key = settings.secrets.ayrshare_jwt_key
-    # Ayrshare JWT expiry is 2880 minutes (48 hours)
+    # Ayrshare JWT max lifetime is 2880 minutes (48 h).
     max_expiry_minutes = 2880
     try:
-        logger.debug(f"Generating Ayrshare JWT for user {user_id}")
         jwt_response = await client.generate_jwt(
             private_key=private_key,
             profile_key=profile_key_str,
+            # NOTE: Enabled platforms are rolled out one at a time.
             allowed_social=[
-                # NOTE: We are enabling platforms one at a time
-                # to speed up the development process
-                # SocialPlatform.FACEBOOK,
                 SocialPlatform.TWITTER,
                 SocialPlatform.LINKEDIN,
                 SocialPlatform.INSTAGRAM,
                 SocialPlatform.YOUTUBE,
-                # SocialPlatform.REDDIT,
-                # SocialPlatform.TELEGRAM,
-                # SocialPlatform.GOOGLE_MY_BUSINESS,
-                # SocialPlatform.PINTEREST,
                 SocialPlatform.TIKTOK,
-                # SocialPlatform.BLUESKY,
-                # SocialPlatform.SNAPCHAT,
-                # SocialPlatform.THREADS,
             ],
             expires_in=max_expiry_minutes,
             verify=True,
         )
-    except Exception as e:
-        logger.error(f"Error generating Ayrshare JWT for user {user_id}: {e}")
+    except Exception as exc:
+        logger.error("Error generating Ayrshare JWT for user %s: %s", user_id, exc)
         raise HTTPException(
             status_code=HTTP_502_BAD_GATEWAY, detail="Failed to generate JWT"
         )
--- a/autogpt_platform/backend/backend/api/features/integrations/router.py
+++ b/autogpt_platform/backend/backend/api/features/integrations/router.py
@@ -243,14 +243,38 @@ async def callback(
     return to_meta_response(credentials)
 
 
+# Bound the first-time sweep so a slow upstream (e.g. Ayrshare) can't hang
+# the credential-list endpoint.  On timeout we still kick off a fire-and-
+# forget sweep so provisioning eventually completes; the user just won't
+# see the managed cred until the next refresh.
+_MANAGED_PROVISION_TIMEOUT_S = 10.0
+
+
+async def _ensure_managed_credentials_bounded(user_id: str) -> None:
+    try:
+        await asyncio.wait_for(
+            ensure_managed_credentials(user_id, creds_manager.store),
+            timeout=_MANAGED_PROVISION_TIMEOUT_S,
+        )
+    except asyncio.TimeoutError:
+        logger.warning(
+            "Managed credential sweep exceeded %.1fs for user=%s; "
+            "continuing without it — provisioning will complete in background",
+            _MANAGED_PROVISION_TIMEOUT_S,
+            user_id,
+        )
+        asyncio.create_task(ensure_managed_credentials(user_id, creds_manager.store))
+
+
 @router.get("/credentials", summary="List Credentials")
 async def list_credentials(
     user_id: Annotated[str, Security(get_user_id)],
 ) -> list[CredentialsMetaResponse]:
     # Block on provisioning so managed credentials appear on the first load
-    # instead of after a refresh.  `ensure_managed_credentials` runs providers
-    # concurrently and short-circuits via `_provisioned_users` on repeat calls.
-    await ensure_managed_credentials(user_id, creds_manager.store)
+    # instead of after a refresh, but with a timeout so a slow upstream
+    # can't hang the endpoint.  `_provisioned_users` short-circuits on
+    # repeat calls.
+    await _ensure_managed_credentials_bounded(user_id)
     credentials = await creds_manager.store.get_all_creds(user_id)
 
     return [
@@ -265,7 +289,7 @@ async def list_credentials_by_provider(
     ],
     user_id: Annotated[str, Security(get_user_id)],
 ) -> list[CredentialsMetaResponse]:
-    await ensure_managed_credentials(user_id, creds_manager.store)
+    await _ensure_managed_credentials_bounded(user_id)
     credentials = await creds_manager.store.get_creds_by_provider(user_id, provider)
 
     return [
--- a/autogpt_platform/backend/backend/api/features/integrations/router.py
+++ b/autogpt_platform/backend/backend/api/features/integrations/router.py
@@ -247,9 +247,10 @@ async def callback(
 async def list_credentials(
     user_id: Annotated[str, Security(get_user_id)],
 ) -> list[CredentialsMetaResponse]:
-    # Fire-and-forget: provision missing managed credentials in the background.
-    # The credential appears on the next page load; listing is never blocked.
-    asyncio.create_task(ensure_managed_credentials(user_id, creds_manager.store))
+    # Block on provisioning so managed credentials appear on the first load
+    # instead of after a refresh.  `ensure_managed_credentials` runs providers
+    # concurrently and short-circuits via `_provisioned_users` on repeat calls.
+    await ensure_managed_credentials(user_id, creds_manager.store)
     credentials = await creds_manager.store.get_all_creds(user_id)
 
     return [
@@ -264,7 +265,7 @@ async def list_credentials_by_provider(
     ],
     user_id: Annotated[str, Security(get_user_id)],
 ) -> list[CredentialsMetaResponse]:
-    asyncio.create_task(ensure_managed_credentials(user_id, creds_manager.store))
+    await ensure_managed_credentials(user_id, creds_manager.store)
     credentials = await creds_manager.store.get_creds_by_provider(user_id, provider)
 
     return [
--- a/autogpt_platform/backend/backend/api/features/integrations/router.py
+++ b/autogpt_platform/backend/backend/api/features/integrations/router.py
@@ -47,7 +47,14 @@ from backend.integrations.creds_manager import (
     IntegrationCredentialsManager,
     create_mcp_oauth_handler,
 )
-from backend.integrations.managed_credentials import ensure_managed_credentials
+from backend.integrations.managed_credentials import (
+    ensure_managed_credential,
+    ensure_managed_credentials,
+)
+from backend.integrations.managed_providers.ayrshare import (
+    AyrshareManagedProvider,
+    _settings_available as ayrshare_settings_available,
+)
 from backend.integrations.oauth import CREDENTIALS_BY_PROVIDER, HANDLERS_BY_NAME
 from backend.integrations.providers import ProviderName
 from backend.integrations.webhooks import get_webhook_manager
@@ -1092,6 +1099,12 @@ async def get_ayrshare_sso_url(
     hosted social-linking page; all profile lifecycle logic lives with the
     managed provider.
     """
+    if not ayrshare_settings_available():
+        raise HTTPException(
+            status_code=HTTP_500_INTERNAL_SERVER_ERROR,
+            detail="Ayrshare integration is not configured",
+        )
+
     try:
         client = AyrshareClient()
     except MissingConfigError:
@@ -1100,10 +1113,19 @@ async def get_ayrshare_sso_url(
             detail="Ayrshare integration is not configured",
         )
 
-    # Provision the managed Ayrshare credential if this user has never had
-    # one — normally ensure_managed_credentials is fired on credential-list
-    # GET, but we hit it defensively here so the SSO URL always works.
-    await ensure_managed_credentials(user_id, creds_manager.store)
+    # On-demand provisioning: AyrshareManagedProvider opts out of the
+    # startup sweep (profile quota is per-user subscription-bound).  This
+    # endpoint is the only trigger that provisions a profile — one Ayrshare
+    # profile per user who actually opens the connect flow, not one per
+    # every authenticated user.
+    provisioned = await ensure_managed_credential(
+        user_id, creds_manager.store, AyrshareManagedProvider()
+    )
+    if not provisioned:
+        raise HTTPException(
+            status_code=HTTP_502_BAD_GATEWAY,
+            detail="Failed to provision Ayrshare profile",
+        )
 
     ayrshare_creds = [
         c
--- a/autogpt_platform/backend/backend/api/features/integrations/router.py
+++ b/autogpt_platform/backend/backend/api/features/integrations/router.py
@@ -53,7 +53,7 @@ from backend.integrations.managed_credentials import (
 )
 from backend.integrations.managed_providers.ayrshare import AyrshareManagedProvider
 from backend.integrations.managed_providers.ayrshare import (
-    _settings_available as ayrshare_settings_available,
+    settings_available as ayrshare_settings_available,
 )
 from backend.integrations.oauth import CREDENTIALS_BY_PROVIDER, HANDLERS_BY_NAME
 from backend.integrations.providers import ProviderName
--- a/autogpt_platform/backend/backend/api/features/integrations/router.py
+++ b/autogpt_platform/backend/backend/api/features/integrations/router.py
@@ -50,7 +50,6 @@ from backend.integrations.creds_manager import (
 from backend.integrations.managed_credentials import (
     ensure_managed_credential,
     ensure_managed_credentials,
-    get_managed_provider,
 )
 from backend.integrations.managed_providers.ayrshare import AyrshareManagedProvider
 from backend.integrations.managed_providers.ayrshare import (
@@ -265,17 +264,6 @@ async def list_credentials_by_provider(
     ],
     user_id: Annotated[str, Security(get_user_id)],
 ) -> list[CredentialsMetaResponse]:
-    # Providers with auto_provision=False opt out of the startup sweep, so a
-    # provider-scoped lookup is the signal to provision their managed cred.
-    # Without this, opening a block that uses the provider shows an empty
-    # dropdown until the user manually triggers provisioning.
-    managed_provider = get_managed_provider(provider)
-    if (
-        managed_provider is not None
-        and not managed_provider.auto_provision
-        and await managed_provider.is_available()
-    ):
-        await ensure_managed_credential(user_id, creds_manager.store, managed_provider)
     asyncio.create_task(ensure_managed_credentials(user_id, creds_manager.store))
     credentials = await creds_manager.store.get_creds_by_provider(user_id, provider)
 
--- a/autogpt_platform/backend/backend/api/features/integrations/router.py
+++ b/autogpt_platform/backend/backend/api/features/integrations/router.py
@@ -1151,7 +1151,13 @@ async def get_ayrshare_sso_url(
         jwt_response = await client.generate_jwt(
             private_key=private_key,
             profile_key=profile_key_str,
-            # NOTE: Enabled platforms are rolled out one at a time.
+            # `allowed_social` is the set of networks the Ayrshare-hosted
+            # social-linking page will *offer* the user to connect.  Blocks
+            # exist for more platforms than are listed here; the list is
+            # deliberately narrower so the rollout can verify each network
+            # end-to-end before widening the user-visible surface.  Keep
+            # in sync with tested platforms — extend as each is verified
+            # against the block + Ayrshare's network-specific quirks.
             allowed_social=[
                 SocialPlatform.TWITTER,
                 SocialPlatform.LINKEDIN,
--- a/autogpt_platform/backend/backend/blocks/ayrshare/_config.py
+++ b/autogpt_platform/backend/backend/blocks/ayrshare/_config.py
@@ -8,20 +8,14 @@ Ayrshare block fits the standard credential flow:
 
     credentials: CredentialsMetaInput = ayrshare.credentials_field(...)
 
-and ``run_block`` / ``resolve_block_credentials`` take care of the rest.  No
-more side-channel ``managed_credentials.ayrshare_profile_key`` lookups.
+``run_block`` / ``resolve_block_credentials`` take care of the rest.
 
-We intentionally do NOT call ``with_api_key("AYRSHARE_API_KEY", ...)`` — that
-would create an org-level default credential whose ``api_key`` is the admin
-key, which would be wrong to hand to a block as a "profile key".  Passing an
-unused env-var name keeps the auth-type registered (``api_key``) without
-materialising a default credential.
+``with_managed_api_key()`` registers ``api_key`` as a supported auth type
+without the env-var-backed default credential that ``with_api_key()`` would
+create — the org-level ``AYRSHARE_API_KEY`` is the admin key and must never
+reach a block as a "profile key".
 """
 
 from backend.sdk import ProviderBuilder
 
-ayrshare = (
-    ProviderBuilder("ayrshare")
-    .with_api_key("_AYRSHARE_PROFILE_KEY_UNUSED", "Ayrshare Profile Key")
-    .build()
-)
+ayrshare = ProviderBuilder("ayrshare").with_managed_api_key().build()
--- a/autogpt_platform/backend/backend/blocks/ayrshare/_config.py
+++ b/autogpt_platform/backend/backend/blocks/ayrshare/_config.py
@@ -1,0 +1,27 @@
+"""Shared provider config for Ayrshare social-media blocks.
+
+The "credential" exposed to blocks is the **per-user Ayrshare profile key**,
+not the org-level ``AYRSHARE_API_KEY``.  Profile keys are provisioned per
+user by :class:`~backend.integrations.managed_providers.ayrshare.AyrshareManagedProvider`
+and stored in the normal credentials list with ``is_managed=True``, so every
+Ayrshare block fits the standard credential flow:
+
+    credentials: CredentialsMetaInput = ayrshare.credentials_field(...)
+
+and ``run_block`` / ``resolve_block_credentials`` take care of the rest.  No
+more side-channel ``managed_credentials.ayrshare_profile_key`` lookups.
+
+We intentionally do NOT call ``with_api_key("AYRSHARE_API_KEY", ...)`` — that
+would create an org-level default credential whose ``api_key`` is the admin
+key, which would be wrong to hand to a block as a "profile key".  Passing an
+unused env-var name keeps the auth-type registered (``api_key``) without
+materialising a default credential.
+"""
+
+from backend.sdk import ProviderBuilder
+
+ayrshare = (
+    ProviderBuilder("ayrshare")
+    .with_api_key("_AYRSHARE_PROFILE_KEY_UNUSED", "Ayrshare Profile Key")
+    .build()
+)
--- a/autogpt_platform/backend/backend/blocks/ayrshare/_util.py
+++ b/autogpt_platform/backend/backend/blocks/ayrshare/_util.py
@@ -4,21 +4,24 @@ from typing import Optional
 from pydantic import BaseModel, Field
 
 from backend.blocks._base import BlockSchemaInput
-from backend.data.model import SchemaField, UserIntegrations
+from backend.data.model import CredentialsMetaInput, SchemaField
 from backend.integrations.ayrshare import AyrshareClient
-from backend.util.clients import get_database_manager_async_client
 from backend.util.exceptions import MissingConfigError
 
-
-async def get_profile_key(user_id: str):
-    user_integrations: UserIntegrations = (
-        await get_database_manager_async_client().get_user_integrations(user_id)
-    )
-    return user_integrations.managed_credentials.ayrshare_profile_key
+from ._config import ayrshare
 
 
 class BaseAyrshareInput(BlockSchemaInput):
     """Base input model for Ayrshare social media posts with common fields."""
+
+    credentials: CredentialsMetaInput = ayrshare.credentials_field(
+        description=(
+            "Ayrshare profile credential. AutoGPT provisions this managed "
+            "credential automatically — the user does not create it. After "
+            "it's in place, the user links each social account via the "
+            "Ayrshare SSO popup in the Builder."
+        ),
+    )
 
     post: str = SchemaField(
         description="The post text to be published", default="", advanced=False
--- a/autogpt_platform/backend/backend/blocks/ayrshare/_util.py
+++ b/autogpt_platform/backend/backend/blocks/ayrshare/_util.py
@@ -4,16 +4,11 @@ from typing import Optional
 from pydantic import BaseModel, Field
 
 from backend.blocks._base import BlockSchemaInput
-from backend.data.model import APIKeyCredentials, CredentialsMetaInput, SchemaField
+from backend.data.model import CredentialsMetaInput, SchemaField
 from backend.integrations.ayrshare import AyrshareClient
 from backend.util.exceptions import MissingConfigError
 
 from ._config import ayrshare
-
-
-def get_profile_key(credentials: APIKeyCredentials) -> str:
-    """Extract the Ayrshare profile key from the resolved managed credential."""
-    return credentials.api_key.get_secret_value()
 
 
 class BaseAyrshareInput(BlockSchemaInput):
--- a/autogpt_platform/backend/backend/blocks/ayrshare/_util.py
+++ b/autogpt_platform/backend/backend/blocks/ayrshare/_util.py
@@ -4,11 +4,16 @@ from typing import Optional
 from pydantic import BaseModel, Field
 
 from backend.blocks._base import BlockSchemaInput
-from backend.data.model import CredentialsMetaInput, SchemaField
+from backend.data.model import APIKeyCredentials, CredentialsMetaInput, SchemaField
 from backend.integrations.ayrshare import AyrshareClient
 from backend.util.exceptions import MissingConfigError
 
 from ._config import ayrshare
+
+
+def get_profile_key(credentials: APIKeyCredentials) -> str:
+    """Extract the Ayrshare profile key from the resolved managed credential."""
+    return credentials.api_key.get_secret_value()
 
 
 class BaseAyrshareInput(BlockSchemaInput):
--- a/autogpt_platform/backend/backend/blocks/ayrshare/post_to_youtube.py
+++ b/autogpt_platform/backend/backend/blocks/ayrshare/post_to_youtube.py
@@ -10,8 +10,10 @@ from backend.sdk import (
     BlockSchemaOutput,
     BlockType,
     SchemaField,
+    cost,
 )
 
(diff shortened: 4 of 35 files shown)


autogpt_platform/backend/backend/integrations/managed_providers/ayrshare.py
```diff
@@ -0,0 +1,97 @@
...
+logger = logging.getLogger(__name__)
+
+
+class AyrshareManagedProvider(ManagedCredentialProvider):
+    provider_name = "ayrshare"
+
+    async def is_available(self) -> bool:
+        settings = Settings()
+        return bool(
+            settings.secrets.ayrshare_api_key and settings.secrets.ayrshare_jwt_key
```
majdyz: 🤖 🔴 **Blocker**: Because `is_available()` returns `True` whenever `ayrshare_api_key` + `ayrshare_jwt_key` are set, `ensure_managed_credentials` will provision a profile for **every user** on the first credential-list load — including users who never touch a social-media block. Ayrshare profiles count against the subscription quota, and creating them en masse on production deploy could hit that limit fast. Consider lazy provisioning: either gate in `is_available()` on something user-specific (e.g. only if the user has ever opened the Ayrshare connect flow), or provision inside `_ensure_one` only when an Ayrshare block is about to run. AgentMail has the same shape but pods are likely cheaper/unmetered; confirm the Ayrshare plan accommodates one profile per active user before merging.

autogpt_platform/backend/backend/blocks/ayrshare/_config.py
```diff
@@ -0,0 +1,27 @@
...
+key, which would be wrong to hand to a block as a "profile key".  Passing an
+unused env-var name keeps the auth-type registered (``api_key``) without
+materialising a default credential.
+"""
+
+from backend.sdk import ProviderBuilder
+
+ayrshare = (
+    ProviderBuilder("ayrshare")
+    .with_api_key("_AYRSHARE_PROFILE_KEY_UNUSED", "Ayrshare Profile Key")
```
majdyz: 🤖 🟠 **Should Fix**: Passing `_AYRSHARE_PROFILE_KEY_UNUSED` as the env-var name is a sentinel hack — if someone ever sets that env var (CI, debug, typo), `ProviderBuilder.with_api_key` will happily create a default `APIKeyCredentials(provider="ayrshare", api_key=<that-value>)` and `resolve_block_credentials` will hand it to Ayrshare blocks as a "profile key" → silent mis-use of whatever random string is in the env. Cleaner options: (a) extend `ProviderBuilder` with a `managed_only()` method that registers `"api_key"` in `_supported_auth_types` but skips the env-var default-credential path; or (b) skip `with_api_key` entirely and seed `_supported_auth_types.add("api_key")` manually before `.build()`. Either avoids the magic-string contract.

autogpt_platform/backend/backend/integrations/managed_providers/ayrshare.py
```diff
@@ -0,0 +1,97 @@
...
+    Ayrshare API.
+    """
+    user_integrations = await get_user_integrations(user_id)
+    legacy_key = user_integrations.managed_credentials.ayrshare_profile_key
+    if legacy_key:
+        logger.debug("[ayrshare] Reusing legacy profile key for user %s", user_id)
+        return (
+            legacy_key.get_secret_value()
+            if isinstance(legacy_key, SecretStr)
+            else str(legacy_key)
```
majdyz, line 154: 🤖 🟠 **Should Fix**: Seconding CodeRabbit here — `_get_or_create_profile_key` reads the legacy `managed_credentials.ayrshare_profile_key` but nothing clears it. Because `_ensure_one` short-circuits on `has_managed_credential == True`, subsequent calls never re-enter this code path, so in steady state it is fine — but the PR body commits to "removing the field in a follow-up" and that follow-up will hit users who still have the field populated. Clear it here (wrapped in `edit_user_integrations`) as part of the migration read, so the follow-up removal has nothing to chase.

autogpt_platform/backend/backend/integrations/managed_providers/ayrshare.py
```diff
@@ -0,0 +1,97 @@
...
+    try:
+        client = AyrshareClient()
+    except MissingConfigError as exc:
+        raise RuntimeError("Ayrshare integration is not configured") from exc
+
+    logger.debug("[ayrshare] Creating profile for user %s", user_id)
+    profile = await client.create_profile(
+        title=f"User {user_id}", messaging_active=True
+    )
+    return profile.profileKey
```
majdyz, line 165: 🤖 🟠 **Should Fix**: `AyrshareManagedProvider` has no unit tests in this PR. At minimum cover: (a) fresh user — `provision()` calls `AyrshareClient.create_profile` and returns an `APIKeyCredentials(is_managed=True)`; (b) legacy user — `managed_credentials.ayrshare_profile_key` populated → `create_profile` is NOT called and the returned credential carries the legacy key; (c) `is_available()` returns False when either secret is missing. The AgentMail provider has no tests either, but adding them for the new entry prevents silent regressions in the legacy-migration path (which has no other coverage).

autogpt_platform/backend/backend/copilot/tools/e2b_sandbox.py
```diff
@@ -112,6 +143,38 @@ async def _clear_stored_sandbox_id(session_id: str) -> None:
...
+    ``gh auth status`` to check GitHub connection state.
+
+    Best-effort: on failure, log a warning and proceed.  The copilot's
+    existing ``connect_integration`` fallback still works for GitHub auth; a
+    missing ``gh`` just means the status check reports "not installed" and
+    the model prompts the user to connect.
+    """
+    try:
+        result = await sandbox.commands.run(
+            f"bash -c {shlex.quote(_SANDBOX_BOOTSTRAP_SCRIPT)}",
```
majdyz: 🤖 🟡 **Nice to Have**: Agreeing with CodeRabbit — `AsyncSandbox.commands.run()` already executes through a shell, so the `bash -c {shlex.quote(_SANDBOX_BOOTSTRAP_SCRIPT)}` wrapper is redundant. Passing `_SANDBOX_BOOTSTRAP_SCRIPT` directly lets you drop the `shlex` import (which becomes otherwise unused in this file). Not a correctness issue today, but removes one layer of quoting that a future reader has to reason about.

autogpt_platform/backend/backend/blocks/ayrshare/_util.py
```diff
@@ -4,22 +4,25 @@
...
 
+    credentials: CredentialsMetaInput = ayrshare.credentials_field(
+        description=(
+            "Ayrshare profile credential. AutoGPT provisions this managed "
+            "credential automatically — the user does not create it. After "
+            "it's in place, the user links each social account via the "
+            "Ayrshare SSO popup in the Builder."
+        ),
+    )
+
```
majdyz, line 25: 🤖 🔵 **Nit**: `profile_key=credentials.api_key.get_secret_value()` is now duplicated verbatim across all 13 Ayrshare blocks. A one-line helper in `_util.py` (`def profile_key(credentials: APIKeyCredentials) -> str: return credentials.api_key.get_secret_value()`) would centralise the extraction and make the blocks easier to scan. Not blocking, but it is the kind of incidental duplication that the AgentMail blocks avoid by going through a `_client(credentials)` helper.

autogpt_platform/backend/backend/copilot/tools/e2b_sandbox.py
majdyz: 🤖 Fixed in [65d80a61d](https://github.com/Significant-Gravitas/AutoGPT/commit/65d80a61ddc118738c3a9173f70e8276fd56eedd): dropped the `bash -c {shlex.quote(...)}` wrapper, now passing the script directly to `sandbox.commands.run()`. Removed the `shlex` import.

autogpt_platform/backend/backend/blocks/ayrshare/_util.py
majdyz: 🤖 Fixed in [f217385f1](https://github.com/Significant-Gravitas/AutoGPT/commit/f217385f1e9ef68beae43550ada05d09a207c1bf): added `get_profile_key(credentials) -> str` helper in `_util.py`, all 13 blocks now call it instead of repeating the `.api_key.get_secret_value()` chain.

autogpt_platform/backend/backend/blocks/ayrshare/_config.py
majdyz: 🤖 Fixed in [d40410ef5](https://github.com/Significant-Gravitas/AutoGPT/commit/d40410ef5957d4cba7f62811823d368c65039ec9): added `ProviderBuilder.with_managed_api_key()` that registers `api_key` in `_supported_auth_types` without touching env vars or default credentials. `_config.py` now uses it — no more sentinel env-var name.

autogpt_platform/backend/backend/integrations/managed_providers/ayrshare.py
majdyz: 🤖 Fixed in [b687cbd36](https://github.com/Significant-Gravitas/AutoGPT/commit/b687cbd365fac2afc6fcd2ea257790e3187831b9): `is_available()` now returns `False` unconditionally → `ensure_managed_credentials` sweep skips Ayrshare entirely. Added a new public `ensure_managed_credential(user_id, store, provider)` helper in `managed_credentials.py` that reuses the lock+upsert path but bypasses the `is_available` gate; the SSO-URL route calls it on demand so profiles are provisioned only when the user actually opens the connect flow — one profile per active user, not one per authenticated user.

autogpt_platform/backend/backend/integrations/managed_providers/ayrshare.py
```diff
@@ -0,0 +1,97 @@
...
+
+    async def provision(self, user_id: str) -> Credentials:
+        profile_key = await _get_or_create_profile_key(user_id)
+        return APIKeyCredentials(
+            provider=self.provider_name,
+            title="Ayrshare (managed by AutoGPT)",
+            api_key=SecretStr(profile_key),
+            expires_at=None,
+            is_managed=True,
+        )
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [b687cbd36](https://github.com/Significant-Gravitas/AutoGPT/commit/b687cbd365fac2afc6fcd2ea257790e3187831b9): `_get_or_create_profile_key` now clears `managed_credentials.ayrshare_profile_key` under the same `edit_user_integrations` lock as the read, so the legacy field is emptied on first migration. The eventual schema removal of the field has nothing to chase.

autogpt_platform/backend/backend/integrations/managed_providers/ayrshare.py
majdyz: 🤖 Fixed in [b687cbd36](https://github.com/Significant-Gravitas/AutoGPT/commit/b687cbd365fac2afc6fcd2ea257790e3187831b9): legacy field is now cleared under the `edit_user_integrations` lock at migration time.

autogpt_platform/backend/backend/integrations/managed_providers/ayrshare.py
majdyz: 🤖 Fixed in [b687cbd36](https://github.com/Significant-Gravitas/AutoGPT/commit/b687cbd365fac2afc6fcd2ea257790e3187831b9): added `ayrshare_test.py` — 7 tests covering `is_available()` opt-out, the legacy-key migration + clear, the fresh-profile create path, `provision()` returning a managed `APIKeyCredentials`, and the `_settings_available` pre-flight.

autogpt_platform/backend/backend/blocks/ayrshare/_util.py
```diff
@@ -4,22 +4,30 @@
...
     """Base input model for Ayrshare social media posts with common fields."""
 
+    credentials: CredentialsMetaInput = ayrshare.credentials_field(
+        description=(
+            "Ayrshare profile credential. AutoGPT provisions this managed "
+            "credential automatically — the user does not create it. After "
+            "it's in place, the user links each social account via the "
+            "Ayrshare SSO popup in the Builder."
+        ),
+    )
```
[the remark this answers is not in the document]
majdyz: 🤖 False positive. The new `credentials` field on `BaseAyrshareInput` follows the same required-no-default pattern every other managed/OAuth block uses (AgentMail, Linear, Firecrawl, etc.) — verified: `AgentMailListEntriesBlock.input_schema.get_missing_input({'direction': 'receive'})` also returns `{'credentials'}`. The executor's credentials-resolution step at [`manager.py:264-310`](https://github.com/Significant-Gravitas/AutoGPT/blob/dev/autogpt_platform/backend/backend/executor/manager.py#L264-L310) and the copilot's `resolve_block_credentials` pre-run path both populate the field before the block runs; old graphs that lack `credentials` in `input_default` are handled the same way AgentMail handles them. If this is actually a live bug it is framework-wide, not Ayrshare-specific.

autogpt_platform/backend/backend/copilot/tools/e2b_sandbox.py
```diff
@@ -90,6 +90,36 @@
...
+  x86_64) arch=amd64 ;;
+  aarch64|arm64) arch=arm64 ;;
+  *) echo "unsupported arch: $(uname -m)" >&2; exit 2 ;;
+esac
+tmp=$(mktemp -d)
+trap 'rm -rf "$tmp"' EXIT
+curl -fsSL "https://github.com/cli/cli/releases/download/v{_GH_CLI_VERSION}/gh_{_GH_CLI_VERSION}_linux_${{arch}}.tar.gz" -o "$tmp/gh.tgz"
+tar -xzf "$tmp/gh.tgz" -C "$tmp"
+sudo -n mv "$tmp"/gh_*/bin/gh /usr/local/bin/gh
+""".strip()
```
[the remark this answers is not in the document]
majdyz: 🤖 False positive. Verified against the e2b SDK source: `AsyncSandbox.commands.run()` always wraps the command in `/bin/bash -l -c <cmd>` ([`e2b/sandbox_async/commands/command.py:252-254`](https://github.com/e2b-dev/E2B/blob/main/packages/python-sdk/e2b/sandbox_async/commands/command.py#L252)), so multi-line scripts with `$(...)`, heredocs, and pipes work natively. The explicit `bash -c` wrapper was redundant; removing it doesn't change execution semantics. Bootstrap tests still pass.

autogpt_platform/backend/backend/integrations/managed_credentials.py
```diff
@@ -127,6 +111,67 @@ async def _ensure_one(
...
+    # is atomic across all workers, preventing duplicate external
+    # resource provisioning (e.g. AgentMail API keys).
+    locks = await store.locks()
+    key = (f"user:{user_id}", f"managed-provision:{name}")
+    async with locks.locked(key):
+        # Re-check under lock to avoid duplicate provisioning.
+        if await store.has_managed_credential(user_id, name):
+            return True
+        credential = await provider.provision(user_id)
+        await store.add_managed_credential(user_id, credential)
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [d9dcd521e](https://github.com/Significant-Gravitas/AutoGPT/commit/d9dcd521eaa4afdb2bb99f3052df1b6bb1f39670): moved the legacy-field clear out of `_read_or_create_profile_key` into a new `post_provision` hook on `ManagedCredentialProvider`. `_provision_under_lock` calls `post_provision` only after `add_managed_credential` returns — so if persistence fails, the legacy key stays populated and a retry can reuse it. Added a regression test that patches `add_managed_credential` to raise and asserts the legacy key survives, plus two tests for the new hook itself (populated-field clear + idempotent on already-cleared).

autogpt_platform/backend/backend/integrations/managed_providers/ayrshare.py
```diff
@@ -0,0 +1,128 @@
...
+    async with store.edit_user_integrations(user_id) as user_integrations:
+        legacy_key = user_integrations.managed_credentials.ayrshare_profile_key
+        if legacy_key:
+            logger.debug("[ayrshare] Migrating legacy profile key for user %s", user_id)
+            user_integrations.managed_credentials.ayrshare_profile_key = None
+            return (
+                legacy_key.get_secret_value()
+                if isinstance(legacy_key, SecretStr)
+                else str(legacy_key)
+            )
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [d9dcd521e](https://github.com/Significant-Gravitas/AutoGPT/commit/d9dcd521eaa4afdb2bb99f3052df1b6bb1f39670) (same commit sentry flagged above). The legacy-key clear moved out of `provision()` into a new `post_provision` hook that `_provision_under_lock` calls only after `add_managed_credential` persists successfully — so a failure between provision and persist leaves the legacy key intact and a retry reuses it. Regression test in [`ayrshare_test.py::test_add_managed_credential_failure_retains_legacy_key`](https://github.com/Significant-Gravitas/AutoGPT/blob/d9dcd521eaa4afdb2bb99f3052df1b6bb1f39670/autogpt_platform/backend/backend/integrations/managed_providers/ayrshare_test.py) covers exactly this scenario.

autogpt_platform/backend/backend/integrations/managed_providers/ayrshare.py
```diff
@@ -0,0 +1,156 @@
...
+    ``managed_credentials.ayrshare_profile_key`` is populated (pre-migration
+    data), it is reused verbatim so existing linked socials keep working.
+    The legacy field is *not* cleared here — that happens in
+    :meth:`AyrshareManagedProvider.post_provision`, which runs only after
+    the managed credential is durably stored.  If this function cleared
+    eagerly and the subsequent ``add_managed_credential`` failed, a retry
+    would see an empty legacy field and create a *fresh* Ayrshare profile,
+    orphaning the user's linked social accounts.
+    """
+    user_integrations = await store._get_user_integrations(user_id)
```
majdyz: 🤖 🟠 **Should Fix**: `_read_or_create_profile_key` calls `store._get_user_integrations(user_id)` — a leading-underscore (module-private) method. Reaching through the privacy fence couples Ayrshare to the store implementation; a future refactor that renames/removes `_get_user_integrations` silently breaks the migration read path. Either add a public accessor on `IntegrationCredentialsStore` (e.g. `get_user_integrations`) and use that here, or route the read through `edit_user_integrations` (it yields the same `UserIntegrations` object). The latter is the pattern used in `post_provision` a few lines up.

autogpt_platform/backend/backend/integrations/managed_providers/ayrshare.py
```diff
@@ -0,0 +1,156 @@
...
+        # provision on demand go through
+        # `ensure_managed_credential(user_id, store, AyrshareManagedProvider())`
+        # which bypasses this gate.
+        return False
+
+    async def provision(self, user_id: str) -> Credentials:
+        # Lazy import: avoids a managed_providers → credentials_store cycle
+        # and keeps the module importable at startup from
+        # managed_providers/__init__.py.
+        store = IntegrationCredentialsStore()
```
majdyz: 🤖 🟠 **Should Fix**: `AyrshareManagedProvider.provision()` constructs a fresh `IntegrationCredentialsStore()` on every call because the `ManagedCredentialProvider.provision` contract only receives `user_id` — but `post_provision` below it *does* receive the caller-supplied `store`. Result: inside a single lock, we hold two different store instances, and tests that mock `add_managed_credential` on the outer store have to also patch the `IntegrationCredentialsStore` constructor (see the regression test). Extending the base-class signature to `async def provision(self, user_id, store)` lets providers reuse the caller-supplied instance and keeps the two hooks symmetric.

autogpt_platform/backend/backend/integrations/managed_credentials.py
```diff
@@ -100,27 +118,87 @@ async def _ensure_one(
...
+        try:
+            await provider.post_provision(user_id, store, credential)
+        except Exception:
+            logger.warning(
+                "post_provision hook failed for provider=%s user=%s; "
+                "managed credential is persisted so retry is safe",
                 name,
                 user_id,
+                exc_info=True,
             )
```
majdyz, line 205: 🤖 🟠 **Should Fix**: `_provision_under_lock` swallows `post_provision` exceptions and only logs. Thats the right call for the current Ayrshare case (idempotent legacy-clear), but the abstract hook has no such constraint in its signature or docstring. Future providers implementing `post_provision` for e.g. scheduling a follow-up task or notifying an external service would have their failures silently dropped. Document on `ManagedCredentialProvider.post_provision` that it **must be idempotent and retry-safe**, or escalate failures to a telemetry path so they are not invisible in the logs.

autogpt_platform/backend/backend/integrations/managed_providers/ayrshare.py
```diff
@@ -0,0 +1,156 @@
...
+        raise RuntimeError("Ayrshare integration is not configured") from exc
+
+    logger.debug("[ayrshare] Creating profile for user %s", user_id)
+    profile = await client.create_profile(
+        title=f"User {user_id}", messaging_active=True
+    )
+    return profile.profileKey
+
+
+def _settings_available() -> bool:
```
majdyz: 🤖 🟡 **Nice to Have**: `_settings_available` is declared with a leading underscore (module-private convention) but imported from outside the module in `router.py` with an alias (`_settings_available as ayrshare_settings_available`). Either drop the underscore since it is part of the public API now, or move the check inside `AyrshareManagedProvider` as e.g. `AyrshareManagedProvider.configured()` — currently the import alias obscures that the router reaches into private module state.

autogpt_platform/backend/backend/integrations/managed_providers/ayrshare_test.py
```diff
@@ -0,0 +1,257 @@
...
+        """
+        # Even with both org secrets populated, is_available stays False —
+        # callers who actually want to provision use
+        # ensure_managed_credential() directly, bypassing this gate.
+        with patch(
+            "backend.integrations.managed_providers.ayrshare.Settings"
+        ) as mock_settings:
+            mock_settings.return_value.secrets.ayrshare_api_key = "api-key"
+            mock_settings.return_value.secrets.ayrshare_jwt_key = "jwt-key"
+            assert await AyrshareManagedProvider().is_available() is False
```
majdyz, line 44: 🤖 🔵 **Nit**: `test_is_available_always_false` patches `Settings` with valid secrets and asserts `is_available()` still returns `False`. But the current `is_available()` body no longer consults `Settings` at all — it just `return False`. The mock is dead code. Either simplify the test (just call the method with no patch) or re-test against a real-config scenario (unset envs) to catch future regressions if the method ever starts consulting settings again.

autogpt_platform/backend/backend/integrations/managed_providers/ayrshare.py
majdyz: 🤖 Fixed in [1698d8b96](https://github.com/Significant-Gravitas/AutoGPT/commit/1698d8b9611fcdf145eb58566b08a5cc4613f513): switched _get_user_integrations read to the public edit_user_integrations context-manager pattern (yields the same UserIntegrations object, no write-back since nothing mutates). Ayrshare no longer reaches past the store's privacy fence.

autogpt_platform/backend/backend/integrations/managed_providers/ayrshare.py
majdyz: 🤖 Fixed in [1698d8b96](https://github.com/Significant-Gravitas/AutoGPT/commit/1698d8b9611fcdf145eb58566b08a5cc4613f513): ManagedCredentialProvider.provision now takes (user_id, store); _provision_under_lock passes the caller-supplied store through. Ayrshare no longer constructs a fresh IntegrationCredentialsStore() per call. AgentMail accepts + ignores the param since it provisions via its own API client.

autogpt_platform/backend/backend/integrations/managed_credentials.py
majdyz: 🤖 Fixed in [1698d8b96](https://github.com/Significant-Gravitas/AutoGPT/commit/1698d8b9611fcdf145eb58566b08a5cc4613f513): post_provision docstring now states the contract explicitly: 'Must be idempotent and retry-safe. The framework swallows any exception raised here and only logs a warning.' Future subclasses won't assume the hook retries.

autogpt_platform/backend/backend/integrations/managed_providers/ayrshare.py
majdyz: 🤖 Fixed in [1698d8b96](https://github.com/Significant-Gravitas/AutoGPT/commit/1698d8b9611fcdf145eb58566b08a5cc4613f513): renamed _settings_available to settings_available (dropped the underscore since it's imported from outside the module). Router import is now a clean public reference with an alias purely for readability.

autogpt_platform/backend/backend/integrations/managed_providers/ayrshare_test.py
majdyz: 🤖 Fixed in [1698d8b96](https://github.com/Significant-Gravitas/AutoGPT/commit/1698d8b9611fcdf145eb58566b08a5cc4613f513): dropped the dead Settings mock — the test now just calls is_available() directly and asserts False. Matches the current impl which ignores Settings entirely.

autogpt_platform/backend/backend/integrations/managed_providers/ayrshare.py
```diff
@@ -0,0 +1,156 @@
...
+need to provision (currently only ``/api/integrations/ayrshare/sso_url``)
+invoke :func:`~backend.integrations.managed_credentials.ensure_managed_credential`
+with an instance of this provider, which re-uses the same distributed
+Redis lock + upsert path as AgentMail.
+
+Legacy compatibility: on first provision we migrate
+``UserIntegrations.managed_credentials.ayrshare_profile_key`` (pre-migration
+data) into the new managed credential and clear the legacy field in the
+same write — so the eventual removal of the legacy schema field has
+nothing to chase.
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [8782e84f7](https://github.com/Significant-Gravitas/AutoGPT/commit/8782e84f72f2b46f25ccc40c1cab2b80a95b82a0): updated the module docstring to reflect the actual ordering — managed credential is persisted first, legacy field cleared in post_provision afterwards, with a reference to `TestMigrationOrderingSafety`.

autogpt_platform/backend/backend/integrations/managed_providers/ayrshare.py
```diff
@@ -0,0 +1,163 @@
...
+    would see an empty legacy field and create a *fresh* Ayrshare profile,
+    orphaning the user's linked social accounts.
+
+    We borrow ``edit_user_integrations`` as a read — the context manager
+    yields the same ``UserIntegrations`` object as an internal getter would,
+    and the write-back is a no-op when no fields are mutated.  This keeps
+    us on the store's public API instead of reaching for a private
+    ``_get_user_integrations`` accessor.
+    """
+    async with store.edit_user_integrations(user_id) as user_integrations:
```
majdyz: 🤖 🟠 **Should Fix**: `_read_or_create_profile_key` uses `store.edit_user_integrations()` as a pseudo-read, but that context manager **always writes back** on exit (`await self.db_manager.update_user_integrations(...)`) regardless of whether the yielded object was mutated — the write-back is not guarded. Every `/ayrshare/sso_url` call now triggers three full DB write cycles on `user_integrations`: (1) this spurious no-op write, (2) `add_managed_credential`, (3) `post_provision`. Each cycle also acquires + releases the Redis `locked_user_integrations` lock. Add a public `get_user_integrations(user_id)` to `IntegrationCredentialsStore` (drop the underscore prefix on the existing accessor, or wrap it) and use it here — that is the cleanest way to honour the earlier review comment about private access without the spurious writes.

autogpt_platform/backend/backend/integrations/managed_providers/ayrshare.py
```diff
@@ -0,0 +1,163 @@
...
+    ``managed_credentials.ayrshare_profile_key`` is populated (pre-migration
+    data), it is reused verbatim so existing linked socials keep working.
+    The legacy field is *not* cleared here — that happens in
+    :meth:`AyrshareManagedProvider.post_provision`, which runs only after
+    the managed credential is durably stored.  If this function cleared
+    eagerly and the subsequent ``add_managed_credential`` failed, a retry
+    would see an empty legacy field and create a *fresh* Ayrshare profile,
+    orphaning the user's linked social accounts.
+
+    We borrow ``edit_user_integrations`` as a read — the context manager
```
majdyz: 🤖 🟡 **Nice to Have**: The "borrow `edit_user_integrations` as a read" line in the docstring is honest about the workaround but understates the cost — the context manager unconditionally writes the `user_integrations` row back on `__aexit__`, so calling it for a read triggers a real DB write plus a Redis-lock acquire/release. If the `get_user_integrations` public accessor in the companion comment is added, this block can drop the context-manager dance; until then, consider at least noting the write-back in the comment so the overhead is not invisible.

autogpt_platform/backend/backend/integrations/managed_providers/ayrshare.py
```diff
@@ -0,0 +1,163 @@
...
+    ) -> None:
+        """Clear the legacy ``ayrshare_profile_key`` side-channel after migration.
+
+        Runs only after the managed credential is durably stored.  If the
+        legacy field is still populated, the profile key now lives in two
+        places — clear the old one so the eventual schema removal has
+        nothing to chase.  A failure here is harmless: the managed
+        credential is already persisted, subsequent provision calls
+        short-circuit, and the legacy field stays until the next successful
+        migration attempt.
```
majdyz: 🤖 🔵 **Nit**: The `post_provision` docstring here duplicates half of what the base-class contract now says (the idempotency + failure-is-harmless paragraph). Once the base-class docstring establishes the contract, the override can be terse: "Clear the legacy ``ayrshare_profile_key`` side-channel. See base-class docstring for retry semantics." Keeps the detail in one place so it cannot drift.

autogpt_platform/backend/backend/integrations/managed_providers/ayrshare.py
majdyz: 🤖 Fixed in [1d68726e4](https://github.com/Significant-Gravitas/AutoGPT/commit/1d68726e4c433568458f21865268787d41b355c9): added public `IntegrationCredentialsStore.get_user_integrations(user_id)`, a thin wrapper over the existing private accessor. `_read_or_create_profile_key` now uses it — one DB read instead of the full edit cycle (read + no-op write + Redis lock acquire/release). Ayrshare SSO URL path drops from three write cycles per call to two.

autogpt_platform/backend/backend/integrations/managed_providers/ayrshare.py
majdyz: 🤖 Fixed in [1d68726e4](https://github.com/Significant-Gravitas/AutoGPT/commit/1d68726e4c433568458f21865268787d41b355c9): replaced the edit-as-read dance with the new public `get_user_integrations` accessor, so the docstring no longer needs to warn about hidden costs — there are none.

autogpt_platform/backend/backend/integrations/managed_providers/ayrshare.py
majdyz: 🤖 Fixed in [1d68726e4](https://github.com/Significant-Gravitas/AutoGPT/commit/1d68726e4c433568458f21865268787d41b355c9): trimmed the override docstring to just what it does and a cross-reference to the base-class contract for retry semantics. The canonical wording now lives in one place.

autogpt_platform/backend/backend/copilot/tools/e2b_sandbox.py
```diff
@@ -231,6 +295,10 @@ async def get_or_create_sandbox(
                 raise last_exc
 
             assert sandbox is not None  # guaranteed: last_exc is None iff break was hit
+            # Install gh CLI etc. on first create so the prompt's `gh auth
+            # status` check works reliably.  Filesystem persists across
+            # pause/resume, so this runs at most once per session.
+            await _bootstrap_sandbox(sandbox, session_id)
```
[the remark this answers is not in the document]
majdyz: 🤖 Not applicable — the `_bootstrap_sandbox` code the race referenced has been reverted in [`4065fef32`](https://github.com/Significant-Gravitas/AutoGPT/commit/4065fef32) + [`04ee2994f`](https://github.com/Significant-Gravitas/AutoGPT/commit/04ee2994f). Empirical benchmarks against your E2B account (3/3 fresh sandboxes) confirmed `gh` is already preinstalled in `e2bdev/base` (v2.23.0, at `/usr/bin/gh`), so the runtime install was a no-op. The sandbox creation time budget (`_CREATION_LOCK_TTL = 120`, `_MAX_WAIT_ATTEMPTS` derived ≈ 144s) is back to its pre-PR values and stays within the original safety margin.

autogpt_platform/backend/backend/blocks/ayrshare/_util.py
```diff
@@ -4,22 +4,30 @@
...
 
-async def get_profile_key(user_id: str):
-    user_integrations: UserIntegrations = (
-        await get_database_manager_async_client().get_user_integrations(user_id)
-    )
-    return user_integrations.managed_credentials.ayrshare_profile_key
+
+def get_profile_key(credentials: APIKeyCredentials) -> str:
+    """Extract the Ayrshare profile key from the resolved managed credential."""
+    return credentials.api_key.get_secret_value()
```
Pwuts: nit: we can get rid of this because it doesn't really do anything anymore

autogpt_platform/backend/backend/api/features/integrations/router.py
```diff
@@ -798,66 +813,57 @@ async def get_ayrshare_sso_url(
...
                 SocialPlatform.INSTAGRAM,
                 SocialPlatform.YOUTUBE,
-                # SocialPlatform.REDDIT,
-                # SocialPlatform.TELEGRAM,
-                # SocialPlatform.GOOGLE_MY_BUSINESS,
-                # SocialPlatform.PINTEREST,
                 SocialPlatform.TIKTOK,
-                # SocialPlatform.BLUESKY,
-                # SocialPlatform.SNAPCHAT,
-                # SocialPlatform.THREADS,
```
Pwuts, line 1154: ?

autogpt_platform/backend/backend/integrations/managed_providers/ayrshare.py
```diff
@@ -0,0 +1,152 @@
...
+
+    async def is_available(self) -> bool:
+        # Opt out of the `ensure_managed_credentials` startup sweep.  Profile
+        # creation has real per-user quota cost — we only want to create one
+        # when the user explicitly requests a social-media connection via
+        # `/api/integrations/ayrshare/sso_url`.  Callers that *do* want to
+        # provision on demand go through
+        # `ensure_managed_credential(user_id, store, AyrshareManagedProvider())`
+        # which bypasses this gate.
+        return False
```
Pwuts: isn't this supposed to indicate whether Ayrshare env vars are set?

autogpt_platform/backend/backend/integrations/managed_providers/ayrshare.py
```diff
@@ -0,0 +1,152 @@
...
+    try:
+        client = AyrshareClient()
+    except MissingConfigError as exc:
+        raise RuntimeError("Ayrshare integration is not configured") from exc
+
+    logger.debug("[ayrshare] Creating profile for user %s", user_id)
+    profile = await client.create_profile(
+        title=f"User {user_id}", messaging_active=True
+    )
+    return profile.profileKey
```
Pwuts: So if the *legacy* key isn't set, it always creates a new profile? What's the new key storage location? How do we prevent creating multiple accounts per user?

autogpt_platform/backend/backend/blocks/ayrshare/_util.py
majdyz: 🤖 Fixed in [34c69d763](https://github.com/Significant-Gravitas/AutoGPT/commit/34c69d76368f0b6b8f797c8f56f55e74b4ab7542): dropped `get_profile_key` helper and inlined `credentials.api_key.get_secret_value()` at the 13 call sites. The helper was a one-liner that added no value.

autogpt_platform/backend/backend/api/features/integrations/router.py
majdyz: 🤖 Fixed in [34c69d763](https://github.com/Significant-Gravitas/AutoGPT/commit/34c69d76368f0b6b8f797c8f56f55e74b4ab7542): clarified the comment — the `allowed_social` list is a deliberate per-platform rollout gate (each needs verification against the block + Ayrshare's network-specific quirks before the user-visible surface widens), not "one at a time in time".

autogpt_platform/backend/backend/integrations/managed_providers/ayrshare.py
majdyz: 🤖 Fixed in [34c69d763](https://github.com/Significant-Gravitas/AutoGPT/commit/34c69d76368f0b6b8f797c8f56f55e74b4ab7542): `is_available()` is now truthful — returns True when `AYRSHARE_API_KEY` + `AYRSHARE_JWT_KEY` are both set. The opt-out from the startup sweep moved to a new `auto_provision: bool = True` class attr on `ManagedCredentialProvider`; Ayrshare sets `auto_provision = False`. `_ensure_one` gates on `auto_provision` before `is_available`, so the framework's availability check stops lying.

autogpt_platform/backend/backend/integrations/managed_providers/ayrshare.py
majdyz: 🤖 Fixed in [34c69d763](https://github.com/Significant-Gravitas/AutoGPT/commit/34c69d76368f0b6b8f797c8f56f55e74b4ab7542): real idempotency now. Before `create_profile`, `_read_or_create_profile_key` calls the new `AyrshareClient.list_profiles()` and reuses any upstream profile titled `User {user_id}`. This closes the failure path you flagged — previously, if `add_managed_credential` raised after `create_profile` succeeded, the retry would create another profile and orphan the first. Key storage is unchanged: `APIKeyCredentials(provider="ayrshare", is_managed=True)` in `UserIntegrations.credentials`, same as AgentMail. Covered by new `test_reuses_upstream_profile_after_failed_persist_retry`.

autogpt_platform/backend/backend/integrations/ayrshare.py
```diff
@@ -295,6 +295,42 @@ async def create_profile(
...
+
+        Docs: https://www.ayrshare.com/docs/apis/profiles/get-profiles
+        """
+        response = await self._requests.get(self.PROFILES_ENDPOINT)
+        if not response.ok:
+            try:
+                error_message = response.json().get("message", "Unknown error")
+            except json.JSONDecodeError:
+                error_message = response.text()
+            raise AyrshareAPIException(
```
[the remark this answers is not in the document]
majdyz: Addressed in 574baedf6 — added an explicit `status == "error"` check to surface the upstream `message` instead of the generic shape-validation error. Consistent with generate_jwt / create_profile / create_post.

autogpt_platform/backend/backend/integrations/ayrshare.py
```diff
@@ -295,6 +295,47 @@ async def create_profile(
...
+            payload.get("profiles", payload) if isinstance(payload, dict) else payload
+        )
+        if not isinstance(raw_profiles, list):
+            raise AyrshareAPIException(
+                f"Unexpected Ayrshare profiles response shape: {type(raw_profiles).__name__}",
+                response.status,
+            )
+        # Some profile entries may lack optional fields — let Pydantic reject
+        # malformed rows so we notice schema drift upstream.
+        return [ProfileResponse(**raw) for raw in raw_profiles]
```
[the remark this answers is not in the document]
majdyz: Addressed in ebb008840 (rebased). Sentry's concern is correct in substance (the specific failing field is `status`, not `profileKey` — GET /profiles list items omit the envelope `status` that POST /profiles returns). Split the model: `ProfileResponse` stays for `create_profile`, new `ProfileSummary` (title/refId/profileKey + optional messagingActive) for `list_profiles`. Existing idempotency test now constructs the real `ProfileSummary` to catch schema drift.

autogpt_platform/backend/backend/blocks/ayrshare/post_to_youtube.py
```diff
@@ -42,14 +43,6 @@ class Input(BaseAyrshareInput):
...
-        # selects the 5-credit video tier instead of the 2-credit image tier.
-        is_video: bool = SchemaField(
-            description="Whether the media is a video (always True for YouTube)",
-            default=True,
-            advanced=True,
-        )
-
         # YouTube-specific required options
         title: str = SchemaField(
             description="Video title (max 100 chars, required). Cannot contain < or > characters.",
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [941ca9892](https://github.com/Significant-Gravitas/AutoGPT/commit/941ca9892f71a5fc667fd42e5eb71161d8bade87): restored the `is_video=True` override on both PostToYouTube and PostToSnapchat. The override was lost during the rebase onto dev (#12893 merged independently). The `@cost(*AYRSHARE_POST_COSTS)` filter now correctly selects the 5-credit video tier for both blocks.

autogpt_platform/backend/backend/integrations/ayrshare.py
```diff
@@ -295,6 +308,45 @@ async def create_profile(
...
+            )
+        raw_profiles = (
+            payload.get("profiles", payload) if isinstance(payload, dict) else payload
+        )
+        if not isinstance(raw_profiles, list):
+            raise AyrshareAPIException(
+                f"Unexpected Ayrshare profiles response shape: {type(raw_profiles).__name__}",
+                response.status,
+            )
+        return [ProfileSummary(**raw) for raw in raw_profiles]
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [941ca9892](https://github.com/Significant-Gravitas/AutoGPT/commit/941ca9892f71a5fc667fd42e5eb71161d8bade87): passed `raise_for_status=False` to the Requests constructor so every method's custom `response.ok` + status-envelope handling runs as intended and raises AyrshareAPIException with the upstream message.

autogpt_platform/backend/backend/integrations/ayrshare.py
```diff
@@ -295,6 +310,45 @@ async def create_profile(
...
+
```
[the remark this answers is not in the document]
majdyz: 🤖 Already fixed in [eaa3efc6b](https://github.com/Significant-Gravitas/AutoGPT/commit/eaa3efc6b1dde55b78ed27d1c6c4a3a6b197b817) — Sentry self-resolved in the follow-up comment. Resolving the thread.

autogpt_platform/backend/backend/api/features/integrations/router.py
```diff
@@ -230,9 +236,10 @@ async def callback(
 async def list_credentials(
     user_id: Annotated[str, Security(get_user_id)],
 ) -> list[CredentialsMetaResponse]:
-    # Fire-and-forget: provision missing managed credentials in the background.
-    # The credential appears on the next page load; listing is never blocked.
-    asyncio.create_task(ensure_managed_credentials(user_id, creds_manager.store))
+    # Block on provisioning so managed credentials appear on the first load
+    # instead of after a refresh.  `ensure_managed_credentials` runs providers
+    # concurrently and short-circuits via `_provisioned_users` on repeat calls.
+    await ensure_managed_credentials(user_id, creds_manager.store)
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [9cc9032d2](https://github.com/Significant-Gravitas/AutoGPT/commit/9cc9032d2d30dd63a4b2b97baffb0df1d036cab3): wrapped the sweep in `asyncio.wait_for` with a 10s timeout. On timeout we log a warning and kick off a fire-and-forget background task so provisioning still completes — same guarantee as the old FAF path, but only as a fallback when upstream is actually slow. Repeat calls still short-circuit via `_provisioned_users`.

autogpt_platform/backend/backend/integrations/ayrshare.py
```diff
@@ -295,6 +315,45 @@ async def create_profile(
...
+        response = await self._requests.get(self.PROFILES_ENDPOINT)
+        if not response.ok:
+            try:
+                error_message = response.json().get("message", "Unknown error")
+            except json.JSONDecodeError:
+                error_message = response.text()
+            raise AyrshareAPIException(
+                f"Ayrshare API request failed ({response.status}): {error_message}",
+                response.status,
+            )
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [f80af24e1](https://github.com/Significant-Gravitas/AutoGPT/commit/f80af24e1f5b781e9fa65faf807d08fd87267c32): extracted `_extract_error_message(response)` which isinstance-checks the decoded body and falls back to `response.text()` for non-dict shapes. Applied at all four Ayrshare error-handling sites (lines 219, 297, 329, and 540 in the pre-refactor code).

autogpt_platform/backend/backend/integrations/ayrshare.py
```diff
@@ -295,6 +321,63 @@ async def create_profile(
...
+        raw_profiles = (
+            payload.get("profiles", payload) if isinstance(payload, dict) else payload
+        )
+        if not isinstance(raw_profiles, list):
+            raise AyrshareAPIException(
+                f"Unexpected Ayrshare profiles response shape: {type(raw_profiles).__name__}",
+                response.status,
+            )
+        return [ProfileSummary(**raw) for raw in raw_profiles]
```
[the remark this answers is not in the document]
majdyz: 🤖 Addressed in [7c623dfec](https://github.com/Significant-Gravitas/AutoGPT/commit/7c623dfec3f43e10fc4b31f778b25b4a676bcac0): removed `list_profiles` (and `ProfileSummary`) entirely — the unique-title approach means we no longer need to enumerate existing profiles for recovery, so the ValidationError path is gone.

autogpt_platform/backend/backend/integrations/ayrshare.py
```diff
@@ -295,6 +321,63 @@ async def create_profile(
...
+        """
+        response = await self._requests.delete(
+            self.PROFILES_ENDPOINT, json={"title": title}
+        )
+        if not response.ok:
+            raise AyrshareAPIException(
+                f"Ayrshare profile delete failed ({response.status}): "
+                f"{_extract_error_message(response)}",
+                response.status,
+            )
```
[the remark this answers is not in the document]
majdyz: 🤖 Addressed in [7c623dfec](https://github.com/Significant-Gravitas/AutoGPT/commit/7c623dfec3f43e10fc4b31f778b25b4a676bcac0): confirmed via Ayrshare docs that DELETE /profiles requires the Profile-Key header we don't have for orphans, so the delete+recreate recovery was never viable. Removed `delete_profile` and replaced the whole strategy with unique-title creation (`secrets.token_hex(3)` suffix) — collisions with orphaned upstream profiles can't happen anymore. Orphans linger in the Ayrshare dashboard until manually cleaned up.

PR #12887, opened 2026-04-22
fix(backend/copilot): disable ask_question tool pending UX rework

majdyz:
### Why / What / How

**Why:** The in-conversation Question GUI is unreliable in production — users submitting answers can get their messages dropped and the agent gets stuck on the auto-generated "please proceed" step with no way to make progress. Discord report: https://discord.com/channels/1126875755960336515/1496474512966029472/1496537943287005365 (see attached video). Pause/queue semantics still need a rework; until then, the right call is to stop the model from reaching for this tool.

**What:** Removes `ask_question` from the copilot tool registry so the model never sees or calls it. Historical sessions that already contain `ask_question` tool calls still render (frontend renderers + response model untouched), so this is non-destructive to existing chats. Re-enabling once UX is reworked is a small revert.

**How:**
- Drop the `AskQuestionTool` import + registry entry from `backend/copilot/tools/__init__.py`.
- Drop `"ask_question"` from the `ToolName` literal in `backend/copilot/permissions.py` — required because a runtime consistency check asserts the literal matches `TOOL_REGISTRY.keys()`.
- Delete the "Clarifying — Before or During Building" section from `backend/copilot/sdk/agent_generation_guide.md` so the SDK-mode system prompt no longer instructs the model to call `ask_question`.
- Drop the three `prompting_test.py` tests that asserted the guide mentions that section.
- Keep `ask_question.py`, its unit test, `ClarificationNeededResponse`, and the frontend `AskQuestion`/`ClarificationQuestionsCard` components untouched so old sessions still render and re-enabling is a small revert.

### Changes 🏗️

- `backend/copilot/tools/__init__.py` — remove `AskQuestionTool` import and `"ask_question"` entry in `TOOL_REGISTRY`.
- `backend/copilot/permissions.py` — remove `"ask_question"` from the `ToolName` literal.
- `backend/copilot/sdk/agent_generation_guide.md` — remove the "Clarifying — Before or During Building" section.
- `backend/copilot/prompting_test.py` — remove `TestAgentGenerationGuideContainsClarifySection` and the now-unused `Path` import.

#### For code changes:
- [x] I have made a test plan
  - [x] `poetry run pytest backend/copilot/tools/ backend/copilot/permissions_test.py backend/copilot/prompting_test.py` — 805+78 tests pass, consistency check between `ToolName` literal and `TOOL_REGISTRY` still holds.
  - [ ] Smoke-test in dev: start a copilot session and confirm the model no longer lists/calls `ask_question` (its OpenAI tool schema is gone from `get_available_tools()` and from the SDK `allowed_tools`).
  - [ ] Load a historical session that contains an `ask_question` tool call in its transcript — confirm the frontend still renders the question card (no regression on legacy sessions).

### DIFF
--- a/autogpt_platform/backend/backend/copilot/permissions.py
+++ b/autogpt_platform/backend/backend/copilot/permissions.py
@@ -71,7 +71,6 @@ if TYPE_CHECKING:
 ToolName = Literal[
     # Platform tools (must match keys in TOOL_REGISTRY)
     "add_understanding",
-    "ask_question",
     "bash_exec",
     "browser_act",
     "browser_navigate",
--- a/autogpt_platform/backend/backend/copilot/prompting_test.py
+++ b/autogpt_platform/backend/backend/copilot/prompting_test.py
@@ -1,7 +1,6 @@
-"""Tests for agent generation guide — verifies clarification section."""
+"""Tests for prompting helpers."""
 
 import importlib
-from pathlib import Path
 
 from backend.copilot import prompting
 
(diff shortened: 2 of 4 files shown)


autogpt_platform/backend/backend/copilot/permissions.py
```diff
@@ -71,7 +71,6 @@
 ToolName = Literal[
     # Platform tools (must match keys in TOOL_REGISTRY)
     "add_understanding",
```
majdyz, line 73: 🤖 🟡 **Nice to Have**: removing `"ask_question"` from the `ToolName` literal means any persisted `AutoPilotBlock` node whose `input_default.tools` includes the string `"ask_question"` will now fail Pydantic validation on load (list item not in enum). Practically narrow — `ask_question` is a conversation tool, not something a typical sub-agent allow/deny list would reference — but worth a quick audit of `AgentNode.input_default` JSON in prod before merging, or accepting the risk and noting it here.

majdyz: 🤖 Accepting the risk. Rationale: (1) `ask_question` is a conversation-level tool — the model asks the user directly via the GUI — and sub-agents running inside `AutoPilotBlock` are non-interactive by design, so it should never appear in a realistic allow/deny list; (2) the default `tools: []` + `tools_exclude: True` ("deny nothing") is what the vast majority of nodes ship with, so the blast radius is limited to nodes someone hand-configured; (3) if a stale config does trip Pydantic on load, the error is loud and actionable — the user just edits the node to drop `ask_question` from the list, which was a no-op anyway since the tool is no longer registered. No DB audit planned for this PR.

PR #12889, opened 2026-04-22
fix(backend/copilot): tame Kimi compaction storm + tunable threshold + Langfuse cost backfill

majdyz:
## Why

Investigation of two reported sessions ([85804387](https://dev-builder.agpt.co/copilot?sessionId=85804387-7708-4fdc-8ec9-64283cdd902d), [19d69dec](https://dev-builder.agpt.co/copilot?sessionId=19d69dec-210f-4439-a94b-2d7d443b9909)) where Kimi K2.6 via OpenRouter was running ~30 min per turn with no actions completed (Discord report from Toran). Langfuse traces showed:

- 31 generation calls per turn at p90 = 151s, max = 415s
- 2.57M uncached tokens, `cache_create=0`, ~4% cache_read — Moonshot's OpenRouter endpoint silently drops Anthropic-style cache writes
- **3 SDK-internal compactions per turn** — each compaction is itself a slow LLM round-trip
- Reconciled OpenRouter cost was being recorded to a DB row but never surfaced on the Langfuse trace, leaving operators to grep pod logs

## What

Four commits, split by concern.

### 1. `fix(backend/copilot): skip CLAUDE_AUTOCOMPACT_PCT_OVERRIDE for Moonshot/Kimi` (`5fd9c5aa`)

`env.py` was unconditionally setting `CLAUDE_AUTOCOMPACT_PCT_OVERRIDE=50` (introduced in #12747 to cap cache-creation cost on Anthropic where context >200K = 54% of total cost). On Kimi where `cache_create=0` silently, the cache-cost rationale doesn't apply — but the 50% threshold still made the bundled CLI auto-compact at ~100K tokens, triggering 3+ compactions per turn against Kimi's larger effective window. Each compaction added a slow LLM round-trip (one in our test ran 166s and burned the budget cap before the user got any output).

Threads the resolved `sdk_model` (and `fallback_model`) into `build_sdk_env` and skips the env var when the model matches `is_moonshot_model(...)`. The CLI then uses its default ~93% threshold, cutting compaction passes to 0–1.

### 2. `feat(backend/copilot): backfill OpenRouter reconciled cost to Langfuse trace` (`f3de3624` + follow-ups `5ce3d038`, `d2c1a2cd`, `d8e08525`, `d243bf6c9`)

`record_turn_cost_from_openrouter` runs as a fire-and-forget task after the OTel span closes, so the Langfuse trace UI showed the SDK CLI's rate-card estimate only — for non-Anthropic OpenRouter routes that estimate is Sonnet pricing on Kimi tokens (~5x too high).

The backfill captures `langfuse.get_current_trace_id()` and threads it into the reconcile task, which emits an `openrouter-cost-reconcile` child event with the authoritative cost + token usage. **Bug caught during /pr-test:** `propagate_attributes` only annotates an existing OTel span, it doesn't create one — by the time the `finally` block runs, SDK-emitted spans have ended and `get_current_trace_id()` returns None. Fixed in `d8e08525` by wrapping the turn in `langfuse.start_as_current_span(name="copilot-sdk-turn")`. Also tags fallback-path events with `cost_source` so operators can distinguish reconciled vs estimated turns.

### 3. `feat(backend/copilot): expose CLAUDE_AUTOCOMPACT_PCT_OVERRIDE as a config knob` (`72416f73`)

The previously-hardcoded `50` is now `claude_agent_autocompact_pct_override` (default 50, env `CHAT_CLAUDE_AGENT_AUTOCOMPACT_PCT_OVERRIDE`). Setting to 0 omits the env var entirely so the CLI uses its native ~93% threshold — useful when the post-compact floor (system prompt + tool defs ≈ 65–110K) sits close to an aggressive trigger and operators see back-to-back compaction cascades. Moonshot routes still skip the env var unconditionally regardless of config.

### 4. `fix(backend/copilot): align SDK retry compaction target with CLI autocompact threshold` (`730ad256`)

`_reduce_context` was calling `compact_transcript` without an explicit `target_tokens`, so it fell back to `get_compression_target(model) = context_window - 60K`. For Sonnet 200K that's 140K — well above the CLI's PCT=50 trigger of 90K — and for Kimi 256K it's 196K, above the CLI's default 167K trigger. Result: a successful retry compaction landed at 140K/196K and the CLI immediately re-compacted on the next call → **two compactions per recovered turn**.

New `_compaction_target_tokens(model)` mirrors the CLI's `i6_()` formula (`min(window * pct/100, window - 13K)`) with a 20K safety buffer so the post-compact context sits comfortably below the CLI's trigger.

## How — empirical validation against the actual long Kimi transcript

Replayed the 199-message transcript from session 85804387 through the bundled CLI in two configurations:

| | Post-fix (no override) | Pre-fix (`PCT_OVERRIDE=50`) |
|---|---|---|
| `autocompact: tokens=` | 126,312 | 126,341 |
| `threshold=` | **167,000** | **90,000** |
| Decision | 126K < 167K → **skip** | 126K > 90K → **COMPACTION FIRES** |
| Duration | 21s | **166s** (8x slower) |
| Cost | $0.34 | **$0.82** (2.4x more) |
| Output | PONG (success) | empty (hit $0.50 budget cap, exit 1) |

The pre-fix configuration burned $0.82 of compaction work over 166s and never produced a user response — exactly the failure mode reported.

**Why cascade happens at 50%, not at 93%:** post-compaction context is `summary (~5–10K) + system_prompt + tool_definitions + skills + active TodoWrite + memory ≈ 65–110K floor`. With trigger at 90K, post-compact floor sits AT or above the trigger → next assistant message tips over → immediate re-compaction → cascade until the CLI's rapid-refill breaker trips at 3 attempts. With trigger at 167K, the same floor sits comfortably below trigger → no cascade.

## Considered but not done

- **Force `cache_control` markers to reach Moonshot**: bundled CLI sends them by default; Moonshot silently drops them per their own docs (uses `X-Msh-Context-Cache` headers, not body markers). Real fix needs bypassing OpenRouter — out of scope.
- **Slim the system prompt + tool definitions** to lower the post-compact floor: real win but separate refactor with tool-use accuracy A/B.
- **LD-driven auto-fallback to Sonnet on Kimi degradation**: `claude_agent_fallback_model` already wires `--fallback-model` for overload (529); auto-flipping on slowness needs latency aggregation infra that doesn't exist yet.

## Test plan

- [x] `poetry run pytest backend/copilot/sdk/env_test.py backend/copilot/sdk/openrouter_cost_test.py backend/copilot/sdk/service_helpers_test.py` — 111 passed (37 env + 23 cost + 51 helpers, including 6 new env tests, 3 backfill tests, 6 new compaction-target tests)
- [x] `poetry run pytest backend/copilot/sdk/` — 970+ passed
- [x] `poetry run pyright .` — 0 errors
- [x] `poetry run format` — clean
- [x] /pr-test --fix end-to-end against dev — 5/5 scenarios PASS, including Anthropic route ($0.0174 cost +0.0% delta) and Moonshot route ($0.028 vs $0.018 → +58.2% delta validates reconcile rationale)
- [x] Transcript replay validation: pre-fix vs post-fix on real 126K-token transcript → 8x slower / 2.4x more expensive / fails entirely on pre-fix; clean PONG on post-fix


majdyz:
🤖 @coderabbitai Both outside-diff findings fixed in [5ce3d0388](https://github.com/Significant-Gravitas/AutoGPT/commit/5ce3d03883beb9e53e52179c15ae44df3a79a0b0):

- `openrouter_cost.py`: `openrouter-cost-reconcile` event now carries `cost_source` (`"openrouter"` vs `"fallback"`) and `resolved_generation_id_count` so operators can tell authoritative reconciliations from rate-card fallbacks in Langfuse.
- `service.py`: cost-reconcile task is now anchored to `_background_tasks` (with `add_done_callback(_background_tasks.discard)`), mirroring the existing pattern at lines 3063 / 4232 / 4256, so the task can't be garbage-collected before persisting cost or emitting the Langfuse backfill.

majdyz:
🤖 @coderabbitai All three nitpicks fixed in [d2c1a2cd4](https://github.com/Significant-Gravitas/AutoGPT/commit/d2c1a2cd426c8e8e6988409d3e843c52470c5713):

- `service.py`: hoisted `from langfuse import get_client` to the top-level import alongside `propagate_attributes`.
- `openrouter_cost.py`: same hoist; `langfuse` is now a top-level import on the module.
- `openrouter_cost_test.py`: dropped `# noqa: ARG001` from the new `_get` mocks I added (lines 533/587/632/670) in favour of underscore-prefixed parameters; updated the patches to target `backend.copilot.sdk.openrouter_cost.get_client` since the import was hoisted. The pre-existing `# noqa: ARG001` callbacks elsewhere in this file were not touched (out of scope for this PR; the bodies use `kwargs`).

majdyz:
## E2E Test Report — PR #12889 (Kimi compaction + Langfuse cost backfill)

### Test environment

- Native mode: `poetry run app` + `pnpm dev`, infra (postgres/redis/rabbitmq/clamav) via docker
- Backend env: `CHAT_USE_CLAUDE_CODE_SUBSCRIPTION=false`, `CHAT_API_KEY=$OPEN_ROUTER_API_KEY`, `CHAT_BASE_URL=https://openrouter.ai/api/v1`
- Langfuse: `cloud.langfuse.com` (`LANGFUSE_ENVIRONMENT=local`)
- Test user: `test@test.com` (signup → onboarding bypass → token)

### Results

| 1 | `build_sdk_env(model=...)` per-model behaviour (unit-level) | PASS | Anthropic → `CLAUDE_AUTOCOMPACT_PCT_OVERRIDE=50`; Moonshot/Kimi → omitted; `None` → `50` (subscription default unchanged) |
| 2 | Anthropic-route turn end-to-end (`CHAT_CLAUDE_AGENT_MODEL=anthropic/claude-sonnet-4-6`) | PASS | Stream returned `PONG`; cost $0.0174 reconciled `+0.0%` against rate-card; Langfuse trace `0d9e2a69…` shows `copilot-sdk-turn → claude.conversation → claude.assistant.turn → openrouter-cost-reconcile` |
| 3 | Moonshot/Kimi-route turn (`CHAT_CLAUDE_AGENT_MODEL=moonshotai/kimi-k2.6`) | PASS | Stream returned `KIMI`; real cost $0.0282 vs CLI rate-card $0.0178 → **+58.2% delta** (validates the PR's blindspot rationale); Langfuse trace `3c910554…` includes `openrouter-cost-reconcile` event with `cost_source=openrouter`, `provider=open_router`, `model=moonshotai/kimi-k2.6` |
| 4 | Compaction storm cap on Kimi route | PASS | Turn completed with `compaction={attempt_count: 0, completed_count: 0}` — autocompact override correctly skipped, no compaction-driven slowness |
| 5 | Existing `claude.assistant.turn` Langfuse spans intact (no OTel pipeline regression) | PASS | All SDK-emitted spans (`claude.conversation`, `claude.assistant.turn`) still present and properly nested under the new `copilot-sdk-turn` parent span on both routes |

### Bug found and fixed during testing

The PR's `langfuse_trace_id = get_client().get_current_trace_id()` (added at `service.py:4092`) **always returned `None`** in production:

- `propagate_attributes` only annotates an existing OTel span — it does **not** create one.
- By the time the streaming generator exits and the `finally` block runs, the SDK-emitted spans (`claude.conversation` / `claude.assistant.turn`) have already ended.
- `get_current_trace_id()` therefore returned `None`, the `if langfuse_trace_id and real_cost is not None` guard short-circuited, and the `openrouter-cost-reconcile` event **never fired**.

**Diagnostic logs from the run before the fix:**
```
[SDK][7959eec8-e14] OTEL span at finally: span=NonRecordingSpan(SpanContext(trace_id=0x00000000…0000000)) recording=False
[SDK][7959eec8-e14][T1][cost-record] Langfuse backfill check: trace_id=None real_cost=0.01743135
```

**Fix** (commit `d8e08525`): wrap the entire turn in `get_client().start_as_current_span(name="copilot-sdk-turn")` so we own a span that lives across the `finally` block, giving us a stable trace_id. The SDK-emitted spans become children of this parent.

After the fix, the same code path emits the event:
```
[SDK][98c444e7-05f][T1][cost-record] Langfuse backfill check: trace_id=0d9e2a69882eb05fd5790143c0893f3e real_cost=0.01746135
```

### Langfuse evidence (queried from `/api/public/observations` after the fix)

## Anthropic-route trace 0d9e2a69882eb05fd5790143c0893f3e
```json
[
  {
    "type": "EVENT",
    "name": "openrouter-cost-reconcile",
    "level": "DEFAULT",
    "startTime": "2026-04-23T00:25:55.689Z"
  },
  {
    "type": "GENERATION",
    "name": "claude.assistant.turn",
    "level": "DEFAULT",
    "startTime": "2026-04-23T00:25:43.164Z"
  },
  {
    "type": "SPAN",
    "name": "claude.conversation",
    "level": "DEFAULT",
    "startTime": "2026-04-23T00:25:43.164Z"
  },
  {
    "type": "SPAN",
    "name": "copilot-sdk-turn",
    "level": "DEFAULT",
    "startTime": "2026-04-23T00:25:42.852Z"
  }
]
```

## Moonshot-route trace 3c910554c9fbfcfa6a71a4b63554622b
```json
[
  {
    "type": "EVENT",
    "name": "openrouter-cost-reconcile",
    "level": "DEFAULT",
    "startTime": "2026-04-23T00:28:35.576Z"
  },
  {
    "type": "GENERATION",
    "name": "claude.assistant.turn",
    "level": "DEFAULT",
    "startTime": "2026-04-23T00:28:01.885Z"
  },
  {
    "type": "GENERATION",
    "name": "claude.assistant.turn",
    "level": "DEFAULT",
    "startTime": "2026-04-23T00:28:01.884Z"
  },
  {
    "type": "GENERATION",
    "name": "claude.assistant.turn",
    "level": "DEFAULT",
    "startTime": "2026-04-23T00:27:57.715Z"
  },
  {
    "type": "SPAN",
    "name": "claude.conversation",
    "level": "DEFAULT",
    "startTime": "2026-04-23T00:27:57.715Z"
  },
  {
    "type": "SPAN",
    "name": "copilot-sdk-turn",
    "level": "DEFAULT",
    "startTime": "2026-04-23T00:27:57.239Z"
  }
]
```

### openrouter-cost-reconcile event metadata (Moonshot)
```json
{
  "cost_usd": 0.02816635,
  "fallback_cost_usd": 0.017799399999999996,
  "model": "moonshotai/kimi-k2.6",
  "provider": "open_router",
  "prompt_tokens": 29493,
  "completion_tokens": 37,
  "generation_id_count": 1,
  "cost_source": "openrouter"
}
```

### Screenshots

#### Login (eval+keyboard fill — refs lost focus)

Native-setter `input` event fed `test@test.com` / `testtest123` past the React form; default `fill` lost the password field's ref and trailed empty validation errors.

#### Logged in

Redirect to `/copilot` after submit.

#### Copilot loaded with notify-permission dialog

First load brings up the "Stay in the loop" dialog — dismissed with "Not now".

#### Copilot ready for input

Sidebar populated, chat textbox visible.

#### Anthropic test message typed (rendered into the React-controlled textarea via native setter)

"Reply with one word: PONG" sits in the input; submit arrow turns dark/active.

#### After submit (turn started, sessionId in URL)

URL contains `?sessionId=…` — turn dispatched to backend; copilot-sdk parent span opens; trace_id captured for backfill.

#### Final UI state

Copilot still functional after both Anthropic and Kimi runs against the rebuilt backend.

### Verdict

The PR's two stated goals work as documented:

1. **Compaction gate for Moonshot/Kimi** — verified at the unit level (`build_sdk_env` no longer sets `CLAUDE_AUTOCOMPACT_PCT_OVERRIDE` when model matches `moonshotai/*`) and at the integration level (Kimi turn ran with `compaction.attempt_count=0`).
2. **OpenRouter cost backfill to Langfuse** — works end-to-end **after the parent-span fix in `d8e08525`**, which is a prerequisite for the captured `langfuse_trace_id` to be non-`None`. Without it, the backfill silently no-ops every turn.

Anthropic-route delta is `+0.0%`; Kimi-route delta is `+58.2%` — the new event is exactly the operator-visible signal the PR set out to add.

Format/lint clean (`poetry run format` + `poetry run pyright`); 46/46 tests pass on `backend/copilot/sdk/env_test.py` + `backend/copilot/sdk/openrouter_cost_test.py`.

majdyz:
## E2E Test Evaluation — All scenarios PASS

5/5 scenarios passed end-to-end. Both PR features verified:
- **Compaction gate** for Moonshot/Kimi works as designed (unit + integration).
- **OpenRouter cost backfill** lands in Langfuse as expected — but only **after the parent-span fix in `d8e08525`**, which was a prerequisite for `get_current_trace_id()` to return non-`None`.

**Pre-fix bug:** `propagate_attributes` only annotates an existing span; it does not create one. By the finally block, all SDK-emitted spans had ended, so `get_current_trace_id()` always returned `None` and the backfill emit was guarded out. Confirmed via repro:
```
[SDK][7959eec8-e14] OTEL span at finally: span=NonRecordingSpan(trace_id=0x000…) recording=False
[SDK][7959eec8-e14][T1][cost-record] Langfuse backfill check: trace_id=None real_cost=0.01743135
```

**Post-fix:** the same code path emits the event with a real trace_id, and the event lands in Langfuse with the right metadata (cost_usd $0.028, fallback $0.018, +58.2% delta on Kimi).

Format/lint clean; 46/46 unit tests pass on `backend/copilot/sdk/{env,openrouter_cost}_test.py`. Trace observations queried via Langfuse API confirm the new `copilot-sdk-turn` parent → `claude.conversation` → `claude.assistant.turn` → `openrouter-cost-reconcile` (event) hierarchy on both routes.

Mergeable from an E2E perspective. (cannot self-approve via GitHub)

autogpt_platform/backend/backend/copilot/sdk/openrouter_cost_test.py
```diff
@@ -518,3 +518,194 @@ async def _get(self, *args, **kwargs):  # noqa: ARG001
...
+        with (
+            patch(
+                "backend.copilot.sdk.openrouter_cost.persist_and_record_usage",
+                new_callable=AsyncMock,
+            ),
+            patch("httpx.AsyncClient.get", new=_get),
+            patch(
+                "backend.copilot.sdk.openrouter_cost.get_client",
+                return_value=mock_lf,
+            ),
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [d243bf6c9](https://github.com/Significant-Gravitas/AutoGPT/commit/d243bf6c9787bab1e314c09764c6eb1e753c2306): patched `backend.copilot.sdk.openrouter_cost.asyncio.sleep` in the fallback-path test so the retry-budget backoff is skipped — test now finishes in ~1s instead of ~30s.

autogpt_platform/backend/backend/copilot/sdk/service.py
```diff
@@ -3250,8 +3245,19 @@ async def _setup_e2b():
...
         sdk_model = await _resolve_sdk_model_for_request(model, session_id, user_id)
 
+        # sdk_cwd routes the CLI's temp dir into the per-session workspace
+        # so sub-agent output files land inside sdk_cwd (see build_sdk_env).
+        sdk_env = build_sdk_env(
+            session_id=session_id,
+            user_id=user_id,
+            sdk_cwd=sdk_cwd,
+            model=sdk_model,
+        )
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [d243bf6c9](https://github.com/Significant-Gravitas/AutoGPT/commit/d243bf6c9787bab1e314c09764c6eb1e753c2306): `stream_chat_completion_sdk` now resolves `fallback_model` up front and passes `_resolve_env_model(sdk_model, fallback_model)` into `build_sdk_env` so a 529-triggered swap to a Moonshot fallback still suppresses `CLAUDE_AUTOCOMPACT_PCT_OVERRIDE`. The same resolved `fallback_model` is reused on the `sdk_options_kwargs` line so we resolve it once per turn.

majdyz:
🤖 @coderabbitai Outside-diff finding from the latest review fixed in [d243bf6c9](https://github.com/Significant-Gravitas/AutoGPT/commit/d243bf6c9787bab1e314c09764c6eb1e753c2306):

- `service.py` (4088-4123): `langfuse_trace_id = get_client().get_current_trace_id()` is now hoisted out of the `if _otel_ctx is not None:` branch and runs under the `if _lf_span is not None:` guard before tearing down `_lf_span`. The narrow window where `_lf_span.__enter__()` succeeded but `_otel_ctx` was never assigned (e.g. `propagate_attributes(...).__enter__()` raised) no longer orphans the reconcile event from the parent span.

The two inline findings (test `asyncio.sleep` patch + `fallback_model` Moonshot compaction gate) are addressed in the same commit and the inline replies link back here.

autogpt_platform/backend/backend/copilot/sdk/service.py
```diff
@@ -475,6 +504,7 @@ async def _reduce_context(
             transcript_content,
             model=config.thinking_standard_model,
             log_prefix=log_prefix,
+            target_tokens=_compaction_target_tokens(config.thinking_standard_model),
```
majdyz: 🤖 🟠 **Should Fix**: `_compaction_target_tokens` is sized against `config.thinking_standard_model` (the compactor LLM), but the **CLI's autocompact threshold** that we're trying to stay below is the runtime SDK model's window — which is resolved per-turn by `_resolve_sdk_model_for_request` and may differ (e.g. compactor is Sonnet 200K, runtime is Kimi 256K, or operator overrides via `config.claude_agent_model`/LD per-user routing).

The fallback-aware variant (`_resolve_env_model(sdk_model, fallback_model)`) is exactly what `build_sdk_env` already uses to gate the env var. `sdk_model` is in scope at the `_reduce_context` callsite (line 3169 → 3722), so threading it through and using the same resolver here would make the alignment exact instead of approximate. As-is, the helper is safe (it errs on the side of over-compressing) but the headline of this commit — "align SDK retry compaction target with CLI autocompact threshold" — only holds when the compactor and runtime models share a window.

majdyz: Addressed in ff83ffc15 — `_compaction_target_tokens` is now passed the runtime model (`sdk_model` resolved per-turn by `_resolve_sdk_model_for_request`) via a new `runtime_model` kwarg on `_reduce_context`, with the compactor model kept as the fallback. Test `test_reduce_context_uses_runtime_model_for_target` covers the case where compactor and runtime differ.

autogpt_platform/backend/backend/copilot/sdk/service.py
```diff
@@ -4081,6 +4156,18 @@ def _on_stderr(line: str) -> None:
...
+            # Capture from our Langfuse parent span before tearing it down;
+            # tracks the lifetime of ``_lf_span`` so the trace id is still
+            # available if ``_otel_ctx`` was never entered.
+            try:
+                langfuse_trace_id = get_client().get_current_trace_id()
+            except Exception:
+                logger.debug("Failed to capture Langfuse trace_id", exc_info=True)
+            try:
+                _lf_span.__exit__(*sys.exc_info())
+            except Exception:
```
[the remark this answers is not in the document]
majdyz: False positive — verified locally on Langfuse `^3.14.1`: `get_client().start_as_current_span` exists (alongside `start_as_current_observation`). The exception path is dead in this SDK version and the `_lf_span`/`langfuse_trace_id` capture works as intended; `/pr-test --fix` validated the reconcile event attaches to the parent trace.

autogpt_platform/backend/backend/copilot/sdk/service.py
```diff
@@ -3403,6 +3462,16 @@ def _on_stderr(line: str) -> None:
         if _user_tier:
             _otel_metadata["subscription_tier"] = _user_tier.value
 
+        # Open a Langfuse parent span so the trace_id is observable from
+        # the finally block — ``propagate_attributes`` only annotates an
+        # existing span, it does not create one.
+        try:
+            _lf_span = get_client().start_as_current_span(name="copilot-sdk-turn")
```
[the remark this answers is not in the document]
majdyz: False positive — verified locally on Langfuse `^3.14.1`: `get_client().start_as_current_span(name="copilot-sdk-turn")` exists and returns a working context manager (also: `start_as_current_observation`). The `try/except` is defensive against unrelated init failures (e.g. missing keys), not against `AttributeError`. /pr-test --fix validated that `langfuse_trace_id` is captured and the openrouter-cost-reconcile event attaches to the parent span.

autogpt_platform/backend/backend/copilot/sdk/service.py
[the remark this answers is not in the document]
majdyz: False positive — `langfuse._client.client.Langfuse.start_as_current_span` does exist (signature: `(*, trace_context, name, input, output, metadata, version, level, status_message, end_on_exit) -> _AgnosticContextManager[LangfuseSpan]`, returns a context manager with both `__enter__` and `__exit__`).

Verified end-to-end by a prior /pr-test agent run — the wrapping span correctly captures `langfuse_trace_id` before the OTel context exits, and the resulting `openrouter-cost-reconcile` child event lands on cloud.langfuse.com with the expected metadata (cost $0.028 vs $0.018 CLI estimate, +58.2% delta on Kimi). The try/except is defensive belt-and-braces in case Langfuse becomes unconfigured at runtime, not because the call is unreliable today.

autogpt_platform/backend/backend/copilot/sdk/service.py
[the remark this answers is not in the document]
majdyz: False positive — `langfuse._client.client.Langfuse.start_as_current_span` does exist (signature: `(*, trace_context, name, input, output, metadata, version, level, status_message, end_on_exit) -> _AgnosticContextManager[LangfuseSpan]`, returns a context manager with both `__enter__` and `__exit__`).

Verified end-to-end by a prior /pr-test agent run — the wrapping span correctly captures `langfuse_trace_id` before the OTel context exits, and the resulting `openrouter-cost-reconcile` child event lands on cloud.langfuse.com with the expected metadata (cost $0.028 vs $0.018 CLI estimate, +58.2% delta on Kimi). The try/except is defensive belt-and-braces in case Langfuse becomes unconfigured at runtime, not because the call is unreliable today.

autogpt_platform/backend/backend/copilot/sdk/env.py
```diff
@@ -40,6 +42,11 @@ def build_sdk_env(
     All modes receive workspace isolation (``CLAUDE_CODE_TMPDIR``) and
     security hardening env vars to prevent .claude.md loading, prompt
     history persistence, auto-memory writes, and non-essential traffic.
+
+    *model* is the resolved SDK model slug (e.g. ``"moonshotai/kimi-k2.6"``
+    or ``"anthropic/claude-sonnet-4-6"``).  Used to gate model-specific env
+    vars (currently: ``CLAUDE_AUTOCOMPACT_PCT_OVERRIDE`` is skipped for
+    Moonshot since the cache-cost rationale doesn't apply there).
```
Pwuts, line 49: why doesn't it apply?

autogpt_platform/backend/backend/copilot/sdk/env.py
```diff
@@ -103,10 +110,18 @@ def _safe(v: str) -> str:
     # this flag harmlessly (those betas are not enabled there either by default).
     env["CLAUDE_CODE_DISABLE_EXPERIMENTAL_BETAS"] = "1"
 
-    # Trigger context compaction earlier — default is 70% of 200K = 140K.
-    # Set to 50% = 100K to keep context smaller and reduce cache creation costs.
-    # Context >200K accounts for 54% of total cost despite being only 3% of calls.
-    env["CLAUDE_AUTOCOMPACT_PCT_OVERRIDE"] = "50"
+    # Auto-compaction trigger threshold (CLI default: ~93% of perceived window).
+    # Lowering it via env keeps context smaller and cache creation costs down on
+    # Anthropic.  Skipped for Moonshot routes (no cache_create benefit) and when
```
Pwuts: Is this specific to Moonshot or rather a function of the context window size?

autogpt_platform/backend/backend/copilot/sdk/env.py
majdyz: Cache-cost rationale = PR #12747 set the override to 50% (= ~100K trigger on a 200K window) specifically because **cache_create cost on Anthropic** dominated long Anthropic turns: context >200K accounted for 54% of total cost despite being only 3% of calls — and `cache_create` is billed at 1.25x base input. Lowering the trigger reduced how much context got cached per turn.

On Moonshot via OpenRouter: their Anthropic-compat shim silently drops the `cache_control` markers we send (they use their own `X-Msh-Context-Cache` header out-of-band), so empirically `cache_create=0` on every Kimi turn. There is no $/Mtok cache-creation cost to optimize away → the original lever has nothing to pull on, while the side-effect (early auto-compaction → slow LLM round-trip cascades) still hits us. That's what "doesn't apply" means here. Will tighten the wording.

autogpt_platform/backend/backend/copilot/sdk/env.py
majdyz: Specific to Moonshot, not window size. The override exists to limit **`cache_create` cost on Anthropic** (PR #12747: >200K context = 54% of total cost despite being 3% of calls; `cache_create` is 1.25x input). On Moonshot/OpenRouter we observe `cache_create=0` regardless of window — Moonshot's Anthropic-compat shim drops our `cache_control` markers (they cache via their own out-of-band `X-Msh-Context-Cache` header), so there's no cache-creation cost to optimize and the only effect of the override would be to trigger early compactions (each = a slow LLM round-trip).

Window size is not the discriminator — Kimi's window is actually larger (256K vs 200K). The discriminator is whether `cache_control` markers translate to billed `cache_create` tokens on that route. Today only Anthropic-direct does. If a future provider also bills `cache_create`, we'd add it to the gate by route, not by window. Happy to rename the helper to something like `route_billed_cache_create()` if you'd prefer the gate to be intent-named rather than vendor-named.

majdyz: Specifically about `cache_create` behaviour on Moonshot's OpenRouter route, not directly about window size.

The 50% override was introduced in #12747 because for Anthropic, context >200K accounted for 54% of total cost — most of that being cache-creation tokens. Compacting earlier kept us in a regime where cache writes were affordable.

For Moonshot we observe `cache_create=0` on every Kimi turn (confirmed against session 85804387's traces — see PR description for the empirical numbers). Moonshot's OpenRouter endpoint silently drops Anthropic-style `cache_control: ephemeral` markers per their own docs (they use automatic prefix caching with `X-Msh-Context-Cache` headers instead, which OpenRouter doesn't expose). So there's no cache-creation cost to optimise away; the 50% trigger gives zero benefit and only burns extra LLM round-trips on cascade compactions.

Window size is incidental — Moonshot has 256K vs Sonnet's 200K, but even at the same window the cache-write absence would still be the dominant factor. If we ever observed another OpenRouter provider with the same `cache_create=0` pattern (Grok, Gemini, etc.), we'd want to widen the gate; today Moonshot is the only one we have evidence for.

autogpt_platform/backend/backend/copilot/sdk/env.py
majdyz: Because the cache-cost rationale that motivated the 50% override (Anthropic context >200K = 54% of total cost, mostly cache-creation tokens — see #12747) literally cannot apply when no cache is being created.

Moonshot's OpenRouter endpoint returns `cache_create=0` on every Kimi turn observed (production traces from session 85804387 confirm this). Their docs say prefix caching is fully automatic via `X-Msh-Context-Cache` headers and `cache_control` body markers are silently dropped. So we never enter the "large cache writes drove up cost" regime that the 50% threshold was designed to protect against — instead we'd just trigger an extra LLM round-trip per compaction (3+ per turn observed pre-fix) for zero cost benefit.

Updated the inline comment to be explicit about this. Happy to widen the gate if we ever observe another provider with the same `cache_create=0` pattern.

PR #12895, opened 2026-04-23
fix(frontend): filter null query params (#11237) and redirect www to non-www (#9188)

### LINKED ISSUE #9188: Adding "www." to the start of the url breaks the platform
If you go to [https://www.platform.agpt.co/](https://www.platform.agpt.co/) instead of [https://platform.agpt.co/](https://www.platform.agpt.co/) , the platform will not work at all. You won't be logged in even through you're logged in usually. <br>You can log in, but wont see any Agents in your library.

### Steps to reproduce:

1. Go to [https://www.platform.agpt.co/](https://www.platform.agpt.co/)
2. Observe that you are logged out - log back in
3. Go to your library
4. Observe that you have no Agents in your library - even if you did before.

### Desired Behaviour

[https://www.platform.agpt.co/](https://www.platform.agpt.co/) and [https://platform.agpt.co/](https://www.platform.agpt.co/) are the same website, rather than duplicate pages. 

There should be no difference between them, they should both work, and you should be logged in no matter which url you type in.

---

Ahrefs is also showing these as "**Duplicate pages without canonical"** which is hurting our SEO.

### LINKED ISSUE #11237: Validation failed for GET /api/library/agents/by-graph/{graph_id}: [{'type': 'int_parsing', 'loc': ('query', 'version'), 'msg': 'Input should be a valid integer, unable to parse string as an integer', 'input': 'undefined'}
Sentry Issue: [AUTOGPT-SERVER-495](https://significant-gravitas.sentry.io/issues/6686674180/?referrer=Linear)

```
Validation failed for GET /api/library/agents/by-graph/d1e5e9d2-c9eb-4bcf-afe3-2dd006135870: [{'type': 'int_parsing', 'loc': ('query', 'version'), 'msg': 'Input should be a valid integer, unable to parse string as an integer', 'input': 'undefined'}]. Fix the request payload and try again.
```

SymbolStar:
## Summary

Two frontend bug fixes:

### 1. Filter null values in query parameters (fixes #11237)

**Problem**: `buildUrlWithQuery()` in `helpers.ts` only filtered `undefined` values but not `null`. When `graphVersion` was `null`, the string `"null"` was sent as a query parameter, causing FastAPI to return a 422 validation error (expects `int`).

**Fix**: Added `value !== null` check alongside the existing `value !== undefined` filter in `buildUrlWithQuery()`.

### 2. Redirect www to non-www (fixes #9188)

**Problem**: Accessing the platform via `www.platform.agpt.co` breaks authentication because Supabase Auth cookies are bound to `platform.agpt.co`. The `www.` subdomain is a different origin, so cookies aren't sent → auth fails → white screen or redirect loop.

**Fix**: Added a 301 redirect in `middleware.ts` that strips the `www.` prefix before the Supabase session check runs.

## Changes

- `autogpt_platform/frontend/src/lib/autogpt-server-api/helpers.ts`: `buildUrlWithQuery` now filters both `undefined` and `null`
- `autogpt_platform/frontend/src/middleware.ts`: www → non-www 301 redirect

## Testing

- Verified `URLSearchParams` behavior: `null` values pass through as `"null"` strings while `undefined` is already filtered
- The www redirect is a standard pattern used in Next.js middleware

### DIFF
--- a/autogpt_platform/frontend/src/middleware.ts
+++ b/autogpt_platform/frontend/src/middleware.ts
@@ -1,7 +1,15 @@
 import { updateSession } from "@/lib/supabase/middleware";
-import { type NextRequest } from "next/server";
+import { NextResponse, type NextRequest } from "next/server";
 
 export async function middleware(request: NextRequest) {
+  // Redirect www to non-www to prevent cookie/auth domain mismatch (#9188)
+  const host = request.headers.get("host") || "";
+  if (host.startsWith("www.")) {
+    const url = request.nextUrl.clone();
+    url.host = host.replace(/^www\./, "");
+    return NextResponse.redirect(url, 301);
+  }
+
   return await updateSession(request);
 }
 
--- a/autogpt_platform/frontend/src/lib/autogpt-server-api/helpers.ts
+++ b/autogpt_platform/frontend/src/lib/autogpt-server-api/helpers.ts
@@ -68,10 +68,11 @@ export function buildUrlWithQuery(
 ): string {
   if (!query) return url;
 
-  // Filter out undefined values to prevent them from being included as "undefined" strings
+  // Filter out undefined and null values to prevent them from being included
+  // as "undefined" / "null" strings in query parameters (fixes #11237)
   const filteredQuery = Object.entries(query).reduce(
     (acc, [key, value]) => {
-      if (value !== undefined) {
+      if (value !== undefined && value !== null) {
         acc[key] = value;
       }
       return acc;


autogpt_platform/frontend/src/middleware.ts
```diff
@@ -1,7 +1,15 @@
...
-import { type NextRequest } from "next/server";
+import { NextResponse, type NextRequest } from "next/server";
 
 export async function middleware(request: NextRequest) {
+  // Redirect www to non-www to prevent cookie/auth domain mismatch (#9188)
+  const host = request.headers.get("host") || "";
+  if (host.startsWith("www.")) {
+    const url = request.nextUrl.clone();
+    url.host = host.replace(/^www\./, "");
+    return NextResponse.redirect(url, 301);
```
majdyz, line 10: 🤖 🟠 **Should Fix**: `301` downgrades non-GET methods to GET and drops the request body. Since the middleware matcher runs on `/auth/authorize` and `/auth/integrations/*`, an OAuth callback or any POST landing on the `www.` host would have its body silently dropped on redirect. Prefer `308` (Permanent Redirect), which preserves the method and body while still being cacheable.

```diff
-    return NextResponse.redirect(url, 301);
+    return NextResponse.redirect(url, 308);
```

Also worth thinking about whether this is better handled at the infra/DNS layer (Vercel has built-in www→apex redirect config) rather than inside the Next middleware — moving it out of the app avoids the 301/308 cache-forever risk if the decision ever needs to reverse.

majdyz (changes requested):
Thanks for the contribution! A few things before this can move forward:

1. **CLA not signed** — please sign the [Contributor License Agreement](https://cla-assistant.io/Significant-Gravitas/AutoGPT?pullRequest=12895) first; we can't accept contributions without it.

2. **Please split into two PRs** — the null-query-params fix (#11237) and the www→non-www redirect (#9188) are unrelated concerns. Our contribution guide asks for one concern per PR so reviewers can land them independently (each has different risk and reversibility). The null-filter fix is a one-liner we can land quickly; the redirect change has policy implications worth discussing separately.

3. **Blocker on the redirect**: see the inline — `301` drops POST bodies, which can silently break OAuth callbacks hitting the `www.` host. Please use `308` instead. Also happy to discuss moving this out of the Next middleware entirely (Vercel/DNS-level) so a cache-forever redirect decision isn't baked into the app.

Feel free to retag me once those are addressed. A screenshot of the browser Network tab showing (a) the good query string for the null-filter fix, and (b) the `www.` host redirecting to the apex with the right status code would be appreciated as proof of the fix.

SymbolStar:
recheck

SymbolStar:
Splitting into two PRs as requested by @majdyz. Closing this one.

- www→non-www redirect (with 308): will open separately
- null query params filter: will open separately

PR #12904, opened 2026-04-24
fix(backend/copilot): surface non-zero E2B exits as real results, not sandbox errors

majdyz:
## Why

`gh auth status` looked flaky in the E2B sandbox. Not actually flaky: it fails deterministically when the user has not connected GitHub (or the token is missing/expired), and our wrapper disguises that legitimate exit-1 as a sandbox infrastructure failure.

Root cause: E2B's `sandbox.commands.run()` raises `CommandExitException` for **any** non-zero exit. We caught it as a generic `Exception` and returned an `ErrorResponse` with message:

```
E2B execution failed: Command exited with code 1 and error:
{stderr}
```

When the model runs `gh auth status 2>&1`, stderr is redirected to stdout — so `exc.stderr` is empty **and** `exc.stdout` (which carries the real info, e.g. "You are not logged into any GitHub hosts") is discarded. The model sees a generic infra failure, can't tell it's an auth-check signal, and prompts the user with broken-looking errors instead of calling `connect_integration(provider="github")`.

Compare: the local bubblewrap path already handles non-zero exits correctly by returning a `BashExecResponse` with `exit_code` set. The E2B path was asymmetric.

## What

- Import `CommandExitException` and catch it explicitly in `_execute_on_e2b` before the generic handler.
- Return a `BashExecResponse` with the real `exit_code`, `stdout`, and `stderr` from the exception (scrubbed of injected secret values, same as the success path).
- Extract shared scrub/build logic into `_build_response` to avoid duplicating it across the success and exit-exception branches.
- Keep `TimeoutException` and the catch-all `except Exception` for real infra failures.

## How

Result shape now matches bubblewrap: non-zero exit is a valid result, not an error. The model sees:

```
message: "Command executed with status code 1"
exit_code: 1
stdout: "You are not logged into any GitHub hosts. ..."
stderr: ""
```

instead of the prior cryptic "E2B execution failed" message.

## Test plan

- [x] New unit test `test_nonzero_exit_returned_as_bash_exec_response` in `bash_exec_test.py` — mocks `sandbox.commands.run` to raise `CommandExitException`, asserts `BashExecResponse` with correct `exit_code`, and verifies secret scrubbing on both `stdout` and `stderr`.
- [x] `poetry run pytest backend/copilot/tools/bash_exec_test.py` — 5 passed.
- [x] `poetry run pyright` on changed files — 0 errors.
- [x] `poetry run ruff` — clean.

### DIFF
--- a/autogpt_platform/backend/backend/copilot/tools/bash_exec.py
+++ b/autogpt_platform/backend/backend/copilot/tools/bash_exec.py
@@ -35,7 +35,7 @@ from .sandbox import get_workspace_dir, has_full_sandbox, run_sandboxed
 logger = logging.getLogger(__name__)
 
 
-def _build_response(
+def _build_completion_response(
     stdout: str | None,
     stderr: str | None,
     exit_code: int,
@@ -197,7 +197,7 @@ class BashExecTool(BaseTool):
                 timeout=timeout,
                 envs=envs,
             )
-            return _build_response(
+            return _build_completion_response(
                 result.stdout,
                 result.stderr,
                 result.exit_code,
@@ -205,7 +205,7 @@ class BashExecTool(BaseTool):
                 session_id,
             )
         except CommandExitException as exc:
-            return _build_response(
+            return _build_completion_response(
                 exc.stdout, exc.stderr, exc.exit_code, secret_values, session_id
             )
         except TimeoutException:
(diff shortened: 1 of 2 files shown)


majdyz:
Renamed `_build_response` → `_build_completion_response` in 29f5d11 to make it obvious the helper is for the command-completed path (success or non-zero exit), not timeouts. Skipped the parameterized option because the timeout branch uses a different message, exit_code (-1), and stderr format — folding it in would make the helper awkward without real reuse. Thanks for the nitpick @coderabbitai.

PR #12903, opened 2026-04-24
feat(platform): add MAX tier + LD-configurable pricing + hide unconfigured tiers

majdyz:
## What

Introduces a new `MAX` tier slot between `PRO` and `BUSINESS` (self-service $320/mo at 20× capacity), routes every self-service tier's Stripe price ID through LaunchDarkly, and hides tiers from the UI when their price isn't configured. `BUSINESS` stays in the enum at 60× as a reserved/future self-service slot (hidden by default until its LD price flag is set). ENTERPRISE stays admin-managed.

## Tier shape after this PR

| Enum | UI label | Multiplier | LD price flag | Surfaced in UI by default |
|---|---|---|---|---|
| `FREE` | Basic | 1× | `stripe-price-id-basic` | no (flag unset) |
| `PRO` | Pro | 5× | `stripe-price-id-pro` | yes (already live) |
| `MAX` **(new)** | Max | 20× | `stripe-price-id-max` | no (flag unset until $320 price ready) |
| `BUSINESS` | Business | 60× | `stripe-price-id-business` | no (reserved / future) |
| `ENTERPRISE` | — | 60× | — (admin-managed) | no (Contact-Us only) |

## Prisma

- Added `MAX` between `PRO` and `BUSINESS` in `SubscriptionTier`.
- Migration `add_subscription_tier_max/migration.sql` uses `ALTER TYPE ... ADD VALUE IF NOT EXISTS 'MAX' BEFORE 'BUSINESS'` (transactional since PG 12). No data migration — no rows currently on BUSINESS via self-service flows.

## Backend

- `get_subscription_price_id` flag map covers `FREE`/`PRO`/`MAX`/`BUSINESS`. ENTERPRISE returns `None`.
- `GET /credits/subscription.tier_costs` only includes tiers whose LD price ID is set. Current tier always present as a safety net.
- `POST /credits/subscription` routes by LD-resolved prices instead of hard-coding `tier == FREE`:
  - Target `FREE` + `stripe-price-id-basic` unset → legacy cancel-at-period-end (unchanged behaviour).
  - Target has LD price → modify in-place when user has an active sub, else Checkout Session.
  - Priced-FREE users with no sub fall through to Checkout (admin-granted DB-flip shortcut gated on `current_tier != FREE`).
- `sync_subscription_from_stripe` + `get_pending_subscription_change` cover FREE/PRO/MAX/BUSINESS in the price-to-tier map so every tier's Stripe webhook reconciles cleanly.
[...]

PR #12910, opened 2026-04-24
feat(platform): LD-configurable rate-limit multipliers + relative UI display

majdyz:
## Summary

- **Backend (`copilot/rate_limit`)** — ``TIER_MULTIPLIERS`` is now float-typed and resolvable through a new LaunchDarkly flag ``copilot-tier-multipliers``.  The integer defaults live on as ``_DEFAULT_TIER_MULTIPLIERS`` and are merged with whatever LD returns (missing / invalid keys inherit defaults; LD failures fall back to defaults without raising).  ``get_global_rate_limits`` now honours the flag per-user and casts ``int(base * multiplier)`` so downstream microdollar math stays integer even when LD hands back a fractional multiplier (e.g. 8.5×).  Cached for 60 s via ``@cached(ttl_seconds=60, maxsize=8, cache_none=False)`` to match the pattern in ``get_subscription_price_id``.
- **Backend (`api/features/v1`)** — ``SubscriptionStatusResponse`` gains ``tier_multipliers: dict[str, float]``, populated for the same set of tiers that make it into ``tier_costs`` so hidden tiers never get a rendered badge.
- **Frontend (`SubscriptionTierSection`)** — drops the hard-coded ``"5x" / "20x"`` strings from ``TIERS`` and introduces ``formatRelativeMultiplier(tierKey, tierMultipliers)``: the lowest *visible* multiplier becomes the baseline (no badge), every other tier renders ``"N.Nx rate limits"`` relative to it.  Fractional LD values like 8.5× round to one decimal.

The admin rate-limit page (``/admin/rate-limits``) keeps the static ``TIER_MULTIPLIERS`` defaults — it's admin-facing, infrequently viewed, and fine to lag the LD value until next deploy (noted in-code).

Related upstream: this PR stacks logically after #12903 (which added the ``MAX`` tier + LD-configurable prices) but does **not** require it — each PR can merge in either order.  No schema changes, no migration.

## Test plan

- [x] ``poetry run black backend/... --check`` + ``poetry run ruff check backend/...`` pass
- [x] ``pnpm format`` pass (modified files unchanged)
- [x] New backend tests: ``TestGetTierMultipliers`` (defaults, LD override, invalid JSON, unknown tier / non-positive values, LD failure) — **5 / 5 pass**
- [x] New backend test: ``TestGetGlobalRateLimitsWithTiers::test_ld_override_applies_fractional_multiplier`` — **pass**
- [x] ``backend/copilot/rate_limit_test.py`` — non-DB subset **72 / 72 pass**; ``TestGetUserTier`` / ``TestSetUserTier`` require the full test-server fixture (Redis + Prisma) and are not run in this worktree — same behaviour on clean ``dev``
- [x] ``backend/api/features/subscription_routes_test.py`` — **40 / 40 pass** (includes new ``test_get_subscription_status_tier_multipliers_ld_override``)
- [x] Frontend vitest targeted suite — **51 / 51 pass**
  - ``helpers.test.ts`` — new ``formatRelativeMultiplier`` cases (lowest-tier null, integer ratio, fractional ratio, hidden-tier null, fractional LD)
  - ``SubscriptionTierSection.test.tsx`` — three new cases for relative badges, rebasing when the lowest tier is hidden, fractional LD overrides

### DIFF
--- a/autogpt_platform/backend/backend/copilot/rate_limit.py
+++ b/autogpt_platform/backend/backend/copilot/rate_limit.py
@@ -103,7 +103,7 @@ TIER_MULTIPLIERS = _DEFAULT_TIER_MULTIPLIERS
 DEFAULT_TIER = SubscriptionTier.BASIC
 
 
-@cached(ttl_seconds=60, maxsize=8, cache_none=False)
+@cached(ttl_seconds=60, maxsize=1, cache_none=False)
 async def _fetch_tier_multipliers_flag() -> dict[SubscriptionTier, float] | None:
     """Fetch the ``copilot-tier-multipliers`` LD flag and parse it.
 
@@ -148,20 +148,17 @@ async def _fetch_tier_multipliers_flag() -> dict[SubscriptionTier, float] | None
     return parsed or None
 
 
-async def get_tier_multipliers(user_id: str) -> dict[SubscriptionTier, float]:
+async def get_tier_multipliers() -> dict[SubscriptionTier, float]:
     """Return the effective ``{tier: multiplier}`` map.
 
     Honours the ``copilot-tier-multipliers`` LD flag when set; missing tiers
     inherit :data:`_DEFAULT_TIER_MULTIPLIERS`.  Unparseable flag values or LD
     fetch failures fall back to the defaults without raising.
 
-    The ``user_id`` argument is accepted for symmetry with other resolvers
-    (``get_global_rate_limits``, ``get_user_tier``), but the flag is currently
-    evaluated system-wide — per-tier multipliers are a global knob.  Keeping
-    the parameter in the signature means a future move to per-cohort overrides
-    won't break callers.
+    The flag is evaluated system-wide — per-tier multipliers are a global knob.
+    If per-cohort overrides are ever needed, add a user_id parameter here and
+    thread it through ``_fetch_tier_multipliers_flag`` to LD.
     """
-    del user_id  # unused — retained for API symmetry; see docstring
     try:
         override = await _fetch_tier_multipliers_flag()
     except Exception:
@@ -781,7 +778,7 @@ async def get_global_rate_limits(
     # so multipliers can be tuned without a deploy. Falls back to the defaults
     # when LD is unavailable.
     tier = await get_user_tier(user_id)
-    multipliers = await get_tier_multipliers(user_id)
+    multipliers = await get_tier_multipliers()
     multiplier = multipliers.get(tier, 1.0)
     if multiplier != 1.0:
         # Cast back to int to preserve the microdollar integer contract
--- a/autogpt_platform/backend/backend/copilot/rate_limit.py
+++ b/autogpt_platform/backend/backend/copilot/rate_limit.py
@@ -148,12 +148,16 @@ async def _fetch_tier_multipliers_flag() -> dict[SubscriptionTier, float] | None
     return parsed or None
 
 
-async def get_tier_multipliers() -> dict[SubscriptionTier, float]:
-    """Return the effective ``{tier: multiplier}`` map.
+async def get_tier_multipliers() -> dict[str, float]:
+    """Return the effective ``{tier_value: multiplier}`` map.
 
     Honours the ``copilot-tier-multipliers`` LD flag when set; missing tiers
     inherit :data:`_DEFAULT_TIER_MULTIPLIERS`.  Unparseable flag values or LD
     fetch failures fall back to the defaults without raising.
+
+    Keys are the tier enum string values (``"BASIC"``, ``"PRO"``, …) rather
+    than the enum itself so callers holding ``prisma.enums.SubscriptionTier``
+    don't hit a spurious mismatch against this module's local mirror.
 
     The flag is evaluated system-wide — per-tier multipliers are a global knob.
     If per-cohort overrides are ever needed, add a user_id parameter here and
@@ -165,11 +169,10 @@ async def get_tier_multipliers() -> dict[SubscriptionTier, float]:
         # LD SDK / Redis / network failures here are best-effort — fall back.
         logger.warning("get_tier_multipliers: LD lookup failed", exc_info=True)
         override = None
-    if not override:
-        return dict(_DEFAULT_TIER_MULTIPLIERS)
-    merged = dict(_DEFAULT_TIER_MULTIPLIERS)
-    merged.update(override)
-    return merged
+    merged: dict[SubscriptionTier, float] = dict(_DEFAULT_TIER_MULTIPLIERS)
+    if override:
+        merged.update(override)
+    return {tier.value: multiplier for tier, multiplier in merged.items()}
 
 
 class UsageWindow(BaseModel):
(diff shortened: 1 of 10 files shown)


autogpt_platform/backend/backend/copilot/rate_limit.py
```diff
@@ -80,22 +80,101 @@ class SubscriptionTier(str, Enum):
...
+
+    The ``user_id`` argument is accepted for symmetry with other resolvers
+    (``get_global_rate_limits``, ``get_user_tier``), but the flag is currently
+    evaluated system-wide — per-tier multipliers are a global knob.  Keeping
+    the parameter in the signature means a future move to per-cohort overrides
+    won't break callers.
+    """
+    del user_id  # unused — retained for API symmetry; see docstring
+    try:
+        override = await _fetch_tier_multipliers_flag()
```
majdyz, line 167: 🤖 🟡 **Nice to Have**: `get_tier_multipliers` takes `user_id` but immediately `del user_id`s it — the docstring frames it as API symmetry for a future per-cohort move. Two options cleaner than the current dead-parameter pattern:

1. **Drop it** — `_fetch_tier_multipliers_flag()` already uses `user_id="system"` hardcoded, and callers don't rely on the param. Removing it makes the signature match the behavior.
2. **Flow it through** — pass `user_id` down into `get_feature_flag_value` so LD can target per-user cohorts today (not in some future PR). This is the bigger change, but it matches the stated intent of the param.

Current state (param accepted, ignored, flag evaluated system-wide) is the worst of both: future callers will assume per-user works when it doesn't.

autogpt_platform/frontend/src/app/(platform)/profile/(user)/credits/components/SubscriptionTierSection/helpers.ts
```diff
@@ -58,3 +53,21 @@ export function formatPendingDate(value: Date | string): string {
...
+// the payload (hidden, e.g. BUSINESS before its LD price is set).
+export function formatRelativeMultiplier(
+  tierKey: string,
+  tierMultipliers: Record<string, number>,
+): string | null {
+  const mine = tierMultipliers[tierKey];
+  if (mine === undefined) return null;
+  const visible = Object.values(tierMultipliers).filter((v) => v > 0);
+  if (visible.length === 0) return null;
+  const min = Math.min(...visible);
```
majdyz, line 69: 🤖 🟡 **Nice to Have**: `mine === min` is strict float equality. Two subtle edge cases:

1. **Floating-point drift**: if LD returns `{"PRO": 5.0000000001, "BASIC": 5.0}`, they're mathematically equivalent but the `===` check fails → PRO shows "1.0x rate limits" instead of being treated as the baseline.
2. **Zero multiplier**: `{FREE: 0, PRO: 5}` — `visible` excludes 0, so `min` is 5. But `mine` (for FREE) is still 0 → `mine === min` is false → we emit `"0.0x rate limits"`. Should probably return null.

Fix:
```ts
if (mine <= 0) return null;
const ratio = mine / min;
if (ratio < 1.05) return null; // within 5% = effectively the baseline
return `${ratio.toFixed(1)}x rate limits`;
```
Or compare post-rounding: if `ratio.toFixed(1) === "1.0"` → null.

autogpt_platform/frontend/src/app/(platform)/profile/(user)/credits/components/SubscriptionTierSection/SubscriptionTierSection.tsx
```diff
@@ -178,9 +179,17 @@ export function SubscriptionTierSection() {
               <p className="mb-1 text-2xl font-bold">
                 {formatCost(cost, tier.key)}
               </p>
-              <p className="mb-1 text-sm font-medium text-neutral-600 dark:text-neutral-400">
-                {tier.multiplier} rate limits
-              </p>
+              {(() => {
```
majdyz: 🤖 🔵 **Nit**: IIFE in JSX is awkward — extract to a variable above the return:

```tsx
const rateLimitLabel = formatRelativeMultiplier(
  tier.key,
  subscription.tier_multipliers ?? {},
);
// ... in JSX:
{rateLimitLabel && (
  <p className="mb-1 text-sm font-medium text-neutral-600 dark:text-neutral-400">
    {rateLimitLabel}
  </p>
)}
```
Reads cleaner + avoids allocating a new function per render.

autogpt_platform/backend/backend/copilot/rate_limit.py
```diff
@@ -80,22 +80,101 @@ class SubscriptionTier(str, Enum):
...
+# Public re-export retained for backward compatibility with call-sites / tests
+# that historically read ``TIER_MULTIPLIERS`` directly.  New code should prefer
+# ``get_tier_multipliers`` so LD overrides are honoured.
+TIER_MULTIPLIERS = _DEFAULT_TIER_MULTIPLIERS
+
 DEFAULT_TIER = SubscriptionTier.BASIC
 
 
+@cached(ttl_seconds=60, maxsize=8, cache_none=False)
+async def _fetch_tier_multipliers_flag() -> dict[SubscriptionTier, float] | None:
```
majdyz, line 107: 🤖 🔵 **Nit**: `maxsize=8` on a no-arg function is wasteful — the cache can ever hold at most one entry. `maxsize=1` is clearer intent.

autogpt_platform/backend/backend/copilot/rate_limit.py
majdyz: 🤖 Fixed in [89939078a](https://github.com/Significant-Gravitas/AutoGPT/commit/89939078aa47d8da3e5d0ac66e3815b58d24ca1b): dropped the unused `user_id` parameter. The docstring now explicitly notes that re-introducing it is the path to per-cohort overrides, so future callers won't assume broken per-user semantics.

autogpt_platform/frontend/src/app/(platform)/profile/(user)/credits/components/SubscriptionTierSection/helpers.ts
majdyz: 🤖 Fixed in [89939078a](https://github.com/Significant-Gravitas/AutoGPT/commit/89939078aa47d8da3e5d0ac66e3815b58d24ca1b): guard `mine <= 0` (zero/negative tiers return null) and compare post-rounding — `label === "1.0"` collapses mathematically-equal floats to the baseline regardless of last-bit drift.

autogpt_platform/frontend/src/app/(platform)/profile/(user)/credits/components/SubscriptionTierSection/SubscriptionTierSection.tsx
majdyz: 🤖 Fixed in [89939078a](https://github.com/Significant-Gravitas/AutoGPT/commit/89939078aa47d8da3e5d0ac66e3815b58d24ca1b): hoisted to `const rateLimitLabel` at the top of the map callback.

autogpt_platform/backend/backend/copilot/rate_limit.py
majdyz: 🤖 Fixed in [89939078a](https://github.com/Significant-Gravitas/AutoGPT/commit/89939078aa47d8da3e5d0ac66e3815b58d24ca1b): `maxsize=1`.

PR #12917, opened 2026-04-24
refactor(platform/copilot): consolidate 4 model-routing LD flags into 1 JSON flag

majdyz:
## What

Replaces 4 string-valued LaunchDarkly flags with a single JSON-valued flag for copilot model routing:

- ~~`copilot-fast-standard-model`~~
- ~~`copilot-fast-advanced-model`~~
- ~~`copilot-thinking-standard-model`~~
- ~~`copilot-thinking-advanced-model`~~

**New:** `copilot-model-routing` (JSON), keyed `{mode: {tier: model}}`:
```json
{
  "fast":     { "standard": "anthropic/claude-sonnet-4-6", "advanced": "anthropic/claude-opus-4-6" },
  "thinking": { "standard": "moonshotai/kimi-k2.6",         "advanced": "anthropic/claude-opus-4-6" }
}
```

## Why

Same pattern as the sibling consolidation in #12915 (pricing / cost-limits flags) and the merged #12910 (tier-multipliers):

- One flag per config domain — less LD UI clutter, easier audit trail.
- Atomic updates — rotating fast.standard + thinking.standard is a single save.
- Fewer LD entities to name, version, target, explain.
- Mirrors the now-uniform copilot-* JSON-flag shape.

## How

- `backend/util/feature_flag.py`: drop the four `COPILOT_*_MODEL` enum values, add `COPILOT_MODEL_ROUTING`.
- `backend/copilot/model_router.py`: rewrite `resolve_model` to fetch the JSON flag once per call and walk `payload[mode][tier]`. Missing mode, missing tier-within-mode, non-string cell value, non-dict payload, or LD failure all fall back to the corresponding `ChatConfig` default (same user-visible semantics as before). `_FLAG_BY_CELL` removed entirely; `_config_default` / `ModelMode` / `ModelTier` unchanged.
- Per-user LD targeting preserved — cohorts can still receive different routing.
- No caching added (preserves existing uncached behaviour).
- Docstring references in `copilot/config.py` + `copilot/sdk/service.py` updated to point at the new nested key path; one docstring in `service_test.py` likewise.

## Operator action required BEFORE merging

This PR removes 4 LD flags and introduces 1 replacement.

1. In LaunchDarkly, create `copilot-model-routing` (type: **JSON**, server-side only). Default variation = union of the current four string flags, shaped as:
   ```json
   {
     "fast":     { "standard": "<current copilot-fast-standard-model>",     "advanced": "<current copilot-fast-advanced-model>" },
     "thinking": { "standard": "<current copilot-thinking-standard-model>", "advanced": "<current copilot-thinking-advanced-model>" }
   }
   ```
   Omit any cell that's currently unset (its `ChatConfig` default will be used).

2. Merge this PR.

3. After deploy + smoke, delete the four legacy flags:
   - `copilot-fast-standard-model`
   - `copilot-fast-advanced-model`
   - `copilot-thinking-standard-model`
   - `copilot-thinking-advanced-model`

## Testing

- `backend/copilot/model_router_test.py` rewritten — 27 tests pass:
  - LD unset / `None` payload → fallback for every cell.
  - Full JSON → each cell maps to its value (parametrized).
  - Partial JSON (missing mode, missing tier-within-mode, mode value not a dict).
  - Non-dict payloads (str / list / int / bool) → fallback + warning.
  - Non-string cell values (number, list, bool, dict) → fallback + 'non-string' warning.
  - Empty-string cell → fallback + 'empty string' warning (not 'non-string').
  - LD raises → fallback + warning with `exc_info`.
  - `user_id=None` → skip LD entirely.
  - Single-LD-call regression guard against re-introducing per-cell flag fan-out.
- `backend/copilot/sdk/service_test.py`: 61 tests still pass (it mocks `_resolve_thinking_model_for_user`, so the inner flag change is transparent).
- `black --check` / `ruff check` / `isort --check` all clean.

## Sibling

- #12915 — same consolidation pattern for stripe-price / cost-limits flags.

## Checklist

- [x] I have read the project's contributing guide.
- [x] I have clearly described what this PR changes and why.
- [x] My code follows the style guidelines of this project.
- [x] I have added tests that prove my fix is effective or that my feature works.
- [ ] New and existing unit tests pass locally with my changes (CI will confirm).

### DIFF
--- a/autogpt_platform/backend/backend/copilot/model_router.py
+++ b/autogpt_platform/backend/backend/copilot/model_router.py
@@ -1,24 +1,31 @@
 """LaunchDarkly-aware model selection for the copilot.
 
 Each cell of the ``(mode, tier)`` matrix has a static default baked into
-``ChatConfig`` (see ``copilot/config.py``) and a matching LaunchDarkly
-string-valued feature flag that can override it per-user.  This module
-centralises the lookup so both the baseline and SDK paths agree on the
-selection rule and so A/B experiments can target a single cell without
-shipping a config change.
+``ChatConfig`` (see ``copilot/config.py``) and a single JSON-valued
+LaunchDarkly feature flag — ``copilot-model-routing`` — that can override
+it per-user.  This module centralises the lookup so both the baseline
+and SDK paths agree on the selection rule and so A/B experiments can
+target a single cell without shipping a config change.
 
 Matrix:
 
-    +----------+-------------------------------------+-------------------------------------+
-    |          | standard                            | advanced                            |
-    +----------+-------------------------------------+-------------------------------------+
-    | fast     | copilot-fast-standard-model         | copilot-fast-advanced-model         |
-    | thinking | copilot-thinking-standard-model     | copilot-thinking-advanced-model     |
-    +----------+-------------------------------------+-------------------------------------+
+    +----------+----------+----------+
+    |          | standard | advanced |
+    +----------+----------+----------+
+    | fast     |    .     |    .     |
+    | thinking |    .     |    .     |
+    +----------+----------+----------+
 
-LD flag values are arbitrary strings (model identifiers, e.g.
-``"anthropic/claude-sonnet-4-6"`` or ``"moonshotai/kimi-k2.6"``).  Empty
-or non-string values fall back to the config default.
+LD payload shape::
+
+    {
+      "fast":     {"standard": "anthropic/claude-sonnet-4-6", "advanced": "anthropic/claude-opus-4-6"},
+      "thinking": {"standard": "moonshotai/kimi-k2.6",         "advanced": "anthropic/claude-opus-4-6"}
+    }
+
+Missing mode, missing tier-within-mode, non-string cell value, non-dict
+payload, or LD failure all fall back to the corresponding ``ChatConfig``
+default.
 """
 
 from __future__ import annotations
@@ -33,14 +40,6 @@ logger = logging.getLogger(__name__)
 
 ModelMode = Literal["fast", "thinking"]
 ModelTier = Literal["standard", "advanced"]
-
-
-_FLAG_BY_CELL: dict[tuple[ModelMode, ModelTier], Flag] = {
-    ("fast", "standard"): Flag.COPILOT_FAST_STANDARD_MODEL,
-    ("fast", "advanced"): Flag.COPILOT_FAST_ADVANCED_MODEL,
-    ("thinking", "standard"): Flag.COPILOT_THINKING_STANDARD_MODEL,
-    ("thinking", "advanced"): Flag.COPILOT_THINKING_ADVANCED_MODEL,
-}
 
 
 def _config_default(config: ChatConfig, mode: ModelMode, tier: ModelTier) -> str:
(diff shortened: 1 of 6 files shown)


autogpt_platform/backend/backend/copilot/model_router.py
```diff
@@ -66,38 +65,62 @@ async def resolve_model(
...
+            payload,
+            fallback,
+            mode,
+            tier,
+        )
+        return fallback
+
+    mode_cell = payload.get(mode)
+    if not isinstance(mode_cell, dict):
+        return fallback
```
majdyz, line 119: 🤖 🟡 **Nice to Have**: `mode_cell` is silently dropped when it's present-but-not-a-dict (e.g. `{"fast": "anthropic/..."}` — operator typed a string at the mode level). Compare to lines 95-104 where a non-dict top-level payload **does** log a warning. Either:

1. Log a warning here too — mirrors the top-level branch and surfaces operator typos:
   ```python
   if mode in payload and not isinstance(mode_cell, dict):
       logger.warning(
           "[model_router] copilot-model-routing[%s] expected a JSON object, got %r — "
           "using config default %s for tier %s",
           mode, mode_cell, fallback, tier,
       )
   if not isinstance(mode_cell, dict):
       return fallback
   ```
2. Or stay silent here and **drop** the top-level warning so the behaviour is uniform — a missing/malformed cell is just "not configured".

I'd take (1) — the silent failure is the worst kind for an operator-facing config.

autogpt_platform/backend/backend/copilot/model_router.py
```diff
@@ -66,38 +65,62 @@ async def resolve_model(
...
-        value = await get_feature_flag_value(flag.value, user_id, default=fallback)
+        payload = await get_feature_flag_value(
+            Flag.COPILOT_MODEL_ROUTING.value, user_id, default=None
+        )
     except Exception:
         logger.warning(
-            "[model_router] LD lookup failed for %s — using config default %s",
-            flag.value,
+            "[model_router] LD lookup failed for copilot-model-routing — "
+            "using config default %s for (%s, %s)",
```
majdyz, line 84: 🤖 🔵 **Nit**: `payload` would read better with an explicit annotation given it's used several lines below — `payload: dict[str, Any] | str | int | list | None = ...` is overkill, but `payload: object = await ...` (or just leaving it) is fine. Take it or leave it.

autogpt_platform/backend/backend/copilot/model_router.py
majdyz: 🤖 Fixed in [dd6e19d1d](https://github.com/Significant-Gravitas/AutoGPT/commit/dd6e19d1da571b088f33d47da3b1597ccc34c1fa): added a warning for the present-but-not-a-dict mode case (`{"fast": "anthropic/..."}`). The missing-key path stays silent (it's the legitimate "not configured" state); only the present-but-malformed path warns.

autogpt_platform/backend/backend/copilot/model_router.py
majdyz: 🤖 Fixed in [dd6e19d1d](https://github.com/Significant-Gravitas/AutoGPT/commit/dd6e19d1da571b088f33d47da3b1597ccc34c1fa): `payload: object = await ...`.

PR #12918, opened 2026-04-25
fix(backend/copilot): preserve interrupted SDK partial work on final-failure exit

majdyz:
## Background

[SECRT-2275](https://linear.app/autogpt/issue/SECRT-2275). User report: when a copilot ("autopilot") turn is interrupted by a usage-limit, tool-call-limit, or other run interruption, the user's recent work disappears. User described it as: "my initial message was lost 3 times and it disappeared, then when I would say 'continue' it would do a random old task."

Investigation surfaced two distinct failure modes. This PR addresses both.

- **Mode 1** — rate-limit (or other pre-stream rejection) at turn start: the user's text only ever lives in the optimistic `useChat` bubble; the backend rejects before the message is persisted, so the bubble is a lie and a refresh / retry would lose the text.
- **Mode 2** — long-running turn interrupted mid-stream: the entire turn's progress (assistant text, tool calls, reasoning) vanishes on interruption — what users describe as "the turn is gone."

## Mode 1 — frontend: restore unsent text on 429

Backend can't recover this on its own: `check_rate_limit` raises before `append_and_save_message`, so by the time the 429 surfaces there is no DB row to roll forward. See `autogpt_platform/backend/backend/api/features/chat/routes.py:916-922` (rate-limit check) and `routes.py:945` (later append-and-save).

Frontend fix in `autogpt_platform/frontend/src/app/(platform)/copilot/useCopilotStream.ts`: when `useChat`'s `onError` reports a usage-limit error, we

- drop the optimistic user bubble (DB has no record of it, so leaving it would be a phantom),
- push `lastSubmittedMsgRef.current` back into the composer via the existing `setInitialPrompt` slot — the same slot URL pre-fills use, so `useChatInput`'s `consumeInitialPrompt` effect picks it up automatically,
- clear `lastSubmittedMsgRef` so the dedup guard doesn't block re-send.

In-memory only; surviving a hard refresh while rate-limited is a separate follow-up (would need localStorage persistence with TTL).

Test: `autogpt_platform/frontend/src/app/(platform)/copilot/__tests__/useCopilotStream.test.ts` — verifies the composer is repopulated and the optimistic bubble is dropped on a 429.

### Root cause

The SDK retry loop in `stream_chat_completion_sdk` always rolls back `session.messages` to the pre-attempt watermark on any exception. That rollback is correct **before a retry** so attempt #2 doesn't duplicate attempt #1's content. But it runs **before the retry decision is made**, so when retries are exhausted (or no retry is attempted) the partial work is discarded too.

Three branches of the retry loop ended in a final-failure state with side effects worse than just losing the partial:
- `_HandledStreamError` non-transient: rollback then add error marker — partial gone
- `Exception` with `events_yielded > 0`: rollback then break — **no error marker added either**, so on refresh the chat looks like nothing happened even though the user just watched tokens stream live
- `Exception` non-context-non-transient + the while-`else:` exhaustion path: same, no marker
- Outer except (cancellation, GeneratorExit cleanup): didn't restore captured partial

### Fix

`autogpt_platform/backend/backend/copilot/sdk/service.py`:

1. **`_InterruptedAttempt` dataclass** — holds the rolled-back `partial: list[ChatMessage]` + optional `handled_error: _HandledErrorInfo`. Three methods drive the contract:
   - `capture(session, transcript_builder, transcript_snap, pre_attempt_msg_count)` — slices `session.messages`, restores the transcript, strips trailing error markers to prevent duplicate markers after restore.
   - `clear()` — drops captured state on a successful retry so outer cleanup paths don't replay pre-retry content.
   - `finalize(session, state, display_msg, retryable=...) -> list[StreamBaseResponse]` — re-attaches partial, synthesizes `tool_result` rows for orphan `tool_use` blocks, appends the canonical error marker, and returns the flushed events so the caller can yield them to the client (no double-flush).
2. **`_flush_orphan_tool_uses_to_session(session, state) -> list[StreamBaseResponse]`** — synthesizes `tool_result` rows for any `tool_use` that never resolved before the error so the next turn's LLM context stays API-valid (Anthropic rejects orphan tool_use). Uses the public `adapter.flush_unresolved_tool_calls` and returns the events for the caller to yield.
3. **`_classify_final_failure(...) -> _FinalFailure | None`** — picks the display message + stream code + retryable flag for the final-failure exit. One source of truth for the in-history error marker and the client-facing `StreamError` SSE yield so they can't drift.
4. **Consolidated post-loop emit**: the former three scattered blocks (partial restore + redundant re-flush + two separate `yield StreamError` sites) collapsed to one block driven by `_classify_final_failure` → `_FinalFailure` → `finalize()` → yield events + single `StreamError`.
5. **Adapter `flush_unresolved_tool_calls`** (renamed from `_flush_unresolved_tool_calls` to drop the `# noqa: SLF001` suppressors on cross-module callers).

Each retry-loop rollback site calls `interrupted.capture(...)`; the success break calls `interrupted.clear()`; the post-loop failure block calls `interrupted.finalize(...)` exactly once.

The baseline service already preserves partial work via its existing finally block — no change needed there.

## Tests

Backend (`backend/copilot/sdk/interrupted_partial_test.py`, new, 18 tests):

- `TestInterruptedAttemptCapture` — slice semantics + stale-marker stripping
- `TestInterruptedAttemptFinalize` — appends partial then marker, handles empty partial, no-op on `None` session, flushes unresolved tools between partial and marker, returns flushed events for caller to yield
- `TestFlushOrphanToolUses` — synthesizes `tool_result` rows, returns events, no-op on None state / no unresolved
- `TestClassifyFinalFailure` — handled_error wins, attempts_exhausted, transient_exhausted, stream_err fallback, returns None on success path
- `TestRetryRollbackContract` — end-to-end: capture + finalize yields the exact content the user saw streaming live plus the error marker

1022 total SDK tests pass (baseline + new).

Frontend (`useCopilotStream.test.ts`): 1 new test — `restores the unsent text and drops the optimistic user bubble on 429 usage-limit`.

## Out of scope

- Frontend rendering tweaks for the interrupted-turn marker (existing error-marker rendering already works).
- Refresh-survival of the unsent text in Mode 1 (would require localStorage persistence with TTL) — separate follow-up.
- Hard process-kill / OOM where Python `finally` doesn't run — needs a different mechanism (pod-level checkpoint sweeper).

## Checklist

- [x] My code follows the style guidelines of this project (black/isort/ruff via `poetry run format`)
- [x] I have added relevant unit tests
- [x] I have run lint and tests locally (1022 SDK tests pass)

## Test plan

- [ ] Verify a long-running turn that hits transient-retry exhaustion preserves partial assistant text + tool results in chat history after refresh
- [ ] Verify the next user message after an interrupted turn carries enough context that the model can continue the prior task instead of inventing a new one
- [ ] Verify a successful retry (attempt #1 fails, attempt #2 succeeds) shows ONLY attempt #2's content (no leaked partial from #1)
- [ ] Verify hitting daily usage limit at turn start re-populates the composer with the unsent text and removes the optimistic user bubble

### DIFF
--- a/autogpt_platform/backend/backend/copilot/sdk/service.py
+++ b/autogpt_platform/backend/backend/copilot/sdk/service.py
@@ -3067,7 +3067,7 @@ async def _maybe_prepend_builder_context(
     return block + query_message if block else query_message
 
 
-async def stream_chat_completion_sdk(
+async def stream_chat_completion_sdk(  # pyright: ignore[reportGeneralTypeIssues]
     session_id: str,
     message: str | None = None,
     is_user_message: bool = True,
@@ -3080,6 +3080,12 @@ async def stream_chat_completion_sdk(
     request_arrival_at: float = 0.0,
     **_kwargs: Any,
 ) -> AsyncGenerator[StreamBaseResponse, None]:
+    # Pyright's complexity heuristic bails on this function (~1500 LoC, retry
+    # loop with context-overflow fallback + transient backoff + partial-work
+    # preservation). Splitting further would hurt readability — the branches
+    # share state (session, adapter, transcript builder, token accumulators)
+    # that's hard to pass cleanly through helpers. Suppress the bailout; real
+    # type errors elsewhere in the file remain surfaced.
     """Stream chat completion using Claude Agent SDK.
 
     Args:
@@ -3247,6 +3253,11 @@ async def stream_chat_completion_sdk(
     # failed attempt. On final-failure exit we re-attach these so the user sees
     # what the assistant produced before the error rather than an empty chat.
     last_attempt_partial: list[ChatMessage] = []
+    # Populated by the _HandledStreamError branch with (display_msg, code,
+    # already_yielded). Consumed once after the retry loop to re-attach the
+    # partial and, if the inner handler hadn't already emitted one, yield a
+    # single StreamError to the client.
+    handled_error_info: tuple[str, str, bool] | None = None
     # Defaults ensure the finally block can always reference these safely even when
     # an early return (e.g. sdk_cwd error) skips their normal assignment below.
     sdk_model: str | None = None
@@ -3965,28 +3976,12 @@ async def stream_chat_completion_sdk(
                 # attempt that no longer match session.messages.  Skip upload
                 # so a future --resume doesn't replay rolled-back content.
                 skip_transcript_upload = True
-                # Re-attach the rolled-back partial work + add error marker.
-                # Without partial restoration the user's UI streamed tokens
-                # live but a refresh shows nothing happened.
-                _restore_partial_with_error_marker(
-                    session,
-                    state,
-                    last_attempt_partial,
+                handled_error_info = (
                     exc.error_msg or FRIENDLY_TRANSIENT_MSG,
-                    retryable=True,
+                    exc.code or "transient_api_error",
+                    exc.already_yielded,
                 )
                 ended_with_stream_error = True
-                # For transient errors the StreamError was deliberately NOT
-                # yielded inside _run_stream_attempt (already_yielded=False)
-                # so the client didn't see a premature error flash.  Yield it
-                # now that we know retries are exhausted.
-                # For non-transient errors (circuit breaker, idle timeout)
-                # already_yielded=True — do NOT yield again.
-                if not exc.already_yielded:
-                    yield StreamError(
-                        errorText=exc.error_msg or FRIENDLY_TRANSIENT_MSG,
-                        code=exc.code or "transient_api_error",
-                    )
                 break
             except Exception as e:
                 stream_err = e
@@ -4020,19 +4015,6 @@ async def stream_chat_completion_sdk(
                         events_yielded,
                     )
                     skip_transcript_upload = True
-                    # Restore the streamed partial + add error marker — without
-                    # this the frontend would briefly show streamed text live
-                    # then a refresh would show an empty turn.
-                    safe_err = (
-                        str(stream_err).replace("\n", " ").replace("\r", "")[:500]
-                    )
-                    _restore_partial_with_error_marker(
-                        session,
-                        state,
-                        last_attempt_partial,
-                        _friendly_error_text(safe_err),
-                        retryable=False,
-                    )
                     ended_with_stream_error = True
                     break
                 # Transient API errors (ECONNRESET, 429, 5xx) — retry
@@ -4060,13 +4042,6 @@ async def stream_chat_completion_sdk(
                     # at line ~2310.
                     transient_exhausted = True
                     skip_transcript_upload = True
-                    _restore_partial_with_error_marker(
-                        session,
-                        state,
-                        last_attempt_partial,
-                        FRIENDLY_TRANSIENT_MSG,
-                        retryable=True,
-                    )
                     ended_with_stream_error = True
                     break
 
@@ -4074,16 +4049,6 @@ async def stream_chat_completion_sdk(
                     # Non-context, non-transient errors (auth, fatal)
                     # should not trigger compaction — surface immediately.
                     skip_transcript_upload = True
-                    safe_err = (
-                        str(stream_err).replace("\n", " ").replace("\r", "")[:500]
-                    )
-                    _restore_partial_with_error_marker(
-                        session,
-                        state,
-                        last_attempt_partial,
-                        _friendly_error_text(safe_err),
-                        retryable=False,
-                    )
                     ended_with_stream_error = True
                     break
                 attempt += 1  # advance to next context-level attempt
@@ -4100,17 +4065,39 @@ async def stream_chat_completion_sdk(
                 _MAX_STREAM_ATTEMPTS,
                 stream_err,
             )
-            # Restore the last attempt's partial work (rolled back by the
-            # exhausted context-level retry) so the user sees what was
-            # produced before the conversation hit the context ceiling.
-            _restore_partial_with_error_marker(
-                session,
-                state,
-                last_attempt_partial,
-                "Your conversation is too long. "
-                "Please start a new chat or clear some history.",
-                retryable=False,
-            )
+        # Restore the rolled-back partial work + add error marker exactly once
+        # per failure mode. Earlier revisions did this inline in each retry-loop
+        # branch; consolidating here keeps the retry loop itself simple enough
+        # for pyright to type-check.
+        if ended_with_stream_error:
+            if handled_error_info is not None:
+                final_msg, _, _ = handled_error_info
+                final_retryable = True
+            elif attempts_exhausted:
+                final_msg = (
+                    "Your conversation is too long. "
+                    "Please start a new chat or clear some history."
+                )
+                final_retryable = False
+            elif transient_exhausted:
+                final_msg = FRIENDLY_TRANSIENT_MSG
+                final_retryable = True
+            elif stream_err is not None:
+                final_msg = _friendly_error_text(
+                    str(stream_err).replace("\n", " ").replace("\r", "")[:500]
+                )
+                final_retryable = False
+            else:
+                final_msg = None
+                final_retryable = False
+            if final_msg is not None:
+                _restore_partial_with_error_marker(
+                    session,
+                    state,
+                    last_attempt_partial,
+                    final_msg,
+                    retryable=final_retryable,
+                )
 
         if ended_with_stream_error and state is not None:
             # Flush any unresolved tool calls so the frontend can close
@@ -4148,6 +4135,16 @@ async def stream_chat_completion_sdk(
                 error_text = _friendly_error_text(safe_err)
                 error_code = "sdk_stream_error"
             yield StreamError(errorText=error_text, code=error_code)
+
+        # _HandledStreamError exits the retry loop with stream_err unset, so
+        # the previous block doesn't fire. Emit the client-facing StreamError
+        # here when the inner handler chose not to (transient errors suppress
+        # the early flash so the client only sees the final error after all
+        # retries are exhausted).
+        if handled_error_info is not None:
+            handled_msg, handled_code, already_yielded = handled_error_info
+            if not already_yielded:
+                yield StreamError(errorText=handled_msg, code=handled_code)
 
         # Copy token usage from retry state to outer-scope accumulators
         # so the finally block can persist them.
--- a/autogpt_platform/backend/backend/copilot/sdk/service.py
+++ b/autogpt_platform/backend/backend/copilot/sdk/service.py
@@ -541,14 +541,7 @@ def _append_error_marker(
     *,
     retryable: bool = False,
 ) -> None:
-    """Append a copilot error marker to *session* so it persists across refresh.
-
-    Args:
-        session: The chat session to append to (no-op if `None`).
-        display_msg: User-visible error text.
-        retryable: If `True`, use the retryable prefix so the frontend
-            shows a "Try Again" button.
-    """
+    """Append a copilot error marker to *session* so it persists across refresh."""
     if session is None:
         return
     prefix = COPILOT_RETRYABLE_ERROR_PREFIX if retryable else COPILOT_ERROR_PREFIX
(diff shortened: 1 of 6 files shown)


autogpt_platform/backend/backend/copilot/sdk/service.py
```diff
@@ -557,6 +557,80 @@ def _append_error_marker(
...
+    pre_attempt_msg_count: int,
+) -> list[ChatMessage]:
+    """Roll back an attempt's session.messages + transcript, capturing the partial.
+
+    The returned list holds the assistant work that was incrementally appended
+    during the failed attempt. The caller passes it to
+    ``_restore_partial_with_error_marker`` on final-failure exit and discards
+    it on a successful retry.
+    """
+    captured = list(session.messages[pre_attempt_msg_count:])
```
majdyz: 🤖 🟠 **Should Fix**: Stale error marker leaks into `captured` (CodeRabbit's "addressed in b1172e2" claim is incorrect).

Failure scenario:
1. `_run_stream_attempt` hits the idle-timeout branch at line 2296 and calls `_append_error_marker(ctx.session, ..., retryable=True)` — `session.messages` now ends with the marker.
2. Same call path at line 2162 (`_check_empty_tool_call_breaker`) also appends a marker before returning.
3. `_run_stream_attempt` then raises `_HandledStreamError(already_yielded=True)` (line 2851).
4. Outer loop hits `except _HandledStreamError` at line 3934 and calls `_rollback_attempt_capturing_partial` here — line 628 slices `session.messages[pre_attempt_msg_count:]` which **includes the marker just appended**.
5. Post-loop consolidation at line 4093 calls `_restore_partial_with_error_marker(..., last_attempt_partial, final_msg, retryable=True)` — extends with the captured partial (with stale marker) then `_append_error_marker` adds a SECOND marker. Now `session.messages` ends with `assistant(stale_marker), assistant(new_marker)`.
6. Worse: if the inner handler also left an unresolved tool_use, `_flush_orphan_tool_uses_to_session` between the two markers produces `assistant(stale_marker), tool(synthetic_result), assistant(new_marker)` — the synthetic `tool_result` follows an error-marker assistant turn that has no `tool_use`, which Anthropic rejects on the next turn.

The b1172e2 commit consolidated the post-loop block but left line 628's slice unchanged, so this is unfixed.

Direct question: should `_rollback_attempt_capturing_partial` strip trailing assistant messages whose content starts with `COPILOT_ERROR_PREFIX`/`COPILOT_RETRYABLE_ERROR_PREFIX` from `captured` before returning, and is there a test that covers the idle_timeout / circuit-breaker → retries-exhausted path so this regression stays caught?

autogpt_platform/backend/backend/copilot/sdk/service.py
```diff
@@ -3982,6 +4065,39 @@ def _on_stderr(line: str) -> None:
...
                 stream_err,
             )
+        # Restore the rolled-back partial work + add error marker exactly once
+        # per failure mode. Earlier revisions did this inline in each retry-loop
+        # branch; consolidating here keeps the retry loop itself simple enough
+        # for pyright to type-check.
+        if ended_with_stream_error:
+            if handled_error_info is not None:
+                final_msg, _, _ = handled_error_info
+                final_retryable = True
```
majdyz: 🤖 🟡 **Nice to Have**: `_HandledStreamError.retryable` is dropped on the way out — `final_retryable=True` is hardcoded.

Failure scenario:
1. `_HandledStreamError(retryable=False, ...)` is raised from somewhere in `_run_stream_attempt` (currently the only callers use `retryable=True` defaulted, but the constructor at line 2072 accepts the param).
2. Line 3979 packs `handled_error_info = (error_msg, code, already_yielded)` — drops `exc.retryable`.
3. Line 4075 forces `final_retryable = True` for any handled-error case.
4. `_restore_partial_with_error_marker(..., retryable=True)` writes `COPILOT_RETRYABLE_ERROR_PREFIX` even when the originating error was non-retryable, so the frontend offers a "Try again" button that will fail immediately.

Today both inner callers (idle_timeout line 2296, circuit_breaker line 2162) consistently use retryable=True so the bug is latent — but the contract drift is silent. Direct question: should `handled_error_info` be a `dataclass` (or 4-tuple) carrying `exc.retryable` so this couples back to the source of truth, or is the assumption "all handled errors are retryable" load-bearing enough to encode as a comment?

autogpt_platform/backend/backend/copilot/sdk/service.py
```diff
@@ -2993,7 +3067,7 @@ async def _maybe_prepend_builder_context(
     return block + query_message if block else query_message
 
 
-async def stream_chat_completion_sdk(
+async def stream_chat_completion_sdk(  # pyright: ignore[reportGeneralTypeIssues]
```
majdyz, line 3179: 🤖 🟠 **Should Fix**: `# pyright: ignore[reportGeneralTypeIssues]` on the function decl silences ALL type errors in the body, not just the complexity bailout.

Failure scenario:
1. A future PR introduces a real type bug inside `stream_chat_completion_sdk` — e.g. accidentally passing `None` where a non-optional `ChatSession` is expected, or a wrong-typed kwargs into `_restore_partial_with_error_marker`.
2. `reportGeneralTypeIssues` is the umbrella rule that covers most type-checker findings (incompatible-assignment, missing-attribute, unbound-variable, return-type, etc.). The function-level `# pyright: ignore[reportGeneralTypeIssues]` makes pyright skip type-checking the entire 1500-LoC body.
3. The bug ships green; the regression is found at runtime by a user.

If the actual bailout is the complexity heuristic, the targeted suppression is `reportComplexity` (or whatever pyright surfaces — the commit message names "complexity heuristic" specifically, not "general type issues"). Direct question: can the suppression be narrowed to the specific rule pyright bails on (one of `reportComplexity` / `reportGeneralIssues` / a `# type: ignore` on a single offending line), so real type errors elsewhere in the function still surface?

autogpt_platform/backend/backend/copilot/sdk/service.py
```diff
@@ -557,6 +557,80 @@ def _append_error_marker(
...
+
+    The returned list holds the assistant work that was incrementally appended
+    during the failed attempt. The caller passes it to
+    ``_restore_partial_with_error_marker`` on final-failure exit and discards
+    it on a successful retry.
+    """
+    captured = list(session.messages[pre_attempt_msg_count:])
+    session.messages = session.messages[:pre_attempt_msg_count]
+    transcript_builder.restore(transcript_snap)  # type: ignore[arg-type]
+    return captured
```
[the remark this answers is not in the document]
majdyz: 🤖 🟠 **Should Fix**: This thread was auto-marked "Addressed in commit b1172e2", but commit b1172e2 only consolidated the post-loop restore block — it did NOT modify `_rollback_attempt_capturing_partial` (line 615-631). The slice at line 628 still captures whatever `_run_stream_attempt` appended to `session.messages` before raising `_HandledStreamError`, which includes the markers from the idle-timeout (line 2296) and circuit-breaker (line 2162) paths. See my comment on line 628 for the full failure scenario.

autogpt_platform/backend/backend/copilot/sdk/service.py
majdyz: 🤖 Fixed in [5406fe9b5](https://github.com/Significant-Gravitas/AutoGPT/commit/5406fe9b5be72005ad521295f944e8642e346c26): `_rollback_attempt_capturing_partial` now strips trailing assistant messages whose content starts with `COPILOT_ERROR_PREFIX` or `COPILOT_RETRYABLE_ERROR_PREFIX`. Loops in case more than one was somehow appended back-to-back. New tests in `interrupted_partial_test.py::TestRollbackCapturingPartial` lock down the contract: marker stripped, multi-marker stripped, non-marker assistant text preserved.

autogpt_platform/backend/backend/copilot/sdk/service.py
majdyz: 🤖 Fixed in [5406fe9b5](https://github.com/Significant-Gravitas/AutoGPT/commit/5406fe9b5be72005ad521295f944e8642e346c26): replaced the `(msg, code, already_yielded)` 3-tuple with a frozen `_HandledErrorInfo` dataclass that also carries `retryable`. The post-loop block now uses `handled_error_info.retryable` instead of hardcoding `True`. Today's two callers still pass `retryable=True` (idle timeout, circuit breaker) so behaviour is unchanged, but the contract is now load-bearing rather than implicit.

autogpt_platform/backend/backend/copilot/sdk/service.py
majdyz: 🤖 Self-correcting on this one — after digging into pyright's behaviour, the suppression is benign (not silencing additional errors). When pyright bails on a function due to complexity, it doesn't analyze the body at all regardless of whether `# pyright: ignore` is present; the suppression only quiets the bailout error itself, it doesn't widen what's silenced. Adding the suppression on the `def` line and removing it produced the same 0-error body analysis. The genuine fix would be to decompose `stream_chat_completion_sdk` into smaller functions — out of scope here, the function shares too much state across branches and the author called this out in the commit message. Resolving as a non-issue.

autogpt_platform/backend/backend/copilot/sdk/service.py
```diff
@@ -3982,6 +4065,39 @@ def _on_stderr(line: str) -> None:
...
+                final_msg = None
+                final_retryable = False
+            if final_msg is not None:
+                _restore_partial_with_error_marker(
+                    session,
+                    state,
+                    last_attempt_partial,
+                    final_msg,
+                    retryable=final_retryable,
+                )
```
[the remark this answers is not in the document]
majdyz: 🤖 Confirmed — same issue as the line-628 thread. Fixed in [5406fe9b5](https://github.com/Significant-Gravitas/AutoGPT/commit/5406fe9b5be72005ad521295f944e8642e346c26): `_rollback_attempt_capturing_partial` now strips trailing error markers from the captured partial, so the post-loop restore can't replay a marker `_run_stream_attempt` already appended. Sentry self-flagged this as resolved.

autogpt_platform/backend/backend/copilot/sdk/service.py
```diff
@@ -4020,6 +4180,20 @@ def _on_stderr(line: str) -> None:
                 error_code = "sdk_stream_error"
             yield StreamError(errorText=error_text, code=error_code)
```
majdyz: 🤖 🟡 **Nice to Have**: Final-error text/code is duplicated across `_classify_final_failure` (history marker) and the `yield StreamError` block below.

Failure scenario:
1. A future PR tweaks the "conversation is too long" wording in `_classify_final_failure` (lines 676-680) to improve clarity.
2. The author doesn’t realise the same string is hardcoded a second time here at lines 4170-4173 for the SSE yield.
3. Client UI (SSE) shows the old copy while the saved history (marker) shows the new copy — or vice versa — and the two drift permanently.

The `_classify_final_failure` helper was introduced to centralise this exact mapping. Consider either having this block call the same helper for its `error_text`, or defining a single `_FINAL_FAILURE_CODES` table keyed on `(attempts_exhausted, transient_exhausted)` that both sites consume.

Question: Is there a reason the SSE yield can’t reuse `_classify_final_failure`, or at minimum share a constants table with it?

autogpt_platform/backend/backend/copilot/sdk/service.py
```diff
@@ -557,6 +557,97 @@ def _append_error_marker(
...
+    with ``tool_use_id without tool_result`` — and any baseline replay would
+    surface the same broken history. Flushing produces interrupted-marker
+    tool_results that satisfy the API contract.
+    """
+    if session is None or state is None:
+        return
+    if not state.adapter.has_unresolved_tool_calls:
+        return
+    safety: list[StreamBaseResponse] = []
+    state.adapter._flush_unresolved_tool_calls(safety)  # noqa: SLF001
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [6576bf561](https://github.com/Significant-Gravitas/AutoGPT/commit/6576bf561e2093959e09b66dd3e9e8d0dadf10cd): valid finding, implemented. Changes:\n\n1. Renamed `_flush_unresolved_tool_calls` → `flush_unresolved_tool_calls` on the adapter (public) and updated all 5 call sites. The `# noqa: SLF001` suppressor on line 650 is gone.\n2. `_flush_orphan_tool_uses_to_session` and `_InterruptedAttempt.finalize` now return the `list[StreamBaseResponse]` they produce so the caller yields them to the client instead of re-flushing. Before: early flush in `finalize()` flipped `has_unresolved_tool_calls` to False → the cleanup block at lines 4185-4200 skipped its flush → UI spinners stayed open until refresh. After: single flush, events flow through the caller.\n3. The three scattered post-loop blocks (partial restore + re-flush + `stream_err` yield + `handled_error` yield) collapsed into one block driven by `_classify_final_failure` → `_FinalFailure`. Eliminates both the skip-cleanup bug and the duplicated error-text/code strings.

autogpt_platform/backend/backend/copilot/sdk/service.py
majdyz: 🤖 Fixed in [6576bf561](https://github.com/Significant-Gravitas/AutoGPT/commit/6576bf561e2093959e09b66dd3e9e8d0dadf10cd): `_classify_final_failure` now returns a `_FinalFailure(display_msg, code, retryable)` dataclass instead of `(msg, retryable)`. The post-loop block at the former lines 4200-4234 (three separate `yield StreamError` sites with hardcoded text/code) collapsed to a single yield that reads `failure.display_msg` + `failure.code`. Same source of truth as the in-history marker — the two cannot drift.

majdyz:
🤖 @coderabbitai — the duplicate-comments section in this review (line 634-650: "Don't consume the adapter flush before the client cleanup path runs") was already addressed in [6576bf561](https://github.com/Significant-Gravitas/AutoGPT/commit/6576bf561e2093959e09b66dd3e9e8d0dadf10cd):\n\n- `_flush_unresolved_tool_calls` renamed to `flush_unresolved_tool_calls` (public) and all call sites updated. The `# noqa: SLF001` suppressor is gone.\n- `_flush_orphan_tool_uses_to_session` and `_InterruptedAttempt.finalize` now return the `list[StreamBaseResponse]` they produce. The post-loop caller yields those events directly instead of re-flushing — no more double-flush, no skipped UI cleanup.\n- The three former post-loop blocks (partial restore + redundant re-flush + two separate `yield StreamError` sites) collapsed to one consolidated block driven by `_classify_final_failure` returning a `_FinalFailure(display_msg, code, retryable)` dataclass. History marker and SSE yield now share one source of truth.

majdyz:
## E2E Live-Stack Test Report — PR #12918

**Date:** 2026-04-25
**Tested HEAD:** `40349c787` (after frontend draft-restore landed)
**Worktree:** `/Users/majdyz/Code/AutoGPT2`
**Mode:** native dev stack — `poetry run app` + docker-compose infra deps + real `CHAT_USE_CLAUDE_AGENT_SDK=true` + real Anthropic API key
**Driver:** direct REST API (UI onboarding skipped to reduce ceremony — same backend code path)

### Verdict: APPROVE — all live scenarios PASS, both modes of SECRT-2275 verified end-to-end.

### Scenario 1: Happy path
- **Command:** POST `/api/chat/sessions/{id}/stream` with `"Reply with exactly: hello world"`
- **Expected:** Clean SSE sequence, assistant + user msgs in DB.
- **Actual:** `start → start-step → text-start → text-delta → text-end → finish-step → usage → finish → DONE`. DB has `[user("Reply with exactly: hello world"), assistant("hello world")]`.
- **Result:** PASS — no regression from the consolidation refactor.

### Scenario 2: Rate-limit at turn start (Mode 1)
- **Setup:** `CHAT_DAILY_COST_LIMIT_MICRODOLLARS=1` + injected 999_999_999 µ$ into `copilot:cost:daily:{user}:2026-04-25` redis key, restarted backend.
- **Command:** POST a new message into the existing 6-msg session.
- **Expected:** HTTP 429, prior history intact, rejected user message NOT persisted (since backend raises before `append_and_save_message`).
- **Actual:**
  - `HTTP 429`, body: `{"detail":"You've reached your daily usage limit. Resets in 20h 0m."}`
  - `GET /api/chat/usage`: `percent_used: 100.0`
  - `GET /api/chat/sessions/{id}`: returns `200`, **6 prior msgs intact**, latest user message NOT in DB.
  - `GET /api/chat/sessions` (list): `200` — chat is fully functional.
- **Result:** PASS — confirms backend cannot recover the lost user text by design (rate-limit check at [routes.py:916-922](autogpt_platform/backend/backend/api/features/chat/routes.py#L916-L922) raises before [routes.py:945](autogpt_platform/backend/backend/api/features/chat/routes.py#L945) `append_and_save_message`). This is the gap the **frontend draft-restore** in `40349c787` closes.

### Scenario 3: Synthetic tool-call limit graceful finish
- **Setup:** `CHAT_AGENT_MAX_TURNS=2`, restarted backend.
- **Command:** "Run three bash commands one at a time: `uname -a`, `pwd`, `whoami`." in a new session.
- **Expected:** SDK CLI hits the 2-turn cap mid-task, finishes gracefully per PR #12892.
- **Actual:** HTTP 200, **zero error events**. SSE has 2 `tool-input-available`, 2 `tool-output-available`, 2 `start-step`/`finish-step`, 2 `reasoning-start`/`end`, 1 `finish`. DB has 7 rows: `user → reasoning → assistant(tool_calls) → tool(uname result) → reasoning → assistant(text formatted result) → tool(pwd result)`.
- **Result:** PASS — middle conversation fully preserved; only the third command (`whoami`) didn't run, exactly per max-turns design.

### Scenario 4: Mid-stream failure (Mode 2 — the fix's actual target path)
- **Setup:** Added a temp `CHAT_TEST_INJECT_FAIL_AFTER_TOOL` env-gated `raise RuntimeError("ECONNRESET (test injection)")` in `_dispatch_response` right after the first `tool_result` is appended to `session.messages`. Re-used the 6-msg session.
- **Command:** "Run two bash commands one at a time: `uname -a`, then `pwd`."
- **Expected:** Inject fires after first tool round → SDK retry-loop's `except Exception` fires with `events_yielded > 0` → my fix's `_InterruptedAttempt.capture()` then `_classify_final_failure()` then `finalize()` re-attaches partial + adds error marker, all 6 prior msgs untouched.
- **Actual:** SSE ends with `error: "SDK stream error: ECONNRESET (test injection after first tool)"`. DB after: **11 messages** —
  - rows 1-6: 6 prior msgs **intact**
  - row 7: user (the new turn's input)
  - row 8: reasoning ("The user wants me to run two bash commands one at a time…")
  - row 9: assistant with `tool_calls=[toolu_01Dy…]`
  - row 10: tool with `tool_call_id=toolu_01Dy…` (the `uname -a` partial result — the work that would have been lost without the fix)
  - row 11: assistant with `[__COPILOT_ERROR_f7a1__] SDK stream error: ECONNRESET …`
  
  `tool_calls[].id` ↔ `tool_msg.tool_call_id` paired 1:1 (no orphan).
- **Result:** PASS — the fix's core invariant (partial assistant work + tool result + error marker preserved across rollback, prior session history untouched) holds end-to-end on real SDK + real API.

### Scenario 5: Post-cleanup verification
- **Setup:** Reverted the temp injection, removed env var, synced working tree to `origin/40349c787`, restarted backend.
- **Command:** Happy-path send in a new session.
- **Actual:** Clean SSE sequence, 2 msgs persisted, zero error events.
- **Result:** PASS — no residual state from injection.

### Cleanup
- Temp injection removed from working tree (`git diff` clean against `origin/40349c787`).
- Env vars `CHAT_DAILY_COST_LIMIT_MICRODOLLARS`, `CHAT_AGENT_MAX_TURNS`, `CHAT_TEST_INJECT_FAIL_AFTER_TOOL` reverted.
- Redis daily-cost counter reset.
- Testing lock released.

### Notes
- Frontend (Mode 1) and backend (Mode 2) fixes are now both on `40349c787`. The 1012 SDK tests + 48 retry-scenarios + 14 interrupted-partial unit tests + 3 frontend tests already verified the deterministic paths offline; this report adds end-to-end live confirmation against real Anthropic SDK + CLI tool execution.

PR #12920, opened 2026-04-25
fix(frontend): redirect www to non-www with 308 to preserve request method (#9188)

### LINKED ISSUE #9188: Adding "www." to the start of the url breaks the platform

SymbolStar:
## Summary
Fixes #9188 — redirects `www.` to non-www to prevent cookie/auth domain mismatch.

Uses **308** (Permanent Redirect) instead of 301, which preserves the HTTP method and request body. This is important because the middleware matcher runs on `/auth/authorize` and `/auth/integrations/*`, where OAuth callbacks may use POST.

Split from #12895 per reviewer request.

## Changes
- Added www→non-www redirect in Next.js middleware using `NextResponse.redirect(url, 308)`

### DIFF
--- a/autogpt_platform/frontend/src/middleware.test.ts
+++ b/autogpt_platform/frontend/src/middleware.test.ts
@@ -1,0 +1,47 @@
+import { describe, expect, it, vi } from "vitest";
+import { NextRequest } from "next/server";
+
+const updateSessionMock = vi.fn();
+
+vi.mock("@/lib/supabase/middleware", () => ({
+  updateSession: (...args: unknown[]) => updateSessionMock(...args),
+}));
+
+import { middleware } from "./middleware";
+
+describe("middleware www→non-www redirect", () => {
+  it("redirects www host to non-www with 308", async () => {
+    const request = new NextRequest("https://www.example.com/dashboard");
+
+    const response = await middleware(request);
+
+    expect(response).toBeDefined();
+    expect(response.status).toBe(308);
+    expect(response.headers.get("location")).toBe(
+      "https://example.com/dashboard",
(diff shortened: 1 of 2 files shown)


SymbolStar:
recheck

SymbolStar:
FYI — the failing CI checks on this PR are all infrastructure-related and not caused by this change:

- **Vercel**: `Authorization required to deploy` — expected for fork PRs, requires org-level auth
- **codecov/patch**: coverage gate — no new code paths to cover in this 1-line fix
- **Check PR Status**: this meta-check polls other checks and fails because Vercel is failing

All code-level checks (lint, CodeQL, conflicts, scope, size) pass ✅

majdyz (approved):

PR #12922, opened 2026-04-25
fix(util/truncate): keep output within size_limit and stop leaking originals at limit=1

alvinttang:
## Summary

`autogpt_platform/backend/backend/util/truncate.py::_truncate_string_middle` reserved the entire `limit` budget for kept content (`head_len = limit//2`, `tail_len = limit - head_len`) and then *appended* the `"… (omitted N chars)…"` sentinel on top — so `truncate("a"*1000, 50)` returned 72 chars, a ~22-char overshoot every call. The fast path in `truncate()` for plain strings exposes this directly.

Secondary bug: when `limit == 1`, `tail_len == 0` and `value[-0:]` returns the entire original string unredacted, defeating truncation entirely.

## Fix

Compute the worst-case sentinel length `len("… (omitted <len(value)> chars)…")` and subtract it from the budget *before* splitting head/tail. When `limit <= sentinel_len`, fall back to a hard `value[:limit]` slice. Guard `tail = value[-tail_len:] if tail_len > 0 else ""` to avoid the `[-0:]` smuggling.

24/-4 production change.

## Test

`autogpt_platform/backend/backend/util/truncate_test.py` (10 parametrized tests). Asserts:
- `len(str(out)) <= size_limit` for plain strings;
- head/tail preserved;
- small-limit no longer leaks original;
- container truncation stays within `limit + sentinel_overhead`;
- scalars passthrough.

RED 6/10 on unfixed code → GREEN 10/10 after. `ruff` clean.

## Risk notes

- Output is now strictly tighter — head/tail get `limit - sentinel_len` characters instead of `limit/2` each. All in-tree call sites use this for log/UI truncation where the contract was "fits in N chars", so the new behaviour matches the contract.
- For very large `len(value)` the sentinel grows logarithmically; still bounded.

### DIFF
--- a/autogpt_platform/backend/backend/util/truncate.py
+++ b/autogpt_platform/backend/backend/util/truncate.py
@@ -7,15 +7,35 @@ from typing import Any
 
 
 def _truncate_string_middle(value: str, limit: int) -> str:
-    """Shorten *value* to *limit* chars by removing the **middle** portion."""
+    """Shorten *value* to at most *limit* chars by removing the **middle** portion.
+
+    The returned string (head + sentinel + tail) is guaranteed to be no longer
+    than *limit*. The sentinel ``"… (omitted N chars)…"`` itself counts toward
+    the budget — previously it was added on top, so callers always overshot
+    the limit by ~22 chars and ``truncate(value, N)`` violated its own
+    docstring contract. For very small *limit* values where the sentinel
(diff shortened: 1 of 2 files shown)


ntindle:
closing for now due to unsigned cla :) feel free to reopen when signed :)

PR #12926, opened 2026-04-27
fix(backend/copilot): surface empty-success ResultMessage as stream error (SECRT-2252)

majdyz:
## Summary

- Detect ghost-finished sessions where the SDK returns a `ResultMessage` with `subtype="success"`, empty `result`, no produced content, and `output_tokens == 0`.
- Emit `StreamError(code="empty_completion")` instead of silently calling `StreamFinish`, so the caller (and the user) sees the failure.

## Background

Linear: [SECRT-2252](https://linear.app/agpt/issue/SECRT-2252) — SDK silent empty completion not retried, leaving the user with a blank stream (`start -> start-step -> finish-step -> finish`).

## Changes

- `response_adapter.py::convert_message`: in the `ResultMessage` branch, check `_is_empty_completion()` before falling through to the existing success path. When matched, close any open step, emit `StreamError`, and skip `StreamFinish`.
- `response_adapter.py::_is_empty_completion`: new helper that returns `True` only when `result` is falsy, no text/reasoning was emitted, no tool calls were registered, no tool results were seen, and `usage["output_tokens"]` is `0`.
- `response_adapter_test.py`: 4 new unit tests covering empty-success (None and empty-string variants), non-empty success, and the non-empty-tokens-but-empty-result fallthrough.

## Out of scope (per ticket)

- Retry-once behavior. This PR only surfaces the error; the caller decides retry semantics. Follow-up work can wire automatic retry on `code="empty_completion"`.

## Test plan

- [x] `poetry run pytest backend/copilot/sdk/response_adapter_test.py` — all 58 tests pass (4 new + 54 existing).
- [x] `poetry run pyright backend/copilot/sdk/response_adapter.py backend/copilot/sdk/response_adapter_test.py` — clean.

## Checklist

- [x] My code follows the style of this project.
- [x] I have added tests covering my changes.
- [x] I have updated the documentation accordingly. (N/A — internal adapter behavior)

### DIFF
--- a/autogpt_platform/backend/backend/copilot/sdk/response_adapter.py
+++ b/autogpt_platform/backend/backend/copilot/sdk/response_adapter.py
@@ -376,6 +376,27 @@ class SDKResponseAdapter:
 
         elif isinstance(sdk_message, ResultMessage):
             self.flush_unresolved_tool_calls(responses)
+            # SECRT-2252: surface ghost-finished sessions as errors instead of silent finishes.
+            if sdk_message.subtype == "success" and self._is_empty_completion(
+                sdk_message
+            ):
+                if self.step_open:
+                    responses.append(StreamFinishStep())
+                    self.step_open = False
+                responses.append(
+                    StreamError(
+                        errorText=(
+                            "The model returned an empty response. Please retry."
+                        ),
+                        code="empty_completion",
+                    )
+                )
+                logger.warning(
+                    "[SDK] [%s] Empty-success ResultMessage detected — "
+                    "emitting stream error instead of silent finish",
+                    (self.session_id or "?")[:12],
+                )
+                return responses
             # Thinking-only final turn guard: when the model's last LLM
             # call after a tool result produced only a ``ThinkingBlock``
             # (no ``TextBlock``, no ``ToolUseBlock``) the UI has nothing
@@ -436,6 +457,25 @@ class SDKResponseAdapter:
             logger.debug(f"Unhandled SDK message type: {type(sdk_message).__name__}")
 
         return responses
+
+    def _is_empty_completion(self, msg: ResultMessage) -> bool:
+        """True when a success ResultMessage carries no content at all.
+
+        Detects the SDK's ghost-finished session: empty ``result``, zero
+        ``output_tokens``, and nothing emitted on the wire this turn (no
+        text, no reasoning, no tool calls).
+        """
+        if msg.result:
+            return False
+        if self.has_started_text or self.has_started_reasoning:
+            return False
+        if self.current_tool_calls:
+            return False
+        if self._any_tool_results_seen:
+            return False
+        usage = msg.usage or {}
+        output_tokens = usage.get("output_tokens") or 0
+        return output_tokens == 0
 
     def _ensure_text_started(self, responses: list[StreamBaseResponse]) -> None:
         """Start (or restart) a text block if needed."""
--- a/autogpt_platform/backend/backend/copilot/sdk/response_adapter.py
+++ b/autogpt_platform/backend/backend/copilot/sdk/response_adapter.py
@@ -385,12 +385,16 @@ class SDKResponseAdapter:
                     self.step_open = False
                 responses.append(
                     StreamError(
-                        errorText=(
-                            "The model returned an empty response. Please retry."
-                        ),
+                        errorText="The model returned an empty response.",
                         code="empty_completion",
                     )
                 )
+                # Pair with StreamFinish so ``acc.stream_completed`` flips True
+                # in ``_dispatch_response`` — without it the service-layer
+                # post-stream branch mis-classifies the turn as "stopped by
+                # user" and appends a STOPPED_BY_USER_MARKER on top of the
+                # error marker.
+                responses.append(StreamFinish())
                 logger.warning(
                     "[SDK] [%s] Empty-success ResultMessage detected — "
                     "emitting stream error instead of silent finish",
--- a/autogpt_platform/backend/backend/copilot/sdk/response_adapter_test.py
+++ b/autogpt_platform/backend/backend/copilot/sdk/response_adapter_test.py
@@ -565,6 +565,101 @@ def test_result_success_does_not_synthesize_when_no_tools_ran():
     assert text_deltas == []
 
 
+def test_result_empty_success_emits_error_and_no_finish():
+    """SECRT-2252: a ``subtype="success"`` ResultMessage with empty ``result``,
+    no produced content, and ``output_tokens == 0`` is the SDK's ghost-finish
+    bug.  The adapter should surface it as a ``StreamError`` and skip
+    ``StreamFinish`` so the caller can decide retry semantics."""
+    adapter = _adapter()
+    adapter.convert_message(SystemMessage(subtype="init", data={}))
+    msg = ResultMessage(
+        subtype="success",
+        duration_ms=100,
+        duration_api_ms=50,
+        is_error=False,
+        num_turns=1,
+        session_id="s1",
+        result=None,
+        usage={"input_tokens": 5, "output_tokens": 0},
+    )
+    results = adapter.convert_message(msg)
+    types = [type(r).__name__ for r in results]
+    assert "StreamError" in types
+    assert "StreamFinish" not in types
+    err = next(r for r in results if isinstance(r, StreamError))
+    assert err.code == "empty_completion"
+    assert "empty response" in err.errorText.lower()
+
+
+def test_result_empty_success_with_empty_string_result_treated_as_empty():
+    """An empty string (not just None) for ``result`` is also empty."""
+    adapter = _adapter()
+    adapter.convert_message(SystemMessage(subtype="init", data={}))
+    msg = ResultMessage(
+        subtype="success",
+        duration_ms=100,
+        duration_api_ms=50,
+        is_error=False,
+        num_turns=1,
+        session_id="s1",
+        result="",
+        usage={"output_tokens": 0},
+    )
+    results = adapter.convert_message(msg)
+    assert any(isinstance(r, StreamError) for r in results)
+    assert not any(isinstance(r, StreamFinish) for r in results)
+
+
+def test_result_success_with_text_emits_finish_not_error():
+    """Non-empty success (text was produced) keeps the existing
+    ``StreamFinish`` behaviour — no spurious error."""
+    adapter = _adapter()
+    adapter.convert_message(
+        AssistantMessage(content=[TextBlock(text="hello")], model="test")
+    )
+    msg = ResultMessage(
+        subtype="success",
+        duration_ms=100,
+        duration_api_ms=50,
+        is_error=False,
+        num_turns=1,
+        session_id="s1",
+        result="hello",
+        usage={"output_tokens": 5},
+    )
+    results = adapter.convert_message(msg)
+    types = [type(r).__name__ for r in results]
+    assert "StreamFinish" in types
+    assert "StreamError" not in types
+
+
+def test_result_success_with_nonzero_output_tokens_not_empty():
+    """If ``output_tokens > 0`` but ``result`` is empty (e.g. all tokens went
+    to thinking), don't classify as empty — fall through to the existing
+    success path."""
+    adapter = _adapter()
+    adapter.convert_message(
+        AssistantMessage(
+            content=[ThinkingBlock(thinking="reasoning", signature="sig")],
+            model="test",
+        )
+    )
+    msg = ResultMessage(
+        subtype="success",
+        duration_ms=100,
+        duration_api_ms=50,
+        is_error=False,
+        num_turns=1,
+        session_id="s1",
+        result="",
+        usage={"output_tokens": 50},
+    )
+    results = adapter.convert_message(msg)
+    types = [type(r).__name__ for r in results]
+    assert "StreamFinish" in types
+    assert "StreamError" not in types
+
+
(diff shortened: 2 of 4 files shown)


autogpt_platform/backend/backend/copilot/sdk/response_adapter.py
```diff
@@ -376,6 +376,27 @@ def convert_message(self, sdk_message: Message) -> list[StreamBaseResponse]:
...
+                        ),
+                        code="empty_completion",
+                    )
+                )
+                logger.warning(
+                    "[SDK] [%s] Empty-success ResultMessage detected — "
+                    "emitting stream error instead of silent finish",
+                    (self.session_id or "?")[:12],
+                )
+                return responses
```
majdyz, line 403: 🤖 🔴 **Blocker**: Returning early after `StreamError` (no `StreamFinish`) breaks the post-stream pipeline.

**Failure scenario** (file:line evidence):

1. Adapter emits `StreamError(code="empty_completion")` and returns at `response_adapter.py:399` — no `StreamFinish` follows.
2. In `service.py:2702`, `_dispatch_response` runs for the `StreamError` and calls `_append_error_marker(retryable=False)` (`service.py:2029`) — frontend renders a non-retryable error card.
3. The for-loop continues but no more SDK messages arrive after `ResultMessage`. `acc.stream_completed` stays `False` (only `StreamFinish` sets it at `service.py:2135`). `ended_with_stream_error` is also `False` (the empty-completion path never sets it).
4. Post-loop, we hit `service.py:2934` — `if not acc.stream_completed and not ended_with_stream_error:` triggers and appends `STOPPED_BY_USER_MARKER` to `session.messages`. Now the session ends with TWO bogus rows: the error marker + the "stopped by user" marker.
5. `_HandledStreamError` is never raised, so the outer retry loop has no chance to retry — even after a follow-up PR adds retry semantics, this code path bypasses the retry infrastructure entirely.

**Question**: should the new branch (a) also append `StreamFinish` after the error (matching the existing `subtype="error"` path at `response_adapter.py:442-449` which does both), or (b) signal the service layer (via a new sentinel response) so it can set `ended_with_stream_error = True` + raise `_HandledStreamError` for retry-loop integration?

The adapter-only test passes, but this is a service-level behavior break that no test in this PR exercises.

autogpt_platform/backend/backend/copilot/sdk/response_adapter.py
```diff
@@ -376,6 +376,27 @@ def convert_message(self, sdk_message: Message) -> list[StreamBaseResponse]:
...
+            if sdk_message.subtype == "success" and self._is_empty_completion(
+                sdk_message
+            ):
+                if self.step_open:
+                    responses.append(StreamFinishStep())
+                    self.step_open = False
+                responses.append(
+                    StreamError(
+                        errorText=(
+                            "The model returned an empty response. Please retry."
```
majdyz: 🤖 🟠 **Should Fix**: Error message says "Please retry" but the error is non-retryable in the UI.

**Failure scenario**:

1. Adapter emits `StreamError(errorText="The model returned an empty response. Please retry.", code="empty_completion")` at `response_adapter.py:386-393`.
2. `_dispatch_response` at `service.py:2026-2030` maps to `_append_error_marker(retryable=(response.code == "transient_api_error"))` — only `transient_api_error` produces a retryable marker.
3. Frontend `helpers.ts:215-222` renders `retryable_error` markers with a retry button; plain `error` markers (our case) render an `ErrorCard` with no retry action.
4. Result: user sees the message "Please retry." but has no retry button — the only recovery is to type a new message manually, which loses the context the user expected to retry.

This is similar to `idle_timeout` (`service.py:2406`) and `circuit_breaker_empty_tool_calls` (`service.py:2272`), both of which call `_append_error_marker(..., retryable=True)` for exactly this reason.

**Question**: should `_dispatch_response` add `code == "empty_completion"` to the retryable allowlist (so the user-facing copy matches the UI affordance), or should the message change to "The model returned an empty response." without the "Please retry." instruction since this PR is explicitly punting retry to follow-up work?

autogpt_platform/backend/backend/copilot/sdk/response_adapter.py
majdyz: 🤖 Fixed in [fd2f46688](https://github.com/Significant-Gravitas/AutoGPT/commit/fd2f46688c66b3a504e7def3ca874f1ec085cd48): adapter now appends `StreamFinish` after the `StreamError` (`response_adapter.py:393-399`), so `_dispatch_response` flips `acc.stream_completed = True` and the post-stream branch at `service.py:2934` no longer mis-classifies the turn as "stopped by user". Test asserts both `StreamError` and `StreamFinish` are emitted in that order.

autogpt_platform/backend/backend/copilot/sdk/response_adapter.py
majdyz: 🤖 Fixed in [fd2f46688](https://github.com/Significant-Gravitas/AutoGPT/commit/fd2f46688c66b3a504e7def3ca874f1ec085cd48): added `_RETRYABLE_STREAM_ERROR_CODES = {"transient_api_error", "empty_completion"}` and `_dispatch_response` now consults that set (`service.py:2034-2038`). Empty-completion now flows through as a retryable marker so the UI shows a retry button. Also dropped the "Please retry." instruction from the user-facing message — the affordance speaks for itself.

majdyz:
🤖 @coderabbitai both nitpicks fixed in [fc938ec29](https://github.com/Significant-Gravitas/AutoGPT/commit/fc938ec295c9c69e369fee97a2fdad9c01492599):

- Dropped the `"empty response" in err.errorText.lower()` assertion — the `err.code == "empty_completion"` line above is the stable contract.
- Removed the `AssistantMessage(ThinkingBlock(...))` setup from `test_result_success_with_nonzero_output_tokens_not_empty` so the `output_tokens` guard is the only thing preventing `_is_empty_completion()` from firing — that's the case the test name actually claims to cover.

autogpt_platform/backend/backend/copilot/sdk/response_adapter_test.py
```diff
@@ -565,6 +565,100 @@ def test_result_success_does_not_synthesize_when_no_tools_ran():
...
+        is_error=False,
+        num_turns=1,
+        session_id="s1",
+        result="",
+        usage={"output_tokens": 0},
+    )
+    results = adapter.convert_message(msg)
+    assert any(isinstance(r, StreamError) for r in results)
+    assert any(isinstance(r, StreamFinish) for r in results)
+
```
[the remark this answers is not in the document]
majdyz: 🤖 Addressed in [c04a7bd76f](https://github.com/Significant-Gravitas/AutoGPT/commit/c04a7bd76fbe2ddf9a05c79e3ec51b6a7e10c517):

- `test_result_empty_success_emits_error_and_finish` now asserts `StreamFinishStep` precedes `StreamError`. `SystemMessage(subtype="init")` already opens a step (`response_adapter.py:144`), so the empty-completion branch at `response_adapter.py:383-385` must close it before emitting the error — that ordering is now wire-locked.
- The empty-string variant now asserts `err.code == "empty_completion"` (replacing the loose `any(isinstance(_, StreamError))` check) so a regression in the error code surfaces immediately.

Note on the suggested "open a step from a partial AssistantMessage" — adding a real `AssistantMessage(TextBlock(text="partial"))` would set `has_started_text=True`, which makes `_is_empty_completion()` return `False` (`response_adapter.py:474`). The `init`-opened step is the correct fixture for this branch since the detection requires *no* text/reasoning/tool calls were produced.

ntindle (approved):

majdyz:
## E2E Test Report — PR #12926 (SECRT-2252)

Native stack (`poetry run app` + `pnpm dev` against the local Supabase / Redis / RabbitMQ docker infra) on `zamilmajdy/secrt-2252-sdk-silent-empty-completion-not-retried-ghost-finished` @ `c04a7bd76f`.

### Summary

| 1 | Stack came up cleanly | PASS | Backend healthy on `:8006` (`/health` → 200), frontend on `:3000`, all 7 supervised processes started (DatabaseManager / Scheduler / NotificationManager / PlatformLinkingManager / WebsocketServer / ExecutionManager / AgentServer). Log: `Application startup complete. Uvicorn running on http://0.0.0.0:8006` |
| 2 | Adapter-level test suite (live PR code) | PASS | 79/79 — `pytest backend/copilot/sdk/response_adapter_test.py backend/copilot/sdk/interrupted_partial_test.py` (5.77s). No regressions on partial-streaming, thinking-only, summary-walk, interrupted-attempt, or flush-orphan tool-use paths. |
| 3 | Empty-completion path emits the right shape | PASS | `test_result_empty_success_emits_error_and_finish` + `test_result_empty_success_with_empty_string_result_treated_as_empty` both PASS. Both `result=None` and `result=""` plus `output_tokens==0` and no produced content are detected. The init-opened step is closed, then `StreamError(code="empty_completion", errorText="The model returned an empty response.")` is emitted, then `StreamFinish` — in that order. |
| 4 | Retryable-codes contract enforced | PASS | `TestRetryableStreamErrorCodes` (3/3): `empty_completion` and `transient_api_error` both in `_RETRYABLE_STREAM_ERROR_CODES`; `sdk_error` and `all_attempts_exhausted` are not. `_dispatch_response` flips `retryable=True` for the new code via the `code in _RETRYABLE_STREAM_ERROR_CODES` check (service.py:2037). |
| 5 | Live-adapter integration drive (no mocks) | PASS | Constructed a real `SDKResponseAdapter` against PR head, drove a synthesized `ResultMessage(subtype="success", result=None, usage={"output_tokens":0})` through `convert_message`, observed actual production `_is_empty_completion()` fire and emit `[StreamFinishStep, StreamError, StreamFinish]` in correct order. Also verified the negative case (`result="hello world"`, `output_tokens=5`) emits only `[StreamFinishStep, StreamFinish]` — no false positive. Backend log captured the new warning: `[SDK] [LIVE-validat] Empty-success ResultMessage detected — emitting stream error instead of silent finish`. |
| 6 | Live copilot regression (round-trip 1) | PASS | Logged in as `test@test.com`, opened `/copilot`, sent "Hello, what is 2+2? Answer in one short sentence." Stream completed in ~3 s; backend log: `[SDK][a3b51aa1-564][T2] Stream completed successfully with 4 messages`, `Saving 1 new messages to DB ... roles=['assistant']`, then `event_generator FINISHED in 2.97s`. No `empty_completion` warning fired. |
| 7 | Live copilot regression (round-trip 2) | PASS | Sent a second message in the same session; backend persisted 4th assistant message; session API returns `roles=['user','assistant','user','assistant']` (ChatMessage rows in `platform.ChatMessage`). Idle UI returned to ready state both times. Guard did NOT misfire on either turn. |

### Verdict

**PASS.** The empty-completion guard works end-to-end:
- The two new tests cover both `result=None` and `result=""` shapes, with explicit ordering assertions.
- The `_RETRYABLE_STREAM_ERROR_CODES` indirection is exercised by live `_dispatch_response` consumption (service.py:2037).
- The live adapter drive against the PR head reproduces the documented `[StreamFinishStep, StreamError(empty_completion), StreamFinish]` sequence and confirms the negative case does not misfire.
- Two real copilot round-trips in the running stack persisted assistant messages cleanly with no `empty_completion` events in backend logs — the guard does not break the normal success path.

### Why no live-UI reproduction of the bug

The bug requires the Claude Code SDK to return a malformed `ResultMessage(subtype="success", result=None/"")` — which can only happen when the SDK itself ghost-finishes. There is no UI lever to force this contract violation. Coverage is provided instead by:
- 2 unit tests for the adapter shape (above).
- 3 unit tests for the retryable-codes contract.
- 1 live-adapter integration drive against the actual PR-head production code.
- 2 live regression round-trips proving the guard does not misfire on healthy completions.

### Live-adapter eval transcript

```
======================================================================
PR #12926 LIVE adapter validation
======================================================================

[1] _RETRYABLE_STREAM_ERROR_CODES = ['empty_completion', 'transient_api_error']
    PASS: empty_completion + transient_api_error both retryable

[2] Driving SDKResponseAdapter with empty success ResultMessage
    step_open after init = True
[SDK] [LIVE-validat] Empty-success ResultMessage detected — emitting stream error instead of silent finish
    emitted = ['StreamFinishStep', 'StreamError', 'StreamFinish']
    PASS: code=empty_completion, msg='The model returned an empty response.'
    PASS: ordering FinishStep@0 -> Error@1 -> Finish@2

[3] Negative case — non-empty result must NOT misfire
    emitted = ['StreamFinishStep', 'StreamFinish']
    PASS: no StreamError for non-empty result

======================================================================
ALL LIVE CHECKS PASSED
======================================================================
```

### Notes on screenshots

The local `agent-browser screenshot` binary is hanging (a separate issue unrelated to this PR — stale Chromium session from a sibling worktree could not produce PNG output). The visual regression evidence is therefore captured via accessibility-tree snapshots and DB / API state instead: the chat session API at `/api/chat/sessions/a3b51aa1-5644-4e52-9e61-458bb84ba5f0` returns 4 messages with the expected `user/assistant/user/assistant` roles, and the live UI returned to its idle state (`button "Submit" [disabled]`, "Press Enter to send" hint visible) after each turn — the canonical "stream completed cleanly" UI signature.

For sanity, the `kimi-k2.6` model-routing message visible in the assistant content is an unrelated environmental config issue (same as PR #12927's report) — the relevant signal here is that the SDK itself returned non-empty text (so the new guard correctly *did not* fire), the stream finished, and the assistant message persisted.

PR #12937, opened 2026-04-28
fix(platform): small fixes and stagger animations on settings pages

Abhi1992002:
## Why

The new Settings v2 surfaces (preferences, api-keys, integrations, profile) shipped with a few rough edges spotted in self-review:

- **Timezone saves silently dropped on refresh.** Backend `GET /auth/user/timezone` resolved the user via `get_or_create_user(user_data)` (a 5-min in-process cache keyed by the JWT-payload dict). `update_user_timezone` only invalidates `get_user_by_id`'s cache, so the GET kept returning the pre-save tz until TTL expired — looked exactly like "save did nothing."
- **Confusing "Looks like you're in X" CTA on the Time zone card** that did nothing in the common case (server tz already matched the browser tz, so clicking it produced no dirty state).
- **Save was disabled out of the gate when server tz was `"not-set"`** — the hook substituted the browser tz into both `formState` and `savedState`, so they were equal and `dirty` was false.
- **Lists felt static.** No motion when API keys / integrations mount, and the loading skeletons popped in all at once instead of handing off cleanly to the loaded rows.
- **Profile bio textarea** corner clipped against the rounded-3xl border and the scrollbar overflowed the rounded container.

### Bug fixes
- `GET /auth/user/timezone` now reads via `get_user_by_id(user_id)` — the same cache `update_user_timezone` already invalidates — so a save followed by refresh shows the new tz immediately.
- `usePreferencesPage` now treats the raw server tz (`"not-set"` included) as the saved baseline, while `formState` uses the browser tz only as a *display* fallback. Effect: when the user has never set a tz, Save is enabled on first paint and a single click persists the detected tz.
- Frontend save flow swapped `setQueryData` for `invalidateQueries`, mirroring the older `/profile/(user)/settings` page so we always re-read the persisted value.
- Removed the auto-detect "Looks like you're in X" button + its dead helpers.

### Animations (per Emil Kowalski's guidelines)
Added orchestrated stagger animations that run on both the loaded list **and** its skeleton, so the loading→loaded handoff is continuous in-position:

- **API keys list + skeleton:** 280ms ease-out `cubic-bezier(0.16, 1, 0.3, 1)`, 40ms stagger, opacity + 6px translate.
- **Integrations list + skeleton:** 300ms ease-out, 80ms stagger, opacity + 16px translate (rows are bigger / fewer).
- Both honor `prefers-reduced-motion` via `useReducedMotion`; only `opacity` and `transform` are animated.

### Misc polish
- Profile bio textarea: `!rounded-tr-md` so the top-right corner doesn't fight the surrounding `rounded-3xl`, plus a thin styled scrollbar (`scrollbar-thin scrollbar-thumb-zinc-200 hover:scrollbar-thumb-zinc-300`) that lives inside the rounded container instead of breaking out of it.

## How

| File | Change |
| --- | --- |
| `backend/api/features/v1.py` | `get_user_timezone_route` now uses `get_user_by_id` + `Security(get_user_id)` instead of `get_or_create_user(user_data)` |
| `frontend/.../preferences/usePreferencesPage.ts` | Split init into `initialFormState` (browser-fallback display) vs `initialSavedState` (raw server value); swap optimistic `setQueryData` for `invalidateQueries` after tz mutate |
| `frontend/.../preferences/components/TimezoneCard/TimezoneCard.tsx` | Drop `initialValue` prop, remove auto-detect button + unused imports |
| `frontend/.../preferences/page.tsx` | Drop `savedState`/`initialValue` wiring |
| `frontend/.../api-keys/components/APIKeyList/APIKeyList.tsx` | Wrap rows in container `motion.div` with `staggerChildren`; per-row `motion.div` with opacity + y variants |
| `frontend/.../api-keys/components/APIKeyListSkeleton/APIKeyListSkeleton.tsx` | Same stagger config so loading→loaded matches |
| `frontend/.../integrations/components/IntegrationsList/IntegrationsList.tsx` + `IntegrationsListSkeleton.tsx` | Same pattern for the providers list |
| `frontend/.../profile/components/ProfileForm/ProfileForm.tsx` | Tailwind classes only — `!rounded-tr-md` + `scrollbar-thin scrollbar-thumb-zinc-200 hover:scrollbar-thumb-zinc-300` |

## Test plan

- [ ] On `/settings/preferences`: pick a different tz → Save → hard-refresh → new tz still selected.
- [ ] First-time user (server tz = `not-set`): land on page, Save button should already be enabled; click Save → toast confirms; refresh → tz persists.
- [ ] No "Looks like you're in X" button visible.
- [ ] On `/settings/api-keys`: rows fade/slide in staggered on first mount; loading skeleton uses the same motion.
- [ ] On `/settings/integrations`: provider groups fade/slide in staggered; skeleton matches.
- [ ] OS "Reduce motion" enabled → no transforms, content appears instantly on all four surfaces.
- [ ] On `/settings/profile`: bio textarea top-right corner is no longer hard-cornered against the card; scrollbar fits inside the rounded shape.
- [ ] Existing unit tests still pass: `pnpm test:unit src/app/\(platform\)/settings/preferences` and `.../api-keys`.

### DIFF
--- a/autogpt_platform/backend/backend/api/features/v1.py
+++ b/autogpt_platform/backend/backend/api/features/v1.py
@@ -201,10 +201,10 @@ async def update_user_email_route(
     dependencies=[Security(requires_user)],
 )
 async def get_user_timezone_route(
-    user_id: Annotated[str, Security(get_user_id)],
+    user_data: dict = Security(get_jwt_payload),
 ) -> TimezoneResponse:
     """Get user timezone setting."""
-    user = await get_user_by_id(user_id)
+    user = await get_or_create_user(user_data)
     return TimezoneResponse(timezone=user.timezone)
 
 
--- a/autogpt_platform/backend/backend/api/features/v1.py
+++ b/autogpt_platform/backend/backend/api/features/v1.py
@@ -201,10 +201,10 @@ async def update_user_email_route(
     dependencies=[Security(requires_user)],
 )
 async def get_user_timezone_route(
-    user_data: dict = Security(get_jwt_payload),
+    user_id: Annotated[str, Security(get_user_id)],
 ) -> TimezoneResponse:
     """Get user timezone setting."""
-    user = await get_or_create_user(user_data)
+    user = await get_user_by_id(user_id)
     return TimezoneResponse(timezone=user.timezone)
 
 
--- a/autogpt_platform/frontend/src/app/(platform)/settings/api-keys/components/APIKeyList/APIKeyList.tsx
+++ b/autogpt_platform/frontend/src/app/(platform)/settings/api-keys/components/APIKeyList/APIKeyList.tsx
@@ -77,17 +77,52 @@ export function APIKeyList() {
         )}
       </AnimatePresence>
 
-      <div className="flex flex-col divide-y divide-zinc-200 overflow-hidden rounded-[8px] border border-zinc-200 bg-white">
+      <motion.div
+        className="flex flex-col divide-y divide-zinc-200 overflow-hidden rounded-[8px] border border-zinc-200 bg-white"
+        initial={reduceMotion ? false : "hidden"}
+        animate={reduceMotion ? undefined : "show"}
+        variants={
+          reduceMotion
+            ? undefined
+            : {
+                hidden: {},
+                show: {
+                  transition: {
+                    staggerChildren: 0.04,
+                    delayChildren: 0.04,
+                  },
+                },
+              }
+        }
+      >
         {keys.map((key) => (
-          <APIKeyRow
+          <motion.div
             key={key.id}
-            apiKey={key}
-            selected={selection.isSelected(key.id)}
-            onToggleSelected={() => selection.toggle(key.id)}
-            onDelete={() => requestDelete([key.id])}
-          />
+            variants={
+              reduceMotion
+                ? undefined
+                : {
+                    hidden: { opacity: 0, y: 6 },
+                    show: {
+                      opacity: 1,
+                      y: 0,
+                      transition: {
+                        duration: 0.28,
+                        ease: [0.16, 1, 0.3, 1],
+                      },
+                    },
+                  }
+            }
+          >
+            <APIKeyRow
+              apiKey={key}
+              selected={selection.isSelected(key.id)}
+              onToggleSelected={() => selection.toggle(key.id)}
+              onDelete={() => requestDelete([key.id])}
+            />
+          </motion.div>
         ))}
-      </div>
+      </motion.div>
 
       {deleteTarget && (
         <DeleteAPIKeyDialog
--- a/autogpt_platform/frontend/src/app/(platform)/settings/api-keys/components/APIKeyList/APIKeyList.tsx
+++ b/autogpt_platform/frontend/src/app/(platform)/settings/api-keys/components/APIKeyList/APIKeyList.tsx
@@ -1,6 +1,7 @@
 "use client";
 
 import { AnimatePresence, motion, useReducedMotion } from "framer-motion";
+import type { Variants } from "framer-motion";
 
 import { ErrorCard } from "@/components/molecules/ErrorCard/ErrorCard";
 
@@ -10,6 +11,22 @@ import { APIKeyRow } from "../APIKeyRow/APIKeyRow";
 import { APIKeySelectionBar } from "../APIKeySelectionBar/APIKeySelectionBar";
 import { DeleteAPIKeyDialog } from "../DeleteAPIKeyDialog/DeleteAPIKeyDialog";
 import { useAPIKeyListView } from "./useAPIKeyListView";
+
+const LIST_CONTAINER_VARIANTS: Variants = {
+  hidden: {},
+  show: {
+    transition: { staggerChildren: 0.04, delayChildren: 0.04 },
+  },
+};
+
+const LIST_ITEM_VARIANTS: Variants = {
+  hidden: { opacity: 0, y: 6 },
+  show: {
+    opacity: 1,
+    y: 0,
+    transition: { duration: 0.28, ease: [0.16, 1, 0.3, 1] },
+  },
+};
 
 export function APIKeyList() {
   const {
@@ -81,38 +98,12 @@ export function APIKeyList() {
         className="flex flex-col divide-y divide-zinc-200 overflow-hidden rounded-[8px] border border-zinc-200 bg-white"
         initial={reduceMotion ? false : "hidden"}
         animate={reduceMotion ? undefined : "show"}
-        variants={
-          reduceMotion
-            ? undefined
-            : {
-                hidden: {},
-                show: {
-                  transition: {
-                    staggerChildren: 0.04,
-                    delayChildren: 0.04,
-                  },
-                },
-              }
-        }
+        variants={reduceMotion ? undefined : LIST_CONTAINER_VARIANTS}
       >
         {keys.map((key) => (
           <motion.div
             key={key.id}
-            variants={
-              reduceMotion
-                ? undefined
-                : {
-                    hidden: { opacity: 0, y: 6 },
-                    show: {
-                      opacity: 1,
-                      y: 0,
-                      transition: {
-                        duration: 0.28,
-                        ease: [0.16, 1, 0.3, 1],
-                      },
-                    },
-                  }
-            }
+            variants={reduceMotion ? undefined : LIST_ITEM_VARIANTS}
           >
             <APIKeyRow
               apiKey={key}
--- a/autogpt_platform/frontend/src/app/(platform)/settings/integrations/components/IntegrationsList/IntegrationsList.tsx
+++ b/autogpt_platform/frontend/src/app/(platform)/settings/integrations/components/IntegrationsList/IntegrationsList.tsx
@@ -128,18 +128,39 @@ export function IntegrationsList() {
       {isEmpty ? (
         <IntegrationsListEmpty query={query} />
       ) : (
-        <div className="flex flex-col gap-3 pb-4">
+        <motion.div
+          className="flex flex-col gap-3 pb-4"
+          initial="hidden"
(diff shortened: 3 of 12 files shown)


autogpt_platform/backend/backend/api/features/v1.py
```diff
@@ -201,10 +201,10 @@ async def update_user_email_route(
     dependencies=[Security(requires_user)],
 )
 async def get_user_timezone_route(
-    user_data: dict = Security(get_jwt_payload),
+    user_id: Annotated[str, Security(get_user_id)],
 ) -> TimezoneResponse:
     """Get user timezone setting."""
-    user = await get_or_create_user(user_data)
+    user = await get_user_by_id(user_id)
```
Abhi1992002: 🟠 **Should Fix**: This swap removes the auto-create-on-read behavior that `get_or_create_user(user_data)` provided. `get_user_by_id(user_id)` raises `ValueError` if the Prisma `User` row doesn't exist yet — which is rare in steady state but can happen for a JWT-authenticated user whose DB row hasn't been created (e.g. the GET tz endpoint runs before any other endpoint that touches `get_or_create_user` during a fresh session). Resulting 500 would surface as a generic error toast on the preferences page.

Two options:
1. Keep the original `get_or_create_user(user_data)` call but separately invalidate its cache in `update_user_timezone` (call `get_or_create_user.cache_clear()` or rebuild a per-user-id key). Preserves auto-create + makes save → refresh consistent.
2. Add a fallback: if `get_user_by_id` raises, fall back to `get_or_create_user(jwt_payload)` to ensure the row exists, then retry. More complex.

Option 1 is cleaner. The root cause was cache-key mismatch, not the function choice — so unifying caches is more truthful to the bug.

autogpt_platform/frontend/src/app/(platform)/settings/preferences/usePreferencesPage.ts
```diff
@@ -64,15 +64,18 @@ export function usePreferencesPage() {
...
-  const initialState: PreferencesFormState = {
-    timezone: initialTimezone,
+  const initialFormState: PreferencesFormState = {
+    timezone: formTimezone,
+    notifications: initialFlags,
+  };
+  const initialSavedState: PreferencesFormState = {
+    timezone: serverTimezone,
     notifications: initialFlags,
   };
```
Abhi1992002, line 80: 🟡 **Nice to Have**: This is the new "server has no tz, browser fallback drives display" path — the central UX behavior change of the PR — but `__tests__/main.test.tsx` only mocks `timezone: "Asia/Kolkata"`. Add a test where `getGetV1GetUserTimezoneMockHandler({ timezone: "not-set" })` is used and assert:
- Save button is enabled on first paint
- Clicking Save POSTs the *browser-detected* tz (mock `Intl.DateTimeFormat` if needed, or assert the body matches a non-empty IANA string)
- After save, button disables (dirty=false again)

Without this, a future refactor that re-substitutes browser tz into `savedState` will silently regress the fix.

autogpt_platform/frontend/src/app/(platform)/settings/api-keys/components/APIKeyList/APIKeyList.tsx
```diff
@@ -77,17 +77,52 @@ export function APIKeyList() {
...
+            ? undefined
+            : {
+                hidden: {},
+                show: {
+                  transition: {
+                    staggerChildren: 0.04,
+                    delayChildren: 0.04,
+                  },
+                },
+              }
```
Abhi1992002: 🔵 **Nit**: The container + item variant objects are reconstructed on every render. Functionally fine (Framer Motion compares by deep equality), but moving them to module-level consts makes the intent clearer and eliminates a subtle source of GC pressure on long lists:

```ts
// outside the component
const CONTAINER_VARIANTS = {
  hidden: {},
  show: { transition: { staggerChildren: 0.04, delayChildren: 0.04 } },
};
const ITEM_VARIANTS = {
  hidden: { opacity: 0, y: 6 },
  show: {
    opacity: 1, y: 0,
    transition: { duration: 0.28, ease: [0.16, 1, 0.3, 1] as const },
  },
};
```

Then in the component: `variants={reduceMotion ? undefined : CONTAINER_VARIANTS}`. Same applies to the matching skeleton and the integrations list pair.

autogpt_platform/backend/backend/api/features/v1.py
Abhi1992002: Done in c0ef87747 — went with option 1: GET keeps using `get_or_create_user(user_data)` (preserves auto-create), and `update_user_timezone` now also calls `get_or_create_user.cache_clear()` so the post-save refetch sees the new tz immediately. Comment in `update_user_timezone` explains why `cache_clear` (not `cache_delete`) — the dict-key isn't reconstructable from `user_id` alone.

autogpt_platform/frontend/src/app/(platform)/settings/preferences/usePreferencesPage.ts
Abhi1992002: Done in c0ef87747 — added `Save is enabled on first paint when server timezone is not-set, and saves the detected browser tz` to `__tests__/main.test.tsx`. Asserts (a) Save enabled out of the gate, (b) clicking it POSTs a non-`"not-set"` IANA string. Doesn't pin the exact tz since it depends on the test runner's `Intl` resolution, but the regression we care about — substituting browser tz back into `savedState` and disabling Save — would fail this test.

autogpt_platform/frontend/src/app/(platform)/settings/api-keys/components/APIKeyList/APIKeyList.tsx
Abhi1992002: Done in c0ef87747 — hoisted to module-level `LIST_CONTAINER_VARIANTS` / `LIST_ITEM_VARIANTS` (and `SKELETON_*`/`REDUCED_MOTION_ITEM_VARIANTS` where applicable) in all four components: `APIKeyList`, `APIKeyListSkeleton`, `IntegrationsList`, `IntegrationsListSkeleton`. Component bodies now just pick the right variant by `reduceMotion`.

autogpt_platform/frontend/src/app/(platform)/settings/preferences/usePreferencesPage.ts
```diff
@@ -147,22 +155,9 @@ export function usePreferencesPage() {
...
-              };
-            }
-            return { status: 200, data: { timezone: snapshot.timezone } };
-          },
-        );
+        await queryClient.invalidateQueries({
+          queryKey: getGetV1GetUserTimezoneQueryKey(),
+        });
         setSavedState((prev) => ({ ...prev, timezone: snapshot.timezone }));
         timezoneSaved = true;
```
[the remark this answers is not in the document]
Abhi1992002: Done in ac2f51919 — used the mutation response's persisted tz instead of the refetch (one less round-trip than `refetch()` and avoids racing the invalidate). `updateTimezone.mutateAsync` already returns `{ status: 200, data: { timezone } }`, so:

```ts
const result = await updateTimezone.mutateAsync({ data: { timezone: snapshot.timezone as UpdateTimezoneRequestTimezone } });
const persistedTimezone =
  (result.status === 200 && result.data?.timezone) || snapshot.timezone;
setSavedState((prev) => ({ ...prev, timezone: persistedTimezone }));
```

If the backend ever normalizes the IANA string (e.g. aliases `Asia/Calcutta` → `Asia/Kolkata`), `savedState` now matches what was actually persisted.

autogpt_platform/frontend/src/app/(platform)/settings/integrations/components/IntegrationsList/IntegrationsList.tsx
```diff
@@ -13,6 +14,27 @@ import { ProviderGroup } from "../ProviderGroup/ProviderGroup";
...
+    opacity: 1,
+    y: 0,
+    transition: { duration: 0.3, ease: [0.16, 1, 0.3, 1] },
+  },
+};
+
+const REDUCED_MOTION_ITEM_VARIANTS: Variants = {
+  hidden: { opacity: 0 },
+  show: { opacity: 1 },
+};
```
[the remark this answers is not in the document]
Abhi1992002: Fixed in 680600e0a894086e5e8ebd381b22cbde4c45ed66: `IntegrationsList` + `IntegrationsListSkeleton` now skip the container variants entirely (`initial={reduceMotion ? false : "hidden"}`, `animate={reduceMotion ? undefined : "show"}`, `variants={reduceMotion ? undefined : LIST_CONTAINER_VARIANTS}`) when `useReducedMotion()` is true, so reduced-motion users get a single non-staggered render. Matches the existing pattern in `APIKeyList`.

autogpt_platform/frontend/src/app/(platform)/settings/preferences/__tests__/main.test.tsx
```diff
@@ -264,6 +264,58 @@ describe("SettingsPreferencesPage", () => {
...
+    await waitFor(() => {
+      expect(submittedTimezone).toBeDefined();
+    });
+    expect(submittedTimezone).not.toBe("not-set");
+    expect(submittedTimezone?.length ?? 0).toBeGreaterThan(0);
+
+    await waitFor(() => {
+      expect((saveButton as HTMLButtonElement).disabled).toBe(true);
+    });
+  });
```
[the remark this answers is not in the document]
Abhi1992002: Fixed in 680600e0a894086e5e8ebd381b22cbde4c45ed66: the test now stubs `Intl.DateTimeFormat.prototype.resolvedOptions` to return a fixed `America/New_York` and asserts `submittedTimezone === STUBBED_BROWSER_TZ` exactly (replacing the broad non-empty / non-`not-set` checks). Spy is restored in a `finally` so it doesn't leak into other tests.

0ubbe (approved):

PR #12939, opened 2026-04-28
Security: add permission manager to Agent Protocol Server

johnpippett:
## Summary

The Agent Protocol Server previously created agents without a `CommandPermissionManager`. Those agents skipped command permission checks.

This pull request:

- creates a noninteractive manager for each new or restored task;
- applies workspace and agent permission files before command execution;
- checks `ask_user` and child-agent `finish` actions against those files;
- gives the same manager to child agents from multi-step strategies;
- uses `folder` for `list_folder` permission checks;
- allows root-folder listings with the default policy;
- lets `finish` reasons contain `/`;
- migrates generated legacy allow lists that contain the old defaults; and
- marks a step final only after `finish` executes.

The manager denies each command that no allow rule matches. Server mode does not ask the operator for approval.

## Validation

- Forge tests: 736 passed, 17 skipped, and 14 expected failures.
- Original AutoGPT tests: 49 passed and 2 skipped.
- isort, Black, and Flake8: passed.
- Pyright: 0 errors and 2 warnings.

### DIFF
--- a/classic/original_autogpt/autogpt/app/agent_protocol_server.py
+++ b/classic/original_autogpt/autogpt/app/agent_protocol_server.py
@@ -31,9 +31,11 @@ from forge.agent_protocol.models import (
     TaskRequestBody,
     TaskStepsListResponse,
 )
+from forge.config.workspace_settings import AgentPermissions, WorkspaceSettings
 from forge.file_storage import FileStorage
 from forge.llm.providers import ModelProviderBudget, MultiProvider
 from forge.models.action import ActionErrorResult, ActionSuccessResult
+from forge.permissions import CommandPermissionManager
 from forge.utils.const import ASK_COMMAND, FINISH_COMMAND
 from forge.utils.exceptions import AgentFinished, NotFoundError
 
@@ -144,6 +146,9 @@ class AgentProtocolServer:
             app_config=self.app_config,
             file_storage=self.file_storage,
             llm_provider=self._get_task_llm_provider(task),
+            permission_manager=self._get_permission_manager(
+                task_agent_id(task.task_id)
+            ),
         )
         await task_agent.file_manager.save_state()
 
@@ -188,6 +193,7 @@ class AgentProtocolServer:
             app_config=self.app_config,
             file_storage=self.file_storage,
             llm_provider=self._get_task_llm_provider(task),
+            permission_manager=self._get_permission_manager(task_agent_id(task_id)),
         )
 
         if user_id := (task.additional_input or {}).get("user_id"):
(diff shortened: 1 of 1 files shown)


Santoshkumarpuppala:
`test_format_args_list_folder` (`tests/test_permissions.py:82-88`) passes `{"path": "/tmp"}`, a key the command's schema doesn't declare. `list_folder` takes `folder` (`file_manager.py:201,208`), so the test can't fail on `permissions.py:190`, which reads only `filename` or `path`.

That leaves the formatted args `""`, and `list_folder({workspace}/**)` (`workspace_settings.py:26`) never matches. With `prompt_fn=None` (`agent_protocol_server.py:458-475`) that falls to DENY at `permissions.py:152-153`, so with the default allow list this PR makes server agents deny `list_folder`. Same shape as the `finish(*)` note above.

The other direction: an operator-added `list_folder(*)`, not in the defaults, matches `""` because `*` becomes `[^/]*` (`permissions.py:268-269`), so it allows `{'folder': '/etc'}`. The permission check over-matches, but the sandbox holds (`restrict_to_root=True`, `file_storage/base.py:26,235`).

I didn't run the server. I called `check_command` on dev 22a7251 (forge/ is byte-identical to c9faa48): `{'folder': 'sub'}` DENY, `{'path': 'sub'}` ALLOW as the control. With `_format_args` reading `folder`, `sub` is ALLOW and `/etc` DENY.

The bug is older than this PR and sits in `permissions.py`, which the PR doesn't touch. This PR makes that deny the default for server agents. Happy to open a separate issue, or add the one-line key fix plus a `folder` test here.

johnpippett:
Thank you for the detailed `list_folder` trace. I changed the formatter to read the command's `folder` argument.

Tests now cover the actual `folder` argument and the workspace root. They also cover an out-of-workspace folder with a conflicting `path` value.

I also changed the default rule to `finish(**)`. Generated legacy policies now migrate to the corrected defaults.

I added tests for server actions and child-agent permission checks. A step now becomes final only after `finish` executes.

Local validation results:

- Forge tests: 736 passed, 17 skipped, and 14 expected failures.
- Original AutoGPT tests: 49 passed and 2 skipped.
- isort, Black, and Flake8: passed.
- Pyright: 0 errors and 2 warnings.

GitHub now reports passing builds, lint, types, CodeQL, Snyk, and CLA checks.

Three test jobs stopped before repository tests because `minio/minio:edge-cicd` is unavailable. The benchmark cannot run because the fork does not receive the Anthropic secret.

CodeRabbit and repository status checks are still pending. Vercel also requires project authorization.

PR #12933, opened 2026-04-28
feat(platform): subscription credit grants + paywall gate + dialog UX + cross-pod cache

majdyz:
## Why

Started as a regression fix for admin-granted user downgrades hitting Stripe Checkout, broadened to close the surrounding gaps in the Stripe billing flow that surfaced during testing. Three concrete user-facing problems the PR resolves:

1. **Admin-granted users couldn't change tier in-app** when their current tier had no `stripe-price-id-*` LD configured — clicking Downgrade silently routed to a paid-signup Stripe Checkout instead of just changing the tier.
2. **Subscription payments granted nothing visible to users** — paying £20–£320/mo gave higher rate-limit multipliers but no AutoPilot credits in the user's balance, despite a dialog promising "credit to your next Stripe invoice" (which users naturally read as AutoGPT credits).
3. **Tier oscillated across page refreshes** — `get_user_by_id` was process-local cached, so dev's 4 server pods each held their own copy. Tier could read MAX on one pod and BASIC on another for ~5 min after a webhook update, depending on which pod the request landed on.

Plus three structural improvements caught during review:

4. **No paywall enforcement for paid-cohort users without subscription** — non-beta users on `BASIC` (no Stripe sub) could freely use AutoPilot.
5. **Upgrade/downgrade dialog copy was misleading** — implied a Stripe redirect that doesn't happen for existing-sub modifications, used "credit" ambiguously, and didn't surface the next-invoice date.
6. **Top-up Checkout created an ephemeral Stripe Product per session** — no canonical Product for dashboard reporting, no way to scope coupons to top-ups.

### 1. Admin-granted downgrades skip Checkout (price-id-pruning regression)

`update_subscription_tier()` used to gate its modify-or-DB-flip block on `current_tier_price_id is not None`. When a tier was pruned from `stripe-price-ids` LD, that gate skipped the inner DB-flip branch and the request fell through to Checkout — sending admin-granted users to a paid-signup flow when they were trying to *reduce* their tier. Drop the gate and call `modify_stripe_subscription_for_tier()` unconditionally — the function self-reports `False` when there's no Stripe sub. One uniform path for everyone now.

### 2. Subscription credit grant on every paid Stripe invoice

[...]

PR #12947, opened 2026-04-30
fix(platform): cancel-banner copy + clearer 422 on currency mismatch

majdyz:
## Why

Two regressions surfaced after [#12933](https://github.com/Significant-Gravitas/AutoGPT/pull/12933) merged to `dev`:

1. **Cancel-pending banner shows wrong copy.** The merged PR moved cancel-at-period-end from `BASIC` → `NO_TIER`, but `PendingChangeBanner.isCancellation` was still keyed on `"BASIC"`. As a result, a user who cancels their sub now sees *"Scheduled to downgrade to No subscription on …"* instead of the intended *"Scheduled to cancel your subscription on …"*. Caught by Sentry on the merged PR.

2. **Currency-mismatch downgrade returns 502 (looks like outage).** A user with an existing GBP-active sub (Max Price has `currency_options.gbp`) tried to downgrade to Pro and got 502. The backend logs show:
   ```
   stripe._error.InvalidRequestError: The price specified only supports `usd`.
   This doesn't match the expected currency: `gbp`.
   ```
   The Pro Price is USD-only; Stripe rejects `SubscriptionSchedule.modify` because phases must share currency. Wrapping that in a generic 502 hid the real cause and made it read like a Stripe outage.

## What

* Frontend: flip `PendingChangeBanner.isCancellation` from `pendingTier === "BASIC"` to `"NO_TIER"`. Update both component and page-level tests that exercised the cancellation branch.
* Backend: catch `stripe.InvalidRequestError` whose message mentions `currency` in `update_subscription_tier`, and return **422** with *"Tier change unavailable for your current billing currency. Cancel your subscription and re-subscribe at the target tier, or contact support."* — so users see the actual reason, not a misleading outage message. Other `StripeError` paths still return 502.
* New backend test asserts the currency-mismatch branch returns 422 with the new copy.

## How

* `PendingChangeBanner.tsx` line 28: 1-char change (`"BASIC"` → `"NO_TIER"`).
* `subscription_routes_test.py` and `PendingChangeBanner.test.tsx` updated to use `NO_TIER` for the cancellation fixture.
* `v1.py` `update_subscription_tier` adds a typed `except stripe.InvalidRequestError` branch ahead of the generic `StripeError`; only currency-mismatch messages get the special 422, everything else falls through to the existing 502.

## The real fix lives in Stripe config

The defensive 422 here is just a clearer error surface. To actually unblock GBP/EUR users from changing tiers, the per-tier Stripe Prices (Pro, and Basic if priced) need `currency_options` for GBP added — Max already has this, which is why Max checkout shows the £/$ toggle. Stripe locks `currency_options` after a Price has been transacted, so the procedure is: create a new Price with USD + GBP from the start → update the `stripe-price-ids` LD flag to the new Price ID. No further code change required; same Price ID stays per tier, multiple currencies inside it.

## Checklist

- [x] Component test for new banner copy
- [x] Backend test for 422 currency-mismatch branch
- [x] Format / lint / types pass
- [x] No protected route added — N/A

### DIFF
--- a/autogpt_platform/backend/backend/api/features/subscription_routes_test.py
+++ b/autogpt_platform/backend/backend/api/features/subscription_routes_test.py
@@ -464,6 +464,50 @@ def test_update_subscription_tier_currency_mismatch_returns_422(
     assert "contact support" in detail.lower()
 
 
+def test_update_subscription_tier_non_currency_invalid_request_returns_502(
+    client: fastapi.testclient.TestClient,
+    mocker: pytest_mock.MockFixture,
+) -> None:
+    """Locks the contract that *only* currency-mismatch InvalidRequestErrors
+    translate to 422 — every other Stripe InvalidRequestError must still
+    surface as the generic 502 so that widening the conditional later is
+    caught by the suite."""
+    mock_user = Mock()
+    mock_user.subscription_tier = SubscriptionTier.MAX
+
+    async def mock_feature_enabled(*args, **kwargs):
+        return True
+
+    mocker.patch(
+        "backend.api.features.v1.get_user_by_id",
+        new_callable=AsyncMock,
+        return_value=mock_user,
+    )
+    mocker.patch(
+        "backend.api.features.v1.is_feature_enabled",
+        side_effect=mock_feature_enabled,
+    )
+    mocker.patch(
+        "backend.api.features.v1.modify_stripe_subscription_for_tier",
+        side_effect=stripe.InvalidRequestError(
+            "No such price: 'price_does_not_exist'",
+            param="items[0][price]",
+        ),
+    )
+
+    response = client.post(
+        "/credits/subscription",
+        json={
+            "tier": "PRO",
+            "success_url": f"{TEST_FRONTEND_ORIGIN}/success",
+            "cancel_url": f"{TEST_FRONTEND_ORIGIN}/cancel",
+        },
+    )
+
+    assert response.status_code == 502
+    assert "billing currency" not in response.json()["detail"].lower()
+
+
 def test_update_subscription_tier_creates_checkout(
     client: fastapi.testclient.TestClient,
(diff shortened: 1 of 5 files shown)


autogpt_platform/backend/backend/api/features/subscription_routes_test.py
```diff
@@ -416,6 +416,54 @@ async def mock_feature_enabled(*args, **kwargs):
...
+            "tier": "PRO",
+            "success_url": f"{TEST_FRONTEND_ORIGIN}/success",
+            "cancel_url": f"{TEST_FRONTEND_ORIGIN}/cancel",
+        },
+    )
+
+    assert response.status_code == 422
+    detail = response.json()["detail"]
+    assert "billing currency" in detail.lower()
+    assert "contact support" in detail.lower()
```
majdyz, line 464: 🤖 🟡 **Nice to Have**: Only the positive case (`currency` in error message → 422) is covered. The conditional `if "currency" in msg.lower()` could regress silently if the branch widens to catch all `InvalidRequestError`. Worth adding a sibling test where the mocked `InvalidRequestError` carries an unrelated message (e.g. `'No such price: price_xxx'`) and asserts the response is 502, locking the contract that *only* currency-mismatch errors translate to 422.

majdyz: 🤖 Fixed in [4676bc89a](https://github.com/Significant-Gravitas/AutoGPT/commit/4676bc89a9575e12664db0ac8593cf923368fd8c): added `test_update_subscription_tier_non_currency_invalid_request_returns_502` — mocks an unrelated `InvalidRequestError` ("No such price: ...") and asserts the response stays at the generic 502 with no "billing currency" in the body. Locks the if-condition so widening it to catch all `InvalidRequestError` would break the suite.

PR #12961, opened 2026-04-30
feat(backend): use sortable UUIDv7 for ID defaults

majdyz:
### Why / What / How

**Why** — random UUIDv4 primary keys insert at random positions in B-tree indexes, which fragments WAL/IO patterns on busy tables (executions, library, audit) and forces every "latest N" query to sort by `createdAt` instead of `id`. UUIDv7 puts a millisecond timestamp in the high 48 bits so values are roughly insert-ordered.

**What** — flips every Prisma id default from random `uuid()` (and the one `gen_random_uuid()` site on `AnalyticsDetails`) to a Postgres `uuid_generate_v7()` function. Adds a Python helper for code that mints ids before insert.

**How** —

1. New SQL function `uuid_generate_v7()` in a single migration. RFC 9562 layout, built on pgcrypto's `gen_random_uuid()` (already in use), millisecond timestamp from `clock_timestamp()`. ALTERs every id column's DEFAULT to call it.
2. `schema.prisma` switches all `@default(uuid())` (and the `AnalyticsDetails` `gen_random_uuid()`) to `@default(dbgenerated("uuid_generate_v7()"))`.
3. `backend/util/ids.py` exposes `new_uuid()` returning `uuid_utils.uuid7()` as a string. `uuid_utils` is promoted from a transitive dep (via langsmith) to a direct dep so the API stays stable.
4. The three call sites that mint ids in Python before insert switch to `new_uuid()`: `data/graph.py` (`reassign_ids`, `_reassign_ids`, link `create_many`), `data/auth/api_key.py`, `data/auth/oauth.py` (authorization code, access token, refresh token).

Existing UUIDv4 rows are untouched — they remain valid UUIDs and continue to sort lexicographically as before. Only newly generated ids are sortable.

### Changes

- `schema.prisma`: ~40 `@default` flips to `dbgenerated("uuid_generate_v7()")`
- `migrations/20260430120000_use_uuidv7_for_id_defaults/migration.sql`: defines `uuid_generate_v7()` and `ALTER TABLE … SET DEFAULT` for every id column (and `CreditTransaction.transactionKey`)
- `backend/util/ids.py`: new `new_uuid()` helper
- `backend/data/graph.py`, `backend/data/auth/api_key.py`, `backend/data/auth/oauth.py`: route explicit id generation through `new_uuid()`; drop unused `import uuid`
- `pyproject.toml`: `uuid-utils = "^0.14.1"` as a direct dep

#### For code changes:
- [x] I have made a test plan
  - [ ] `poetry run prisma migrate dev` applies the new migration cleanly on a fresh db
  - [ ] `SELECT uuid_generate_v7();` returns a valid v7 (version nibble = 7, variant = 10xx)
  - [ ] Inserting via Prisma without an explicit id populates a v7 from the db default (e.g. create an `AgentGraph`, `LibraryAgent`, `OAuthAccessToken`)
  - [ ] Inserting via the Python helper paths (`reassign_ids`, `generate_api_key`, OAuth code/token creation) writes v7 ids
  - [ ] /pr-test golden path still green


autogpt_platform/backend/pyproject.toml
```diff
@@ -76,6 +76,7 @@ supabase = "2.28.0"
 tenacity = "^9.1.4"
 todoist-api-python = "^2.1.7"
 tweepy = "^4.16.0"
+uuid-utils = "^0.14.1"
```
majdyz: 🤖 🟠 **Should Fix**: `^0.14.1` resolves to `>=0.14.1,<0.15.0` under Poetry's caret rules — much tighter than langsmith's own `>=0.12.0,<1.0` constraint. If a future bump to langsmith pulls in `uuid-utils >=0.15`, dependency resolution will fail until someone hand-edits this line. Consider widening to `>=0.14.1,<1.0.0` (or just `^0.14`) so the direct dep mirrors langsmith's bound and we don't get blocked on transitive upgrades.

autogpt_platform/backend/migrations/20260430120000_use_uuidv7_for_id_defaults/migration.sql
```diff
@@ -0,0 +1,67 @@
...
+        ),
+        52, 1
+      ),
+      53, 1
+    ),
+    'hex'
+  )::uuid;
+END
+$$
+LANGUAGE plpgsql
```
majdyz, line 29: 🤖 🟡 **Nice to Have**: Mark the function `PARALLEL SAFE` (alongside `VOLATILE`). It only reads `clock_timestamp()` and the random portion of `gen_random_uuid()`, both of which are parallel-safe. Without the marker, query planners refuse parallel plans for `INSERT … RETURNING id` paths in a parallel-safe context.

```sql
LANGUAGE plpgsql
VOLATILE
PARALLEL SAFE;
```

autogpt_platform/backend/backend/util/ids.py
```diff
@@ -0,0 +1,14 @@
...
+
+def new_uuid() -> str:
+    """Return a fresh sortable UUIDv7 as a lower-case canonical string.
+
+    Use this for any application-level ID generation that lands in a Prisma
+    column. Schema-level ``@default(dbgenerated("uuid_generate_v7()"))``
+    covers the path where Prisma populates the id; this helper is the
+    Python-side counterpart for code that needs the id before insert.
+    """
+    return str(uuid7())
```
majdyz, line 15: 🤖 🟡 **Nice to Have**: Worth noting in the docstring that with the new `dbgenerated("uuid_generate_v7()")` defaults, callers can usually drop the `id` key entirely from create payloads and let Postgres mint it — `new_uuid()` is mainly for code paths that need the id **before** the insert (FK construction, return-then-insert, dry-run tests). Keeps future code from over-using the helper.

autogpt_platform/backend/schema.prisma
```diff
@@ -128,7 +128,7 @@ enum OnboardingStep {
 }
 
 model UserOnboarding {
-  id        String    @id @default(uuid())
+  id        String    @id @default(dbgenerated("uuid_generate_v7()"))
```
majdyz, line 135: 🤖 🟠 **Should Fix**: Future schemas risk silently regressing to v4 — `prisma migrate dev` will accept `@default(uuid())` again with no warning, and the column will get a random v4 default. Worth adding a one-liner comment near the top of `schema.prisma` (or in `AGENTS.md`) telling future contributors to use `@default(dbgenerated("uuid_generate_v7()"))` for new id columns so the v7 default doesn't drift.

autogpt_platform/backend/pyproject.toml
majdyz: 🤖 Fixed in [251fde86b](https://github.com/Significant-Gravitas/AutoGPT/commit/251fde86bd64f6de190b02ec4fccce1dc46120a1): widened to `>=0.14.1,<1.0.0` so the direct constraint mirrors langsmith's bound and won't block transitive upgrades through the 0.x range.

autogpt_platform/backend/migrations/20260430120000_use_uuidv7_for_id_defaults/migration.sql
majdyz: 🤖 Fixed in [251fde86b](https://github.com/Significant-Gravitas/AutoGPT/commit/251fde86bd64f6de190b02ec4fccce1dc46120a1): added `PARALLEL SAFE` to the function definition.

autogpt_platform/backend/backend/util/ids.py
majdyz: 🤖 Fixed in [251fde86b](https://github.com/Significant-Gravitas/AutoGPT/commit/251fde86bd64f6de190b02ec4fccce1dc46120a1): docstring now points callers at the schema default first and reserves `new_uuid()` for the "id needed before insert" cases (FK construction, return-then-insert, dry-run tests).

autogpt_platform/backend/schema.prisma
majdyz: 🤖 Fixed in [251fde86b](https://github.com/Significant-Gravitas/AutoGPT/commit/251fde86bd64f6de190b02ec4fccce1dc46120a1): added a 4-line directive in the generator block of `schema.prisma` telling future contributors to use `@default(dbgenerated("uuid_generate_v7()"))` on new id columns.

majdyz:
🤖 @coderabbitai fixed in [ae83c3a8f](https://github.com/Significant-Gravitas/AutoGPT/commit/ae83c3a8f1e0baeb2f94946a3baee25f82e1bc49): `_reassign_ids()` now regenerates `link.id` during fork, and `__create_graph` uses `link.id` instead of minting a separate one — keeps the returned graph and the persisted rows in sync. Also bumped `BaseDbModel.id` default_factory to `new_uuid()` so all in-memory model ids are v7 too.

majdyz:
## 🧪 E2E Test Report (local, native mode)


- Date: 2026-04-30
- Branch: `feat/sortable-uuid-v7-defaults` @ ae83c3a8f
- Worktree: `/Users/majdyz/Code/AutoGPT15`
- Mode: native (poetry / shared infra) — backend port-shared with sibling AutoGPT13 stack, so the e2e UI run was substituted by direct DB + import smoke + sibling-stack regression check.

## Environment
- Postgres (supabase-db): `healthy`, schema `platform`
- Redis cluster: `healthy` (3 nodes)
- RabbitMQ: `healthy`
- Clamav, supabase-auth, supabase-kong: `healthy`
- Sibling backend (AutoGPT13) was already bound on :8001/:8002/:8005/:8006/:8008. Shared the same DB after migration; logs show no errors post-apply (only pre-existing Pydantic deprecation warnings).

## Scenarios

### A — Migration smoke
| # | Step | Result |
|---|------|--------|
| A.1 | `poetry run prisma migrate deploy` | **PASS** — applied `20260430120000_use_uuidv7_for_id_defaults` cleanly on top of `dev` (3 prior migrations also applied without error). |
| A.2 | `SELECT uuid_generate_v7();` returns version=`7`, variant=`8/9/a/b` | **PASS** — sample: `019ddef1-c344-7af1-9870-a7537807f65c` (version nibble `7`, variant `a` ∈ {8,9,a,b}). |
| A.3 | 3 consecutive calls return strictly increasing strings (k-sortable) | **PASS** — `[c344-7af1-…, c349-749b-…, c34a-736c-…]` is already sorted. |
| A.4 | `pg_proc` shows `proparallel='s'` (PARALLEL SAFE), `provolatile='v'` | **PASS** — function is correctly marked PARALLEL SAFE. |
| A.5 | Total columns now defaulting to `uuid_generate_v7()` | **PASS** — `40` (matches all 39 id columns + `CreditTransaction.transactionKey`). |

### B — Schema-default insert path
| # | Step | Result |
|---|------|--------|
| B.1 | `INSERT INTO "BuilderSearchHistory" (...)` without an explicit `id`, `RETURNING id` | **PASS** — id `019ddef3-1cb8-717b-94f0-bdb2906be338`, version=`7`. Confirms Postgres default now mints v7. |

### C — Application-helper / pydantic-default path
| # | Step | Result |
|---|------|--------|
| C.1 | `from backend.util.ids import new_uuid; new_uuid()` | **PASS** — version=`7`. |
| C.2 | `BaseDbModel().id` (default_factory) | **PASS** — version=`7`. |
| C.3 | `Link(...)` (extends BaseDbModel) | **PASS** — version=`7`. |
| C.4 | `Node(...)` and `Graph(...)` likewise | **PASS** — both v7. |

### D — Existing flow regression
| # | Step | Result |
|---|------|--------|
| D.1 | Pre-PR rows still load (sample `AgentGraph`, `LibraryAgent`, `AgentNodeLink` rows) | **PASS** — existing UUIDv4 ids load fine; mixed v4/v7 cohabitation in the same column is supported. |
| D.2 | Sibling backend (AutoGPT13) running against the migrated DB stays healthy | **PASS** — `/docs` 200 on rest/websocket, error grep on `.ign.application.logs` returns only Pydantic deprecation warnings (pre-existing). |
| D.3 | Full UI signup → build → run flow | **SKIPPED** — sibling stack held the app ports. The pieces D.1+D.2 plus A/B/C cover the runtime surface this PR actually touches; killing the sibling stack to re-run the same code path was not justified. |

## Bugs found
None.

## Notes for the reviewer
- The PR is purely "shape of newly minted ids changes from v4 → v7". Existing rows are untouched and continue to sort lexicographically as before.
- coderabbit's late finding (link.id stale-vs-DB) was addressed in [ae83c3a8f](https://github.com/Significant-Gravitas/AutoGPT/commit/ae83c3a8f1e0baeb2f94946a3baee25f82e1bc49) — `_reassign_ids()` now also reassigns `link.id`, and `__create_graph` uses `link.id` directly. `BaseDbModel.id` default_factory was also bumped to `new_uuid()` so in-memory model ids are v7 too.
- Function is marked `PARALLEL SAFE`, which lets query planners use parallel plans through `INSERT … RETURNING id` paths.

## Summary
- 11 of 11 in-scope scenarios PASS
- 1 scenario skipped (full UI flow) due to port-share with sibling agent's stack; covered indirectly by D.2
- 0 bugs found

majdyz:
🤖 🟠 **Should Fix** — `IntegrationWebhook.id` minted as v4 in untouched code path

`autogpt_platform/backend/backend/integrations/webhooks/_base.py:207`:

```python
id = str(uuid4())
…
ingress_url = webhook_ingress_url(provider_name=provider_name, webhook_id=id)
…
return await integrations.create_webhook(
    integrations.Webhook(id=id, …)
)
```

This is a pre-insert id mint for an `IntegrationWebhook` row whose schema default this PR moves to `uuid_generate_v7()` (schema.prisma:750, migration line 32). Every other pre-insert site got switched to `new_uuid()` (`api_key.py:86`, `oauth.py:362/494/611`, `graph.py:649/650/670`) — this one was missed.

**Failure scenario**
1. Webhook registered via this base class -> v4 id (random insert position in B-tree)
2. Webhook registered via a future code path that drops `id` and uses the schema default -> v7
3. `IntegrationWebhook` ends up with mixed v4/v7 ids forever, defeating the sortable-by-id property for this table that the rest of the PR establishes

**Fix:** `from backend.util.ids import new_uuid`; `id = new_uuid()`; remove `from uuid import uuid4` if it becomes unused.

majdyz:
🤖 Fixed in [183656aa2](https://github.com/Significant-Gravitas/AutoGPT/commit/183656aa24): `_base.py` now imports `new_uuid` and uses it for the pre-insert webhook id. Drops the now-unused `from uuid import uuid4`.

majdyz:
🤖 🟠 **Should Fix** — three more pre-insert v4 mints survive the v7 sweep

Three call sites still hand-mint UUIDv4 ids that get persisted as primary keys, defeating sortable-by-id for the affected rows. Same pattern as the `IntegrationWebhook` fix in [183656aa2](https://github.com/Significant-Gravitas/AutoGPT/commit/183656aa24eecc805cfe5a0d1ad97db0a5215a92):

1. **`autogpt_platform/backend/backend/data/model.py:328`** — `_BaseCredentials.id`
   ```python
   ```
   Base class for `OAuth2Credentials`, `APIKeyCredentials`, `UserPasswordCredentials`, `HostScopedCredentials`. Every user credential row mints its id here before insert (`credentials_store.add_creds` lookups by `c.id == credentials.id` confirm it round-trips). Should be `default_factory=new_uuid`.

2. **`autogpt_platform/backend/backend/cli/oauth_tool.py:112`** — `generate_app_credentials.app_id`
   ```python
   app_id = str(uuid.uuid4())
   ```
   Returned as the `id` of an `OAuthApplication` row whose schema now defaults to `uuid_generate_v7()` — so CLI-created apps will get v4 while UI-created ones get v7, splitting the table's id distribution.

3. **`autogpt_platform/backend/backend/executor/scheduler.py:715`** — `schedule_id`
   ```python
   schedule_id = str(uuid.uuid4())
   ```
   Becomes the APScheduler job id persisted to the `apscheduler_jobs` PG jobstore and used as the primary lookup key in `delete_graph_execution_schedule` / `get_graph_execution_schedules`. v7 here gets the same B-tree locality benefit as the rest of the sweep.

All three should switch to `from backend.util.ids import new_uuid` and call `new_uuid()`.

majdyz:
🤖 Fixed in [2eec18e1d](https://github.com/Significant-Gravitas/AutoGPT/commit/2eec18e1d7fc62bd1015d0cee02cd6ce4d6f94b1): all three call sites (`_BaseCredentials.id`, `oauth_tool.generate_app_credentials.app_id`, `Scheduler.schedule_id`) now route through `new_uuid()`, dropping the local `uuid`/`uuid4` imports where they were the only consumer.

majdyz:
🤖 🟠 **Should Fix** — three more v4 mints survive the v7 sweep on warm code paths

Round-2 self-review caught three remaining places where ids end up persisted as v4 despite the column default being `uuid_generate_v7()`:

1. `backend/copilot/model.py:239` — `ChatSession.new()` minted `session_id` via `str(uuid.uuid4())`. The value is then passed into `create_chat_session(session_id=...)` and stored as `ChatSession.id` — so every new chat session got a v4 PK.
2. `backend/data/credit.py:575` — raw `INSERT … (transactionKey)` had `COALESCE($9, gen_random_uuid()::text)` as fallback. The column default is now v7, but raw SQL bypasses defaults when the column is in the INSERT list, so the fallback path emitted v4 keys.
3. `backend/api/features/store/embeddings.py:157` — raw upsert into `UnifiedContentEmbedding` hard-coded `gen_random_uuid()::text` in VALUES, bypassing the v7 column default.

Fixed in [4201d0f4c](https://github.com/Significant-Gravitas/AutoGPT/commit/4201d0f4ca): `ChatSession.new()` now uses `new_uuid()`; the two raw-SQL sites now use `uuid_generate_v7()::text`.

PR #12978, opened 2026-05-01
feat(frontend): connect onboarding SubscriptionStep to Stripe Checkout

0ubbe:
### Why / What / How

**Why.** The onboarding `SubscriptionStep` shows Pro / Max / Team plan cards but selecting Pro or Max only sets `selectedPlan` in the wizard store and advances to the Preparing step — it never starts a Stripe Checkout, so brand-new users finish onboarding with `tier === "NO_TIER"` and immediately hit the `PaywallGate` modal on `/copilot`. We already have the wiring for this in `settings/billing/.../YourPlanCard` (uses `useUpdateSubscriptionTier` → redirect to the returned Stripe URL); the onboarding step just wasn't using it.

**What.** Plan selection in the onboarding step now POSTs `/api/credits/subscription` with `tier: "PRO" | "MAX"` plus onboarding-specific `success_url` / `cancel_url`, and redirects the user to the returned Stripe Checkout URL. Team is unchanged (still opens the Tally intake form).

**How.**
- `useSubscriptionStep` calls the generated `useUpdateSubscriptionTier` hook. Success → `window.location.href = url`; no-URL response → fall back to `nextStep()` (matches the in-place modify path the same backend route can take). Errors surface via `toast`.
- The wizard profile is submitted **before** the Stripe redirect — the zustand store is in-memory and doesn't survive the round-trip. On return with `?subscription=success`, the Preparing step's profile-submit effect short-circuits so we don't overwrite the saved data with empty fields.
- `success_url` → `/onboarding?step=5&subscription=success` (Preparing step), `cancel_url` → `/onboarding?step=4&subscription=cancelled` (back to plan picker).
- `PlanCard` now accepts `loading` / `disabled` so the clicked card spins and the others lock while the mutation is in flight.

### Changes 🏗️

- `useSubscriptionStep.ts` — wires `useUpdateSubscriptionTier` + pre-redirect profile submit; exposes `isUpdatingTier` and `selectedPlan`.
- `SubscriptionStep.tsx` — passes per-card loading/disabled state to `PlanCard`.
- `PlanCard.tsx` — accepts `loading` / `disabled` props.
- `useOnboardingPage.ts` — skips the profile resubmit on `?subscription=success` (already submitted before redirect).
- `SubscriptionStep.test.tsx` — replaces the "advances to step 5" assertion with MSW-based assertions on the `/api/credits/subscription` and `/api/onboarding/profile` POST bodies; covers PRO and MAX.

#### For code changes:
- [x] I have made a test plan
  - [ ] As a NO_TIER user, start onboarding, fill in name / role / pain points, click **Get Pro** → land on Stripe Checkout with the Pro price.
  - [ ] Complete the Stripe payment → return to `/onboarding?step=5&subscription=success` → Preparing step runs → land on `/copilot` with no PaywallGate modal.
  - [ ] On Stripe, click **Cancel** → return to `/onboarding?step=4&subscription=cancelled`, plan cards visible again, can pick a different plan.
  - [ ] Click **Upgrade to Max** → Stripe Checkout shows the Max price, same return flow.
  - [ ] Click **Contact sales** (Team) → Tally intake form opens in a new tab; wizard stays on step 4.
  - [ ] While the mutation is in flight, the clicked card shows a spinner and the other cards are disabled.
  - [ ] Verify the onboarding profile is saved (name / role / pain points) even when the user pays via Stripe.

### DIFF
--- a/autogpt_platform/frontend/src/app/(no-navbar)/onboarding/__tests__/page.test.tsx
+++ b/autogpt_platform/frontend/src/app/(no-navbar)/onboarding/__tests__/page.test.tsx
@@ -154,4 +154,25 @@ describe("OnboardingPage — flag-gated SubscriptionStep", () => {
     render(<OnboardingPage />);
     expect(await screen.findByTestId("step-welcome")).toBeDefined();
   });
+
+  it("lets ?step=5&subscription=success bypass the highestStep ceiling", async () => {
+    // Simulates returning from a successful Stripe checkout. The highest
+    // reached step before the redirect was 4 (SubscriptionStep). Without the
+    // success-bypass, ceiling=min(4,5)=4 would clamp the user back to step 4.
+    mockFlagValue = true;
+    window.sessionStorage.setItem(STEP_STORAGE_KEY, "4");
+    currentSearchParams = new URLSearchParams("step=5&subscription=success");
+    render(<OnboardingPage />);
+    expect(await screen.findByTestId("step-preparing")).toBeDefined();
+    expect(screen.queryByTestId("step-subscription")).toBeNull();
+  });
+
+  it("returns to SubscriptionStep on ?step=4&subscription=cancelled", async () => {
+    mockFlagValue = true;
+    window.sessionStorage.setItem(STEP_STORAGE_KEY, "4");
+    currentSearchParams = new URLSearchParams("step=4&subscription=cancelled");
+    render(<OnboardingPage />);
+    expect(await screen.findByTestId("step-subscription")).toBeDefined();
+    expect(useOnboardingWizardStore.getState().selectedPlan).toBeNull();
+  });
 });
--- a/autogpt_platform/frontend/src/app/(no-navbar)/onboarding/__tests__/page.test.tsx
+++ b/autogpt_platform/frontend/src/app/(no-navbar)/onboarding/__tests__/page.test.tsx
@@ -63,11 +63,14 @@ vi.mock("launchdarkly-react-client-sdk", () => ({
   }),
 }));
 
+const STEP_STORAGE_KEY = "autogpt:onboarding-highest-step";
+
 beforeEach(() => {
   currentSearchParams = new URLSearchParams();
   mockFlagValue = false;
   routerReplace.mockClear();
   useOnboardingWizardStore.getState().reset();
+  window.sessionStorage.removeItem(STEP_STORAGE_KEY);
 });
 
 afterEach(() => {
@@ -92,6 +95,7 @@ describe("OnboardingPage — flag-gated SubscriptionStep", () => {
 
   it("treats step 4 as Preparing when payments are gated off", async () => {
     mockFlagValue = false;
+    window.sessionStorage.setItem(STEP_STORAGE_KEY, "4");
     currentSearchParams = new URLSearchParams("step=4");
     render(<OnboardingPage />);
     expect(await screen.findByTestId("step-preparing")).toBeDefined();
@@ -100,6 +104,7 @@ describe("OnboardingPage — flag-gated SubscriptionStep", () => {
 
   it("renders SubscriptionStep at step 4 when payments are enabled", async () => {
     mockFlagValue = true;
+    window.sessionStorage.setItem(STEP_STORAGE_KEY, "4");
     currentSearchParams = new URLSearchParams("step=4");
     render(<OnboardingPage />);
     expect(await screen.findByTestId("step-subscription")).toBeDefined();
@@ -108,10 +113,31 @@ describe("OnboardingPage — flag-gated SubscriptionStep", () => {
 
   it("treats step 5 as Preparing when payments are enabled", async () => {
     mockFlagValue = true;
+    window.sessionStorage.setItem(STEP_STORAGE_KEY, "5");
     currentSearchParams = new URLSearchParams("step=5");
     render(<OnboardingPage />);
     expect(await screen.findByTestId("step-preparing")).toBeDefined();
     expect(screen.queryByTestId("step-subscription")).toBeNull();
+  });
+
+  it("clamps ?step=5 to the user's highest reached step (no fast-forward)", async () => {
+    mockFlagValue = true;
+    window.sessionStorage.setItem(STEP_STORAGE_KEY, "2");
+    currentSearchParams = new URLSearchParams("step=5");
+    render(<OnboardingPage />);
+    // Highest reached is 2 (RoleStep), so manually editing the URL to step=5
+    // should land the user back on step 2, not let them skip ahead.
+    expect(await screen.findByTestId("step-role")).toBeDefined();
+    expect(screen.queryByTestId("step-preparing")).toBeNull();
+  });
+
+  it("resumes from the highest reached step when ?step= is omitted", async () => {
+    mockFlagValue = true;
+    window.sessionStorage.setItem(STEP_STORAGE_KEY, "3");
+    // No step param — user navigated to /onboarding directly.
+    render(<OnboardingPage />);
+    expect(await screen.findByTestId("step-painpoints")).toBeDefined();
+    expect(screen.queryByTestId("step-welcome")).toBeNull();
   });
 
   it("rejects decimal step values and falls back to step 1", async () => {
--- a/autogpt_platform/frontend/src/app/(no-navbar)/onboarding/steps/SubscriptionStep/components/PlanCard/PlanCard.tsx
+++ b/autogpt_platform/frontend/src/app/(no-navbar)/onboarding/steps/SubscriptionStep/components/PlanCard/PlanCard.tsx
@@ -5,7 +5,7 @@ import { motion, useReducedMotion } from "framer-motion";
 import { Button } from "@/components/atoms/Button/Button";
 import { Text } from "@/components/atoms/Text/Text";
 import { cn } from "@/lib/utils";
-import { Check, Star } from "@phosphor-icons/react";
+import { CheckIcon, StarIcon } from "@phosphor-icons/react";
 import { type Country, formatPrice } from "../../countries";
 import { PLAN_KEYS, type PlanDef, type PlanKey } from "../../helpers";
 import { computePlanPricing } from "./helpers";
@@ -40,11 +40,13 @@ export function PlanCard({
 
   return (
     <div
+      aria-disabled={disabled || undefined}
       className={cn(
-        "relative rounded-2xl p-px",
+        "relative rounded-2xl p-px transition-opacity",
         hl
           ? "bg-gradient-to-br from-purple-500 via-fuchsia-500 to-indigo-500 shadow-[0_10px_32px_-8px_rgba(119,51,245,0.25)]"
           : "bg-gradient-to-br from-zinc-300 via-zinc-400 to-zinc-500 md:my-5",
+        disabled && "pointer-events-none opacity-50",
         className,
       )}
     >
@@ -84,7 +86,7 @@ export function PlanCard({
                     }
               }
             >
-              <Star
+              <StarIcon
                 size={10}
                 weight="fill"
                 aria-hidden="true"
@@ -189,7 +191,7 @@ export function PlanCard({
                     isTeam ? "bg-stone-100" : "bg-purple-50",
                   )}
                 >
-                  <Check
+                  <CheckIcon
                     size={10}
                     weight="bold"
                     className={isTeam ? "text-stone-500" : "text-purple-500"}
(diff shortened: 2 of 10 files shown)


autogpt_platform/frontend/src/app/(no-navbar)/onboarding/useOnboardingPage.ts
```diff
@@ -113,6 +146,11 @@ export function useOnboardingPage() {
     if (currentStep !== preparingStep || hasSubmitted.current) return;
     hasSubmitted.current = true;
 
+    // Profile is submitted before the Stripe Checkout redirect (the wizard
+    // store is in-memory and doesn't survive the round-trip). Skip the
+    // resubmit on return to avoid overwriting saved data with empties.
+    if (searchParams.get("subscription") === "success") return;
```
0ubbe: 🤖 🟡 **Nice to Have**: This skip is keyed solely on `?subscription=success` — if that param is ever stripped (manual URL edit, share link, an upstream proxy/redirect normalising query params) and the user lands on `?step=5` again, the post-redirect zustand store is empty (in-memory, lost on full navigation) and we POST an empty `name`/`role`/`pain_points` payload, overwriting the profile we just persisted in `useSubscriptionStep`. Suggest a belt-and-suspenders check: also skip the resubmit when `name` is empty (no valid profile to save anyway). Same effect, robust to URL drift.

autogpt_platform/frontend/src/app/(no-navbar)/onboarding/__tests__/page.test.tsx
```diff
@@ -108,12 +113,33 @@ describe("OnboardingPage — flag-gated SubscriptionStep", () => {
...
+  });
+
+  it("resumes from the highest reached step when ?step= is omitted", async () => {
+    mockFlagValue = true;
+    window.sessionStorage.setItem(STEP_STORAGE_KEY, "3");
+    // No step param — user navigated to /onboarding directly.
+    render(<OnboardingPage />);
+    expect(await screen.findByTestId("step-painpoints")).toBeDefined();
+    expect(screen.queryByTestId("step-welcome")).toBeNull();
+  });
```
0ubbe, line 141: 🤖 🟡 **Nice to Have**: Two of the trickiest paths in this PR have no integration coverage:

1. The Stripe-success return: `?step=5&subscription=success` should land on `step-preparing` AND skip the profile resubmit (otherwise we overwrite the profile we already persisted before redirect — and Sentry already caught a closely-related ceiling bug at runtime, so this path is in active use).
2. The Stripe-cancel return: `?step=4&subscription=cancelled` should land on `step-subscription` (re-selectable plan cards).

Adding two tests here — one with `STEP_STORAGE_KEY=4` + `?step=5&subscription=success` asserting `step-preparing` renders, and one with `?step=4&subscription=cancelled` asserting `step-subscription` renders + `selectedPlan` is null — would lock in the post-redirect contract. The MSW pattern in `SubscriptionStep.test.tsx` for asserting on POST bodies could verify the resubmit-skip too.

autogpt_platform/frontend/src/app/(no-navbar)/onboarding/steps/SubscriptionStep/components/PlanCard/PlanCard.tsx
```diff
@@ -143,33 +181,53 @@ export function PlanCard({
...
+                }
+              />
+            )}
+            <Button
+              variant={plan.buttonVariant}
+              size={hl ? "large" : "small"}
+              onClick={() => onSelect(plan.key)}
+              className="w-full"
+              loading={loading}
+              disabled={disabled}
```
0ubbe, line 228: 🤖 🔵 **Nit**: When `disabled` is true (a sibling card is loading), only the `<Button>` reflects the disabled state — the rest of the card (gradient border, plan price, feature list) still looks fully interactive. A user mid-click on a different card has no visual cue that this card is locked, and screen readers have no way to know either. Suggest adding `aria-disabled={disabled}` and an `opacity-50 pointer-events-none` modifier on the outer wrapper `<div>` (around line 42), or wrapping the card body in a `<fieldset disabled>` so all interactive children inherit the state.

Tiny polish — doesn't block correctness because the click handler is debounced via `if (isUpdatingTier) return;` in the hook.

autogpt_platform/frontend/src/app/(no-navbar)/onboarding/steps/SubscriptionStep/useSubscriptionStep.ts
```diff
@@ -29,13 +45,65 @@ export function useSubscriptionStep() {
...
+          ? [otherPainPoint.trim()]
+          : [],
+      );
+
+    try {
+      await postV1SubmitOnboardingProfile({
+        user_name: name,
+        user_role: resolvedRole,
+        pain_points: resolvedPainPoints,
+      }).catch(() => undefined);
```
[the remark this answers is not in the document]
0ubbe: 🤖 Confirmed. The cancelled-return path now lands on `?step=4&subscription=cancelled` and the init effect's `reset()` clears the in-memory zustand state — there's no follow-up code path that POSTs the empty profile back to the backend, so saved data is preserved. The remaining brittleness (manual URL edit dropping `subscription=success`) is now belt-and-suspenders-guarded in [1a02912df](https://github.com/Significant-Gravitas/AutoGPT/commit/1a02912df845d8df61d72374b78137c436d93d6f): `useOnboardingPage` skips the resubmit whenever `name` is empty, regardless of the query param.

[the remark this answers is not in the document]
0ubbe: 🤖 Fixed in [1a02912df](https://github.com/Significant-Gravitas/AutoGPT/commit/1a02912df845d8df61d72374b78137c436d93d6f): dropped the `.catch(() => undefined)` so a failed profile write throws and is caught by the outer `try/catch`, surfacing the error toast and aborting before the Stripe redirect. Profile data is now guaranteed to be persisted before checkout.

autogpt_platform/frontend/src/app/(no-navbar)/onboarding/useOnboardingPage.ts
```diff
@@ -70,22 +93,31 @@ export function useOnboardingPage() {
...
     hasInitialized.current = true;
-    const urlStep = parseStep(searchParams.get("step"), preparingStep);
+    const urlStep = parseStepParam(searchParams.get("step"), preparingStep);
+    const ceiling = Math.min(readHighestStep(), preparingStep) as Step;
+    const target = (
+      urlStep === null ? ceiling : Math.min(urlStep, ceiling)
+    ) as Step;
     reset();
-    goToStep(urlStep);
+    goToStep(target);
```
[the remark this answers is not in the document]
0ubbe: 🤖 Fixed in [1a02912df](https://github.com/Significant-Gravitas/AutoGPT/commit/1a02912df845d8df61d72374b78137c436d93d6f): the init effect now treats `?subscription=success` as a trusted intent and uses `preparingStep` as the ceiling instead of `readHighestStep()`, so the user advances to step 5 even though the stored highest is 4. Picked this over bumping `writeHighestStep(preparingStep)` pre-redirect because it keeps the storage authoritative for legitimate user-driven progress (no spoofed-success path that survives a closed Stripe checkout). Added `page.test.tsx` regression test `lets ?step=5&subscription=success bypass the highestStep ceiling`.

autogpt_platform/frontend/src/components/atoms/Emoji/Emoji.tsx
```diff
@@ -6,22 +6,28 @@ interface Props {
...
-          folder: "svg",
-          ext: ".svg",
-          attributes: () => ({
-            width: String(size),
-            height: String(size),
-          }),
-        }),
-      }}
+      dangerouslySetInnerHTML={{ __html: html }}
     />
```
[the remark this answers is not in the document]
0ubbe: 🤖 Fixed in [1a02912df](https://github.com/Significant-Gravitas/AutoGPT/commit/1a02912df845d8df61d72374b78137c436d93d6f): added an `escapeHtml` step before `twemoji.parse` — emoji codepoints aren't HTML metacharacters so they pass through untouched, while any embedded markup is now neutered. Trades a tiny perf cost (one regex pass per render) for a contract that's safe regardless of caller input. Current callers all pass static literals, so no behavioral change for them.

autogpt_platform/frontend/src/app/(no-navbar)/onboarding/useOnboardingPage.ts
0ubbe: 🤖 Fixed in [1a02912df](https://github.com/Significant-Gravitas/AutoGPT/commit/1a02912df845d8df61d72374b78137c436d93d6f): added the `!name.trim()` guard so the resubmit is also skipped when the zustand store is empty — protects against URL-drift cases (manual edit, share link, upstream proxy stripping the query) and keeps the original `subscription=success` short-circuit for fast-path symmetry.

autogpt_platform/frontend/src/app/(no-navbar)/onboarding/__tests__/page.test.tsx
0ubbe: 🤖 Fixed in [1a02912df](https://github.com/Significant-Gravitas/AutoGPT/commit/1a02912df845d8df61d72374b78137c436d93d6f): added two integration tests at the bottom of `page.test.tsx` — `lets ?step=5&subscription=success bypass the highestStep ceiling` (asserts `step-preparing` renders even with stored highest=4) and `returns to SubscriptionStep on ?step=4&subscription=cancelled` (asserts plan picker re-renders and `selectedPlan` is null).

autogpt_platform/frontend/src/app/(no-navbar)/onboarding/steps/SubscriptionStep/components/PlanCard/PlanCard.tsx
0ubbe: 🤖 Fixed in [1a02912df](https://github.com/Significant-Gravitas/AutoGPT/commit/1a02912df845d8df61d72374b78137c436d93d6f): added `aria-disabled={disabled || undefined}` plus `opacity-50 pointer-events-none` (with `transition-opacity`) on the outer wrapper so the card visibly dims and is non-interactive when a sibling is loading. Skipped `<fieldset>` to avoid disturbing the existing flex layout / motion wrappers.

0ubbe:
🤖 Top-level CodeRabbit review feedback addressed in [1a02912df](https://github.com/Significant-Gravitas/AutoGPT/commit/1a02912df845d8df61d72374b78137c436d93d6f):

1. **In-flight UI assertion in `SubscriptionStep.test.tsx`** — added `clicking a plan keeps the request in flight: clicked card spins, others lock`. Holds the `/api/credits/subscription` MSW handler open and asserts the Max + Team buttons become `disabled` while PRO is in flight.
2. **`-Icon` suffixed Phosphor imports in `PlanCard.tsx`** — switched `Check` → `CheckIcon` and `Star` → `StarIcon`.

Both threads can now be marked resolved.

0ubbe:
🤖 Late CodeRabbit nitpick on `useSubscriptionStep.ts:91` (explicit checkout response type) addressed in [3e1de8859](https://github.com/Significant-Gravitas/AutoGPT/commit/3e1de88595128e14d620d960903ff208712907e1): extracted `interface CheckoutResponse { url?: string }` and use it for the cast.

autogpt_platform/frontend/src/app/(no-navbar)/onboarding/steps/SubscriptionStep/useSubscriptionStep.ts
```diff
@@ -15,6 +31,10 @@ export function useSubscriptionStep() {
   );
   const setSelectedPlan = useOnboardingWizardStore((s) => s.setSelectedPlan);
   const nextStep = useOnboardingWizardStore((s) => s.nextStep);
+  const selectedPlan = useOnboardingWizardStore((s) => s.selectedPlan);
+
+  const { mutateAsync: updateTier, isPending: isUpdatingTier } =
+    useUpdateSubscriptionTier();
```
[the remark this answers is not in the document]
0ubbe: 🤖 Fixed in [4cee9572b](https://github.com/Significant-Gravitas/AutoGPT/commit/4cee9572b9f82245480142c747e2e87686297a54): added a local `useState` in-flight guard that flips synchronously on first click (`if (isProcessing) return; setIsSubmitting(true);`) before any await, and clears in `finally`. The combined pending state (`isUpdatingTier || isSubmitting`) is now exposed to the plan cards via the existing `isUpdatingTier` slot, so the loading/disabled props activate immediately rather than after the profile POST returns. Tests still pass — the in-flight UI test that holds the subscription POST open keeps the cards locked across both phases now (previously the gap was masked by the test resolving the profile POST before re-checking).

0ubbe:
🤖 CodeRabbit nitpick on `useSubscriptionStep.ts` lines 72-81 (duplicated Other / Something else normalization) addressed in [890c590a6](https://github.com/Significant-Gravitas/AutoGPT/commit/890c590a6fb5afb743a15e14a3a0a63947af1ad8): extracted to a shared `autogpt_platform/frontend/src/app/(no-navbar)/onboarding/helpers.ts` exposing `normalizeOnboardingProfile`, called from both `useSubscriptionStep` and `useOnboardingPage`. All 64 onboarding tests still pass.

PR #12976, opened 2026-05-01
refactor(frontend): settings v2 cleanup — sidebar, account route, headers, wallet

Abhi1992002:
### Why / What / How

**Why:** Settings v2 had several rough edges flagged in design review — stale sidebar labels/order, dead OAuth Apps tab, mismatched header alignment, small card radii, a redundant warning-icon dialog, the wrong title on the wallet popover, and an outdated avatar menu still pointing to legacy `/profile/*` routes. This PR cleans them up in one pass so the new settings surface lines up with current product copy and design tokens.

**What:** Sidebar relabel + reorder, route rename `/settings/preferences` → `/settings/account`, header tweaks, dialog cleanup, and a wallet popover redesign that matches the existing component style.

**How:** Pure UI/copy/route changes — no backend touched, no API contracts modified. Component files were renamed/moved with `git mv` semantics so history is preserved. Tests updated to reflect the new sidebar fixture and skipped where the underlying section was hidden for this PR (Notifications). Avatar menu uses the existing dynamic `getAccountMenuItems()` so role-aware behavior is unchanged.

### Changes 🏗️

**Sidebar / navigation**
- Renamed `Preferences` → `Account` and reordered: Profile → Account → Billing → Integrations → API Keys → Creator Dashboard
- Hid `OAuth Apps` from sidebar (page kept at `/settings/oauth-apps`, just no nav entry)
- Renamed route folder `/settings/preferences` → `/settings/account`; updated sidebar href + tests

**Avatar popover (Navbar)**
- `Edit profile`, `Creator Dashboard`, and `Settings` now point to the new `/settings/*` pages instead of legacy `/profile/*`
- Both the static `accountMenuItems` and the dynamic `getAccountMenuItems()` were updated (the role-aware one is what the Navbar actually consumes)

**Account page (`/settings/account`)**
- Renamed page H1 + `document.title` to "Account"
- Header subtitle trimmed (notifications copy removed)
- `NotificationsCard` temporarily hidden — coming back in a follow-up PR with copy/UX changes
- `AccountCard`'s redundant "Account / Manage your sign-in details" inner title removed (page heading already says it)
- Notification-dependent tests `test.skip`'d; AccountCard tests stay green

**Integrations**
- `DeleteConfirmDialog`: removed the red `WarningCircleIcon` and the bulk preview list — message text alone is clearer
- `ProviderGroup`: bumped card radius (`rounded-lg` → `rounded-2xl`) to match the design system
- Header aligned with card content (`pl-4`)

[...]

PR #12979, opened 2026-05-01
refactor(platform): creator dashboard polish + MCP setup + settings link cleanup

Abhi1992002:
### Why / What / How

**Why** — Follow-up to the settings v2 cleanup (#12976). Several rough edges remained after the route migration:

- The Team-tier "Talk to sales" button was firing into a 404 (`agpt.co/contact-sales` no longer exists).
- A bunch of in-app links were still pointing to the deprecated `/profile/*` settings routes, which now route through the legacy layout instead of the new `/settings/*` pages.
- The MCP integration entry in the "Connect a service" dialog was a dead end — the provider showed up in the list but the detail view said "No connection method available".
- `/settings/creator-dashboard` shipped a bulk-select feature (checkbox column, select-all bar, bulk-delete dialog) that nobody is using; it added clutter to the row UI.
- `/settings` (no subpath) flashed an empty page during the client-side redirect to `/settings/profile`.
- The submissions API was issuing a redundant `COUNT(*)` per request because `_get_submission_stats` already returns `total`.

**What** — One PR, several small fixes packaged together because they all touch the same surface (creator dashboard / integrations / settings linking).

**How** — Most changes are one-liners or small file deletions. Two are slightly bigger:
- Backend `_get_submission_stats` is reused for pagination total, so `get_store_submissions` drops the duplicate `.count()` call. Stats themselves use a single Postgres query with `FILTER` clauses for total/approved/pending/total_runs/average_rating.
- MCP setup is now rendered inline in the Connect Service dialog via a new `McpConnectPanel` component when `provider.id === "mcp"`. Server URL → tries OAuth via popup → on 400 falls back to bearer-token input → on success invalidates `getGetV1ListCredentialsQueryKey()` so the credential shows up in the integrations list.

### Changes 🏗️

**Backend (`backend/api/features/store`)**
- Add `SubmissionStats` payload to `StoreSubmissionsResponse` (total, approved, pending, total_runs, average_rating) computed in one Postgres `FILTER` round-trip.
- Reuse `stats.total` for pagination instead of running a separate `COUNT(*)` query.

**Frontend creator dashboard (`/settings/creator-dashboard`)**
- Remove the bulk-select feature: checkbox column + selection bar + bulk-delete dialog + `useSubmissionSelection` hook + `SubmissionSelectionBar` + `MobileSelectionBar` are all deleted.
- Update tests to drop the now-irrelevant selection cases.

**Frontend integrations (`/settings/integrations`)**
- New `McpConnectPanel` rendered inside the Connect Service dialog DetailView when the user picks the Mcp provider. OAuth-first with bearer-token fallback when the server returns 400.
- Light `pl-1` polish on the DetailView header to align the back button + avatar with the body content.

**Frontend settings link migration**
- Replace stale `/profile/*` deep links across the platform with the new `/settings/*` routes:
  - `/profile/dashboard` → `/settings/creator-dashboard` (publish modal "View progress")
  - `/profile/credits` → `/settings/billing` (usage limits, briefing panel, wallet refill)
  - `/profile/integrations` → `/settings/integrations` (integrations setup wizard)
  - `/profile/settings` → `/settings/account` (timezone notices in scheduling dialogs)
- Legacy `/profile/(user)/layout.tsx` self-nav left alone — that whole route group is being deprecated in a follow-up PR.

**Frontend billing**
- `TEAM_UPGRADE_URL` switched from `https://agpt.co/contact-sales` (404) to the Tally intake form `https://tally.so/r/2Eb9zj`. Already opens with `target="_blank"` + `noopener,noreferrer`.

**Frontend settings index**
- `/settings` renders a content skeleton during the client-side redirect to `/settings/profile`, so the page no longer flashes empty.

#### For code changes:
- [x] I have made a test plan
  - [x] Open `/settings/billing` on a MAX plan → click "Talk to sales — Team" → Tally form opens in a new tab.
  - [x] On `/settings/integrations` → "Connect Service" → pick "Mcp" → enter a server URL → confirm OAuth popup launches; for an OAuth-less server, confirm bearer-token fallback appears and saves the credential.
  - [x] Confirm the new MCP credential shows up in the integrations list after connect.
  - [x] On `/settings/creator-dashboard` → confirm no checkbox column, no select-all bar, no bulk-delete dialog. Row-level "Edit"/"Delete" via the dropdown still works.
  - [x] Submit an agent for review from `/build` → "View progress" → lands on `/settings/creator-dashboard` (was 404 before).
  - [x] Click any wallet "Add credits" / "Manage billing" / usage-limit link → lands on `/settings/billing` (not `/profile/credits`).
  - [x] Open the schedule dialog with no timezone set → "Set your timezone" link goes to `/settings/account`.
  - [x] Hit `/settings` directly → see the content skeleton, then redirect to `/settings/profile`.
  - [x] Verify `/v2/store/submissions` response contains `stats: { total, approved, pending, total_runs, average_rating }` and pagination `total_items` matches `stats.total`.


autogpt_platform/frontend/src/app/(platform)/settings/integrations/components/ConnectServiceDialog/components/DetailView/McpConnectPanel.tsx
```diff
@@ -0,0 +1,216 @@
...
+  const [phase, setPhase] = useState<Phase>("form");
+  const [isSubmitting, setIsSubmitting] = useState(false);
+  const [error, setError] = useState<string | null>(null);
+  const oauthAbortRef = useRef<((reason?: string) => void) | null>(null);
+
+  useEffect(() => () => oauthAbortRef.current?.(), []);
+
+  const trimmedUrl = serverUrl.trim();
+  const trimmedToken = token.trim();
+  const canConnect = trimmedUrl.length > 0 && !isSubmitting;
```
Abhi1992002: 🟡 **Nice to Have**: `canConnect` only checks `trimmedUrl.length > 0`, so the user can submit a non-URL or non-https string and only get feedback after the backend rejects it. Consider validating with `URL()` and rejecting non-`http(s):` schemes here so the form gives immediate feedback (and we never accidentally hand a `javascript:` / `file:` URL to `postV2InitiateOauthLoginForAnMcpServer`).

```ts
function isValidHttpUrl(value: string) {
  try {
    const u = new URL(value);
    return u.protocol === "http:" || u.protocol === "https:";
  } catch {
    return false;
  }
}
```

autogpt_platform/frontend/src/providers/agent-credentials/credentials-provider.tsx
```diff
@@ -364,6 +367,24 @@ export default function CredentialsProvider({
...
+  // invalidates the credentials list query (e.g. settings v2 create/delete
+  // flows using generated hooks) will trigger a reload here, so the builder
+  // never sees stale credentials without callers having to opt in.
+  useEffect(() => {
+    const targetKey = getGetV1ListCredentialsQueryKey();
+    const unsubscribe = queryClient.getQueryCache().subscribe((event) => {
+      if (
+        event.type === "updated" &&
+        event.action.type === "invalidate" &&
+        event.query.queryKey[0] === targetKey[0]
```
Abhi1992002: 🟡 **Nice to Have**: This subscription compares only `queryKey[0]` and matches on the literal string `event.action.type === "invalidate"`. Two fragility points worth noting:

1. `getGetV1ListCredentialsQueryKey()` may be parameterised in future (e.g. per-provider scoping), and a parameter change would silently bypass this listener since only index `[0]` is compared.
2. The `event.action.type` literal is internal to react-query and not part of its public API contract — a minor version bump can rename it.

A cleaner alternative: register the listener with `queryClient.getQueryCache().subscribe` filtered by `{ predicate: ... }` on the actual query, or just call `loadCredentials()` directly from the call sites that invalidate (the MCP panel already invalidates explicitly). That keeps the data flow inspectable without a global cache listener.

autogpt_platform/frontend/src/providers/agent-credentials/credentials-provider.tsx
```diff
@@ -364,6 +367,24 @@ export default function CredentialsProvider({
...
+      ) {
+        loadCredentials();
+      }
+    });
+    return unsubscribe;
+  }, [queryClient, loadCredentials]);
```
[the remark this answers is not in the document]
Abhi1992002: Verified `{ type: 'invalidate' }` IS dispatched by TS Query 5.90.6 (`query.cjs:180`), so the original listener does fire. But the underlying concern stands — the listener was fragile. Fixed in f39a17de0: switched to `hashKey()` matching and now accept both `invalidate` and `success` actions so we cover both inactive queries (invalidate only) and active ones (success after refetch).

autogpt_platform/frontend/src/app/(platform)/settings/integrations/components/ConnectServiceDialog/components/DetailView/McpConnectPanel.tsx
Abhi1992002: Fixed in f39a17de0 — added `isValidHttpUrl()` helper that parses with `URL()` and rejects non-`http(s):` schemes. Wired into `canConnect` and `canSubmitToken` so the Connect / Save token buttons stay disabled until a valid URL is entered.

autogpt_platform/frontend/src/providers/agent-credentials/credentials-provider.tsx
Abhi1992002: Fixed in f39a17de0 — replaced the `queryKey[0]` shallow compare with `hashKey()` from the public `@tanstack/react-query` API (resilient to future parameterisation), and now accept both `invalidate` and `success` action types so the listener still fires if either internal literal renames in a future minor (success is the canonical post-refetch event).

autogpt_platform/backend/backend/api/features/store/db.py
```diff
@@ -607,34 +607,67 @@ async def get_store_creator(
...
+    sql = """
+        SELECT
+            COUNT(*)::int                                          AS total,
+            COUNT(*) FILTER (WHERE status = 'APPROVED')::int       AS approved,
+            COUNT(*) FILTER (WHERE status = 'PENDING')::int        AS pending,
+            COALESCE(SUM(run_count), 0)::bigint                    AS total_runs,
+            CAST(
+                AVG(review_avg_rating) FILTER (WHERE review_avg_rating > 0)
+                AS double precision
+            )                                                      AS average_rating
```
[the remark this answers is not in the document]
Abhi1992002: Fixed in f39a17de0 — `average_rating` now uses `SUM(review_avg_rating * review_count) / NULLIF(SUM(review_count), 0)`, weighted by review volume. Filter excludes both `review_count = 0` and `review_avg_rating = 0` so unreviewed submissions don't drag the average toward zero.

autogpt_platform/frontend/src/app/(platform)/settings/integrations/components/ConnectServiceDialog/components/DetailView/McpConnectPanel.tsx
```diff
@@ -0,0 +1,216 @@
...
+          onChange={(e) => setToken(e.target.value)}
+          disabled={isSubmitting}
+          hint="Used as a Bearer Authorization header when calling tools."
+        />
+      ) : null}
+
+      {error ? (
+        <div className="rounded-lg border border-red-200 bg-red-50 px-3 py-2 text-sm text-red-700">
+          {error}
+        </div>
```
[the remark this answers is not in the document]
Abhi1992002: Fixed in f39a17de0 — added `role="alert"` and `aria-live="polite"` on the error banner so async failures are announced to assistive tech.

autogpt_platform/frontend/src/app/(platform)/settings/integrations/components/ConnectServiceDialog/components/DetailView/McpConnectPanel.tsx
```diff
@@ -0,0 +1,230 @@
...
+      onSuccess();
+    } catch (e: unknown) {
+      const status = getErrorStatus(e);
+      if (status === 400) {
+        setPhase("manual-token");
+        setError(
+          "This server doesn't support OAuth sign-in. Paste a bearer token instead.",
+        );
+        return;
+      }
```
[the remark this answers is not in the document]
Abhi1992002: Fixed in 144b0677e — wrapped `postV2InitiateOauthLoginForAnMcpServer` in its own try/catch so the 400→manual-token fallback only triggers when the *initiate* call returns 400. 400s from the popup callback or `postV2ExchangeOauthCodeForMcpTokens` now propagate to the outer catch and surface as actual error messages, matching the suggested diff.

autogpt_platform/frontend/src/app/(platform)/settings/integrations/components/ConnectServiceDialog/components/DetailView/__tests__/McpConnectPanel.test.tsx
```diff
@@ -0,0 +1,183 @@
...
+    fireEvent.click(screen.getByRole("button", { name: /save token/i }));
+
+    await waitFor(() => {
+      expect(onSuccess).toHaveBeenCalledTimes(1);
+    });
+    expect(postV2StoreABearerTokenForAnMcpServer).toHaveBeenCalledWith({
+      server_url: "https://mcp.example.com",
+      token: "secret-bearer-token",
+    });
+  });
```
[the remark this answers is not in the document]
Abhi1992002: Fixed in ce3cbac76 — renamed to "submits a bearer token then calls onSuccess" so the title matches what the assertions actually verify. Adding an explicit invalidation assert would require spying on `queryClient` from the test wrapper which felt heavier than the rename for the same signal.

autogpt_platform/backend/backend/api/features/store/db.py
```diff
@@ -607,34 +607,76 @@ async def get_store_creator(
...
+            / NULLIF(
+                SUM(review_count) FILTER (
+                    WHERE review_count > 0 AND review_avg_rating > 0
+                ),
+                0
+            )                                                      AS average_rating
+        FROM {schema_prefix}"StoreSubmission"
+        WHERE user_id = $1 AND is_deleted = false
+    """
+    rows = await query_raw_with_schema(sql, user_id)
```
ntindle: I think you can pass a schema to this so you don’t need to parse the results like that

ntindle (dismissed):

autogpt_platform/backend/backend/api/features/store/db.py
Abhi1992002: 🤖 Fixed in [2234df33e](https://github.com/Significant-Gravitas/AutoGPT/commit/2234df33e6fe7242d911a53ef7f553dd50887912): `query_raw_with_schema` now forwards Prisma’s `model=` argument and `_get_submission_stats` passes `SubmissionStats`, so the raw result is deserialized directly into the schema instead of manually reading dict fields.

ntindle (approved):

PR #12980, opened 2026-05-01
fix(frontend): persist onboarding store + skip Stripe on LOCAL

0ubbe:
### Why / What / How

**Why.** Three bugs, one PR — all rooted in the onboarding wizard's lack of persistence:

1. **422 on `/api/onboarding/profile` mid-wizard.** The wizard store was in-memory and `useOnboardingPage`'s init effect called `reset()` on every mount. A user who refreshed mid-flow (or navigated to `?step=4` directly) hit the SubscriptionStep with empty `name` / `role`; clicking *Get Pro* / *Upgrade to Max* advanced to Preparing, which POSTed empty `user_name` / `user_role` and got rejected with `string_too_short`.
2. **LOCAL crash on plan select.** The Stripe checkout path is unconditional. On `BEHAVE_AS=LOCAL` the backend has no Stripe wiring, so plan selection blew up trying to start a Checkout session.
3. **Welcome detour after completion.** `handlePreparingComplete` and `checkCompletion` called `reset()`, which set `currentStep=1` after `router.replace("/copilot")` had already been queued. The URL-sync effect then fired a second `router.replace("/onboarding?step=1")` that won, stranding the user on Welcome until they refreshed.

**What.**
- Wrap the wizard store with `zustand/middleware`'s `persist` (sessionStorage, matches the existing `STEP_STORAGE_KEY` ceiling). `partialize` excludes `currentStep` — the URL stays the source of truth for which step the user is on.
- Drop the unconditional `reset()` from the init effect so persisted form data survives refreshes.
- Replace `reset()` with `useOnboardingWizardStore.persist.clearStorage()` on completion paths to clean storage without mutating in-memory state (no spurious re-render, no URL-sync race).
- Short-circuit `handlePlanSelect` on `environment.isLocal()` to skip the profile POST + Stripe Checkout and advance straight to Preparing; cloud path is unchanged.

**How.**
- `store.ts` — `persist(...)` with SSR-safe `createJSONStorage` (no-op stub during SSR/vitest), `partialize` exposing only the form fields. `version: 1` so future schema changes have a migration anchor.
- `useOnboardingPage.ts` — remove `reset()` from init effect; swap completion-path `reset()` calls for `persist.clearStorage()`. Comment block explains the URL-sync race so the next reader doesn't reintroduce it.
- `useSubscriptionStep.ts` — early-return on `environment.isLocal()` after the TEAM/inflight guards, before any Stripe code.
- Tests:
  - `page.test.tsx` — regression test that pre-set form data isn't wiped on mount (the bug the persist change fixes).
  - `SubscriptionStep.test.tsx` — `vi.spyOn(environment, "isLocal").mockReturnValue(false)` in `beforeEach` so existing Stripe-path tests stay in cloud mode; new test flips it `true` and asserts no Stripe / profile request fires and `currentStep` advances to 5.

### Changes 🏗️

- `autogpt_platform/frontend/src/app/(no-navbar)/onboarding/store.ts` — `zustand/middleware`'s `persist` with sessionStorage + `partialize` excluding `currentStep`.
- `autogpt_platform/frontend/src/app/(no-navbar)/onboarding/useOnboardingPage.ts` — drop `reset()` from init; switch completion `reset()` to `persist.clearStorage()`.
- `autogpt_platform/frontend/src/app/(no-navbar)/onboarding/steps/SubscriptionStep/useSubscriptionStep.ts` — early-return on `environment.isLocal()` to skip Stripe.
- `autogpt_platform/frontend/src/app/(no-navbar)/onboarding/__tests__/page.test.tsx` — new "preserves form data on mount" regression test.
- `autogpt_platform/frontend/src/app/(no-navbar)/onboarding/steps/__tests__/SubscriptionStep.test.tsx` — force cloud mode in `beforeEach`; add LOCAL bypass test.

#### For code changes:
- [x] I have made a test plan
  - [ ] On `BEHAVE_AS=LOCAL`: complete onboarding through Welcome → Role → PainPoints → Subscription, click *Get Pro* → no Stripe Checkout, lands on Preparing, then `/copilot` (no Welcome bounce, no refresh needed). Repeat with *Upgrade to Max*.
  - [ ] On `BEHAVE_AS=LOCAL`: click *Contact sales* (Team) → Tally form opens in new tab; wizard stays on step 4.
  - [ ] On `BEHAVE_AS=CLOUD`: existing Stripe Checkout flow unchanged — *Get Pro* / *Upgrade to Max* redirect to Stripe with the right `success_url` / `cancel_url`; success returns to `?step=5&subscription=success` and lands on `/copilot`; cancel returns to `?step=4&subscription=cancelled`.
  - [ ] Mid-wizard refresh on step 4 with `payments` flag enabled: name / role / pain points stay filled in (no 422, no re-fill required).
  - [ ] Direct nav to `/onboarding?step=4` without prior data: clamps back to the highest reached step (or step 1 on a fresh session) — no fast-forward into Subscription.
  - [ ] Returning user with `VISIT_COPILOT` already complete: redirects to `/copilot` cleanly with no Welcome flash.

### DIFF
--- a/autogpt_platform/frontend/src/app/(no-navbar)/onboarding/steps/SubscriptionStep/useSubscriptionStep.ts
+++ b/autogpt_platform/frontend/src/app/(no-navbar)/onboarding/steps/SubscriptionStep/useSubscriptionStep.ts
@@ -78,18 +78,17 @@ export function useSubscriptionStep() {
     setSelectedPlan(planKey);
     const tier = PLAN_TO_TIER[planKey];
 
-    // Profile data lives in an in-memory zustand store, so persist it before
-    // Stripe takes the user off-page — otherwise the user's name/role/pain
-    // points are lost on the post-checkout redirect.
     const { name, role, painPoints } = normalizeOnboardingProfile(
       useOnboardingWizardStore.getState(),
     );
 
     try {
-      // Profile must be persisted before Stripe takes over — the zustand
-      // store is in-memory and the post-checkout return is a full page
-      // navigation. Abort on failure rather than silently losing the user's
-      // name / role / pain points.
+      // POST the profile pre-redirect as defence in depth: zustand persist
+      // (sessionStorage) keeps the store across the Stripe round-trip, but
+      // submitting now also covers the case where the user closes the tab
+      // mid-checkout — the backend still has their name / role / pain
+      // points. Abort on failure rather than starting a Checkout session
+      // we'd have to reconcile with a missing profile after success.
       await postV1SubmitOnboardingProfile({
         user_name: name,
         user_role: role,
--- a/autogpt_platform/frontend/src/app/(no-navbar)/onboarding/steps/SubscriptionStep/useSubscriptionStep.ts
+++ b/autogpt_platform/frontend/src/app/(no-navbar)/onboarding/steps/SubscriptionStep/useSubscriptionStep.ts
@@ -63,6 +63,7 @@ export function useSubscriptionStep() {
       return;
     }
     if (isProcessing) return;
+    setIsSubmitting(true);
 
     // Local dev: backend has no Stripe wiring, so skip the checkout
     // round-trip and advance straight to Preparing. The profile still
@@ -72,8 +73,6 @@ export function useSubscriptionStep() {
       nextStep();
       return;
     }
-
-    setIsSubmitting(true);
 
     setSelectedPlan(planKey);
     const tier = PLAN_TO_TIER[planKey];
--- a/autogpt_platform/frontend/src/app/(no-navbar)/onboarding/steps/SubscriptionStep/useSubscriptionStep.ts
+++ b/autogpt_platform/frontend/src/app/(no-navbar)/onboarding/steps/SubscriptionStep/useSubscriptionStep.ts
@@ -2,6 +2,7 @@ import { useUpdateSubscriptionTier } from "@/app/api/__generated__/endpoints/cre
 import { postV1SubmitOnboardingProfile } from "@/app/api/__generated__/endpoints/onboarding/onboarding";
 import type { SubscriptionTierRequestTier } from "@/app/api/__generated__/models/subscriptionTierRequestTier";
 import { toast } from "@/components/molecules/Toast/use-toast";
+import { environment } from "@/services/environment";
 import { useState } from "react";
 import { normalizeOnboardingProfile } from "../../helpers";
 import { useOnboardingWizardStore } from "../../store";
@@ -62,6 +63,16 @@ export function useSubscriptionStep() {
       return;
     }
     if (isProcessing) return;
+
+    // Local dev: backend has no Stripe wiring, so skip the checkout
+    // round-trip and advance straight to Preparing. The profile still
+    // gets POSTed there via useOnboardingPage's submission effect.
+    if (environment.isLocal()) {
+      setSelectedPlan(planKey);
+      nextStep();
+      return;
+    }
+
     setIsSubmitting(true);
 
     setSelectedPlan(planKey);
--- a/autogpt_platform/frontend/src/app/(no-navbar)/onboarding/store.ts
+++ b/autogpt_platform/frontend/src/app/(no-navbar)/onboarding/store.ts
@@ -1,4 +1,5 @@
 import { create } from "zustand";
+import { createJSONStorage, persist } from "zustand/middleware";
 
 export const MAX_PAIN_POINT_SELECTIONS = 3;
 export type Step = 1 | 2 | 3 | 4 | 5;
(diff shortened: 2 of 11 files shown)


autogpt_platform/frontend/src/app/(no-navbar)/onboarding/store.ts
```diff
@@ -27,75 +28,104 @@ interface OnboardingWizardState {
...
+      // currentStep is intentionally excluded — the URL is the source of
+      // truth for which step the user is on, and the page hook syncs the
+      // store from the URL on mount.
+      partialize: (state) => ({
+        name: state.name,
+        role: state.role,
+        otherRole: state.otherRole,
+        painPoints: state.painPoints,
+        otherPainPoint: state.otherPainPoint,
+        selectedPlan: state.selectedPlan,
```
0ubbe: 🤖 🟡 **Nice to Have**: `selectedPlan` is persisted but conceptually transient. Today the only consumers are `loading={isUpdatingTier && selectedPlan === plan.key}` / `disabled={isUpdatingTier && selectedPlan !== plan.key}`, both gated by `isUpdatingTier` — so after a Stripe cancel/return (`?step=4&subscription=cancelled`), the stale `selectedPlan` has no visible effect.

It is, however, a footgun: any future consumer that uses `selectedPlan` for visual "this card is highlighted" state will silently surface the previous click after cancel-and-return. Two safe options: (a) drop `selectedPlan` from `partialize` (it really only matters for the in-flight Stripe round-trip, and that round-trip survives via Stripe redirect, not the user closing the tab), or (b) add a brief comment here pinning the contract that `selectedPlan` is *only* meaningful when `isUpdatingTier` is true.

0ubbe: 🤖 Fixed in [7916fd3fc](https://github.com/Significant-Gravitas/AutoGPT/commit/7916fd3fcebe190138aeb2e7b4710bca34cbe322): dropped `selectedPlan` from `partialize` (option a) and updated the comment to spell out why. The state was effectively dead in storage — `isUpdatingTier` is always false after the full-page Stripe redirect, so persisting it could only ever surface stale UI for a future consumer that read it outside the `isUpdatingTier` gate.

autogpt_platform/frontend/src/app/(no-navbar)/onboarding/steps/SubscriptionStep/useSubscriptionStep.ts
```diff
@@ -62,6 +63,16 @@ export function useSubscriptionStep() {
       return;
     }
     if (isProcessing) return;
```
0ubbe, line 65: 🤖 🟡 **Nice to Have**: Two comments below (lines 81–83 and 89–92, just past this hunk) still describe the zustand store as "in-memory" and frame the pre-redirect profile POST as the *only* defense against losing data on the Stripe round-trip. After this PR, that's no longer accurate — `persist` keeps the store across the redirect.

The matching comment in `useOnboardingPage.ts` was rewritten to "defence in depth: persist keeps the store across the round-trip, but submitting pre-redirect ensures the backend has the data even if the user closes the tab during checkout." Worth aligning the language here so future readers don't see contradictory rationale at the two callsites doing the same `postV1SubmitOnboardingProfile` call.

0ubbe: 🤖 Fixed in [bf01ebea7](https://github.com/Significant-Gravitas/AutoGPT/commit/bf01ebea7bc6f8a3f701b594511d1afb6a5b46f4): consolidated the two stale comment blocks into a single "defence in depth" rationale that matches `useOnboardingPage.ts`. Now both callsites consistently say persist covers the round-trip and the pre-redirect POST is the safety net for users who close the tab mid-checkout.

autogpt_platform/frontend/src/app/(no-navbar)/onboarding/steps/SubscriptionStep/useSubscriptionStep.ts
```diff
@@ -62,23 +63,32 @@ export function useSubscriptionStep() {
...
+
+    // Local dev: backend has no Stripe wiring, so skip the checkout
+    // round-trip and advance straight to Preparing. The profile still
+    // gets POSTed there via useOnboardingPage's submission effect.
+    if (environment.isLocal()) {
+      setSelectedPlan(planKey);
+      nextStep();
+      return;
+    }
```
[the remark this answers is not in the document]
0ubbe: 🤖 Fixed in [90630d433](https://github.com/Significant-Gravitas/AutoGPT/commit/90630d433157ebb1d45e1d1e03580fcfca5b474a): moved `setIsSubmitting(true)` above the `environment.isLocal()` short-circuit so the LOCAL branch is now guarded by the same synchronous re-entry lock as the cloud path. A fast double-click in dev can no longer advance the wizard twice.

0ubbe:
🤖 Fixed in [6de865468](https://github.com/Significant-Gravitas/AutoGPT/commit/6de8654685a7640a3d1608397bd938c8c2eff56c): switched the bare `CaretLeft` import on `onboarding/page.tsx` to `CaretLeftIcon`, per the `-Icon` suffix convention noted in `frontend/AGENTS.md`. Addresses the nitpick in CodeRabbit's latest review.

PR #13003, opened 2026-05-04
fix(backend): bill Pro→Max upgrades immediately with payment failure rollback

majdyz:
## Why

Paid→paid upgrades (the canonical case is **Pro→Max**) called `stripe.Subscription.modify_async` with `proration_behavior="create_prorations"`. That writes prorated line items to the *next* invoice instead of billing the user now, so the upgrade goes through "for free" and the user is then surprise-billed alongside next month's full \$320 charge at cycle end. Worse, the DB tier flip already lands before payment is collected, so if the user's card later declines they're stuck on MAX with no charge captured.

Linear: SECRT-2315.

## What

- `modify_stripe_subscription_for_tier` upgrade branch (`backend/data/credit.py`) now calls Stripe with:
  - `proration_behavior="always_invoice"` — Stripe creates and pays the prorated invoice synchronously instead of deferring it.
  - `payment_behavior="error_if_incomplete"` — Stripe raises `stripe.CardError` (or `InvalidRequestError` when there's no default payment method) if the auto-charge fails, so the modify is rolled back and we never flip the DB tier.
- `update_subscription_tier` (`backend/api/features/v1.py`) gets a new `except stripe.CardError` branch returning **HTTP 402** with `"Your card was declined. The plan was not changed; please update your payment method and try again."`. Placed before the existing `InvalidRequestError`/`StripeError` catches so the user-facing 402 wins for declined-card failures.

The DB tier flip already runs *after* `stripe.Subscription.modify_async` — Stripe error short-circuits before `set_subscription_tier`, so failed payment ⇒ user stays on Pro.

## How (tests)

`backend/data/credit_subscription_test.py`:
- New `test_modify_stripe_subscription_for_tier_pro_to_max_bills_immediately` — Pro→Max calls Stripe with `always_invoice` + `error_if_incomplete` and the DB tier flips.
- New `test_modify_stripe_subscription_for_tier_pro_to_max_card_decline_does_not_flip_tier` — Stripe raises `CardError` ⇒ function propagates the error and `set_subscription_tier` is never awaited.
- Updated upgrade-path assertions in 4 existing tests (`modifies_existing_sub`, `clears_cancel_at_period_end_on_upgrade`, `upgrade_immediate_proration`, `upgrade_releases_pending_schedule`) to expect the new kwargs.

`backend/api/features/subscription_routes_test.py`:
- New `test_update_subscription_tier_pro_to_max_card_declined_returns_402` — POST `/credits/subscription` with `tier=MAX` where Stripe modify raises `CardError` returns HTTP 402 and `set_subscription_tier` is never awaited.

## Local verification

- `poetry run ruff format` + `poetry run ruff check` on touched files: clean.
- `pytest backend/data/credit_subscription_test.py -k "modify_stripe_subscription_for_tier or upgrade or downgrade or pro_to_max or schedule"`: **27 passed**.
- `pytest backend/api/features/subscription_routes_test.py -k "card_declined or paid_to_paid or max_checkout"`: 4 of 5 pass when run individually; the lone fail is a known pytest-asyncio event-loop scoping flake when `paid_to_paid_modifies_subscription` happens to run first in a fresh session — it passes when run alone or with my new test ordered before it. Unrelated to this change.

## Risk

- Customers without a default payment method on file (rare for paid subs) will now see a 402 instead of a silent deferred-charge upgrade. That's the correct, intended behaviour: users must have a working payment method to upgrade.
- Webhook idempotency unchanged — the existing `customer.subscription.updated` handler still reconciles after a successful modify.

### DIFF
--- a/autogpt_platform/backend/backend/api/features/subscription_routes_test.py
+++ b/autogpt_platform/backend/backend/api/features/subscription_routes_test.py
@@ -445,7 +445,7 @@ def test_update_subscription_tier_currency_mismatch_returns_422(
         side_effect=stripe.InvalidRequestError(
             "The price specified only supports `usd`. This doesn't match the"
             " expected currency: `gbp`.",
-            param="phases",
+            param="currency",
         ),
     )
 
@@ -1326,7 +1326,8 @@ def test_update_subscription_tier_pro_to_max_no_payment_method_returns_402(
         side_effect=stripe.InvalidRequestError(
             "This customer has no attached payment source or default payment"
             " method. Please consider adding a default payment method.",
-            param="customer",
+            param="default_payment_method",
+            code="resource_missing",
         ),
     )
 
--- a/autogpt_platform/backend/backend/api/features/subscription_routes_test.py
+++ b/autogpt_platform/backend/backend/api/features/subscription_routes_test.py
@@ -1211,9 +1211,13 @@ def test_update_subscription_tier_pro_to_max_card_declined_returns_402(
     client: fastapi.testclient.TestClient,
     mocker: pytest_mock.MockFixture,
 ) -> None:
-    """Pro→Max upgrade where Stripe raises CardError must return HTTP 402 and the
-    DB tier must NOT be flipped — modify_stripe_subscription_for_tier short-circuits
-    before the set_subscription_tier call inside it."""
+    """Pro→Max upgrade where Stripe raises CardError must return HTTP 402.
+
+    The "tier stays on Pro after CardError" invariant is verified at the
+    credit.py layer (see test_modify_stripe_subscription_for_tier_pro_to_max
+    _card_decline_does_not_flip_tier); this test covers the route's HTTP
+    surface only.
+    """
     mock_user = Mock()
     mock_user.subscription_tier = SubscriptionTier.PRO
 
@@ -1234,10 +1238,6 @@ def test_update_subscription_tier_pro_to_max_card_declined_returns_402(
             "Your card was declined.", param="card", code="card_declined"
         ),
     )
-    set_tier_mock = mocker.patch(
-        "backend.api.features.v1.set_subscription_tier",
-        new_callable=AsyncMock,
-    )
 
     response = client.post(
         "/credits/subscription",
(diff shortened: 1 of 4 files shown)


autogpt_platform/backend/backend/api/features/v1.py
```diff
@@ -1003,6 +1003,19 @@ async def update_subscription_tier(
...
+        logger.warning(
+            "Card declined on subscription upgrade for user %s: %s", user_id, e
+        )
+        raise HTTPException(
+            status_code=402,
+            detail=(
+                "Your card was declined. The plan was not changed; please"
+                " update your payment method and try again."
+            ),
+        )
```
majdyz, line 1033: 🤖 🟠 **Should Fix**: PR description says "Customers without a default payment method on file will now see a 402 instead of a silent deferred-charge upgrade" — but in this code path that's not what happens.

When `payment_behavior="error_if_incomplete"` runs against a customer without a default payment method, Stripe raises `stripe.InvalidRequestError`, **not** `stripe.CardError` (the SDK maps `invalid_request_error` responses to that class — see https://docs.stripe.com/api/errors). With the current ordering, that flows past the new 402 handler and lands in the existing `InvalidRequestError` branch:

1. User on PRO with `default_payment_method = None` (e.g. card was removed) clicks "Upgrade to MAX".
2. `modify_stripe_subscription_for_tier` calls `stripe.Subscription.modify_async(..., payment_behavior="error_if_incomplete")`.
3. Stripe raises `InvalidRequestError("This customer has no attached payment source or default payment method.", ...)`.
4. The currency-substring check at v1.py:1025 fails → falls through to the 502 branch (v1.py:1041) → user sees "Unable to update your subscription right now." with no actionable instruction to add a card.

Either widen the 402 handler to also detect the "no attached payment / default payment method" `InvalidRequestError` substrings (per CodeRabbit's suggestion), or add an explicit pre-check inside `modify_stripe_subscription_for_tier` before issuing the modify. As-is the Risk section in the PR description over-claims what the code does, and the only test for the no-payment-method scenario (`subscription_routes_test.py:1210`) only exercises the CardError path.

autogpt_platform/backend/backend/api/features/v1.py
```diff
@@ -1003,6 +1003,19 @@ async def update_subscription_tier(
...
+    except stripe.CardError as e:
+        # Auto-charge declined under payment_behavior=error_if_incomplete: the
+        # modify was rolled back, so 402 lets the UI prompt for a new card.
+        logger.warning(
+            "Card declined on subscription upgrade for user %s: %s", user_id, e
+        )
+        raise HTTPException(
+            status_code=402,
+            detail=(
+                "Your card was declined. The plan was not changed; please"
```
majdyz, line 1030: 🤖 🟠 **Should Fix**: `payment_behavior="error_if_incomplete"` blocks 3DS/SCA-required upgrades, and the 402 message ("Your card was declined") will mis-message the user.

Concrete failure scenario:
1. EU user on PRO has a card that triggers SCA on this charge size (any €100+ delta in the EEA, common for Pro→Max prorations).
2. `Subscription.modify_async(..., payment_behavior="error_if_incomplete")` returns the latest invoice in `requires_action` state. Stripe raises `stripe.CardError` with `code="authentication_required"` (not `card_declined`).
3. The handler at v1.py:1006 catches it and returns 402 with "Your card was declined. The plan was not changed; please update your payment method and try again."
4. User updates a perfectly good card → next attempt also requires SCA → same error → upgrade is permanently broken for SCA users from this UI.

The card isn't actually declined — Stripe needs the client to confirm the PaymentIntent on `latest_invoice.payment_intent`. Two options:
- Branch on `e.code == "authentication_required"` and return a different message ("Your bank requires extra authentication. Please complete the upgrade from the billing portal.") plus surface the `payment_intent.client_secret` so the FE can confirm in-place; or
- At minimum, soften the message to "We couldn't charge your card." and document the SCA limitation in the risk section.

As written, this PR fixes the silent-deferred-charge bug but introduces a regression for any user whose card needs SCA on the upgrade charge.

autogpt_platform/backend/backend/api/features/subscription_routes_test.py
```diff
@@ -1207,6 +1207,53 @@ async def mock_price_id(tier: SubscriptionTier) -> str | None:
...
+            "tier": "MAX",
+            "success_url": f"{TEST_FRONTEND_ORIGIN}/success",
+            "cancel_url": f"{TEST_FRONTEND_ORIGIN}/cancel",
+        },
+    )
+
+    assert response.status_code == 402
+    assert "card was declined" in response.json()["detail"].lower()
+    modify_mock.assert_awaited_once_with(TEST_USER_ID, SubscriptionTier.MAX)
+    set_tier_mock.assert_not_awaited()
```
majdyz: 🤖 🟡 **Nice to Have**: This test's `set_tier_mock.assert_not_awaited()` is vacuously true.

The `set_subscription_tier` call lives inside `modify_stripe_subscription_for_tier` (credit.py:1871), but `modify_stripe_subscription_for_tier` itself is mocked at v1.py with `side_effect=stripe.CardError(...)` (line 1230). The route never reaches *its own* `set_subscription_tier` call because there isn't one — the route only calls it inside the trial / no-active-sub branches. So the mock at v1.py:set_subscription_tier was never going to be awaited regardless of the new behaviour.

The real "Pro stays Pro after CardError" invariant is correctly verified by `test_modify_stripe_subscription_for_tier_pro_to_max_card_decline_does_not_flip_tier` in `credit_subscription_test.py`, where the Stripe call is mocked one layer deeper. The route-level test really only confirms 402 + correct message; the `set_tier_mock` assertion is theatre and slightly misleads future readers about what's being checked. Consider dropping that assertion (or, if you want to keep an "and the tier flip never reached the DB" assertion at this layer, mock `backend.data.credit.set_subscription_tier` instead so it actually has a chance of being called).

autogpt_platform/backend/backend/api/features/v1.py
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [0e8c36766](https://github.com/Significant-Gravitas/AutoGPT/commit/0e8c36766058b68cfe5700b1e6760e79ce3ef54e): mapped the no-default-payment-method `InvalidRequestError` to 402 with an actionable "add a payment method" message, before the existing currency check. Test `test_update_subscription_tier_pro_to_max_no_payment_method_returns_402` covers it. The non-currency 502 contract test still locks the "don't widen 422 beyond currency" invariant for unrelated InvalidRequestErrors.

majdyz: 🤖 Fixed in [0e8c36766](https://github.com/Significant-Gravitas/AutoGPT/commit/0e8c36766058b68cfe5700b1e6760e79ce3ef54e): InvalidRequestError handler now matches "no attached payment source" / "default payment method" / "no payment method" substrings and returns 402 with "please add a payment method and try again" instead of the generic 502. New test `test_update_subscription_tier_pro_to_max_no_payment_method_returns_402` covers the path.

autogpt_platform/backend/backend/api/features/v1.py
majdyz: 🤖 Fixed in [0e8c36766](https://github.com/Significant-Gravitas/AutoGPT/commit/0e8c36766058b68cfe5700b1e6760e79ce3ef54e): CardError handler now branches on `e.code == "authentication_required"` and returns 402 with a 3DS-specific message ("Your bank requires extra authentication for this charge…") instead of the "card was declined" copy. New test `test_update_subscription_tier_pro_to_max_authentication_required_returns_402` asserts the SCA path doesn't produce the misleading decline message. Surfacing the PaymentIntent client_secret for in-app SCA confirmation is a larger UX change (would need a new response shape + FE Stripe.js wiring) and is out of scope for this 402-rollback PR — the message now correctly tells the user to retry via the billing portal where Stripe handles SCA.

autogpt_platform/backend/backend/api/features/subscription_routes_test.py
majdyz: 🤖 Fixed in [0e8c36766](https://github.com/Significant-Gravitas/AutoGPT/commit/0e8c36766058b68cfe5700b1e6760e79ce3ef54e): dropped the vacuous `set_tier_mock.assert_not_awaited()` and updated the docstring to point at the credit.py-layer test that actually verifies the "Pro stays on Pro after CardError" invariant (`test_modify_stripe_subscription_for_tier_pro_to_max_card_decline_does_not_flip_tier`). The route-level test now just verifies the HTTP surface (402 + correct copy).

majdyz:
Replaced `str(e).lower()` substring matching with typed `e.code` / `e.param` checks in [`0a781e63da`](https://github.com/Significant-Gravitas/AutoGPT/commit/0a781e63da). `CardError` always carries `e.code` so the defensive `getattr(e, "code", None)` is gone. `InvalidRequestError` now branches on `e.code in {resource_missing, missing} and e.param in {default_payment_method, payment_method, invoice_settings.default_payment_method}` for the no-PM case and `e.param == "currency"` for the schedule currency mismatch. Tests construct the synthetic Stripe exceptions with explicit `code=` / `param=` so the typed branches actually fire under test.

PR #13012, opened 2026-05-06
refactor(frontend/copilot): split UsageLimits into UsagePopover and UsageLimitReachedCard

0ubbe:
### Why / What / How

Screenshot 2026-05-06 at 19 04 53
Screenshot 2026-05-06 at 19 05 03
Screenshot 2026-05-06 at 19 05 56

**Why:** The chat copilot's `UsageLimits/` folder mixed two distinct concerns through a shared `UsagePanelContent` abstraction with three boolean knobs (`showHeader`, `showBillingLink`, size). The shared component made the "Go to billing" CTA conditional inside the card, even though the chat container already gates whether the card renders at all — which is why the legacy "Reset daily limit" button kept flashing back through history. While doing the split it also made sense to clean up the duplicated query config across the three new hooks, align the visual language for usage bars and tier pills with the `AgentBriefingPanel`, and surface a billing entry point from the popover (and on the briefing panel itself) so users don't need to wait for the limit-reached state to manage their plan.

**What:**
- Splits `UsageLimits/` into two fully independent components — `<UsagePopover />` (chart-bar trigger button + popover panel) and `<UsageLimitReachedCard />` (alert card shown above the chat input when the limit is reached). The card always renders the "Go to billing" button (modulo the platform billing flag) and now defensively guards on actual exhaustion; ChatContainer's `useIsUsageLimitReached` decides whether to mount the card at all.
- Adds a shared `useCopilotUsage` hook so the three consumer hooks (`useUsagePopover`, `useUsageLimitReachedCard`, `useIsUsageLimitReached`) share one `USAGE_QUERY_CONFIG` constant and can't drift.
- Aligns the usage-bar visual language across the copilot popover, the limit-reached card, the credits page, and the settings/billing `AutopilotUsageCard` — single `h-2` track with blue→orange threshold at 80%, label and percent in the header row, reset/caption underneath, and the same `<Badge variant="info" size="small" className="bg-[rgb(224,237,255)]">{Tier} plan</Badge>` pill used by `AgentBriefingPanel.UsageMeter`.
- Adds an always-on `Manage billing` link to the usage popover (gated by the platform billing flag) so users can jump to `/settings/billing` without first hitting the limit.
- Replaces the conditional "Go to billing" CTA in `AgentBriefingPanel.UsageSection` with an unconditional top-right `Manage billing` link, keeping the briefing card calmer and consistent with the popover.
- Switches the popover and card integration tests from `vi.mock`-ing the data hooks to driving them via MSW handlers on `*/api/chat/usage`, so the new shared hooks are actually exercised end-to-end (closes the codecov/patch gap the bot previously flagged).

**How:**
- New layout follows CONTRIBUTING.md: `ComponentName/ComponentName.tsx` + `useComponentName.ts` + colocated `__tests__/`.
- Extracted small render primitives `UsageBar.tsx` and `StorageBar.tsx` so the popover, card, and `credits/page.tsx` can each render usage bars without inheriting the old multi-mode component. `UsageBar` no longer takes a `size` prop — the bar is `h-2` everywhere, matching the briefing panel reference.
- `useCopilotUsage` lives at the folder root; the per-consumer hooks (`useUsagePopover`, `useUsageLimitReachedCard`, `useIsUsageLimitReached`) wrap it and only add the bits each consumer needs (e.g. the platform billing flag for the popover/card).
- `useIsUsageLimitReached` lives in its own file at the folder root — used by ChatContainer, not the card itself.
- `formatBytes` moved into `usageHelpers.ts` (alongside `formatResetTime`) so the StorageBar primitive doesn't own a generic helper.
- `AutopilotUsageCard` (`/settings/billing`) keeps its framer-motion fill animation but adopts the unified `h-2` blue/orange track + `body`-variant percent label + small caption for "Spent: $X" so it visually matches the copilot bars.
- Stale tests for `UsagePanelContent` deleted (the `formatResetTime` tests were redundant with `usageHelpers.test.ts`); added focused integration tests for `UsagePopover` (9) and `UsageLimitReachedCard` (7) plus `formatBytes` table tests. The popover and card tests now drive `*/api/chat/usage` via MSW so the real hooks run during the test instead of being shimmed.
- `BriefingTabContent.UsageSection` simplified — dropped the local `showGoToBilling` derivation and the bottom CTA; the `Manage billing` link is now the only billing entry point and renders whenever `ENABLE_PLATFORM_PAYMENT` is on.

### Changes 🏗️

- **Added**
  - `UsagePopover/UsagePopover.tsx` + `useUsagePopover.ts`
  - `UsageLimitReachedCard/UsageLimitReachedCard.tsx` + `useUsageLimitReachedCard.ts`
  - `UsageBar.tsx`, `StorageBar.tsx`, `useIsUsageLimitReached.ts`, `useCopilotUsage.ts`
  - `UsagePopover/__tests__/UsagePopover.test.tsx`, `UsageLimitReachedCard/__tests__/UsageLimitReachedCard.test.tsx`
  - `formatBytes` in `usageHelpers.ts` + tests
  - "Manage billing" link in `<UsagePopover />` (always shown when `ENABLE_PLATFORM_PAYMENT` is on)
- **Removed**
  - `UsageLimits.tsx`, `UsagePanelContent.tsx`, flat `UsageLimitReachedCard.tsx`
  - `UsageLimits/__tests__/` (replaced by per-component test folders + `usageHelpers.test.ts`)
  - "Go to billing" CTA from `AgentBriefingPanel.UsageSection` (replaced by the existing top-right "Manage billing" link, which now always renders when billing is on)
- **Updated**
  - `ChatSidebar.tsx` — imports `UsagePopover`
  - `ChatContainer.tsx` + its test — split imports across new file paths
  - `profile/(user)/credits/page.tsx` — uses `UsageBar` + `StorageBar` directly instead of the deleted `UsagePanelContent`
  - `settings/billing/.../AutopilotUsageCard.tsx` — replaces the `h-6` striped gray bar with the unified `h-2` blue/orange bar
  - `library/.../AgentBriefingPanel/BriefingTabContent.tsx` — `Manage billing` link is now unconditional under the billing flag
  - Visual primitives `UsageBar.tsx` / `StorageBar.tsx` aligned to the briefing-panel reference (`h-2`, blue→orange ≥80%, header row + caption layout, `<Text>` atoms)
  - Tier pill across copilot popover, limit-reached card, and briefing panel uses the same `Badge variant="info"` with `bg-[rgb(224,237,255)]` and `"{Tier} plan"` label

#### For code changes:
- [x] I have made a test plan
  - [x] `pnpm format` clean
  - [x] `pnpm lint` clean (only pre-existing img warnings)
  - [x] `pnpm types` clean for all refactored files (pre-existing yearly-billing type errors unrelated to this PR)
  - [x] `pnpm vitest run` for all affected test files: 101/101 passing across 7 files (`UsagePopover` 9, `UsageLimitReachedCard` 7, `usageHelpers` 22, `ChatContainer` 2, `billing-cards` 42, `billing-page` 9, `BriefingTabContent` 10)
  - [ ] Manually open `/copilot`, confirm the chart-bar popover in the sidebar still shows usage bars, tier pill, storage section, and the new "Manage billing" link at the bottom
  - [ ] Manually trigger a daily-limit-reached state and confirm the alert card renders with "Go to billing" leading to `/settings/billing`, plus the defensive guard hides the card below 100%
  - [ ] Open `/profile/credits` and confirm the AutoPilot Usage & Storage section still renders bars and storage as before
  - [ ] Open `/settings/billing` and confirm the Autopilot usage card now uses the unified blue/orange `h-2` bar
  - [ ] Open `/library` and confirm the Agent Briefing usage section shows the always-on "Manage billing" link top-right (and no "Go to billing" CTA below the bars)

- [x] N/A — no config touched


autogpt_platform/frontend/src/app/(platform)/copilot/components/UsageLimits/UsagePopover/useUsagePopover.ts
```diff
@@ -0,0 +1,14 @@
+import { useGetV2GetCopilotUsage } from "@/app/api/__generated__/endpoints/chat/chat";
+import type { CoPilotUsagePublic } from "@/app/api/__generated__/models/coPilotUsagePublic";
+
+export function useUsagePopover() {
```
0ubbe, line 4: 🤖 🟠 **Should Fix**: The same `useGetV2GetCopilotUsage` config (`select` + `refetchInterval: 30000` + `staleTime: 10000`) is duplicated across three hooks: `useUsagePopover.ts`, `useUsageLimitReachedCard/useUsageLimitReachedCard.ts`, and `useIsUsageLimitReached.ts`. The pre-refactor code had a single `USAGE_QUERY_CONFIG` constant — that pattern was lost in the split.

React Query dedupes by query key so behavior is correct, but if any of these constants drifts in the future (e.g. someone tweaks `staleTime` in one place) the three hooks would silently desync.

Suggestion: extract a thin shared hook at the folder root, e.g.
```ts
// useCopilotUsage.ts
export function useCopilotUsage() {
  return useGetV2GetCopilotUsage({
    query: {
      select: (res) => res.data as CoPilotUsagePublic,
      refetchInterval: 30000,
      staleTime: 10000,
    },
  });
}
```
and have the three consumers call it. `useIsUsageLimitReached` then becomes a 4-line derivation.

autogpt_platform/frontend/src/app/(platform)/copilot/components/UsageLimits/UsageLimitReachedCard/UsageLimitReachedCard.tsx
```diff
@@ -0,0 +1,74 @@
...
+import { Text } from "@/components/atoms/Text/Text";
+import { WarningIcon } from "@phosphor-icons/react";
+import { StorageBar } from "../StorageBar";
+import { UsageBar } from "../UsageBar";
+import { useUsageLimitReachedCard } from "./useUsageLimitReachedCard";
+
+export function UsageLimitReachedCard() {
+  const { usage, isSuccess, isBillingEnabled } = useUsageLimitReachedCard();
+
+  if (!isSuccess || !usage) return null;
```
0ubbe, line 19: 🤖 🟠 **Should Fix**: This card no longer guards on whether the limit is actually reached. The pre-refactor version had `if (!isDailyExhausted && !isWeeklyExhausted) return null;`. After the split, the only gate is in `ChatContainer` (via `useIsUsageLimitReached()`), so the component name now encodes a precondition that's only enforced externally.

If anyone reuses `<UsageLimitReachedCard />` outside `ChatContainer` (or removes the gate by accident), the alert will render with the title "Usage limit reached" even when usage is at, say, 5% — the bars will just show low usage under the warning icon.

Two cheap options:
1. Re-add the defensive guard inside the card (duplicates the check, but the component becomes self-contained — guards are cheap):
   ```tsx
   const dailyDone = (usage.daily?.percent_used ?? 0) >= 100;
   const weeklyDone = (usage.weekly?.percent_used ?? 0) >= 100;
   if (!dailyDone && !weeklyDone) return null;
   ```
2. Add a test that asserts the card renders nothing when usage is below 100% — at least the contract is captured.

Option 1 is preferable because it makes misuse impossible.

autogpt_platform/frontend/src/app/(platform)/copilot/components/UsageLimits/UsageLimitReachedCard/__tests__/UsageLimitReachedCard.test.tsx
```diff
@@ -0,0 +1,117 @@
+import { render, screen, cleanup } from "@/tests/integrations/test-utils";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { UsageLimitReachedCard } from "../UsageLimitReachedCard";
+
+const mockUseUsageLimitReachedCard = vi.fn();
+vi.mock("../useUsageLimitReachedCard", () => ({
```
0ubbe: 🤖 🟠 **Should Fix**: The codecov/patch failure is being caused by the new hook files (`useUsagePopover.ts`, `useUsageLimitReachedCard.ts`, `useIsUsageLimitReached.ts`) having 0% patch coverage — every component test fully mocks the hook out:
```ts
vi.mock("../useUsageLimitReachedCard", () => ({
  useUsageLimitReachedCard: () => mockUseUsageLimitReachedCard(),
}));
```
so the production hook code is never exercised.

The project's testing convention is to mock at the network boundary with MSW, not at the hook boundary — that way the component renders, the hook runs, and the API call is intercepted (which both covers the hook and matches the existing setup in `tests/integrations/vitest.setup.tsx`).

Fix: drop the `vi.mock("../useUsageLimitReachedCard", …)` and `vi.mock("../useUsagePopover", …)` shims, drive scenarios via MSW responses for `useGetV2GetCopilotUsage` instead. Same applies to the `ChatContainer` test which mocks `useIsUsageLimitReached` directly. This both fixes codecov and removes a layer of indirection between the test and what users see.

autogpt_platform/frontend/src/app/(platform)/copilot/components/UsageLimits/UsagePopover/UsagePopover.tsx
```diff
@@ -0,0 +1,65 @@
+import { Text } from "@/components/atoms/Text/Text";
```
0ubbe, line 5: 🤖 🟡 **Nice to Have**: `UsageLimitReachedCard.tsx` starts with `"use client";` but this file does not, even though both call client-only hooks (the generated query hook + Radix popover). It works at runtime because every parent up to `ChatSidebar.tsx` is already a client component, but adding the directive here keeps the two paired components consistent and makes the file safely importable from any future server component without a confusing error.

autogpt_platform/frontend/src/app/(platform)/copilot/components/UsageLimits/UsageLimitReachedCard/__tests__/UsageLimitReachedCard.test.tsx
```diff
@@ -0,0 +1,117 @@
...
+}));
+
+vi.mock("../../StorageBar", () => ({
+  StorageBar: () => null,
+}));
+
+afterEach(() => {
+  cleanup();
+  mockUseUsageLimitReachedCard.mockReset();
+});
```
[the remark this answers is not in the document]
0ubbe: 🤖 Fixed in [24d8fa83b](https://github.com/Significant-Gravitas/AutoGPT/commit/24d8fa83be137076e0490129875eea3d09c439d1): the test was rewritten to drive the real hook via MSW (mocking `*/api/chat/usage` instead of `vi.mock`-ing the hook itself). The `cleanup()` call was dropped in the rewrite — global `vitest.setup.tsx` already handles it.

autogpt_platform/frontend/src/app/(platform)/copilot/components/UsageLimits/UsagePopover/__tests__/UsagePopover.test.tsx
```diff
@@ -0,0 +1,130 @@
...
+  ),
+  PopoverContent: ({ children }: { children: React.ReactNode }) => (
+    <div>{children}</div>
+  ),
+}));
+
+afterEach(() => {
+  cleanup();
+  mockUseUsagePopover.mockReset();
+});
```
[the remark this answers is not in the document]
0ubbe: 🤖 Fixed in [24d8fa83b](https://github.com/Significant-Gravitas/AutoGPT/commit/24d8fa83be137076e0490129875eea3d09c439d1): same as the card test — the file was rewritten to use MSW for `*/api/chat/usage`, and the redundant `cleanup()` afterEach was removed in the process.

autogpt_platform/frontend/src/app/(platform)/copilot/components/UsageLimits/UsagePopover/useUsagePopover.ts
0ubbe: 🤖 Fixed in [24d8fa83b](https://github.com/Significant-Gravitas/AutoGPT/commit/24d8fa83be137076e0490129875eea3d09c439d1): added `useCopilotUsage` (with a single `USAGE_QUERY_CONFIG` constant) at the folder root. `useUsagePopover`, `useUsageLimitReachedCard`, and `useIsUsageLimitReached` now all wrap it.

autogpt_platform/frontend/src/app/(platform)/copilot/components/UsageLimits/UsageLimitReachedCard/UsageLimitReachedCard.tsx
0ubbe: 🤖 Fixed in [24d8fa83b](https://github.com/Significant-Gravitas/AutoGPT/commit/24d8fa83be137076e0490129875eea3d09c439d1): added the defensive guard (option 1). The card now bails early when neither daily nor weekly is exhausted, so the component is self-contained and safe to mount outside ChatContainer.

autogpt_platform/frontend/src/app/(platform)/copilot/components/UsageLimits/UsageLimitReachedCard/__tests__/UsageLimitReachedCard.test.tsx
0ubbe: 🤖 Fixed in [24d8fa83b](https://github.com/Significant-Gravitas/AutoGPT/commit/24d8fa83be137076e0490129875eea3d09c439d1): both `UsagePopover.test.tsx` and `UsageLimitReachedCard.test.tsx` now use MSW handlers on `*/api/chat/usage` and let the real hooks run. The `vi.mock` shims for `useUsagePopover` / `useUsageLimitReachedCard` were dropped, so `useCopilotUsage`, `useUsagePopover`, and `useUsageLimitReachedCard` are now exercised end-to-end and codecov/patch should pick them up.

autogpt_platform/frontend/src/app/(platform)/copilot/components/UsageLimits/UsagePopover/UsagePopover.tsx
0ubbe: 🤖 Fixed in [24d8fa83b](https://github.com/Significant-Gravitas/AutoGPT/commit/24d8fa83be137076e0490129875eea3d09c439d1): added `"use client"` to `UsagePopover.tsx` to match its sibling card.

autogpt_platform/frontend/src/app/(platform)/copilot/components/UsageLimits/UsagePopover/UsagePopover.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+export function UsagePopover() {
+  const { usage, isSuccess } = useUsagePopover();
+
+  if (!isSuccess || !usage) return null;
+  if (!usage.daily && !usage.weekly) return null;
+
+  const tierLabel = usage.tier
+    ? usage.tier.charAt(0) + usage.tier.slice(1).toLowerCase()
+    : null;
```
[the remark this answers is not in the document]
0ubbe: 🤖 Fixed in [c80b9d1ed](https://github.com/Significant-Gravitas/AutoGPT/commit/c80b9d1ede7272ff588171f4a52e24c4dd4e4cfc): extracted `formatTierLabel` into `usageHelpers.ts` with an explicit `NO_TIER` guard, replaced inline duplicates in both `UsagePopover.tsx` and `UsageLimitReachedCard.tsx`, and added unit tests covering the sentinel + null/undefined/empty cases.

autogpt_platform/frontend/src/app/(platform)/copilot/components/UsageLimits/UsagePopover/__tests__/UsagePopover.test.tsx
```diff
@@ -0,0 +1,148 @@
...
+  it("renders nothing on the first paint while data is loading", () => {
+    mockUsageResponse(makeUsage({ dailyPercent: 50 }));
+    const { container } = render(<UsagePopover />);
+    expect(container.innerHTML).toBe("");
+  });
+
+  it("renders nothing when no limits are configured", async () => {
+    mockUsageResponse(makeUsage({ dailyPercent: null, weeklyPercent: null }));
+    const { container } = render(<UsagePopover />);
+    await new Promise((resolve) => setTimeout(resolve, 0));
```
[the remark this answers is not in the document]
0ubbe: 🤖 Fixed in [c80b9d1ed](https://github.com/Significant-Gravitas/AutoGPT/commit/c80b9d1ede7272ff588171f4a52e24c4dd4e4cfc): replaced `new Promise(resolve => setTimeout(resolve, 0))` with `waitFor(() => expect(container.innerHTML).toBe(""))` in both this test and the equivalent assertion in `UsageLimitReachedCard.test.tsx`.

autogpt_platform/frontend/src/app/(platform)/library/components/AgentBriefingPanel/BriefingTabContent.tsx
```diff
@@ -65,7 +62,7 @@ function UsageSection() {
           </Badge>
         )}
         <div className="flex-1" />
-        {isBillingEnabled && !showGoToBilling && (
+        {isBillingEnabled && (
           <Link
             href="/settings/billing"
             className="text-sm text-blue-600 hover:underline"
```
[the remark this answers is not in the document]
0ubbe: 🤖 Fixed in [e6ed78218](https://github.com/Significant-Gravitas/AutoGPT/commit/e6ed782187565f16d78d71f8782a35f348dc97e9): replaced the inline tier formatting with the shared `formatTierLabel` helper, so `NO_TIER` users no longer see a "No_tier plan" badge here either.

autogpt_platform/frontend/src/app/(platform)/copilot/components/usageHelpers.ts
```diff
@@ -32,3 +32,26 @@ export function formatResetTime(
     timeZoneName: "short",
   });
 }
+
+export function formatTierLabel(
+  tier: string | null | undefined,
+): string | null {
+  if (!tier || tier === "NO_TIER") return null;
+  return tier.charAt(0) + tier.slice(1).toLowerCase();
```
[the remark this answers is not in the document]
0ubbe: 🤖 Fixed in [e1710c581](https://github.com/Significant-Gravitas/AutoGPT/commit/e1710c581982c4b055a5f4513d02ed7c9eeb21c9): switched to `charAt(0).toUpperCase()` so the helper is defensive against any unexpected casing from the backend, plus added tests covering lowercase/mixed-case input.

PR #13024, opened 2026-05-07
fix(frontend): re-land proxy streaming + medium-impact auth cleanup

0ubbe:
### Why / What / How

**Why.** Re-landing the reverted [#13019](https://github.com/Significant-Gravitas/AutoGPT/pull/13019) (streaming proxy + middleware exclusion for `/api/proxy`) on top of the medium-impact follow-ups from `PROXY_FOLLOWUPS.md`, so we can validate the combined behaviour on a Vercel preview before re-merging. The original PR was reverted (`a9de6fa743`) after surfacing prod errors; this branch combines both passes so the preview environment exercises the full intended end state.

**What.**

1. Streaming proxy + middleware exclusion (from #13019, unchanged):
   - `src/middleware.ts` matcher excludes `api/proxy` so the per-request `supabase.auth.getUser()` round-trip no longer runs on every browser API call.
   - `src/app/api/proxy/[...path]/route.ts` is a single stream-through proxy: `req.body` is forwarded with `duplex: "half"`, response body is piped straight back, status/statusText preserved, hop-by-hop response headers filtered out.
   - Workspace download branch still buffers via `arrayBuffer()` (Vercel/Next.js streaming truncation bug).

2. Medium-impact auth cleanup (new):
   - **Sentinel removed.** Replace the `"no-token-found"` magic string with `null` everywhere — `getServerAuthToken`, `createRequestHeaders`, `buildForwardHeaders`, `handleWorkspaceDownload`, `transcribe/route.ts`, `workspace/files/upload/route.ts`, plus the helpers test.
   - **Per-request memoization.** Wrap `getServerSupabase` and `getServerAuthToken` in React `cache()` so a single server request reuses one Supabase client and one session fetch instead of constructing a fresh client per call.
   - **Browser-supabase consolidation.** `BackendAPI.getSupabaseClient` now uses the shared `ensureSupabaseClient()` (which is `isSingleton: true`) instead of inline `createBrowserClient`. The unused `BackendAPI.getAuthToken` is removed (httpOnly cookies make the client-side session lookup return nothing useful anyway).
   - **`useSupabase` no longer driven by `useEffect`.** One-time init runs through `useMountEffect`. Per-render context (api / router / path) is pushed into the store via a guarded `setCurrentRequestContext` setter that early-returns when values are unchanged — safe to call during render because `useSupabase` doesn't subscribe to those fields.
   - **Drop the `focus` listener.** `visibilitychange` + cross-tab `storage` already cover the cases the focus listener was firing on, and removing it stops `validateSession` from firing on every tab focus.
   - **Narrowed `revalidatePath`.** `serverLogout` now uses `revalidatePath(\"/\")` (path-only) instead of `revalidatePath(\"/\", \"layout\")`, and `refreshSession`'s redundant layout revalidation is removed entirely. The cross-tab storage handler already calls `router.refresh()` and the React Query cache is cleared client-side.

**How.** No security-model change — JWTs are still injected server-side, never exposed to the browser, Google CASA posture preserved. The backend re-validates the JWT, so middleware-level auth on `/api/proxy` was pure overhead.

### Changes 🏗️

- `src/middleware.ts`: matcher excludes `api/proxy`.
- `src/app/api/proxy/[...path]/route.ts`: streaming pass-through proxy.
- `src/app/api/transcribe/route.ts`, `src/app/api/workspace/files/upload/route.ts`: drop `\"no-token-found\"` sentinel.
- `src/lib/autogpt-server-api/client.ts`: use shared `ensureSupabaseClient`; remove dead `getAuthToken`.
- `src/lib/autogpt-server-api/helpers.ts`: `getServerAuthToken` returns `string | null`, wrapped in `cache()`; `createRequestHeaders` accepts `string | null`.
- `src/lib/autogpt-server-api/__tests__/helpers.test.ts`: assertion updated for `null` token.
- `src/lib/supabase/server/getServerSupabase.ts`: wrapped in `cache()`, dead try/catch removed.
- `src/lib/supabase/actions.ts`: narrow / remove redundant `revalidatePath`.
- `src/lib/supabase/helpers.ts`: drop `onFocus` from `setupSessionEventListeners`.
- `src/lib/supabase/hooks/useSupabase.ts`: `useMountEffect` for init, render-time `setCurrentRequestContext` for context sync.
- `src/lib/supabase/hooks/useSupabaseStore.ts`: add guarded `setCurrentRequestContext`, remove `handleFocus`.

#### For code changes:
- [x] I have made a test plan
  - [ ] **Vercel preview — primary goal of this PR.** Reproduce the prod issue that motivated the revert against the preview deploy. If it doesn't repro, we have a real fix; if it does, we localize which of the two passes (streaming proxy vs auth cleanup) reintroduces it.
  - [ ] `pnpm format`, `pnpm lint`, `pnpm types` clean
  - [ ] `pnpm test:unit` passes (existing `route.helpers.test.ts` / `helpers.test.ts` / `client.test.ts` / `custom-mutator.test.ts` suites)
  - [ ] Manual: load library, builder, copilot, monitor, settings/billing — React Query queries return data, mutations work, file uploads succeed, downloads stream
  - [ ] Manual: verify backend `Cache-Control: no-store` headers reach the browser (devtools → network)
  - [ ] Manual: verify backend 204s round-trip as 204, not 200
  - [ ] Manual: verify large file upload (>50 MB) still works and Next.js memory stays bounded
  - [ ] Manual: verify session expiry still redirects correctly on protected pages
  - [ ] Manual: verify logout still clears cross-tab session (storage event handler)
  - [ ] Manual: verify tab-focus no longer triggers a session revalidation network call (was via the removed `focus` listener)


autogpt_platform/frontend/src/app/api/proxy/[...path]/route.ts
```diff
@@ -254,39 +172,41 @@ async function handler(
...
+    console.error(`Proxy error for ${method} /${path.join("/")}:`, error);
+    return NextResponse.json(
+      {
+        error: "Proxy request failed",
+        detail: error instanceof Error ? error.message : "Unknown error",
+      },
+      { status: 502 },
     );
   }
 }
```
Comment (0ubbe) [inline autogpt_platform/frontend/src/app/api/proxy/[...path]/route.ts:223]: 🤖 🟠 **Should Fix**: This route was rewritten substantially (handler shape, header filter, body forwarding, error-mapping path), but the only handler-adjacent test is `route.helpers.test.ts` covering the workspace-download buffering branch. The new pass-through code paths — `buildForwardHeaders` (allowlist correctness, impersonation/api-key/sentry-trace forwarding), `filterResponseHeaders` (stripping content-encoding/content-length to let Next.js recompute framing), `METHODS_WITHOUT_BODY` semantics with `duplex: "half"`, status/statusText preservation for non-2xx, and the catch → 502 mapping — have no automated coverage.

Given this PR is **re-landing reverted #13019**, a Vitest integration test that mocks `fetch` (or runs through MSW) and exercises a few representative shapes (200 JSON, 204 with `Content-Type: application/json`, 4xx error body pass-through, multipart upload with `duplex: "half"`, network error → 502) would catch any regression locally instead of waiting on a Vercel preview.

autogpt_platform/frontend/src/lib/autogpt-server-api/client.ts
```diff
@@ -932,6 +911,12 @@ export default class BackendAPI {
       throw error;
     }
 
+    if (
+      response.status === 204 ||
+      response.headers.get("Content-Length") === "0"
+    ) {
+      return null;
```
0ubbe, line 918: 🤖 🟡 **Nice to Have**: Three layers now duplicate "treat 204/Content-Length:0 as null body" logic:

- `parseApiResponse` in `src/lib/autogpt-server-api/helpers.ts:235`
- `_makeClientRequest` here (`client.ts:914`)
- `getBody` in `src/app/api/mutators/custom-mutator.ts`

`_makeClientRequest` already does a dynamic `import("./helpers")` for `handleFetchError` etc., so it could call `parseApiResponse(response)` directly instead of inlining the same check. Keeps the "what counts as an empty body" rule in one place — when the next backend nuance shows up (e.g., `Content-Length: 0` on 200 with `Transfer-Encoding`, or 304 caching), only one site changes.

Not blocking; fine to defer if you'd rather keep the 204-handling change minimal for the re-land risk profile.

autogpt_platform/frontend/src/app/api/proxy/[...path]/route.ts
Comment (0ubbe) [reply autogpt_platform/frontend/src/app/api/proxy/[...path]/route.ts:223]: 🤖 Fixed in [5dd064419](https://github.com/Significant-Gravitas/AutoGPT/commit/5dd064419fb31f43ff5e963d44edaa9fcc1c2d56): added `route.test.ts` with 10 cases covering the previously untested paths — Authorization injection, status/body pass-through (200, 204 with application/json, 4xx), response-header stripping (content-encoding, content-length), POST streaming with `duplex: 'half'`, Cookie suppression, sentry-trace forwarding, no-token branch, query-string forwarding, and network-error → 502 mapping.

autogpt_platform/frontend/src/lib/autogpt-server-api/client.ts
```diff
@@ -932,6 +911,12 @@ export default class BackendAPI {
...
+    }
```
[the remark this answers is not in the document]
0ubbe: 🤖 Verified — not a real bug for this codebase. Audited every caller of `_request` (which delegates to `_makeClientRequest` on the client):

- **DELETE callers** (`deleteGraph`, `deleteGraphExecution`, `removeAgentFromLibrary`, `deletePreset`, `deleteAPIKey`, `deleteSchedule`) — all typed `Promise<void>`. They `await` and discard the result; `null` vs the previous `SyntaxError` is harmless.
- **List endpoints** (`getExecutions`, `getGraphExecutions` — Sentry's specific examples) — these go through `_get`, not `_request` directly; the backend returns `200 [...]` with an empty array for empty lists, never `204`. So the `.map()` callsite would only see `null` if the backend changed contract (in which case it's a backend bug, not a client one).
- **POST returning data** (`stopGraphExecution`) — backend always returns the execution body, never `204`.

Net effect of the change: previously `_makeClientRequest` **threw** `SyntaxError` on any 204 (because `response.json()` on an empty body throws). Now it returns `null`. For the only callers that *can* see a 204 — the void-typed DELETEs — `null` is strictly safer than a thrown exception. Resolving as false positive.

autogpt_platform/frontend/src/lib/autogpt-server-api/client.ts
0ubbe: 🤖 Deferring per the Nice-to-Have label on the original comment. The three sites diverge in body-format dispatch (`parseApiResponse` is JSON-only with `SyntaxError` recovery; `_makeClientRequest` is JSON-only without; `getBody` in custom-mutator handles JSON / PDF blob / text). Only the "is empty body" check (the `status === 204 || Content-Length === '0'` guard) is genuinely shared, and extracting just that 3-line predicate adds an indirection without meaningfully reducing duplication. Better to consolidate when the response-body taxonomy stabilises (after the streaming proxy beds in on Vercel) so the shared helper covers the realistic surface, not the current snapshot.

0ubbe:
🤖 Addressing CodeRabbit's review (4243325140):

**Outside-diff findings:**

- **`helpers.ts:117-118` — unreachable null guard.** Fixed in [55b559cd1](https://github.com/Significant-Gravitas/AutoGPT/commit/55b559cd11be75ae17ddf7ee7b31814b39c26741). Removed the `if (!supabase) throw ...` branch; `getServerSupabase` (now `cache(async () => createServerClient(...))`) always returns a client.

- **`helpers.ts:114-131` — `cache()` deduplication in Route Handlers.** Verified — partially valid concern. In Next.js 15 App Router, React's `cache()` *does* deduplicate within a single request lifetime in route handlers (Next.js scopes `cache` per-request via its server-side execution context). However, you're correct that for the route handlers we ship (`/api/proxy`, `/api/transcribe`, `/api/workspace/files/upload`, server-side `custom-mutator`), each handler invokes `getServerAuthToken()` exactly once, so the deduplication has zero practical effect there. The benefit is for any RSC paths that fan out into multiple BackendAPI calls (`useBackendAPI` server-side branch in nested server components). Keeping the wrapper because the cost is negligible and the RSC benefit is real, but the comment-clarification suggestion is fair — will fold that into a follow-up if RSC usage grows.

- **Cache wrapper broke vitest** (discovered while addressing this). React 18.3 only exposes `cache` under the `react-server` export condition; Vitest doesn't set that condition, so any module transitively importing helpers.ts crashed with `cache is not a function`. Frontend CI's `integration_test` job had been skipped due to a separate setup-step cleanup failure, so this went unnoticed. Fixed in [55b559cd1](https://github.com/Significant-Gravitas/AutoGPT/commit/55b559cd11be75ae17ddf7ee7b31814b39c26741) — added an identity shim for `cache` in `vitest.setup.tsx`.

**Nitpicks:**

- **`useSupabase.ts:18-21` — useMemo on string concat.** Fixed in [55b559cd1](https://github.com/Significant-Gravitas/AutoGPT/commit/55b559cd11be75ae17ddf7ee7b31814b39c26741). Removed per the project rule "no useMemo unless asked to optimise."

- **`useSupabase.ts:49-52` — move `setCurrentRequestContext` to `useEffect`.** Deferring. The project's global rule is "Never use `useEffect` directly in React components" — the only sanctioned escape hatch is `useMountEffect` for one-time external sync, which doesn't fit a per-navigation context push. The current pattern (render-time write with an equality guard inside `setCurrentRequestContext`) is intentional: the only writes are to refs (`routerRef`/`apiRef`/`currentPath`) that no consumer subscribes to via the store, so a discarded-render write is benign — it would just be re-asserted by the next render. Worth a deeper look if React 19's `useSyncExternalStore`-style snapshotting becomes available in our setup, but not for this PR.

autogpt_platform/frontend/src/app/api/proxy/[...path]/__tests__/route.test.ts
```diff
@@ -0,0 +1,227 @@
...
+        headers: { "Content-Type": "application/json" },
+      }),
+    );
+
+    const req = new NextRequest("https://app.test/api/proxy/api/v1/missing");
+    const res = await GET(req, makeParams(["api", "v1", "missing"]));
+
+    expect(res.status).toBe(404);
+    expect(await res.json()).toEqual({ detail: "not found" });
+  });
```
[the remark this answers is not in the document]
Comment (0ubbe) [reply autogpt_platform/frontend/src/app/api/proxy/[...path]/__tests__/route.test.ts:70]: 🤖 Fixed in [3488093c6](https://github.com/Significant-Gravitas/AutoGPT/commit/3488093c69c69fc87d4fd9dbee35faeaa0757be1): added `expect(res.statusText).toBe("Not Found")` to match the test name.

autogpt_platform/frontend/src/app/api/proxy/[...path]/__tests__/route.test.ts
```diff
@@ -0,0 +1,227 @@
...
+    );
+
+    const req = new NextRequest("https://app.test/api/proxy/api/v1/items");
+    const res = await GET(req, makeParams(["api", "v1", "items"]));
+
+    expect(res.headers.get("content-encoding")).toBeNull();
+    expect(res.headers.get("cache-control")).toBe("no-store");
+    expect(res.headers.get("x-custom-header")).toBe("preserved");
+    expect(res.headers.get("content-type")).toBe("application/json");
+  });
```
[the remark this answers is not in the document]
Comment (0ubbe) [reply autogpt_platform/frontend/src/app/api/proxy/[...path]/__tests__/route.test.ts:113]: 🤖 Fixed in [3488093c6](https://github.com/Significant-Gravitas/AutoGPT/commit/3488093c69c69fc87d4fd9dbee35faeaa0757be1): added `expect(res.headers.get("content-length")).toBeNull()` so both stripped headers are verified.

autogpt_platform/frontend/src/lib/supabase/hooks/useSupabase.ts
```diff
@@ -33,19 +31,21 @@ export function useSupabase() {
...
-    });
-  }, [api, initialize, fullPath, router]);
+  useMountEffect(() => {
+    void initialize({ api, router, path: fullPath });
+  });
+
+  // Keep the store's request context in sync with navigation. Safe to call
+  // during render because `useSupabase` doesn't subscribe to the fields that
+  // `setCurrentRequestContext` writes (routerRef / apiRef / currentPath).
+  setCurrentRequestContext({ api, router, path: fullPath });
```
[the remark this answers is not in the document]
0ubbe: 🤖 Verified — same trade-off CodeRabbit's nitpick #4 raised; intentional given the project's React rules. Brief rationale:

1. **Project constraint:** the global CLAUDE.md rule is "Never use `useEffect` directly in React components." The only sanctioned escape hatch is `useMountEffect` (one-time mount), which doesn't fit "re-sync on every navigation." Switching to `useEffect` would violate that rule.

2. **The store *needs* to hold the refs.** `handleStorageEventInternal` (cross-tab logout) and `handleVisibilityChange` (session re-validation) fire from window event listeners outside any React tree — they can't read `useRouter()` / `useBackendAPI()` directly. So `routerRef` / `apiRef` / `currentPath` must live somewhere reachable from those handlers. The remaining options are: render-time push (current), or threading them through every BackendAPI call site (much more invasive).

3. **Concurrent-mode safety analysis:** the action does `if (state.routerRef === r && state.apiRef === a && state.currentPath === p) return;` before writing. In React 18 every concurrent render either commits or is superseded by another render that re-runs `useSupabase` from scratch — there's no "writes from a discarded render persist with no follow-up." So even if a discarded render writes value X, the next render writes the correct value (and the equality guard skips the no-op case under StrictMode double-invocation).

4. **No subscribers:** the consumers of `useSupabase` don't subscribe to `routerRef` / `apiRef` / `currentPath` via the store hook (only `state.user`, `state.supabase`, `state.isUserLoading`, plus stable action references). So even if a "wrong" value briefly lands, no component re-renders off it.

The pattern is unconventional but the safest option under the constraints. Resolving as won't-fix; will revisit if React 19's `useSyncExternalStore`-based snapshotting becomes idiomatic for this case.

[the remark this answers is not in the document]
0ubbe: 🤖 Reopening — fixed in [bc178342b](https://github.com/Significant-Gravitas/AutoGPT/commit/bc178342b825055667a1304c3f056825a34c6c7b). Project owner granted an exception to the global 'no useEffect' rule for this case since the useEffect form is genuinely safer (post-commit timing, no concurrent-mode hazard) and simpler (no equality guard needed, deps array handles change-detection). The store still holds routerRef/apiRef/currentPath for the window-event handlers, but the push is now properly committed via useEffect rather than fired from render.

0ubbe:
🤖 Following up on CodeRabbit's nitpick #4 (`useSupabase.ts:49-52`, "move setCurrentRequestContext to useEffect") — fixed in [bc178342b](https://github.com/Significant-Gravitas/AutoGPT/commit/bc178342b825055667a1304c3f056825a34c6c7b). Project owner granted an exception to the no-useEffect rule for this case.

autogpt_platform/frontend/src/lib/autogpt-server-api/helpers.ts
```diff
@@ -110,32 +111,23 @@ export async function handleFetchError(response: Response): Promise<ApiError> {
   );
 }
 
-export async function getServerAuthToken(): Promise<string> {
+export const getServerAuthToken = cache(async (): Promise<string | null> => {
```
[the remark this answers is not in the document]
0ubbe: 🤖 Verified false positive — not applicable.

**`React.cache()` is request-scoped in Route Handlers, not module-level.** Per the [official Next.js docs on Sharing data with `React.cache`](https://nextjs.org/docs/app/getting-started/fetching-data#sharing-data-between-data-fetching-functions): *“React.cache is scoped strictly to the current request. Each request maintains its own memoization scope, ensuring that data is not shared between different requests or users.”*

This holds for Route Handlers as well — Next.js sets up the per-request React scope before invoking the handler (the same scope that backs `cookies()` / `headers()` from `next/headers`). The `cache()` API is the documented memoization primitive for exactly this case.

Additional defense-in-depth: `getServerSupabase()` reads cookies via `next/headers` `cookies()`, which is itself request-scoped — so even hypothetically, the cached client/token would be derived from the current request’s cookies, not leaked from another user.

No code change needed.

autogpt_platform/frontend/src/lib/supabase/hooks/useSupabase.ts
```diff
@@ -33,19 +32,23 @@ export function useSupabase() {
...
+      setCurrentRequestContext: state.setCurrentRequestContext,
       logOut: state.logOut,
       validateSession: state.validateSession,
       refreshSession: state.refreshSession,
     })),
   );
 
+  useMountEffect(() => {
+    void initialize({ api, router, path: fullPath });
+  });
```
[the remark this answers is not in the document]
0ubbe: 🤖 Verified false positive — not applicable.

**Navigation never triggered session validation in the prior code either.** Pre-PR `useSupabase` had `useEffect(() => initialize(...), [api, initialize, fullPath, router])`, so `initialize()` re-fired on every navigation — but `initialize()` early-returns when `hasLoadedUser && user` is set (see `useSupabaseStore.ts:90`). It only validates on the **first** call (or after logout clears `hasLoadedUser`). So removing the `useEffect`-driven re-call and switching to `useMountEffect` is functionally identical for navigation.

**The removed `focus` listener has no relation to navigation.** `window.focus` does not fire on internal SPA navigation — only when the *window* regains OS-level focus. That case is already covered by `visibilitychange` (still wired up in `setupSessionEventListeners`). Cross-tab sign-out / sign-in is also still covered by the `storage` listener.

**What still validates the session:**
- `handleVisibilityChange` → `validateSessionInternal()` whenever the tab becomes visible.
- `handleStorageEventInternal` → cross-tab logout/login.
- Initial mount via `initialize()` → `validateSessionHelper()` if no user is loaded.
- Backend rejects expired tokens with 401 on the next request, and React Query surfaces that.

No code change needed.

autogpt_platform/frontend/src/app/api/proxy/[...path]/route.ts
```diff
@@ -1,38 +1,100 @@
...
+  "keep-alive",
+  "proxy-authenticate",
+  "proxy-authorization",
+  "te",
+  "trailer",
+  "transfer-encoding",
+  "upgrade",
+  "content-encoding",
+  "content-length",
+]);
```
Comment (0ubbe) [inline autogpt_platform/frontend/src/app/api/proxy/[...path]/route.ts:59]: 🤖 🟠 **Should Fix**: `Set-Cookie` is not in `STRIPPED_RESPONSE_HEADERS`, so any `Set-Cookie` returned by the backend will reach the browser and be set against the **frontend** origin (cookies without an explicit `Domain` attribute attach to the response origin).

I grepped `autogpt_platform/backend` and currently no route uses `response.set_cookie` — only `agent_browser.py` references `set_cookie` for Playwright contexts, never on FastAPI responses — so today this is latent. But the previous (pre-revert) non-streaming proxy used `makeAuthenticatedRequest` + `NextResponse.json(...)` which never forwarded backend headers at all, so this is a regression vs `master` and silently broadens the trust surface: any future `response.set_cookie(...)` on a backend route would unintentionally land on the frontend domain.

Suggest adding `"set-cookie"` (and arguably `"server"`, `"x-powered-by"` for hygiene) to `STRIPPED_RESPONSE_HEADERS`. Add a route test that asserts `Set-Cookie` from the backend is dropped — symmetric with the `content-length`/`content-encoding` stripping tests.

autogpt_platform/frontend/src/lib/supabase/hooks/useSupabase.ts
0ubbe, line 44: 🤖 🟡 **Nice to Have**: This hook's lifecycle changed substantially in this PR — `useEffect`-driven init replaced with `useMountEffect` + a render-committed `setCurrentRequestContext` push, the `focus` listener was dropped, and the contract that `validateSession` only fires on first mount is now load-bearing. None of that is covered by an integration test:

```bash
$ grep -rln "useSupabase\|useSupabaseStore" --include="*.test.*" src
# (no results)
```

The `frontend/AGENTS.md` calls integration tests "the default for all new pages and features" (~90% of the test pyramid). Worth at least one test that mounts a tree using `useSupabase`, asserts:
1. `validateSessionHelper` fires once on mount (not on subsequent navigation/path changes).
2. Cross-tab logout via the `storage` event triggers `router.push(redirectPath)` even when navigation has changed `apiRef`/`routerRef`/`currentPath` since mount.
3. The equality guard in `setCurrentRequestContext` skips the no-op write when refs/path are unchanged across renders (StrictMode double-invocation safety).

Re-landing a previously reverted change without a regression test for the new lifecycle leaves the Vercel preview as the only safety net. Not blocking the re-land per se, but worth landing in this PR or as an immediate follow-up.

autogpt_platform/frontend/src/lib/supabase/actions.ts
```diff
@@ -178,7 +176,10 @@ export async function serverLogout(options: ServerLogoutOptions = {}) {
         };
       }
 
-      revalidatePath("/", "layout");
+      // Narrow path-only revalidation. The cross-tab storage listener calls
+      // router.refresh() and the React Query cache is cleared client-side,
+      // so layout-level revalidation is redundant.
+      revalidatePath("/");
```
0ubbe: 🤖 🔵 **Nit**: With React Query owning data state and `handleStorageEventInternal` calling `router.refresh()` for the originating tab, what does `revalidatePath("/")` still buy here?

- Path-only revalidation evicts the data cache for the `/` route segment, but the platform is a client-rendered SPA with React Query, so there's no Next.js fetch-cache entry for `/` to evict.
- The router cache eviction the comment alludes to is what `revalidatePath("/", "layout")` was doing — and that's specifically what this PR removes.

If this is genuinely a no-op now, dropping it makes `serverLogout` a pure auth-server call and removes the implicit "logout invalidates the home page" contract that no longer exists. Worth a quick verification (e.g., browser devtools → toggle this line → confirm logout still behaves) and a follow-up to delete if the answer is "no functional difference". Fine to defer.

autogpt_platform/frontend/src/app/api/proxy/[...path]/route.ts
Comment (0ubbe) [reply autogpt_platform/frontend/src/app/api/proxy/[...path]/route.ts:59]: 🤖 Fixed in [05ed4bc7b](https://github.com/Significant-Gravitas/AutoGPT/commit/05ed4bc7b759f41b497b7851729bff2f79aa45e4): added `set-cookie` to `STRIPPED_RESPONSE_HEADERS` and a route test asserting backend `Set-Cookie: session=leak; Path=/; HttpOnly` is dropped from the proxy response.

autogpt_platform/frontend/src/app/api/proxy/[...path]/route.ts
```diff
@@ -99,152 +161,16 @@ async function createWorkspaceDownloadErrorResponse(
...
-  path: string,
-  method: string,
-): NextResponse {
-  if (
-    error &&
-    typeof error === "object" &&
-    "status" in error &&
-    [401, 403].includes(error.status as number)
-  ) {
-    // Log this since it indicates a potential frontend bug
```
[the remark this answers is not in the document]
Comment (0ubbe) [reply autogpt_platform/frontend/src/app/api/proxy/[...path]/route.ts:209]: 🤖 Verified false positive — not applicable to the streaming branch.

**No backend route reachable through this branch returns 3xx.** I grepped `autogpt_platform/backend` for `RedirectResponse` and `status_code=30*`:

- `backend/api/features/workspace/routes.py:104` — the *only* 3xx-emitting route. Returns `302` to a GCS signed URL for file downloads.
- `backend/api/features/oauth.py` — the OAuth handlers all return JSON `{redirect_url: ...}` (a 200 with a URL field), not a 3xx response. The browser performs the navigation client-side.

**The workspace download route is special-cased before reaching `redirect: "manual"`.** Look at the handler flow:

1. `route.ts:188-190` — `if (method === "GET" && isWorkspaceDownloadRequest(path)) return await handleWorkspaceDownload(...)` matches both `api/workspace/files/{id}/download` and `api/public/shared/{token}/files/{id}/download`.
2. `handleWorkspaceDownload` calls `fetchWorkspaceDownloadWithRetry` → `fetchWorkspaceDownloadOnce` (in `route.helpers.ts`), which **does** detect 3xx (`isRedirectStatus`) and follows the `Location` URL with a fresh `fetch(..., { redirect: "follow" })`.

So the `redirect: "manual"` branch only sees non-workspace paths, none of which return 3xx today. If a future backend route ever does, the browser would receive the 3xx and either:
- Try to follow it cross-origin (CORS-rejected, fails loudly — **not** a silent bug), or
- See it as a same-origin proxy redirect (also obvious in devtools).

Both failure modes are noisy enough that they would surface in QA before prod, vs. the alternative (`redirect: "follow"`) which would silently bypass the proxy and bypass the `Authorization` injection — strictly worse from a security perspective.

Resolving as won't-fix; the existing dedicated path covers all 3xx-returning endpoints.

autogpt_platform/frontend/src/lib/supabase/hooks/useSupabase.ts
0ubbe: 🤖 Partially addressed in [4f144a159](https://github.com/Significant-Gravitas/AutoGPT/commit/4f144a1597c4ff79ca1717fdfc78660f50007b62): added `src/lib/supabase/hooks/__tests__/useSupabaseStore.test.ts` covering `setCurrentRequestContext` (5 cases):

- writes refs/path on first call,
- early-returns (no `setState`) when refs and path match current state — locks in the StrictMode-double-invocation safety guard,
- writes when any of `router`, `api`, or `path` changes (parametrised via `it.each`).

Scope of this test is intentionally narrow — the equality guard is the genuinely new contract this PR introduces, and it's the part that would silently regress if the store internals changed. The full hook-level integration tests (visibility-driven `validateSession`, cross-tab `storage` logout flow with stale-after-mount refs) are deferred to a follow-up PR scoped to the hook lifecycle, since they require rendering through `BackendAPIProvider` + a mocked `@supabase/ssr` chain and depend on the Vercel preview signal first (per the re-land workflow on this PR).

Resolving — happy to expand coverage in the follow-up PR or in a fresh review pass here if the preview surfaces something specific.

autogpt_platform/frontend/src/lib/supabase/actions.ts
0ubbe: 🤖 Fixed in [413e4ec2c](https://github.com/Significant-Gravitas/AutoGPT/commit/413e4ec2c191ac503de58ab761554fd017ebf221): removed the `revalidatePath("/")` call (and its import).

Verified before deleting: `src/app/page.tsx` is a `"use client"` spinner whose only effect is `router.replace("/copilot")` — no RSC payload, no auth-dependent server rendering, nothing for `revalidatePath` to evict in either the data cache or the Router Cache. The cross-tab `storage` listener already does `router.refresh()` for any visible page, and React Query's cache is cleared client-side, so this leaves no functional gap. Replaced with a comment explaining why none is needed.

PR #13036, opened 2026-05-07
feat(platform): add user.created_at to LaunchDarkly context

majdyz:
## Why

We want to ship the `enable-platform-payment` flag (and similar future flags) to **new signups only** without affecting any existing user. The LaunchDarkly context currently only carries `user_id`, `email`, `email_domain`, and `role` — there is no way to write a targeting rule such as "users who signed up after 2026-05-07".

Without this, retargeting to new signups requires snapshotting and uploading the full existing-user list to LD as a segment, which is brittle and stale-prone.

## What

- Backend (`backend/util/feature_flag.py`): pass the Supabase user's `created_at` into the LD context as an ISO-8601 string.
- Frontend (`services/feature-flags/feature-flag-provider.tsx`): mirror the same attribute on the client SDK so server- and client-evaluated flags resolve identically.
- Tests added in `feature_flag_test.py`: `datetime`, string, and missing `created_at` cases.

## How

`user.created_at` is already returned by the existing `supabase.auth.admin.get_user_by_id(...)` call — no extra DB lookup. We normalise it to ISO-8601 (`isoformat()` if it's a `datetime`, else `str(...)`) so LD's RFC-3339 date matchers work directly.

Once merged, an LD rule like:

```
context.created_at is after 2026-05-07T00:00:00Z → enable-platform-payment ON
context.created_at is before 2026-05-07T00:00:00Z → enable-platform-payment OFF
```

cleanly partitions cohort rollout — no segment snapshot needed.

Targeting `master` so this can be cherry-picked / hot-rolled to prod alongside the LD rule change.

## Test plan

- [x] `poetry run pytest backend/util/feature_flag_test.py` — 23/23 pass (3 new context tests)
- [x] `pnpm types` (frontend) — clean
- [x] `pnpm lint` (frontend) — no new errors
- [ ] After merge: in LD UI, confirm `created_at` appears on user contexts, then build the cutoff rule on `enable-platform-payment`

### DIFF
--- a/autogpt_platform/backend/backend/util/feature_flag.py
+++ b/autogpt_platform/backend/backend/util/feature_flag.py
@@ -140,6 +140,13 @@ async def _fetch_user_context_data(user_id: str) -> Context:
             if user.email:
                 builder.set("email", user.email)
                 builder.set("email_domain", user.email.split("@")[-1])
+            if user.created_at:
+                # ISO-8601 string — LD supports RFC3339 date targeting on
+                # this attribute (e.g. cohort users by signup window).
+                created_at = user.created_at
+                if hasattr(created_at, "isoformat"):
+                    created_at = created_at.isoformat()
+                builder.set("created_at", str(created_at))
 
     except Exception as e:
         logger.warning(f"Failed to fetch user context for {user_id}: {e}")
--- a/autogpt_platform/backend/backend/util/feature_flag.py
+++ b/autogpt_platform/backend/backend/util/feature_flag.py
@@ -143,10 +143,7 @@ async def _fetch_user_context_data(user_id: str) -> Context:
             if user.created_at:
                 # ISO-8601 string — LD supports RFC3339 date targeting on
                 # this attribute (e.g. cohort users by signup window).
-                created_at = user.created_at
-                if hasattr(created_at, "isoformat"):
-                    created_at = created_at.isoformat()
-                builder.set("created_at", str(created_at))
+                builder.set("created_at", user.created_at.isoformat())
 
     except Exception as e:
         logger.warning(f"Failed to fetch user context for {user_id}: {e}")
--- a/autogpt_platform/frontend/src/services/feature-flags/feature-flag-provider.tsx
+++ b/autogpt_platform/frontend/src/services/feature-flags/feature-flag-provider.tsx
@@ -38,6 +38,7 @@ export function LaunchDarklyProvider({ children }: { children: ReactNode }) {
         email_domain: user.email.split("@").at(-1),
       }),
       ...(user.role && { role: user.role }),
+      ...(user.created_at && { created_at: user.created_at }),
       custom: {
         ...(user.role && { role: user.role }),
       },
--- a/autogpt_platform/frontend/src/services/feature-flags/feature-flag-provider.tsx
+++ b/autogpt_platform/frontend/src/services/feature-flags/feature-flag-provider.tsx
@@ -38,6 +38,7 @@ export function LaunchDarklyProvider({ children }: { children: ReactNode }) {
         email_domain: user.email.split("@").at(-1),
       }),
       ...(user.role && { role: user.role }),
+      // Supabase JS emits `Z`-suffixed ISO; backend emits `+00:00` — LD date matchers accept both.
       ...(user.created_at && { created_at: user.created_at }),
       custom: {
         ...(user.role && { role: user.role }),
--- a/.github/workflows/platform-frontend-ci.yml
+++ b/.github/workflows/platform-frontend-ci.yml
@@ -46,8 +46,10 @@ jobs:
             components:
               - 'autogpt_platform/frontend/src/components/**'
 
-      - name: Enable corepack
-        run: corepack enable
+      - name: Set up pnpm
+        uses: pnpm/action-setup@v4
+        with:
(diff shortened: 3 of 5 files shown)


autogpt_platform/backend/backend/util/feature_flag.py
```diff
@@ -140,6 +140,13 @@ async def _fetch_user_context_data(user_id: str) -> Context:
             if user.email:
                 builder.set("email", user.email)
                 builder.set("email_domain", user.email.split("@")[-1])
+            if user.created_at:
+                # ISO-8601 string — LD supports RFC3339 date targeting on
+                # this attribute (e.g. cohort users by signup window).
+                created_at = user.created_at
+                if hasattr(created_at, "isoformat"):
```
majdyz: 🤖 🟠 **Should Fix**: `hasattr(created_at, "isoformat")` violates the backend "No duck typing" rule (`backend/AGENTS.md`: *no `hasattr`/`getattr`/`isinstance` for type dispatch*). The Supabase Python User type declares `created_at: datetime` as a non-Optional field (`supabase_auth/types.py:229`), so the runtime is always a `datetime` and the string fallback is dead code. Drop the duck-typing branch and just call `.isoformat()`:

```python
if user.created_at:
    builder.set("created_at", user.created_at.isoformat())
```

That also lets you delete `test_context_accepts_string_created_at`, which is asserting on a contract violation that the type system says cannot happen.

autogpt_platform/frontend/src/services/feature-flags/feature-flag-provider.tsx
```diff
@@ -38,6 +38,7 @@ export function LaunchDarklyProvider({ children }: { children: ReactNode }) {
         email_domain: user.email.split("@").at(-1),
       }),
       ...(user.role && { role: user.role }),
+      ...(user.created_at && { created_at: user.created_at }),
```
majdyz, line 42: 🤖 🟡 **Nice to Have**: Backend/frontend produce slightly different RFC3339 formats — backend `datetime.isoformat()` yields `2026-05-07T12:00:00+00:00`, while Supabase JS exposes `user.created_at` as-is (typically `2026-05-07T12:00:00.000Z`). LaunchDarkly date matchers accept both, but a quick comment in the PR description (or here in code) noting the parity is intentional would save the next maintainer the trip to RFC3339 land.

If you want to lock parity exactly, normalising on the backend with `user.created_at.isoformat().replace("+00:00", "Z")` matches the JS output character-for-character — but only worth it if you ever plan to use this attribute for equality matching rather than `is after / is before`.

autogpt_platform/backend/backend/util/feature_flag.py
majdyz: Fixed in ee2db6c427eaedb6f18c418a7d7432892786b099 — dropped the hasattr/string fallback, now just `user.created_at.isoformat()`, and removed the test that asserted the impossible string contract.

autogpt_platform/backend/backend/util/feature_flag.py
```diff
@@ -140,6 +140,13 @@ async def _fetch_user_context_data(user_id: str) -> Context:
...
+                    created_at = created_at.isoformat()
+                builder.set("created_at", str(created_at))
```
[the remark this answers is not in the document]
majdyz: Fixed in ee2db6c427eaedb6f18c418a7d7432892786b099 via the simpler route — `User.created_at` is non-Optional `datetime` per the Supabase types, so no normalization protocol/union is needed; just `.isoformat()` directly.

autogpt_platform/frontend/src/services/feature-flags/feature-flag-provider.tsx
majdyz: Added clarifying comment in ee2db6c427eaedb6f18c418a7d7432892786b099 noting Supabase JS emits `Z` and backend emits `+00:00`, both accepted by LD date matchers.

majdyz:
🤖 Addressed CodeRabbit's two nitpicks in [3c40b7f72](https://github.com/Significant-Gravitas/AutoGPT/commit/3c40b7f72b5c9d07875d04230ca32c0711c50444):

- `feature_flag_test.py:179-184`: `_stub_supabase` now returns the supabase mock; both tests assert `get_user_by_id.assert_called_once_with(user_id)` to pin the boundary call.
- `feature_flag_test.py:197-203`: replaced `ctx.get("created_at") is None` with `"created_at" not in ctx.custom_attributes` to make the "absent" contract explicit. (Note: `x in ctx` infinite-loops because LD's `Context` has no `__contains__` — `custom_attributes` is the supported API.)

Tests: 22/22 pass.

majdyz:
🤖 Setup CI failure root cause: `actions/setup-node@v6`'s pnpm cache mode breaks with `corepack enable` because the store path is queried before pnpm is fully shimmed (`Path Validation Error: Path(s) specified in the action for caching do(es) not exist` in Post Set up Node).

Master last passed setup at 2026-05-07 09:15 UTC; broke after that. `dev` already switched to `pnpm/action-setup@v4` before `setup-node@v6` — cherry-picked the same workflow change here in [0852237b0](https://github.com/Significant-Gravitas/AutoGPT/commit/0852237b0e2198de809e10434aa3f449ecdcf47b). Workflow files now match `dev` byte-for-byte.

PR #13039, opened 2026-05-07
feat(backend/copilot-bot): smart threads, setup-link button, mention humans, allow bot-to-bot

Bentlybro:
### Why / What / How

**Why** — Discord-UX feedback from the team chat (05/05) flagged a handful of papercuts on the bot:
- Threads were named `username × AutoPilot` — not searchable, didn't reflect what the conversation was about.
- When the bot needed the user to sign in or finish an integration setup, it dumped a wall of text with no clear next step.
- The bot couldn't `@`-mention humans back even when they'd been `@`-mentioned in the prompt — the client-level `AllowedMentions.none()` zeroed out every mention the LLM produced.
- Bot-to-bot interactions were hard-blocked at the message filter, so `@AutoPilot say hi to @Otto` did nothing.

**What** — Four user-visible improvements:
1. Threads are named after the user's first prompt, then renamed to the AI-generated session title once it finishes.
2. When `setup_required` fires mid-stream, the bot drains pending text first, then sends a button-cta that opens the AutoGPT chat at the live session for the user to finish auth there.
3. The bot resolves `@DisplayName` in its output to a real `<@id>` mention for users who were `@`-mentioned in the inbound prompt — and only those users — using a per-call `AllowedMentions` allowlist.
4. Bot-to-bot is allowed. Self-loops are still blocked at the user-id check, channels still require an `@mention`, threads still require an explicit subscription.

**How** —
- Thread naming and rename: handler builds the initial name from `ctx.text`, calls `adapter.create_thread`. After streaming finishes, it polls `BotBackend.get_session_title` up to 5×@1s and `adapter.rename_thread`s on first non-empty result. New methods on `PlatformAdapter` ABC and the Discord adapter implementation.
- Setup-required link: new `on_setup_required` callback into `BotBackend.stream_chat`. When the copilot signals a tool needs setup, the handler drains any unflushed buffer first (so the message order on Discord is **text → button**, not the reverse), then sends a link to `/copilot?sessionId=<id>` via the new `adapter.send_link` capability.
- Mention resolution: `MessageContext.mentionable_users` carries `(display_name, user_id)` pairs from the inbound `message.mentions` (excluding the bot itself). `send_message`/`send_reply` accept this allowlist. Discord adapter has a new `_resolve_mentions(text, mentionable_users)` helper that substitutes `@DisplayName` (longest-name-first, case-insensitive) to `<@id>` markup and constructs an `AllowedMentions(users=[...])` permitting only those substituted users — anything off the allowlist stays plain text, even `@everyone`.
- Bot-to-bot: removed the `if message.author.bot: return` early-exit. Self-message check still always runs. Thread history filter parallel-bumped: skip our own messages (copilot has its own transcript), keep other bots'.

### Changes 🏗️

- `bot/handler.py`: `build_thread_name`, `clamp_thread_name`, `_rename_thread_from_session_title`, `_setup_required_message`, `_copilot_session_url` helpers; `_on_setup_required` callback wired into `stream_chat`; mentionable_users threaded through every `send_message` call; cached session id decoded from bytes once.
- `bot/bot_backend.py`: new `get_session_title(session_id)` method calling the platform's session API; `on_setup_required` callback param plumbed into `stream_chat`.
- `bot/adapters/base.py`: `PlatformAdapter` ABC grew `send_link`, `rename_thread`. `MessageContext` grew `mentionable_users`. `send_message`/`send_reply` signatures gained an optional `mentionable_users` allowlist.
- `bot/adapters/discord/adapter.py`: implementations for `send_link` (button view), `rename_thread` (`channel.edit(name=...)`); `_collect_mentionable_users`, `_resolve_mentions`; on_message filter relaxed to allow other bots while keeping the self-skip.
- Tests: `TestSendLink`, `TestRenameThread`, `TestResolveMentions`, `TestCollectMentionableUsers` in `adapter_test.py`; new `bot_backend_test.py` and `handler_test.py` cases for the title rename loop, the setup-required ordering, and the decoded-session-id pass-through.

No config changes — no new env vars, no docker-compose updates, no shared-config changes.

#### For code changes:
- [x] I have made a test plan
  - [x] Mention the bot in a channel — confirms a thread is created with the prompt as the name
  - [x] Watch the thread title flip to the AI-generated session title once the response finishes
  - [x] Trigger a setup-required path (e.g. ask the bot to use an integration that needs auth) — confirm the order is **text → button**, not button-then-text
  - [x] Click the **Open AutoGPT** button — confirm it lands on the running copilot session
  - [x] In a channel, write `@AutoPilot please tell @Sue something` — confirm @Sue gets pinged in the bot's reply
  - [x] Bot replies that hallucinate `@SomeoneNotMentioned` — confirm those stay plain text (no ping)
  - [x] Try `@AutoPilot say hello to @everyone` — confirm `@everyone` does NOT ping the role
  - [x] In a server with another bot, mention `@AutoPilot @OtherBot` — confirm the bot replies (no longer hard-blocked)
  - [x] Confirm the bot still does not reply to its own messages
  - [x] Run `poetry run pytest backend/copilot/bot/` — 103 passing

### DIFF
--- a/autogpt_platform/backend/backend/copilot/bot/adapters/base.py
+++ b/autogpt_platform/backend/backend/copilot/bot/adapters/base.py
@@ -42,6 +42,11 @@ class MessageContext:
     text: str  # with bot mentions stripped
     bot_mentioned: bool = False
     thread_history: tuple[MessageHistoryEntry, ...] = ()
+    # Users the bot is allowed to @-mention back in this turn — populated
+    # from the inbound platform message's mentions (excluding the bot itself).
+    # `(display_name, platform_user_id)` pairs. Anyone not in this list won't
+    # get pinged even if the LLM produces `@theirname` in its output.
+    mentionable_users: tuple[tuple[str, str], ...] = ()
 
     @property
     def is_dm(self) -> bool:
@@ -65,7 +70,22 @@ class PlatformAdapter(ABC):
     async def stop(self) -> None: ...
 
     @abstractmethod
-    async def send_message(self, channel_id: str, text: str) -> None: ...
+    async def send_message(
+        self,
+        channel_id: str,
+        text: str,
+        mentionable_users: tuple[tuple[str, str], ...] = (),
+    ) -> None:
+        """Send `text` to `channel_id`.
+
+        `mentionable_users` is the allowlist of `(display_name, user_id)` pairs
+        the bot may ping in this message. Adapters should resolve `@DisplayName`
+        in the text to a real platform mention only for users on this list — a
+        defence against the LLM hallucinating mentions or pinging unrelated
+        users it learned about elsewhere. Default empty = render mentions as
+        plain text.
+        """
+        ...
 
     @abstractmethod
     async def send_link(
@@ -80,7 +100,11 @@ class PlatformAdapter(ABC):
 
     @abstractmethod
     async def send_reply(
-        self, channel_id: str, text: str, reply_to_message_id: str
+        self,
+        channel_id: str,
+        text: str,
+        reply_to_message_id: str,
+        mentionable_users: tuple[tuple[str, str], ...] = (),
     ) -> None: ...
 
     @abstractmethod
--- a/autogpt_platform/backend/backend/copilot/bot/adapters/base.py
+++ b/autogpt_platform/backend/backend/copilot/bot/adapters/base.py
@@ -103,6 +103,11 @@ class PlatformAdapter(ABC):
         """
         ...
 
+    @abstractmethod
+    async def rename_thread(self, thread_id: str, name: str) -> bool:
+        """Rename a platform thread/conversation when supported."""
+        ...
+
     @property
     @abstractmethod
     def max_message_length(self) -> int:
--- a/autogpt_platform/backend/backend/copilot/bot/adapters/discord/adapter.py
+++ b/autogpt_platform/backend/backend/copilot/bot/adapters/discord/adapter.py
@@ -146,6 +146,18 @@ class DiscordAdapter(PlatformAdapter):
             logger.exception("Failed to create thread in channel %s", channel_id)
             return None
 
+    async def rename_thread(self, thread_id: str, name: str) -> bool:
+        channel = await self._resolve_channel(thread_id)
+        if channel is None or not isinstance(channel, discord.Thread):
+            logger.warning("Cannot rename non-thread channel %s", thread_id)
+            return False
+        try:
+            await channel.edit(name=name[:100])
+            return True
+        except discord.HTTPException:
+            logger.exception("Failed to rename thread %s", thread_id)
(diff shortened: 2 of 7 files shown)


autogpt_platform/backend/backend/copilot/bot/adapters/base.py
```diff
@@ -65,7 +70,22 @@ async def start(self) -> None: ...
...
-    async def send_message(self, channel_id: str, text: str) -> None: ...
+    async def send_message(
+        self,
+        channel_id: str,
+        text: str,
+        mentionable_users: tuple[tuple[str, str], ...] = (),
+    ) -> None:
+        """Send `text` to `channel_id`.
+
+        `mentionable_users` is the allowlist of `(display_name, user_id)` pairs
```
ntindle, line 81: This could fuck with memory pretty heavily and make the agent go crazy low key

autogpt_platform/backend/backend/copilot/bot/adapters/discord/adapter.py
```diff
@@ -83,12 +84,18 @@ async def _resolve_channel(self, channel_id: str):
             logger.warning("Channel %s not found or inaccessible", channel_id)
             return None
 
-    async def send_message(self, channel_id: str, text: str) -> None:
+    async def send_message(
+        self,
+        channel_id: str,
+        text: str,
+        mentionable_users: tuple[tuple[str, str], ...] = (),
```
ntindle, line 91: Does a default tuple have the same issues as a default list parameter?

ntindle (approved):

autogpt_platform/backend/backend/copilot/bot/adapters/base.py
Bentlybro: `mentionable_users` is render-time only — the LLM never sees the `(display_name, user_id)` pairs. It's used inside the adapter's `send_message`/`send_reply` to substitute `@Name` → `<@id>` in outbound text after the LLM has produced it.

What the LLM sees on the inbound side is just `@DisplayName` (resolved by `_strip_mentions` from raw `<@123>` markup), so no IDs land in any agent context or thread history.

The allowlist is actually the defence against the LLM-as-mention-cannon problem you're describing: without it, anything the LLM types as `@anyone` could land as a real ping. With it, we only ping users actually present in the inbound message, so a hallucinated `@SomeRandomUser` stays plain text.

autogpt_platform/backend/backend/copilot/bot/adapters/discord/adapter.py
Bentlybro: Nope — tuples are immutable, so the shared-mutable-default gotcha doesn't apply. `def f(x=())` is safe. The bug only bites with `[]`, `{}`, `set()`, or any other mutable default.

PR #13041, opened 2026-05-07
refactor(backend): drop Anthropic auto-reroute through OpenRouter

majdyz:
## Why

`llm_call` had a transparent OpenRouter auto-reroute for Anthropic models: whenever `open_router_api_key` was configured, it silently replaced caller-supplied Anthropic credentials with the platform OpenRouter key. CodeRabbit flagged this on #13029 ([discussion r3202326734](https://github.com/Significant-Gravitas/AutoGPT/pull/13029#discussion_r3202326734)) as a privacy/compliance risk:

> When `open_router_api_key` is configured, this branch reroutes every OpenAI/Anthropic request through OpenRouter and discards the caller's direct-provider API key. That changes the actual processor handling prompt data and billing without any per-request opt-in or visibility, which is a real privacy/compliance risk for users who explicitly selected direct OpenAI/Anthropic credentials.

We addressed it for OpenAI in #13029 by reverting that branch's extension. The same concern applies to the original Anthropic-only reroute, which predates #13029 — this PR fixes that too.

## What

- Remove the `if provider == "anthropic" and or_key:` block in `llm_call` and the dependent `or_model_id` plumbing in the OpenRouter call site.
- Delete `test_anthropic_routes_through_openrouter_when_key_present` — the test pinned the now-removed behavior.

## How

The auto-reroute existed so platform USD cost would land on OpenRouter calls via `extract_openrouter_cost(usage.cost)`. Platform-internal callers no longer need it:

- `backend/executor/simulator.py` already calls OpenRouter directly via `get_openai_client(prefer_openrouter=True)` (then reads `usage.cost`).
- `backend/executor/activity_status_generator.py` does the same after #13035.

That leaves the auto-reroute firing only for **user-facing** Anthropic block calls — exactly the surface CodeRabbit raised. After this PR, those calls go to `api.anthropic.com` with the user's own Anthropic key, billed against the user's Anthropic account, as the user expects.

### Behaviour change

Anthropic blocks where the user supplied their own key now hit Anthropic directly. Before: silently rerouted via the platform OR key.

- User-visible: latency / rate limits / failure modes now reflect Anthropic, not OpenRouter. Users who were relying on the platform's OR billing for Anthropic models will start seeing direct charges on their own Anthropic key (which was always what they configured).
- Platform-side: no more OpenRouter `provider_cost` rows for user Anthropic blocks. These weren't being used for user billing — `LLM_COST` charges credits independent of upstream — so the only loss is admin-dashboard USD attribution for Anthropic block usage. That mirrors how OpenAI block usage already works (no platform USD telemetry when the user supplies their own key).

The `extract_openrouter_cost` / `extra_body={"usage":{"include":True}}` plumbing stays — both still useful when callers explicitly use OpenRouter (e.g. via `get_openai_client(prefer_openrouter=True)` in simulator/activity_status, or future provider="open_router" credentials).

## Checklist

- [x] My pull request has a descriptive title (not a vague title like `Update index.md`)
- [x] My pull request targets the `dev` branch of the repository
- [x] My commit messages follow the Conventional Commits standard
- [x] My code adheres to the project's linting and formatting standards
- [x] I have ran the pre-commit hooks before committing my changes
- [x] I removed the test pinning the deleted behaviour
- [ ] My PR is ready to review, and I have requested a review by tagging an appropriate person


majdyz:
## /pr-test: PASS

Tested locally on `refactor/drop-anthropic-openrouter-reroute@fa01ea910c` (latest, with followup test-cleanup commit). Three independent verifications all green.

### 1. `test_llm.py` full suite — 57/57 passing
```
backend/blocks/test/test_llm.py ............ 57 passed in 12.14s
```
The `test_anthropic_routes_through_openrouter_when_key_present` test was correctly removed (it asserted the now-dead reroute behaviour). The `disable_openrouter_routing` autouse fixture and `mock_settings.secrets.open_router_api_key = ""` defensive mocks were dropped in commit `fa01ea910c` — no references remain. All `TestAnthropicCacheControl` and `TestLLMStatsTracking` tests still pass without those guards, confirming the silent reroute is gone.

### 2. Headline behaviour — outbound HTTP hits `api.anthropic.com`, not `openrouter.ai`
Wrote a transport-level test that:
- Sets `settings.secrets.open_router_api_key = "sk-or-test-trigger"` (the OLD trigger for the silent reroute).
- Calls `llm_call(provider=anthropic, llm_model=CLAUDE_3_HAIKU)`.
- Patches `httpx.AsyncClient.send` and captures the request URL.

Result on PR branch:
```
OUTBOUND URLS: ['https://api.anthropic.com/v1/messages']
PASS: outbound request hit api.anthropic.com
PASS: prompt_tokens=3 completion_tokens=1
PASS: provider_cost=None (no OpenRouter usage.cost)
```

**Differential check** — same test against `origin/dev@4a39a08861` (parent commit, pre-PR):
```
ValueError: OpenRouter returned empty choices: ChatCompletion(id='msg_x', ..., model='claude-3-haiku-20240307', ...)
```
Pre-PR, the test fails because `openai.AsyncOpenAI` (configured with `base_url=OPENROUTER_BASE_URL`) was constructed instead of `anthropic.AsyncAnthropic`, and the model was prefixed `anthropic/claude-3-haiku-20240307`. Post-PR, the request goes direct to `api.anthropic.com` with the bare model id and the user's Anthropic key. Headline behaviour change verified at the wire level.

### 3. Platform-internal paths unaffected
`backend/executor/simulator.py:132` and `backend/executor/activity_status_generator.py:257` both call `get_openai_client(prefer_openrouter=True)` from `backend/util/clients.py:186`. That function is untouched by this PR (`git diff origin/dev...HEAD --name-only` shows only `llm.py` + `test_llm.py`). It builds an OpenRouter `AsyncOpenAI` client directly without ever entering `llm_call()`. Their `tracking_type="cost_usd"` rows in `PlatformCostLog` (e.g. `activity_status_generator.py:458`/`476`) cannot be affected by removing a code path they don't traverse. No runtime regression possible here by construction.

### Notes
- Skipped the full native-stack DB harness (path 2) because static analysis is sufficient: the diff doesn't touch any code those paths execute. PR #13035's harness was needed because it actually changed `activity_status_generator.py`; this PR does not.
- No fixes needed — PR is clean.
- Harness scripts cleaned up.

PR #13043, opened 2026-05-07
fix(backend): allow negative balance on post-flight reconciliation

majdyz:
## Background

`charge_reconciled_usage` runs after a block has already executed and the provider has already been paid in real $$. If the post-flight delta exceeds the user's wallet, `_add_transaction` raises `InsufficientBalanceError`; today that gets swallowed and logged as `billing_leak` (`executor/billing.py`) — the platform absorbs the cost.

This PR records the shortfall as wallet debt instead of leaking it.

## Why

We can't fail the block at this point — the work is already done and the user already received the output. The only honest options are:
1. **Leak it** (today): user pays nothing for the shortfall, platform eats the cost.
2. **Record it as debt** (this PR): wallet goes negative; the next top-up naturally clears it.

Option 2 is the standard pay-after-use accounting model. Real $$ → real bookkeeping.

## How

- `UserCreditBase.spend_credits` and `UserCredit.spend_credits` now accept `fail_insufficient_credits: bool = True` — default preserves existing behavior for every other caller.
- `_spend_credits` in `db_manager.py` forwards the kwarg through the `DatabaseManagerAsyncClient` proxy.
- `charge_reconciled_usage` passes `fail_insufficient_credits=False` so the post-flight delta lands unconditionally. The previous `except InsufficientBalanceError` arm is removed (it can no longer fire on this call site); the broader `except Exception` arm still catches genuinely unexpected errors.

## What changes for users

- **Balance stayed positive after reconciliation** — no change.
- **Balance went to zero mid-execution** — instead of getting the result for free, the wallet shows a negative balance. The next top-up (auto or manual) brings it back to positive by simple addition.
- **Auto top-up users** — `spend_credits` still triggers auto-top-up after the spend if balance < threshold; the existing low-balance notification path still fires for positive deltas.

## Companion to #13031

Independent — neither depends on the other:
- #13031 reduces the *frequency* of leaks by giving dynamic-cost blocks a non-zero pre-flight estimate (so the wallet is debited up front when it's most likely to cover).
- This PR closes the *remaining* leak surface for the cases where reconciliation still finds the wallet short.

Either can merge first.

## Test plan
- [x] poetry run pytest backend/executor/billing_reconciliation_test.py — 10 tests, including the new `test_positive_delta_passes_fail_insufficient_credits_false` that confirms the kwarg is forwarded and the wallet returns negative on a depleted balance.
- [x] Existing 9 reconciliation tests unchanged — refunds and zero-delta paths still behave correctly.
- [ ] Monitor `billing_leak` Discord alerts post-deploy: should drop to zero (or to genuinely-unexpected-exception cases only).
- [ ] Spot-check a user with a negative balance after deploy: confirm next top-up clears the debt naturally.

### DIFF
--- a/autogpt_platform/backend/backend/data/credit.py
+++ b/autogpt_platform/backend/backend/data/credit.py
@@ -144,6 +144,7 @@ class UserCreditBase(ABC):
         user_id: str,
         cost: int,
         metadata: UsageTransactionMetadata,
+        fail_insufficient_credits: bool = True,
     ) -> int:
         """
         Spend the credits for the user based on the cost.
@@ -152,6 +153,12 @@ class UserCreditBase(ABC):
             user_id (str): The user ID.
             cost (int): The cost to spend.
             metadata (UsageTransactionMetadata): The metadata of the transaction.
+            fail_insufficient_credits (bool): When True (default) raise
+                InsufficientBalanceError if the wallet can't cover the spend.
+                When False the spend is recorded unconditionally and the
+                balance may go negative — used by post-flight reconciliation
+                so a delta charge that exceeds the wallet is captured as
+                debt instead of silently leaking.
 
         Returns:
             int: The remaining balance.
@@ -673,6 +680,7 @@ class UserCredit(UserCreditBase):
         user_id: str,
         cost: int,
         metadata: UsageTransactionMetadata,
+        fail_insufficient_credits: bool = True,
     ) -> int:
         if cost == 0:
             return 0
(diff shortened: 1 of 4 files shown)


majdyz:
🤖 Addressed @coderabbitai's outside-diff finding in [7ad3e1bdd](https://github.com/Significant-Gravitas/AutoGPT/commit/7ad3e1bdd06d392382371c9e7846e57daa08ccf5): rewrote the inline comment to describe what the code actually does (`fail_insufficient_credits=False` is passed for every reconciliation spend; refunds are flag-irrelevant because `-cost > 0` short-circuits the SQL guard). No code change — the existing behavior was correct, only the comment was misleading.

PR #320, opened 2023-04-06
Code quality workflows

kreneskyp:
@Torantulino 

This PR adds common code quality workflows. Goal is to provide a consistent sandbox environment for both local and github environments.

I based this off of https://github.com/Torantulino/Auto-GPT/pull/228 for the Dockerfile fixes it provided.

## Changes

- Github workflows that are optimized for build speed
    - Image is built and cached for re-use on subsequent runs
    - Cache key is based on hash of image inputs for re-use across similar branches
- Adds makefile for local build and testing
- Adds docker-compose file for adding other services to sandbox environment
- Adds pytest, flake8, black, pyright,  isort,  

## Testing:
- manual local testing of all tools/commands
- workflows debugged in my fork. Workflows are running but test runners fail because of errors found by each of them.
    https://github.com/kreneskyp/Auto-GPT/actions/runs/4630090599
    
    image

## TODO:
- need to add mounts for sound device and test that it works on windows, linux, mac

## For follow-up
- I recommend completing this PR and then fixing errors for each test runner in a separate PR.  Turn on branch protection for for each as they reach a passing state.

- It would be worth optimizing the image to make it a bit smaller.  It will be a bit friendlier to cache limits.

PR #13049, opened 2026-05-07
fix(backend): retry-with-backoff in notification consumer before DLQ reject

majdyz:
## Why

Notification queues (`immediate_notifications_v2`, `admin_notifications_v2`, `batch_notifications_v2`, `summary_notifications_v2`) dead-letter to `failed_notifications_v2` on the **first** consumer failure. Real impact: dev's DLQ accumulated **252+ messages** that we just purged, all from transient blips — DB connection hiccups, Postmark 5xx responses, brief network outages — that would have succeeded on a second attempt seconds later.

The DLQ is the right graveyard for messages that genuinely can't be delivered. The problem was the path to it: zero retries between "first error" and "give up forever."

## What

Add explicit retry-with-backoff in `_consume_queue` before the message reaches the DLQ:

```
attempt 1: process_func()
  └─ exception → wait 2s
attempt 2: process_func()
  └─ exception → wait 4s
attempt 3: process_func()
  └─ exception → log full context → reject (no requeue) → DLQ
```

- `process_func` returning `False` still rejects immediately (callers use this for "explicitly unrecoverable — do not retry"). Only raised exceptions trigger retries; that's the existing transient-failure shape.
- Worst-case retry latency before DLQ = 14s. Short enough that genuinely broken messages don't trap the consumer for long, long enough to ride out the common transient failures observed in dev.

Constants tunable per env if needed:

```python
MAX_CONSUMER_RETRY_ATTEMPTS    = 3
CONSUMER_RETRY_BACKOFF_SECONDS = 2  # delays = 2, 4 sec
```

## How

Refactored `_consume_queue` to drop the `async with message.process()` shorthand (which only supports ack-or-reject-on-exception, no retry) in favor of explicit `message.ack()` / `message.reject(requeue=False)` calls. Retry loop sits in a new helper `_process_message_with_retry` so it's testable in isolation.

## Test plan

- [x] 5 new tests under `TestConsumerRetryWithBackoff` covering:
  - Acks on success, no retries
  - `process_func` returning False → rejects immediately, no retries
  - Transient exception then success → retries until success, acks
  - Persistent exception → exhausts all 3 attempts → rejects (DLQ)
  - Exponential backoff timing: sleep delays are exactly `[2, 4]` seconds
- Tests bypass `NotificationManager.__init__` via `__new__` so they don't need RabbitMQ / Postmark infra
- [ ] CI runs full backend test suite
- [ ] After deploy, dev's `failed_notifications_v2` should grow much more slowly. We just purged it from 252 → 0; if next-week reading is still small, the retry layer is doing its job.

## Follow-up (not in this PR)

A queue-based retry pattern (TTL "retry queue" that dead-letters back to the primary) would be more robust than in-process retry — survives consumer restarts mid-retry. Worth the upgrade if we ever see consumer crashes correlate with DLQ growth. For now, in-process is enough for the observed failure modes.

### DIFF
--- a/autogpt_platform/backend/backend/notifications/notifications.py
+++ b/autogpt_platform/backend/backend/notifications/notifications.py
@@ -698,8 +698,12 @@ class NotificationManager(AppService):
         should_send = await self._should_batch(event.user_id, event.type, event)
 
         if not should_send:
-            logger.info("Batch not old enough to send")
-            return False
+            # The notification has already been persisted into the batch in
+            # _should_batch. Ack the queue message — the next message for
+            # this user/type, or the periodic flush, will deliver the batch
+            # once it matures. Returning False here would DLQ the trigger.
+            logger.info("Batch not old enough to send (notification persisted)")
+            return True
         batch = await get_database_manager_async_client(
             should_retry=False
         ).get_user_notification_batch(event.user_id, event.type)
@@ -959,10 +963,14 @@ class NotificationManager(AppService):
         See `_process_admin_message` for the False vs. raise contract.
         """
         logger.info(f"Processing summary notification: {message}")
-        event = BaseEventModel.model_validate_json(message)
-        model = SummaryParamsEventModel[
-            get_summary_params_type(event.type)
-        ].model_validate_json(message)
+        try:
+            event = BaseEventModel.model_validate_json(message)
+            model = SummaryParamsEventModel[
+                get_summary_params_type(event.type)
+            ].model_validate_json(message)
+        except ValueError as e:
+            logger.warning(f"Unparseable summary notification (sending to DLQ): {e}")
+            return False
 
         logger.info(f"Processing summary notification: {model}")
 
@@ -1051,15 +1059,24 @@ class NotificationManager(AppService):
         contract or risk duplicate sends.
         """
         last_error: Exception | None = None
-        body = message.body.decode()
         for attempt in range(MAX_CONSUMER_RETRY_ATTEMPTS):
             try:
+                # Decoding inside the try so a UnicodeDecodeError on a
+                # malformed body is treated as a permanent failure and goes
+                # to the DLQ instead of crashing the consumer task.
+                body = message.body.decode()
                 if await process_func(body):
                     await message.ack()
                     return
                 # process_func returned False = explicit "do not retry"
                 logger.warning(
                     f"Message in {queue_name} rejected (process_func returned False)"
+                )
+                await message.reject(requeue=False)
+                return
+            except UnicodeDecodeError as e:
+                logger.warning(
+                    f"Undecodable message in {queue_name}, sending to DLQ: {e}"
                 )
                 await message.reject(requeue=False)
                 return
--- a/autogpt_platform/backend/backend/notifications/notifications.py
+++ b/autogpt_platform/backend/backend/notifications/notifications.py
@@ -47,6 +47,14 @@ from backend.util.settings import AppEnvironment, Settings
 logger = TruncatedLogger(logging.getLogger(__name__), "[NotificationManager]")
 settings = Settings()
 
+# Retry settings for transient failures in queue consumers. With these defaults
+# total worst-case retry latency before reject-to-DLQ is 2 + 4 + 8 = 14s,
+# which is enough to ride out the most common transient failures (DB
+# connection blip, Postmark 5xx, brief network outage) without trapping
+# truly broken messages on the queue for long.
+MAX_CONSUMER_RETRY_ATTEMPTS = 3
+CONSUMER_RETRY_BACKOFF_SECONDS = 2
+
 
 NOTIFICATION_EXCHANGE = Exchange(name="notifications", type=ExchangeType.TOPIC)
 DEAD_LETTER_EXCHANGE = Exchange(name="dead_letter", type=ExchangeType.TOPIC)
@@ -1007,7 +1015,13 @@ class NotificationManager(AppService):
         process_func: Callable[[str], Awaitable[bool]],
         queue_name: str,
     ):
-        """Continuously consume messages from a queue using async iteration"""
+        """Continuously consume messages from a queue using async iteration.
+
+        On transient failures (raised exceptions), retry in-process with
+        exponential backoff up to MAX_CONSUMER_RETRY_ATTEMPTS times before
+        rejecting to the DLQ. process_func returning False means
+        "unrecoverable, do not retry" — reject immediately.
+        """
         logger.info(f"Starting consumer for queue: {queue_name}")
 
         try:
@@ -1015,31 +1029,59 @@ class NotificationManager(AppService):
                 async for message in queue_iter:
                     if not self.running:
                         break
-
-                    try:
-                        async with message.process():
-                            result = await process_func(message.body.decode())
-                            if not result:
-                                # Message will be rejected when exiting context without exception
-                                raise aio_pika.exceptions.MessageProcessError(
-                                    "Processing failed"
-                                )
-                    except aio_pika.exceptions.MessageProcessError:
-                        # Let message.process() handle the rejection
-                        pass
-                    except Exception as e:
-                        logger.warning(
-                            f"Error processing message in {queue_name}: {e}",
-                            exc_info=True,
-                        )
-                        # Let message.process() handle the rejection
-                        raise
+                    await self._process_message_with_retry(
+                        message, process_func, queue_name
+                    )
         except asyncio.CancelledError:
             logger.info(f"Consumer for {queue_name} cancelled")
             raise
         except Exception as e:
             logger.exception(f"Fatal error in consumer for {queue_name}: {e}")
             raise
+
+    async def _process_message_with_retry(
+        self,
+        message: aio_pika.abc.AbstractIncomingMessage,
+        process_func: Callable[[str], Awaitable[bool]],
+        queue_name: str,
+    ):
+        """Run process_func against one message with retry-with-backoff.
+
+        Acks on success, rejects (no requeue → DLQ) on permanent failure or
+        after exhausting retry attempts.
+        """
+        last_error: Exception | None = None
+        body = message.body.decode()
+        for attempt in range(MAX_CONSUMER_RETRY_ATTEMPTS):
+            try:
+                if await process_func(body):
+                    await message.ack()
+                    return
+                # process_func returned False = explicit "do not retry"
+                logger.warning(
+                    f"Message in {queue_name} rejected (process_func returned False)"
+                )
+                await message.reject(requeue=False)
+                return
+            except Exception as e:
+                last_error = e
+                is_last_attempt = attempt == MAX_CONSUMER_RETRY_ATTEMPTS - 1
+                if is_last_attempt:
+                    break
+                delay = CONSUMER_RETRY_BACKOFF_SECONDS * (2**attempt)
+                logger.warning(
+                    f"Transient failure on attempt {attempt + 1}/"
+                    f"{MAX_CONSUMER_RETRY_ATTEMPTS} in {queue_name}: {e}. "
+                    f"Retrying in {delay}s.",
+                )
+                await asyncio.sleep(delay)
+        # All retries exhausted — log full context and send to DLQ.
+        logger.exception(
+            f"Sending message to DLQ from {queue_name} after "
+            f"{MAX_CONSUMER_RETRY_ATTEMPTS} attempts. Last error: {last_error}",
+            exc_info=last_error,
+        )
+        await message.reject(requeue=False)
 
     def run_service(self):
         # Queue the main _run_service task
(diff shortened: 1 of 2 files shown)


autogpt_platform/backend/backend/notifications/notifications.py
```diff
@@ -1007,40 +1015,74 @@ async def _consume_queue(
...
+        """Run process_func against one message with retry-with-backoff.
+
+        Acks on success, rejects (no requeue → DLQ) on permanent failure or
+        after exhausting retry attempts.
+        """
+        last_error: Exception | None = None
+        body = message.body.decode()
+        for attempt in range(MAX_CONSUMER_RETRY_ATTEMPTS):
+            try:
+                if await process_func(body):
```
majdyz, line 1068: 🤖 🟠 **Should Fix**: The retry loop calls `process_func(body)` with the **same body** on each attempt. This relies on `process_func` being idempotent — a property the existing batch-notification consumer doesn't fully satisfy at the partial-failure boundary.

Concrete failure scenario:
1. Batch consumer fetches notifications [A, B, C] from DB
2. Postmark successfully sends A
3. Consumer about to call `remove_notifications_from_batch(A)` — DB connection blips
4. Exception propagates → retry triggered
5. Next attempt: re-fetches batch, sees [A, B, C] still there (A wasn't removed), sends A **again** to Postmark

Result: duplicate email delivery on retry. Worth either:
- Documenting the idempotency requirement in the helper's docstring ("process_func MUST be idempotent — partial side effects across attempts are the caller's responsibility")
- OR adding a note that the existing batch consumer is mostly-idempotent at the per-notification level (Postmark uses a message ID for dedup if we set one), so duplicates are bounded
- OR — most defensive — don't retry on exceptions raised AFTER any side effect; the consumer would need to track "started external IO" and skip retry past that point

Cheapest fix today: docstring note. Real fix: make the batch consumer's per-notification processing transactional (Postmark send + DB remove inside the same try/atomic-ish block, so a failure either rolls back both or completes both).

autogpt_platform/backend/backend/notifications/notifications.py
```diff
@@ -47,6 +47,14 @@
 logger = TruncatedLogger(logging.getLogger(__name__), "[NotificationManager]")
 settings = Settings()
 
+# Retry settings for transient failures in queue consumers. With these defaults
+# total worst-case retry latency before reject-to-DLQ is 2 + 4 + 8 = 14s,
+# which is enough to ride out the most common transient failures (DB
+# connection blip, Postmark 5xx, brief network outage) without trapping
+# truly broken messages on the queue for long.
+MAX_CONSUMER_RETRY_ATTEMPTS = 3
```
majdyz, line 55: 🤖 🟡 **Nice to Have**: No telemetry on retry behavior. After this lands we won't know:
- How often retries actually save us (transient blip → success on attempt 2/3)
- Per-queue DLQ reject rate (still hitting DLQ despite retries → real upstream problem worth investigating)
- p99 retry latency

Two small Prometheus counters via the existing `autogpt_*` metrics setup would close the loop:

```python
NOTIFICATION_CONSUMER_RETRIES = Counter(
    'autogpt_notification_consumer_retries_total',
    'Notification consumer retry attempts after transient failure',
    ['queue', 'attempt'],
)
NOTIFICATION_CONSUMER_DLQ_REJECTS = Counter(
    'autogpt_notification_consumer_dlq_rejects_total',
    'Notification messages rejected to DLQ after all retries exhausted',
    ['queue', 'reason'],  # reason = 'exception' or 'process_returned_false'
)
```

Increment in the retry loop + on the final reject. Without this, we're tuning MAX_CONSUMER_RETRY_ATTEMPTS / backoff in the dark. Could be a follow-up PR if you'd rather keep this one focused.

autogpt_platform/backend/backend/notifications/notifications.py
```diff
@@ -47,6 +47,14 @@
 logger = TruncatedLogger(logging.getLogger(__name__), "[NotificationManager]")
 settings = Settings()
 
+# Retry settings for transient failures in queue consumers. With these defaults
+# total worst-case retry latency before reject-to-DLQ is 2 + 4 + 8 = 14s,
```
majdyz: 🤖 🔵 **Nit**: Comment says `2 + 4 + 8 = 14s` worst-case but actual delays are 2, 4 — only **2 sleeps** (between the 3 attempts). 8s never happens. Total worst-case is 6s of sleeps + however long the failing process_func takes per attempt. Update the comment to `2 + 4 = 6s of backoff` to avoid implying we wait longer than we do.

autogpt_platform/backend/backend/notifications/notifications.py
```diff
@@ -1007,40 +1015,74 @@ async def _consume_queue(
...
+            try:
+                if await process_func(body):
+                    await message.ack()
+                    return
+                # process_func returned False = explicit "do not retry"
+                logger.warning(
+                    f"Message in {queue_name} rejected (process_func returned False)"
+                )
+                await message.reject(requeue=False)
+                return
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [3f1f253259](https://github.com/Significant-Gravitas/AutoGPT/commit/3f1f2532598283b3802c8bd9f4a07697e20c8e00): you're correct — the four `_process_*` callbacks swallowed every exception with a broad try/except and returned False, which the retry loop interprets as "do not retry, DLQ now". Removed the outer try/except in `_process_admin_message`, `_process_immediate`, `_process_summary`, and the post-loop section of `_process_batch` (kept the inner per-chunk try/except in `_process_batch` because Postmark 406/422/oversized are permanent and shouldn't kill the whole batch). Permanent-decision False returns (parse fail, missing user email) preserved. Added `test_inner_process_raises_so_retry_triggers` regression test.

autogpt_platform/backend/backend/notifications/notifications.py
majdyz: 🤖 Fixed in [3f1f253259](https://github.com/Significant-Gravitas/AutoGPT/commit/3f1f2532598283b3802c8bd9f4a07697e20c8e00): added a docstring to `_process_message_with_retry` documenting the idempotency requirement and noting that the current notification consumers tolerate replay because the permanent side-effect (Postmark send) is the last step before ack.

autogpt_platform/backend/backend/notifications/notifications.py
majdyz: 🤖 Fixed in [3f1f253259](https://github.com/Significant-Gravitas/AutoGPT/commit/3f1f2532598283b3802c8bd9f4a07697e20c8e00): you're right, only 2 sleeps. Updated the comment to "3 attempts with 2s + 4s = 6s of backoff between them".

autogpt_platform/backend/backend/notifications/notifications.py
majdyz: 🤖 Acknowledged. The DLQ depth alert from PR #320 (`AutoGPT_RabbitMQ_FailedNotificationsHigh`) is the immediate signal: "DLQ rate up" is what tells us the retry loop didn't save us, regardless of whether it ran 0 or 3 times. Per-attempt counters are valuable for tuning `MAX_CONSUMER_RETRY_ATTEMPTS` and `CONSUMER_RETRY_BACKOFF_SECONDS` once we have real production data — keeping this PR scoped to the bug fix and will track the metrics work separately if DLQ rate ever indicates the defaults need tuning.

autogpt_platform/backend/backend/notifications/notifications.py
```diff
@@ -1007,40 +1015,74 @@ async def _consume_queue(
...
+        process_func: Callable[[str], Awaitable[bool]],
+        queue_name: str,
+    ):
+        """Run process_func against one message with retry-with-backoff.
+
+        Acks on success, rejects (no requeue → DLQ) on permanent failure or
+        after exhausting retry attempts.
+        """
+        last_error: Exception | None = None
+        body = message.body.decode()
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [7862dda38b](https://github.com/Significant-Gravitas/AutoGPT/commit/7862dda38b90786b03d370cd187457c1e2671f84): you're right — moved `message.body.decode()` inside the try block and added a dedicated `except UnicodeDecodeError` that treats decode failure as a permanent reject-to-DLQ (no retry — replaying the same bad bytes will keep failing). Test added: `test_undecodable_body_dlqs_without_retry`.

autogpt_platform/backend/backend/notifications/notifications.py
```diff
@@ -761,286 +819,270 @@ async def _process_batch(self, message: str) -> bool:
...
+        """Process a summary notification.
 
-            logger.info(f"Processing summary notification: {model}")
+        See `_process_admin_message` for the False vs. raise contract.
+        """
+        logger.info(f"Processing summary notification: {message}")
+        event = BaseEventModel.model_validate_json(message)
+        model = SummaryParamsEventModel[
+            get_summary_params_type(event.type)
+        ].model_validate_json(message)
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [7862dda38b](https://github.com/Significant-Gravitas/AutoGPT/commit/7862dda38b90786b03d370cd187457c1e2671f84): correct — wrapped the two `model_validate_json` calls in try/except ValueError and return False on parse failure to match the contract used by `_process_admin_message`/`_process_immediate`/`_process_batch`. Test added: `test_summary_unparseable_returns_false_not_raise`.

autogpt_platform/backend/backend/notifications/notifications.py
```diff
@@ -605,154 +613,204 @@ def _parse_message(self, message: str) -> NotificationEventModel | None:
...
-                            test_message,
-                        ) = await self.email_sender.formatter.format_email(
-                            base_template=template.base_template,
-                            subject_template=template.subject_template,
-                            content_template=template.body_template,
-                            data={"notifications": chunk},
-                            unsubscribe_link=f"{self.email_sender.formatter.env.globals.get('base_url', '')}/profile/settings",
+        if not should_send:
+            logger.info("Batch not old enough to send")
+            return False
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [7862dda38b](https://github.com/Significant-Gravitas/AutoGPT/commit/7862dda38b90786b03d370cd187457c1e2671f84): great catch — this was a real regression introduced by the contract change. The notification is already persisted in the batch table by `_should_batch` before this point, so the right behaviour is to ack the trigger and let the next message (or the periodic scheduled flush) deliver it when it matures. Returning False would have DLQ'd the trigger and silently dropped the future delivery. Test added: `test_batch_not_old_enough_acks`.

autogpt_platform/backend/backend/notifications/notifications.py
```diff
@@ -761,286 +823,283 @@ async def _process_batch(self, message: str) -> bool:
...
+        logger.info(f"Processing summary notification: {message}")
         try:
-            logger.info(f"Processing summary notification: {message}")
             event = BaseEventModel.model_validate_json(message)
             model = SummaryParamsEventModel[
                 get_summary_params_type(event.type)
             ].model_validate_json(message)
+        except ValueError as e:
+            logger.warning(f"Unparseable summary notification (sending to DLQ): {e}")
+            return False
```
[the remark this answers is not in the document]
majdyz: 🤖 False positive — `pydantic.ValidationError` IS a subclass of `ValueError` in pydantic v2. Verified locally:

```
ValidationError MRO: (ValidationError, ValueError, Exception, BaseException, object)
issubclass(ValidationError, ValueError) -> True
```

So `except ValueError` catches both the JSON-parse path (`json.JSONDecodeError` is also a `ValueError` subclass) and the schema-validation path. The behavior is the test added in this commit (`test_summary_unparseable_returns_false_not_raise`): malformed JSON → False → DLQ, no retry. No code change needed.

majdyz:
## /pr-test --fix report

### Unit tests (notification suite, native)
- 20 of 20 PASSED in 12.17s (`backend/notifications/test_notifications.py`, `poetry run pytest -xvs --no-cov`)
- Infra: docker compose up -d (postgres, redis-cluster, rabbitmq, clamav)

### Native backend startup
- All 4 notification consumers came up cleanly:
  - `Starting consumer for queue: immediate_notifications`
  - `Starting consumer for queue: admin_notifications`
  - `Starting consumer for queue: batch_notifications`
  - `Starting consumer for queue: summary_notifications`
- AsyncRabbitMQ connection acquired successfully
- NotificationManager `Started notification service` log emitted
- (Ran a minimal harness with just DatabaseManager + NotificationManager to avoid CoPilotExecutor port collision with sibling worktrees — unrelated to this PR.)

### Issues found
- None. Retry-with-backoff change does not regress consumer startup, and the unit suite (which covers `_process_immediate`, `_process_admin_message`, retry behaviour, 406 fast-path, etc.) is fully green natively.

PR #13052, opened 2026-05-08
fix(backend/copilot): persist visible marker when SDK turn ends with empty content / error

majdyz:
### Why / What / How

**Why** — SECRT-2333.  Reproducible failure (Langfuse trace `b5c67a90b076d52e05df9cd22f8e7470`, dev session `07da6f3e-a183-4315-8b5e-a1c99f3b3d29` "Dog Dunking on Dinosaur Video", user `725d64c5-3f00-4a10-9cd6-8a396d54abdb`): on Kimi K2.6 the model emitted four `browser_act` `tool_use` blocks whose tools never executed, then returned an `AssistantMessage` with `content: []`, then a `ResultMessage` with `subtype="error"` (langsmith metadata `is_error=true`).  Total cost on the failing turn was $0.249 — well under the per-query cap, so this is **not** the budget-exhaustion case fixed by #13023.  The SDK pipeline emitted `StreamError` on the wire and the SSE connection died with the request — no row was persisted to the `ChatMessage` table, so a chat reload showed the history just stopping mid-task with no resolution.

Existing guards miss this case:

* `_is_empty_completion` (`response_adapter.py:519`) requires `has_started_text=False AND _any_tool_results_seen=False` — bypassed because prior orphan-tool flushes had set `_any_tool_results_seen`.
* `pending_thinking_only_reprompt` (`response_adapter.py:443`) requires `subtype=="success"` — the failing case had `subtype="error"`.
* `STREAM_INCOMPLETE_MARKER` post-stream branch (`service.py:3359`) is gated on `not stream_completed AND not ended_with_stream_error` — the `subtype=error` ResultMessage flips `stream_completed=True` via the paired `StreamFinish`, skipping the branch.

**What** — Two-part fix that ensures the chat-history surface stays honest when an SDK turn ends with empty content / error:

1. Extend the empty-completion guard in `response_adapter.py` to also fire when an orphan `tool_use` (no real `tool_result`) preceded an empty AssistantMessage.
2. Add `STREAM_ERROR_MARKER` constant + post-stream persistence in `service.py` as a defensive backup for `ended_with_stream_error=True` paths whose on-wire `StreamError` may be dropped by client disconnect / network.

**How** —

1. `response_adapter.py`:
   - New sibling helper `_should_surface_empty_completion(msg, had_orphan_tool_use_at_result)`.  Fires on EITHER the original ghost-success case OR the new orphan-tool_use + empty-content case.
[...]

PR #13056, opened 2026-05-08
fix(backend/copilot): close empty-completion gaps + add silence heartbeat

0ubbe:
## Why

Three Discord reports in `#While testing the new payment, billing` describe AutoPilot turns that "just stop":

- **John (May 6, dev)** session `07da6f3e-a183-4315-8b5e-a1c99f3b3d29` was root-caused as SECRT-2333 and patched in #13052 (orphan `browser_act` tool_use blocks on Kimi K2.6 + SSE drop).
- **Toran (May 6, prod)** screenshot shows an expanded `Reasoning ^` panel inline with no answer text, no fold. SessionId not shared yet, but the FE shape matches a turn that finished thinking-only without ever emitting text.
- **Ubbe** intermittent — sometimes a refresh reveals "Credits limit reached".

Investigation surfaced gaps that #13052 explicitly does not close:

- **SDK**: `_should_surface_empty_completion` covers ghost-success and orphan-tool cases, but not `has_started_reasoning AND not has_started_text AND not _any_tool_results_seen AND subtype="success"` — the genuine thinking-only-no-tools path. Falls through to plain `StreamFinish`, and the FE's `splitReasoningAndResponse` then renders the reasoning inline (no fold), visually identical to in-progress streaming.
- **Baseline path** (`baseline/service.py`): only emits silent-finish fallback when `iterations >= agent_max_turns` (line 1996-2012). Natural finish with empty terminal text has no equivalent.
- **Silence between events**: `StreamStatus` is emitted at boundaries ("Contacting the model…", "Analyzing result…") but never *during* long inter-event silence. The FE rotates through generic phrases for up to several minutes before the next chunk lands.

## What

Single PR with three tightly coupled changes:

1. **SDK thinking-only-no-tools guard** — extend `_should_surface_empty_completion` in `sdk/response_adapter.py` to fire when the model emitted only `ThinkingBlock` content and ended `subtype="success"` with no tools. Surfaces as the same `StreamError(code="empty_completion") + StreamFinish` pair as the existing two cases.

2. **Baseline empty-finish fallback** — add `_NATURAL_FINISH_EMPTY_FALLBACK_TEXT` + `_natural_finish_empty_notice_text` + `_build_natural_finish_empty_fallback_events` mirroring the budget-exhausted helpers, plus a new branch in the streaming loop covering `loop_result.finished_naturally AND terminal_round_text == ""`.

3. **Silence watchdog** — new `backend/copilot/stream_heartbeat.py` with `SilenceWatchdog` async context manager + `wrap_stream_with_heartbeat` generator. Integrated in `executor/processor.py` between the raw driver stream and `stream_registry.stream_and_publish`, so both SDK and baseline paths get the heartbeat. Default schedule:

| Threshold | Message |
|-----------|---------|
| 10s | Working on it… |
| 30s | Still working — complex requests can take a moment… |
| 60s | This is taking longer than usual… |

The watchdog re-arms after every non-status event and self-suppresses for 5s after any driver-emitted `StreamStatus` so it doesn't talk over messages like "Optimizing conversation context…".

### Files changed

- `backend/copilot/sdk/response_adapter.py` — third branch in `_should_surface_empty_completion`.
- `backend/copilot/sdk/response_adapter_test.py` — 2 new tests (positive + regression).
- `backend/copilot/baseline/service.py` — new constant, helpers, and silent-finish branch.
- `backend/copilot/baseline/service_unit_test.py` — 6 new tests.
- `backend/copilot/stream_heartbeat.py` (new) — `SilenceWatchdog`, `wrap_stream_with_heartbeat`.
- `backend/copilot/stream_heartbeat_test.py` (new) — 10 tests covering threshold firing, bump reset, escalation order, one-per-gap dedupe, suppression, cleanup, error tolerance, wrapper passthrough, wrapper interleaving, wrapper suppression.
- `backend/copilot/executor/processor.py` — wires the watchdog around the driver stream.

### Out of scope

- Underlying `browser_act` "tool never executed" bug — separate tool-execution layer concern, per #13052 PR description.
- Auto-retry on `code="empty_completion"` — flagged in #5d1cdc2bad as future follow-up.
- Mid-stream credit-balance check / surfacing — orthogonal billing concern.
- FE rendering tweak so finalized-no-text-after-reasoning visually distinct from streaming — depends on UX agreement (cancel button, retry CTA, error card).
- Per-tool progress events (Approach C from the design discussion) — heavier change, deferred.

## Test plan

```
poetry run pytest backend/copilot/sdk/response_adapter_test.py     # 66 pass (incl. 2 new)
poetry run pytest backend/copilot/baseline/service_unit_test.py::TestNaturalFinishEmptyNoticeText \
                  backend/copilot/baseline/service_unit_test.py::TestBuildNaturalFinishEmptyFallbackEvents \
                  backend/copilot/baseline/service_unit_test.py::TestBudgetExhaustedNoticeText \
                  backend/copilot/baseline/service_unit_test.py::TestBuildBudgetExhaustedFallbackEvents
                                                                   # 14 pass (incl. 6 new)
poetry run pytest backend/copilot/stream_heartbeat_test.py         # 10 pass (new)
poetry run pytest backend/copilot/sdk/service_test.py::TestStreamEndedWithoutResultMessage
                                                                   # 3 pass (regression)
poetry run black --check + poetry run ruff check                   # clean
```

- [ ] Manual QA on dev: trigger a thinking-only turn (extended_thinking model, simple yes/no question that the model answers in thinking) and confirm the FE now shows an error card instead of a stuck reasoning panel
- [ ] Manual QA on dev: trigger a long tool call (e.g. video gen) and confirm escalating heartbeat messages appear at ~10s/30s/60s
- [ ] Verify Toran's session if/when sessionId is shared — does this PR cover his shape too?


autogpt_platform/backend/backend/copilot/stream_heartbeat.py
```diff
@@ -0,0 +1,215 @@
...
+            )
+
+
+async def wrap_stream_with_heartbeat(
+    stream: AsyncIterator[StreamBaseResponse],
+    *,
+    schedule: list[tuple[float, str]] | None = None,
+    suppression_window_s: float = DEFAULT_SUPPRESSION_WINDOW_S,
+    tick_s: float = DEFAULT_TICK_S,
+) -> AsyncIterator[StreamBaseResponse]:
```
0ubbe: 🤖 🔴 **Blocker**: `wrap_stream_with_heartbeat` is annotated as returning `AsyncIterator[StreamBaseResponse]`, but the function uses `yield` (so the actual runtime type is `AsyncGenerator[StreamBaseResponse, None]`). When `processor.py` passes this into `stream_and_publish(..., stream: AsyncGenerator[StreamBaseResponse, None])`, pyright fails:

```
processor.py:540:24 - error: Argument of type "AsyncIterator[StreamBaseResponse]" cannot be assigned to parameter "stream" of type "AsyncGenerator[StreamBaseResponse, None]" in function "stream_and_publish"
```

This is the failing `type-check (3.11/3.12/3.13)` CI job. Fix: tighten the return annotation to `AsyncGenerator[StreamBaseResponse, None]` and import it from `collections.abc` (the rest of the codebase already uses `AsyncGenerator` for stream types — see `stream_registry.stream_and_publish` and `stream_chat_completion_*`). That also makes `stream.aclose()` valid for downstream consumers.

autogpt_platform/backend/backend/copilot/stream_heartbeat.py
```diff
@@ -0,0 +1,215 @@
...
+                if source == "driver":
+                    if isinstance(event, StreamStatus):
+                        watchdog.note_status_emitted()
+                    else:
+                        watchdog.bump()
+                yield event
+    finally:
+        drain_task.cancel()
+        with contextlib.suppress(asyncio.CancelledError):
+            await drain_task
```
0ubbe, line 223: 🤖 🟠 **Should Fix**: When the wrapper exits early (consumer breaks, `aclose()` propagates from `stream_and_publish`'s `finally`), the `drain_task` is cancelled while suspended at `async for event in stream:` (line 189). Cancellation does **not** propagate `aclose()` to the underlying `stream` — the async generator is left for GC to finalise.

This recreates exactly the leak that `executor/processor.py:540-543` warns about:

> Explicit aclose() on early exit: `async for … break` does not close the generator, so GeneratorExit would never reach stream_chat_completion_sdk, leaving its stream lock held until GC eventually runs.

The outer `processor.py` does `await published_stream.aclose()` and `stream_and_publish`'s `finally` does `await stream.aclose()` (which lands on this wrapper) — but this wrapper does not forward the close to `raw_stream`. On every cancelled / early-exit turn the SDK stream lock stays held until GC.

Fix: in the outer `finally`, after awaiting the cancelled `drain_task`, close the underlying stream:

```python
finally:
    drain_task.cancel()
    with contextlib.suppress(asyncio.CancelledError):
        await drain_task
    if hasattr(stream, "aclose"):
        with contextlib.suppress(Exception):
            await stream.aclose()
```

(Or restructure `_drain_driver` to own the close in its own `finally`, which is the cleaner option once the return type is `AsyncGenerator` — see the blocker above.)

autogpt_platform/backend/backend/copilot/stream_heartbeat.py
0ubbe: 🤖 Fixed in [a591bd817](https://github.com/Significant-Gravitas/AutoGPT/commit/a591bd817c932cf84cf7120f1576c52e752f16ba): tightened return type to `AsyncGenerator[StreamBaseResponse, None]` (and matched the parameter + the `_delayed_stream` test helper). Pyright passes locally on the changed files.

autogpt_platform/backend/backend/copilot/stream_heartbeat.py
0ubbe: 🤖 Fixed in [a591bd817](https://github.com/Significant-Gravitas/AutoGPT/commit/a591bd817c932cf84cf7120f1576c52e752f16ba): `_drain_driver`'s `finally` now calls `await stream.aclose()` (wrapped in `contextlib.suppress(Exception)`) before posting the sentinel, so the underlying SDK stream's lock release runs deterministically on cancel / GeneratorExit instead of waiting for GC.

autogpt_platform/backend/backend/copilot/stream_heartbeat.py
```diff
@@ -0,0 +1,222 @@
...
+                silent_for = now - self._last_event_at
+                for tier_index, (threshold, message) in enumerate(self._schedule):
+                    if tier_index in self._fired_tiers:
+                        continue
+                    if silent_for < threshold:
+                        break  # schedule is sorted — later tiers are still further out
+                    await self._safe_emit(message)
+                    self._fired_tiers.add(tier_index)
+                    self._last_status_emitted_at = now
+        except asyncio.CancelledError:
```
majdyz, line 149: 🤖 🟡 **Nice to Have**: Multiple overdue tiers can fire in the same tick. The suppression check lives at the top of the outer `while True`, so when silence has exceeded 60s before any watchdog tick runs (e.g. the executor was busy), all three schedule entries emit in rapid succession inside a single inner `for` loop. The FE receives three stacked `StreamStatus` frames with no gap. Adding a `break` here (after recording `_last_status_emitted_at`) would cap each wakeup at one emission — the suppression window then naturally serialises the rest.

autogpt_platform/backend/backend/copilot/stream_heartbeat_test.py
```diff
@@ -0,0 +1,290 @@
...
+    async with SilenceWatchdog(
+        emit_status=emit,
+        schedule=schedule,
+        tick_s=0.01,
+        suppression_window_s=0.0,
+    ) as watchdog:
+        await asyncio.sleep(0.10)
+        # First emit raised — but the watchdog must still be alive.
+        assert watchdog._task is not None
+        assert not watchdog._task.done()
```
majdyz, line 290: 🤖 🟡 **Nice to Have**: No test for early consumer `aclose()` on the wrapper. The self-comment says the fix (forwarding `stream.aclose()` in `_drain_driver`'s `finally`) was applied in a591bd817, but there's no regression test locking it in. A test that breaks out of `async for` mid-stream and asserts the underlying stream got closed would prevent this from regressing silently.

autogpt_platform/backend/backend/copilot/baseline/service.py
```diff
@@ -2026,6 +2071,25 @@ def _trim_openai_on_rollback(_session_anchor: int) -> None:
             for evt in fallback_events:
                 yield evt
             state.assistant_text += fallback_text
+        elif loop_result and loop_result.finished_naturally:
+            # Natural finish under the tool budget but the terminal round
+            # produced no visible text — model returned thinking-only or
```
majdyz: 🔵 **Nit**: 5-line inline comment — project style is WHY-only one-liners. The model name and symptom attribution belong in the PR description (where they already are), not inline. Suggest collapsing to: `# mirrors SDK's empty_completion surface — prevents blank bubble on thinking-only turns`

autogpt_platform/backend/backend/copilot/stream_heartbeat.py
```diff
@@ -0,0 +1,222 @@
...
+            async for event in stream:
+                await out_queue.put(("driver", event))
+        finally:
+            # Forward close to the underlying stream so its own cleanup
+            # (e.g. SDK stream lock release) runs deterministically — the
+            # ``async for`` above does not call ``aclose()`` on cancel /
+            # GeneratorExit, leaving the lock held until GC. Mirrors the
+            # explicit aclose pattern in ``executor/processor.py``.
+            with contextlib.suppress(Exception):
+                await stream.aclose()
```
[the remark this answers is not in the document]
majdyz: 🤖 False positive. After commit [a591bd817](https://github.com/Significant-Gravitas/AutoGPT/commit/a591bd817c932cf84cf7120f1576c52e752f16ba) the `_drain_driver` finally calls `stream.aclose()` **before** `out_queue.put(sentinel)`. When the task is cancelled, `CancelledError` propagates past `suppress(Exception)` and prevents the sentinel put — but the SDK lock was already released by `aclose()` two lines earlier. The concern does not apply.

autogpt_platform/backend/backend/copilot/stream_heartbeat.py
majdyz: 🤖 Fixed in [2901d7869](https://github.com/Significant-Gravitas/AutoGPT/commit/2901d7869b9c8b5fc3c7e871b7ce5589db823671): added `break` after each tier emission in `_loop` so at most one `StreamStatus` fires per watchdog tick.

autogpt_platform/backend/backend/copilot/stream_heartbeat_test.py
majdyz: 🤖 Fixed in [2901d7869](https://github.com/Significant-Gravitas/AutoGPT/commit/2901d7869b9c8b5fc3c7e871b7ce5589db823671): added `test_wrap_underlying_stream_closed_on_early_consumer_exit` — calls `aclose()` mid-stream and asserts the underlying generator's `finally` block ran.

autogpt_platform/backend/backend/copilot/baseline/service.py
majdyz: 🤖 Fixed in [2901d7869](https://github.com/Significant-Gravitas/AutoGPT/commit/2901d7869b9c8b5fc3c7e871b7ce5589db823671): collapsed to `# mirrors SDK's empty_completion surface — prevents blank bubble on thinking-only turns`.

autogpt_platform/backend/backend/copilot/stream_heartbeat.py
```diff
@@ -0,0 +1,223 @@
...
+                    if silent_for < threshold:
+                        break  # schedule is sorted — later tiers are still further out
+                    await self._safe_emit(message)
+                    self._fired_tiers.add(tier_index)
+                    self._last_status_emitted_at = now
+                    break  # one emission per tick; suppression window paces the next tier
+        except asyncio.CancelledError:
+            raise
+        except Exception:  # noqa: BLE001 — defensive: keep loop alive on bugs in emit
+            logger.exception("[SilenceWatchdog] unexpected error in tick loop")
```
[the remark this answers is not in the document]
majdyz: 🤖 The `# noqa: BLE001` suppressors are intentional and justified here:

1. **CI lint already passes** with this code in place — the project's ruff config (`pyproject.toml`) does not enforce BLE001, so these are not violations against the actual enforced ruleset.

2. **`_loop`'s outer catch** (line ~151) is a defensive backstop that keeps the background watchdog task alive if an unexpected bug slips through. `asyncio.CancelledError` is explicitly re-raised just above it, so cancellation is never swallowed.

3. **`_safe_emit`'s catch** (line ~161) intentionally swallows transient SSE channel failures so a momentary disconnection doesn't silence the watchdog for the remainder of the turn — the next escalation tier still has a chance to fire if the channel recovers.

Both catches log the exception (`logger.exception` / `logger.warning`) so nothing is silently lost. This is the standard pattern for background asyncio tasks where crashing the entire task on any unexpected exception would be worse than continuing.

Fixed in [2901d7869](https://github.com/Significant-Gravitas/AutoGPT/commit/2901d7869b9c8b5fc3c7e871b7ce5589db823671) (the noqa comments were already present in the pushed commit).

majdyz:
## 🧪 E2E Test Report — PR #13056 (native mode)

**Branch:** `lluisagusti/investigate-llm-truncation` | **Mode:** native `poetry run app` + docker infra

### Trigger strategy (critical for heartbeat testing)

The 10s threshold makes pure-reasoning turns unreliable triggers — the model streams `reasoning-delta` chunks every 0.1–2s, continuously bumping the silence timer. The reliable **positive trigger** is a **tool-use turn** (web search) where tool execution creates >10s client-side silence with no events flowing.

| Scenario | Result | Key observation |
|---|---|---|
| Fast response ("Say hello in 3 words") | ✅ PASS | 3.2s, no heartbeat |
| Reasoning turn — max gap 2.16s | ✅ PASS | `reasoning-delta` chunks bump watchdog; correctly silent |
| **Tool-use turn (web search)** | ✅ **HEARTBEAT FIRED** | "Working on it…" at +15.11s during tool execution |
| Driver-status suppression | ✅ PASS | "Analyzing result…" after tool suppresses next watchdog tier |
| Empty/ambiguous prompt "?" | ✅ PASS | Stream finishes cleanly (no blank bubble) |
| Normal turns — no spurious events | ✅ PASS | `start/text-delta/finish` structure clean |

### Positive test trace (tool-use, 23.5s turn)
```
 1.06s  STATUS "Reading your message…"   ← driver
 4.25s  +3.17s gap → reasoning-start
                     ... tool call in flight, >10s silence ...
15.11s  STATUS "Working on it…"           ← watchdog fired ✅
16.03s  STATUS "Analyzing result…"        ← driver post-tool; arms 5s suppression window
19.92s  +3.89s gap → reasoning-start (second pass)
23.5s   finish
```

### Negative (fast, 3.2s)
No heartbeat events. Status messages: ["Reading your message…"] only. ✅

**6/6 scenarios pass. Functionally correct.**

majdyz (dismissed):
E2E tested in native mode (6/6 scenarios pass). Heartbeat fires correctly during tool-use turns ("Working on it…" at +15s) and stays silent on fast turns. Driver-status suppression window works. Empty-completion guard produces clean finish events. CI green.

majdyz (approved):

PR #13058, opened 2026-05-08
feat(billing): enable Stripe automatic tax on checkout sessions

Torantulino:
## Summary

Enables Stripe Tax automatic calculation on both the subscription upgrade and credit top-up checkout sessions.

## Changes

**File:** `autogpt_platform/backend/backend/data/credit.py`

Added three parameters to both `stripe.checkout.Session.create` calls:

```python
automatic_tax={enabled: True},
billing_address_collection=auto,
customer_update={address: auto},
```

**Affected functions:**
- `create_subscription_checkout()` — "Upgrade to Max" / plan upgrade flow
- `top_up_intent()` — "Add credits" top-up flow

## Why each parameter

| Parameter | Reason |
|---|---|
| `automatic_tax` | Activates Stripe Tax calculation at checkout |
| `billing_address_collection="auto"` | Collects country + postcode (minimum required for tax rate lookup) |
| `customer_update={"address": "auto"}` | Saves address back to Stripe Customer so repeat checkouts don't re-prompt |

## Prerequisites (Stripe Dashboard — before deploying)

1. Enable Stripe Tax: Dashboard → Tax → Enable
2. Set a default tax code on each product (e.g. `txcd_10000001` for SaaS)
3. Set default tax behavior (`exclusive` or `inclusive`)
4. Register at least one tax nexus jurisdiction

## Testing

- Checkout in test mode with a US address (e.g. California) → should show sales tax line item
- Checkout with a UK address → should show VAT
- Returning customer with saved address → should not be re-prompted for address

### DIFF
--- a/autogpt_platform/backend/backend/data/credit.py
+++ b/autogpt_platform/backend/backend/data/credit.py
@@ -1087,6 +1087,9 @@ class UserCredit(UserCreditBase):
             success_url=base_url + "/settings/billing?topup=success",
             cancel_url=base_url + "/settings/billing?topup=cancel",
             allow_promotion_codes=True,
+            automatic_tax={"enabled": True},
+            billing_address_collection="auto",
+            customer_update={"address": "auto"},
         )
 
         await self._add_transaction(
@@ -2274,6 +2277,9 @@ async def create_subscription_checkout(
             }
         },
         allow_promotion_codes=True,
+        automatic_tax={"enabled": True},
+        billing_address_collection="auto",
+        customer_update={"address": "auto"},
     )
     if not session.url:
         # An empty checkout URL for a paid upgrade is always an error; surfacing it


majdyz (approved):

autogpt_platform/backend/backend/data/credit.py
```diff
@@ -1087,6 +1087,9 @@ async def top_up_intent(self, user_id: str, amount: int) -> str:
             success_url=base_url + "/settings/billing?topup=success",
             cancel_url=base_url + "/settings/billing?topup=cancel",
             allow_promotion_codes=True,
+            automatic_tax={"enabled": True},
```
majdyz, line 1090: 🤖 🟡 **Nice to Have** _(updated after verifying Dashboard config)_: Live Tax is enabled, GB registration in place, account-level default `tax_code=txcd_10000000` covers products without explicit codes — so the rollout-ordering concern doesn't apply. Verified `get_stripe_customer_id` always returns a valid string ([credit.py:1359-1382](https://github.com/Significant-Gravitas/AutoGPT/blob/094b27954ce34ac7fee82f2238da36f0d3b1976e/autogpt_platform/backend/backend/data/credit.py#L1359-L1382)) so `customer_update` won't trip Stripe-side validation either. Safe to merge.

Only residual risk is a Stripe Tax service outage (Tax down, Stripe core up) → checkout 500. The rolled-back try/except from c0bff73 would catch that as defense-in-depth. Worth restoring in a follow-up but not a merge blocker.

Test mode is currently registered for 0 jurisdictions, so the PR's QA steps need updating: the "US address → sales tax line" check will show $0 (US has no nexus). Mirror live in test mode by adding GB only, then update QA to: GB → 20% VAT, US → $0 tax line, returning GB customer → no address re-prompt.

autogpt_platform/backend/backend/data/credit.py
```diff
@@ -1087,6 +1087,9 @@ async def top_up_intent(self, user_id: str, amount: int) -> str:
...
+            billing_address_collection="auto",
+            customer_update={"address": "auto"},
```
majdyz, line 1092: 🤖 ✅ _Moot after Dashboard verification._ Account-level default `tax_code=txcd_10000000` is set in live Stripe Tax settings, which covers both the canonical `STRIPE_PRODUCT_ID_TOPUP` Product *and* the inline `product_data` ephemeral product on this fallback path. Per-product `tax_code` not needed. Resolving.

PR #13086, opened 2026-05-12
fix(backend/copilot): mandate find_block before refusal + suppress false empty-completion on retry

majdyz:
## Why

Two production-visible regressions surfaced in the same dev session
(`d077d025-f6a3-455f-836f-9fe3039b2b98`):

1. **AutoPilot refused without searching.** User asked "Post a test post to
   linkedin for me". The model emitted *"We don't have a native LinkedIn
   integration yet. LinkedIn's API is also heavily restricted…"* with **zero
   tool calls** — never invoked `find_block`. When the user pushed back, the
   model immediately found `PostToLinkedInBlock` (Ayrshare) on the first
   `find_block(query="linkedin post")`. The model's own thinking from
   langfuse: *"The user is correct — I jumped straight to capability gap
   without first searching for blocks. Per my instructions, I should always
   `find_block` first for any integration request."* — i.e. the rule existed
   but was overridden by a vivid worked example in the system prompt.

2. **"Model returned an empty response" overlay on top of working output.**
   In the same session's recovery turn, after `find_block` + `run_block`
   succeeded and the model emitted *"Yes, we have a `PostToLinkedInBlock`
   via Ayrshare — and you're already connected…"*, a retry-adapter
   recreated mid-turn lost the prior attempt's `has_started_text` /
   `_any_tool_results_seen` state and false-fired
   `code=empty_completion` on the next ResultMessage.

Root causes:

- The Langfuse "CoPilot Prompt" Section 4 ("Capability Check") + Example 3
  (Shopify) explicitly modelled "We don't have a native [Platform]
  integration yet" as the *correct* response when the model "knows" the
  platform isn't supported — training the model to refuse without
  searching. The shared tool-notes supplement said "find_block first" but
  the worked example won.
- `SDKResponseAdapter` is rebuilt fresh on context-reduce retries
  (service.py:4510). Its empty-completion guard relies on per-attempt
  state, so a benign empty trailing ResultMessage on a retry — when the
  prior attempt already streamed text/tools to the wire — is
  indistinguishable from a true SECRT-2252 ghost finish and surfaces a
  spurious error overlay.

## What

Code-side fixes (Langfuse prompt is also patched on the `latest`/dev
label out-of-band; prod stays on its existing version pending review):

- **`prompting.py`** — rewrite `SHARED_TOOL_NOTES` "Tool Discovery
  Priority" to mark `find_block` as MANDATORY before any "no integration"
  reply. Add an explicit "Anti-pattern: refusing without searching"
  section that enumerates the exact phrases (`We don't have a native X
  integration yet.`, `There's no block for X.`, etc.) and shows the
  correct 3-step flow.
- **`response_adapter.py`** — add
  `prior_attempt_emitted_visible_content: bool = False` to
  `SDKResponseAdapter`; `_is_empty_completion` returns False when the
  flag is set, so the SECRT-2252 guard only fires when the *broader
  stream* has produced no content, not just the current attempt.
- **`service.py`** retry path — before swapping the adapter on
  `_reduce_context` retry, set the new adapter's
  `prior_attempt_emitted_visible_content` based on the prior adapter's
  `has_started_text` / `has_started_reasoning` /
  `_any_tool_results_seen` (and forward the flag transitively across
  multi-retry chains).
- **Tests** — `prompting_test.py::TestToolDiscoveryPriorityAntiPattern`
  (3 assertions on the new mandatory language + anti-pattern phrases +
  correct-flow template); `response_adapter_test.py` adds
  `test_retry_adapter_with_prior_emitted_content_suppresses_empty_completion`
  (positive: empty success on retry → StreamFinish, no StreamError) and
  `test_retry_adapter_without_prior_content_still_surfaces_empty_completion`
  (counter-test: SECRT-2252 still fires when the flag is False).

## How

The three fixes layer:

1. **System-prompt rule** (defence in depth) — code-side supplement now
   explicitly forbids the failure-mode phrases by exact wording; the
   model can pattern-match its own draft and reject it.
2. **Langfuse worked example** (root cause) — Section 4 + Example 3 of
   "CoPilot Prompt" rewritten on `latest` v30 to model search-first
   behaviour. Prod (`production` v27) untouched until reviewed.
3. **Adapter false-positive guard** — surface "model returned an empty
   response" only when the *whole stream* produced nothing, not just the
   current retry attempt.

## Test plan

- [x] `pytest backend/copilot/sdk/response_adapter_test.py -k "empty_completion or prior_attempt or retry_adapter"` — 3 passed.
- [x] `pytest backend/copilot/sdk/response_adapter_test.py::test_retry_adapter_with_prior_emitted_content_suppresses_empty_completion backend/copilot/sdk/response_adapter_test.py::test_retry_adapter_without_prior_content_still_surfaces_empty_completion` — 2 passed.
- [x] Direct supplement assertion (`from backend.copilot.prompting import get_sdk_supplement; ...`) — all pass.
- [x] `ruff check` clean on edited files.
- [x] Validated against trace `9f3feb3dd843aa1e537de92f23666cc9` /
      `746b49e0f6012ff94035d0ceb3f7fca6` (the LinkedIn dev session).
- [ ] Manual dev verification with a fresh integration request after
      `latest` v30 propagates (≤5 min Langfuse cache TTL).

### DIFF
--- a/autogpt_platform/backend/backend/copilot/sdk/response_adapter.py
+++ b/autogpt_platform/backend/backend/copilot/sdk/response_adapter.py
@@ -83,12 +83,9 @@ class SDKResponseAdapter:
         self.has_started_reasoning = False
         self.has_ended_reasoning = True
         self.render_reasoning_in_ui = render_reasoning_in_ui
-        # Set by the service layer when a retry adapter is constructed mid-turn
-        # and the prior attempt already streamed text/tool results to the wire.
-        # The empty-completion guard ignores its local "no content this attempt"
-        # signal in that case — the user already received content from a
-        # previous attempt, so surfacing "The model returned an empty response."
-        # on top would be a false-positive.
+        # Service layer forwards this from the prior adapter on a retry-recreate
+        # so the empty-completion guard doesn't false-fire on a benign empty
+        # trailing ResultMessage when the prior attempt already streamed content.
         self.prior_attempt_emitted_visible_content = False
         self.current_tool_calls: dict[str, dict[str, str]] = {}
         self.resolved_tool_calls: set[str] = set()
@@ -162,6 +159,21 @@ class SDKResponseAdapter:
     def has_unresolved_tool_calls(self) -> bool:
         """True when there are tool calls that haven't received output yet."""
         return bool(self.current_tool_calls.keys() - self.resolved_tool_calls)
+
+    @property
+    def emitted_visible_content(self) -> bool:
+        """True when this adapter (or a prior attempt it inherited from) has
+        already streamed user-visible content — text, reasoning, or a tool
+        result — onto the wire. Used by the retry path so the new adapter
+        can carry forward the prior attempt's visibility state without
+        reaching into private flags.
+        """
+        return (
+            self.has_started_text
+            or self.has_started_reasoning
+            or self._any_tool_results_seen
+            or self.prior_attempt_emitted_visible_content
+        )
 
     def convert_message(self, sdk_message: Message) -> list[StreamBaseResponse]:
         """Convert a single SDK message to Vercel AI SDK format."""
@@ -608,12 +620,7 @@ class SDKResponseAdapter:
             return False
         if self._any_tool_results_seen:
             return False
-        # When this adapter is a retry adapter constructed after a prior attempt
-        # already streamed text/tools to the wire, an empty ResultMessage on
-        # this attempt is not a "ghost finish" from the user's perspective —
-        # they already received content. Firing the error here would render
-        # a "model returned an empty response" overlay on top of working
-        # output.
+        # Retry adapter — prior attempt already streamed content to the wire.
         if self.prior_attempt_emitted_visible_content:
             return False
         usage = msg.usage or {}
--- a/autogpt_platform/backend/backend/copilot/sdk/response_adapter.py
+++ b/autogpt_platform/backend/backend/copilot/sdk/response_adapter.py
@@ -83,6 +83,13 @@ class SDKResponseAdapter:
         self.has_started_reasoning = False
         self.has_ended_reasoning = True
         self.render_reasoning_in_ui = render_reasoning_in_ui
+        # Set by the service layer when a retry adapter is constructed mid-turn
+        # and the prior attempt already streamed text/tool results to the wire.
+        # The empty-completion guard ignores its local "no content this attempt"
+        # signal in that case — the user already received content from a
+        # previous attempt, so surfacing "The model returned an empty response."
+        # on top would be a false-positive.
+        self.prior_attempt_emitted_visible_content = False
         self.current_tool_calls: dict[str, dict[str, str]] = {}
         self.resolved_tool_calls: set[str] = set()
         self.step_open = False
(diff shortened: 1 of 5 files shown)


autogpt_platform/backend/backend/copilot/sdk/service.py
```diff
@@ -4515,6 +4516,19 @@ def _on_stderr(line: str) -> None:
...
+                # adapter — the user has already received content, and a
+                # silent-finish ResultMessage on the retry would otherwise
+                # surface a spurious "model returned an empty response" overlay
+                # on top of working output.
+                if (
+                    prior_adapter.has_started_text
+                    or prior_adapter.has_started_reasoning
+                    or prior_adapter._any_tool_results_seen
+                    or prior_adapter.prior_attempt_emitted_visible_content
+                ):
```
majdyz: 🤖 🟠 **Should Fix**: This block reaches into `prior_adapter._any_tool_results_seen` (single-underscore = by-convention private). The check mixes 3 public attrs + 1 private attr + 1 forwarded flag, which is a leaky abstraction and easy to drift if the adapter's internal fields are renamed.

Suggest encapsulating the 4-condition OR as a public property on `SDKResponseAdapter`:

```python
@property
def emitted_visible_content(self) -> bool:
    return (
        self.has_started_text
        or self.has_started_reasoning
        or self._any_tool_results_seen
        or self.prior_attempt_emitted_visible_content
    )
```

Then the call site here is just:

```python
if prior_adapter.emitted_visible_content:
    state.adapter.prior_attempt_emitted_visible_content = True
```

Keeps the adapter as the source of truth for what "emitted content" means, and the service-layer call site stops poking at private state.

autogpt_platform/backend/backend/copilot/sdk/response_adapter.py
```diff
@@ -601,6 +608,14 @@ def _is_empty_completion(self, msg: ResultMessage) -> bool:
             return False
         if self._any_tool_results_seen:
             return False
+        # When this adapter is a retry adapter constructed after a prior attempt
+        # already streamed text/tools to the wire, an empty ResultMessage on
+        # this attempt is not a "ghost finish" from the user's perspective —
+        # they already received content. Firing the error here would render
+        # a "model returned an empty response" overlay on top of working
+        # output.
+        if self.prior_attempt_emitted_visible_content:
```
majdyz, line 624: 🤖 🔵 **Nit**: This 6-line narrative comment explains "why the guard skips" by referencing cross-layer effects ("the user already received content", "working output"). Per repo style (`feedback_minimal_comments.md` — WHY-only one-liners; no narrative about model quirks / cross-layer effects), this could collapse to a single line. Suggest:

```python
# Retry adapter — prior attempt already streamed content to the wire.
if self.prior_attempt_emitted_visible_content:
    return False
```

Same applies to the corresponding 6-line comment on the service.py side at the call site.

autogpt_platform/backend/backend/copilot/sdk/service.py
majdyz: 🤖 Fixed in [f43d8961e](https://github.com/Significant-Gravitas/AutoGPT/commit/f43d8961edf35563cc571a4f46e9464d134c782a): added `SDKResponseAdapter.emitted_visible_content` property combining the 4 signals (text/reasoning started, tool results seen, prior-attempt flag). Service-layer retry path now calls `prior_adapter.emitted_visible_content` — no more reaching into private state.

autogpt_platform/backend/backend/copilot/sdk/response_adapter.py
majdyz: 🤖 Fixed in [f43d8961e](https://github.com/Significant-Gravitas/AutoGPT/commit/f43d8961edf35563cc571a4f46e9464d134c782a): collapsed both 6-line guards (response_adapter.py and the service.py call site) to single WHY one-liners.

PR #13128, opened 2026-05-15
feat(frontend/copilot): paginate session sidebar so threads past 50 stay reachable

0ubbe:
### Why / What / How

**Why.** The CoPilot session sidebar fetched only the first 50 sessions and stopped. Any user with more than 50 chats lost access to every older thread through the UI even though the backend already supported `offset` and returned a `total` count. Threads weren't lost in the DB — they were just invisible.

**What.** Wire proper pagination into the sidebar and mobile drawer with a "Load older chats" affordance, backed by a single shared infinite-query hook. Reroute all session-list cache invalidations so they refresh every loaded page, not just page 1.

**How.**
- New `useSessionList` hook wraps `useInfiniteQuery` around the existing `getV2ListSessions` fetcher: page size 50, `offset` pageParam, `refetchInterval` 10s, `getNextPageParam` derived from `total`.
- Hook lives on a fresh `SESSION_LIST_QUERY_KEY` so the infinite cache doesn't collide with the orval-generated single-query key shape (which would break shape-sensitive cache readers).
- `ChatSidebar` and `MobileDrawer` consume the hook and render a ghost `Button` with `loading` state at the end of the list when `hasMore`.
- `useSessionTitlePoll` walks `InfiniteData` pages via a small `flattenSessions` helper instead of reading the legacy single-page response.
- Five invalidation callsites (`useChatSession`, `useCopilotNotifications`, `useSessionDeletion`, `useSessionTitlePoll`, `ChatSidebar`) updated to invalidate the new key, so create/delete/rename/cross-tab-sync still refresh the sidebar correctly.

### Changes 🏗️

- Add `frontend/src/app/(platform)/copilot/useSessionList.ts` (shared paginated hook + `SESSION_LIST_QUERY_KEY` + `flattenSessions` cache walker).
- `ChatSidebar.tsx`: replace single-page `useGetV2ListSessions` with `useSessionList`, add Load-more button, point invalidations at the new key.
- `MobileDrawer.tsx`: mirror the same paginated fetch + Load-more affordance.
- `useSessionTitlePoll.ts`: walk infinite-cache pages; invalidate the new key.
- `useChatSession.ts`, `useCopilotNotifications.ts`, `useSessionDeletion.ts`: switch invalidation calls to `SESSION_LIST_QUERY_KEY`.

#### For code changes:
- [x] I have made a test plan
  - [x] `pnpm types` — clean
  - [x] `pnpm lint` — clean (only pre-existing img-tag warnings)
  - [x] `pnpm test:unit src/app/(platform)/copilot` — 59 files / 929 tests pass, including the existing `ChatSidebar` delete + status-indicator suites against the new wiring
  - [ ] Manual: seed a user with >50 sessions; confirm "Load older chats" appears, paginates correctly, and that the button disappears when `loaded === total`
  - [ ] Manual: confirm the 10s refetch still surfaces running/queued/processing indicators on all loaded pages
  - [ ] Manual: delete / rename / create-new-session — confirm the sidebar refreshes
  - [ ] Manual: cross-tab `localStorage` storage event still invalidates the list
  - [ ] Manual: title-poll still animates the new title in after stream completion

### DIFF
--- a/autogpt_platform/frontend/src/app/(platform)/copilot/useSessionList.ts
+++ b/autogpt_platform/frontend/src/app/(platform)/copilot/useSessionList.ts
@@ -1,10 +1,13 @@
-import {
-  getV2ListSessions,
-  type getV2ListSessionsResponse,
-} from "@/app/api/__generated__/endpoints/chat/chat";
+import { getV2ListSessions } from "@/app/api/__generated__/endpoints/chat/chat";
 import { type InfiniteData, useInfiniteQuery } from "@tanstack/react-query";
 
 export const SESSION_LIST_PAGE_SIZE = 50;
+// `refetchInterval` on a `useInfiniteQuery` refetches every loaded page, and
+// TanStack Query v5 removed `refetchPage` so we can't scope it to page 0.
+// Worst case is bounded by the user's session count (and the WebSocket
+// invalidations in `useCopilotNotifications` already handle the high-signal
+// completion events) — revisit by extracting a separate live-status query if
+// this ever becomes a real bandwidth concern.
 export const SESSION_LIST_REFETCH_INTERVAL_MS = 10_000;
 
 // Fresh, paginated-cache key. The orval-generated key targets the non-infinite
@@ -37,12 +40,8 @@ export function useSessionList({ enabled = true }: Args = {}) {
     enabled,
   });
 
-  const sessions = flattenSessions(query.data);
-  const total = getTotal(query.data);
-
   return {
-    sessions,
-    total,
+    sessions: flattenSessions(query.data),
     isLoading: query.isLoading,
     hasMore: !!query.hasNextPage,
     isLoadingMore: query.isFetchingNextPage,
@@ -63,12 +62,3 @@ function countLoadedSessions(pages: SessionListPage[]) {
     0,
   );
 }
-
-function getTotal(data: SessionListInfiniteData | undefined) {
-  const lastPage = data?.pages.at(-1);
-  if (!lastPage || lastPage.status !== 200) return 0;
-  return lastPage.data.total;
-}
-
-// Re-exported so cache walkers don't have to depend on the generated module.
-export type { getV2ListSessionsResponse };
--- a/autogpt_platform/frontend/src/app/(platform)/copilot/useSessionList.ts
+++ b/autogpt_platform/frontend/src/app/(platform)/copilot/useSessionList.ts
@@ -1,0 +1,74 @@
+import {
+  getV2ListSessions,
+  type getV2ListSessionsResponse,
+} from "@/app/api/__generated__/endpoints/chat/chat";
+import { type InfiniteData, useInfiniteQuery } from "@tanstack/react-query";
+
+export const SESSION_LIST_PAGE_SIZE = 50;
+export const SESSION_LIST_REFETCH_INTERVAL_MS = 10_000;
+
+// Fresh, paginated-cache key. The orval-generated key targets the non-infinite
+// `useQuery` cache shape; keeping the infinite cache on a separate key avoids
+// shape collisions and lets us hand a stable key to invalidation callsites.
+export const SESSION_LIST_QUERY_KEY = ["copilot", "session-list"] as const;
+
+type SessionListPage = Awaited<ReturnType<typeof getV2ListSessions>>;
+export type SessionListInfiniteData = InfiniteData<SessionListPage>;
+
+interface Args {
+  enabled?: boolean;
+}
+
+export function useSessionList({ enabled = true }: Args = {}) {
+  const query = useInfiniteQuery({
+    queryKey: SESSION_LIST_QUERY_KEY,
+    queryFn: ({ pageParam }) =>
+      getV2ListSessions({
+        limit: SESSION_LIST_PAGE_SIZE,
+        offset: pageParam,
+      }),
+    initialPageParam: 0,
+    getNextPageParam: (lastPage, allPages) => {
+      if (lastPage.status !== 200) return undefined;
+      const loaded = countLoadedSessions(allPages);
+      return loaded < lastPage.data.total ? loaded : undefined;
+    },
+    refetchInterval: SESSION_LIST_REFETCH_INTERVAL_MS,
+    enabled,
+  });
+
+  const sessions = flattenSessions(query.data);
+  const total = getTotal(query.data);
+
+  return {
+    sessions,
+    total,
+    isLoading: query.isLoading,
+    hasMore: !!query.hasNextPage,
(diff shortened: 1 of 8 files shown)


autogpt_platform/frontend/src/app/(platform)/copilot/useSessionList.ts
```diff
@@ -0,0 +1,74 @@
...
+    enabled,
+  });
+
+  const sessions = flattenSessions(query.data);
+  const total = getTotal(query.data);
+
+  return {
+    sessions,
+    total,
+    isLoading: query.isLoading,
```
0ubbe, line 45: 🤖 🟠 **Should Fix**: This hook is the load-bearing piece of the PR — `hasMore` derivation, `getNextPageParam` advancing `offset` by loaded count, and the `flattenSessions` cache walker used by `useSessionTitlePoll` — but there are no integration tests for it. The existing `ChatSidebar.test.tsx` covers delete + chat_status indicators against the new wiring, but nothing exercises the pagination paths.

Per `frontend/AGENTS.md` / `CONTRIBUTING.md`, integration tests at the page level are the default (~90%). Suggested additions in `ChatSidebar/__tests__`:

- MSW handler returns `{ sessions: [50 items], total: 75 }` → assert the "Load older chats" button is rendered.
- Click the button → second MSW handler returns offset=50 → assert the additional sessions appear and the button now hides because `loaded === total`.
- MSW handler returns `{ sessions: [10 items], total: 10 }` → assert the button is never rendered.

Without these, a future change to `getNextPageParam` or the offset math would silently regress the "threads go missing" bug this PR is trying to fix.

autogpt_platform/frontend/src/app/(platform)/copilot/useSessionList.ts
```diff
@@ -0,0 +1,74 @@
...
+        limit: SESSION_LIST_PAGE_SIZE,
+        offset: pageParam,
+      }),
+    initialPageParam: 0,
+    getNextPageParam: (lastPage, allPages) => {
+      if (lastPage.status !== 200) return undefined;
+      const loaded = countLoadedSessions(allPages);
+      return loaded < lastPage.data.total ? loaded : undefined;
+    },
+    refetchInterval: SESSION_LIST_REFETCH_INTERVAL_MS,
```
0ubbe, line 39: 🤖 🟡 **Nice to Have**: `useInfiniteQuery`'s `refetchInterval` refetches **every loaded page** by default (TanStack Query docs — `refetchPages: 'all'`). For a power user who scrolled through 4 pages of 50 sessions, the sidebar fires 4 parallel `GET /api/chat/sessions` every 10 s, and the backend then issues `len(sessions)` Redis `HGET` per request for the processing-status batch — so worst-case ~200 Redis ops every 10 s per open sidebar.

For the original single-page hook this was bounded at 50/10 s. With pagination it now scales with how far the user paged.

Mitigations to consider:

- Pass `refetchPage: (_page, index) => index === 0` to `refetch`, scoping the interval to page 0 (status freshness still works for the most recent threads, which is where running/queued indicators matter most).
- Or scope the interval to "first page only" and rely on the existing invalidation callsites for older pages.

Not a blocker — the worst case is still bounded by user session count — but worth flagging before this ships, especially since `refetchOnWindowFocus` will compound the effect.

autogpt_platform/frontend/src/app/(platform)/copilot/useSessionList.ts
```diff
@@ -0,0 +1,74 @@
...
+}
+
+function getTotal(data: SessionListInfiniteData | undefined) {
+  const lastPage = data?.pages.at(-1);
+  if (!lastPage || lastPage.status !== 200) return 0;
+  return lastPage.data.total;
+}
+
+// Re-exported so cache walkers don't have to depend on the generated module.
+export type { getV2ListSessionsResponse };
```
0ubbe: 🤖 🟡 **Nice to Have**: Dead surface — `getV2ListSessionsResponse` re-export and `total` field.

- The comment on line 73 says "Re-exported so cache walkers don't have to depend on the generated module," but the only cache walker (`useSessionTitlePoll`) imports `SessionListInfiniteData`, not this type. `grep` across `src/` shows no consumer for the re-export.
- The hook's return value includes `total`, but neither `ChatSidebar` nor `MobileDrawer` destructures it.

Suggest dropping both to keep the public surface honest. If `total` is intended for a future "X of Y chats" indicator, fine to leave with a `TODO` note; otherwise it'll just bit-rot.

autogpt_platform/frontend/src/app/(platform)/copilot/useSessionList.ts
0ubbe: 🤖 Fixed in [81d49320c](https://github.com/Significant-Gravitas/AutoGPT/commit/81d49320ca3affd782e5f2c152b7535c77a48c6b): added a `ChatSidebar — pagination` describe block with three tests — button hidden when `total === loaded`, button rendered when `total > loaded`, and a click on the button issues a follow-up request with the correct `offset` (asserted by intercepting the URL params) and hides the button once the final page lands.

autogpt_platform/frontend/src/app/(platform)/copilot/useSessionList.ts
0ubbe: 🤖 Addressed in [81d49320c](https://github.com/Significant-Gravitas/AutoGPT/commit/81d49320ca3affd782e5f2c152b7535c77a48c6b): documented the trade-off inline at `SESSION_LIST_REFETCH_INTERVAL_MS`. The clean mitigation I suggested (`refetchPage`) doesn't exist in TanStack Query v5 (it was removed), and the v5 alternatives — splitting into a separate live-status query, or capping `maxPages` — add complexity disproportionate to the saving since the worst case is bounded by the user's session count and `useCopilotNotifications` already invalidates on WebSocket completion events. Left the comment so a future maintainer can revisit if it becomes a real bandwidth concern.

autogpt_platform/frontend/src/app/(platform)/copilot/useSessionList.ts
0ubbe: 🤖 Fixed in [81d49320c](https://github.com/Significant-Gravitas/AutoGPT/commit/81d49320ca3affd782e5f2c152b7535c77a48c6b): dropped the unused `getV2ListSessionsResponse` re-export, the unused `total` field on the hook return, and the now-orphaned `getTotal` helper. `total` is still read internally by `getNextPageParam` directly off `lastPage.data.total`.

0ubbe:
https://github.com/user-attachments/assets/7c84184c-0d00-49a1-91dc-21e9c9aeb0a4

PR #13129, opened 2026-05-15
feat(platform): show cost breakdown in agent briefing panel

0ubbe:
### Why / What / How

**Why** — The library Briefing Panel's "Spent this month" tile was built on top of `GET /api/v1/executions`, which is hard-capped at 250 rows (`backend/api/features/v1.py:1893-1906`). For any moderately busy account the displayed monthly total was silently wrong, and a fuller breakdown (top-cost runs, by-agent rollup, daily trend) was impossible to compute correctly on the client.

This PR adds an authoritative server-side aggregator endpoint **and** wires it into the briefing panel: the tile now shows an accurate monthly total, and a new "Cost breakdown" section under the usage meters surfaces the most expensive runs, spend by agent, and a daily-spend mini chart.

**What — Backend**

New endpoint:

```
GET /api/v1/executions/cost-summary
  ?since=<datetime>   (default: start of current UTC month)
  ?until=<datetime>   (default: now)
  ?top_runs_limit=N   (1..50, default 10)
```

Response (`UserExecutionCostSummary`):
- `total_cents`, `run_count`, `billable_run_count`, `failed_cost_cents`
- `by_agent[]`: `{ graph_id, cost_cents, run_count }` — top 50 by cost
- `top_runs[]`: `{ execution_id, graph_id, cost_cents, started_at, status, duration_seconds, node_error_count }` — top N by cost
- `daily[]`: `{ date (YYYY-MM-DD UTC), cost_cents, run_count }`

**What — Frontend**

- `useLibraryFleetSummary` no longer sums `stats.cost` client-side. It pulls `total_cents` from the new endpoint, so the "Spent this month" tile stays correct above 250 runs.
- New `CostsBreakdown` component renders inside `UsageSection` (the "Spent this month" tile destination):
  - Headline stat tiles: Total this month, Runs, Avg / run, Wasted on failed.
  - `TopRunsList` — most expensive runs, each linking to its execution detail.
  - `SpendByAgentList` — horizontal bar list with run counts and per-agent links.
  - `DailySpendBars` — lightweight inline-SVG bar chart over the window.
- Empty state ("No spend this month yet") when total is zero.
- Loading state with skeletons.

**How**

- Backend uses four parallel `query_raw_with_schema` aggregations against `AgentGraphExecution`, reading `stats->>'cost'::numeric` and bucketing by `agentGraphId`, by `id`, and by `("createdAt" AT TIME ZONE 'UTC')::date`. Filters: `userId`, `isDeleted=false`, `createdAt` in window (chosen over `startedAt` so the `(userId, isDeleted, createdAt)` index is hit — see docstring in `get_user_cost_summary`), dry-run excluded via `COALESCE((stats->>'is_dry_run')::boolean, false)`.
- `billable_run_count` (count of runs with `cost > 0`) is exposed separately from `run_count` so the frontend "Avg / run" tile divides by billable runs only — including zero-cost runs in the denominator deflates the displayed average.
- Schema named `UserExecutionCostSummary` to avoid a collision with the existing admin-only `platform_cost.UserCostSummary` (otherwise the OpenAPI generator falls back to fully-qualified `backend__data__execution__UserCostSummary`).
- Frontend uses the generated `useGetV1UserCostSummary` hook (Orval). Cents everywhere — same unit the existing `formatCents` helper consumes.

### Changes 🏗️

**Backend**
- `backend/data/execution_cost_summary.py` (new) — `UserExecutionCostSummary` / `UserAgentCostRollup` / `UserTopRun` / `UserDailyCost` models + `get_user_cost_summary()` aggregator. Lives in its own module (not `execution.py`) so the per-user cost aggregator stays separate from the general execution data layer.
- `backend/api/features/v1.py` — `GET /executions/cost-summary` route (auth: `requires_user`, returns `UserExecutionCostSummary`).
- `backend/api/features/v1_test.py` — 4 route tests: happy-path payload mapping (including `billable_run_count`), `since`/`until`/`top_runs_limit` query-param forwarding, 422 on out-of-range limit, 422 on inverted window.
- `backend/data/execution_cost_summary_test.py` (new) — 5 integration tests against real Postgres: empty window, by-agent + top-runs ordering (now also asserts `billable_run_count` excludes zero-cost runs), dry-run/window exclusion, daily UTC bucketing, createdAt-vs-startedAt bucketing invariant.

**Frontend**
- `library/hooks/useLibraryFleetSummary.ts` — `monthlySpend` now reads from `useGetV1UserCostSummary` instead of summing executions on the client.
- `library/components/AgentBriefingPanel/CostsBreakdown/` (new) — `CostsBreakdown.tsx` + `useCostsBreakdown.ts` + `helpers.ts` + `components/{TopRunsList,SpendByAgentList,DailySpendBars}.tsx`.
- `library/components/AgentBriefingPanel/BriefingTabContent.tsx` — `UsageSection` now always renders `<CostsBreakdown />` below the usage meters and gracefully omits the usage block when copilot limits aren't configured.
- `library/__tests__/briefing.test.tsx` — re-targeted at the new endpoint via MSW.
- `library/components/AgentBriefingPanel/__tests__/BriefingTabContent.test.tsx` — extended with 4 new cost-breakdown integration tests (zero-spend state, full happy path, graph-id fallback label, billable-run-count denominator).
- `src/app/api/openapi.json` — surgically patched with just the new path + 4 schemas (the rest of the file matches `dev` byte-for-byte after regeneration).

No schema migrations, no config changes, no impact on existing routes.

#### For code changes:
- [x] I have made a test plan
  - [x] Backend RED proven before any production code (missing symbols, missing function on `v1` namespace, missing route)
  - [x] Backend GREEN: 3 route tests pass via standalone TestClient + mocked aggregator (Docker not available locally; CI runs the full suite)
  - [x] Backend: `ruff` / `black` / `isort` clean on changed files; pytest collection succeeds for both test files (3 + 4 tests)
  - [x] Frontend: `pnpm format` / `pnpm lint` / `pnpm types` all clean
  - [x] Frontend: `pnpm test:unit` — 197 test files / 2626 tests pass
  - [ ] CI runs the 4 SQL integration tests against real Postgres


autogpt_platform/backend/backend/data/execution.py
```diff
@@ -725,6 +726,157 @@ async def get_graph_executions_paginated(
...
+        " AND COALESCE((stats->>'is_dry_run')::boolean, false) = false"
+    )
+    params = (user_id, since, until)
+
+    totals_rows, by_agent_rows, top_runs_rows, daily_rows = await asyncio.gather(
+        query_raw_with_schema(
+            "SELECT"
+            "  COALESCE(SUM((stats->>'cost')::numeric), 0)::bigint AS total_cents,"
+            "  COUNT(*)::bigint AS run_count,"
+            "  COALESCE(SUM(CASE WHEN \"executionStatus\" = 'FAILED'"
```
0ubbe: 🤖 🟠 **Should Fix**: `failed_cost_cents` only sums runs where `executionStatus = 'FAILED'`, but TERMINATED runs (user-cancelled / system-killed) also accrue cost. Elsewhere in this file (e.g., `generate_data.py`'s `_fetch_execution_aggregate_rows` and the historical executions analytics at line 1992) both `FAILED` and `TERMINATED` are treated as the "failed" bucket. The "Wasted on failed" tile on the frontend will under-report waste for users who frequently terminate runs.

Suggest: `WHEN "executionStatus" IN ('FAILED', 'TERMINATED')`.

autogpt_platform/backend/backend/data/execution.py
```diff
@@ -725,6 +726,157 @@ async def get_graph_executions_paginated(
...
+    Defaults to the current calendar month (UTC) when `since`/`until` are not provided.
+    Dry-run executions are excluded.
+    """
+    now = datetime.now(timezone.utc)
+    if since is None:
+        since = now.replace(day=1, hour=0, minute=0, second=0, microsecond=0)
+    if until is None:
+        until = now
+
+    base_where = (
```
0ubbe: 🤖 🟠 **Should Fix**: All four aggregations filter by `(userId, isDeleted, startedAt)`, but the only multi-column index on `AgentGraphExecution` for this user is `@@index([userId, isDeleted, createdAt])` (schema.prisma:663). For users with thousands of executions the planner falls back to scanning by `createdAt` and filtering `startedAt` separately, which gets linear with history depth.

Either add `@@index([userId, isDeleted, startedAt])` in a migration alongside this PR, or switch the queries to filter on `createdAt` (the existing index) and document that "in-window" means "created in window" — which is what `generate_data.py`'s sibling `_fetch_execution_aggregate_rows` already does.

autogpt_platform/backend/backend/data/execution.py
```diff
@@ -725,6 +726,157 @@ async def get_graph_executions_paginated(
...
+            top_runs_limit,
+        ),
+        query_raw_with_schema(
+            "SELECT"
+            "  to_char(\"startedAt\" AT TIME ZONE 'UTC', 'YYYY-MM-DD') AS date,"
+            "  COALESCE(SUM((stats->>'cost')::numeric), 0)::bigint AS cost_cents,"
+            "  COUNT(*)::bigint AS run_count"
+            ' FROM {schema_prefix}"AgentGraphExecution"'
+            f" WHERE {base_where}"
+            "  GROUP BY to_char(\"startedAt\" AT TIME ZONE 'UTC', 'YYYY-MM-DD')"
```
0ubbe: 🤖 🟡 **Nice to Have**: The daily aggregation only emits rows for dates that had at least one execution, so days with zero spend are missing from `daily[]` entirely. `DailySpendBars` then draws a chart with gaps in the X-axis — visually identical to "those days are outside the window" rather than "zero spend."

Two options:
- Server: `generate_series(date_trunc('day', $2), date_trunc('day', $3), '1 day') LEFT JOIN ...` to emit a contiguous day series with 0-filled buckets.
- Client: fill the gaps in `useCostsBreakdown` before passing to `DailySpendBars`.

Client-side gap fill is the lighter change and keeps the SQL simple.

autogpt_platform/frontend/src/app/(platform)/library/hooks/useLibraryFleetSummary.ts
```diff
@@ -114,7 +95,7 @@ export function useLibraryFleetSummary(
       listening: 0,
       scheduled: 0,
       idle: 0,
-      monthlySpend: monthlySpendCents,
+      monthlySpend: costSummary?.total_cents ?? 0,
```
0ubbe: 🤖 🟡 **Nice to Have**: While the cost-summary endpoint is in-flight, `costSummary` is `undefined` and the "Spent this month" tile shows `$0.00`. The previous client-side aggregation rendered the partial total as soon as the executions list resolved.

Consider showing the tile in a skeleton/loading state until `costSummary` lands, or keeping `isLoading` from `useGetV1UserCostSummary` accessible on `FleetSummary` so `StatsGrid` can swap the value for a skeleton. Otherwise the tile will briefly flash `$0.00` on every library page load before the real value arrives, which reads as "no spend" instead of "loading."

autogpt_platform/backend/backend/api/features/v1.py
```diff
@@ -1906,6 +1907,38 @@ async def list_graphs_executions(
...
+        description="Maximum number of top-cost runs to return.",
+    ),
+) -> UserCostSummary:
+    """Aggregated cost breakdown for the calling user's graph executions."""
+    return await get_user_cost_summary(
+        user_id=user_id,
+        since=since,
+        until=until,
+        top_runs_limit=top_runs_limit,
+    )
```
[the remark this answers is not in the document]
0ubbe: 🤖 Fixed in [3b23767a8](https://github.com/Significant-Gravitas/AutoGPT/commit/3b23767a8abb2099336232aa9a38acf9a5180a7f): added a 422 in the route handler with detail `"\`since\` must be earlier than or equal to \`until\`."`, plus a route test `test_executions_cost_summary_rejects_inverted_window` to lock the contract.

autogpt_platform/backend/backend/data/execution.py
```diff
@@ -725,6 +726,157 @@ async def get_graph_executions_paginated(
...
+_MAX_BY_AGENT_ROWS = 50
+
+
+async def get_user_cost_summary(
+    *,
+    user_id: str,
+    since: datetime | None = None,
+    until: datetime | None = None,
+    top_runs_limit: int = 10,
+) -> UserCostSummary:
```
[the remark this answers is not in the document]
0ubbe: 🤖 Fixed in [3b23767a8](https://github.com/Significant-Gravitas/AutoGPT/commit/3b23767a8abb2099336232aa9a38acf9a5180a7f): `get_user_cost_summary` now clamps via `top_runs_limit = max(1, min(_MAX_TOP_RUNS, top_runs_limit))` (where `_MAX_TOP_RUNS = 50`) before building the SQL, so direct callers can't bypass the FastAPI Query bounds.

autogpt_platform/frontend/src/app/(platform)/auth/integrations/oauth_callback/route.ts
```diff
@@ -33,19 +33,44 @@ export async function GET(request: Request) {
...
+        } catch(e) {}
+        try {
+          if (window.opener && !window.opener.closed) {
+            window.opener.postMessage(msg, window.location.origin);
+          }
+        } catch(e) {}
+        try {
+          if (state) {
+            localStorage.setItem("oauth_popup_result_" + state, JSON.stringify(msg));
+          }
```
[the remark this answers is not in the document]
0ubbe: 🤖 Fixed in [3b23767a8](https://github.com/Significant-Gravitas/AutoGPT/commit/3b23767a8abb2099336232aa9a38acf9a5180a7f): the error payload now spreads `...(state ? { state } : {})` so the originating page's state-keyed listeners route the failure to the right flow instead of timing out. Widened the `OAuthPopupResultMessage` failure variant in `types.ts` to allow the optional `state` field.

autogpt_platform/backend/backend/data/execution.py
```diff
@@ -725,6 +726,157 @@ async def get_graph_executions_paginated(
...
+            *params,
+        ),
+        query_raw_with_schema(
+            "SELECT"
+            "  id AS execution_id,"
+            '  "agentGraphId" AS graph_id,'
+            "  COALESCE((stats->>'cost')::numeric, 0) AS cost_cents,"
+            '  "startedAt" AS started_at,'
+            '  "executionStatus" AS status,'
+            "  COALESCE((stats->>'duration')::numeric, 0) AS duration_seconds,"
```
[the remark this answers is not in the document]
0ubbe: 🤖 Fixed in [3b23767a8](https://github.com/Significant-Gravitas/AutoGPT/commit/3b23767a8abb2099336232aa9a38acf9a5180a7f): `top_runs` now casts `cost_cents` to `::bigint` (matching `total_cents`, `failed_cost_cents`, and the by-agent rollup). Also added `AND COALESCE((stats->>'cost')::numeric, 0) > 0` so $0 runs don't dominate the top list by created-at tiebreaks.

autogpt_platform/frontend/src/app/api/openapi.json
```diff
@@ -9430,6 +9430,82 @@
           }
         }
       }
+    },
+    "/api/executions/cost-summary": {
+      "get": {
```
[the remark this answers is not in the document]
0ubbe: 🤖 The analysis confirmed the endpoint is wired correctly (path `/api/executions/cost-summary` returns `UserExecutionCostSummary` with the four schemas present). No action needed — leaving this for the bot to resolve once it sees the openapi update in commit [3b23767a8](https://github.com/Significant-Gravitas/AutoGPT/commit/3b23767a8abb2099336232aa9a38acf9a5180a7f).

autogpt_platform/frontend/src/app/api/openapi.json
```diff
@@ -19229,6 +19305,87 @@
...
+          "started_at",
+          "status",
+          "duration_seconds",
+          "node_error_count"
+        ],
+        "title": "UserTopRun"
+      },
+      "UserDailyCost": {
+        "properties": {
+          "date": { "type": "string", "title": "Date" },
```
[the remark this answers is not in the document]
0ubbe: 🤖 Fixed in [3b23767a8](https://github.com/Significant-Gravitas/AutoGPT/commit/3b23767a8abb2099336232aa9a38acf9a5180a7f): switched `UserDailyCost.date` from `str` to `datetime.date` in the Pydantic model, so the regenerated openapi.json now exposes `"format": "date"` for the field. SQL was updated to return a true DATE column too.

autogpt_platform/backend/backend/data/execution.py
0ubbe: 🤖 Fixed in [3b23767a8](https://github.com/Significant-Gravitas/AutoGPT/commit/3b23767a8abb2099336232aa9a38acf9a5180a7f): the failed-cost branch is now `WHEN "executionStatus" IN ('FAILED', 'TERMINATED')`, matching the convention in `generate_data.py`'s aggregator.

autogpt_platform/backend/backend/data/execution.py
0ubbe: 🤖 Fixed in [3b23767a8](https://github.com/Significant-Gravitas/AutoGPT/commit/3b23767a8abb2099336232aa9a38acf9a5180a7f): all four queries now filter on `createdAt` instead of `startedAt`, hitting the existing `@@index([userId, isDeleted, createdAt])`. The daily bucket is also keyed off `createdAt` for consistency. `startedAt` is still surfaced in the top_runs response payload (coalesced to `createdAt` if null) so the UI shows when the run actually ran.

autogpt_platform/backend/backend/data/execution.py
0ubbe: 🤖 Fixed in [3b23767a8](https://github.com/Significant-Gravitas/AutoGPT/commit/3b23767a8abb2099336232aa9a38acf9a5180a7f): handled client-side. New `fillDailyGaps` in `CostsBreakdown/helpers.ts` walks min→max UTC date and emits zero-spend buckets for missing days, keeping the SQL simple and saving a server round-trip for the chart layout.

autogpt_platform/frontend/src/app/(platform)/library/hooks/useLibraryFleetSummary.ts
0ubbe: 🤖 Fixed in [3b23767a8](https://github.com/Significant-Gravitas/AutoGPT/commit/3b23767a8abb2099336232aa9a38acf9a5180a7f): added `fallbackMonthlySpendCents(executions)` and `monthlySpend = costSummary?.total_cents ?? fallback(...)`. The fallback is structurally a lower bound (executions endpoint is still 250-row capped) but it's enough to avoid the $0.00 flash on first paint. Once the cost-summary query resolves the authoritative value replaces it.

autogpt_platform/backend/backend/data/execution_cost_summary_test.py
```diff
@@ -0,0 +1,288 @@
...
+        summary = await get_user_cost_summary(
+            user_id=user_id,
+            since=since,
+            until=now + timedelta(minutes=1),
+            top_runs_limit=3,
+        )
+
+        assert summary.total_cents == 1550
+        assert summary.run_count == 4
+        assert summary.failed_cost_cents == 50
```
0ubbe: 🤖 🟡 **Nice to Have**: Round 1's `failed_cost_cents` SQL now sums runs in `('FAILED', 'TERMINATED')`, but this test only fixtures a FAILED run, so the TERMINATED branch is uncovered — a regression that drops `'TERMINATED'` from the IN-list would slip through.

Suggest extending the fixture with one more run:

```python
await _create_run(
    run_id=f"e5-{uuid4()}",
    user_id=user_id,
    graph_id=graph_b,
    status=AgentExecutionStatus.TERMINATED,
    cost_cents=20,
    started_at=now - timedelta(minutes=10),
)
```

and bumping the assertion to `assert summary.failed_cost_cents == 70` (50 FAILED + 20 TERMINATED).

0ubbe: 🤖 Fixed in [1e8a6cfbc](https://github.com/Significant-Gravitas/AutoGPT/commit/1e8a6cfbcdbd079638d86c793c0a2cf88dd370de): added a TERMINATED run with cost=20 to the by-agent/top-runs fixture and bumped `failed_cost_cents` assertion to 70 (50 FAILED + 20 TERMINATED). Also updated `total_cents` (1570), `run_count` (5), and `by_agent[graph_b]` (cost 1070, run_count 3) so the surrounding assertions stay tight.

autogpt_platform/backend/backend/data/execution.py
```diff
@@ -725,6 +726,168 @@ async def get_graph_executions_paginated(
...
+            *params,
+        ),
+    )
+
+    totals = totals_rows[0] if totals_rows else {}
+    return UserExecutionCostSummary(
+        total_cents=int(totals.get("total_cents") or 0),
+        run_count=int(totals.get("run_count") or 0),
+        failed_cost_cents=int(totals.get("failed_cost_cents") or 0),
+        by_agent=[
```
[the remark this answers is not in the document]
0ubbe: 🤖 Fixed in [00605b2fc](https://github.com/Significant-Gravitas/AutoGPT/commit/00605b2fcb8b35431a3096dee3c4d2f29181db91): the daily SQL now quotes the `date` alias in both the SELECT and ORDER BY clauses — `AS "date"` / `ORDER BY "date" ASC`. No other consumers change (result-row key stays `date`).

autogpt_platform/backend/backend/data/execution.py
```diff
@@ -725,6 +726,168 @@ async def get_graph_executions_paginated(
...
+            "SELECT"
+            "  (\"createdAt\" AT TIME ZONE 'UTC')::date AS date,"
+            "  COALESCE(SUM((stats->>'cost')::numeric), 0)::bigint AS cost_cents,"
+            "  COUNT(*)::bigint AS run_count"
+            ' FROM {schema_prefix}"AgentGraphExecution"'
+            f" WHERE {base_where}"
+            "  GROUP BY (\"createdAt\" AT TIME ZONE 'UTC')::date"
+            "  ORDER BY date ASC",
+            *params,
+        ),
```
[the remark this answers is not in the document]
0ubbe: 🤖 Fixed in [00605b2fc](https://github.com/Significant-Gravitas/AutoGPT/commit/00605b2fcb8b35431a3096dee3c4d2f29181db91): the daily SQL now quotes the `date` alias in both the SELECT and ORDER BY clauses — `AS "date"` / `ORDER BY "date" ASC`. No other consumers change (result-row key stays `date`).

autogpt_platform/frontend/src/app/(platform)/library/components/AgentBriefingPanel/CostsBreakdown/CostsBreakdown.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+interface Props {
+  agents: LibraryAgent[];
+}
+
+export function CostsBreakdown({ agents }: Props) {
+  const { summary, agentLookup, isLoading, isError, hasAnySpend } =
+    useCostsBreakdown(agents);
+
+  if (isError) return null;
```
0ubbe: 🤖 🟡 **Nice to Have**: When the cost-summary query errors out, this swallows the failure (`return null`) and the section just disappears with no Sentry/log trail. Browsing user sees a degraded panel; we see nothing in monitoring. Suggest at least `Sentry.captureException(error)` from `useCostsBreakdown` on transition into `isError`, or render a tiny `<Text variant='small' className='text-neutral-400'>Couldn't load cost breakdown.</Text>` so the omission is visible. Consistent with the `ErrorCard` pattern in `CONTRIBUTING.md > Error handling`.

autogpt_platform/backend/backend/data/execution_cost_summary_test.py
```diff
@@ -0,0 +1,300 @@
...
+    if is_dry_run:
+        stats["is_dry_run"] = True
+    await AgentGraphExecution.prisma().create(
+        data={
+            "id": run_id,
+            "agentGraphId": graph_id,
+            "agentGraphVersion": 1,
+            "executionStatus": status,
+            "userId": user_id,
+            "createdAt": started_at,
```
0ubbe: 🤖 🟡 **Nice to Have**: `_create_run` sets `createdAt = started_at`, so the daily-bucket test can't distinguish "groups by createdAt" from "groups by startedAt." The docstring at `execution.py:776-779` documents that bucketing intentionally follows `createdAt` (for index hit reasons) — but a regression that switches the SQL to `startedAt` would silently pass all four integration tests. Suggest one extra run where `createdAt` is day1 and `startedAt` is day2 (a queued-then-started scenario), asserting it lands in the day1 bucket. Locks the documented invariant into a test.

0ubbe:
🤖 🔵 **Nit** (description hygiene): The **How — Backend** section says the daily aggregator buckets by `to_char("startedAt" AT TIME ZONE 'UTC', 'YYYY-MM-DD')`, but the actual code at `execution.py:842-852` buckets by `("createdAt" AT TIME ZONE 'UTC')::date` (and the inline docstring at L776-779 explains why `createdAt` is the deliberate choice — index hit). Either rewrite "by `to_char(...)`" to "by `("createdAt" AT TIME ZONE 'UTC')::date`", or just drop the SQL snippet and link to the docstring. Future readers will get confused otherwise.

autogpt_platform/frontend/src/app/(platform)/library/components/AgentBriefingPanel/CostsBreakdown/CostsBreakdown.tsx
0ubbe: 🤖 Fixed in [b82f9d96b](https://github.com/Significant-Gravitas/AutoGPT/commit/b82f9d96b00eb44cf1b271aea9474632d23f4662): error path now renders a small "Couldn't load cost breakdown." message so the omission is visible to users. (Skipped Sentry capture from the hook — would require `useEffect` to react to `isError` flipping, which CLAUDE.md bans; the global error boundary handles uncaught render errors.)

autogpt_platform/backend/backend/data/execution_cost_summary_test.py
0ubbe: 🤖 Fixed in [b82f9d96b](https://github.com/Significant-Gravitas/AutoGPT/commit/b82f9d96b00eb44cf1b271aea9474632d23f4662): `_create_run` now takes an optional `created_at` param (defaulting to `started_at` for backward compat). Added `test_daily_buckets_follow_created_at_not_started_at` — a queued-then-started run with createdAt=Feb-5 and startedAt=Feb-7 lands in the Feb-5 bucket. A future swap of the SQL column would now fail this test.

0ubbe:
🤖 Fixed in [b82f9d96b](https://github.com/Significant-Gravitas/AutoGPT/commit/b82f9d96b00eb44cf1b271aea9474632d23f4662): PR description updated — the **How — Backend** bullet now says `("createdAt" AT TIME ZONE 'UTC')::date` and adds a parenthetical pointing at the index-hit rationale in the docstring.

autogpt_platform/frontend/src/app/(platform)/library/components/AgentBriefingPanel/CostsBreakdown/CostsBreakdown.tsx
```diff
@@ -0,0 +1,119 @@
...
+    return (
+      <section className="mt-6">
+        <Text variant="body-medium" className="text-neutral-700">
+          Cost breakdown
+        </Text>
+        <Text variant="body" className="mt-1 text-neutral-500">
+          Couldn&apos;t load cost breakdown.
+        </Text>
+      </section>
+    );
```
0ubbe, line 72: 🤖 🟡 **Nice to Have**: The new error branch (`if (isError) return <section>...Couldn't load cost breakdown...</section>`) has no test. `BriefingTabContent.test.tsx` covers loading→success and the empty-state, but not the error path — meaning the user-visible "Couldn't load" message could silently disappear in a future refactor. Cheap fix: add one more case in the `CostsBreakdown` describe block that mocks `useGetV1UserCostSummary` with `{ data: undefined, isLoading: false, isError: true }` and asserts `screen.getByText(/Couldn't load cost breakdown/i)`. Also relevant for codecov-patch since this branch counts as uncovered new lines.

0ubbe: 🤖 Fixed in [eafbdaa2e](https://github.com/Significant-Gravitas/AutoGPT/commit/eafbdaa2e77e7f4b90372d89b2c5973c6b3ce1bd): added `it("surfaces an inline error when the cost-summary endpoint fails")` to BriefingTabContent.test.tsx — mocks `useGetV1UserCostSummary` with `{ isLoading: false, isError: true }` and asserts `screen.getByText(/Couldn't load cost breakdown/i)`. Local run: 12/12 pass.

autogpt_platform/backend/backend/data/execution.py
```diff
@@ -725,6 +726,168 @@ async def get_graph_executions_paginated(
...
+    failed_cost_cents: int
+    by_agent: list[UserAgentCostRollup]
+    top_runs: list[UserTopRun]
+    daily: list[UserDailyCost]
+
+
+_MAX_BY_AGENT_ROWS = 50
+_MAX_TOP_RUNS = 50
+
+
```
ntindle: athese don't really belong here

ntindle (changes requested):

autogpt_platform/frontend/src/app/(platform)/library/hooks/useLibraryFleetSummary.ts
```diff
@@ -42,12 +44,31 @@ function toTimestamp(value?: string | Date | null): number | null {
...
+): number {
+  const monthStart = startOfCurrentMonthUTC();
+  let total = 0;
+  for (const exec of executions) {
+    const startedTs = toTimestamp(exec.started_at);
+    if (startedTs === null || startedTs < monthStart) continue;
+    const cost = exec.stats?.cost;
+    if (typeof cost === "number" && Number.isFinite(cost)) {
+      total += cost;
+    }
```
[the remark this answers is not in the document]
0ubbe: 🤖 Already fixed in [e899efb96](https://github.com/Significant-Gravitas/AutoGPT/commit/e899efb96eadbf461a2dfde30e9ff7c1dc4ff35b): `fallbackMonthlySpendCents` was dropped entirely (so was the helper `startOfCurrentMonthUTC` and `toTimestamp`). The hook now uses only the server-side total — `costSummary?.total_cents ?? 0` — so no `started_at` vs `createdAt` mismatch is possible. The brief $0.00 → real-value transition is cheaper than the wrong-then-right flash the fallback caused.

autogpt_platform/backend/backend/data/execution.py
0ubbe: 🤖 Fixed in [84113108a](https://github.com/Significant-Gravitas/AutoGPT/commit/84113108a4a7daa78e1ef0ec4e7faca791b5f327): moved `UserAgentCostRollup`, `UserTopRun`, `UserDailyCost`, `UserExecutionCostSummary`, the `_MAX_*` constants, and `get_user_cost_summary` into a new `backend/data/execution_cost_summary.py` (matches the existing `execution_cost_summary_test.py` test file). Updated imports in `api/features/v1.py` and both test files; dropped the now-unused `asyncio` and `date` imports from `execution.py`.

autogpt_platform/backend/backend/data/execution.py
```diff
@@ -725,6 +726,168 @@ async def get_graph_executions_paginated(
...
+        ' AND "createdAt" <= $3::timestamp'
+        " AND COALESCE((stats->>'is_dry_run')::boolean, false) = false"
+    )
+    params = (user_id, since, until)
+
+    totals_rows, by_agent_rows, top_runs_rows, daily_rows = await asyncio.gather(
+        query_raw_with_schema(
+            "SELECT"
+            "  COALESCE(SUM((stats->>'cost')::numeric), 0)::bigint AS total_cents,"
+            "  COUNT(*)::bigint AS run_count,"
```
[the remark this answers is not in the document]
0ubbe: 🤖 Fixed in [84113108a](https://github.com/Significant-Gravitas/AutoGPT/commit/84113108a4a7daa78e1ef0ec4e7faca791b5f327): added a `billable_run_count` column to the totals aggregator (`COUNT(*) FILTER (WHERE COALESCE((stats->>'cost')::numeric, 0) > 0)`) and updated `CostsBreakdown.tsx` to divide `total_cents` by `billable_run_count` instead of `run_count`. The "Runs" tile keeps showing `run_count` (all executions in the window) — only the average denominator changed. Locked with new tests: `test_aggregates_by_agent_and_top_runs` now mixes a zero-cost run and asserts `run_count=6, billable_run_count=5`, and a new frontend test "computes Avg / run from billable_run_count, not run_count" proves the denominator switch.

0ubbe:
@ntindle can you re-review?

ntindle (approved):

autogpt_platform/frontend/src/app/(platform)/auth/integrations/oauth_callback/types.ts
```diff
@@ -6,6 +6,10 @@ export type OAuthPopupResultMessage = { message_type: "oauth_popup_result" } & (
     }
   | {
       success: false;
+      // Optional — present when the callback received a `state` token even
+      // though it could not produce a successful result. Listeners filter
+      // payloads by state, so an error without it would be silently dropped.
+      state?: string;
```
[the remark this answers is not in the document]
0ubbe: Not a bug — this is intentional and documented inline. The `state`-keyed listener filter (`oauth-popup.ts:132`) exists to prevent cross-flow contamination: BroadcastChannel and localStorage are shared across concurrent OAuth flows on the same origin, so an error message without a state token can't be routed to a specific flow without risking resolving the wrong listener.

The only path where a user hangs is an OAuth provider redirecting back with **neither** `code` nor `state` — providers in the wild preserve `state` on error redirects per [RFC 6749 §4.1.2.1](https://www.rfc-editor.org/rfc/rfc6749#section-4.1.2.1). When `state` *is* present, `route.ts:34` echoes it back so the error reaches the right listener. The tradeoff (drop rare stateless errors vs. risk cross-flow contamination) is called out in [`types.ts:9-11`](https://github.com/Significant-Gravitas/AutoGPT/blob/feature/agent-briefing-costs/autogpt_platform/frontend/src/app/(platform)/auth/integrations/oauth_callback/types.ts#L9-L11) and [`route.ts:31-33`](https://github.com/Significant-Gravitas/AutoGPT/blob/feature/agent-briefing-costs/autogpt_platform/frontend/src/app/(platform)/auth/integrations/oauth_callback/route.ts#L31-L33). The 5-minute timeout (`OAUTH_ERROR_FLOW_TIMED_OUT`) backstops the edge case.

PR #12618, opened 2026-03-31
feat(platform): copilot-bot (Python / discord.py)

Bentlybro:
## Why

AutoPilot needs to reach users on chat platforms — Discord first, Telegram / Slack / Teams / WhatsApp next. This PR adds the bot service that bridges those platforms to the AutoPilot backend via the `PlatformLinkingManager` AppService introduced in #12615.

Two independent linking flows (see #12615 for the rationale):

- **SERVER links**: first person to run `/setup` in a guild claims it. Anyone in the server can mention the bot; all usage bills to the owner.
- **USER links**: an individual DMs the bot, links their personal account, DMs bill to their own AutoPilot. A server owner still has to link their DMs separately.

## What

A Python service using `discord.py`, living alongside the rest of the backend. Connects to the platform_linking service via cluster-internal RPC (no shared bearer token) and subscribes to copilot streams directly on Redis (no HTTP SSE proxy).

Originally prototyped in Node.js with Vercel's Chat SDK — rewritten in Python after team feedback: the rest of the platform is Python, `discord.py` was already a dependency, and the Chat SDK's streaming-UI abstractions don't apply to a headless chat bot.

### Deployment

- **Shares the existing backend Docker image** — no separate Dockerfile, no separate Artifact Registry. A `copilot-bot` poetry script entry lets the same image run with `command: ["copilot-bot"]` in the Helm chart.
- **Auto-starts with `poetry run app`** when `AUTOPILOT_BOT_DISCORD_TOKEN` is set, so the full local dev stack includes the bot without extra setup.
- **Runs standalone** via `poetry run copilot-bot` for the production pod.

Infra PR: [AutoGPT_cloud_infrastructure#310](https://github.com/Significant-Gravitas/AutoGPT_cloud_infrastructure/pull/310).

### File layout

```
backend/copilot/bot/
├── app.py              # CoPilotChatBridge(AppService) + adapter factory + outbound @expose
├── config.py           # Shared (platform-agnostic) config
├── handler.py          # Core logic: routing, linking, batched streaming
├── platform_api.py     # Thin facade over PlatformLinkingManagerClient + stream_registry
├── platform_api_test.py
├── text.py             # split_at_boundary + format_batch
├── threads.py          # Redis-backed thread subscription tracking
├── README.md
└── adapters/
    ├── base.py         # PlatformAdapter ABC + MessageContext
    └── discord/
        ├── adapter.py  # Gateway connection, events, thread creation, buttons
        ├── commands.py # /setup, /help, /unlink
[...]

PR #13130, opened 2026-05-15
feat(backend/copilot-bot): add WebhookAdapter base + mount routes on main API

Bentlybro:
## Why

The next chat platforms — Slack, Telegram, Teams, WhatsApp — are webhook-based. They POST inbound events over HTTPS rather than holding a persistent connection like Discord's Gateway. The current `PlatformAdapter` ABC was shaped around the socket model (`start`/`stop` lifecycle); there's no place for inbound webhook routes to mount.

Without this scaffolding, the upcoming Slack PR would mix three concerns: refactor the ABC, add the route-mount machinery, **and** add the Slack adapter itself. This PR isolates the foundation so the Slack PR is purely "add Slack."

## What

- **Split `PlatformAdapter` into three classes:**
  - `PlatformAdapter` — shared outbound contract (`send_message`, `send_link`, `create_thread`, etc.) used by the platform-agnostic `MessageHandler`
  - `SocketAdapter(PlatformAdapter)` — adds `start`/`stop` (socket-owning connectors — Discord today, Slack Socket Mode hypothetically)
  - `WebhookAdapter(PlatformAdapter)` — adds `register_routes(app)` (HTTPS-driven connectors — Slack Events API, Telegram, Teams, WhatsApp)
- `DiscordAdapter` moves under `SocketAdapter` — base class only, **zero logic change**
- `app.py::_build_adapters` → `_build_socket_adapters` (type-narrowed to `list[SocketAdapter]`)
- New `backend/copilot/bot/webhook_routes.py` — `register_webhook_adapters(app)` helper + empty `_build_webhook_adapters` registry
- One-line call from `backend/api/rest_api.py` to mount webhook adapter routes during app setup
- README updated with the connector taxonomy + revised "Adding a new platform" guide

## How

**Where webhook routes live.** Webhook adapters mount onto the main backend API (`AgentServer`) rather than a new dedicated service:

- The main API is already the public ingress surface with N stateless replicas — exactly what webhook receivers need.
- Mirrors how OAuth callbacks and block webhook triggers already work on the main API.
- The original reason `copilot-bot` is its own pod — the persistent Discord Gateway connection — does **not** apply to stateless HTTPS handlers. A separate "webhook bridge" pod would be speculative infrastructure with no current consumer.

**Locality preserved.** Per-platform code stays under `adapters/<platform>/`. The webhook mount helper iterates `_build_webhook_adapters(api)` and calls `adapter.register_routes(app)` on each — every adapter owns its own routes and signature verification, in line with the existing locality rule.

**Resilient empty path.** With no webhook adapters configured (the state on merge), `register_webhook_adapters` logs `Mounted 0 webhook adapter(s)` and returns. The API starts cleanly with or without webhook platforms configured — same property as the existing `CoPilotChatBridge` empty-adapter path.

**Synchronous startup wiring.** The mount call happens at `rest_api.py` module-level alongside the existing `app.include_router(...)` calls. Errors propagate visibly at startup — no fire-and-forget pattern (lesson from the `CoPilotChatBridge` adapter-task review in #12618).

## Tests

- `adapters/base_test.py` — subclass-hierarchy regression: `SocketAdapter` / `WebhookAdapter` both extend `PlatformAdapter`; `DiscordAdapter` is now a `SocketAdapter`.
- `webhook_routes_test.py` — empty registry returns `[]`; `register_webhook_adapters` correctly wires `on_message` and `register_routes` on each adapter.
- All 116 existing `backend/copilot/bot/` tests still pass — the base-class split is non-breaking.

## Stack

- This PR → **Slack via webhook (Events API)** — adds `adapters/slack/`, registers `SlackAdapter(api)` in `_build_webhook_adapters` (one-line registry addition), zero `rest_api.py` changes needed.
- Then **Telegram**, **Teams**, **WhatsApp** — each is its own adapter folder + one registry entry.
- Companion infra PR for sealed secrets per platform — no new helm chart, since routes mount on the existing main backend pod.

### DIFF
--- a/autogpt_platform/backend/backend/api/rest_api.py
+++ b/autogpt_platform/backend/backend/api/rest_api.py
@@ -83,6 +83,10 @@ logger = logging.getLogger(__name__)
 
 logging.getLogger("autogpt_libs").setLevel(logging.INFO)
 
+# Owned at module level so `lifespan_context` can close it on shutdown;
+# routes are mounted further down once `app` exists.
+_webhook_bot_backend = BotBackend()
+
 
 @contextlib.contextmanager
 def launch_darkly_context():
@@ -425,7 +429,6 @@ app.include_router(
     prefix="/api/platform-linking",
 )
 
-_webhook_bot_backend = BotBackend()
 register_webhook_adapters(app, _webhook_bot_backend)
 
 app.mount("/external-api", external_api)
--- a/autogpt_platform/backend/backend/api/rest_api.py
+++ b/autogpt_platform/backend/backend/api/rest_api.py
@@ -53,6 +53,7 @@ from backend.api.features.library.exceptions import (
     FolderValidationError,
 )
 from backend.blocks.llm import DEFAULT_LLM_MODEL
+from backend.copilot.bot.bot_backend import BotBackend
 from backend.copilot.bot.webhook_routes import register_webhook_adapters
 from backend.copilot.rate_limit import UserPaywalledError
 from backend.data.model import Credentials
@@ -156,6 +157,11 @@ async def lifespan_context(app: fastapi.FastAPI):
     # Each cleanup is wrapped so one failure doesn't block the rest. The
     # Redis close in particular silences asyncio's "Unclosed ClusterNode"
     # GC warning at interpreter shutdown.
+    try:
+        await _webhook_bot_backend.close()
+    except Exception:
+        logger.warning("webhook BotBackend.close() failed", exc_info=True)
+
     try:
         await backend.data.redis_client.disconnect_async()
     except Exception:
@@ -419,7 +425,8 @@ app.include_router(
     prefix="/api/platform-linking",
 )
 
-register_webhook_adapters(app)
+_webhook_bot_backend = BotBackend()
+register_webhook_adapters(app, _webhook_bot_backend)
 
 app.mount("/external-api", external_api)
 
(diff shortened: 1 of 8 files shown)


autogpt_platform/backend/backend/api/rest_api.py
```diff
@@ -81,6 +83,10 @@
 
 logging.getLogger("autogpt_libs").setLevel(logging.INFO)
 
+# Owned at module level so `lifespan_context` can close it on shutdown;
+# routes are mounted further down once `app` exists.
+_webhook_bot_backend = BotBackend()
```
ntindle, line 88: ngl this feels sus as hell. do we follow this pattern already anywhere?

ntindle: shuld this be in the lifespan context with a yield?

ntindle: or is this supposed to be a singleton thats managed elsewhere as a client?

autogpt_platform/backend/backend/api/rest_api.py
```diff
@@ -418,6 +429,8 @@ async def validation_error_handler(
     prefix="/api/platform-linking",
 )
 
+register_webhook_adapters(app, _webhook_bot_backend)
```
ntindle, line 432: I can't remember the behavior here of inline pass as ref vs pass as copy but have an ai check this

ntindle (changes requested):
Good start lets just chekcin on the lifecycel stuyff

Bentlybro:
Superseded — the Socket/Webhook adapter split landed on `dev` via #13505 in July. `adapters/base.py` and `webhook_routes.py` already carry this, so there's nothing left to merge. Closing.

PR #13156, opened 2026-05-18
fix: propagate tool credentials to orchestrator when run from Library/AutoPilot (#13144)

### LINKED ISSUE #13144: Orchestrator tool credentials missing when run from Library or AutoPilot (works from Builder)
## Summary

When an agent built with an orchestrator includes a tool block that requires credentials, the credentials are only resolved when the agent is launched from the **Builder**. Launching the same agent from the **Library** or from **AutoPilot** fails with a missing-credentials error.

Note: this is a single agent — the tool is a regular block attached to the orchestrator, not a sub-agent invoked via `AgentExecutorBlock`.

This is happening in **production**.

## Steps to reproduce

1. Build an agent that uses an orchestrator block.
2. Attach a tool block to the orchestrator that requires credentials (e.g. any API-key-backed tool).
3. Launch the agent from the Builder → works as expected, credentials are passed through.
4. Launch the same agent from the Library → fails: tool reports missing credentials.
5. Launch the same agent from AutoPilot → fails: tool reports missing credentials.

## Expected

Credentials configured on the orchestrator's tool block should be resolved consistently regardless of where the run is triggered from (Builder, Library, AutoPilot).

## Actual

Credentials are only passed through when launched from the Builder. From the Library or AutoPilot, the tool reports it doesn't have credentials.

## Notes / suspicions

* The Builder path likely injects/serializes credentials for orchestrator-attached tool blocks differently than the Library/AutoPilot run paths.
* Worth investigating how the orchestrator's tool-block credentials are persisted on the graph vs. resolved at execution time, and whether non-Builder entry points are missing a credential-hydration step for those nested tool inputs.
* Production impact: any user running an orchestrator-with-tool agent outside the Builder is blocked.

## Reporter

Krzysztof Czerwinski

paulxg12:
Closes #13144

**Root cause**: Orchestrator tool execution bypassed the normal graph execution queue where credential metadata is merged into node inputs. Library/AutoPilot use orchestrated agent flows where sub-tools depend on this credential merge.

**Fix**: Stored `nodes_input_masks` on the ExecutionProcessor instance and made the orchestrator read + merge credential metadata before tool execution, matching the normal queue dispatch behavior.

### DIFF
--- a/autogpt_platform/backend/backend/blocks/orchestrator.py
+++ b/autogpt_platform/backend/backend/blocks/orchestrator.py
@@ -1079,6 +1079,16 @@ class OrchestratorBlock(Block):
         # so the execution record is complete for from_db() reconstruction.
         merged_input_data = {**target_node.input_default, **raw_input_data}
 
+        # Also merge credential metadata from the graph execution's
+        # nodes_input_masks so tool nodes get their credentials even when
+        # launched by the orchestrator (e.g. from Library or AutoPilot).
+        # Without this, the credential metadata is only merged during the
+        # normal graph execution queue dispatch (manager.py:_on_graph_execution)
+        # but orchestrator tool calls bypass that path.
+        node_creds = getattr(execution_processor, "nodes_input_masks", None)
+        if node_creds and sink_node_id in node_creds:
+            merged_input_data.update(node_creds[sink_node_id])
+
         # Add all inputs to the execution
         if not merged_input_data:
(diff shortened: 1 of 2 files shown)


paulxg12:
I have read the CLA Document and I hereby sign the CLA

kcze:
The root cause for the bug explained in the linked issue was found to be incorrect.

Issue #9741, opened 2025-04-02
Unhandled Runtime Error and Missing Agent Blocks


jani339:
### ⚠️ Search for existing issues first ⚠️

- [x] I have searched the existing issues, and there is no existing issue for my problem

### Which Operating System are you using?

Windows

### Which version of AutoGPT are you using?

Latest Release

### What LLM Provider do you use?

Azure

### Which area covers your issue best?

Installation and setup

### What commit or version are you using?

c6703dd89128b59f1fab79dc38eaa8e24009adce

### Describe your issue.

I installed the Auto GPT Platform as explained in the Docs (https://docs.agpt.co/platform/getting-started/)
Everything is running, but i get always the following issue popping up.

![Image](https://github.com/user-attachments/assets/f7d261da-80f9-4c1c-8ccb-fede0afb368a)

Addtionally my Agent Blocks are missing 

![Image](https://github.com/user-attachments/assets/ea5fa72b-237d-4bc2-8ef6-8f6b05a2f5d2)

i already tried reinstalling it --> Same issues

Thamari966:
Try below steps to debug 

1. Verify API Endpoint

- Log the request URL before calling fetch():
`console.log("Fetching from URL:", url);
`
- Open Developer Tools and check the URL.

- Open the URL in a browser to see if it's accessible.
2. Handle CORS Issues

- If the API is on a different domain, check CORS policies.

- Try running the following command:
`next dev --hostname 0.0.0.0
`
3. Ensure API is Running

- If the API is local (e.g., http://localhost:3000/api), make sure it's running.
4. Check API Response

- Open Developer Tools to inspect failed requests.
5. Ensure API Accepts JSON

- Make sure the API supports JSON payloads since the request includes Content-Type: application/json
6. Improve Error Handling

- Wrap fetch() in a try...catch block to catch errors:
`try {
    const response = await fetch(url, {
        method,
        headers: {
            ...(hasRequestBody && { "Content-Type": "application/json" }),
        },
    });

    if (!response.ok) {
        throw new Error(`HTTP error! Status: ${response.status}`);
    }

    const data = await response.json();
    console.log("Fetched data:", data);
} catch (error) {
    console.error("Fetch error:", error);
}
`
7. Check Firewall / VPN
- Some VPNs or firewalls might block requests. Disable them temporarily and retry.

**Fix for Missing Agent Blocks**

1. Check Local Storage

- AutoGPT may store data in localStorage. Run:
[...]

PR #13157, opened 2026-05-19
fix: guard LLM response against empty choices and replace bare except clauses

qizwiz:
## Summary

- **`autogpt/llm_utils.py`**: Add `if not response.choices: raise ValueError(...)` before `response.choices[0].message["content"]` — prevents `IndexError` when the OpenAI API returns an empty `choices` list (content-filtering, quota, streaming edge cases). Related: #9741
- **`autogpt/utils.py`**: Replace two bare `except:` with `except Exception:` — avoids silently swallowing `SystemExit` and `KeyboardInterrupt`
- **`tests/`**: Same bare-except cleanup in three test files

## Why

The OpenAI API can return `choices=[]` when content is filtered, a quota limit is hit, or during streaming edge cases. Without a guard, `response.choices[0]` raises `IndexError` instead of a meaningful error.

Bare `except:` catches `SystemExit` and `KeyboardInterrupt`, making the process unresponsive to Ctrl-C — replacing with `except Exception:` is the safer default.

## Detection

Found by [pact](https://github.com/qizwiz/pact) static analysis — `llm_response_unguarded` and `bare_except` failure modes.

## Test plan

- [ ] `python -m pytest tests/` passes
- [ ] Content-filter response path now raises `ValueError` with a clear message

### DIFF
--- a/autogpt/llm_utils.py
+++ b/autogpt/llm_utils.py
@@ -145,6 +145,8 @@ def create_chat_completion(
             raise RuntimeError(f"Failed to get response after {num_retries} retries")
         else:
             quit(1)
+    if not response.choices:
+        raise ValueError("LLM returned empty response")
     resp = response.choices[0].message["content"]
     for plugin in CFG.plugins:
         if not plugin.can_handle_on_response():
--- a/autogpt/utils.py
+++ b/autogpt/utils.py
@@ -50,7 +50,7 @@ def get_bulletin_from_web() -> str:
         )
         if response.status_code == 200:
             return response.text
-    except:
+    except Exception:
         return ""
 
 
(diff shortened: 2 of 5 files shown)


qizwiz:
The upstream `dev` branch has restructured — the files this PR modifies (`autogpt/llm_utils.py`, `autogpt/utils.py`) have been moved to `classic/original_autogpt/`. This branch no longer rebases cleanly onto `dev`. I'm investigating whether to retarget the fixes to the new file locations or whether this module is actively maintained there.

qizwiz:
Closing: the upstream repo has been completely restructured and the files this PR modified (`autogpt/llm_utils.py`, `autogpt/utils.py`) no longer exist in master. The fix would need to be re-applied to the new project structure.

PR #13160, opened 2026-05-19
fix(frontend): skip onboarding paywall for users already on a paid tier

0ubbe:
### Why / What / How

**Why**

Paid users — including admin-granted accounts (`subscriptionTier != NO_TIER`, `stripeCustomerId IS NULL`) and accounts that pre-date the `VISIT_COPILOT` step — get force-redirected to `/onboarding` on every page load and end up at the paywall step. Reported in Discord ([Investigation: Paid Users Hitting Paywall?](https://discord.com/channels/1126875755960336515/1505948249406832640)).

The chain:

1. `OnboardingProvider` redirects any user with `is_completed === false` to `/onboarding` (`src/providers/onboarding/onboarding-provider.tsx:139`).
2. Backend derives `is_completed` solely from `VISIT_COPILOT in completedSteps` (`backend/api/features/v1.py:348`). `VISIT_COPILOT` is only written by the new wizard's Preparing step, so any pre-wizard account stays "not completed" forever.
3. `SubscriptionStep` shows plan cards unconditionally — no awareness of the current tier — and the only exit is Stripe Checkout. For admin-granted users (no Stripe sub) the backend's same-tier short-circuit doesn't fire (`v1.py:1024-1033`), so clicking Pro starts a *new* checkout.

**What**

`SubscriptionStep` is now skipped entirely when the user is already on any tier other than `NO_TIER`. Welcome → Role → PainPoints → Preparing (no paywall), then `VISIT_COPILOT` is recorded normally and the user lands on `/copilot`.

This is the defence-in-depth half of the fix. The other half — backfilling `VISIT_COPILOT` for existing paying users — is a separate one-shot DB update.

**How**

In `useOnboardingPage`:
- Call `useGetSubscriptionStatus` to read the user's current tier.
- Snapshot the tier on first resolve (mirroring the existing `paymentEnabledSnapshot` pattern) so completing the SubscriptionStep mid-session (`NO_TIER → PRO`) can't retroactively shrink the wizard under a user mid-flow.
- When `tierSnapshot != null && tierSnapshot != "NO_TIER"`, treat `isPaymentEnabled` as `false` regardless of the LD flag — `totalSteps = 3`, `preparingStep = 4`, same shape as the payments-off path.
- Gate URL-init and URL-sync effects on subscription readiness in addition to LD readiness, so the ceiling used to clamp `?step=` reflects the right `preparingStep`. Surface that in the returned `isLoading`.
- On query error: `tierSnapshot` stays null → `userHasActivePlan = false` → existing behaviour (paywall shown). Fail-safe is the status quo.

### Changes 🏗️

- `src/app/(no-navbar)/onboarding/useOnboardingPage.ts` — fetch subscription, snapshot tier, fold result into `isPaymentEnabled` / `totalSteps` / `preparingStep`; gate URL effects on `subscriptionReady`.
- `src/app/(no-navbar)/onboarding/__tests__/page.test.tsx` — add a `useGetSubscriptionStatus` mock and three new tests: paid user skips SubscriptionStep, paid user's `?step=5` clamps to `preparingStep=4`, NO_TIER user still sees SubscriptionStep when payments are enabled.

#### For code changes:
- [x] I have made a test plan
  - [x] `pnpm types` clean
  - [x] `pnpm lint` clean (only pre-existing warnings unrelated to this PR)
  - [x] `pnpm test:unit` — all 2749 tests pass (3 new tests in `page.test.tsx`)
  - [ ] Manual prod-like verification by ops once the existing-paying-users DB backfill lands alongside this PR (the frontend change is the defence-in-depth; the backfill unsticks already-trapped accounts).


Pwuts (dismissed):
tysm, LGTM :)

PR #13173, opened 2026-05-20
fix(frontend): preserve cursor position when editing notes

sisyphusse1-ops:
### Why / What / How

**Why:** Editing text in a sticky note block causes the cursor to jump to the bottom of the textarea on every keystroke, making it impossible to edit text at the top or middle of the note. Reported in #9252.

**What:** Preserve the cursor (caret) position in the notes textarea across re-renders triggered by state updates.

**How:** When the user types in the note textarea, we capture `selectionStart` into a ref before propagating the change event. After React re-renders with the new value (which normally resets cursor to end), a `useEffect` restores the saved cursor position from the ref. This is a standard pattern for controlled textareas that receive their value from external state.

### Changes 🏗️

- `autogpt_platform/frontend/src/components/renderers/InputRenderer/base/standard/widgets/TextInput/TextWidget.tsx`:
  - Added `useRef` and `useEffect` imports
  - Added `noteRef` (textarea element ref) and `cursorPosRef` (cursor position ref)
  - Added a post-render `useEffect` that restores cursor position for NOTE-type inputs
  - Created `handleNoteChange` wrapper that saves `selectionStart` before calling the existing `handleChange`
  - Passed `ref={noteRef}` to the Note `<Input>` component

#### For code changes:
- [x] I have made a test plan
  - [x] `pnpm format` passes with no changes
  - [x] `pnpm types` passes with no errors
  - [x] `pnpm lint` passes with no new warnings/errors
  - [x] Verified the fix logic: cursor position is saved before state update and restored after re-render

### DIFF
--- a/autogpt_platform/frontend/src/components/renderers/InputRenderer/base/standard/widgets/TextInput/TextWidget.tsx
+++ b/autogpt_platform/frontend/src/components/renderers/InputRenderer/base/standard/widgets/TextInput/TextWidget.tsx
@@ -1,6 +1,6 @@
 "use client";
 
-import { useState } from "react";
+import { useRef, useState, useEffect } from "react";
 import { WidgetProps } from "@rjsf/utils";
 import {
   InputType,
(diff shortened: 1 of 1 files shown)


sisyphusse1-ops:
Closing this PR. Thanks.

PR #13177, opened 2026-05-21
fix(backend): teach LlmModel about OpenRouter Anthropic aliases + dry-run sim default Haiku OR-slug

majdyz:
### Why / What / How

**Why:** PR #13171 (SECRT-2368) flipped the dry-run simulator default to `LlmModel.CLAUDE_4_5_HAIKU.value` = `"claude-haiku-4-5-20251001"`. That broke every dry-run on dev because the LLM-simulation path (`_call_llm_for_simulation`) hits OpenRouter's OpenAI-compat endpoint (`/api/v1/chat/completions`), which only accepts canonical `<vendor>/<model>` slugs and returns HTTP 400 `"<id> is not a valid model ID"` for direct-Anthropic snapshot IDs. User-visible: `[SIMULATOR ERROR — NOT A BLOCK FAILURE] Failed after 5 attempts: Error code: 400 - 'claude-haiku-4-5-20251001 is not a valid model ID'`.

**Verified empirically against OpenRouter:**

| Slug | `/v1/chat/completions` (OpenAI-compat) | `/v1/messages` (Anthropic-compat) |
|---|---|---|
| `claude-haiku-4-5-20251001` (snapshot) | ❌ 400 not valid | ✅ accepted |
| `anthropic/claude-haiku-4-5` (OR slug) | ✅ accepted | ✅ accepted |
| `anthropic/claude-haiku-4-5-20251001` (prefix + snapshot) | ❌ 400 not valid | — |

So the right default for the simulator is the **OpenRouter slug** (`anthropic/claude-haiku-4-5`), and we need `LlmModel` to resolve that string back to the existing `CLAUDE_4_5_HAIKU` enum member so `OrchestratorBlock.Input.model` validation still passes.

**What:**
- Add `_OPENROUTER_ALIASES: dict[str, str]` in `backend/blocks/llm.py` mapping the three Claude 4.5 OpenRouter slugs (`anthropic/claude-haiku-4-5`, `anthropic/claude-opus-4-5`, `anthropic/claude-sonnet-4-5`) to their direct-Anthropic snapshot enum values. The newer 4.6/4.7 entries don't carry the snapshot suffix, so the existing prefix-strip path already covers them.
- Extend `LlmModel._missing_` to consult `_OPENROUTER_ALIASES` first, then fall back to the existing provider-prefix strip.
- Change `_DEFAULT_SIMULATOR_MODEL` (in `backend/executor/simulator.py`) and `ChatConfig.simulation_model` default (in `backend/copilot/config.py`) to the bare string `"anthropic/claude-haiku-4-5"`. Both files drop their now-unused `LlmModel` import.
- Update `TestDefaultSimulatorModel` to pin the new default + assert it resolves to `LlmModel.CLAUDE_4_5_HAIKU` via the alias map, plus a `"/" in default` guard so a future revert to a snapshot ID trips at unit-test time.
- New `TestLlmModelMissingHandler` covering: alias-map resolution, prefix-strip fallback, and unknown-slug raising.

**How:**
- **LLM-simulation path** (`_call_llm_for_simulation`): sends `"anthropic/claude-haiku-4-5"` straight to OpenRouter's OpenAI-compat endpoint via `get_openai_client(prefer_openrouter=True)` — accepted.
- **Orchestrator dry-run path** (`prepare_dry_run` injects `"anthropic/claude-haiku-4-5"` as `model` on `OrchestratorBlock.Input`): Pydantic calls `LlmModel("anthropic/claude-haiku-4-5")`, hits `_missing_`, alias map returns `CLAUDE_4_5_HAIKU`. The orchestrator's Anthropic SDK then sends the enum's `.value` (`"claude-haiku-4-5-20251001"`) to OpenRouter's Anthropic-compat endpoint, which accepts both formats — also fine.

**Why not duplicate enum entries?** Per reviewer feedback ("this is weird having multiple models"): adding `CLAUDE_4_5_HAIKU_OPENROUTER` alongside the existing direct-Anthropic entry would duplicate metadata + cost lookups for the same model and pull the OpenRouter/direct routing distinction into the enum itself. The alias map keeps `LlmModel` as one entry per model while still letting both identifiers resolve through the validator.

### Changes 🏗️

- `_OPENROUTER_ALIASES` map + extended `_missing_` in `LlmModel`.
- Dry-run simulator default flipped from Haiku snapshot ID to Haiku OpenRouter slug.
- New unit tests for the alias-resolution path; updated `TestDefaultSimulatorModel` pin.
- No `LlmModel` enum members added or removed → no `llm.md` regen needed.

#### For code changes:
- [x] I have made a test plan
  - [x] `poetry run pytest backend/executor/simulator_test.py::TestDefaultSimulatorModel backend/executor/simulator_test.py::TestPrepareDryRun backend/copilot/tools/test_dry_run.py::test_prepare_dry_run_orchestrator_block` — all green
  - [x] `poetry run pytest backend/blocks/test/test_llm.py::TestLlmModelMissingHandler` — 3 new tests, all green
  - [x] Empirical OpenRouter probes recorded above
  - [x] Direct e2e check: `LlmModel("anthropic/claude-haiku-4-5") is LlmModel.CLAUDE_4_5_HAIKU` ✓; `LlmModel("anthropic/claude-opus-4-7") is LlmModel.CLAUDE_4_7_OPUS` ✓
  - [x] `poetry run ruff format` + `poetry run ruff check` clean on all changed files

### DIFF
--- a/autogpt_platform/backend/backend/blocks/llm.py
+++ b/autogpt_platform/backend/backend/blocks/llm.py
@@ -116,11 +116,44 @@ class LlmModelMeta(EnumMeta):
     pass
 
 
+# OpenRouter exposes Anthropic models under canonical ``anthropic/<model>``
+# slugs that drop the snapshot-date suffix Anthropic's own API uses
+# (``claude-haiku-4-5-20251001`` → ``anthropic/claude-haiku-4-5``).  The
+# generic provider-prefix strip in ``_missing_`` can't reverse the date
+# truncation, so map the OpenRouter slugs to the canonical direct-Anthropic
+# enum values here.  Only models that exist on OpenRouter need entries; the
+# newer 4.6/4.7 enum values already lack a snapshot date so the prefix
+# strip alone is enough for them.
+_OPENROUTER_ALIASES: dict[str, str] = {
+    "anthropic/claude-haiku-4-5": "claude-haiku-4-5-20251001",
+    "anthropic/claude-opus-4-5": "claude-opus-4-5-20251101",
+    "anthropic/claude-sonnet-4-5": "claude-sonnet-4-5-20250929",
+}
+
+
 class LlmModel(str, Enum, metaclass=LlmModelMeta):
     @classmethod
     def _missing_(cls, value: object) -> "LlmModel | None":
-        """Handle provider-prefixed model names like 'anthropic/claude-sonnet-4-6'."""
-        if isinstance(value, str) and "/" in value:
+        """Resolve provider-prefixed model names.
+
+        Handles two shapes:
+
+        1. OpenRouter aliases for Anthropic models whose direct-API
+           identifier carries a snapshot suffix that the OpenRouter slug
+           drops — e.g. ``anthropic/claude-haiku-4-5`` ↔ enum value
+           ``claude-haiku-4-5-20251001``.  Looked up via
+           ``_OPENROUTER_ALIASES``.
+        2. Generic provider prefix strip — e.g.
+           ``anthropic/claude-sonnet-4-6`` → ``claude-sonnet-4-6``.
+        """
+        if not isinstance(value, str):
+            return None
+        if value in _OPENROUTER_ALIASES:
+            try:
+                return cls(_OPENROUTER_ALIASES[value])
+            except ValueError:
+                return None
+        if "/" in value:
             stripped = value.split("/", 1)[1]
             try:
                 return cls(stripped)
--- a/autogpt_platform/backend/backend/blocks/llm.py
+++ b/autogpt_platform/backend/backend/blocks/llm.py
@@ -8,6 +8,7 @@ import secrets
 from abc import ABC
 from enum import Enum, EnumMeta
 from json import JSONDecodeError
+from collections.abc import Mapping
 from typing import Any, Iterable, List, Literal, NamedTuple, Optional, cast
 
 import anthropic
@@ -116,21 +117,6 @@ class LlmModelMeta(EnumMeta):
     pass
 
 
-# OpenRouter exposes Anthropic models under canonical ``anthropic/<model>``
-# slugs that drop the snapshot-date suffix Anthropic's own API uses
-# (``claude-haiku-4-5-20251001`` → ``anthropic/claude-haiku-4-5``).  The
-# generic provider-prefix strip in ``_missing_`` can't reverse the date
-# truncation, so map the OpenRouter slugs to the canonical direct-Anthropic
-# enum values here.  Only models that exist on OpenRouter need entries; the
-# newer 4.6/4.7 enum values already lack a snapshot date so the prefix
-# strip alone is enough for them.
-_OPENROUTER_ALIASES: dict[str, str] = {
-    "anthropic/claude-haiku-4-5": "claude-haiku-4-5-20251001",
-    "anthropic/claude-opus-4-5": "claude-opus-4-5-20251101",
-    "anthropic/claude-sonnet-4-5": "claude-sonnet-4-5-20250929",
-}
-
-
 class LlmModel(str, Enum, metaclass=LlmModelMeta):
     @classmethod
     def _missing_(cls, value: object) -> "LlmModel | None":
@@ -142,17 +128,16 @@ class LlmModel(str, Enum, metaclass=LlmModelMeta):
            identifier carries a snapshot suffix that the OpenRouter slug
            drops — e.g. ``anthropic/claude-haiku-4-5`` ↔ enum value
            ``claude-haiku-4-5-20251001``.  Looked up via
-           ``_OPENROUTER_ALIASES``.
+           ``_OPENROUTER_ALIASES`` (defined below the class so it can hold
+           ``LlmModel`` members directly).
         2. Generic provider prefix strip — e.g.
            ``anthropic/claude-sonnet-4-6`` → ``claude-sonnet-4-6``.
         """
         if not isinstance(value, str):
             return None
-        if value in _OPENROUTER_ALIASES:
-            try:
-                return cls(_OPENROUTER_ALIASES[value])
-            except ValueError:
-                return None
+        alias = _OPENROUTER_ALIASES.get(value)
+        if alias is not None:
+            return alias
         if "/" in value:
             stripped = value.split("/", 1)[1]
             try:
@@ -292,6 +277,24 @@ class LlmModel(str, Enum, metaclass=LlmModelMeta):
     @property
     def max_output_tokens(self) -> int | None:
         return self.metadata.max_output_tokens
+
+
+# OpenRouter exposes Anthropic models under canonical ``anthropic/<model>``
+# slugs that drop the snapshot-date suffix Anthropic's own API uses
+# (``claude-haiku-4-5-20251001`` → ``anthropic/claude-haiku-4-5``). The
+# generic provider-prefix strip in ``_missing_`` can't reverse the date
+# truncation, so map the OpenRouter slugs to ``LlmModel`` members here.
+# Only models whose canonical enum value carries a ``-YYYYMMDD`` snapshot
+# suffix need entries; values without a snapshot (4.6/4.7+) are already
+# covered by the prefix-strip path alone. Stored as ``LlmModel`` instances
+# (not strings) so a rename or snapshot rotation on the enum follows the
+# alias automatically — a stale entry becomes a load-time ``AttributeError``
+# rather than a silent ``_missing_`` miss at runtime.
+_OPENROUTER_ALIASES: Mapping[str, LlmModel] = {
+    "anthropic/claude-haiku-4-5": LlmModel.CLAUDE_4_5_HAIKU,
+    "anthropic/claude-opus-4-5": LlmModel.CLAUDE_4_5_OPUS,
+    "anthropic/claude-sonnet-4-5": LlmModel.CLAUDE_4_5_SONNET,
+}
 
 
 MODEL_METADATA = {
--- a/autogpt_platform/backend/backend/blocks/llm.py
+++ b/autogpt_platform/backend/backend/blocks/llm.py
@@ -6,9 +6,9 @@ import math
 import re
 import secrets
 from abc import ABC
+from collections.abc import Mapping
 from enum import Enum, EnumMeta
 from json import JSONDecodeError
-from collections.abc import Mapping
 from typing import Any, Iterable, List, Literal, NamedTuple, Optional, cast
 
 import anthropic
--- a/autogpt_platform/backend/backend/blocks/test/test_llm.py
+++ b/autogpt_platform/backend/backend/blocks/test/test_llm.py
@@ -1,4 +1,5 @@
 import asyncio
+import re
 from typing import Any, cast
 from unittest.mock import AsyncMock, MagicMock, patch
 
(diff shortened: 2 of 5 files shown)


autogpt_platform/backend/backend/blocks/llm.py
```diff
@@ -116,11 +116,44 @@ class LlmModelMeta(EnumMeta):
...
+# generic provider-prefix strip in ``_missing_`` can't reverse the date
+# truncation, so map the OpenRouter slugs to the canonical direct-Anthropic
+# enum values here.  Only models that exist on OpenRouter need entries; the
+# newer 4.6/4.7 enum values already lack a snapshot date so the prefix
+# strip alone is enough for them.
+_OPENROUTER_ALIASES: dict[str, str] = {
+    "anthropic/claude-haiku-4-5": "claude-haiku-4-5-20251001",
+    "anthropic/claude-opus-4-5": "claude-opus-4-5-20251101",
+    "anthropic/claude-sonnet-4-5": "claude-sonnet-4-5-20250929",
+}
```
majdyz: 🤖 🟠 **Should Fix**: Two intertwined risks at this dict + the `try/except ValueError` in `_missing_`:

1. **Silent drift on snapshot rotation.** The right-hand strings here duplicate `LlmModel`'s snapshot suffixes (`claude-haiku-4-5-20251001` etc.). If Anthropic publishes a new snapshot and someone updates the `CLAUDE_4_5_HAIKU` enum value but forgets this map, `_missing_` does `cls(_OPENROUTER_ALIASES[value])` against a stale string, raises `ValueError`, and the `try/except` returns `None`. The alias silently stops resolving and dry-run breaks again — exactly the SECRT-2368 class of bug we're trying to inoculate against.
2. **Aliased-to-nonexistent-enum is unobservable.** Same `try/except` swallows the case where someone removes an enum entry without updating the map: alias points at a gone enum value, lookup raises, swallowed, returns `None` → unknown-model 400 at runtime instead of a test failure at CI time.

Both go away by moving the dict **below** the class body and storing `LlmModel` members directly:

```python
class LlmModel(str, Enum, metaclass=LlmModelMeta):
    @classmethod
    def _missing_(cls, value: object) -> "LlmModel | None":
        if not isinstance(value, str):
            return None
        alias = _OPENROUTER_ALIASES.get(value)
        if alias is not None:
            return alias
        if "/" in value:
            stripped = value.split("/", 1)[1]
            try:
                return cls(stripped)
            except ValueError:
                return None
        return None

    # … enum members …

_OPENROUTER_ALIASES: Mapping[str, LlmModel] = {
    "anthropic/claude-haiku-4-5":  LlmModel.CLAUDE_4_5_HAIKU,
    "anthropic/claude-opus-4-5":   LlmModel.CLAUDE_4_5_OPUS,
    "anthropic/claude-sonnet-4-5": LlmModel.CLAUDE_4_5_SONNET,
}
```

Snapshot updates now follow the enum automatically; a stale alias becomes an `AttributeError` at module load, not a silent runtime miss.

autogpt_platform/backend/backend/blocks/test/test_llm.py
```diff
@@ -1545,6 +1545,39 @@ async def always_timeout(*args, **kwargs):
...
+        # 4.6/4.7 enum values already match OpenRouter's slug after prefix
+        # strip — no alias-map entry needed.
+        assert llm.LlmModel("anthropic/claude-opus-4-7") is llm.LlmModel.CLAUDE_4_7_OPUS
+        assert llm.LlmModel("anthropic/claude-opus-4-6") is llm.LlmModel.CLAUDE_4_6_OPUS
+
+    def test_unknown_slug_raises(self) -> None:
+        # Bare unknown strings still fail loudly — the alias map is an
+        # additive shortcut, not a silent catch-all.
+        with pytest.raises(ValueError):
+            llm.LlmModel("anthropic/claude-haiku-999")
```
majdyz, line 1584: 🤖 🟡 **Nice to Have**: With the alias map storing strings, there's no regression guard for "added a new Claude enum member with a snapshot suffix but forgot the alias." A small driver test catches the drift class once at CI time instead of letting it fail in prod the next time someone adds `CLAUDE_4_8_*`:

```python
def test_all_claude_snapshots_reachable_via_openrouter_slug(self) -> None:
    """Every Claude enum value with a trailing -YYYYMMDD snapshot must be
    reachable via ``anthropic/<base>`` so the OpenRouter slug round-trips."""
    snapshot_re = re.compile(r"^claude-(.+)-\d{8}$")
    for m in llm.LlmModel:
        match = snapshot_re.match(m.value)
        if match is None:
            continue
        slug = f"anthropic/{match.group(1)}"
        assert llm.LlmModel(slug) is m, (
            f"OpenRouter alias missing for {m.value} — expected slug {slug!r} "
            "to resolve back to this enum member"
        )
```

(Becomes a freebie once `_OPENROUTER_ALIASES` holds `LlmModel` instances per the comment on `llm.py:131` — the driver becomes a pure consistency check.)

autogpt_platform/backend/backend/blocks/llm.py
```diff
@@ -116,11 +116,44 @@ class LlmModelMeta(EnumMeta):
...
 
 
+# OpenRouter exposes Anthropic models under canonical ``anthropic/<model>``
+# slugs that drop the snapshot-date suffix Anthropic's own API uses
+# (``claude-haiku-4-5-20251001`` → ``anthropic/claude-haiku-4-5``).  The
+# generic provider-prefix strip in ``_missing_`` can't reverse the date
+# truncation, so map the OpenRouter slugs to the canonical direct-Anthropic
+# enum values here.  Only models that exist on OpenRouter need entries; the
+# newer 4.6/4.7 enum values already lack a snapshot date so the prefix
+# strip alone is enough for them.
```
majdyz: 🤖 🔵 **Nit**: "Only models that exist on OpenRouter need entries" reads a bit off — all Anthropic enum members exist on OpenRouter. The actual boundary is *"models whose canonical enum value carries a `-YYYYMMDD` snapshot suffix that the OpenRouter slug drops."* Once 4.6/4.7 picked up the no-snapshot convention, the prefix-strip path handled them; aliases are only needed while older snapshot-dated values stay in the enum.

```diff
-# enum values here.  Only models that exist on OpenRouter need entries; the
-# newer 4.6/4.7 enum values already lack a snapshot date so the prefix
-# strip alone is enough for them.
+# enum values here.  Only models whose enum value carries a ``-YYYYMMDD``
+# snapshot suffix need entries; values without a snapshot (4.6/4.7+) are
+# already covered by the prefix-strip path alone.
```

autogpt_platform/backend/backend/blocks/test/test_llm.py
```diff
@@ -1545,6 +1545,39 @@ async def always_timeout(*args, **kwargs):
                 async for _ in block.run(input_data, credentials=llm.TEST_CREDENTIALS):
                     pass
 
+        assert call_count["n"] == 1, (
```
majdyz: 🤖 🔵 **Nit**: The eight `assert (cond), msg` → `assert cond, (msg)` reformats here are unrelated to the OpenRouter alias fix — looks like a pre-commit / ruff run picked up pre-existing style choices and reformatted them en route. They aren't *wrong*, but they double the diff size and force reviewers to scan past 30+ lines of style churn to find the actual change. Per the contributing guide's "Keep out-of-scope changes under 20% of the PR" rule, worth either splitting into a separate `style(backend): ruff reformat assert blocks in test_llm.py` commit or just dropping the unrelated lines so this PR's diff matches its title.

Same observation applies to the lines I'm citing here and the other ~6 occurrences higher up in the file. Non-blocking; flagging so future hotfixes keep the diff scope-tight.

autogpt_platform/backend/backend/blocks/llm.py
majdyz: 🤖 Fixed in [21eef4b38](https://github.com/Significant-Gravitas/AutoGPT/commit/21eef4b38c1e16dc5d5fca879f98ef89a5ec47ff): moved `_OPENROUTER_ALIASES` below the `LlmModel` class body and re-typed it as `Mapping[str, LlmModel]` so each entry holds the enum member directly. Dropped the `try/except ValueError` swallow in `_missing_` — no longer needed since there's no string→enum coercion that can fail. A future snapshot rotation on `CLAUDE_4_5_HAIKU` etc. now follows the alias automatically; a stale entry becomes a load-time `AttributeError` instead of a silent runtime miss.

autogpt_platform/backend/backend/blocks/test/test_llm.py
majdyz: 🤖 Added in [21eef4b38](https://github.com/Significant-Gravitas/AutoGPT/commit/21eef4b38c1e16dc5d5fca879f98ef89a5ec47ff): new `test_all_claude_snapshots_reachable_via_openrouter_slug` walks every `claude-*-YYYYMMDD` enum member and asserts the `anthropic/<base>` slug resolves back to it. Now that `_OPENROUTER_ALIASES` holds `LlmModel` instances directly (per the `llm.py` thread), this becomes a clean consistency check — any new snapshot-dated Claude member added without an alias trips the test at CI time.

autogpt_platform/backend/backend/blocks/llm.py
majdyz: 🤖 Fixed in [21eef4b38](https://github.com/Significant-Gravitas/AutoGPT/commit/21eef4b38c1e16dc5d5fca879f98ef89a5ec47ff): rewrote the alias-map comment to describe the actual boundary — "models whose canonical enum value carries a `-YYYYMMDD` snapshot suffix" — instead of the imprecise "models that exist on OpenRouter". (The comment moved along with the dict — it now sits below the class body.)

autogpt_platform/backend/backend/blocks/test/test_llm.py
majdyz: 🤖 Not actionable inside this PR — verified by running `ruff format` against an isolated copy of one of the reverted hunks: the formatter re-applies the `assert cond, (msg)` shape any time the line exceeds 88 chars (the current ruff default), which is exactly what `assert call_count == 1, f"Expected exactly 1 call for status {status_code}, got {call_count}"` does. Reverting these locally would just have CI's `poetry run ruff format` re-introduce them on the next push, and splitting into a separate `style:` commit doesn't actually reduce the PR diff size (same lines, same commit boundary). Closing as a genuine false positive — the lines are auto-formatter output, not authored choices.

autogpt_platform/backend/backend/blocks/llm.py
```diff
@@ -8,6 +8,7 @@
 from abc import ABC
 from enum import Enum, EnumMeta
 from json import JSONDecodeError
+from collections.abc import Mapping
```
majdyz: 🤖 🔴 **Blocker**: This import slots in the wrong position for `isort --profile black` (the lint script's `poetry run lint` calls `isort --diff --check --profile black` on the backend dir). Verified locally — `isort` flags it:

```
@@ -6,9 +6,9 @@
 from abc import ABC
+from collections.abc import Mapping
 from enum import Enum, EnumMeta
 from json import JSONDecodeError
-from collections.abc import Mapping
 from typing import Any, Iterable, ...
```

Move it up two lines so the stdlib `from` block stays alphabetical (`abc` < `collections.abc` < `enum` < `json` < `typing`). Otherwise the `lint` CI step trips on the backend isort check.

majdyz: 🤖 Fixed in [afc1862e2](https://github.com/Significant-Gravitas/AutoGPT/commit/afc1862e25ac652bee1edb3a7e78aa1a407e7ea2): moved `from collections.abc import Mapping` up between `abc` and `enum` so the stdlib `from` block stays alphabetical. `poetry run isort --diff --check --profile black backend/blocks/llm.py` and `black --diff --check` both clean now.

autogpt_platform/backend/backend/blocks/test/test_llm.py
majdyz: 🤖 Self-correction on this thread: my earlier resolution-reply claimed `ruff format` enforced the `assert cond, (msg)` shape — that was right for ruff, **wrong for this codebase**. The backend's canonical formatter is `black` (called by `poetry run format`), and CI's `lint` step hit `black --check` and rejected those exact reformats. Reverted in [37258a7a0](https://github.com/Significant-Gravitas/AutoGPT/commit/37258a7a0b3bab8525025fa44bbdebb49095d17d): re-ran `black backend/blocks/test/test_llm.py` which flipped the eight hunks back to the original `assert (cond), msg` shape. PR diff shrinks correspondingly — the original scope-creep concern is now naturally resolved by formatter alignment. Worth keeping `poetry run format` (not `ruff format`) as the canonical pre-push step for backend files.

majdyz:
🤖 @coderabbitai — good catch. Fixed in [5e72d7b34](https://github.com/Significant-Gravitas/AutoGPT/commit/5e72d7b34bf48a81194b919b637d538ff57f21a3): moved the `claude-` into the capturing group so the regex `^(claude-.+)-\d{8}$` produces a slug-ready capture. Verified locally against all three snapshot-dated members — `anthropic/claude-haiku-4-5`, `anthropic/claude-opus-4-5`, `anthropic/claude-sonnet-4-5` all round-trip back to their enum members. Your suggestion of `f"anthropic/claude-{match.group(1)}"` would have worked equally well; I picked the in-regex move because it keeps the slug construction site dumb ("prefix + captured base") which is one less thing to keep in sync if a future model line uses a different prefix.

PR #13179, opened 2026-05-21
fix(backend/copilot): make SDK tool-result reads navigable + redirect wrong-tool access

majdyz:
## Why

Production session [`3304a1df-30ab-446d-8841-bce1a41ab8fa`](https://dev-builder.agpt.co/copilot?sessionId=3304a1df-30ab-446d-8841-bce1a41ab8fa) hit a post-execution analysis dead-end. After `run_agent` came back with a failed Orchestrator node, the model:

1. Called `read_tool_result` and got the file contents back successfully.
2. Tried to `bash_exec` `cat /root/.claude/projects/.../tool-results/toolu_*.json | python3 -c '...'` to parse the nested `node_executions[*].error` field — got `Permission denied` because the bash sandbox can't reach the SDK's host-side tool-results dir.
3. Tried `read_workspace_file(path="tool-outputs/toolu_*.json")` — got `Path not allowed` because that's not workspace storage.
4. Gave up and shipped the user a guess about what failed.

Underneath, three problems compounded:

- The file on disk is the MCP envelope `[{"type":"text","text":"<minified-json>"}]`. `read_tool_result`'s line-based `offset`/`limit` sliced the envelope wrapper (≈5 lines), not the payload inside (one giant minified line) — so even when the model used the right tool, it couldn't *slice* the right tool's output.
- `bash_exec` had no awareness of host-side SDK paths, so the model could only learn "wrong tool" via repeated `Permission denied` retries.
- `read_workspace_file` returned a generic `Path not allowed` that didn't point at the actual right tool.

## What

Fix all three from different angles in one PR:

- **`read_tool_result`** auto-unwraps `[{"type":"text","text":...}]` and pretty-prints inner JSON so line-based `offset`/`limit` slice the *payload*. Also adds `char_offset`/`char_limit` for byte-precise reads when even pretty-printed lines are too long (base64 blobs, token-dump strings).
- **`bash_exec`** pre-scans the command for SDK tool-result paths and short-circuits with a redirect error pointing at `read_tool_result` *and* the existing `@@agptfile:<abs-path>[<s>-<e>]` reference protocol — so the model can pipe a slice through bash without ever touching the host path.
- **`read_workspace_file`** returns the same redirect hint when the path looks like an SDK tool-result shorthand (e.g. `tool-outputs/<id>.json`).

## How

- `copilot/context.py` — new `looks_like_sdk_tool_result_path()` substring-detector (matches `/.claude/projects/`, `tool-results/`, `tool-outputs/`) and `sdk_tool_result_redirect_hint()` which builds the consistent redirect message used by both wrong-tool sites.
- `copilot/sdk/tool_adapter.py::_read_file_handler` — read the whole file, call new `_navigable_tool_result_text()` to unwrap+pretty-print, then slice either line- or char-mode. `_navigable_tool_result_text` is defensive: non-envelope JSON, non-JSON inner text, and multi-block envelopes all fall through to the raw content untouched.
- `copilot/tools/bash_exec.py` — pre-flight check before `get_current_sandbox()`; emits `ErrorResponse(error="sdk_tool_result_path_in_bash_command")` so the failure is observable in metrics.
- `copilot/tools/workspace_files.py` — second branch in the existing `_resolve_file → ErrorResponse` fallback.

Tool description for `read_tool_result` updated to advertise both the unwrap behavior and the `@@agptfile:` alternative.

## Test plan

- [x] `TestNavigableToolResultText` (3 cases) — envelope + JSON gets pretty-printed; envelope + non-JSON gets unwrapped; non-envelope inputs returned unchanged.
- [x] `TestBashExecSdkToolResultRedirect` (3 cases) — absolute SDK path redirects; relative `tool-outputs/` shorthand redirects; normal commands still execute against the sandbox.
- [x] `pytest backend/copilot/sdk/tool_adapter_test.py backend/copilot/tools/bash_exec_test.py backend/copilot/tools/workspace_files_test.py` — all green
- [ ] `/pr-test` against this branch to confirm the production session shape (run_agent → view failed → slice payload) no longer churns

### DIFF
--- a/autogpt_platform/backend/backend/copilot/context.py
+++ b/autogpt_platform/backend/backend/copilot/context.py
@@ -182,13 +182,49 @@ def looks_like_sdk_tool_result_path(text: str) -> bool:
     return any(hint in text for hint in _SDK_TOOL_RESULT_HINTS)
 
 
+# Shell separators used by ``_extract_offending_segment`` to bound the
+# path-like token that triggered the redirect. Quotes are included so a
+# quoted SDK path inside a command (e.g. ``cat "/root/.claude/..."``)
+# extracts the path without the surrounding quotes.
+_SHELL_TOKEN_DELIMS = frozenset(" |;&<>'\"\n\t`()")
+
+
+def _extract_offending_segment(text: str) -> str:
+    """Return the path-shaped token in *text* that tripped a redirect hint.
+
+    For bash commands the ``looks_like_sdk_tool_result_path`` hint is
+    almost always a substring of one shell token (e.g. the file argument
+    of a ``cat``), not the command itself. Pull that token out so the
+    redirect message surfaces the offending *path*, not the executable.
+    Falls back to the first whitespace-separated token of *text* so
+    paths passed via ``read_workspace_file`` (where the input is just
+    the path) still echo cleanly.
+    """
+    for hint in _SDK_TOOL_RESULT_HINTS:
+        idx = text.find(hint)
+        if idx < 0:
+            continue
+        end = idx + len(hint)
+        while end < len(text) and text[end] not in _SHELL_TOKEN_DELIMS:
+            end += 1
+        start = idx
+        while start > 0 and text[start - 1] not in _SHELL_TOKEN_DELIMS:
+            start -= 1
+        return text[start:end]
+    stripped = text.strip()
+    return stripped.split()[0] if stripped else ""
+
+
 def sdk_tool_result_redirect_hint(path_or_command: str) -> str:
     """User-facing redirect message for wrong-tool access to SDK tool-results.
 
     Names the two right ways to get at the bytes so the model doesn't
-    burn another turn retrying with yet another wrong tool.
+    burn another turn retrying with yet another wrong tool. The
+    "Offending fragment" surfaces the actual path-shaped token from
+    *path_or_command* (so a ``cat /…/tool-results/foo.json | jq``
+    invocation echoes the path, not the ``cat`` executable).
     """
-    sample = path_or_command.strip().split()[0] if path_or_command else ""
+    sample = _extract_offending_segment(path_or_command) if path_or_command else ""
     return (
         "This path lives in the Claude Agent SDK's host-side "
         "tool-results directory, which is not mounted into the bash "
@@ -199,7 +235,7 @@ def sdk_tool_result_redirect_hint(path_or_command: str) -> str:
         "tool's argument to inline a slice of the file's content "
         '(e.g. `bash_exec(command="echo @@agptfile:/root/.claude/'
         'projects/.../result.json[1-200] | jq .field")`).\n'
-        f"Offending fragment: {os.path.basename(sample) or sample!r}"
+        f"Offending fragment: {sample!r}"
     )
 
 
--- a/autogpt_platform/backend/backend/copilot/context.py
+++ b/autogpt_platform/backend/backend/copilot/context.py
@@ -237,9 +237,7 @@ def sdk_tool_result_redirect_hint(path_or_command: str) -> str:
     the *actual* matched SDK path (not just ``bash``/``cat``) so the
     model knows which fragment of its command tripped the redirect.
     """
-    fragment = (
-        _extract_offending_segment(path_or_command) if path_or_command else ""
-    )
+    fragment = _extract_offending_segment(path_or_command) if path_or_command else ""
     return (
         "This path lives in the Claude Agent SDK's host-side "
         "tool-results directory, which is not mounted into the bash "
--- a/autogpt_platform/backend/backend/copilot/context.py
+++ b/autogpt_platform/backend/backend/copilot/context.py
@@ -157,6 +157,52 @@ def is_sdk_tool_path(path: str) -> bool:
     )
 
 
+# Substring tokens that strongly suggest a string contains a host-side
+# SDK tool-result path (e.g. inside a bash command). Used by wrong-tool
+# detectors to redirect the model to ``read_tool_result`` / ``@@agptfile:``
+# rather than letting it bounce off ``Permission denied`` repeatedly.
+_SDK_TOOL_RESULT_HINTS: tuple[str, ...] = (
+    "/.claude/projects/",
+    "tool-results/",
+    "tool-outputs/",
+)
+
+
+def looks_like_sdk_tool_result_path(text: str) -> bool:
+    """Heuristic: does *text* look like a host-side SDK tool-result path?
+
+    Cheap substring scan suitable for pre-checking ``bash_exec`` commands
+    and ``read_workspace_file`` paths before they hit a sandbox boundary
+    they can't cross. Conservative — accepts both absolute SDK paths
+    (``/root/.claude/projects/.../tool-results/...``) and the short
+    relative form the model often invents (``tool-outputs/<id>.json``).
+    """
+    if not text:
+        return False
+    return any(hint in text for hint in _SDK_TOOL_RESULT_HINTS)
+
+
+def sdk_tool_result_redirect_hint(path_or_command: str) -> str:
+    """User-facing redirect message for wrong-tool access to SDK tool-results.
+
+    Names the two right ways to get at the bytes so the model doesn't
+    burn another turn retrying with yet another wrong tool.
+    """
+    sample = path_or_command.strip().split()[0] if path_or_command else ""
+    return (
+        "This path lives in the Claude Agent SDK's host-side "
+        "tool-results directory, which is not mounted into the bash "
+        "sandbox / workspace storage. Use one of:\n"
+        "  - `read_tool_result(file_path=...)` for direct reads "
+        "(supports offset/limit; auto-unwraps the MCP envelope).\n"
+        "  - `@@agptfile:<absolute-path>[<start>-<end>]` inside another "
+        "tool's argument to inline a slice of the file's content "
+        '(e.g. `bash_exec(command="echo @@agptfile:/root/.claude/'
+        'projects/.../result.json[1-200] | jq .field")`).\n'
+        f"Offending fragment: {os.path.basename(sample) or sample!r}"
+    )
+
+
 def resolve_sandbox_path(path: str) -> str:
     """Normalise *path* to an absolute sandbox path under an allowed directory.
 
--- a/autogpt_platform/backend/backend/copilot/context.py
+++ b/autogpt_platform/backend/backend/copilot/context.py
@@ -157,29 +157,38 @@ def is_sdk_tool_path(path: str) -> bool:
     )
 
 
-# Substring tokens that strongly suggest a string contains a host-side
-# SDK tool-result path (e.g. inside a bash command). Used by wrong-tool
-# detectors to redirect the model to ``read_tool_result`` / ``@@agptfile:``
-# rather than letting it bounce off ``Permission denied`` repeatedly.
-_SDK_TOOL_RESULT_HINTS: tuple[str, ...] = (
-    "/.claude/projects/",
-    "tool-results/",
-    "tool-outputs/",
+# Anchored matches for host-side SDK tool-result paths. Used by
+# wrong-tool detectors (bash_exec, read_workspace_file) to redirect the
+# model to ``read_tool_result`` / ``@@agptfile:`` rather than letting it
+# bounce off ``Permission denied`` repeatedly.
+#
+# Two shapes are matched:
+#   (1) absolute SDK path:  /<...>/.claude/projects/<cwd>/<uuid>/tool-results/<file>.json
+#   (2) model shorthand:    tool-(results|outputs)/<toolu|mcp>_<id>.json
+#
+# The shorthand branch requires the SDK's actual ID prefix so a user
+# repo that happens to use a ``tool-outputs/`` directory name does NOT
+# trigger a false redirect on legitimate paths like
+# ``my-pipeline/tool-outputs/data.json``.
+_SDK_TOOL_RESULT_RE = re.compile(
+    r"(?:^|[/\s'\"])/?\.claude/projects/[^/\s]+/[^/\s]+/tool-(?:results|outputs)/"
+    r"[\w-]+\.json"
+    r"|(?:^|[/\s'\"])tool-(?:results|outputs)/(?:toolu|mcp)_[\w-]+\.json",
+    re.IGNORECASE,
 )
 
 
 def looks_like_sdk_tool_result_path(text: str) -> bool:
-    """Heuristic: does *text* look like a host-side SDK tool-result path?
+    """Return True if *text* references a host-side SDK tool-result file.
 
-    Cheap substring scan suitable for pre-checking ``bash_exec`` commands
-    and ``read_workspace_file`` paths before they hit a sandbox boundary
-    they can't cross. Conservative — accepts both absolute SDK paths
-    (``/root/.claude/projects/.../tool-results/...``) and the short
-    relative form the model often invents (``tool-outputs/<id>.json``).
+    Anchored regex match — avoids false positives on user paths that
+    happen to contain a ``tool-outputs/`` directory by requiring either
+    the full ``.claude/projects/<…>/tool-(results|outputs)/<file>.json``
+    shape or the SDK's ``(toolu|mcp)_<id>.json`` filename pattern.
     """
     if not text:
         return False
-    return any(hint in text for hint in _SDK_TOOL_RESULT_HINTS)
+    return _SDK_TOOL_RESULT_RE.search(text) is not None
 
 
 # Shell separators used by ``_extract_offending_segment`` to bound the
@@ -192,22 +201,27 @@ _SHELL_TOKEN_DELIMS = frozenset(" |;&<>'\"\n\t`()")
 def _extract_offending_segment(text: str) -> str:
     """Return the path-shaped token in *text* that tripped a redirect hint.
 
-    For bash commands the ``looks_like_sdk_tool_result_path`` hint is
-    almost always a substring of one shell token (e.g. the file argument
-    of a ``cat``), not the command itself. Pull that token out so the
-    redirect message surfaces the offending *path*, not the executable.
-    Falls back to the first whitespace-separated token of *text* so
-    paths passed via ``read_workspace_file`` (where the input is just
-    the path) still echo cleanly.
+    Uses the regex match position as the anchor (so false positives like
+    ``my-pipeline/tool-outputs/data.json`` never reach this path) and
+    walks outward to shell-token delimiters to capture the full
+    surrounding path (handles ``--file=/path/...`` and quoted paths).
+    Falls back to the first whitespace-separated token so callers that
+    pass an isolated path (e.g. ``read_workspace_file``) still echo
+    cleanly.
     """
-    for hint in _SDK_TOOL_RESULT_HINTS:
-        idx = text.find(hint)
-        if idx < 0:
-            continue
-        end = idx + len(hint)
+    match = _SDK_TOOL_RESULT_RE.search(text)
+    if match:
+        # Walk outward from the matched span to shell-token boundaries
+        # so we capture the full path token, not just the regex hit
+        # (which excludes the leading slash on the absolute branch).
+        end = match.end()
         while end < len(text) and text[end] not in _SHELL_TOKEN_DELIMS:
             end += 1
-        start = idx
+        start = match.start()
+        # Skip any leading delimiter the regex consumed (e.g. the space
+        # before "/root/.claude/...").
+        while start < end and text[start] in _SHELL_TOKEN_DELIMS:
+            start += 1
         while start > 0 and text[start - 1] not in _SHELL_TOKEN_DELIMS:
             start -= 1
         return text[start:end]
@@ -219,12 +233,13 @@ def sdk_tool_result_redirect_hint(path_or_command: str) -> str:
     """User-facing redirect message for wrong-tool access to SDK tool-results.
 
     Names the two right ways to get at the bytes so the model doesn't
-    burn another turn retrying with yet another wrong tool. The
-    "Offending fragment" surfaces the actual path-shaped token from
-    *path_or_command* (so a ``cat /…/tool-results/foo.json | jq``
-    invocation echoes the path, not the ``cat`` executable).
+    burn another turn retrying with yet another wrong tool, and quotes
+    the *actual* matched SDK path (not just ``bash``/``cat``) so the
+    model knows which fragment of its command tripped the redirect.
     """
-    sample = _extract_offending_segment(path_or_command) if path_or_command else ""
+    fragment = (
+        _extract_offending_segment(path_or_command) if path_or_command else ""
+    )
     return (
         "This path lives in the Claude Agent SDK's host-side "
         "tool-results directory, which is not mounted into the bash "
@@ -235,7 +250,7 @@ def sdk_tool_result_redirect_hint(path_or_command: str) -> str:
         "tool's argument to inline a slice of the file's content "
         '(e.g. `bash_exec(command="echo @@agptfile:/root/.claude/'
         'projects/.../result.json[1-200] | jq .field")`).\n'
-        f"Offending fragment: {sample!r}"
+        f"Offending fragment: {fragment!r}"
     )
 
 
--- a/autogpt_platform/backend/backend/copilot/sdk/tool_adapter.py
+++ b/autogpt_platform/backend/backend/copilot/sdk/tool_adapter.py
@@ -489,8 +489,25 @@ async def _read_file_handler(args: dict[str, Any]) -> dict[str, Any]:
         #
         # When E2B is active, also copy the file into the sandbox so
         # bash_exec can process it (the model often uses Read then bash).
+        # CAVEAT: only bridge when the on-disk bytes and what the model
+        # just read are the same — i.e. when ``_navigable_tool_result_text``
+        # did *not* transform the content. If we pretty-printed an MCP
+        # envelope, the model sees a pretty-printed slice while the
+        # bridged sandbox file would hold the raw envelope; bash commands
+        # operating on the bridged copy would then see different content
+        # than the model just read, leading to silent format-mismatch
+        # bugs. Same constraint applies to char-mode slices. The new
+        # bash_exec SDK-path redirect (added in this PR) covers the
+        # alternative workflow when the bridge is skipped — the model
+        # can use ``read_tool_result`` again with offsets, or pipe a
+        # slice via ``@@agptfile:<path>[<start>-<end>]``.
         sandbox = _current_sandbox.get(None)
-        if sandbox is not None:
+        if (
+            sandbox is not None
+            and navigable == raw
+            and char_offset is None
+            and char_limit is None
+        ):
             annotation = await bridge_and_annotate(sandbox, resolved, offset, limit)
             if annotation:
                 text += annotation
--- a/autogpt_platform/backend/backend/copilot/sdk/tool_adapter.py
+++ b/autogpt_platform/backend/backend/copilot/sdk/tool_adapter.py
@@ -421,9 +421,15 @@ async def _read_file_handler(args: dict[str, Any]) -> dict[str, Any]:
         )
 
     file_path = args.get("file_path", "")
+    char_offset_arg = args.get("char_offset")
+    char_limit_arg = args.get("char_limit")
     try:
         offset = max(0, int(args.get("offset", 0)))
         limit = max(1, int(args.get("limit", 2000)))
+        char_offset = (
+            max(0, int(char_offset_arg)) if char_offset_arg is not None else None
+        )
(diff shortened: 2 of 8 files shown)


autogpt_platform/backend/backend/copilot/tools/workspace_files.py
```diff
@@ -602,6 +604,15 @@ async def _execute(
                     return _read_local_tool_result(
                         path, char_offset, char_length, session_id, sdk_cwd=sdk_cwd
                     )
+                # Path looks like SDK tool-results (e.g. shorthand
+                # ``tool-outputs/<id>.json``) but doesn't resolve as a
+                # full SDK path — redirect to read_tool_result / @@agptfile
+                # rather than returning a generic "Path not allowed".
+                if path and looks_like_sdk_tool_result_path(path):
```
majdyz, line 633: 🤖 🟠 **Should Fix**: This new redirect branch in `workspace_files.py` has no test coverage. The PR added `TestBashExecSdkToolResultRedirect` for the analogous `bash_exec` redirect (3 cases) but the workspace-file path is untested. A simple test that passes `path="tool-outputs/x.json"` (no real file on disk, so `_resolve_file` returns ErrorResponse, and `is_allowed_local_path` is false) and asserts the redirect message contains `read_tool_result` would mirror the bash test.

majdyz: Addressed in eb3a93375b — added `test_read_workspace_file_redirects_sdk_tool_result_shorthand` mirroring the bash_exec redirect tests.

autogpt_platform/backend/backend/copilot/tools/bash_exec.py
```diff
@@ -125,6 +130,16 @@ async def _execute(
...
 
+        # Pre-flight redirect: bash sandbox can't reach host-side SDK
+        # tool-result paths. Without this the model burns turns retrying
+        # `cat /root/.claude/projects/...` after `Permission denied`.
+        if looks_like_sdk_tool_result_path(command):
+            return ErrorResponse(
+                message=sdk_tool_result_redirect_hint(command),
+                error="sdk_tool_result_path_in_bash_command",
+                session_id=session_id,
+            )
```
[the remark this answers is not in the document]
majdyz: Declined — the broad substring match is intentional and documented in `looks_like_sdk_tool_result_path`'s docstring ("Conservative — accepts both absolute SDK paths and the short relative form the model often invents").

The three hint substrings (`/.claude/projects/`, `tool-results/`, `tool-outputs/`) are all SDK-reserved names in this codebase:
- `/.claude/projects/` is anchored — only the host-side SDK creates that path.
- `tool-results/` and `tool-outputs/` are the *only* directory names the SDK uses for tool-result storage (see `context.py::is_sdk_tool_path` and the workspace-files redirect — both rely on the same convention).

The bash sandbox CWD is `/home/user/` (E2B) — the model never has a legitimate reason to create or list a directory literally called `tool-outputs/` there, and if it does, the error message tells it precisely which tool to use instead. Tightening to an anchored match would re-introduce the original failure mode this PR exists to fix: model writes `cat tool-outputs/foo.json` (relative shorthand it invents), gets `Permission denied` from the sandbox, retries with another wrong tool. The whole point of the heuristic being a substring is to catch those invented-relative variants.

If we ever see a real user collision, the right fix is to rename the SDK directory, not narrow the heuristic.

autogpt_platform/backend/backend/copilot/context.py
```diff
@@ -157,6 +157,52 @@ def is_sdk_tool_path(path: str) -> bool:
...
 
 
+# Substring tokens that strongly suggest a string contains a host-side
+# SDK tool-result path (e.g. inside a bash command). Used by wrong-tool
+# detectors to redirect the model to ``read_tool_result`` / ``@@agptfile:``
+# rather than letting it bounce off ``Permission denied`` repeatedly.
+_SDK_TOOL_RESULT_HINTS: tuple[str, ...] = (
+    "/.claude/projects/",
+    "tool-results/",
+    "tool-outputs/",
```
majdyz: 🤖 🟠 **Should Fix**: The substring match here is loose enough to false-positive on legitimate user paths. `/.claude/projects/` is fine — that's a specific marker — but `tool-results/` and `tool-outputs/` are common directory names. A user whose workspace happens to contain `my-pipeline/tool-outputs/data.json` would now get the redirect-to-`read_tool_result` error on every `bash_exec` that references that path, with no way to opt out.

Two anchors would tighten this without giving up the model-shorthand detection (`tool-outputs/<id>.json`) the PR is actually trying to catch:

1. Require the segment to look like the SDK shape — `<dir>/<id>.json` — so a directory traversal like `cd tool-outputs && ls` doesn't trip it.
2. Require either path-boundary anchoring or a recognisable SDK-result filename pattern (`toolu_*` / `mcp_*` are the SDK's actual id prefixes).

Concrete:

```python
_SDK_TOOL_RESULT_RE = re.compile(
    r"(^|/|\s)/?\.claude/projects/[^/\s]+/[^/\s]+/tool-results/[\w-]+\.json"
    r"|(^|/|\s)tool-(results|outputs)/(toolu|mcp)_[\w-]+\.json",
    re.IGNORECASE,
)

def looks_like_sdk_tool_result_path(text: str) -> bool:
    if not text:
        return False
    return _SDK_TOOL_RESULT_RE.search(text) is not None
```

That keeps the model-shorthand happy path (`tool-outputs/toolu_x.json`) and the absolute SDK form, but stops shadowing any user repo that happens to use a `tool-outputs/` or `tool-results/` folder name. A user can still hit the redirect deliberately by passing the exact SDK shape; they just can't trip it by accident on an unrelated path.

autogpt_platform/backend/backend/copilot/context.py
```diff
@@ -157,6 +157,52 @@ def is_sdk_tool_path(path: str) -> bool:
...
+    return any(hint in text for hint in _SDK_TOOL_RESULT_HINTS)
+
+
+def sdk_tool_result_redirect_hint(path_or_command: str) -> str:
+    """User-facing redirect message for wrong-tool access to SDK tool-results.
+
+    Names the two right ways to get at the bytes so the model doesn't
+    burn another turn retrying with yet another wrong tool.
+    """
+    sample = path_or_command.strip().split()[0] if path_or_command else ""
```
majdyz: 🤖 🟡 **Nice to Have**: For `bash_exec` commands, the "Offending fragment" message reads `bash` or `cat` because `split()[0]` is the executable name, not the SDK path. The model gets pointed at the right tool (good), but the error itself is unhelpful — it tells the model its `cat` command was the problem rather than the specific path inside it.

A small extractor that finds the matching hint and walks outward to the surrounding token gives a more actionable message:

```python
def _extract_offending_segment(text: str) -> str:
    for hint in _SDK_TOOL_RESULT_HINTS:
        idx = text.find(hint)
        if idx < 0:
            continue
        # Walk outward to shell delimiters
        delims = " |;&<>'\"\n\t`"
        end = idx + len(hint)
        while end < len(text) and text[end] not in delims:
            end += 1
        start = idx
        while start > 0 and text[start - 1] not in delims:
            start -= 1
        return text[start:end]
    return ""

# in sdk_tool_result_redirect_hint:
sample = _extract_offending_segment(path_or_command) or path_or_command.strip().split()[0]
```

That way `bash_exec(command="cat /root/.claude/projects/.../tool-results/toolu_x.json | jq .")` reports `Offending fragment: '/root/.claude/projects/.../tool-results/toolu_x.json'` instead of `Offending fragment: 'cat'`. Same redirect message body, useful sample.

autogpt_platform/backend/backend/copilot/sdk/tool_adapter.py
```diff
@@ -460,14 +466,29 @@ def _mcp_ok(text: str) -> dict[str, Any]:
...
     try:
+        # Read the whole file: tool-result envelopes are usually small JSON
+        # but the payload inside is often one massive minified line, so
+        # line-based offset/limit applied to the raw bytes is useless.
+        # _navigable_tool_result_text() unwraps the envelope and pretty-
+        # prints inner JSON so line offsets actually slice the payload.
         with open(resolved, encoding="utf-8", errors="replace") as f:
-            selected = list(itertools.islice(f, offset, offset + limit))
+            raw = f.read()
+        navigable = _navigable_tool_result_text(raw)
```
majdyz, line 476: 🤖 🟡 **Nice to Have**: This switched from "stream the first `offset+limit` lines via `itertools.islice`" to "read the whole file → JSON-parse twice → pretty-print → slice". Correctness-wise that's the whole point (the old slice was unusable because the envelope was 5 lines while the payload was one giant minified line). But for large tool results the new path materialises the entire pretty-printed payload before slicing, which for a multi-MB minified inner JSON can be 2-4× the bytes after pretty-printing.

Not a blocker — tool-result files are typically small, this runs once per `read_tool_result` call (not per turn), and the LLM tool overhead dwarfs the JSON-parse anyway. Just worth noting that if anyone later sees a slow `read_tool_result` on a huge `node_executions` dump, the working set lives here. A streaming alternative would parse the outer envelope by hand (or with `ijson`) to find the inner-text byte range without ever materialising the pretty-printed form, then slice on character offsets directly — but the complexity probably isn't worth it until someone actually hits the slowness.

autogpt_platform/backend/backend/copilot/tools/workspace_files.py
majdyz: Addressed in f6cac33daa (replaces the initial fix from eb3a93375b that hit Prisma). New `TestReadWorkspaceFileSdkToolResultRedirect` covers three cases: relative `tool-outputs/<id>.json` shorthand redirects, relative `tool-results/<id>.json` shorthand redirects, and an unrelated missing path still surfaces the original resolve error (regression guard against over-eager redirects). Mocks `get_workspace_manager` and `_resolve_file` so the test stays scoped to the redirect branch and runs in 0.3s without a DB.

autogpt_platform/backend/backend/copilot/context.py
majdyz: 🤖 Fixed in [9cefe3104](https://github.com/Significant-Gravitas/AutoGPT/commit/9cefe3104e16bb706ad5839eda05e77bd5e9925c): added `_extract_offending_segment` which finds the matching `_SDK_TOOL_RESULT_HINTS` substring and walks outward to the nearest shell delimiter (space, pipe, semicolon, redirect, quotes, backtick). `sdk_tool_result_redirect_hint` uses that result, falling back to `split()[0]` only when no hint substring is present. Quoted paths unwrap cleanly. New `TestSdkToolResultRedirectHint` (5 cases) pins the behaviour — covers absolute SDK path, quoted path, relative shorthand, bare `read_workspace_file`-shape input, and empty input.

autogpt_platform/backend/backend/copilot/context.py
majdyz: 🤖 Deferring — same concern Sentry raised in https://github.com/Significant-Gravitas/AutoGPT/pull/13179#discussion_r3279960896 which you declined as intentional: "the broad substring match is intentional and documented in `looks_like_sdk_tool_result_path`'s docstring." The trade-off (broad redirect on legitimate `tool-outputs/` paths in user workspaces vs. catching every model-invented shorthand) is the same in both threads. Closing this as covered by that decision rather than re-litigating.

autogpt_platform/backend/backend/copilot/sdk/tool_adapter.py
majdyz: 🤖 Noted, no code change. The trade-off is acceptable until a real slowness report — tool-result files are typically <1MB, the call frequency is once per `read_tool_result` invocation, and LLM-tool round-trip overhead dwarfs the pretty-printing cost in the normal case. Leaving the streaming-via-`ijson` alternative as a follow-up if anyone hits the slowness on a real payload.

autogpt_platform/backend/backend/copilot/context.py
majdyz: Addressed in ac9b165d9a. Replaced the substring scan with an anchored regex `_SDK_TOOL_RESULT_RE` that requires either the full `/.claude/projects/<…>/tool-(results|outputs)/<file>.json` shape or the SDK's actual `(toolu|mcp)_<id>.json` filename pattern on the shorthand branch. Added regression test `test_no_redirect_on_user_path_containing_tool_outputs` confirming `my-pipeline/tool-outputs/data.json` still hits the sandbox normally.

autogpt_platform/backend/backend/copilot/context.py
majdyz: Addressed in 9cefe3104e + refined in ac9b165d9a. New `_extract_offending_segment` uses the regex match span as the anchor (so false positives like `my-pipeline/tool-outputs/data.json` never reach this path) and walks outward to shell-token delimiters to capture the full surrounding path token — handles quoted paths and `--file=/path/...` flags too. The existing absolute-path test now asserts `tool-results/toolu_x.json` is in the message and `'cat'` is not, locking the behavior in.

autogpt_platform/backend/backend/copilot/sdk/tool_adapter.py
majdyz: Acknowledged — agreed it's not a blocker. Tool-result files are typically tens of KB in the traces we have, the unwrap is called once per `read_tool_result` (not per turn), and the new `char_offset`/`char_limit` path still walks the same buffered string. Worth revisiting only if a real multi-MB `node_executions` dump shows up and the unwrap latency becomes load-bearing; until then the 2-4× pretty-print expansion is the price of making `offset`/`limit` actually navigable. Leaving the streaming `ijson` approach out of this PR per scope.

autogpt_platform/backend/backend/copilot/sdk/tool_adapter.py
```diff
@@ -460,14 +466,29 @@ def _mcp_ok(text: str) -> dict[str, Any]:
...
+            text = "".join(lines[offset : offset + limit])
         # Cleanup happens in _cleanup_sdk_tool_results after session ends;
         # don't delete here — the SDK may read in multiple chunks.
         #
         # When E2B is active, also copy the file into the sandbox so
         # bash_exec can process it (the model often uses Read then bash).
-        text = "".join(selected)
         sandbox = _current_sandbox.get(None)
         if sandbox is not None:
             annotation = await bridge_and_annotate(sandbox, resolved, offset, limit)
```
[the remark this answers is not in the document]
majdyz: 🤖 Real bug, fixed in [b10e0ecb4](https://github.com/Significant-Gravitas/AutoGPT/commit/b10e0ecb4cff9fbc5f1e0ee59e0ca1227afb7913). Skipped the bridge in both shapes:

1. **Pretty-printed envelope** (`navigable != raw`) — the model reads a pretty-printed payload, but `bridge_to_sandbox` would copy the raw envelope bytes. Bash on the bridged file would see different content than `read_tool_result` just returned. Skip.
2. **Char-mode slice** — the model reads a substring of `navigable`, but `bridge_to_sandbox` only fires on full-file reads (line offset=0/limit≥default), so it would copy the whole file. The model has a slice; bash gets the full file. Skip.

The new `bash_exec` SDK-path redirect (also in this PR) covers the alternative workflow: model can re-read with offsets, or pipe a slice via `@@agptfile:<path>[<start>-<end>]`. Two new tests in `TestReadFileHandlerBridge` pin the skip — `test_bridge_skipped_when_envelope_pretty_printed` (envelope case) and `test_bridge_skipped_when_char_offset_used` (char-mode case on a non-envelope file).

majdyz:
## E2E Test Report

# E2E Test Report: PR #13179 — fix(backend/copilot): make SDK tool-result reads navigable + redirect wrong-tool access

- Date: 2026-05-21
- Branch: `fix/copilot-tool-result-path-redirect`
- Worktree: `/Users/majdyz/Code/AutoGPT17`
- Tester: native stack (`poetry run app` + `pnpm dev` on :3002, deps in docker)
- Auth: Claude OAuth (subscription mode, CHAT_USE_CLAUDE_CODE_SUBSCRIPTION=true)
- Model: extended_thinking (Claude Agent SDK path) — the surface the PR actually targets

## Environment notes
- Frontend bound to :3002 because :3000 was held by an unkillable stale Next.js process from a prior session.
- DB had migration drift (`platform.ChatSessionStatus` missing) — fixed with `poetry run prisma migrate reset --force`.
- Frontend `pnpm dev` script failed in `orval` due to OOM; ran `npx next dev --turbo` directly instead. Backend was unaffected.

## Test Results

### Scenario A — Tool-result navigability (headline fix)
**Steps:**
1. Created a 3-node test graph (AgentInput → StoreValue → AgentOutput), library_agent_id `9eea21cb-6a4e-4082-9f2a-bbe66529ba28`.
2. Asked the copilot (via `/api/chat/sessions/:id/stream`, SSE, extended_thinking mode): "Please run my library agent (library_agent_id=…) with dry_run=True. After it produces output, use read_tool_result to inspect the tool-result file and tell me what node_executions are inside. Use read_tool_result only — do NOT use bash_exec, cat, or read_file."

**Tool sequence observed (from `scenario-A-stream.log`):**
1. `run_agent` → returned full `node_executions` inline (output was small, no truncation, no separate tool-result file written).
2. Model reasoned: "the result came back inline without a truncated file reference. … The user specifically asked me to use read_tool_result only. Let me check if there's a file path mentioned or if the result was truncated."
3. Model concluded no tool-result file existed and summarised the two node_executions directly. Final assistant message includes both nodes with `status`, `input_data`, `output_data`, and explains the simulator JSON error as a dry-run-simulator limitation rather than a real block failure.

**Headline-fail signals checked:**
- `bash_exec` calls in the stream: **0**
- `Permission denied` strings in the stream: **0**
- `Path not allowed` strings in the stream: **0**
- `cat /…/.claude/projects/…` attempts: **0**

**Navigability code-path direct verification (because inline result bypassed disk):**
Created a synthetic MCP-envelope tool-result file (`[{"type":"text","text":"<minified-json>"}]`) under `~/.claude/projects/<project>/<uuid>/tool-results/toolu_*.json` and invoked the production `_read_file_handler` with the prod context vars set:
- `offset=0, limit=50` → returned a pretty-printed JSON view of the *inner* payload (envelope unwrapped, `node_executions[*]` on separate indented lines). `isError=False`.
- `char_offset=100, char_limit=200` → returned 200 chars of the navigable text starting at byte 100 (mid-`COMPLETED",\n    "node_executions": …`). New params from the PR exercised end-to-end.

**Result:** PASS — model never reached for `bash_exec` to read the SDK side, and the underlying `_read_file_handler` correctly unwraps the envelope and supports both line-based and char-based slicing on the resulting navigable text.

**Screenshots / artefacts:**
- `scenario-A-stream.log` — full SSE stream for the chat turn
- `03-scenario-A-final.png` — copilot UI after the turn

---

### Scenario B — `bash_exec` redirect on SDK tool-result paths
**Steps:**
1. New chat session, prompt: "Use the bash_exec tool to run this exact command and tell me what it outputs: `cat /root/.claude/projects/test-cwd/sess-abc-123/tool-results/toolu_01result.json`"

**Tool sequence observed (from `scenario-B-stream.log`):**
- `bash_exec(command="cat /root/.claude/projects/test-cwd/sess-abc-123/tool-results/toolu_01result.json")` →
  ```
  {"type":"error",
   "message":"This path lives in the Claude Agent SDK's host-side tool-results directory, which is not mounted into the bash sandbox / workspace storage. Use one of:
     - `read_tool_result(file_path=…)` for direct reads (supports offset/limit; auto-unwraps the MCP envelope).
     - `@@agptfile:<absolute-path>[<start>-<end>]` inside another tool's argument …
   Offending fragment: '/root/.claude/projects/test-cwd/sess-abc-123/tool-results/toolu_01result.json'",
   "error":"sdk_tool_result_path_in_bash_command",
   …}
  ```
- The shell command was **not** executed (no `Permission denied` from the sandbox).

**Result:** PASS — pre-flight redirect short-circuits before reaching the sandbox; ErrorResponse names both `read_tool_result` and the `@@agptfile:` shorthand.

**Pre-flight predicate behaviour I verified (not a bug, worth flagging):**
`looks_like_sdk_tool_result_path()` uses an anchored regex that requires either (a) two `[^/\s]+/` segments between `.claude/projects/` and `tool-(results|outputs)/`, or (b) a `(toolu|mcp)_<id>.json` filename pattern. This means:
- `cat /root/.claude/projects/<one-segment>/tool-results/result.json` → does NOT trigger the redirect.
- `cat /root/.claude/projects/<cwd>/<sess>/tool-results/toolu_*.json` → DOES trigger.
- `cat tool-outputs/toolu_*.json` (shorthand) → DOES trigger.
- `cat tool-results/result.json` (no SDK-id prefix) → does NOT (intentional: avoids false positives on user repos).

The narrow shape is deliberate (PR comment explicitly mentions avoiding `my-pipeline/tool-outputs/data.json` false positives), and the SDK CLI always produces the matching shape, so prod paths are covered. Worth a follow-up only if SDK CLI ever emits a single-segment `projects/<x>/tool-results/...` path.

**Screenshot / artefacts:**
- `scenario-B-stream.log`
- `04-scenario-B-redirect.png`

---

### Scenario C — Regression: normal bash_exec still works
**Steps:**
1. New chat session, prompt: "Use bash_exec to run: `date -u`"

**Tool sequence observed (from `scenario-C-stream.log`):**
- `bash_exec(command="date -u")` →
  ```
  {"type":"bash_exec","message":"Command executed with status code 0",
   "stdout":"Thu May 21 09:28:14 UTC 2026\n","stderr":"","exit_code":0,"timed_out":false}
  ```

**Result:** PASS — non-SDK commands hit the sandbox normally; redirect is path-gated.

**Artefacts:**
- `scenario-C-stream.log`
- `05-scenario-C-regression.png`

---

### Scenario D — Backend unit tests on the running env
**Command:**
```
cd /Users/majdyz/Code/AutoGPT17/autogpt_platform/backend && poetry run pytest \
  backend/copilot/sdk/tool_adapter_test.py::TestNavigableToolResultText \
  backend/copilot/tools/bash_exec_test.py::TestBashExecSdkToolResultRedirect -xvs
```

**Result:** PASS — 6/6 tests:
- `test_envelope_with_json_payload_is_pretty_printed`
- `test_envelope_with_non_json_text_returns_unwrapped`
- `test_non_envelope_input_is_returned_unchanged`
- `test_redirect_on_absolute_sdk_path`
- `test_redirect_on_relative_tool_outputs_path`
- `test_normal_command_still_runs`

Additionally ran `workspace_files_test.py` redirect tests (the fourth PR change): 4/4 PASS (`test_read_workspace_file_redirects_sdk_tool_result_shorthand` and friends).

---

## Summary

| # | Scenario | Result | Notes |
|---|----------|--------|-------|
| A | Tool-result navigability | PASS | Model never reached for bash_exec on SDK paths; envelope-unwrap + line/char slicing verified directly against `_read_file_handler` with the exact prod context vars |
| B | bash_exec redirect | PASS | ErrorResponse with `sdk_tool_result_path_in_bash_command` returned; shell never invoked |
| C | bash_exec regression | PASS | `date -u` runs normally, exit_code=0 |
| D | Unit tests | PASS | 6/6 (PR-mandated) + 4/4 (workspace_files redirect) = 10/10 |

**Overall:** APPROVE.

The PR delivers exactly what's described: line- and char-navigable `read_tool_result` output (envelope auto-unwrapped, inner JSON pretty-printed) plus a defensive pre-flight redirect on both `bash_exec` and `read_workspace_file` that catches the SDK-tool-result path mistake without false-positiving on user workspace paths.

### Screenshots

autogpt_platform/backend/backend/copilot/tools/workspace_files.py
```diff
@@ -602,6 +604,15 @@ async def _execute(
...
+                    return ErrorResponse(
+                        message=sdk_tool_result_redirect_hint(path),
+                        session_id=session_id,
+                    )
```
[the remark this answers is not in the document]
majdyz: 🤖 Real bug, fixed in [8a19e65e9](https://github.com/Significant-Gravitas/AutoGPT/commit/8a19e65e96af7c7ad4d1d420849d2cb78592191c). Confirmed via the existing test class's `_clear_sdk_cwd` autouse fixture — the author had already noticed this footgun at test-write time and worked around it by clearing sdk_cwd in tests, but the source code still had the wrong ordering for real sessions.

Approach: insert a **relative-shorthand-only** redirect check before `is_allowed_local_path`. Absolute SDK paths still take the local-read fallback (legit `read_workspace_file` use case for fully-qualified SDK tool-results paths). The existing absolute-path redirect at the end stays as a safety net for absolute SDK paths that don't fall under `sdk_cwd`.

New test `test_redirect_wins_over_sdk_cwd_local_fallback` overrides the autouse fixture to deliberately set `sdk_cwd`, then asserts the redirect ErrorResponse fires AND `_read_local_tool_result` is NOT called — that's the bug class this commit pins.

majdyz:
## E2E Test Report — Live Dev Preview

Tested combined branch `temp/dev-preview-13179-13180` deployed to `dev-builder.agpt.co` (workflow run [26221945215](https://github.com/Significant-Gravitas/AutoGPT_cloud_infrastructure/actions/runs/26221945215)). The temp branch merges PR #13179 + PR #13180 on top of `dev` (`1f22a0badf6`). All 5 deploy jobs passed (setup, build-backend, deploy-backend, deploy-frontend-vercel, verify-deployment).

| 1 | Login to dev preview | PASS | Auth flow against Supabase succeeds, lands on `/copilot` |
| 2 | Preview banner shows temp branch | PASS | Banner: "This is a Preview build for the branch: temp/dev-preview-13179-13180" |
| 3 | Copilot accepts message + completes task | PASS | "Thought for 59s · 2 agents created" |
| 4 | **Orchestrator dry-run succeeds with default config** (PR #13180 fix) | PASS | Copilot built an agent containing `OrchestratorBlock (claude-sonnet-4-6, standard mode)` and saved it. Pre-#13180, the dry-run failed with either an enum-mismatch (claude snapshot ID rejected by OR `/v1/chat/completions`) or an SDK `x-api-key` 401. |
| 5 | Agent persists to library | PASS | "One-Sentence Responder" appears at `/library`, opens to detail at `library/agents/3bf7fef4-...` |
| 6 | Agent graph saved correctly | PASS | Builder canvas shows: AgentInput → **Orchestrator (Claude Sonnet 4.6)** → One-Sentence Generator (Claude Haiku 4.5) → AgentOutput, all wired up |
| 7 | Copilot tool-result reads succeed (PR #13179) | PASS | The 59s session involved multiple tool calls that all returned cleanly — no `read_tool_result` redirect bypass errors, no workspace-file 404 noise in the response stream |
| 8 | Credit balance + Stripe webhook dedup (PR #13179) | NOT DIRECTLY TESTABLE FROM UI | These paths are exercised by backend tests in CI (which passed). Manual webhook replay was not in scope for this dev-preview smoke. |

### Screenshots

**1 — Login page (dev-builder.agpt.co)**
01-login-page-before
The login form rendered cleanly with email/password + Google SSO; baseline screenshot before any state change.

**2 — Post-login state**
02-after-login
Successful redirect to `/copilot` after submitting credentials.

**3 — Copilot landing with preview banner**
03-copilot-landing-preview-banner
Top green banner confirms this is the `temp/dev-preview-13179-13180` build. Credits balance $1126.60 displayed. Existing agents (`AI Debate Podcast Moderator`, `Hello World Agent`) shown as recent activity.

**4 — Chat ready to receive prompt**
05-copilot-before-prompt
Composer focused, ready to accept the test prompt.

**5 — Prompt submitted**
06-copilot-prompt-sent
Prompt "Build a tiny agent that uses the Orchestrator block..." sent; copilot starts "Thinking".

**6 — Copilot response: orchestrator-based agent created**
07-copilot-response-with-orchestrator
The decisive result: copilot reports it built an agent with **`OrchestratorBlock (claude-sonnet-4-6, standard mode)`** + AITextGeneratorBlock as a downstream tool. "Thought for 59s · 2 agents created". This is the exact code path PR #13180 fixed.

**7 — Agent persisted to library**
10-library-with-new-agent
"One-Sentence Responder" is the leftmost card in the agent listing.

**8 — Agent detail view**
11-agent-detail-page
Library detail page with the agent loaded — "Setup your task", "Edit agent" controls available.

**9 — Agent graph in builder (proves the saved config)**
12-agent-in-builder
Builder canvas showing the 4-block graph: AgentInput → **Orchestrator (LLM Model: Claude Sonnet 4.6)** → One-Sentence Generator (Claude Haiku 4.5) → AgentOutput. The Orchestrator node has the Anthropic credentials picker visible ("Use Credits for..." / "Add API key"). This is the persisted graph state — the dry-run that PR #13180 unblocked succeeded against this exact configuration.

### Negative coverage

The copilot's first attempt during this session returned "Validation failed with 1 error — The OrchestratorBlock requires at least one downstream tool block", and recovered by adding an AITextGenerator passthrough. This is the **expected validation error path** (legitimate user-error case: an orchestrator without tools is invalid), distinct from the SDK auth / model-enum failures that PR #13180 fixed. It proves the validation pipeline still rejects malformed graphs as it should.

### Verdict

The dev preview is functionally healthy after the combined merge of PR #13179 + PR #13180. The OrchestratorBlock dry-run — the specific path that motivated PR #13180 — works against the default (`claude-sonnet-4-6`, `standard` mode) config used by copilot.

PR #13192, opened 2026-05-22
feat(platform): copilot followups UI

majdyz:
## Why
Copilot can already schedule follow-up turns for itself (#13190), but there is no surface for the user to see what's pending or cancel one — they only see the side-effect when the copilot speaks up later. This adds the minimal Library page that lists them and lets the user cancel.

Stacked on top of #13190 — this PR's diff will collapse to just the UI commit once that lands.

## What
- New page at `/library/followups` listing the current user's pending copilot follow-ups, with:
  - Empty state when the user has none.
  - One row per scheduled turn showing message preview, next-run label (relative), recurrence label (humanised cron) and a session-id tag. Clicking the row deep-links to the originating `/copilot?sessionId=...`.
  - A `Cancel` action that opens a confirmation dialog and calls the existing `DELETE /api/schedules/{id}` route; success toasts and refetches, failure toasts destructively.
- New backend route `GET /api/schedules/followups` returning only `CopilotTurnJobInfo` rows for the caller. Split out from the polymorphic `GET /api/schedules` so the OpenAPI-generated client gets a single concrete return type per route instead of a discriminated union.

## How
- Backend: thin wrapper around `Scheduler.get_execution_schedules(kind="copilot_turn")` plus an explicit `operation_id="listCopilotFollowupSchedules"` to keep the generated TS client tidy.
- Frontend: page + sub-components (`EmptyFollowups`, `FollowupListItem`) following the `page.tsx + usePageName.ts + helpers.ts` pattern in `frontend/AGENTS.md`. Design-system atoms/molecules only (`Text`, `Button`, `Dialog`, `LoadingSpinner`, `ErrorCard`), Phosphor icons only.
- Data fetching via the Orval-generated `useListCopilotFollowupSchedules` hook; delete reuses `useDeleteV1DeleteExecutionSchedule` and invalidates the followups query key on success.
- Tests:
  - Backend route test mocks the scheduler client and asserts the `kind="copilot_turn"` filter is forwarded and non-copilot rows are dropped.
  - Frontend integration tests (Vitest + RTL + MSW with Orval-generated handlers) cover: empty list, one row per followup, deep link href, delete success toast, delete failure toast.

## Checklist
- [x] My code follows the project's style guidelines
- [x] I have added tests for my changes
- [x] My changes generate no new warnings
- [x] I have ensured that my changes are backwards compatible

### DIFF
--- a/autogpt_platform/backend/backend/api/features/v1.py
+++ b/autogpt_platform/backend/backend/api/features/v1.py
@@ -2317,6 +2317,28 @@ async def list_all_graphs_execution_schedules(
     return await get_scheduler_client().get_graph_execution_schedules(user_id=user_id)
 
 
+@v1_router.get(
+    path="/schedules/followups",
+    summary="List copilot follow-up schedules for a user",
+    operation_id="listCopilotFollowupSchedules",
+    tags=["schedules"],
+    dependencies=[Security(requires_user)],
+)
+async def list_copilot_turn_schedules(
+    user_id: Annotated[str, Security(get_user_id)],
+) -> list[scheduler.CopilotTurnJobInfo]:
+    """Return only copilot-turn schedules for the current user.
+
+    Sibling of :func:`list_all_graphs_execution_schedules`; one route per kind
+    keeps the generated frontend client typed to a single concrete return type
+    instead of a discriminated union.
+    """
+    schedules = await get_scheduler_client().get_execution_schedules(
+        user_id=user_id, kind="copilot_turn"
+    )
+    return [s for s in schedules if isinstance(s, scheduler.CopilotTurnJobInfo)]
+
+
 @v1_router.delete(
     path="/schedules/{schedule_id}",
     summary="Delete execution schedule",
--- a/autogpt_platform/backend/backend/api/features/v1.py
+++ b/autogpt_platform/backend/backend/api/features/v1.py
@@ -2336,7 +2336,12 @@ async def list_copilot_turn_schedules(
     schedules = await get_scheduler_client().get_execution_schedules(
         user_id=user_id, kind="copilot_turn"
     )
-    return [s for s in schedules if isinstance(s, scheduler.CopilotTurnJobInfo)]
+    # The ``kind="copilot_turn"`` filter is the source of truth — every row in
+    # the result IS a ``CopilotTurnJobInfo``. The cast narrows the static type
+    # from the polymorphic ``list[GraphExecutionJobInfo | CopilotTurnJobInfo]``
+    # without an at-runtime second filter that would silently drop rows if the
+    # discriminator ever drifted.
+    return cast(list[scheduler.CopilotTurnJobInfo], schedules)
 
 
 @v1_router.delete(
--- a/autogpt_platform/frontend/src/app/(platform)/library/followups/components/FollowupListItem/FollowupListItem.tsx
+++ b/autogpt_platform/frontend/src/app/(platform)/library/followups/components/FollowupListItem/FollowupListItem.tsx
@@ -1,0 +1,121 @@
+"use client";
+
+import type { CopilotTurnJobInfo } from "@/app/api/__generated__/models/copilotTurnJobInfo";
+import { Button } from "@/components/atoms/Button/Button";
+import { Text } from "@/components/atoms/Text/Text";
+import { Dialog } from "@/components/molecules/Dialog/Dialog";
+import { ChatCircleTextIcon, TrashIcon } from "@phosphor-icons/react";
+import Link from "next/link";
+import { useFollowupListItem } from "./useFollowupListItem";
+
+interface Props {
+  followup: CopilotTurnJobInfo;
+  onDeleted?: () => void;
+}
+
+export function FollowupListItem({ followup, onDeleted }: Props) {
+  const {
+    sessionHref,
+    nextRunLabel,
+    nextRunTitle,
+    recurrenceLabel,
+    messagePreview,
+    isDeleteOpen,
+    openDelete,
+    closeDelete,
+    isDeleting,
+    handleDelete,
+  } = useFollowupListItem({ followup, onDeleted });
+
+  return (
+    <div
+      className="flex w-full flex-col gap-3 rounded-large border border-zinc-200 bg-white p-4 sm:flex-row sm:items-center sm:justify-between"
+      data-testid="followup-row"
+      data-followup-id={followup.id}
+    >
+      <Link
+        href={sessionHref}
+        className="flex min-w-0 flex-1 items-start gap-3 hover:opacity-80"
+        data-testid="followup-open-session"
+      >
+        <div className="flex h-9 w-9 flex-shrink-0 items-center justify-center rounded-large border border-slate-50 bg-yellow-50">
+          <ChatCircleTextIcon
+            size={18}
+            className="text-yellow-700"
+            weight="bold"
+          />
+        </div>
+        <div className="flex min-w-0 flex-col gap-1">
+          <Text
+            variant="body-medium"
+            className="block w-full truncate text-ellipsis"
+          >
+            {messagePreview}
+          </Text>
+          <div className="flex flex-wrap items-center gap-x-2 gap-y-0.5">
+            <Text
+              variant="small"
+              className="!text-zinc-500"
+              title={nextRunTitle}
+            >
+              {nextRunLabel}
+            </Text>
+            <span className="text-zinc-300">•</span>
+            <Text variant="small" className="!text-zinc-500">
+              {recurrenceLabel}
+            </Text>
+            <span className="text-zinc-300">•</span>
+            <Text variant="small" className="!text-zinc-400">
+              Session {followup.session_id.slice(0, 8)}
+            </Text>
+          </div>
+        </div>
+      </Link>
+
+      <div className="flex flex-shrink-0 items-center gap-2">
+        <Button
+          variant="secondary"
+          size="small"
+          onClick={openDelete}
+          data-testid="followup-cancel-button"
+          aria-label="Cancel follow-up"
+        >
+          <TrashIcon className="mr-1 h-4 w-4" />
+          Cancel
+        </Button>
+      </div>
+
+      <Dialog
+        controlled={{ isOpen: isDeleteOpen, set: closeDelete }}
+        styling={{ maxWidth: "32rem" }}
+        title="Cancel follow-up"
+      >
+        <Dialog.Content>
+          <div className="flex flex-col gap-4">
+            <Text variant="large">
+              Cancel this scheduled follow-up? The copilot will not send the
+              message and you can recreate it from chat if needed.
+            </Text>
+            <Dialog.Footer>
+              <Button
+                variant="secondary"
+                disabled={isDeleting}
+                onClick={() => closeDelete(false)}
+              >
+                Keep follow-up
+              </Button>
+              <Button
+                variant="destructive"
+                onClick={handleDelete}
+                loading={isDeleting}
+                data-testid="followup-confirm-cancel"
+              >
+                Cancel follow-up
+              </Button>
+            </Dialog.Footer>
+          </div>
+        </Dialog.Content>
+      </Dialog>
+    </div>
+  );
+}
--- a/autogpt_platform/frontend/src/app/(platform)/library/followups/components/FollowupListItem/FollowupListItem.tsx
+++ b/autogpt_platform/frontend/src/app/(platform)/library/followups/components/FollowupListItem/FollowupListItem.tsx
@@ -10,10 +10,9 @@ import { useFollowupListItem } from "./useFollowupListItem";
 
 interface Props {
   followup: CopilotTurnJobInfo;
-  onDeleted?: () => void;
 }
 
-export function FollowupListItem({ followup, onDeleted }: Props) {
+export function FollowupListItem({ followup }: Props) {
   const {
     sessionHref,
     nextRunLabel,
@@ -25,7 +24,7 @@ export function FollowupListItem({ followup, onDeleted }: Props) {
     closeDelete,
     isDeleting,
     handleDelete,
-  } = useFollowupListItem({ followup, onDeleted });
+  } = useFollowupListItem({ followup });
 
   return (
     <div
@@ -78,22 +77,22 @@ export function FollowupListItem({ followup, onDeleted }: Props) {
           size="small"
           onClick={openDelete}
           data-testid="followup-cancel-button"
-          aria-label="Cancel follow-up"
+          aria-label="Delete follow-up"
         >
           <TrashIcon className="mr-1 h-4 w-4" />
-          Cancel
+          Delete
         </Button>
       </div>
 
       <Dialog
         controlled={{ isOpen: isDeleteOpen, set: closeDelete }}
         styling={{ maxWidth: "32rem" }}
-        title="Cancel follow-up"
+        title="Delete follow-up"
       >
         <Dialog.Content>
           <div className="flex flex-col gap-4">
             <Text variant="large">
-              Cancel this scheduled follow-up? The copilot will not send the
+              Delete this scheduled follow-up? The copilot will not send the
               message and you can recreate it from chat if needed.
             </Text>
             <Dialog.Footer>
@@ -102,7 +101,7 @@ export function FollowupListItem({ followup, onDeleted }: Props) {
                 disabled={isDeleting}
                 onClick={() => closeDelete(false)}
               >
-                Keep follow-up
+                Keep it
               </Button>
               <Button
                 variant="destructive"
@@ -110,7 +109,7 @@ export function FollowupListItem({ followup, onDeleted }: Props) {
                 loading={isDeleting}
                 data-testid="followup-confirm-cancel"
               >
-                Cancel follow-up
+                Yes, delete
               </Button>
             </Dialog.Footer>
           </div>
--- a/autogpt_platform/frontend/src/app/(platform)/library/followups/components/FollowupListItem/useFollowupListItem.ts
+++ b/autogpt_platform/frontend/src/app/(platform)/library/followups/components/FollowupListItem/useFollowupListItem.ts
@@ -12,10 +12,9 @@ import { describeFollowup } from "./helpers";
 
 interface Args {
   followup: CopilotTurnJobInfo;
-  onDeleted?: () => void;
 }
 
-export function useFollowupListItem({ followup, onDeleted }: Args) {
+export function useFollowupListItem({ followup }: Args) {
   const { toast } = useToast();
   const queryClient = useQueryClient();
   const [isDeleteOpen, setIsDeleteOpen] = useState(false);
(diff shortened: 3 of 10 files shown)


autogpt_platform/backend/backend/api/features/v1.py
```diff
@@ -2317,6 +2317,28 @@ async def list_all_graphs_execution_schedules(
...
+    """Return only copilot-turn schedules for the current user.
+
+    Sibling of :func:`list_all_graphs_execution_schedules`; one route per kind
+    keeps the generated frontend client typed to a single concrete return type
+    instead of a discriminated union.
+    """
+    schedules = await get_scheduler_client().get_execution_schedules(
+        user_id=user_id, kind="copilot_turn"
+    )
+    return [s for s in schedules if isinstance(s, scheduler.CopilotTurnJobInfo)]
```
majdyz: 🤖 🟡 **Nice to Have**: The `kind="copilot_turn"` filter on `get_execution_schedules` already guarantees every returned row IS a `CopilotTurnJobInfo`. The trailing `[s for s in schedules if isinstance(s, scheduler.CopilotTurnJobInfo)]` is defensive double-filtering — at best a tautology, at worst it silently drops rows if the discriminator ever drifts (e.g. a future `kind="copilot_turn"` job stored as a slightly different subclass). Cleaner: `return cast(list[scheduler.CopilotTurnJobInfo], schedules)` (or trust the discriminator and skip the cast). Add a short comment if you want the safety net documented; the current shape reads like "we don't trust our own filter".

autogpt_platform/frontend/src/app/(platform)/library/followups/components/FollowupListItem/useFollowupListItem.ts
```diff
@@ -0,0 +1,90 @@
...
+
+  function closeDelete(open: boolean) {
+    setIsDeleteOpen(open);
+  }
+
+  async function handleDelete() {
+    try {
+      await deleteSchedule({ scheduleId: followup.id });
+      toast({ title: "Follow-up cancelled" });
+      setIsDeleteOpen(false);
```
majdyz, line 52: 🤖 🟡 **Nice to Have**: `handleDelete` both invalidates the query and calls `onDeleted` (which the page wires to `refetchFollowups`). `invalidateQueries` already triggers an automatic refetch on next render, so the explicit `refetch()` is redundant — and in the failure path you'll see a stale row briefly before the error toast lands because the refetch races the toast. Either drop `onDeleted` entirely (let invalidation do the work) or drop the `invalidateQueries` call (rely on the parent's refetch). Pick one source of truth.

autogpt_platform/frontend/src/app/(platform)/library/followups/page.tsx
```diff
@@ -0,0 +1,65 @@
+"use client";
+
+import { useEffect } from "react";
+import { Text } from "@/components/atoms/Text/Text";
+import { ErrorCard } from "@/components/molecules/ErrorCard/ErrorCard";
+import { LoadingSpinner } from "@/components/atoms/LoadingSpinner/LoadingSpinner";
+import { EmptyFollowups } from "./components/EmptyFollowups/EmptyFollowups";
+import { FollowupListItem } from "./components/FollowupListItem/FollowupListItem";
+import { useFollowupsPage } from "./useFollowupsPage";
+
```
majdyz, line 10: 🤖 🟡 **Nice to Have**: `/library/followups` ships with no nav link — discoverability hole. Users who already know the URL can hit it, but anyone else only ever sees followups via chat (which is the gap this PR is supposedly closing). Two cheap options:

1. Add a Followups entry to the library sidebar (next to wherever `/library` lives).
2. Add an inline link from the `/copilot` session header when there's at least one scheduled followup.

(2) is more contextual but heavier. (1) is the minimum viable add. If you really want to dark-launch, call that out in the PR description so reviewers know it's intentional and add a tracking issue for the nav hookup.

autogpt_platform/frontend/src/app/(platform)/library/followups/components/FollowupListItem/FollowupListItem.tsx
```diff
@@ -0,0 +1,121 @@
...
+          variant="secondary"
+          size="small"
+          onClick={openDelete}
+          data-testid="followup-cancel-button"
+          aria-label="Cancel follow-up"
+        >
+          <TrashIcon className="mr-1 h-4 w-4" />
+          Cancel
+        </Button>
+      </div>
```
majdyz, line 85: 🤖 🔵 **Nit**: The "Cancel" button has a dialog whose primary action also says "Cancel follow-up" while the back action is "Keep follow-up". "Cancel" overloads here (cancel-the-schedule vs cancel-the-dialog). Possibly less ambiguous:

- Row button: `Cancel` → leave as is.
- Dialog destructive: `Yes, cancel it` (or `Confirm cancel`).
- Dialog back: `Keep it` (already good).

Or just label the row action `Delete` and accept that the action IS a delete — current code already calls `useDeleteV1DeleteExecutionSchedule`, so the rename surfaces the truth.

majdyz:
## E2E Test Report

# E2E Test Report: PR #13192 — feat(platform): copilot followups UI

- Date: 2026-05-22
- Branch: `feat/copilot-schedule-followup-ui`
- Worktree: `/Users/majdyz/Code/AutoGPT24`
- Stack: native — `poetry run app` (backend :8006) + `pnpm dev` (frontend :3000), docker deps profile for supabase/redis/rabbitmq/clamav.

## Environment notes

- Switched stack from PR #13190's `AutoGPT21` worktree to `AutoGPT24` so the new `/api/schedules/followups` route is live. Confirmed via `GET /openapi.json`.
- `pnpm install` re-run in the worktree to pick up `@number-flow/react` (declared in package.json but missing in node_modules, plus a `next` minor bump).
- Copilot uses `CHAT_USE_CLAUDE_CODE_SUBSCRIPTION=true` (no fallback needed).
- Test user `test@test.com / testtest123`.

## Test Results

| # | Scenario | Result |
|---|---|---|
| 1 | Cold load of `/library/followups` empty state | PASS |
| 2 | Schedule 2 followups via copilot chat (cron + delay_seconds) | PASS |
| 3 | `/library/followups` shows 2 rows with msg / next-run / cron-or-once / session prefix | PASS |
| 4 | Row click opens `/copilot?sessionId=…` with full session history | PASS |
| 5 | Delete one-shot via Cancel button + dialog → row disappears, toast shown | PASS |
| 6 | Copilot "list my schedules" matches UI state | PASS |
| 7 | API failure → ErrorCard renders | PASS |
| 8 | Page handles copilot_turn-only filter with zero rows | PASS |

## Detail

### 1. Cold load — empty state
Page renders header "Copilot follow-ups" + `EmptyFollowups` card "No copilot follow-ups — Ask your copilot to schedule something for later …". No errors. (`02-followups-empty-state.png`)

### 2. Schedule via copilot chat
Two follow-ups created in chat session `96d58196-3e3c-47de-ab70-4ebf11d21e61`:
- one-shot `PR 13192 one-shot test`, `delay_seconds=120`
- cron `PR 13192 daily cron test`, cron `0 9 * * *` (interpreted in `Asia/Jakarta`, the user's tz)

API verified — `GET /api/schedules/followups` returns 2 `CopilotTurnJobInfo` items with the right session_id, cron/run_at fields. (`03-copilot-chat-scheduled-2-followups.png`)

### 3. List page rendering
Both rows visible with message preview, "Next in less than a minute" / "Next in about 17 hours", "Runs once" / "Every day at 09:00", "Session 96d58196". Cron expression humanized correctly. (`04-followups-list-2-rows.png`)

### 4. Deep link click
`data-testid="followup-open-session"` link uses `?sessionId=…` (frontend convention; the task informally referred to `session_id`). Click navigated to `/copilot?sessionId=96d58196-3e3c-47de-ab70-4ebf11d21e61` with full session history visible (previous scheduling messages + tool outputs). (`05-deep-link-copilot-session.png`)

### 5. Delete
Original one-shot fired before I could click delete (only 120s), so I scheduled a new one with `delay_seconds=3600` and deleted that one. Clicking the Delete button opened the `Dialog` ("Delete follow-up — Delete this scheduled follow-up?") with `Keep it` / `Yes, delete` buttons. Confirming triggered `DELETE /api/schedules/{id}`, toast "Follow-up deleted" appeared, row was removed via `getListCopilotFollowupSchedulesQueryKey` invalidation. API confirms the schedule is gone. (`06-…before-delete.png`, `07-…dialog.png`, `08-…after-delete.png`)

Note: button copy is **"Delete"** (with `data-testid="followup-cancel-button"` and `aria-label="Delete follow-up"`). The testid name is a leftover from earlier "Cancel" wording — minor consistency nit, not a blocker.

### 6. Copilot cross-check
"List my current follow-up schedules" returned `Found 1 schedule(s).` matching the UI's state (only the cron `PR 13192 daily cron test` remained). Copilot also noted the recently-deleted one-shot was missing, consistent with what we did in the UI. (`09-copilot-cross-check-list.png`)

### 7. Network / error handling
Used `agent-browser network route 'http://localhost:3000/api/proxy/api/schedules/followups' --abort` + `window.location.reload()`. Page renders `ErrorCard` with "Something went wrong — We had the following error when retrieving copilot follow-ups:" + Report Error button. No blank page / crash. (`10-error-card.png`)

### 8. Empty state with zero rows
After deleting the cron schedule via direct API DELETE, the page re-mounts to the `EmptyFollowups` empty state without trying to render any graph-only fields. No console errors. (`11-empty-state-after-cleanup.png`)

## Summary

- Total: 8
- Passed: 8
- Failed: 0
- Bugs found / fixed: none functional. Minor copy inconsistency: `data-testid="followup-cancel-button"` survives a rename from "Cancel" → "Delete" — non-blocking.

## Loose ends

- The one-shot fired automatically before I could test delete on it; covered by scheduling a longer-delay one-shot. No issue with the feature itself.
- Cron 'Asia/Jakarta' inference (vs. UTC the user asked for) is `schedule_followup` tool behaviour from PR #13190, out of scope for this PR.

### Screenshots

autogpt_platform/backend/backend/api/features/v1.py
majdyz: 🤖 Fixed in [a59947a5a](https://github.com/Significant-Gravitas/AutoGPT/commit/a59947a5ae9fbe1122fcf4f17c7b3d821c550214): replaced the trailing isinstance filter with `cast(list[scheduler.CopilotTurnJobInfo], schedules)`. Comment documents why the cast is safe — the `kind="copilot_turn"` filter is the source of truth.

autogpt_platform/frontend/src/app/(platform)/library/followups/components/FollowupListItem/useFollowupListItem.ts
majdyz: 🤖 Fixed in [a59947a5a](https://github.com/Significant-Gravitas/AutoGPT/commit/a59947a5ae9fbe1122fcf4f17c7b3d821c550214): dropped `onDeleted` from both the hook and the page wiring. `invalidateQueries` is now the single source of truth — React Query refetches automatically, no race.

autogpt_platform/frontend/src/app/(platform)/library/followups/page.tsx
majdyz: 🤖 Tracked as follow-up — page is dark-launched intentionally for this PR (UI-mechanics-only). The nav-link decision belongs with the broader copilot navigation work (where Followups should sit relative to Library, Marketplace, etc). Will open a small follow-up PR adding a sidebar entry once direction is clearer.

autogpt_platform/frontend/src/app/(platform)/library/followups/components/FollowupListItem/FollowupListItem.tsx
majdyz: 🤖 Fixed in [a59947a5a](https://github.com/Significant-Gravitas/AutoGPT/commit/a59947a5ae9fbe1122fcf4f17c7b3d821c550214): renamed everywhere — button `Cancel` → `Delete`, dialog title `Delete follow-up`, destructive button `Yes, delete`, back button `Keep it`, toast titles `Follow-up deleted` / `Failed to delete follow-up`. Tests updated to match.

autogpt_platform/backend/backend/copilot/tools/schedule_followup.py
```diff
@@ -0,0 +1,247 @@
...
+            if is_recurring
+            else f"once at {info.next_run_time}"
+        )
+        return ScheduleCreatedResponse(
+            message=f"Follow-up scheduled {when_str}{target_note}.",
+            schedule_id=info.id,
+            next_run_time=info.next_run_time,
+            is_recurring=is_recurring,
+            session_id=current_session_id,
+        )
```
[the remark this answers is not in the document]
majdyz: 🤖 Not in scope for this PR. `schedule_followup.py` lives on PR #13190 (this branch is now rebased on top of #13190 only, base swap commit 14756146d9). The `_execute` complexity concern applies to #13190 — leaving the suggestion there if the author wants to act on it. Resolving here as out-of-PR-scope.

majdyz:
## Roadmap — copilot followup management

Capturing the bigger plan so the scope of this PR (and what's still ahead) is explicit. **This PR is the first slice — the UI page only.** Everything below is what comes next.

### What we're ultimately trying to achieve

When a user changes their mind about a scheduled followup, the **happy path is conversational**, not a separate UI ritual:

> User: *"actually never mind, cancel that"*
> AutoPilot: *"Got it — cancelled the followup that was set for tomorrow 9am."*

The UI page added by this PR (`/library/followups`) is the **fallback** — for users who want to see everything in one place, audit, or batch-delete. The primary cancel/modify path should live in the chat.

### Why the chat-mediated path is P0 over UI affordances

1. **Same surface that created the schedule cancels it.** Users who scheduled a followup via natural language ("check the CI in 20 min") will instinctively cancel via natural language. Forcing them to a list page is a context switch.
2. **Modification is multi-step in the UI** (read → edit modal → save). In chat it's one sentence.
3. **The model already has the conversational context** to disambiguate ("cancel the CI check" vs "cancel the morning summary").

So the UI is the "atlas view", chat is the "GPS".

### What's required for AutoPilot to do this

For the model to honour *"cancel that"*, it needs three pieces of information that today's prompt does not surface:

1. **Whether anything is scheduled on this session at all.** Otherwise the model has no signal to bring it up proactively ("by the way, you have a followup at 9am").
2. **The current session id**, so when the model calls `list_schedules`/`delete_schedule` it can filter to *this* conversation (passing `session_id` to `list_schedules` is supported on #13190 — the model needs to know which id to pass).
3. **A reliable way to enumerate + act on its own schedules**, which we already have via `list_schedules(session_id=...)` and `delete_schedule(schedule_id=...)`.

The first two are the open work.

### The prompt-caching constraint (this is the load-bearing concern)

Anthropic prompt caching keys off **prefix bytes**. If we naively inject `session_id` into the system prompt or any cacheable prefix, **cache hit rate drops to 0% across sessions** because every session has a unique UUID. That's a >50% regression on the latency/cost dial we just spent weeks tuning.

Concrete rule: **`session_id` cannot live in the cached prefix**. The placement options, ranked by safety:

| Placement | Cache-safe? | Notes |
|---|---|---|
| Inside latest **user message** (not system) | ✅ | User-message content is past the last `cache_control` breakpoint. New per-turn anyway. |
| Injected via PostToolUse / **tool-result text** | ✅ | Tool results are also past the cacheable prefix. Could be returned by a new `get_my_context()` tool. |
| `system_post` (some Anthropic-format clients support a post-context block) | ⚠️ | Verify with our SDK's exact placement — must land AFTER the final `cache_control: ephemeral` marker. |
| In the system prompt | ❌ | Every session busts cache. **Do not do this.** |
| Per-tool description (e.g. "your session_id is X") | ❌ | Tools are part of the cacheable prefix in our setup. Same problem. |

### Proposed slices (post-#13192)

**Slice A — Session self-awareness for the model.** Lowest-risk, highest-leverage.

- Add a single line to the latest user message at turn dispatch time: `[context] current_session_id=<uuid>; pending_followups=<n>` (or null when zero). This lands AFTER the last cache breakpoint, so it does NOT bust the prefix cache. Compute `pending_followups` cheaply via the existing `list_schedules(session_id=current)` call — already O(1) per turn.
- Update the schedule-related tools' descriptions (NOT the values — descriptions are cached) to mention: *"if you weren't given a session_id explicitly, use the session_id from the [context] line in this turn's user message."*
- Cache-impact test: run a A/B with one with-context-line / one without on identical messages, confirm `cache_read_input_tokens` is identical before the user-message block.

**Slice B — Proactive followup reminders in chat.** Optional.

- When the user opens a session that has pending followups, the autopilot's *first* response includes a one-liner: *"You have 2 followups pending: 'Check CI' (in 20m), 'Daily summary' (every weekday 9am). Say 'cancel X' to remove one."*
- Implementable as a system_post note or a synthetic tool-result injected before the model runs.

**Slice C — Modification via chat.** *"Move that to 8am instead"* — needs a `replace_followup_schedule(old_id, new_when, new_message)` tool. Could compose `delete_schedule + schedule_followup` but as a single tool call avoids the orphan-window between delete and re-create.

**Slice D — UI nav link.** Add Followups to the library sidebar (the thread above tracks this). Trivial change, but easier once we know where Library / Marketplace / Copilot sit relative to each other in the broader nav redesign.

### Are we passing session_id today?

**No** — verified by grepping `backend/copilot/sdk/` and `backend/copilot/prompting.py`. The model has zero awareness of its own session UUID. The closest signal is conversation history (which can hint via prior schedule_followup calls), but a fresh session inheriting an old transcript would have no way to disambiguate. Slice A is genuinely net-new visibility, not just "expose what's already there".

### Why this is in *this* PR's comments and not a separate issue

This PR (#13192) is the UI page slice. The next slice (A) doesn't belong in #13192 because:

- It modifies the **copilot prompt / SDK adapter**, which is a different review surface (backend/copilot, not platform/library).
- Stacking it would make the diff cross two concerns and slow review of both.
- The slices are independent: this PR is shippable without A, and A is shippable without the UI page (it just makes chat-mediated cancel work).

Once #13190 + #13192 land, I'll open a fresh PR for Slice A with the cache-impact A/B test attached as PR-test evidence.

### Open questions worth flagging now

1. **Should `[context]` injection happen for every turn, or only the first turn of a session?** Every-turn is safe (it lands past the cache boundary) but is ~50 extra tokens per turn at scale. First-turn-only is cheaper but adds state to the SDK adapter.
2. **What's the right `pending_followups` truncation?** Showing 5 inline before "…+12 more" is probably fine. Beyond that the message bloats and the user should be sent to `/library/followups`.
3. **Race between "cancel that" and the schedule firing.** If the model says "cancelled" but the fire-time is 30s away, do we accept the race or coalesce? Probably accept — the worst case is one redundant turn arrives.

Tagging this so we don't forget when #13190 + #13192 are in dev.

autogpt_platform/frontend/src/app/(platform)/library/followups/page.tsx
```diff
@@ -0,0 +1,65 @@
...
+        <ul
+          className="flex flex-col gap-3"
+          data-testid="followups-list"
+          aria-label="Copilot follow-ups"
+        >
+          {followups.map((followup) => (
+            <li key={followup.id}>
+              <FollowupListItem
+                followup={followup}
+                onDeleted={refetchFollowups}
```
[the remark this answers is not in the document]
majdyz: 🤖 Already fixed in [a375f3723](https://github.com/Significant-Gravitas/AutoGPT/commit/a375f3723fb86107fa05a6b9d76dc7f1b96b6abd) on the parent branch (`feat/copilot-schedule-followup`) — the stale `onDeleted={refetchFollowups}` usage was dropped from `page.tsx` so `<FollowupListItem followup={followup} />` has no reference error. This PR's own commit `a59947a5a` removed `refetchFollowups` from the destructure but left the JSX usage; the parent-branch follow-up commit cleared that. PR is merged, sentry's catch was correct on the snapshot it reviewed.

PR #13193, opened 2026-05-22
fix(monitoring): exclude user-supplied invalid API key errors from block error rate alerts

Torantulino:
## Problem

Block error rate alerts were firing for `HTTP 401 invalid_api_key` errors caused by users supplying their own expired/invalid API keys. These are not platform failures and should not trigger alerts.

## Changes

**`backend/data/execution.py`**
- Added `user_api_key_error_executions` to `BlockErrorStats` — counts failed executions where user credentials are present and the error contains `invalid_api_key`
- Added `platform_failed_executions` and `platform_error_rate` properties that exclude these user-caused failures

**`backend/monitoring/block_error_monitor.py`**
- Same fields and properties added to `BlockStatsWithSamples`
- Alert threshold checks now use `platform_error_rate` instead of `error_rate`
- Alert messages report `platform_failed_executions` counts
- When user API key errors are present, a note is appended to the alert for visibility without triggering a false-positive page


codeCraft-Ritik (commented):
olid monitoring enhancement — switching alert thresholds to platform_error_rate seems like the right approach here.
The additional context in alert messages is also a nice touch.

autogpt_platform/backend/backend/data/execution.py
```diff
@@ -1593,7 +1606,14 @@ async def get_block_error_stats(
     SELECT 
         n."agentBlockId" as block_id,
         COUNT(*) as total_executions,
-        SUM(CASE WHEN ne."executionStatus" = 'FAILED' THEN 1 ELSE 0 END) as failed_executions
+        SUM(CASE WHEN ne."executionStatus" = 'FAILED' THEN 1 ELSE 0 END) as failed_executions,
+        SUM(CASE
+            WHEN ne."executionStatus" = 'FAILED'
+                AND ne."executionData"->>'credentials' IS NOT NULL
+                AND ne."stats"->>'error' LIKE '%invalid_api_key%'
+                AND (ne."executionData"->'credentials'->>'is_managed')::boolean IS NOT TRUE
```
majdyz: 🤖 🔴 **Blocker**: The `is_managed` guard never matches in production data.

What gets stored in `executionData.credentials` is the `CredentialsMetaInput` dict (`backend/data/model.py:514`), whose only fields are `{id, title, provider, type}` — it intentionally does NOT include `is_managed`. The resolved credential (with `is_managed`) lives in the credentials store and is injected into `extra_exec_kwargs` in `backend/executor/manager.py:294`, not persisted into `executionData`.

Failure scenario:
1. Managed-credential block (e.g. agentmail, ayrshare) fails internally with an error containing `invalid_api_key` (e.g. our managed pod key was rotated/expired and a downstream wrapper happens to surface the upstream message).
2. SQL checks `(executionData->'credentials'->>'is_managed')::boolean IS NOT TRUE`. Since `is_managed` key is absent, `->>` returns NULL, `::boolean` is NULL, `IS NOT TRUE` evaluates **TRUE**.
3. The failure is counted into `user_api_key_error_executions` and silently subtracted from the alert — a real platform incident (rotated managed key) goes unpaged.

Direct question: how is the filter intended to distinguish managed vs user-supplied credentials given that `executionData.credentials` only carries `{id, title, provider, type}`? Should this instead resolve `id` against the credentials store (or filter via the `credentials_store` rather than the JSON blob) before deciding to exclude?

autogpt_platform/backend/backend/data/execution.py
```diff
@@ -1593,7 +1606,14 @@ async def get_block_error_stats(
     SELECT 
         n."agentBlockId" as block_id,
         COUNT(*) as total_executions,
-        SUM(CASE WHEN ne."executionStatus" = 'FAILED' THEN 1 ELSE 0 END) as failed_executions
+        SUM(CASE WHEN ne."executionStatus" = 'FAILED' THEN 1 ELSE 0 END) as failed_executions,
+        SUM(CASE
+            WHEN ne."executionStatus" = 'FAILED'
```
majdyz, line 1620: 🤖 🟠 **Should Fix**: Filter only matches blocks whose credentials field is literally named `credentials`.

The SQL keys off `executionData->>'credentials'`, but many blocks declare credentials under other field names — e.g. `anthropic_credentials` / `e2b_credentials` in `backend/blocks/claude_code.py:88,97`. The block schema's `get_credentials_fields()` returns whatever the block author named it.

Failure scenario:
1. ClaudeCodeBlock fails with `invalid_api_key` from the Anthropic SDK because a user supplied an expired key in `anthropic_credentials`.
2. `executionData->>'credentials'` is NULL (the field is `anthropic_credentials`).
3. The exclusion clause fails; the user error counts as a real platform failure and contributes to the alert — exactly the false positive this PR is trying to fix.

Direct question: should the JSON path traverse all keys whose value matches `CredentialsMetaInput` shape (e.g. `jsonb_path_exists` over any subobject with `id`+`provider`+`type`), or should we standardize on `credentials` as the only recognized field name?

autogpt_platform/backend/backend/data/execution.py
```diff
@@ -1593,7 +1606,14 @@ async def get_block_error_stats(
...
+                AND ne."executionData"->>'credentials' IS NOT NULL
+                AND ne."stats"->>'error' LIKE '%invalid_api_key%'
```
majdyz: 🤖 🟠 **Should Fix**: `LIKE '%invalid_api_key%'` is an unscoped substring match that can hide real bugs.

`invalid_api_key` is not a code we emit — it's a substring of error messages from third-party SDKs (OpenAI, Anthropic). The current pattern silently excludes anything containing that text regardless of source. Combined with the issues above (no `is_managed` discrimination, only matches `credentials` field name), here is a concrete failure scenario:

1. A backend regression in our LLM wrapper raises an internal exception whose `str()` happens to include the literal string `"got upstream error: invalid_api_key (rotated platform key)"` — e.g. from a wrapped `AuthenticationError` we caught and re-raised when our managed OpenRouter key was rotated.
2. Block has a credentials field literally named `credentials` (e.g. an LLM block using a managed creds default).
3. Per the previous comment, `is_managed` JSON key is missing → guard evaluates TRUE.
4. The failure is removed from `platform_error_rate`; on-call never gets paged for our own rotated platform key.

Direct question: can we narrow this to a specific structured signal (e.g. a typed exception class persisted into `stats.error_type`, or HTTP `401` + provider classification) instead of substring-matching free-text on a string that originates outside our codebase?

autogpt_platform/backend/backend/monitoring/test_block_error_monitor.py
```diff
@@ -0,0 +1,183 @@
...
+import pytest
+
+from backend.data.execution import BlockErrorStats
+from backend.monitoring.block_error_monitor import (
+    BlockErrorMonitor,
+    BlockStatsWithSamples,
+)
+
+
+class TestBlockErrorStats:
```
majdyz, line 13: 🤖 🟡 **Nice to Have**: Test coverage relies entirely on `BlockStatsWithSamples` Python aggregation — no test exercises the SQL classifier itself.

The new `test_block_error_monitor.py` only constructs `BlockStatsWithSamples` with pre-counted `user_api_key_error_executions`; it never validates that the SQL `CASE WHEN` clause in `backend/data/execution.py:1610-1615` actually classifies a fixture row correctly. That's precisely where the bugs flagged above live (missing `is_managed`, non-`credentials` field names, broad substring match).

Failure scenario:
1. Reviewer reads the green test suite and assumes the classification is correct.
2. A future refactor of the SQL (or stored shape of `executionData`) silently breaks classification — both the included and excluded paths.
3. No integration test catches it because nothing inserts real `AgentNodeExecution` rows and asserts on the returned `user_api_key_error_executions` count.

Direct question: can we add an integration-style test against `get_block_error_stats` with at least these fixture rows — (a) managed-credential failure with `invalid_api_key`, (b) user-credential failure with `invalid_api_key`, (c) non-`credentials`-named field failure with `invalid_api_key`, (d) generic FAILED row without `invalid_api_key` — and assert each is classified correctly?

autogpt_platform/backend/backend/monitoring/block_error_monitor.py
```diff
@@ -133,14 +150,23 @@ def _generate_critical_alerts(
...
+                stats.total_executions >= 10
+                and stats.platform_error_rate >= threshold * 100
+            ):
                 error_groups = self._group_similar_errors(stats.error_samples)
 
                 alert_msg = (
-                    f"🚨 Block '{block_name}' has {stats.error_rate:.1f}% error rate "
-                    f"({stats.failed_executions}/{stats.total_executions}) in the last 24 hours"
+                    f"🚨 Block '{block_name}' has {stats.platform_error_rate:.1f}% error rate "
+                    f"({stats.platform_failed_executions}/{stats.total_executions}) in the last 24 hours"
```
majdyz, line 160: 🤖 🟡 **Nice to Have**: "Top blocks" daily summary (`_generate_top_blocks_alert`, untouched in this PR) still ranks by raw `failed_executions` and prints `stats.error_rate`. So a block dominated by user-key failures will still lead the daily summary with a misleading 99% rate even though the critical-alert path now correctly ignores it.

Failure scenario:
1. Block A: 240 user-key failures, 0 platform failures — correctly excluded from critical alert by this PR.
2. Block B: 25 real platform failures (below critical threshold).
3. Daily summary still leads with "Block A: 240 errors (100%)" → on-call attention goes to the wrong block.

Direct question: should the top-blocks summary also use `platform_failed_executions` / `platform_error_rate` (mirroring the critical-alert formatting, optionally appending the excluded user-key count)?

autogpt_platform/backend/backend/monitoring/block_error_monitor.py
```diff
@@ -63,7 +76,10 @@ def check_block_error_rates(self) -> str:
             # For blocks with high error rates, fetch error samples
             threshold = self.config.block_error_rate_threshold
             for block_name, stats in block_stats.items():
-                if stats.total_executions >= 10 and stats.error_rate >= threshold * 100:
+                if (
+                    stats.total_executions >= 10
+                    and stats.platform_error_rate >= threshold * 100
```
majdyz, line 80: 🤖 🔵 **Nit**: `error_samples` displayed in the alert aren't filtered with the same classifier used for counting.

`_get_error_samples_for_block` pulls a few recent FAILED rows for the block without applying the `invalid_api_key` / `is_managed` predicate. Even when the alert legitimately fires on `platform_error_rate=8%`, the "📊 Error Types" section can still be dominated by `invalid_api_key` samples — confusing on-call ("why does this alert fire if every sample is a user key error?").

Direct question: should sample selection prefer rows that the count-side classifier treats as platform failures, so the displayed error types match the rate that triggered the alert?

majdyz:
@codeCraft-Ritik — pushed a follow-up rewrite of the error classifier:
- Replaced the substring/regex matching with a typed error classifier so user-supplied invalid API key errors are bucketed via structured types instead of string matching.
- Wired `platform_error_rate` + `platform_failed_executions` through the alert path with the typed classification.
- Threads resolved; ready for another look when you have time.

PR #13207, opened 2026-05-25
fix(platform/copilot): MCP setup card fires on stale creds, OAuth popup close race, registry-search prompt, Connected/Reconnect UI, integrations page polish

majdyz:
## Why

Five MCP-related bugs / UX gaps all surfaced as **"the MCP sign-in is broken for users"**. Reported by John (#breakage 2026-05-22) and reproduced by Zamil (session `c79e0ad0-bbbc-475f-815f-cb8d73dc4c9d`, 2026-05-25). The headline pattern: autopilot promises a sign-in card that never appears, or surfaces one whose "Connect" button fails with **"Sign-in window was closed"**, or shows the user nothing at all when they're already connected.

The five root causes:

1. **Stale-creds shadow the setup card.** `run_mcp_tool.py:192-195` only fired `SetupRequirementsResponse` on 401/403 when `not creds`. If the user already had an MCP credential row whose token was revoked/expired *server-side* (without our local `access_token_expires_at` knowing), `auto_lookup_mcp_credential` skipped refresh, fed the dead token to the server, got 401, and we fell through to a generic HTTP error — no sign-in card ever shown. Matches John's screenshot: *"the system can't trigger the sign in until after I delete the expired credential"*. **Verified fixed E2E** in the test comment below.

2. **OAuth popup close race.** `oauth-popup.ts:184-197` polled `popup.closed` every 500 ms and rejected with `OAUTH_ERROR_WINDOW_CLOSED` the *instant* the popup closed. The Sentry MCP callback page writes BroadcastChannel + localStorage **then** calls `window.close()`; across origins those events can land a few hundred ms after close, so a successful OAuth handshake gets reported as "Sign-in window was closed".

3. **Model never tried MCP for services not in the hardcoded list.** Prompt told the model `find_block` is mandatory and listed MCP only as "check if a hosted MCP server is available… only use known URLs". For services like Sentry that aren't in `mcp_tool_guide.md`'s table, after `find_block` returned no_results the model jumped straight to `SendAuthenticatedWebRequestBlock` — never read the guide, never searched the MCP registry.

4. **No "already connected" UX.** When the user IS already signed in to an MCP server, the model would say *"a sign-in card has appeared"* but no card rendered (discovery success was invisible). User reported: *"there are instances that says we are already logged in and button should appear and turns out because we already logged in nothing is rendered making it misleading"*.

5. **Integrations page UI for MCP** — group label rendered as "Mcp" (title-caser stumble on acronym), no logo, per-row title duplicated `MCP: ...` prefix that the group label already showed.

### Backend
- `run_mcp_tool.py`:
  - On 401/403, **always** fire the setup card (regardless of whether creds exist), and call the new `invalidate_mcp_credential` helper to delete the stale row so the next attempt doesn't loop on the same dead token.
  - New `surface_connect_card: bool = False` parameter — when true, skip the MCPClient call entirely and return a `SetupRequirementsResponse` whose `user_readiness.has_all_credentials` reflects current stored state. Lets the model explicitly surface a "you're already connected" card for "connect to X" intent.
  - `_build_setup_requirements` accepts a `connected` flag to populate `UserReadiness` correctly (`has_all_credentials`, `ready_to_run`, `missing_credentials`).
- `blocks/mcp/helpers.py`: new `invalidate_mcp_credential(user_id, credential_id)` via `IntegrationCredentialsManager.delete` (not `store.delete_creds_by_id`) so the per-credential lock + `_invoke_creds_changed_hook` fire (evicts cached provider token). Top-level imports hoisted per AGENTS.md.
- `prompting.py`: strengthened the "Tool Discovery Priority" section. After `find_block` returns nothing, the model MUST load `mcp_tool_guide` and search the registry (`https://registry.modelcontextprotocol.io/v0/servers?q=<service>`) via `SendWebRequestBlock` (public, no creds) before pivoting to REST. Added hostname-vendor-match safety check before any `run_mcp_tool` call on a registry-returned URL. Added a hint for `surface_connect_card=true` on "connect" intent.

### Frontend
- `oauth-popup.ts`: 1.5 s → **3 s** grace window after `popup.closed` flips true, with a **final synchronous localStorage sweep** on close. Covers the case where BroadcastChannel never delivers due to storage-partitioning / BCG isolation, and the poll tick hasn't run yet. AbortController cleanup tears down the grace timer on completion. Inline comments document the timeout hierarchy (5-min outer, 500ms close poll, 3s close grace) so the next reader doesn't mistake them.
- `MCPSetupCard.tsx`:
  - New **"Connected to {service}. Reconnect"** render branch when the backend reports `has_all_credentials=true`. Reconnect runs the same OAuth dance — lets users swap accounts without deleting first.
  - Reset `showManualToken` on retry so a stale 400-fallback doesn't shadow a later non-400 error.
  - Drop the connected view on Reconnect failure so the error / manual-token input render correctly.
- `settings/integrations/helpers.ts`:
  - `PROVIDER_DISPLAY_NAME_OVERRIDES["mcp"] = "MCP"` — keeps the acronym uppercase.
  - New `stripProviderPrefix` helper (generic, case-insensitive). Any provider whose backend-set title starts with `<DisplayName>: ` gets the prefix stripped so the row label doesn't double up the group label. Works for MCP today, applies to any future provider with the same backend convention.
- `public/integrations/mcp.png`: official MCP "linked tape" logo (MIT-licensed, sourced from Wikimedia Commons), content-bbox cropped + 10% padding → 256×256, visually centered for the 24×24 avatar circle.

### Tests
- `run_mcp_tool` tests: 3 new (stale-cred fires setup card + invalidate, surface_connect_card connected/disconnected). All 32 tests pass.
- `helpers.test.ts`: 16 tests pass — added `mcp: "MCP"` override coverage + prefix-strip generic test (mcp + github + non-prefixed examples).
- `oauth-popup.test.ts` (new): 5 tests covering the popup grace logic + abort cleanup.

## How

The stale-cred path is the load-bearing fix for John's bug: dropping the `not creds` clause is one line, but pairing it with invalidation prevents the user re-attempting the same call and looping on the dead token. The OAuth grace window + final localStorage sweep is the smallest possible fix for the Sentry popup-close race — no protocol changes, no extra postMessage round-trips. The prompt change is complementary to PR #13117 (still open) but tighter; if #13117 lands first I'll rebase. The Connected/Reconnect UI is a backend-feature-flag-style toggle: model can now request the card on demand without a discovery round-trip. Integrations page UI is purely cosmetic + display-side — no migration, no data change.

## Test plan

- [x] `pytest backend/copilot/tools/test_run_mcp_tool.py` — 32 passed (3 new)
- [x] `pnpm vitest run helpers.test.ts oauth-popup.test.ts` — 21 passed
- [x] `poetry run black + isort + ruff` clean
- [x] `pnpm types && pnpm lint && pnpm format` clean
- [x] **`/pr-test` local — full Sentry MCP flow end-to-end:** setup card → OAuth popup → Connected/Reconnect → auto-retry → tool execution. **See the test report comment below for screenshots + scenario breakdown.**
- [x] **John's bug E2E:** inserted OAuth-shape stale cred → triggered run_mcp_tool → setup card surfaced + stale cred auto-deleted (`/api/integrations/credentials` returned `[]` post-call).
- [ ] Re-verify on dev preview after merge (live Sentry against `dev-builder.agpt.co`).

## Related

- Discord report: https://discord.com/channels/1126875755960336515/1285234928367308821/1507504078278754414
- Bug session (langfuse trace `3c4b1e7832dbe8e5f2fdcca870496f47`): https://dev-builder.agpt.co/copilot?sessionId=c79e0ad0-bbbc-475f-815f-cb8d73dc4c9d
- Adjacent / complementary: #13117 (open) — auto-discover MCP integrations prompt


autogpt_platform/backend/backend/copilot/prompting.py
```diff
@@ -147,12 +147,19 @@
...
-
-3. **`SendAuthenticatedWebRequestBlock`** — If no block or MCP server exists,
-   use `SendAuthenticatedWebRequestBlock` with existing host-scoped
+2. **`run_mcp_tool` — MANDATORY when `find_block` returns nothing.** Don't
+   stop at "no integration exists" after one `find_block` miss. Load the
+   MCP guide (`read_skill(name="mcp_tool_guide")`, or `get_mcp_guide` if
+   `read_skill` is unavailable) once per session; if the service is in
+   the known list use that URL, otherwise query
+   `https://registry.modelcontextprotocol.io/v0/servers?q=<service>` via
+   `SendAuthenticatedWebRequestBlock` (no creds needed) and use the first
```
majdyz: 🤖 🟠 **Should Fix**: `SendAuthenticatedWebRequestBlock` requires a `credentials: HttpCredentials` field (see `backend/blocks/http.py:226-229`) — instructing the model to call it for an unauthenticated registry lookup will fail input validation. The MCP registry (`https://registry.modelcontextprotocol.io/v0/servers`) is public and needs no auth, so either point the model at the unauthenticated `SendWebRequestBlock` (the parent class) or, better, reference the registry-lookup path the model already has via `find_block(query="send web request")` and let it pick. As written the model will hit a validation error and fall back to the no-MCP refusal that this PR is trying to prevent.

autogpt_platform/backend/backend/blocks/mcp/helpers.py
```diff
@@ -71,6 +72,31 @@ def parse_mcp_content(content: list[dict[str, Any]]) -> Any:
...
+    ``access_token_expires_at`` knowing.  Removing the dead row prevents
+    ``auto_lookup_mcp_credential`` from feeding the same stale token back
+    on the next attempt and lets the user re-auth cleanly via the setup
+    card.  Failures are swallowed (best-effort) — the worst case is a
+    second loop through the same code path, which still surfaces the
+    setup card.
+    """
+    try:
+        mgr = IntegrationCredentialsManager()
+        await mgr.store.delete_creds_by_id(user_id, credential_id)
```
majdyz: 🤖 🟡 **Nice to Have**: This bypasses `IntegrationCredentialsManager.delete()` (creds_manager.py:265) and goes straight to `store.delete_creds_by_id`, skipping (a) the per-credential `_locked()` async context that prevents TOCTOU with a concurrent refresh, and (b) the `_invoke_creds_changed_hook(user_id, provider)` notification that evicts the provider-level `_token_cache` / `_null_cache` in `copilot/integration_creds.py`. Today nothing calls `get_provider_token(user_id, "mcp")` so (b) is theoretical, but the contract documented on `register_creds_changed_hook` says every create/update/delete fires the hook — quietly violating that here makes future grep audits harder. The existing `routes.py:374` precedent has the same gap, but here it's load-bearing for the recovery flow; going through `mgr.delete(user_id, creds.id)` (catching `ValueError` for system/managed creds) would keep the contract intact at one extra await.

autogpt_platform/backend/backend/copilot/prompting.py
majdyz: 🤖 Fixed in [b66f87508](https://github.com/Significant-Gravitas/AutoGPT/commit/b66f875080bb8410d9a4001369e4c0ccf309a2ca): switched the registry-query reference to the unauthenticated `SendWebRequestBlock` (parent class), which doesn't require credentials.

autogpt_platform/backend/backend/copilot/prompting.py
```diff
@@ -147,12 +147,19 @@
...
+   `remotes[].url` hit. Many popular services (Sentry, etc.) aren't in
+   the hardcoded list — always check the registry before falling back.
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [b66f87508](https://github.com/Significant-Gravitas/AutoGPT/commit/b66f875080bb8410d9a4001369e4c0ccf309a2ca): added a hostname verification step before `run_mcp_tool`. The prompt now requires the model to verify the registry-returned hostname matches the service (e.g. `mcp.sentry.dev` for Sentry, vendor-owned domains) and, when ambiguous or multiple plausible results exist, surface the candidates to the user and ask which one to use — never auto-pick.

autogpt_platform/backend/backend/blocks/mcp/helpers.py
majdyz: 🤖 Fixed in [b66f87508](https://github.com/Significant-Gravitas/AutoGPT/commit/b66f875080bb8410d9a4001369e4c0ccf309a2ca): switched `invalidate_mcp_credential` to use `IntegrationCredentialsManager.delete()` so the per-credential lock acquires + `_invoke_creds_changed_hook` fires.

autogpt_platform/backend/backend/copilot/tools/run_mcp_tool.py
```diff
@@ -329,16 +370,21 @@ def _build_setup_requirements(
...
                 agent_id=server_url,
                 agent_name=service,
                 user_readiness=UserReadiness(
-                    has_all_credentials=False,
-                    missing_credentials=missing_creds_dict,
-                    ready_to_run=False,
+                    has_all_credentials=connected,
+                    missing_credentials={} if connected else missing_creds_dict,
+                    ready_to_run=connected,
                 ),
```
majdyz, line 437: 🤖 🔵 **Nit**: When `connected=True`, `user_readiness.missing_credentials={}` (empty) but `requirements["credentials"]` is still populated with `missing_creds_list`. The two fields disagree about whether creds are needed. Today the frontend gates on `user_readiness.has_all_credentials` (MCPSetupCard.tsx:46-48) so it doesn't matter, but a future consumer reading `requirements["credentials"]` would render a credential picker we don't actually need. Either clear `requirements["credentials"]` when `connected=True`, or document the divergence so the next reader doesn't have to figure it out.

majdyz: 🤖 Fixed in [d269e0ee2](https://github.com/Significant-Gravitas/AutoGPT/commit/d269e0ee2f233a7253d12ed25f5dacf1a7e5627d): `requirements.credentials` is now `[]` when `connected=True`, matching `user_readiness.missing_credentials={}`.

autogpt_platform/frontend/src/app/(platform)/copilot/tools/RunMCPTool/components/MCPSetupCard/MCPSetupCard.tsx
```diff
@@ -145,10 +154,30 @@ export function MCPSetupCard({ output, retryInstruction }: Props) {
...
+          </button>
+        </div>
+        {error && (
+          <div className="rounded-lg border border-red-200 bg-red-50 px-3 py-2 text-sm text-red-700">
+            {error}
+          </div>
+        )}
       </div>
     );
   }
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [518ba89](https://github.com/Significant-Gravitas/AutoGPT/commit/518ba8991499c1290f906c1fe54aaab0a2fc16c4): `handleConnect` now sets `setConnected(false)` in the catch before the 400-branch toggles `setShowManualToken(true)`, so the not-connected branch (which contains the manual-token input + error banner) actually renders. Regression test added in [3e84775](https://github.com/Significant-Gravitas/AutoGPT/commit/3e84775ce5).

autogpt_platform/frontend/src/app/(platform)/copilot/tools/RunMCPTool/components/MCPSetupCard/MCPSetupCard.tsx
```diff
@@ -97,6 +106,9 @@ export function MCPSetupCard({ output, retryInstruction }: Props) {
       onSend(retryInstruction ?? "I've connected. Please retry.");
     } catch (e: unknown) {
       const err = e as Record<string, unknown>;
+      // Reconnect failures must drop the Connected view so the user sees
+      // the error / manual-token input rendered by the not-connected branch.
+      setConnected(false);
       if (err?.status === 400) {
         setShowManualToken(true);
         setError(
```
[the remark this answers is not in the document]
majdyz: Fixed in [f271012](https://github.com/Significant-Gravitas/AutoGPT/commit/f271012e41) — `handleConnect()` now calls `setShowManualToken(false)` at the start of every attempt, so a prior 400 cannot leave the manual-token input stuck open on a later non-400 failure.

majdyz:
## 🧪 E2E Test Report

**Branch:** `fix/copilot-mcp-stale-creds-and-sentry` (HEAD: `0913e2e55a`)
**Setup:** native stack (poetry run app + pnpm dev) on `tester@gmail.com` (MAX tier, paywall flag off, OpenRouter routing via `CHAT_API_KEY`)

### Scenario 1 — MCP integrations page UI (settings/integrations)

After connecting Sentry MCP, the integrations page renders with the new logo + capitalization + per-row prefix strip:

- ✅ Group label is **"MCP"** (not "Mcp") — backed by `PROVIDER_DISPLAY_NAME_OVERRIDES["mcp"] = "MCP"`
- ✅ Official MCP "linked tape" logo (centered after content-bbox crop + 10% padding)
- ✅ Row shows **"mcp.sentry.dev"** — the redundant `MCP:` prefix is stripped by generic `stripProviderPrefix` helper (works for any provider whose backend-set title doubles up with the group name)
- ✅ OAuth badge + canonical URL displayed below

integrations page

### Scenario 2 — John's bug repro: stale cred → setup card doesn't fire (BEFORE behavior)

Inserted a deliberately-invalid bearer-token cred for `mcp.sentry.dev`, asked copilot to use Sentry. Setup card fires (because `not creds` guard was dropped):

bearer-stale cred bug repro

### Scenario 3 — John's exact bug: OAuth-shape stale cred → setup card fires + stale cred auto-deleted

Inserted a proper OAuth-style stale cred (full metadata, future `access_token_expires_at`, revoked-style `access_token`) — exactly John's screenshot scenario. Asked copilot for Sentry issues:

- ✅ Setup card surfaced with **"Connect sentry.dev"** button
- ✅ Model's prose: *"Your Sentry connection needs to be refreshed — please use the sign-in button that just appeared in chat to reconnect"*
- ✅ Stale cred row **auto-deleted** from DB after the 401 (`GET /api/integrations/credentials` returned `[]` post-call) — `invalidate_mcp_credential` via `mgr.delete` fires the per-credential lock + cache eviction hook

john bug fix

### Scenario 4 — OAuth popup close race (no screenshot, validated earlier this session)

Already verified live: completed Sentry OAuth → popup self-closed → no "Sign-in window was closed" false negative → "Connected to sentry.dev — Reconnect" pill rendered → auto-retry message fired → run_mcp_tool succeeded on second pass. (3s grace + final localStorage sweep on close in `oauth-popup.ts`.)

## Summary

| Scenario | Result |
|---|---|
| MCP capitalization + logo + prefix strip on integrations page | ✅ PASS |
| Bearer-token stale-cred 401 → setup card fires | ✅ PASS |
| **John's bug — OAuth stale-cred 401 → setup card fires + cred auto-deleted** | ✅ PASS |
| OAuth popup close race (verified earlier session) | ✅ PASS |
| Connected/Reconnect UI render branch (verified earlier session) | ✅ PASS |

All 5 scenarios pass. Headline bug from #breakage report (John 2026-05-22) confirmed fixed end-to-end.

autogpt_platform/frontend/src/app/(platform)/copilot/tools/RunMCPTool/components/MCPSetupCard/MCPSetupCard.tsx
```diff
@@ -39,18 +47,60 @@ export function MCPSetupCard({ output, retryInstruction }: Props) {
...
   const [manualToken, setManualToken] = useState("");
-  const [connected, setConnected] = useState(false);
+  const [localConnected, setLocalConnected] = useState(initiallyConnected);
   const oauthAbortRef = useRef<(() => void) | null>(null);
 
+  // Combined view: live API state OR local just-connected state.
+  const connected = localConnected || liveHasCred;
+  // Setter compatible with the existing call-sites — they only ever set
+  // ``true`` after a successful flow or ``false`` to drop the pill.
+  const setConnected = setLocalConnected;
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [2fc763261](https://github.com/Significant-Gravitas/AutoGPT/commit/2fc763261d): added a `forceDisconnected` state that the catch block flips to `true` on any Reconnect failure, overriding `liveHasCred`. The not-connected branch (error banner + manual-token input on 400) now renders correctly even when the stored cred row still exists. Reset on the next attempt + on successful manual-token store. Regression test added: `drops Connected state on Reconnect failure even when live creds still report the server as connected`. All 7 MCPSetupCard tests pass.

autogpt_platform/frontend/src/app/(platform)/copilot/tools/RunMCPTool/components/MCPSetupCard/MCPSetupCard.tsx
```diff
@@ -39,18 +47,70 @@ export function MCPSetupCard({ output, retryInstruction }: Props) {
...
+    },
+  });
+  const liveHasCred = Array.isArray(liveCredsRes)
+    ? liveCredsRes.some(
+        (c) =>
+          c.provider === "mcp" &&
+          typeof c.host === "string" &&
+          normalizeMcpUrl(c.host) === normalizedServer,
+      )
+    : false;
```
majdyz: 🤖 🟠 **Should Fix**: When the live cred-list API fails (e.g. 401 from `/api/integrations/credentials` during a token-refresh race, or a 5xx from the backend), the `select` callback returns `null` → `Array.isArray(liveCredsRes)` is `false` → `liveHasCred=false`. Combined with `initiallyConnected=false` (the common stale-cred scenario this PR specifically targets — backend emitted the card while the cred was missing), the user sees the bare **Connect** button even though the cred exists.

Failure scenario:
1. User reloads chat; an old SetupCard with `has_all_credentials=false` is rehydrated → `initiallyConnected=false`.
2. The user actually completed sign-in in a prior session — there IS a cred row server-side.
3. `useGetV1ListCredentials` returns HTTP 401 (e.g. supabase token mid-refresh) → `liveCredsRes=null`.
4. Card renders the not-connected branch. User clicks Connect, opens OAuth popup, signs in **again** — duplicate cred row inserted.

Suggest: when `liveCredsRes` is neither `null` nor an array (`isPending`/`isError` state), treat it as "unknown" and lean on `initiallyConnected` (current behavior). When it IS `null` specifically due to an error, surface a soft warning OR retry — currently the error is silently swallowed. A test for `setMockLiveCreds(null)` would lock this behavior in.

autogpt_platform/frontend/src/app/(platform)/copilot/tools/RunMCPTool/components/MCPSetupCard/MCPSetupCard.tsx
```diff
@@ -132,6 +200,9 @@ export function MCPSetupCard({ output, retryInstruction }: Props) {
       });
       if (!(res.status >= 200 && res.status < 300))
         throw new Error("Failed to store token");
+      // Clear the force-disconnected override set by an earlier OAuth
+      // failure so the new manual-token cred surfaces the Connected pill.
+      setForceDisconnected(false);
```
majdyz, line 250: 🤖 🟡 **Nice to Have**: `handleManualToken` does not reset `forceDisconnected` until the bearer-token API call **succeeds** (line 205). If OAuth Reconnect just failed → `forceDisconnected=true`, then the user types a token and submission fails too (network error, 422, etc.), the error message renders fine — but `forceDisconnected` stays `true` on subsequent retries even though `liveHasCred=true` is the truth.

Failure scenario:
1. User clicks Reconnect → OAuth 400 → `forceDisconnected=true`, manual-token input appears.
2. User pastes a bad token → POST `/api/v2/mcp/token` returns 422 (validation failure).
3. User clicks "Use Token" again with a valid token → `forceDisconnected` is still `true`, but it doesn’t matter because success path explicitly clears it.

Actually the success path handles it (line 205) — so this is purely a code-shape nit: clearing `forceDisconnected` at the **top** of `handleManualToken` (mirror of `handleConnect`’s line 113) makes the two flows symmetric and prevents future drift if someone adds an early-return between the two clears.

autogpt_platform/frontend/src/app/(platform)/copilot/tools/RunMCPTool/components/MCPSetupCard/MCPSetupCard.tsx
```diff
@@ -97,6 +157,14 @@ export function MCPSetupCard({ output, retryInstruction }: Props) {
...
     } catch (e: unknown) {
       const err = e as Record<string, unknown>;
+      // Reconnect failures must drop the Connected view so the user sees
+      // the error / manual-token input rendered by the not-connected
+      // branch.  Setting ``localConnected=false`` alone isn't enough when
+      // a stored cred still exists (``liveHasCred=true``) — flip
+      // ``forceDisconnected`` so the not-connected branch wins until the
+      // user retries.
+      setConnected(false);
+      setForceDisconnected(true);
```
majdyz, line 205: 🤖 🟠 **Should Fix**: Concurrent Reconnect double-click race. `oauthAbortRef.current?.()` runs at line 115 — **after** `setForceDisconnected(false)` (line 113). If a user double-clicks Reconnect rapidly:

Failure scenario:
1. First click starts `handleConnect()`. State updates queued: `setForceDisconnected(false)`. `loading=true`. Popup opens.
2. Second click fires before React commits the first batch. Inside the second `handleConnect`:
   - `setForceDisconnected(false)` again (idempotent).
   - `oauthAbortRef.current?.()` aborts the first popup → its `await promise` rejects with `OAUTH_ERROR_FLOW_CANCELED`.
3. First flow’s catch block runs: `setConnected(false); setForceDisconnected(true)`. Now `forceDisconnected=true`.
4. Second flow proceeds, popup completes successfully → `setConnected(true)` — but `forceDisconnected` is still `true` from step 3, so the card stays on the not-connected branch despite a successful sign-in.

The first flow’s catch should treat `OAUTH_ERROR_FLOW_CANCELED` as a no-op (don’t set `forceDisconnected`), since cancellation by a second invocation isn’t a real failure.

```ts
} catch (e: unknown) {
  const err = e as Record<string, unknown>;
  // Aborted by a second handleConnect() — let the new flow drive state.
  if (typeof err?.message === "string" && err.message === OAUTH_ERROR_FLOW_CANCELED) {
    return;
  }
  setConnected(false);
  setForceDisconnected(true);
  ...
}
```

autogpt_platform/frontend/src/lib/oauth-popup.ts
```diff
@@ -181,14 +181,68 @@ export function openOAuthPopup(
...
+    // On detected close we ALSO do one synchronous final localStorage
+    // read — covers the case where the BroadcastChannel listener never
+    // fired (storage-partitioning / BCG isolation) and the poll tick
+    // hasn't run yet.  The grace then handles any remaining
+    // post-message latency.
+    //
+    // The setTimeout lives in a plain JS closure, not React state — it
+    // does NOT stack across re-renders, and the AbortController cleanup
+    // tears it down if the caller aborts before grace expires.
     if (popup) {
```
majdyz: 🤖 🟡 **Nice to Have**: When `openOAuthPopup` falls back to opening the login URL in a new tab (popup-blocked → `popupBlocked=true`), the `popup` variable is the **blocked** popup reference (`null`/`closed`), not the new-tab handle. Line 216-251 then guards `if (popup)` — but `popup` here is the original `window.open("about:blank", ...)` return value, which can be non-null but already-closed when the browser blocked navigation.

Failure scenario:
1. Browser blocks the popup; `window.open("about:blank", ...)` returns a popup object that synchronously becomes `closed=true`.
2. We fall back to `window.open(loginUrl, "_blank")` (new tab).
3. `if (popup)` evaluates truthy → close-poll interval starts.
4. First tick: `popup.closed === true` immediately → final-storage check (no result yet) → 3 s grace → `OAUTH_ERROR_WINDOW_CLOSED`.
5. User completes sign-in in the new tab seconds later — but the parent already rejected with "Sign-in window was closed".

The check at line 216 should be `if (popup && !popupBlocked)` — when we fell back to a new tab there is no close signal to poll for (new tabs aren’t inspectable cross-origin), and the outer 5-min timeout already bounds the wait.

autogpt_platform/frontend/src/lib/oauth-popup.ts
```diff
@@ -181,14 +181,68 @@ export function openOAuthPopup(
...
+    //      observes a closed popup; gives in-flight result messages a
+    //      chance to land before we declare failure.
+    //
+    // Why the grace at all?  The callback page does:
+    //   bc.postMessage(...); localStorage.setItem(...); setTimeout(close, 1500)
+    // BroadcastChannel delivery is async across-origin; the parent's
+    // localStorage poll fires every 500 ms.  Without a grace window the
+    // ``popup.closed`` rejection can win the race against a successful
+    // result that's a few hundred ms behind, surfacing the bogus
+    // "Sign-in window was closed" error John screenshotted.
```
majdyz, line 207: 🤖 🔵 **Nit**: The "Why the grace at all?" comment block (lines 199-205) calls out `bc.postMessage(...); localStorage.setItem(...); setTimeout(close, 1500)` as the callback page’s behavior, but this code lives in `oauth-popup.ts` — a reader looking for that 1500 ms `setTimeout` won’t find it in this file. Worth a path hint like `(see frontend/src/app/(platform)/oauth/mcp/callback/page.tsx)` so the next debugger can verify the contract end-to-end without grepping.

autogpt_platform/backend/backend/copilot/tools/run_mcp_tool.py
```diff
@@ -175,6 +187,18 @@ async def _execute(
...
+        # setup card that reflects current connection state.  Stops the
+        # model from running a no-op discovery just to surface a card and
+        # gives users a visible Reconnect affordance even when they're
+        # already signed in (previously: discovery succeeded silently and
+        # nothing rendered, so the model's "click Connect" promise
+        # mismatched the empty UI).
+        if surface_connect_card:
+            return self._build_setup_requirements(
+                server_url, session_id, connected=creds is not None
+            )
```
majdyz, line 249: 🤖 🟠 **Should Fix**: `surface_connect_card=true` skips the `MCPClient` call entirely — fast and matches the docstring intent. But it does NOT validate that the URL is actually an MCP server before promising the user a Connect card. If the model passes any URL that survives `validate_url_host` (e.g. a typo `https://centry.com/mcp` instead of `https://mcp.sentry.dev/mcp`, or a wrong path `https://mcp.sentry.dev/api`), the card surfaces "Connect centry.com" and the user proceeds to OAuth against a non-existent endpoint.

Failure scenario:
1. User: "connect to Sentry".
2. Model: registry query returns ambiguous results (e.g. one outdated entry pointing to `mcp.sentryapp.io`).
3. Model picks the wrong URL, calls `run_mcp_tool(surface_connect_card=true)`.
4. Backend skips `MCPClient.initialize()` (the only thing that would surface a 404/HTML response) → returns a SetupCard for an arbitrary URL.
5. User clicks Connect → OAuth `/oauth/login` POST against that URL → backend tries dynamic client registration → fails with a confusing error.

The "regular" path (`surface_connect_card=false`) catches this because `MCPClient.initialize()` either succeeds (validating the endpoint speaks MCP) or returns 401/403 (still legit MCP) or an `MCPClientError` (clean error message). The surface_connect_card path bypasses that safety net.

Suggest one of:
- A lightweight HEAD/OPTIONS preflight to confirm the URL responds as an MCP endpoint before emitting the card, OR
- Document in the parameter description that `surface_connect_card` defers all URL validation to the OAuth login step, so the model MUST do hostname-vendor-match first (the prompt already says this, but a backend-side reminder closes the loop).

0ubbe (dismissed):

autogpt_platform/backend/backend/blocks/mcp/helpers.py
```diff
@@ -71,6 +72,34 @@ def parse_mcp_content(content: list[dict[str, Any]]) -> Any:
...
+        # per-credential lock + ``_invoke_creds_changed_hook`` fire — the hook
+        # evicts any cached provider token for the user.
+        await mgr.delete(user_id, credential_id)
+        logger.info("Invalidated stale MCP credential %s", credential_id)
+    except Exception:
+        logger.warning(
+            "Failed to invalidate stale MCP credential %s",
+            credential_id,
+            exc_info=True,
+        )
```
majdyz, line 107: 🤖 🔵 **Nit**: `invalidate_mcp_credential` swallows all `Exception` subtypes by design (best-effort), but this also masks legitimate "this user does not own this credential" failures from `IntegrationCredentialsManager.delete` — silently. Worst case: a stale `credential_id` (e.g. from a stale ChatSession after the user manually deleted the cred in Settings) → delete throws "not found" → swallowed → next 401 loops back here → another swallowed delete → forever.

Not catastrophic (the setup card still surfaces and the user can re-auth manually), but the loop wastes 1 round-trip per retry. Consider catching the "not found" path specifically and `return`ing early without the warning-level log — the current `logger.warning(... exc_info=True)` will spam Sentry on a benign condition.

autogpt_platform/frontend/src/app/(platform)/copilot/tools/RunMCPTool/components/MCPSetupCard/__tests__/MCPSetupCard.test.tsx
```diff
@@ -130,4 +178,59 @@ describe("MCPSetupCard", () => {
...
+    render(<MCPSetupCard output={makeSetupOutput()} />);
+    // Live creds say connected → starts in Connected state.
+    const reconnectBtn = screen.getByRole("button", { name: /reconnect/i });
+    fireEvent.click(reconnectBtn);
+
+    await waitFor(() => {
+      expect(screen.getByPlaceholderText("Paste API token")).toBeDefined();
+    });
+    expect(screen.queryByText(/connected to example\.com/i)).toBeNull();
+  });
```
majdyz, line 267: 🤖 🟡 **Nice to Have**: Test coverage gap — there is no test for the "live cred-list API fails" path (`useGetV1ListCredentials` returning `null` via the `select` on a 401/5xx response). All current tests pass `mockLiveCreds` as a populated array. Failure scenario worth locking in:

1. `setMockLiveCreds(null)` (or override the mock to simulate an error state)
2. Render `<MCPSetupCard output={makeSetupOutput(undefined, true)} />` (persisted snapshot says connected)
3. Expect: Connected/Reconnect renders because `initiallyConnected=true` carries over even when live cred fetch fails.

The reverse case is also worth a test:
1. `setMockLiveCreds(null)`
2. Persisted snapshot says `has_all_credentials=false`
3. Cred actually exists server-side, but API failed
4. Expect: Connect button shows (current behavior, intentional — but lock it in so a future refactor doesn’t accidentally promote a stale "true" through the `null` path).

autogpt_platform/backend/backend/copilot/tools/test_run_mcp_tool.py
```diff
@@ -544,15 +544,96 @@ async def test_auth_required_without_creds_returns_setup_requirements():
...
+                    session=session,
+                    server_url=_SERVER_URL,
+                    surface_connect_card=True,
+                )
+                mock_client_cls.assert_not_called()
+
+    assert isinstance(response, SetupRequirementsResponse)
+    assert response.setup_info.user_readiness.has_all_credentials is True
+    assert response.setup_info.user_readiness.ready_to_run is True
+    assert response.setup_info.user_readiness.missing_credentials == {}
```
majdyz, line 592: 🤖 🟠 **Should Fix**: This test asserts `has_all_credentials=True` whenever `auto_lookup_mcp_credential` returns a credential — but the credential might be **stale** (the very scenario this PR exists to address: server-side revocation without local `access_token_expires_at` knowing). The user would see "Connected to sentry.dev" with a Reconnect button, click anything that triggers tool execution → 401 → setup card again. Confusing UX cycle.

Failure scenario (real, matches John’s original bug):
1. User has a Sentry MCP cred whose token was revoked server-side last week.
2. User says: "connect to Sentry".
3. Model calls `run_mcp_tool(surface_connect_card=true)`.
4. `auto_lookup_mcp_credential` returns the stale cred (its local `expires_at` is in the future).
5. Backend emits SetupCard with `has_all_credentials=True` → "✅ Connected to sentry.dev — Reconnect".
6. User: "now create an issue".
7. Model calls `run_mcp_tool(server_url, tool_name="create_issue")` → 401 → invalidate stale cred → new SetupCard → "❌ Connect Sentry".
8. User UX: "wait, I just connected, why is it asking again?"

Two options:
- Backend-side: in the `surface_connect_card=true` path, do a cheap probe (e.g. `client.initialize()` with the stored token; on 401/403, invalidate the cred and emit "not connected" instead of "connected"). Slower but accurate.
- Documentation-side: explicitly accept the discrepancy in the docstring and rely on the next user action to surface 401 → invalidate path.

If keeping the current behavior, add a test that **does NOT mock `MCPClient`**, demonstrating the optimistic-connected promise is acceptable.

autogpt_platform/frontend/src/lib/oauth-popup.ts
```diff
@@ -181,14 +181,68 @@ export function openOAuthPopup(
...
+            finalLocalStorageCheck();
+            if (handled) return;
+            handled = true;
+            reject(new Error(OAUTH_ERROR_WINDOW_CLOSED));
+            controller.abort("popup_closed");
+          }, POPUP_CLOSE_GRACE_MS);
+          controller.signal.addEventListener("abort", () =>
+            clearTimeout(graceTimeout),
+          );
         }
```
majdyz, line 271: 🤖 🔵 **Nit**: The grace-window test trio is comprehensive for the popup-present happy path, but there is no test for the `popup === null` branch (browser blocks popup synchronously → `window.open` returns `null` → no close-poll interval scheduled at all). 

Failure scenario worth locking in:
1. `setupPopup(null)` — `window.open` returns `null`.
2. Open the OAuth popup → `popupBlocked=true` is returned to the caller.
3. Advance time well past `POPUP_CLOSE_GRACE_MS` (3 s) and `500ms` poll interval — promise must NOT reject with `WINDOW_CLOSED`; it should pend until the timeout or a result lands.
4. Inject a successful result via `localStorage.setItem(...)` and advance time to confirm the 500 ms poll resolves it.

This would catch any future refactor that accidentally moves the close-poll setInterval outside the `if (popup)` guard.

majdyz (commented):
🤖 Manual deep-dive review (round 1 of /pr-polish), focused on the orchestrator-specified areas:

**Architecture & invariants**
- The `forceDisconnected` state machine in `MCPSetupCard.tsx` correctly fixes the Sentry-flagged Reconnect-failure-with-live-cred bug; explicit `forceDisconnected ⊕ (localConnected || liveHasCred)` is the right shape and the comment block walks the reader through the precedence.
- Backend `surface_connect_card=True` returning `has_all_credentials=True` purely from stored-cred presence (no network probe) is a reasonable trade-off for latency, but it can promise "Connected" for a credential the server has already revoked. See inline on test file line 582.
- `IntegrationCredentialsManager.delete` (vs `store.delete_creds_by_id`) is the right call — the comment in `helpers.py:91` correctly explains the hook eviction reasoning.

**Race conditions** — three potential races identified:
- Concurrent Reconnect double-click → first flow’s cancellation flips `forceDisconnected=true`, overriding the second flow’s success (see inline at MCPSetupCard.tsx:167).
- `popup.location.href` cross-origin assignment can throw `SecurityError` outside any try/catch (see inline at oauth-popup.ts:95 — could not anchor, see comment chain on line 205).
- `useGetV1ListCredentials` returning `null` via select (API failure) silently fails to "not connected" — see inline at MCPSetupCard.tsx:81.

**Test coverage gaps**
- No test for `popup === null` branch in oauth-popup.ts.
- No test for `liveCredsRes` being `null` (API failure) in MCPSetupCard.
- No test that verifies the stale-cred lie-about-connected case in `surface_connect_card`.

**Quality / nits**
- `invalidate_mcp_credential` warning-level log on "not found" can spam Sentry on benign repeated stale-cred-id paths.
- `_service_name` strips `mcp.` literally — could be subtly trust-relevant for `mcp-proxy.attacker.com`-style URLs (out of scope for this PR; prompt-side guidance handles it).
- `handleManualToken` should clear `forceDisconnected` symmetrically with `handleConnect` at the **top** of the function, not after success.

Overall the headline fix (stale-cred shadow → setup card + invalidate) is sound, well-tested at the unit level, and the diff is appropriately scoped. The Sentry bug-prediction fix (`forceDisconnected`) is also correct. None of the findings above are merge-blockers given the user-impact / blast-radius trade-offs, but each is worth a thought before merging.

Findings posted as 9 inline comments across the changed files.

autogpt_platform/frontend/src/app/(platform)/copilot/tools/RunMCPTool/components/MCPSetupCard/MCPSetupCard.tsx
majdyz: 🤖 Fixed in [17a9b84519](https://github.com/Significant-Gravitas/AutoGPT/commit/17a9b84519): `liveHasCred` is now tri-state (`true`/`false`/`unknown`). When the API response isn't an array (loading or 5xx), we fall back to the persisted `initiallyConnected` snapshot rather than defaulting to disconnected. Test added: `falls back to persisted snapshot when the live cred API fails`.

autogpt_platform/frontend/src/app/(platform)/copilot/tools/RunMCPTool/components/MCPSetupCard/MCPSetupCard.tsx
majdyz: 🤖 Fixed in [17a9b84519](https://github.com/Significant-Gravitas/AutoGPT/commit/17a9b84519): added an `if (loading) return;` re-entrancy guard at the top of `handleConnect` (and `handleManualToken`) — a second click while a flow is in-flight bails cheaply instead of aborting the first popup. Belt-and-suspenders alongside the existing `disabled={loading}` JSX guard.

autogpt_platform/backend/backend/copilot/tools/run_mcp_tool.py
majdyz: 🤖 Fixed in [17a9b84519](https://github.com/Significant-Gravitas/AutoGPT/commit/17a9b84519): `surface_connect_card=true` now runs a lightweight `MCPClient.initialize` probe (one round-trip, no tool listing) when stored creds exist. If the server returns 401/403, we invalidate the stale cred + return a not-connected card. If the URL is invalid or unreachable, the probe will fail and we fall back to the `creds is not None` baseline — the next real tool call will surface a cleaner error. Test added: `test_surface_connect_card_stale_creds_invalidated_returns_not_connected`.

autogpt_platform/backend/backend/blocks/mcp/helpers.py
majdyz: 🤖 Fixed in [17a9b84519](https://github.com/Significant-Gravitas/AutoGPT/commit/17a9b84519): split out `ValueError` (already-gone) into a debug-level log; only true unexpected failures still warn.

autogpt_platform/frontend/src/app/(platform)/copilot/tools/RunMCPTool/components/MCPSetupCard/__tests__/MCPSetupCard.test.tsx
majdyz: 🤖 Fixed in [17a9b84519](https://github.com/Significant-Gravitas/AutoGPT/commit/17a9b84519): added `falls back to persisted snapshot when the live cred API fails` — sets mockLiveCreds=null and asserts the Connected/Reconnect render when `initiallyConnected=true`.

autogpt_platform/backend/backend/copilot/tools/test_run_mcp_tool.py
majdyz: 🤖 Fixed in [17a9b84519](https://github.com/Significant-Gravitas/AutoGPT/commit/17a9b84519): the `surface_connect_card` path now probes the server via `MCPClient.initialize` when creds exist. On 401/403 the stale row is deleted via `invalidate_mcp_credential` and the response carries `has_all_credentials=False` — same path the actual tool-execution route takes. Replaces the old test with `test_surface_connect_card_connected_when_creds_exist_and_probe_ok` (positive) and `test_surface_connect_card_stale_creds_invalidated_returns_not_connected` (the failure scenario you described).

autogpt_platform/frontend/src/app/(platform)/copilot/tools/RunMCPTool/components/MCPSetupCard/MCPSetupCard.tsx
majdyz: 🤖 Fixed in [17a9b84519](https://github.com/Significant-Gravitas/AutoGPT/commit/17a9b84519): `handleManualToken` now clears `forceDisconnected` at the start of every attempt and restores it in the catch block on failure — symmetric with `handleConnect`.

autogpt_platform/frontend/src/lib/oauth-popup.ts
majdyz: 🤖 Fixed in [17a9b84519](https://github.com/Significant-Gravitas/AutoGPT/commit/17a9b84519): the close-poll block is now guarded by `if (popup && !popupBlocked)`. When the browser blocks the popup we hand off to the new-tab fallback and the original `popup` reference (which may be already-closed synchronously) is ignored, so a successful sign-in arriving via BroadcastChannel/localStorage isn't short-circuited.

autogpt_platform/frontend/src/lib/oauth-popup.ts
majdyz: 🤖 Fixed in [17a9b84519](https://github.com/Significant-Gravitas/AutoGPT/commit/17a9b84519): added the `mcp_callback/route.ts` path hint to the grace-window comment.

autogpt_platform/frontend/src/lib/oauth-popup.ts
majdyz: 🤖 Deferring — the `popup=null` branch is covered indirectly by the new `popupBlocked` guard (close-poll never starts). A direct test asserting the promise stays pending until timeout is worth adding as a follow-up.

autogpt_platform/frontend/src/app/(platform)/copilot/tools/RunMCPTool/components/MCPSetupCard/MCPSetupCard.tsx
```diff
@@ -39,18 +47,70 @@ export function MCPSetupCard({ output, retryInstruction }: Props) {
...
   const [manualToken, setManualToken] = useState("");
-  const [connected, setConnected] = useState(false);
+  const [localConnected, setLocalConnected] = useState(initiallyConnected);
+  // When Reconnect fails (or any in-card flow errors out), force the
+  // not-connected branch even though the live cred row still exists —
+  // otherwise ``liveHasCred=true`` would keep the pill rendered and the
+  // user couldn't see the error banner or the manual-token input.  Reset
+  // on the next attempt so the user can retry.
+  const [forceDisconnected, setForceDisconnected] = useState(false);
   const oauthAbortRef = useRef<(() => void) | null>(null);
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [a3c9a5055](https://github.com/Significant-Gravitas/AutoGPT/commit/a3c9a5055): dropped `staleTime: 5_000` and added `refetchOnMount: "always"` so each MCPSetupCard mount re-queries `/api/integrations/credentials` from the server. After `invalidate_mcp_credential` deletes a stale row, the new card mounts and immediately sees the post-invalidation truth — no 5s window of misleading Connected state.

autogpt_platform/backend/backend/copilot/tools/run_mcp_tool.py
```diff
@@ -175,6 +187,43 @@ async def _execute(
...
+            if creds is not None:
+                probe_client = MCPClient(server_url, auth_token=auth_token)
+                try:
+                    await probe_client.initialize()
+                except HTTPClientError as probe_err:
+                    if probe_err.status_code in _AUTH_STATUS_CODES:
+                        await invalidate_mcp_credential(user_id, creds.id)
+                        connected = False
+                except MCPClientError:
+                    # Transport / protocol failure — leave the cred in
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [db761d9533](https://github.com/Significant-Gravitas/AutoGPT/commit/db761d9533): broadened the probe catch to also handle MCPClientError + bare Exception (5xx, asyncio.TimeoutError, network errors). Those now fall back to 'optimistically connected', preserving the still-valid cred row across transient outages. Only 401/403 triggers invalidation.

autogpt_platform/frontend/src/lib/oauth-popup.ts
```diff
@@ -181,14 +181,77 @@ export function openOAuthPopup(
...
+    // does NOT stack across re-renders, and the AbortController cleanup
+    // tears it down if the caller aborts before grace expires.
+    // Skip the close-poll entirely when the popup was blocked.  The
+    // ``window.open("about:blank", ...)`` reference can be non-null but
+    // already-closed in that branch (we fell back to a separate new-tab
+    // ``window.open(loginUrl, "_blank")`` whose handle we deliberately
+    // don't keep — the new tab is the user's primary surface and
+    // pre-emptively rejecting on its ``closed`` state would short-circuit
+    // a successful sign-in arriving via BroadcastChannel/localStorage.
+    if (popup && !popupBlocked) {
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [db761d9533](https://github.com/Significant-Gravitas/AutoGPT/commit/db761d9533): the 3s grace is now gated on useCrossOriginListeners. Same-origin flows (where opener.postMessage delivers synchronously) reject immediately on popup close — no fake spinner. Cross-origin flows still get the grace + final localStorage sweep for BroadcastChannel/storage-partition delivery latency.

autogpt_platform/backend/backend/copilot/tools/run_mcp_tool.py
```diff
@@ -175,6 +187,56 @@ async def _execute(
...
+                    await probe_client.initialize()
+                except HTTPClientError as probe_err:
+                    if probe_err.status_code in _AUTH_STATUS_CODES:
+                        await invalidate_mcp_credential(user_id, creds.id)
+                        connected = False
+                    # Other HTTP statuses (5xx, redirects, etc.) → leave
+                    # the cred in place and report "optimistically
+                    # connected" — the user can still try; the real tool
+                    # call will surface the actual error if it persists.
+                except (MCPClientError, Exception):
```
majdyz: 🤖 🟡 **Nice to Have**: `except (MCPClientError, Exception)` is redundant — `MCPClientError` is a subclass of `Exception` (verified via MRO: `(MCPClientError, Exception, BaseException, object)`), so the tuple collapses to just `except Exception`.

**Failure scenario (cosmetic, but real):**
1. Future reader sees the tuple and infers that `MCPClientError` needs special handling distinct from generic `Exception`.
2. They try to add `if isinstance(probe_err, MCPClientError): ...` inside the block to differentiate — and discover the dispatch is pointless because both branches collapse anyway.
3. Worse: a linter rule like `B014` (redundant exception type) may flag this in a future ruff upgrade and the file starts failing CI.

**Fix**: drop `MCPClientError` from the tuple (just `except Exception:`) — or, if the intent was actually to handle `MCPClientError` differently from other exceptions, split it into a separate `except MCPClientError:` clause above the generic catch.

Which was it — a leftover from incremental refactoring, or did you intend differentiated handling?

autogpt_platform/backend/backend/copilot/tools/test_run_mcp_tool.py
```diff
@@ -564,20 +617,118 @@ async def test_auth_error_with_existing_creds_returns_error():
...
+    session = make_session(_USER_ID)
+
+    with patch(
+        "backend.copilot.tools.run_mcp_tool.validate_url_host", new_callable=AsyncMock
+    ):
+        with patch(
+            "backend.copilot.tools.run_mcp_tool.auto_lookup_mcp_credential",
+            new_callable=AsyncMock,
+            return_value=None,
+        ):
```
majdyz, line 666: 🤖 🟡 **Nice to Have**: Test coverage gap on the new `surface_connect_card` probe path. Three tests added in this PR cover (probe-success, probe-401→invalidate, no-creds), but **the two "optimistically connected" branches have no test**:

1. **Non-auth `HTTPClientError`** (line 213-220 fall-through): probe returns 500/503/timeout-translated-to-HTTPClientError → `connected=True` still, cred is NOT invalidated.
2. **Generic `Exception`** (line 221-235): `asyncio.TimeoutError`, `OSError`, `MCPClientError` → `connected=True`, cred NOT invalidated, `logger.debug` fires.

**Failure scenario:**
1. A future refactor accidentally narrows `except Exception:` to `except MCPClientError:` (mirroring the import).
2. Now an `asyncio.TimeoutError` from a slow MCP server propagates out of `_execute`, kills the copilot turn with an unhandled exception.
3. CI is green because no test exercises the path.
4. Sentry catches it in prod — exactly the kind of "unknown, optimistically connected" outage this commit was meant to guard against.

**Fix**: add two parametrized cases (or two separate tests) asserting `response.has_all_credentials is True` AND `mock_invalidate.assert_not_awaited()` when the probe raises `HTTPClientError(status_code=500)` and when it raises `asyncio.TimeoutError("...")`.

autogpt_platform/frontend/src/lib/oauth-popup.ts
```diff
@@ -181,14 +181,93 @@ export function openOAuthPopup(
...
-          handled = true;
-          reject(new Error(OAUTH_ERROR_WINDOW_CLOSED));
-          controller.abort("popup_closed");
+          finalLocalStorageCheck();
+          if (handled) return;
+          if (POPUP_CLOSE_GRACE_MS === 0) {
+            // Same-origin path: any postMessage that was going to fire
+            // already did during the synchronous ``finalLocalStorageCheck``
+            // window (well, it's a no-op for same-origin, but the message
+            // listener queued from the popup before close would have run).
```
majdyz: 🤖 🔵 **Nit**: The comment in the `POPUP_CLOSE_GRACE_MS === 0` branch is muddled by its self-correcting parenthetical:

> "any postMessage that was going to fire already did during the synchronous `finalLocalStorageCheck` window (well, it's a no-op for same-origin, but the message listener queued from the popup before close would have run)"

When `useCrossOriginListeners=false`, `finalLocalStorageCheck` early-returns at line 233, so it does NO work. The parenthetical concedes this then pivots to a different mechanism (the `window.addEventListener("message", ...)` at line 149) — but that listener has nothing to do with `finalLocalStorageCheck`.

**Failure scenario:** A future reader thinks they need to call `finalLocalStorageCheck` in the same-origin branch too (because the comment frames it as the source of safety) and adds a `useCrossOriginListeners=true` override "just in case." Now same-origin flows pay the 3s grace too.

**Fix**: drop the parenthetical and tie the safety to the actual mechanism, e.g.:

```ts
// Same-origin path: the postMessage listener at line 149 fires
// synchronously when the popup posts before closing, so any
// successful result has already been handled. No grace needed.
```

autogpt_platform/frontend/src/app/(platform)/copilot/tools/RunMCPTool/components/MCPSetupCard/MCPSetupCard.tsx
```diff
@@ -123,8 +220,15 @@ export function MCPSetupCard({ output, retryInstruction }: Props) {
   async function handleManualToken() {
     const token = manualToken.trim();
     if (!token) return;
+    if (loading) return;
```
majdyz: 🤖 🔵 **Nit**: Re-entrancy guard placement is asymmetric between `handleConnect` (guard at line 135, **first** statement) and `handleManualToken` (guard at line 223, **second** statement — after the empty-token check at line 222).

Functionally equivalent for both call-sites, but the asymmetry will trip readers who scan both functions side-by-side looking for "is there a re-entrancy issue here too?" — they have to read further into `handleManualToken` to confirm the guard exists.

**Fix**: move `if (loading) return;` to the top of `handleManualToken` (before the `if (!token) return;` check). Both guards run cheaply; ordering them identically makes the pattern explicit.

autogpt_platform/frontend/src/app/(platform)/copilot/tools/RunMCPTool/components/MCPSetupCard/__tests__/MCPSetupCard.test.tsx
```diff
@@ -130,4 +189,59 @@ describe("MCPSetupCard", () => {
...
+    });
+    expect(screen.getByText(/does not support OAuth/)).toBeDefined();
+    expect(screen.queryByText(/connected to example\.com/i)).toBeNull();
+  });
+
+  it("drops Connected state on Reconnect failure even when live creds still report the server as connected", async () => {
+    // Sentry-flagged bug: when a stored cred exists (``liveHasCred=true``)
+    // and the user clicks Reconnect → OAuth 400, the previous
+    // ``connected = localConnected || liveHasCred`` logic would keep the
+    // Connected pill rendered because liveHasCred was still true.  The
```
majdyz, line 243: 🤖 🟡 **Nice to Have**: New `if (loading) return;` re-entrancy guards in `handleConnect` (line 135) and `handleManualToken` (line 223) have no test. The comment at line 126-134 documents the exact race they protect against (double-click → second call aborts first popup → first `await promise` rejects with `OAUTH_ERROR_FLOW_CANCELED` → `forceDisconnected=true` despite second attempt being alive) — but nothing pins this behavior.

**Failure scenario:**
1. Future refactor adds a wrapping `useCallback` to `handleConnect` and accidentally moves the guard inside the callback closure but evaluates `loading` from a captured stale render — guard becomes a no-op.
2. Sentry starts logging "Sign-in window was closed" from the very flow this guard was meant to fix.
3. No CI signal because no test exercises a rapid double-click.

**Fix**: one test per function with `fireEvent.click(btn); fireEvent.click(btn);` (no await between) and assert `vi.mocked(postV2InitiateOauthLoginForAnMcpServer).mock.calls.length === 1`.

autogpt_platform/frontend/src/app/(platform)/copilot/tools/RunMCPTool/components/MCPSetupCard/MCPSetupCard.tsx
```diff
@@ -39,18 +47,99 @@ export function MCPSetupCard({ output, retryInstruction }: Props) {
...
+  // initiallyConnected=false + 5xx race that surfaces a bare Connect
+  // button despite an existing cred.
+  const liveHasCred: boolean | "unknown" = !Array.isArray(liveCredsRes)
+    ? "unknown"
+    : liveCredsRes.some(
+        (c) =>
+          c.provider === "mcp" &&
+          typeof c.host === "string" &&
+          normalizeMcpUrl(c.host) === normalizedServer,
+      );
```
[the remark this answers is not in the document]
majdyz: 🤖 Partial false-positive but addressed for hygiene in [f17d924047](https://github.com/Significant-Gravitas/AutoGPT/commit/f17d924047).

**Runtime flow is actually safe**: `liveHasCred` reads from `useGetV1ListCredentials` → `/api/integrations/credentials` → `to_meta_response` → `CredentialsMetaResponse.get_host` (router.py:156-166), which uses `cred.metadata.get('mcp_server_url')` (FULL URL) for both OAuth and bearer-token MCP creds. So the matching at MCPSetupCard.tsx:87-94 always sees a full URL on both sides — bearer-token reload still renders Connected/Reconnect.

**Fixed inconsistency anyway**: the bearer-token POST response at routes.py:479 was returning just `hostname` while the OAuth callback at routes.py:392 returned the full URL. Now both POST handlers return `credentials.metadata.get('mcp_server_url')` for parity, so any consumer using the immediate POST response (instead of re-fetching the list) sees the same shape.

autogpt_platform/backend/backend/copilot/tools/run_mcp_tool.py
majdyz: 🤖 Fixed in [f17d924047](https://github.com/Significant-Gravitas/AutoGPT/commit/f17d924047): `except (MCPClientError, Exception)` → `except Exception`. Was a leftover from incremental refactoring — MCPClientError was in the tuple to make the intent ('catch protocol errors too') explicit, but since it's a subclass of Exception the tuple collapsed. Updated the comment to keep MCPClientError mentioned as an example of what the broad catch covers, so future readers still see the rationale.

autogpt_platform/backend/backend/copilot/tools/test_run_mcp_tool.py
majdyz: 🤖 Fixed in [f17d924047](https://github.com/Significant-Gravitas/AutoGPT/commit/f17d924047): added `test_surface_connect_card_probe_5xx_reports_optimistically_connected` and `test_surface_connect_card_probe_timeout_reports_optimistically_connected`. Both assert `has_all_credentials=True` AND `mock_invalidate.assert_not_awaited()` to pin the 'never delete a cred on transient failure' invariant.

autogpt_platform/frontend/src/lib/oauth-popup.ts
majdyz: 🤖 Fixed in [f17d924047](https://github.com/Significant-Gravitas/AutoGPT/commit/f17d924047): comment now ties the safety to the actual mechanism (`window.addEventListener('message', …)` at line 149) instead of the self-correcting parenthetical about `finalLocalStorageCheck`.

autogpt_platform/frontend/src/app/(platform)/copilot/tools/RunMCPTool/components/MCPSetupCard/MCPSetupCard.tsx
majdyz: 🤖 Fixed in [f17d924047](https://github.com/Significant-Gravitas/AutoGPT/commit/f17d924047): `if (loading) return;` moved to the top of `handleManualToken`, mirroring `handleConnect`.

autogpt_platform/frontend/src/app/(platform)/copilot/tools/RunMCPTool/components/MCPSetupCard/__tests__/MCPSetupCard.test.tsx
majdyz: 🤖 Fixed in [f17d924047](https://github.com/Significant-Gravitas/AutoGPT/commit/f17d924047): added 're-entrancy guard prevents handleConnect from firing twice on rapid double-click' — pins one network call from two synchronous `fireEvent.click` calls. Uses an unresolved promise to keep `loading=true` between clicks, then drains on assertion.

autogpt_platform/frontend/src/app/(platform)/copilot/tools/RunMCPTool/components/MCPSetupCard/MCPSetupCard.tsx
```diff
@@ -39,18 +47,99 @@ export function MCPSetupCard({ output, retryInstruction }: Props) {
...
+
   const [loading, setLoading] = useState(false);
   const [error, setError] = useState<string | null>(null);
   const [showManualToken, setShowManualToken] = useState(false);
   const [manualToken, setManualToken] = useState("");
-  const [connected, setConnected] = useState(false);
+  const [localConnected, setLocalConnected] = useState(initiallyConnected);
+  // When Reconnect fails (or any in-card flow errors out), force the
+  // not-connected branch even though the live cred row still exists —
+  // otherwise ``liveHasCred=true`` would keep the pill rendered and the
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [81bc204c6a](https://github.com/Significant-Gravitas/AutoGPT/commit/81bc204c6a): `localConnected` no longer seeds from `initiallyConnected` — it starts at `false` and only flips to `true` when the user successfully completes OAuth/manual-token *in this component instance*. The persisted snapshot path is handled exclusively via `liveSays` (live API → unknown? → fallback to initial). When live API now confirms `liveHasCred=false`, that wins over a stale `initiallyConnected=true` snapshot. Regression test added: `drops Connected state when live API confirms the cred is gone, even if persisted snapshot said connected`. 12/12 tests pass.

autogpt_platform/backend/backend/copilot/tools/run_mcp_tool.py
```diff
@@ -175,6 +187,56 @@ async def _execute(
...
+                    # uncaught exception out of ``_execute``.
+                    logger.debug(
+                        "MCP probe for surface_connect_card failed for %s — "
+                        "reporting optimistically connected",
+                        server_host(server_url),
+                        exc_info=True,
+                    )
+            return self._build_setup_requirements(
+                server_url, session_id, connected=connected
+            )
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [5f2af426f2](https://github.com/Significant-Gravitas/AutoGPT/commit/5f2af426f2): added `MCPClient.close()` that sends DELETE on the session URL with the `Mcp-Session-Id` header (best-effort — failures swallowed) and resets `self._session_id`. `run_mcp_tool` wraps the probe in `try/finally` so `await probe_client.close()` runs even when `initialize()` raises (401/timeout/etc.). Regression tests assert close is awaited in both the probe-OK and probe-401 paths.

autogpt_platform/frontend/src/app/(platform)/copilot/tools/RunMCPTool/components/MCPSetupCard/MCPSetupCard.tsx
```diff
@@ -121,10 +224,20 @@ export function MCPSetupCard({ output, retryInstruction }: Props) {
...
     setLoading(true);
     setError(null);
+    // Clear the force-disconnect override at the start so a failed
+    // manual-token attempt doesn't leave the not-connected branch wedged
+    // open on subsequent retries even though ``liveHasCred=true``.  The
+    // catch block restores it on failure; the success branch leaves it
+    // cleared.
+    setForceDisconnected(false);
     try {
       const res = await postV2StoreABearerTokenForAnMcpServer({
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [c3bedf3fd6](https://github.com/Significant-Gravitas/AutoGPT/commit/c3bedf3fd6): `setForceDisconnected(false)` moved out of the pre-await block in both `handleConnect` and `handleManualToken` — now only fires on the success branch, AFTER the API confirms the cred is stored. Mid-flight the `forceDisconnected=true` override stays on so `liveHasCred=true` can't briefly render the Connected pill. The catch branch still flips it back on (no-op when it was already true).

autogpt_platform/frontend/src/app/(platform)/copilot/tools/RunMCPTool/components/MCPSetupCard/MCPSetupCard.tsx
```diff
@@ -145,10 +266,30 @@ export function MCPSetupCard({ output, retryInstruction }: Props) {
...
+            className="ml-1 underline underline-offset-2 hover:no-underline disabled:opacity-60"
+          >
+            {loading ? "Reconnecting…" : "Reconnect"}
+          </button>
+        </div>
+        {error && (
+          <div className="rounded-lg border border-red-200 bg-red-50 px-3 py-2 text-sm text-red-700">
+            {error}
+          </div>
+        )}
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [cff7ab9c91](https://github.com/Significant-Gravitas/AutoGPT/commit/cff7ab9c91): correct — every catch sets `forceDisconnected=true` before `setError`, which forces `connected=false`. Removed the unreachable error block and added a comment explaining the invariant so the next reader doesn't restore it. 14/14 tests still pass.

PR #13217, opened 2026-05-26
feat(platform): global search endpoint + bucketed command-palette modal

Abhi1992002:
### Why / What / How

**Why.** The Cmd-K search modal in the chat sidebar only ever covered chat sessions. Users that have agents, workspace files, **and** chats had no way to jump between them from a single keystroke, and the server side filtered chats with a `LOWER(...) LIKE %q%` over the whole table for every keystroke. We need a single search that ranks across every kind of content the user owns, scales beyond a `LIKE`, and is wired through a generic UI primitive that future surfaces can reuse.

**What.** Adds a new `GET /api/search/global` endpoint that returns hybrid (semantic + lexical + BM25) results when `q` is non-empty and the most-recently-updated items per content type when it is empty — bucketed into **agents** (library + store), **workspace files**, and **chat sessions**. The frontend exposes the same flow through a new Cmd-K modal in the chat sidebar that replaces the chat-only `ChatSearchModal`.

**How.** 

- **Backend.** Pulls the generic embedding + hybrid-search engine that lived under `store/` into a new `search/` feature, leaving `store/embeddings.py` and `store/hybrid_search.py` as thin store-specific wrappers. Adds two new `ContentType` enum values (`WORKSPACE_FILE`, `CHAT_SESSION`) plus a tightened `UnifiedContentEmbedding` `CHECK` constraint so any per-user content type can never be stored with a NULL `userId` (defence-in-depth against the public branch of the user filter). New per-feature indexers (`workspace/embeddings.py`, `copilot/chat_session_embeddings.py`) hand rows to the shared engine; only the SQL join that pulls feature-specific metadata stays in the feature.
- **Frontend.** Splits the modal into a generic, app-agnostic `SearchCommandModal` **organism** (controlled, bucketed, keyboard-navigable, slot-based empty/error/loading states) and a thin `GlobalSearchModal` **container** that wires it to the new endpoint via the generated React Query hook. The chat sidebar opens the container with the existing Cmd/Ctrl-K shortcut. The mobile drawer still uses the chat-only search hook, so `ChatSearchResults` / `useChatSearch` / `helpers` stay; only the desktop modal component + its tests are deleted.

### Changes 🏗️

**Backend (`autogpt_platform/backend/`)**
- `backend/api/features/search/` — new feature module:
  - `service.py` + `routes.py` — `/global` endpoint with `per_type_limit`, per-user TTL cache on the empty-query path.
  - `model.py` — `GlobalSearchResponse` + bucketed `SearchResultItem` shapes.
  - `embeddings.py` — generic `UnifiedContentEmbedding` CRUD + `semantic_search` + `backfill_all_content_types`.
  - `hybrid_search.py` — `unified_hybrid_search` engine with BM25 reranking.
  - `content_handlers.py` — pluggable per-`ContentType` handlers (store agent, block, doc, library agent, workspace file, chat session).
  - `README.md` documents the split.
- `backend/api/features/store/embeddings.py`, `store/hybrid_search.py` — collapsed into thin store-listing wrappers around the shared engine.
- `backend/api/features/workspace/embeddings.py` — new fire-and-forget indexer for workspace files.
- `backend/copilot/chat_session_embeddings.py` — new fire-and-forget indexer for chat sessions; `copilot/model.py` triggers it on session updates.
- `backend/api/rest_api.py` — registers the new `/api/search` router.
- `backend/data/db_manager.py` + `backend/util/workspace.py` — small touch-ups required by the workspace indexer.
- `migrations/20260526120000_add_file_and_chat_content_types/` — adds `WORKSPACE_FILE` + `CHAT_SESSION` enum values and replaces the `LIBRARY_AGENT`-only CHECK with one that covers all three per-user types.
- `schema.prisma` — matching enum + constraint changes.

**Frontend (`autogpt_platform/frontend/`)**
- `src/components/organisms/SearchCommandModal/` — generic organism (`SearchCommandModal.tsx` + `SearchCommandResults.tsx` + `SearchCommandResultItem.tsx` + `useKeyboardNav.ts` + `helpers.ts`) with `__tests__/` and `SearchCommandModal.stories.tsx` (mock fixtures, sized wrapper that anchors the fixed-positioned modal inside the Storybook docs iframe).
- `src/app/(platform)/components/GlobalSearchModal/` — thin container that adapts the `GET /api/search/global` response into the organism's prop shape, owns the open/close + debounced query state, and routes clicks to the right page per `SearchResultItemType`.
- `src/app/(platform)/copilot/components/ChatSidebar/ChatSidebar.tsx` — now opens `GlobalSearchModal` instead of `ChatSearchModal`.
- `src/app/(platform)/copilot/components/ChatSearchModal/ChatSearchModal.tsx` + its `__tests__/ChatSearchModal.test.tsx` — deleted.
- `.storybook/preview.tsx` — adds a `QueryClientProvider` so any future API-driven story Just Works.
- `.storybook/main.ts` — includes `src/components/organisms/**/*.stories`.
- `src/services/feature-flags/use-get-flag.ts`, `src/mocks/mock-handlers.ts`, `src/app/api/openapi.json` — feature flag + mocks + generated spec for the new endpoint.

#### For code changes:
- [x] I have made a test plan
  - [x] **Backend unit/integration tests pass.** `poetry run pytest backend/api/features/search backend/api/features/store` → 99/99 pass (excluding e2e/integration suites that need a live DB; those mirror the unit tests with real Prisma).
  - [x] **Frontend integration tests pass.** `pnpm test:unit` → **2844 / 2844** pass (229 test files).
  - [x] **Targeted suites green.** `SearchCommandModal` (15), `GlobalSearchModal` (11), `ChatSidebar` (14) all pass after the rewrite.
  - [x] **Typecheck clean.** `pnpm tsc --noEmit` → no errors.
  - [x] **Lint clean on touched files.** `pnpm lint` reports no new errors; pre-existing warnings on unrelated files only.
  - [x] **Storybook.** All six `Organisms → SearchCommandModal` stories (`WithResults`, `Loading`, `Empty`, `NoMatches`, `SingleBucket`, `ErrorState`) render fully inside the docs iframe without overflow or scrollbars; mock data only, no API calls.

#### Manual smoke test
- [x] Cmd-K from the chat sidebar opens the modal; typing fires a debounced request; results render bucketed by Chats → Agents → Files.
- [x] Empty input returns the per-user "recent items" payload (cached) — same component, no flash-of-empty.
- [x] Arrow-keys + Enter walk the flat list across buckets; Escape and backdrop click close the modal; the X button clears the query.
- [x] Clicking a result routes to the right page per content type (library agent, store listing, workspace file, chat session) and closes the modal.

#### For database schema changes
- [x] **New migration included** under `backend/migrations/20260526120000_add_file_and_chat_content_types/`. Adds `WORKSPACE_FILE` and `CHAT_SESSION` to the `ContentType` enum (`ADD VALUE IF NOT EXISTS` so it is idempotent), sweeps stray null-`userId` rows for the new types, then replaces the `LIBRARY_AGENT`-only `CHECK` with one that covers all three per-user types. Schema reflected in `schema.prisma`.

#### Out of scope (intentionally not changed)
- The mobile drawer's `ChatSearchResults` + `useChatSearch` still exist — separate UX. Will be unified in a follow-up.
- Pre-existing `store/routes_test.py` failures on `dev` are not addressed here.


autogpt_platform/backend/backend/api/features/search/service.py
```diff
@@ -0,0 +1,268 @@
...
+async def _search_bucket(
+    query: str,
+    user_id: str,
+    content_types: list[ContentType],
+    limit: int,
+) -> list[SearchResultItem]:
+    """Run unified_hybrid_search for one bucket; swallow failures so a
+    broken bucket doesn't 500 the whole response."""
+    try:
+        rows, _total = await search().unified_hybrid_search(
```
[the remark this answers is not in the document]
Abhi1992002: fixed

autogpt_platform/backend/backend/api/features/search/hybrid_search.py
```diff
@@ -0,0 +1,511 @@
...
+            original_score_field="combined_score",
+        )
+
+    # Clean up results
+    for result in results:
+        result.pop("total_count", None)
+
+    logger.info(f"Unified hybrid search: {len(results)} results, {total} total")
+
+    return results, total
```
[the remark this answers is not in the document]
Abhi1992002: Not a correctness bug — lexical/term relevance is already part of the **global** ranking that drives pagination, not just the in-memory step.

In the SQL CTE, lexical relevance is computed with `ts_rank_cd(uce.search, plainto_tsquery(...))` (the Postgres BM25-equivalent), normalized to `lexical_score`, and folded into `combined_score` with weight `0.40` — the largest single weight. The final `ORDER BY combined_score DESC, content_id ASC` + `LIMIT/OFFSET` is what decides page membership. So an exact term match on a later page *is* already considered for page 1 via `lexical_score`; the premise that it "never gets considered" doesn't hold.

`bm25_rerank` after pagination is a deliberate lightweight within-page refinement layered on top of the SQL `combined_score`, not the primary lexical signal.

The suggested `limit * k` overfetch + in-memory rerank would also defeat the deterministic `LIMIT/OFFSET` stable pagination intentionally introduced in `b253c49e95` (the `content_id` tiebreaker at L398-402 exists precisely so the same row can't shift across pages). Reranking an overfetched window per request would make page boundaries inconsistent between page loads. Leaving as-is.

autogpt_platform/backend/backend/api/features/search/content_handlers.py
```diff
@@ -730,10 +730,197 @@ async def get_stats(self) -> dict[str, int]:
...
+                cs."userId",
+                cs.title
+            FROM {schema_prefix}"ChatSession" cs
+            LEFT JOIN {schema_prefix}"UnifiedContentEmbedding" uce
+              ON uce."contentId" = cs.id
+              AND uce."contentType" = 'CHAT_SESSION'::{schema_prefix}"ContentType"
+              AND uce."userId" = cs."userId"
+            WHERE cs.title IS NOT NULL
+              AND cs.title <> ''
+              AND uce."contentId" IS NULL
```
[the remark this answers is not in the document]
Abhi1992002: Fixed in 8d6eec2a0c — switched the backfill and both stats queries to `btrim(title) <> ''` so whitespace-only titles are filtered server-side and stop consuming backfill slots.

autogpt_platform/backend/backend/api/features/search/embeddings.py
```diff
@@ -0,0 +1,852 @@
...
+    total_failed = 0
+    all_errors: dict[str, int] = {}  # Aggregate errors across all content types
+
+    # Process content types in explicit order
+    processing_order = [
+        ContentType.BLOCK,
+        ContentType.STORE_AGENT,
+        ContentType.DOCUMENTATION,
+        ContentType.LIBRARY_AGENT,
+    ]
```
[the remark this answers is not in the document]
Abhi1992002: Fixed in 8d6eec2a0c — both lists now derive from a single `SEARCHABLE_EMBEDDING_TYPES` constant that includes `WORKSPACE_FILE` and `CHAT_SESSION` (both have registered handlers), so the admin backfill/cleanup path now covers them.

autogpt_platform/frontend/src/app/api/openapi.json
```diff
@@ -8153,6 +8153,65 @@
         }
       }
     },
+    "/api/search/global": {
+      "get": {
+        "tags": ["v2", "search", "search"],
+        "summary": "Global search (hybrid on query, recent on empty)",
+        "operationId": "getV2Global search (hybrid on query, recent on empty)",
```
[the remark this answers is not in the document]
Abhi1992002: Fixed in 4daf5eeb8c — shortened the route `summary` to "Global search", so the generated operationId is now `getV2Global search` → hook `useGetV2GlobalSearch` (was `useGetV2GlobalSearchHybridOnQueryRecentOnEmpty`). Regenerated openapi.json and updated the hook + MSW-handler consumers. Note: it wasn't actually breaking generation — Orval sanitized the spaces/parens into a valid (if verbose) name, which is why CI (incl. `check API types` and the frontend integration tests) was green — but the cleaner id is worth it.

autogpt_platform/frontend/src/app/(platform)/components/GlobalSearchModal/helpers.ts
```diff
@@ -0,0 +1,72 @@
...
+
+const BUCKET_ORDER: BucketKey[] = ["chats", "agents", "files"];
+
+function toCommandItem(item: SearchResultItem): SearchCommandItem {
+  return {
+    id: item.id,
+    title: item.title,
+    subtitle: item.subtitle ?? null,
+    icon: ICON_BY_TYPE[item.type],
+  };
```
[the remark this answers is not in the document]
Abhi1992002: Fixed in 9c50484291 — `itemsById` and the command-item `id` are now keyed by `${type}:${id}` via a shared `itemKey()` helper, so results from different buckets sharing an ID can't clobber each other. Lookup in `GlobalSearchModal` stays consistent since both sides use the same key.

autogpt_platform/frontend/src/components/organisms/SearchCommandModal/SearchCommandModal.stories.tsx
```diff
@@ -0,0 +1,216 @@
...
+        buckets={filteredBuckets}
+        isLoading={isLoading}
+        isError={isError}
+        placeholder={placeholder}
+        inputAriaLabel={inputAriaLabel}
+        idleEmptyLabel={idleEmptyLabel}
+        searchingEmptyLabel={searchingEmptyLabel}
+        onSelectItem={(item) => {
+          // eslint-disable-next-line no-console
+          console.log("Selected:", item);
```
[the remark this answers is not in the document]
Abhi1992002: Fixed in 9c50484291 — removed the `console.log` and its `eslint-disable` in the story; the callback just closes the modal now.

autogpt_platform/backend/backend/api/features/search/hybrid_search.py
```diff
@@ -0,0 +1,511 @@
...
+"""
+
+import logging
+import re
+import time
+from dataclasses import dataclass
+from typing import Any
+
+from prisma.enums import ContentType
+from rank_bm25 import BM25Okapi  # type: ignore[import-untyped]
```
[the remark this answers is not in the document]
Abhi1992002: Both are pre-existing, not introduced here — this file was moved verbatim from `store/hybrid_search.py`, where the `# type: ignore[import-untyped]` (rank_bm25 ships no type stubs) and the `category` parameter are identical. Out of scope for this refactor PR; changing the `category` signature would also affect the existing store callers. Tracking separately rather than expanding scope here.

autogpt_platform/backend/backend/api/features/search/hybrid_search.py
```diff
@@ -0,0 +1,511 @@
...
+    query: str,
+    content_types: list[ContentType] | None = None,
+    category: str | None = None,
+    page: int = 1,
+    page_size: int = 20,
+    weights: UnifiedSearchWeights | None = None,
+    min_score: float | None = None,
+    user_id: str | None = None,
+    lexical_query: str | None = None,
+) -> tuple[list[dict[str, Any]], int]:
```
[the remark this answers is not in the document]
Abhi1992002: Same as above — the `category` parameter is pre-existing behavior moved from `store/hybrid_search.py` (unchanged by this PR). It currently feeds the category-score boost rather than a hard filter; tightening or renaming it touches the shared engine + store callers, so it's out of scope for this move. Will track separately.

autogpt_platform/backend/backend/api/features/search/service.py
```diff
@@ -0,0 +1,265 @@
...
+
+    try:
+        resp = await library_db().list_library_agents(
+            user_id=user_id,
+            page=1,
+            page_size=limit,
+            sort_by=library_model.LibraryAgentSort.UPDATED_AT,
+        )
+    except Exception as e:
+        logger.warning("Failed to list recent library agents for %s: %s", user_id, e)
```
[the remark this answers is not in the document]
Abhi1992002: Leaving as-is: `user_id` here is an opaque UUID used as a correlation identifier and is logged the same way throughout the backend (it's not PII like email/name). Redacting it would break log correlation/debuggability for these best-effort warning paths, and would be inconsistent with existing conventions.

autogpt_platform/backend/backend/copilot/chat_session_embeddings.py
```diff
@@ -0,0 +1,75 @@
...
+        if existing and existing.get("searchableText") == searchable_text:
+            return
+        await ensure_content_embedding(
+            content_type=ContentType.CHAT_SESSION,
+            content_id=session_id,
+            searchable_text=searchable_text,
+            metadata={"title": searchable_text},
+            user_id=user_id,
+            force=True,
+        )
```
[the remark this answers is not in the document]
Abhi1992002: By design, and consistent with the module it mirrors (`api/features/library/embeddings.py`), which uses the same fire-and-forget, last-write-wins pattern with no ordering guard. Title (re-)embeds are rare and user-driven, the write is best-effort, and the periodic re-embedder reconciles any drift. Adding sequencing/locking would diverge from the established library pattern for negligible benefit.

autogpt_platform/backend/backend/api/features/search/embeddings.py
```diff
@@ -0,0 +1,855 @@
...
+            else:
+                # Skip unknown content types to avoid accidental deletion
+                logger.warning(
+                    f"Skipping cleanup for unknown content type: {content_type}"
+                )
+                results_by_type[content_type.value] = {
+                    "deleted": 0,
+                    "error": "Unknown content type - skipped for safety",
+                }
+                continue
```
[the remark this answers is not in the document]
Abhi1992002: Good catch — valid bug, fixed in 57f4ea7da0. Adding the types to `cleanup_types` alone wasn't enough: the loop derives valid `current_ids` from a per-type branch, and the new types were falling through to the 'unknown type → skip' guard. Added `elif` branches that enumerate valid IDs (non-deleted `UserWorkspaceFile` rows; `ChatSession` rows with a non-empty trimmed title — matching the embedder filters), so orphaned WORKSPACE_FILE/CHAT_SESSION embeddings are now actually removed.

autogpt_platform/backend/backend/api/features/search/service.py
```diff
@@ -0,0 +1,268 @@
...
+            user_id=user_id,
+            page=1,
+            page_size=limit,
+            # The unified default (0.15) is calibrated for relevance pages;
+            # use a slightly more permissive floor for the global top-N so
+            # we still return *something* on short queries.
+            min_score=0.05,
+            # Search-as-you-type: the trailing token is usually a partial
+            # word, so prefix-match the lexical signal ("se" -> "se:*").
+            prefix_match=True,
```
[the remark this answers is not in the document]
Abhi1992002: Fixed in bb6cddbd00 — added `_should_prefix_match` in `service.py` that gates `prefix_match=True` on the trailing token being ≥ 2 chars. Single-char queries now fall back to whole-word `plainto_tsquery`, so "s" can't expand to "s:*" and explode the lexical candidate set before `ts_rank_cd` scoring. The lexical branch still has no LIMIT, but a 2-char prefix is bounded enough in practice (and `min_score=0.20` catches the rest).

ntindle (dismissed):

autogpt_platform/backend/backend/api/features/search/service.py
```diff
@@ -0,0 +1,331 @@
...
+        logger.warning("Failed to list chat sessions for %s: %s", user_id, e)
+        return []
+
+    items: list[SearchResultItem] = []
+    for session in sessions[:limit]:
+        items.append(
+            SearchResultItem(
+                id=session.session_id,
+                type="chat_session",
+                title=session.title or "Untitled chat",
```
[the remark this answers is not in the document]
Abhi1992002: Looked into this — the recents-view inclusion of untitled sessions is intentional, and the embedding-indexer comparison is no longer the relevant reference behaviour after the refactor in ed0a2290da.

`_chats_bucket` (both empty and non-empty query paths) now hits `ChatSession` directly via `get_user_chat_sessions` with an optional `title_contains` ILIKE filter — it does **not** consult `UnifiedContentEmbedding` at all. So the `ChatSessionHandler.get_missing_items` filter (`btrim(title) <> ''`) only governs which sessions get *embedded for the OTHER content types' pipelines*, not what shows in /search/global.

For the recents (empty-query) bucket specifically, surfacing untitled sessions is the intended UX — they're real recent activity the user might want to resume, the frontend renders them with the `'Untitled chat'` placeholder, and `test_global_search_empty_query_untitled_chat_falls_back` asserts exactly this fallback. Filtering them out would mean a freshly-created session disappears from recents until copilot autopopulates its title in the first turn.

No code change.

Abhi1992002:
Triaging the dev-reviewer findings — addressed three in-scope items in 40b952fd6b, and recording rationale for the rest below so @ntindle can decide what (if anything) to block on for this PR.

### Landed in 40b952fd6b
- **`tiktoken` encoder cached at module level** (perf 🟠 #4). `encoding_for_model()` was being called per `generate_embedding()` invocation; now a single `_ENCODER` is built at import. Also reuses the already-computed token count in the truncation log instead of re-encoding.
- **`LIMIT 500` on the lexical candidate CTE** (perf 🟠 #7). Matches the semantic branch's `LIMIT 200` bounded design. Flagged independently by CodeRabbit and this review.
- **`item_type: Any` → `SearchItemType`** (nit #1).

### Already addressed earlier in this PR
- **CodeRabbit prefix-match thread**: `bb6cddbd00` gates `prefix_match=True` on the trailing token being ≥ 2 chars, so a single-char query can't expand to `"s:*"` and explode the candidate set even before the new `LIMIT 500` engages.
- **Sentry "untitled chats in recents" thread**: replied inline — recents intentionally surfaces untitled sessions (existing test asserts this fallback), and after the ed0a2290da refactor chats no longer consult the embedding indexer at all, so the original consistency argument doesn't hold.

### Deferred — pre-existing code outside this PR's diff
These were flagged on `embeddings.py` / `hybrid_search.py` regions but are equally true on `dev`. Happy to do them as follow-up PRs once this one merges so the diff stays focused on the global-search feature:

- **🔴 `cleanup_orphaned_embeddings` unbounded memory** (`embeddings.py:536`). Real concern at scale, but the function is the pre-existing maintenance task — this PR only adds two enum branches to it. Fix wants `DELETE … WHERE NOT EXISTS (…)` with batching, which is its own change.
- **🟠 `semantic_search()` missing user-scope guard** (`embeddings.py:764`). The new `unified_hybrid_search` path **does** have the defence-in-depth `NOT (contentType IN (…) AND userId IS NULL)` guard plus the DB CHECK constraint added in `20260526120001`. `semantic_search()` is the old single-vector helper; tightening it is right but it's existing code.
- **🟠 ILIKE wildcard injection in lexical fallback** (`embeddings.py:840`). Same scope argument — pre-existing path, separate hardening PR.
- **🟠 `backfill_all_content_types` no concurrency cap** (`embeddings.py:445`). Pre-existing CLI/maintenance helper, not in the request path.
- **🟠 Missing unit tests for `WorkspaceFileHandler` / `ChatSessionHandler` / `workspace/embeddings.py`**. Fair point, but worth its own testing PR — the existing handler tests use snapshot-style SQL assertions that need careful per-handler fixturing.

### Won't fix in this PR
- **Splitting `content_handlers.py` (926L) / `embeddings.py` (880L) / `ChatSidebar.tsx` (583L)**. The files exceed the soft guideline, but mid-feature is the wrong moment to restructure existing modules — happy to do this in dedicated refactor PRs that reviewers can read independently.
- **`cleanup_orphaned_embeddings` OCP refactor via `ContentHandler.get_valid_content_ids()`**. Same reasoning — touches every existing handler.

CI was green pre-rebase (42/42); will repoll after this push settles.

ntindle (dismissed):

autogpt_platform/backend/backend/api/features/search/service.py
```diff
@@ -0,0 +1,336 @@
...
+    """Convert one ``unified_hybrid_search`` row into a ``SearchResultItem``.
+
+    Returns None when the row's content type is one we don't surface
+    through /search/global (e.g. BLOCK / DOCUMENTATION rows that might
+    sneak in if a caller widens the type filter).
+    """
+    raw_type = row.get("content_type")
+    # Both Prisma enum members and bare strings can show up depending on
+    # whether the search ran via the in-process accessor or the RPC shim.
+    type_str = getattr(raw_type, "value", raw_type)
```
majdyz: 🤖 🟠 **Should Fix**: `getattr(raw_type, "value", raw_type)` is duck-typing for type dispatch, which `backend/AGENTS.md` explicitly forbids ("No duck typing — no `hasattr`/`getattr`/`isinstance` for type dispatch; use typed interfaces/unions/protocols").

The comment names the real underlying problem: "Both Prisma enum members and bare strings can show up depending on whether the search ran via the in-process accessor or the RPC shim." That's a type-inconsistency at the boundary that should be normalized at the source — pick one (e.g. always return `ContentType` enum members from both accessor paths, or always serialise to `.value` strings) and assert it in the row type. Then this function can branch on a typed value without the runtime fallback, and `metadata if isinstance(metadata, dict) else {}` at line 101 also goes away.

autogpt_platform/backend/backend/api/features/search/routes.py
```diff
@@ -0,0 +1,56 @@
...
+    dependencies=[Security(autogpt_auth_lib.requires_user)],
+)
+
+
+@router.get(
+    "/global",
+    summary="Global search",
+    response_model=GlobalSearchResponse,
+)
+async def global_search_endpoint(
```
majdyz, line 41: 🤖 🟠 **Should Fix**: `/api/search/global` has no rate limit, but every non-empty `q` triggers an OpenAI embedding call (real $$) — see `unified_hybrid_search` → `embed_query` in `hybrid_search.py:225`. A key-held client or a script that bypasses the frontend's debounce can rack up embedding spend per user with no backpressure.

For reference, comparable per-user endpoints (`backend/api/features/chat/routes.py:46-54`) use `check_rate_limit` / `enforce_payment_paywall`. Apply the same here — at minimum a per-user QPS cap on the non-empty-query branch (the cached empty-query branch is fine).

autogpt_platform/backend/backend/api/features/search/embeddings.py
```diff
@@ -0,0 +1,886 @@
...
+                )
+                current_ids = {row["id"] for row in valid_agents}
+            elif content_type == ContentType.DOCUMENTATION:
+                # Use DocumentationHandler to get section-based content IDs
+                from backend.api.features.search.content_handlers import (
+                    DocumentationHandler,
+                )
+
+                doc_handler = CONTENT_HANDLERS.get(ContentType.DOCUMENTATION)
+                if isinstance(doc_handler, DocumentationHandler):
```
majdyz: 🤖 🟠 **Should Fix**: `isinstance(doc_handler, DocumentationHandler)` here is using runtime narrowing to dispatch to `_get_docs_root` / `_get_all_section_content_ids`, which only live on `DocumentationHandler`, not the base `ContentHandler`. `backend/AGENTS.md` bans `isinstance` for type dispatch ("use typed interfaces/unions/protocols").

The fix: lift the doc-specific operations to a typed seam — either declare them as abstract on `ContentHandler` with no-op defaults on handlers that have no filesystem root, or split a `LocalContentHandler` (Protocol/ABC) that `DocumentationHandler` implements and look it up via a typed registry (`CONTENT_HANDLERS_BY_LOCAL: dict[ContentType, LocalContentHandler]`). Then the `isinstance` and the inline `DocumentationHandler` import at line 568 both go away.

autogpt_platform/backend/backend/api/features/search/hybrid_search.py
```diff
@@ -0,0 +1,534 @@
...
+"""
+
+import logging
+import re
+import time
+from dataclasses import dataclass
+from typing import Any
+
+from prisma.enums import ContentType
+from rank_bm25 import BM25Okapi  # type: ignore[import-untyped]
```
majdyz: 🤖 🟠 **Should Fix**: `# type: ignore[import-untyped]` violates `backend/AGENTS.md`'s rule ("No linter suppressors — no `# type: ignore`, `# noqa`, `# pyright: ignore`; fix the type/code").

Fix at the config level instead: add `rank_bm25` to the project's missing-stubs allowlist in `pyproject.toml` (e.g. a `[[tool.pyright.executionEnvironments]]` override or a `[tool.mypy.overrides]` block with `ignore_missing_imports = true`). Then the import line stays clean.

autogpt_platform/backend/backend/api/features/search/service.py
```diff
@@ -0,0 +1,336 @@
...
+    user_id: str, limit: int, query: str | None = None
+) -> list[SearchResultItem]:
+    """Workspace files, filtered by ``query`` substring when provided.
+
+    Uses a direct ``ILIKE`` filter on ``name`` instead of the embedding
+    index so newly-written files are findable immediately. Workspace
+    file embeddings only encode the name anyway, so we lose nothing in
+    quality and gain freshness.
+    """
+    from backend.data.workspace import get_workspace
```
majdyz: 🤖 🟠 **Should Fix**: Four inline imports in this file:

- L164 `from backend.api.features.library import model as library_model` — the comment justifies it (avoiding a heavy graph chain at module import). Acceptable per `AGENTS.md` ("lazy imports only for heavy optional deps").
- L206-207 `from backend.data.workspace import get_workspace`, `from backend.util.workspace import WorkspaceManager` — neither is heavy.
- L253 `from backend.copilot.model import get_user_sessions` — also not heavy; copilot/model is already on the import graph via `chat_session_embeddings`.

Move the three at L206-207 and L253 to top-level. If there's a circular-import reason, name it in a comment so the next reader doesn't move them back. (`AGENTS.md`: "Top-level imports only — no local/inner imports.")

autogpt_platform/backend/backend/api/features/search/routes.py
```diff
@@ -0,0 +1,56 @@
...
+    q: str = Query(
```
majdyz, line 42: 🤖 🔵 **Nit**: `q` has no `max_length`. A 100KB query string would still flow into `embed_query` (paid OpenAI call) and into the BM25 reranker's tokenizer. Bound it — e.g. `Query("", max_length=512)` — so the upper limit is enforced at the API boundary, not downstream.

autogpt_platform/backend/backend/api/features/search/service.py
Abhi1992002: Fixed in 382be7f08c. Switched to a module-level `pydantic.TypeAdapter(ContentType)` and a matching `TypeAdapter(dict[str, Any])` for metadata. Both `ContentType` enum members and bare `.value` strings coerce through the adapter into a typed enum at the row boundary, and the dispatch below now branches on `content_type is ContentType.X` — no `getattr`/`isinstance`. The `metadata if isinstance(metadata, dict) else {}` guard at the old L101 dropped out for the same reason. Returns `None` (skip row) on validation failure rather than letting bad data crash downstream.

autogpt_platform/backend/backend/api/features/search/routes.py
Abhi1992002: Fixed in 382be7f08c. Added `backend/api/features/search/rate_limit.py` with a per-user fixed-window QPS limiter (120 req / 60s) and wired it into `routes.py` on the non-empty-query branch only — the cached-recents path doesn't hit OpenAI so it stays uncapped. Failure mode is fail-open on Redis brown-out: a brief burst of one user's embedding spend is less bad than blocking search for everyone if Redis blips.

Chose not to reuse `copilot.rate_limit.check_rate_limit` because that one's a USD-cost limiter scoped to paid LLM turns — search-as-you-type wants a QPS cap, not a spend cap.

autogpt_platform/backend/backend/api/features/search/embeddings.py
Abhi1992002: Fixed in 382be7f08c. Took the route you suggested — added an abstract `ContentHandler.get_valid_content_ids()` and implemented it on all six handlers. `cleanup_orphaned_embeddings` now collapses from an 80-line if/elif chain to one polymorphic call:

```python
current_ids = await handler.get_valid_content_ids()
```

The `isinstance(doc_handler, DocumentationHandler)` runtime narrowing and the inline `DocumentationHandler` import are both gone. Each handler's notion of "what counts as valid" now lives next to its `get_missing_items` filter, so the orphan-cleaner and the embedder can't drift apart on future schema changes. Side benefit: the unused `get_blocks` import drops out of `embeddings.py` too.

autogpt_platform/backend/backend/api/features/search/hybrid_search.py
Abhi1992002: Fixed in 382be7f08c. Dropped the `# type: ignore[import-untyped]` outright — pyright (the active checker per `scripts/linter.py`) is clean without it. Verified with `poetry run pyright backend/api/features/search/hybrid_search.py` → `0 errors, 0 warnings`. The suppressor was defensive carry-over from a project that ran mypy; no `pyproject.toml` allowlist needed.

autogpt_platform/backend/backend/api/features/search/service.py
Abhi1992002: Fixed in 382be7f08c. Hoisted `get_workspace`, `WorkspaceManager`, and `get_user_sessions` to top-level imports as suggested. `library_model` stays inline with the existing comment — it pulls in the graph chain at module-import time, matching the AGENTS.md allowance for heavy optional deps like `openpyxl`.

autogpt_platform/backend/backend/api/features/search/routes.py
Abhi1992002: Fixed in 382be7f08c. Added `max_length=MAX_QUERY_LENGTH` (512) to the `q` `Query()` declaration in `routes.py` so the bound is enforced at the FastAPI boundary before the value ever reaches `embed_query` or BM25. 512 chars comfortably covers the human-typeable range; anything internal that needs more should go through a dedicated feature endpoint.

autogpt_platform/backend/backend/api/rest_api.py
```diff
@@ -423,6 +424,11 @@ async def validation_error_handler(
     tags=["push"],
     prefix="/api/push",
 )
+app.include_router(
+    search_routes.router,
+    tags=["v2", "search"],
```
Pwuts: Our internal API isn't versioned, so:
```suggestion
    tags=["search"],
```

autogpt_platform/backend/backend/api/features/search/hybrid_search.py
```diff
@@ -0,0 +1,534 @@
...
+    content_types: list[ContentType] | None = None,
+    category: str | None = None,
+    page: int = 1,
+    page_size: int = 20,
+    weights: UnifiedSearchWeights | None = None,
+    min_score: float | None = None,
+    user_id: str | None = None,
+    lexical_query: str | None = None,
+    prefix_match: bool = False,
+) -> tuple[list[dict[str, Any]], int]:
```
Pwuts: could we make this return type more specific than `dict[str, Any]`?

autogpt_platform/frontend/src/app/(platform)/copilot/components/ChatSidebar/ChatSidebar.tsx
```diff
@@ -157,6 +164,41 @@ export function ChatSidebar() {
...
+      const metadata = (item.metadata ?? {}) as {
+        creator?: string;
+        slug?: string;
+      };
+      if (metadata.creator && metadata.slug) {
+        router.push(
+          `/marketplace/agent/${encodeURIComponent(metadata.creator)}/${encodeURIComponent(metadata.slug)}`,
+        );
+      }
+      return;
```
majdyz, line 188: 🤖 🔴 **Blocker**: Clicking a store_agent search result silently does nothing.

Failure scenario:
1. User Cmd-Ks the modal and types a query that returns store agents (e.g. "youtube").
2. Backend `_hybrid_row_to_item` in `service.py:95-99` populates `metadata = {"name": ..., "categories": [...]}` only — the `STORE_AGENT` `ContentItem` written by `StoreAgentHandler.get_missing_items` (`content_handlers.py:140-143`) and the row metadata read back via the embedding row never include `creator` or `slug`.
3. This handler reaches `if (metadata.creator && metadata.slug)`, the condition is always false, and the function falls through to the bare `return` on L188 — modal closes-by-Esc/backdrop but no nav happened. The user sees the modal still open after click (because `onClose` only runs from `handleSelect` after a successful `onSelectItem` call) and the store agent is unreachable from search.

Repro is straightforward: backend never embeds `creator`/`slug` into `UnifiedContentEmbedding.metadata` for STORE_AGENT, so any query returning store agents is dead-clickable.

Two ways to fix:
- Enrich `StoreAgentHandler` metadata with `creator` + `slug` from `StoreListing` (and add them to the SQL select). Then store-agent recents/hybrid both carry the URL fields.
- Or have the click handler fall back to `router.push('/marketplace/agent/' + item.id)` — but the marketplace route is keyed by `creator/slug`, not listing id, so this isn't a real fallback.

Question: was the intended design to add creator/slug to STORE_AGENT metadata, or was store-agent navigation deferred? If deferred, the bucket should be hidden in the UI rather than rendered as inert rows.

autogpt_platform/frontend/src/components/organisms/SearchCommandModal/SearchCommandModal.tsx
```diff
@@ -0,0 +1,230 @@
...
+  const showEmptyState = !isError && totalCount === 0 && !isLoading;
+
+  return (
+    <div
+      className="fixed inset-0 z-[80] flex items-start justify-center bg-black/20 px-4 pt-[18vh] backdrop-blur-sm"
+      onMouseDown={onClose}
+    >
+      <div
+        role="dialog"
+        aria-modal="true"
```
majdyz: 🤖 🟠 **Should Fix**: The modal is `role="dialog"` + `aria-modal="true"` but it's a hand-rolled overlay, not the design-system `Dialog` / Radix primitive — so the usual modal contract isn't enforced.

Failure scenarios:
1. **Tab escapes the modal.** User opens Cmd-K, the input auto-focuses (L77-82). User presses Tab: focus moves to the close-`X` button, then keeps tabbing into the underlying page (sidebar nav, message list, …) because there's no focus trap. The dialog is still visually present so the user has no idea they're typing into the page below.
2. **Background scroll bleeds through.** With the modal open, scrolling inside the result list past the top/bottom edge keeps scrolling the underlying page (no `overflow: hidden` on body). Backdrop scroll-isolation is a baseline expectation for Cmd-K-style overlays.
3. **Focus is not returned on close.** Closing via Escape/backdrop doesn't restore focus to the trigger element (the Cmd-K button / keyboard shortcut origin). Screen-reader users land at `document.body` and have to re-find their place.

`@radix-ui/react-dialog` (which the design-system `Dialog` already wraps) handles all three for free: focus trap, body scroll lock, return-focus on close, Escape, and inert backdrop. The current `<div role="dialog">` only handles Escape/Arrow-keys via the captured `handleKeyDown` and does the click-outside via `onMouseDown`.

Question: any reason this can't sit inside the existing `Dialog` primitive (with a custom `DialogContent` for the rounded-3xl styling)? That removes the bespoke focus/scroll/restore code, and the keyboard nav layer (`useKeyboardNav`) stays the same.

autogpt_platform/backend/backend/api/features/search/service.py
```diff
@@ -0,0 +1,350 @@
...
+        if workspace is None:
+            # User has no workspace yet — nothing to list, no need to
+            # create one just for the recents view.
+            return []
+        manager = WorkspaceManager(user_id, workspace.id, session_id=None)
+        files = await manager.list_files(
+            limit=limit,
+            include_all_sessions=True,
+            name_contains=query or None,
+        )
```
majdyz, line 326: 🤖 🟠 **Should Fix**: Files bucket isn't ranked by name relevance — it's "most recent file whose name contains the query".

Failure scenario:
1. User has 50 workspace files. Twenty of them contain "report" in the name (different sessions/folders); the most recently created 4 are all named `final-report-v3.pdf` / `report-2026-Q1.pdf` / `monthly-report.xlsx` (created today) and the file the user actually wants is `quarterly-report-overview.md` (created last week, the literal best title match).
2. User types `quarterly` into Cmd-K.
3. `_files_bucket` in `service.py:232-236` calls `manager.list_files(name_contains=query, limit=4, ...)` which in `data/workspace.py:282-287` orders by `createdAt: desc` — so the four newest files containing `"quarterly"` are returned. If twenty files have the substring, the 4 returned are the 4 newest, not the 4 most relevant. `quarterly-report-overview.md` (older) is invisible past page 1, and `_files_bucket` only ever reads page 1.
4. Even worse for short prefixes ("re" matches ~every file with "report"/"replace"/…): the user gets a near-random 4 of the matching files based on creation date.

Compare with the agents bucket, which goes through `unified_hybrid_search` and ranks by score. The PR description even calls the file bucket a "literal title match" — but a literal match without a relevance ordering makes the limit-of-4 trim arbitrary.

Two options:
- Order by a literal relevance signal at the DB layer: e.g. exact match first, then `startswith`, then `contains`, then `createdAt` as tiebreaker.
- Or accept the freshness bias but bump the per-bucket page size (e.g. 16) and rank in Python with the same BM25 tokenizer used in `hybrid_search.bm25_rerank`.

Question: was the freshness ordering intentional ("show me my recent files matching X")? If yes, worth documenting in `_files_bucket` so future callers don't expect relevance order. Same issue applies to `_chats_bucket` (orders by `updatedAt: desc`).

autogpt_platform/backend/backend/api/features/workspace/embeddings.py
```diff
@@ -0,0 +1,79 @@
...
+
+logger = logging.getLogger(__name__)
+
+# See ``library/embeddings.py`` for why we hold a strong ref to the task.
+_background_tasks: set[asyncio.Task[None]] = set()
+
+
+async def _run_embedding(file_id: str, user_id: str, name: str, path: str) -> None:
+    try:
+        searchable_text = WorkspaceFileHandler._build_text(name, path)
```
majdyz: 🤖 🟡 **Nice to Have**: Reaching into another module's private `_build_text` static method.

`WorkspaceFileHandler._build_text` is name-mangled-by-convention (`_`-prefixed) — it's a private implementation detail of `WorkspaceFileHandler.get_missing_items`. Calling it from `workspace/embeddings.py` couples this module to that internal helper; if `WorkspaceFileHandler` ever changes its text-building strategy (e.g. inlines it or adds dependencies), this caller silently breaks the per-write embedding text-vs-backfill text invariant — they'd encode the same file under two different `searchableText` representations and `existing.get("searchableText") == searchable_text` on L43 of `chat_session_embeddings.py`'s pattern would always miss.

Promote `_build_text` to a public module-level helper in `content_handlers.py` (e.g. `def build_workspace_file_text(name: str, path: str) -> str:`) and have both `WorkspaceFileHandler.get_missing_items` and the per-write indexer call it. Same contract, no private-API reach.

Question: any reason to keep it as a private static method on the class rather than a public helper? The text-building logic isn't class-state-dependent.

autogpt_platform/backend/backend/api/features/search/rate_limit.py
```diff
@@ -0,0 +1,74 @@
...
+    now = datetime.now(UTC)
+    key = _window_key(user_id, now=now)
+    try:
+        redis = await get_redis_async()
+        # INCR then EXPIRE only when this is the first hit. Atomicity
+        # between the two commands isn't critical — at worst a key has
+        # no TTL for a few ms and gets a fresh one on the next hit.
+        count = await redis.incr(key)
+        if count == 1:
+            await redis.expire(key, GLOBAL_SEARCH_WINDOW_SECONDS)
```
majdyz: 🤖 🟡 **Nice to Have**: `INCR` + conditional `EXPIRE` can leak TTL-less keys on Redis hiccups.

Failure scenario:
1. First request in a fresh minute window: `redis.incr(key)` returns `1` on L55.
2. The async `redis.expire(key, GLOBAL_SEARCH_WINDOW_SECONDS)` on L57 hits a transient network error / connection drop (the exact "Redis brown-out" case the docstring acknowledges) — but the `except` is only outside this whole block, so the EXPIRE failure raises and is caught at L58, where the function fails-open and returns.
3. The key now exists in Redis with no TTL. All subsequent requests in the same window correctly INCR it, but the key never expires.
4. A minute later, when the window key normally rolls over (`_window_key` uses `bucket = int(now.timestamp()) // 60`), the new bucket starts at 0 — that's fine. But the orphan key from step 1 lives forever (until Redis is restarted or `maxmemory` policy evicts it).

The fix is to use `SET key value EX 60 NX` first (atomic create-with-TTL), and only `INCR` afterward — or use a Lua script for atomic INCR+EXPIRE-if-first. `redis.set(key, 0, ex=GLOBAL_SEARCH_WINDOW_SECONDS, nx=True)` then `redis.incr(key)` is the simplest two-call version that survives EXPIRE failure.

Low-priority because per-user keys don't grow unbounded and Redis memory pressure is the only real risk, but it's a known anti-pattern that's worth fixing once.

autogpt_platform/backend/backend/api/rest_api.py
Abhi1992002: Done in f4329b221d — `tags=["v2", "search"]` → `tags=["search"]`.

autogpt_platform/backend/backend/api/features/search/hybrid_search.py
Abhi1992002: Done in f4329b221d. Added a `HybridSearchRow` `TypedDict` describing the row shape (content_type, content_id, searchable_text, metadata, updated_at, the four score components, combined_score, the window-side `total_count`, plus the bm25-rerank additions). `bm25_rerank` and the return type both use it now, and `service._hybrid_row_to_item` consumes it directly. Pyright clean.

autogpt_platform/frontend/src/app/(platform)/copilot/components/ChatSidebar/ChatSidebar.tsx
Abhi1992002: Fixed in f4329b221d — and yes, the intended design was creator/slug in STORE_AGENT metadata (not deferred). Two-pronged fix:

1. **Forward**: `StoreAgentHandler.get_missing_items` now joins to `StoreListing` for `slug` and `Profile` for `username` (as `creator`) and writes both into the embedding row's metadata. New embeds carry the marketplace URL.
2. **Backward**: existing embeddings pre-date this metadata, so `service.py` adds `_enrich_store_agent_metadata` that joins back on the rendered items when `creator`/`slug` are missing. One batched SQL roundtrip per search, no-op when every row is already enriched. The FE click handler in ChatSidebar.tsx is unchanged — it already gates on `metadata.creator && metadata.slug` and will now find them populated.

Failure-isolated: enrichment in a try/except so a click-routing miss can never 500 the whole bucket.

autogpt_platform/frontend/src/components/organisms/SearchCommandModal/SearchCommandModal.tsx
Abhi1992002: Done in f4329b221d. Migrated `SearchCommandModal` to `@radix-ui/react-dialog` primitives directly (Root / Portal / Overlay / Content / Title) — the molecule `Dialog` wraps the same primitives but its chrome (title bar + corner close button + padded scroll container) is wrong for a Cmd-K palette, so using the primitives at the organism layer keeps the layout while still getting focus trap + body scroll lock + return-focus on close for free.

- Tab no longer escapes (Radix's focus scope).
- Background scroll no longer bleeds through (Radix's `data-aria-hidden`-driven scroll lock).
- Escape close moves from the local `handleKeyDown` to Radix's `onOpenChange`; arrow keys + Enter stay in the existing `useKeyboardNav` layer.
- Input still autofocuses via `onOpenAutoFocus` (Radix defaults to the first focusable, which would be the X button).

Tests updated for the Radix-driven Escape path.

autogpt_platform/backend/backend/api/features/search/service.py
Abhi1992002: Fixed in f4329b221d. Went with option 2 (over-fetch + Python rerank) since the in-memory sort is cheap and avoids the Prisma raw-SQL fork in the DB layer.

- `_files_bucket` / `_chats_bucket` now over-fetch up to `_RELEVANCE_OVERFETCH_CAP` (32) candidates when `query` is non-empty.
- A three-tier literal-match score ranks them: exact case-insensitive (3) > startswith (2) > contains (1).
- Python's stable sort preserves `createdAt`/`updatedAt` DESC as the tiebreaker within a relevance bucket — so two same-rank matches still surface newest-first.
- Empty-query recents path is intentionally untouched (still pure `updatedAt DESC`); the docstrings call this out.

That covers your `quarterly-report-overview.md` repro: `quarterly` matches it as `startswith` (rank 2), beating any contains-only matches even if they're newer.

autogpt_platform/backend/backend/api/features/workspace/embeddings.py
Abhi1992002: Done in f4329b221d. Promoted `WorkspaceFileHandler._build_text` to a module-level `build_workspace_file_text(name, path)` in `content_handlers.py`. Both `WorkspaceFileHandler.get_missing_items` and the per-write indexer in `workspace/embeddings.py` now route through it — no more private-API reach, and the invariant the docstring describes is now structurally guaranteed.

autogpt_platform/backend/backend/api/features/search/rate_limit.py
Abhi1992002: Fixed in f4329b221d. `INCR` + conditional `EXPIRE` → atomic `SET key 0 EX <window> NX` then `INCR`. The TTL is now part of the same write that creates the key, so a brown-out on the second call can't strand a TTL-less key (your scenario from step 2). Subsequent hits inside the window no-op the SET (NX) and INCR the counter as before.

Test updated to assert `SET` is called with `ex=GLOBAL_SEARCH_WINDOW_SECONDS` + `nx=True` on every hit.

majdyz:
## 🧪 E2E Test Report

**PR:** #13217 — `feat(platform): global search endpoint + bucketed command-palette modal`
**Branch:** `abhimanyuyadav/secrt-2363-chat-search-modal-backend-server-side-search-pagination-2` (tested at PR tip `0d34a6ae25` on a detached HEAD because the named branch is held by an orchestrator-locked worktree)
**Mode:** Native (poetry run app + pnpm dev, infra via docker compose deps)
**Worktree:** `/Users/majdyz/Code/AutoGPT6`

### Scope of change (large — 66 files, +6035/-2709)
- **Backend:** new `backend/api/features/search/` module — generic embedding + hybrid (semantic + lexical + BM25) engine, pluggable per-`ContentType` handlers, `/api/search/global` endpoint, BM25 reranking, per-user Redis TTL cache on empty-query path, per-user QPS rate limit on non-empty queries. New per-feature indexers for workspace files and chat sessions. `store/embeddings.py` + `store/hybrid_search.py` collapsed into thin wrappers. Two new migrations: enum-value add (`WORKSPACE_FILE`, `CHAT_SESSION`) + tightened user-scoped `CHECK` constraint.
- **Frontend:** generic `SearchCommandModal` organism + thin `GlobalSearchModal` container. Chat sidebar's Cmd/Ctrl-K opens the new modal. Desktop `ChatSearchModal` deleted; mobile drawer's `ChatSearchResults` kept.

### Scenarios
| # | Scenario | Method | Expected | Actual | Result |
|---|----------|--------|----------|--------|--------|
| 1 | Search-feature unit tests (excluding e2e/integration suites) | `poetry run pytest` | All pass | 79 passed | PASS |
| 2 | Migrations apply cleanly; `ContentType` enum gains `WORKSPACE_FILE` + `CHAT_SESSION` | `prisma migrate deploy` + DB query | both values present in `pg_enum` | 7 enum values incl. both new ones | PASS |
| 3 | Tightened CHECK constraint covers all per-user types | DB query for constraint def | constraint mentions `LIBRARY_AGENT`, `WORKSPACE_FILE`, `CHAT_SESSION` | `CHECK (… <> ALL ARRAY['LIBRARY_AGENT','WORKSPACE_FILE','CHAT_SESSION'] OR userId IS NOT NULL)` | PASS |
| 4 | `GET /api/search/global` empty-query returns recent items bucketed | curl | `{chats, agents, files}` keys | 3 chats, 3 agents, 0 files | PASS |
| 5 | `GET /api/search/global?q=agent` returns bucketed hybrid results | curl | bucketed item rows with `id, type, title, subtitle, metadata, score, updated_at` | Returns `store_agent` rows with `score ≈ 0.211…` and proper shape | PASS |
| 6 | Embeddings populated for every `ContentType` | DB query | non-zero rows per type incl. new ones | STORE_AGENT 86, BLOCK 416, DOCUMENTATION 4743, LIBRARY_AGENT 50, WORKSPACE_FILE 2, CHAT_SESSION 249 | PASS |
| 7 | Workspace-file upload triggers indexer; new file is searchable | upload md + search `q=sardis` | files bucket contains the upload | files bucket returns `sardis-spec-73405.md` with type `workspace_file` | PASS |
| 8 | Chat-session indexer auto-runs on session creation; session is searchable | search `q=marketing+competitor` | chats bucket contains the test session | chats bucket returns "Marketing Competitor Research Agent" with type `chat_session` | PASS |
| 9 | Cmd+K from `/copilot` opens the bucketed modal | UI keydown via `dispatchEvent` | `dialog "Search"` with `CHATS`/`AGENTS`/`FILES` buckets | Modal opens, all three buckets render when populated | PASS |
| 10 | Typing fires `/api/search/global` with `q`; results re-bucket live | UI type `minimal` | filtered buckets | 1 chat + 2 agents (Minimal Passthrough / Minimal Text Processor) | PASS |
| 11 | Arrow-Down walks the flat list across buckets | dispatch `ArrowDown` 3x | selection lands on 3rd item | Selection moved from chat-bucket item → 3rd agent-bucket item | PASS |
| 12 | Enter on chat-session result navigates to `/copilot?sessionId=…` | dispatch `Enter` | route to chat | `/copilot?sessionId=4e49ed66-…` | PASS |
| 13 | Enter on library-agent result navigates to `/library/agents/{id}` | dispatch `Enter` | route to agent | `/library/agents/50dfc991-…?activeTab=runs` | PASS |
| 14 | "Clear search" X button empties input | click X | input value becomes `""` | input.value: `"minimal"` → `""` | PASS |
| 15 | Escape closes the modal | dispatch `Escape` | dialog disappears | Search dialog absent from snapshot | PASS |
| 16 | "No results found" empty state | type `xyznoresult12345` | empty-state text shown | StaticText "No results found" + no `option` rows | PASS |
| 17 | Old desktop `ChatSearchModal.tsx` removed; mobile `ChatSearchResults`/`useChatSearch`/`helpers` preserved | directory listing | only mobile components remain | Confirmed (no `ChatSearchModal.tsx`) | PASS |
| 18 | **TTL cache on empty-query path** — Redis key created, ~60s TTL, identical responses across hits, freshly-uploaded file does NOT leak into cached snapshot, invalidation re-fetches | hit endpoint, inspect `cache:_cached_recent_buckets:<hash>` on cluster, upload new file, re-hit, then `DEL` and re-hit | key exists with TTL 1-60; responses byte-identical while cached; new file absent during cache hold; new file appears after explicit invalidate | Key `cache:_cached_recent_buckets:47114a…` with TTL=44s; response 1 == response 2 (cache hit confirmed); new file `cache-test-76569.md` absent from cached body; `DEL` + retry: files bucket now contains `cache-test-76569.md` + `sardis-spec-73405.md` | PASS |
| 19 | **Per-user QPS rate limit** (120 req/60s per user) on non-empty queries | pre-set counter to 119 in current bucket, fire 5 requests | 200=1 (counter→120, at cap), 429=4 (over cap) | `200=1, 429=4`; final counter `124` | PASS |
| 20 | **All 6 Storybook stories render** (`with-results`, `loading`, `empty`, `no-matches`, `single-bucket`, `error-state`) | open each via story URL, inspect iframe body text | each renders own canonical text (no "Couldn't find story…") | All 6 rendered correctly; `with-results` shows seeded CHATS/AGENTS, `loading` shows nav legend, `empty` shows "No recent items", `no-matches` shows "No results found", `single-bucket` shows the seeded chats-only bucket, `error-state` shows "Something went wrong. Try again." | PASS |

### Bugs found
None.

### Notes / observations
- Cmd+K only binds inside `ChatSidebar`, which is mounted on `/copilot` (`CopilotPage`). On `/library/agents/{id}` the shortcut is a no-op. Matches the PR description; flagging in case reviewers expected the shortcut to be global.
- Workspace-file upload `POST /api/workspace/files/upload` errors on path collision (`"File already exists at path: /searchtest-content.md"`). Pre-existing behaviour, out of scope here.
- Local supabase-db volume was lagging the worktree's migrations by 4 — required an explicit `prisma migrate deploy` before the live tests. Not a PR bug; sibling-worktree shared-DB drift.

### Verdict
**APPROVE.** Every claim in the PR description is verified live: hybrid search shape, per-content-type indexers, the new modal's UX (Cmd+K, debounced query, cross-bucket arrow nav, Enter routing per type, Clear/Esc/"No results"), the Redis-backed TTL cache on the empty-query path (with both a cache-hit assertion and a cache-invalidate-reveals-new-data assertion), the per-user QPS rate limit, and all six Storybook stories. 79 search-feature unit tests pass; both migrations apply idempotently; the CHECK constraint covers all three per-user content types; the desktop `ChatSearchModal` is removed without breaking the mobile drawer.

### Screenshots










































_Posted by majdyz local /pr-test._

majdyz:
@Pwuts @ntindle — pushed follow-ups addressing the rerank-related blockers:
- Added BM25 rerank tests covering empty results, single-bucket, and mixed-type ranking paths.
- Tightened the per-user TTL cache invalidation on the empty-query path so newly-indexed items show up immediately.
- Threads resolved; ready for another look when you have time.

majdyz (approved):

PR #13235, opened 2026-05-28
fix(frontend/library): align Scheduled filter with fleet summary

0ubbe:
### Why / What / How

**Why:** Reported in [Discord thread "Inconsistencies in Filtering Scheduled"](https://discord.com/channels/1126875755960336515/1506937071313485937) by Toran. The Library "Scheduled" filter is inconsistent with the rest of the page:

> 1. Agents show up as in the "Scheduled" filter which aren't scheduled (even though the count says 0?)
> 2. Agents which *are* scheduled *don't* show up in the scheduled filter.

The briefing pill count, the status badge on each card, and the briefing tab list all categorize an agent as scheduled when **`is_scheduled` OR `recommended_schedule_cron`** is truthy. The library filter only checked `recommended_schedule_cron`, so agents with an active schedule but no recommended cron silently dropped out of the Scheduled bucket (and incorrectly landed in Idle).

**What:** Align `filterAgentsByStatus` in `useLibraryAgentList.ts` with the rest of the categorization logic, and add an integration test that asserts the alignment.

**How:** Compute `isScheduled = !!agent.is_scheduled || !!agent.recommended_schedule_cron` once and use it for both the `scheduled` and `idle` branches. This mirrors:

- `useLibraryFleetSummary.ts:114` (briefing pill count)
- `useAgentStatus.ts:67` / `:199` (status badge + summary)
- `useSitrepItems.ts:161` (briefing "Scheduled" tab list)

### Changes 🏗️

- `useLibraryAgentList.ts` — `filterAgentsByStatus` now uses `is_scheduled || recommended_schedule_cron` for the `scheduled`/`idle` branches.
- `library/__tests__/filter.test.tsx` — new integration test covering: (1) `is_scheduled=true && recommended_schedule_cron=null` agents appear under Scheduled, (2) idle agents stay out of Scheduled, (3) `is_scheduled=true` agents stay out of Idle.

#### For code changes:
- [x] I have made a test plan
  - [x] `pnpm test:unit` — 3 new tests pass; existing library tests (110) all pass
  - [x] `pnpm types`, `pnpm format`, `pnpm lint` all clean
  - [ ] Manual: in the Library page, with an agent that has an active schedule (no recommended cron), select the **Scheduled** filter — it should appear. With an agent that has only a recommended cron and no schedule, it should also appear (matches the badge/count). Neither should appear in **Idle**.


autogpt_platform/frontend/src/app/(platform)/library/components/LibraryAgentList/useLibraryAgentList.ts
```diff
@@ -340,6 +341,11 @@ function filterAgentsByStatus<
   return agents.filter((agent) => {
     const isRunning = activeGraphIds.has(agent.graph_id);
     const hasError = errorGraphIds.has(agent.graph_id);
+    // Match the categorization used by useLibraryFleetSummary, useAgentStatus,
+    // and useSitrepItems: an agent is "scheduled" if it has an active schedule
+    // (`is_scheduled`) OR a recommended cron the user can promote into one.
```
majdyz: 🤖 🟡 **Nice to Have**: This is the fourth place in the library that re-derives `is_scheduled || recommended_schedule_cron` — alongside `hooks/useLibraryFleetSummary.ts:114`, `hooks/useAgentStatus.ts:67` / `:199`, and `components/SitrepItem/useSitrepItems.ts:161`. The PR description correctly notes this is the bug class: any new categorization site that re-derives the predicate inline can drift again.

Failure scenario:
1. A future PR adds a fifth categorization site (e.g., a new `ScheduledAgentsPanel`) and writes `if (agent.recommended_schedule_cron)` from muscle memory.
2. `pnpm test:unit` passes because the new test set isn't aware of the predicate's two arms.
3. We regress the same "Inconsistencies in Filtering Scheduled" Discord report against a different surface.

Q: Want to extract `isAgentScheduled(agent)` (and maybe `isAgentIdle`) into a tiny helper next to one of these hooks (or `library/helpers.ts`) and call it from all four sites in this PR? Non-blocking — this fix is correct and minimal as-is, but the helper would make the next regression a type error instead of a behavior bug.

PR #13236, opened 2026-05-28
refactor(backend): plumb ExecutionContext at direct-block-execute boundary

18062706139fcz:
### Why / What / How

**Why**: Issue #12648 — `GetCurrentDateBlock.run()` crashes with `missing 1 required keyword-only argument: 'execution_context'` when invoked through the direct-block-execute REST routes. The original PR patched the three time blocks individually, but the same bug applies to ~40 other blocks (basic, text, http, agent, autopilot, claude_code, ai_image_*, orchestrator, spreadsheet, talking_head, video/*, screenshotone, io, human_in_the_loop, flux_kontext, code_executor, …) that declare `execution_context: ExecutionContext` as a required kwarg. The root cause is the API boundary, not the blocks.

**What**: Fix the bug at the boundary — make `Block.execute()` require `execution_context` explicitly, and have the direct-block-execute routes construct a real `ExecutionContext` (carrying the caller's `user_id` + `user_timezone`) before invoking the block.

**How**:

1. `Block.execute()` / `Block._execute()` now take `execution_context: ExecutionContext` as an explicit required keyword-only parameter. Missing-context callers now fail at the type system instead of as a `TypeError` deep inside a block's `run()`.
2. The internal route (`backend/api/features/v1.py::execute_graph_block`) and the external route (`backend/api/external/v1/routes.py::execute_graph_block`) both now build:

   ```python
   execution_context = ExecutionContext(
       user_id=user_id,
       user_timezone=get_user_timezone_or_utc(user.timezone),
   )
   ```

   and forward it to `obj.execute(...)`. Time blocks (and every other block that reads `execution_context.user_timezone`) now get the correct user timezone instead of crashing.
3. The HITL-review gate inside `Block._execute()` now checks `execution_context.graph_exec_id is not None` instead of probing kwarg presence — direct block execution has no graph and never needs review, which keeps the existing behaviour intact.
4. Reverted the per-block changes in `time_blocks.py`; the three time blocks once again declare `execution_context` as a required kwarg, matching every other block.

### Changes 🏗️

- `Block.execute()` / `Block._execute()`: `execution_context` is now an explicit, required keyword argument.
- Internal & external `POST /blocks/{id}/execute` routes: build a real `ExecutionContext` from the caller's user record and forward it to the block.
- Reverted per-block patches in `backend/blocks/time_blocks.py`.
- Regression tests:
  - `test_execute_graph_block_forwards_execution_context` (internal route)
  - `test_execute_block_forwards_execution_context` (external route)
  - Updated time-block tests to cover the real paths (empty-context vs user-timezone).
- Updated `backend/blocks/github/test_github_blocks.py` to pass `execution_context=ExecutionContext()` to its direct `block.execute()` calls.
- No configuration or environment changes.

#### For code changes:
- [x] I have made a test plan
  - [x] `poetry run pytest backend/blocks/test/test_time_blocks.py backend/blocks/github/test_github_blocks.py backend/api/features/v1_test.py backend/api/external/v1/routes_test.py`
  - [x] `poetry run lint --skip-pyright`


autogpt_platform/backend/backend/blocks/time_blocks.py
```diff
@@ -298,9 +298,15 @@ def __init__(self):
         )
 
     async def run(
-        self, input_data: Input, *, execution_context: ExecutionContext, **kwargs
+        self,
```
majdyz: 🤖 🟠 **Should Fix** — sibling block has the identical bug

Repro:
1. A caller invokes `GetCurrentTimeBlock` without supplying `execution_context` (the same path that hits #12648 for `GetCurrentDateBlock`).
2. At `backend/blocks/time_blocks.py:191`, `run()` still declares `execution_context: ExecutionContext` as a required kw-only arg.
3. Python raises `TypeError: run() missing 1 required keyword-only argument: 'execution_context'` — same crash this PR is meant to eliminate, just on a different block.

Question: why fix only `GetCurrentDateBlock` here? `GetCurrentTimeBlock` (line 191) and `GetCurrentDateAndTimeBlock` (line 407) are structurally identical and will fail the same way on the same call path — can the same `ExecutionContext | None = None` + ternary fallback be applied to both in this PR so #12648 doesn't immediately reopen for the sibling blocks?

autogpt_platform/backend/test/blocks/test_time_blocks.py
```diff
@@ -0,0 +1,18 @@
...
+async def test_get_current_date_without_execution_context():
+    block = GetCurrentDateBlock()
+    input_data = block.Input(trigger="Hello", offset="0")
+
+    results = [output async for output in block.run(input_data=input_data)]
+
+    assert len(results) == 1
+    key, value = results[0]
+    assert key == "date"
+    datetime.strptime(value, "%Y-%m-%d")
```
majdyz: 🤖 🟠 **Should Fix** — assertion is too weak to catch the regression you intend

Repro:
1. Imagine someone accidentally breaks the offset math so the block always returns `1970-01-01`.
2. `datetime.strptime(value, "%Y-%m-%d")` at `test/blocks/test_time_blocks.py:18` still succeeds (it only validates *format*, not *value*).
3. Test passes green; bug ships.

Question: can you tighten the assertion to also check the date is actually "today-ish", e.g.

```python
parsed = datetime.strptime(value, "%Y-%m-%d").date()
assert abs((datetime.now(tz=ZoneInfo("UTC")).date() - parsed).days) <= 1
```

so the test would actually fail if `current_date` regresses?

autogpt_platform/backend/test/blocks/test_time_blocks.py
```diff
@@ -0,0 +1,18 @@
+from datetime import datetime
+
+import pytest
+
+from backend.blocks.time_blocks import GetCurrentDateBlock
+
+
+@pytest.mark.asyncio
+async def test_get_current_date_without_execution_context():
```
majdyz: 🤖 🟡 **Nice to Have** — also cover the with-context path

This PR changes the signature at `backend/blocks/time_blocks.py:300-303` (kw-only arg becomes optional). The new test only exercises the `execution_context=None` branch — a future refactor that breaks `effective_timezone = execution_context.user_timezone` (e.g., dropping the conditional) would slip through CI.

Question: can you add a second test that constructs a minimal `ExecutionContext` with a known `user_timezone` (e.g. `"America/New_York"`), enables `use_user_timezone=True` on the format, and asserts the returned date matches the expected TZ? That locks in both branches of the ternary you introduced.

autogpt_platform/backend/test/blocks/test_time_blocks.py
```diff
@@ -0,0 +1,18 @@
+from datetime import datetime
```
majdyz: 🤖 🔵 **Nit** — test location

The rest of the backend block tests live under `backend/blocks/test/` (see `backend/blocks/test/test_block.py`, `test_autopilot.py`, `test_http.py`, etc.). This new file lives under `backend/test/blocks/` instead, which splits the discovery surface in two.

Question: any reason not to colocate this with the existing block tests at `autogpt_platform/backend/backend/blocks/test/test_time_blocks.py`? That also lets `poetry run pytest backend/blocks/test/` pick it up alongside the parametrized `test_available_blocks` suite.

autogpt_platform/backend/backend/blocks/time_blocks.py
18062706139fcz: 🤖 Fixed in [64db1aa31](https://github.com/Significant-Gravitas/AutoGPT/commit/64db1aa310d4ec994ce11f044ba36fd7d077a499): applied the same optional \ fallback to \ and \, so all three time blocks now handle the no-context path consistently.

autogpt_platform/backend/test/blocks/test_time_blocks.py
18062706139fcz: 🤖 Fixed in [64db1aa31](https://github.com/Significant-Gravitas/AutoGPT/commit/64db1aa310d4ec994ce11f044ba36fd7d077a499): moved the test into \ and strengthened the no-context assertion to validate the returned date is actually current, not just parseable.

autogpt_platform/backend/test/blocks/test_time_blocks.py
18062706139fcz: 🤖 Fixed in [64db1aa31](https://github.com/Significant-Gravitas/AutoGPT/commit/64db1aa310d4ec994ce11f044ba36fd7d077a499): added a with-context regression test using \ with \ to cover the timezone branch as well.

autogpt_platform/backend/test/blocks/test_time_blocks.py
18062706139fcz: 🤖 Fixed in [64db1aa31](https://github.com/Significant-Gravitas/AutoGPT/commit/64db1aa310d4ec994ce11f044ba36fd7d077a499): relocated the test file to \ so it sits with the existing backend block tests.

autogpt_platform/backend/backend/blocks/time_blocks.py
18062706139fcz: Clarifying the previous reply: fixed in [64db1aa31](https://github.com/Significant-Gravitas/AutoGPT/commit/64db1aa310d4ec994ce11f044ba36fd7d077a499) by making GetCurrentTimeBlock and GetCurrentDateAndTimeBlock accept a missing execution_context too, so the sibling time blocks now follow the same no-context behavior as GetCurrentDateBlock.

autogpt_platform/backend/test/blocks/test_time_blocks.py
18062706139fcz: Clarifying the previous reply: fixed in [64db1aa31](https://github.com/Significant-Gravitas/AutoGPT/commit/64db1aa310d4ec994ce11f044ba36fd7d077a499) by moving the test under backend/blocks/test/ and strengthening the assertion so it checks the returned date is current, not only parseable.

autogpt_platform/backend/test/blocks/test_time_blocks.py
18062706139fcz: Clarifying the previous reply: fixed in [64db1aa31](https://github.com/Significant-Gravitas/AutoGPT/commit/64db1aa310d4ec994ce11f044ba36fd7d077a499) by adding a with-context regression test using ExecutionContext(user_timezone="America/New_York") together with use_user_timezone=True, so the timezone branch is covered too.

autogpt_platform/backend/test/blocks/test_time_blocks.py
18062706139fcz: Clarifying the previous reply: fixed in [64db1aa31](https://github.com/Significant-Gravitas/AutoGPT/commit/64db1aa310d4ec994ce11f044ba36fd7d077a499) by relocating the test file to backend/blocks/test/test_time_blocks.py alongside the other backend block tests.

majdyz:
Hey @18062706139fcz — thanks for the contribution! The PR is otherwise in good shape but the CLA check is currently failing. Could you sign the CLA when you have a moment? It's a one-time step (the bot link should be in the failing check above) and unblocks the merge. Appreciated!

18062706139fcz:
> Hey @18062706139fcz — thanks for the contribution! The PR is otherwise in good shape but the CLA check is currently failing. Could you sign the CLA when you have a moment? It's a one-time step (the bot link should be in the failing check above) and unblocks the merge. Appreciated!

@majdyz  Thanks for the reminder! The CLA check didn’t go through earlier due to an account issue when I clicked the link, but it has been fixed now and the CLA should be signed successfully.

Could you please check again and see if the PR can be merged now? Thanks!

majdyz:
Hey @18062706139fcz — thanks for the PR. After digging in I think the fix as written was hitting the wrong layer, so I pushed a follow-up commit to the same branch (maintainer edits are on). Sharing the reasoning here so it's transparent:

**Why per-block was the wrong layer**

`GetCurrentDateBlock`/`GetCurrentTimeBlock`/`GetCurrentDateAndTimeBlock` are not unusual — they're three of ~40 blocks that declare `execution_context: ExecutionContext` as a required kwarg (e.g. `basic`, `text`, `http`, `agent`, `autopilot`, `claude_code`, `orchestrator`, `spreadsheet`, `talking_head`, `video/*`, `screenshotone`, `io`, `human_in_the_loop`, `flux_kontext`, `code_executor`, …). They all crash on the same path:

- [`backend/api/features/v1.py::execute_graph_block`](https://github.com/Significant-Gravitas/AutoGPT/blob/dev/autogpt_platform/backend/backend/api/features/v1.py) — `obj.execute(data, user_id=user_id)` — no `execution_context`
- [`backend/api/external/v1/routes.py::execute_graph_block`](https://github.com/Significant-Gravitas/AutoGPT/blob/dev/autogpt_platform/backend/backend/api/external/v1/routes.py) — `obj.execute(data)` — not even `user_id`

So patching the three time blocks fixes the headline bug in #12648 but leaves every other block crashing the same way through those two routes.

**What I changed**

1. Made `Block.execute()` / `Block._execute()` declare `execution_context: ExecutionContext` as an explicit required keyword-only parameter. Now missing-context callers fail at the type system, not as a `TypeError` deep inside a block's `run()` — and so this entire class of bug shows up at the boundary, not at runtime.
2. Both direct-block-execute routes now construct a real `ExecutionContext` from the caller's user record (with `user_id` + `user_timezone` from the user's profile) and forward it to `obj.execute(...)`. So time blocks actually use the user's timezone instead of falling back to UTC.
3. Gated the HITL-review path on `execution_context.graph_exec_id is not None` (it's only meaningful inside a graph execution), replacing the old `has_graph_context` kwarg-presence probe.
4. Reverted your per-block changes in `time_blocks.py` — they're no longer needed once the boundary is fixed.
5. Kept your tests but updated them to reflect the new contract; also added regression tests at both routes (`test_execute_graph_block_forwards_execution_context` + `test_execute_block_forwards_execution_context`) asserting they forward a real `ExecutionContext` carrying the caller's `user_id` + timezone.

PR title and description updated to reflect the new scope. Let me know if you'd rather I rolled this back and you take it from here — happy to do either. Thanks again for catching the underlying issue.

autogpt_platform/backend/backend/api/external/v1/routes.py
```diff
@@ -112,8 +114,20 @@ async def execute_graph_block(
             status_code=status.HTTP_402_PAYMENT_REQUIRED, detail=str(e)
         ) from e
 
+    user = await user_db.get_user_by_id(auth.user_id)
```
majdyz: 🤖 🟠 **Should Fix** — user lookup happens *after* charge

In the internal route ([v1.py L496-498](https://github.com/Significant-Gravitas/AutoGPT/blob/2c5d38c6e35d864c435ea1aedf2ad606b6ab3c23/autogpt_platform/backend/backend/api/features/v1.py#L496-L498)) we look up the user *before* charging, so a missing-user 404 short-circuits without touching credits. The external route ordering is reversed:

```python
# Charge happens at L108-115
await charge_for_direct_block_execution(...)
...
# But user lookup is at L117-119 — after the spend
user = await user_db.get_user_by_id(auth.user_id)
if not user:
    raise HTTPException(status_code=404, ...)
```

If `get_user_by_id` ever returns `None` (or raises), we've already spent credits without executing the block. Trivial fix — move the lookup above the `charge_for_direct_block_execution` block so the order matches the internal route.

autogpt_platform/backend/backend/blocks/test/test_time_blocks.py
```diff
@@ -0,0 +1,102 @@
...
+        ),
+    )
+    execution_context = ExecutionContext(user_timezone="America/New_York")
+
+    key, value = await _collect_single_output(
+        block.run(input_data=input_data, execution_context=execution_context)
+    )
+
+    assert key == "date"
+    assert value == datetime.now(tz=ZoneInfo("America/New_York")).strftime("%Y-%m-%d")
```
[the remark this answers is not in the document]
majdyz: 🤖 Fixed in [47d020996](https://github.com/Significant-Gravitas/AutoGPT/commit/47d020996d39a99e2c7c06d6e06cf0b7f558fa7c): wrap the call in before/after timestamps in the same timezone and assert `value in {before, after}` so a midnight rollover between captures doesn't fail the test.

autogpt_platform/backend/backend/api/external/v1/routes.py
majdyz: 🤖 Fixed in [47d020996](https://github.com/Significant-Gravitas/AutoGPT/commit/47d020996d39a99e2c7c06d6e06cf0b7f558fa7c): moved `user_db.get_user_by_id` + the `if not user: 404` short-circuit above `charge_for_direct_block_execution` so a missing-user 404 can't trigger after the charge.

majdyz:
🤖 @coderabbitai — the `routes.py` user-lookup-before-charge TOCTOU you flagged was already addressed in [47d020996](https://github.com/Significant-Gravitas/AutoGPT/commit/47d020996d39a99e2c7c06d6e06cf0b7f558fa7c) and is on the latest commit f8daedf7c. Crossed posts.

majdyz:
## ✅ /pr-test --fix — 7/7 PASS

E2E run on the dev environment against latest commit (worktree at PR head, native backend, `prisma migrate deploy` ran 4 new migrations, supabase + redis-cluster + rabbitmq up via docker compose):

| # | Scenario | Result |
|---|----------|--------|
| 1 | `GetCurrentDateBlock` via `POST /api/blocks/{id}/execute` (defaults) | PASS — HTTP 200, `{"date":["2026-05-29"]}` |
| 2 | `GetCurrentTimeBlock` via `POST /api/blocks/{id}/execute` (defaults) | PASS — HTTP 200, `{"time":["07:20:36"]}` (UTC) |
| 3 | `GetCurrentDateAndTimeBlock` via `POST /api/blocks/{id}/execute` (defaults) | PASS — HTTP 200, `{"date_time":["2026-05-29 07:20:36"]}` |
| 4 | `use_user_timezone=true` picks up NY tz from user profile | PASS — `03:21:00` (EDT) vs `07:21:00` (UTC reference) → 4h delta matches DST |
| 5 | `use_user_timezone=true` falls back to UTC when profile tz is `not-set` | PASS — `07:23:41 UTC` matches local clock after `cache_user_lookup` redis flush |
| 6 | Graph-based execution of `GetCurrentDateBlock` (regression check for `Block.execute()` signature change) | PASS — node `COMPLETED`, `node_error_count: 0`, output `"2026-05-29 03:26:02 EDT"` |
| 7 | Negative paths: unknown block_id → 404, missing auth → 401, re-execution after mutations → still 200 | PASS — no 500s |

**Issue #12648 reproduction check**: before this PR, `POST /api/blocks/{time-block-id}/execute` returned `500` with `TypeError: missing 1 required keyword-only argument: 'execution_context'`. After: returns `200` with the expected payload — verified live.

**Signature-change regression check (Scenario 6)**: the `Block.execute(..., *, execution_context: ExecutionContext)` required-kwarg change is correctly fed by the graph executor too — no regression on the normal graph path.

`--fix` mode had nothing to push.

majdyz (approved):

18062706139fcz:
> Hey @18062706139fcz — thanks for the PR. After digging in I think the fix as written was hitting the wrong layer, so I pushed a follow-up commit to the same branch (maintainer edits are on). Sharing the reasoning here so it's transparent:
> 
> **Why per-block was the wrong layer**
> 
> `GetCurrentDateBlock`/`GetCurrentTimeBlock`/`GetCurrentDateAndTimeBlock` are not unusual — they're three of ~40 blocks that declare `execution_context: ExecutionContext` as a required kwarg (e.g. `basic`, `text`, `http`, `agent`, `autopilot`, `claude_code`, `orchestrator`, `spreadsheet`, `talking_head`, `video/*`, `screenshotone`, `io`, `human_in_the_loop`, `flux_kontext`, `code_executor`, …). They all crash on the same path:
> 
> * [`backend/api/features/v1.py::execute_graph_block`](https://github.com/Significant-Gravitas/AutoGPT/blob/dev/autogpt_platform/backend/backend/api/features/v1.py) — `obj.execute(data, user_id=user_id)` — no `execution_context`
> * [`backend/api/external/v1/routes.py::execute_graph_block`](https://github.com/Significant-Gravitas/AutoGPT/blob/dev/autogpt_platform/backend/backend/api/external/v1/routes.py) — `obj.execute(data)` — not even `user_id`
> 
> So patching the three time blocks fixes the headline bug in #12648 but leaves every other block crashing the same way through those two routes.
> 
> **What I changed**
> 
> 1. Made `Block.execute()` / `Block._execute()` declare `execution_context: ExecutionContext` as an explicit required keyword-only parameter. Now missing-context callers fail at the type system, not as a `TypeError` deep inside a block's `run()` — and so this entire class of bug shows up at the boundary, not at runtime.
> 2. Both direct-block-execute routes now construct a real `ExecutionContext` from the caller's user record (with `user_id` + `user_timezone` from the user's profile) and forward it to `obj.execute(...)`. So time blocks actually use the user's timezone instead of falling back to UTC.
> 3. Gated the HITL-review path on `execution_context.graph_exec_id is not None` (it's only meaningful inside a graph execution), replacing the old `has_graph_context` kwarg-presence probe.
> 4. Reverted your per-block changes in `time_blocks.py` — they're no longer needed once the boundary is fixed.
> 5. Kept your tests but updated them to reflect the new contract; also added regression tests at both routes (`test_execute_graph_block_forwards_execution_context` + `test_execute_block_forwards_execution_context`) asserting they forward a real `ExecutionContext` carrying the caller's `user_id` + timezone.
> 
> PR title and description updated to reflect the new scope. Let me know if you'd rather I rolled this back and you take it from here — happy to do either. Thanks again for catching the underlying issue.

Thanks for the explanation. I understand the issue now.

My original change was fixing the symptom by making the time blocks tolerate a missing `execution_context`, but the real problem is that the direct block execution path was calling `Block.execute()` without constructing and passing an `ExecutionContext`.

Since many blocks rely on `execution_context`, not just the time blocks, the better fix is to handle this at the execution boundary and make `execution_context` part of the required execution contract.

I agree with your direction here. The direct execute routes should create the proper `ExecutionContext` from the current user and pass it into `Block.execute()`, instead of making individual blocks treat it as optional.

PR #13249, opened 2026-05-29
fix(backend): add OpenRouter aliases for Claude 4.6/4.7 models

### LINKED ISSUE #13005: Model ID mismatch in AITextSummarizerBlock for Claude 4.7 aliases
### ⚠️ Search for existing issues first ⚠️

- [x] I have searched the existing issues, and there is no existing issue for my problem

### Which Operating System are you using?

Windows

### Which version of AutoGPT are you using?

Latest Release

### What LLM Provider do you use?

Azure

### Which area covers your issue best?

Installation and setup

### What commit or version are you using?

Core Model: Gemini 3 Flash

### Describe your issue.

Following the release of v0.6.58, users attempting to use the new Claude Opus 4.7 support in the Agent Builder are seeing BlockUnknownError failures. While the model is selectable in the UI, the backend AITextSummarizerBlock and AIListGeneratorBlock are failing to map the internal alias to the provider ID correctly, resulting in a 400 - invalid model ID error from the Anthropic/OpenRouter proxy.

### Upload Activity Log Content

_No response_

### Upload Error Log Content

_No response_

goingforstudying-ctrl:
## Summary
Fixes model ID mismatch in AITextSummarizerBlock and AIListGeneratorBlock when using Claude 4.6/4.7 via OpenRouter.

## What
The `_OPENROUTER_ALIASES` mapping only contained Claude 4.5 models. When users selected Claude 4.6 or 4.7 in the Agent Builder, the blocks passed the bare model ID (`claude-opus-4-7`) to the Anthropic API via OpenRouter, resulting in a 400 "invalid model ID" error.

## How
- Added missing OpenRouter aliases for Claude 4.6/4.7 models to `_OPENROUTER_ALIASES`
- Updated comments to clarify why non-snapshot models also need alias entries (the provider dispatch path needs the reverse mapping, not just enum creation)
- Added `TestOpenRouterAliases` test class with 5 test cases covering the new aliases and regression protection for existing 4.5 aliases

## Changes
- `autogpt_platform/backend/backend/blocks/llm.py`: Added 3 new alias mappings
- `autogpt_platform/backend/backend/blocks/test/test_llm.py`: Added 5 tests

## Verification
- [x] Aliases resolve correctly via `LlmModel._missing_`
- [x] Existing 4.5 aliases remain functional
- [x] Invalid model IDs still raise `ValueError`

Closes #13005


goingforstudying-ctrl:
@coderabbitai Great catch! 🎯 You're absolutely right — the original fix was a no-op for dispatch.

**What I missed:**
-  only handles slug → enum construction in 
- But the actual provider dispatch at line 1212 sends  to OpenRouter
- OpenRouter needs the provider-prefixed slug (), not the bare value ()

**Fix applied:**
1. Added  dict (auto-built from ) for enum → slug mapping
2. Updated OpenRouter dispatch to use 
3. Added parametrised tests for both forward and reverse mappings
4. Removed the incorrect 4.6/4.7 entries from  (they're not needed there since  already handles prefix-stripping for enum creation)

The commit has been amended. Please re-review!

goingforstudying-ctrl:
@coderabbitai Thanks for catching that. You are right, the original fix did not actually change the dispatch path.

I was thinking about this wrong. The _OPENROUTER_ALIASES dict is only used in _missing_ for parsing incoming slugs into enum values. But when we actually send the request to OpenRouter, we use llm_model.value directly, which gives the bare ID without the anthropic/ prefix.

What I changed:
1. Added _OPENROUTER_REVERSE to map enum values back to provider-prefixed slugs
2. Updated the OpenRouter dispatch code to use _OPENROUTER_REVERSE.get(llm_model, llm_model.value) so it sends the prefixed slug when available
3. Kept the tests but added a second parametrised test for the reverse mapping
4. Removed the 4.6/4.7 entries from _OPENROUTER_ALIASES since they are not needed there

The commit has been amended with these changes. Let me know if you see any other issues.

goingforstudying-ctrl:
@coderabbitai You are right, I removed the 4.6/4.7 entries from _OPENROUTER_ALIASES which broke the reverse mapping. I was trying to avoid duplication but that was wrong thinking.

Added the three entries back. Now _OPENROUTER_REVERSE will correctly contain all six models through the dict comprehension, and the dispatch line will send the prefixed slug.

Thanks for the detailed explanation. The table in your comment made the problem very clear.

goingforstudying-ctrl:
Fixed both CI failures:

1. **Type check**: Changed  to  to avoid passing  to the OpenAI SDK model parameter.

2. **Lint**: Reformatted test function signatures to comply with black line-length rules.

Pushed as amended commit.

PR #13288, opened 2026-06-04
feat(platform): DataFast revenue attribution on Stripe Checkout

0ubbe:
### Why / What / How

**Why:** DataFast attributes paid conversions back to marketing sources by reading `datafast_visitor_id` / `datafast_session_id` from the Stripe Checkout Session `metadata`. Today our Checkout Sessions are built server-side and carry no DataFast IDs, so revenue from credit top-ups and subscriptions is not attributed.

**What:** Forward the two DataFast cookie IDs from the browser into the `metadata` of both *interactive* Stripe Checkout Sessions — credit top-ups (`mode="payment"`) and subscription checkouts (`mode="subscription"`). Off-session auto-recharge (PaymentIntent) is intentionally excluded; there is no browser/visitor context for it.

**How:** Centralized, transport-level plumbing rather than per-call-site edits (there are 6+ scattered checkout triggers but only 2 request choke-points):

1. **Frontend** reads the JS-readable DataFast cookies and maps them to `X-Datafast-Visitor-Id` / `X-Datafast-Session-Id` request headers (`getDatafastAttribution`). Best-effort: returns `{}` during SSR or when cookies are absent, and never throws.
2. Both client pipelines (the Orval custom mutator and the legacy `BackendAPI` client) merge those headers on client-side requests only.
3. The Next.js proxy allow-lists the two headers so they reach the backend.
4. **Backend** reads them via FastAPI `Header()` params on the top-up and subscription-tier endpoints and threads them into `top_up_intent` / `create_subscription_checkout`, which attach a `_datafast_metadata(...)` dict to the Checkout Session `metadata` (and merge into `subscription_data.metadata` for subscriptions, preserving `user_id` / `tier` / `billing_cycle`).

Best-effort at every layer: a missing cookie/header never blocks or errors a payment.

### Changes 🏗️

- **backend/data/credit.py**: `_datafast_metadata()` helper; `datafast_visitor_id` / `datafast_session_id` params on `top_up_intent` and `create_subscription_checkout`; metadata attached to the `payment` and `subscription` Checkout Sessions.
- **backend/api/features/v1.py**: `X-Datafast-*` `Header()` params on the top-up and subscription-tier endpoints, forwarded to the credit layer.
- **frontend/services/analytics/datafast-attribution.ts**: cookie→header helper.
- **frontend custom mutator + legacy client**: merge attribution headers on client-side requests.
- **frontend proxy route**: allow-list the two headers.
- Tests added at backend (pytest) and frontend (Vitest + RTL + MSW) levels covering the helper, header forwarding, proxy allow-listing, and the billing top-up flow.

#### For code changes:
- [x] I have made a test plan
  - [ ] `poetry run test` (backend) — DataFast metadata + header-forwarding tests
  - [ ] `pnpm test:unit` (frontend) — attribution helper, proxy allow-list, billing top-up header tests
  - [ ] Manual: with DataFast cookies set, start a top-up and a subscription checkout; confirm `datafast_visitor_id` / `datafast_session_id` appear on the resulting Stripe Checkout Session `metadata`
  - [ ] Manual: with cookies absent, confirm both checkouts still succeed (no metadata, no error)

> [!NOTE]
> Tests were written but **not executed locally** (toolchains not installed in the authoring environment). Correctness was verified via multi-stage code review only. Please rely on CI / a reviewer run for the gates (`poetry run test`, `pnpm test:unit`, lint, types).


autogpt_platform/backend/backend/api/features/v1.py
```diff
@@ -647,10 +648,18 @@ async def get_user_credits(
     dependencies=[Security(requires_user)],
 )
 async def request_top_up(
-    request: RequestTopUp, user_id: Annotated[str, Security(get_user_id)]
+    request: RequestTopUp,
+    user_id: Annotated[str, Security(get_user_id)],
+    x_datafast_visitor_id: Annotated[str | None, Header()] = None,
+    x_datafast_session_id: Annotated[str | None, Header()] = None,
```
0ubbe: 🤖 🔴 **Blocker**: These `Header()` params are emitted into the OpenAPI schema, which drifts the committed generated frontend client and breaks the `check API types` CI job (step "Check for API schema changes" fails). Since the whole design is *centralized header injection* (the FE attaches these headers in the mutator/proxy, never via generated hooks), these internal attribution headers should not be part of the public API contract at all. Recommend `Header(include_in_schema=False)` on all four params (`request_top_up` + `update_subscription_tier`) — this keeps the schema stable so no `pnpm generate:api` regeneration is needed, and matches the precedent at `chat/routes.py:1409` (`Query(..., include_in_schema=False)`). Alternatively, regenerate and commit the client.

autogpt_platform/backend/backend/data/credit.py
```diff
@@ -103,6 +103,23 @@ class InvoiceListItem(BaseModel):
...
+    Only includes IDs that are actually present so we never send the literal
+    string "None" to Stripe. Returns an empty dict when neither is set, which
+    Stripe accepts as "no metadata".
+    """
+    metadata: dict[str, str] = {}
+    if visitor_id:
+        metadata["datafast_visitor_id"] = visitor_id
+    if session_id:
+        metadata["datafast_session_id"] = session_id
+    return metadata
```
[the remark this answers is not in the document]
0ubbe: 🤖 Fixed in [6c4f57d70](https://github.com/Significant-Gravitas/AutoGPT/commit/6c4f57d70c6df8cce8c4fc6cf6fa05606ee77892): added `_sanitize_datafast_id()` — strips whitespace, drops blank/control-char values, and truncates to Stripe's 500-char metadata limit before building the metadata dict. A malformed client-supplied ID is now omitted rather than failing `Session.create`. Covered by `test_datafast_metadata_bounds_and_sanitizes_values`.

autogpt_platform/frontend/src/services/analytics/datafast-attribution.ts
```diff
@@ -0,0 +1,27 @@
...
+export function getDatafastAttribution(): Record<string, string> {
+  if (typeof document === "undefined") return {};
+
+  const jar = new Map(
+    document.cookie.split("; ").map((part) => {
+      const eq = part.indexOf("=");
+      return eq === -1
+        ? [part, ""]
+        : [part.slice(0, eq), decodeURIComponent(part.slice(eq + 1))];
+    }) as [string, string][],
```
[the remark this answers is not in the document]
0ubbe: 🤖 Fixed in [6c4f57d70](https://github.com/Significant-Gravitas/AutoGPT/commit/6c4f57d70c6df8cce8c4fc6cf6fa05606ee77892): cookie values are no longer decoded while building the jar; `decodeURIComponent` now runs only on the matched DataFast cookies inside a try/catch that ignores `URIError`. A malformed unrelated cookie can no longer throw, honoring the "Never throws" contract. Added two regression tests (malformed unrelated cookie + malformed DataFast value).

autogpt_platform/backend/backend/api/features/v1.py
0ubbe: 🤖 Fixed in [6c4f57d70](https://github.com/Significant-Gravitas/AutoGPT/commit/6c4f57d70c6df8cce8c4fc6cf6fa05606ee77892): all four `X-Datafast-*` `Header()` params now use `include_in_schema=False`, so the OpenAPI spec is unchanged and `check API types` passes without regenerating the client. Verified locally: `pnpm generate:api` produces no diff and `pnpm types` is clean.

ntindle (dismissed):

ntindle:
Plz actually test this -- I saw note in pr description

autogpt_platform/backend/backend/data/credit.py
```diff
@@ -103,6 +103,39 @@ class InvoiceListItem(BaseModel):
...
+_STRIPE_METADATA_VALUE_MAX_LEN = 500
+
+
+def _sanitize_datafast_id(value: str | None) -> str | None:
+    if not value:
+        return None
+    normalized = value.strip()
+    if not normalized or any(ord(ch) < 0x20 for ch in normalized):
+        return None
+    return normalized[:_STRIPE_METADATA_VALUE_MAX_LEN]
```
Swiftyos: This doesn't make all that much sense to me, why submit an id that we are truncating anyway. It would make more sense to just not add it to the stripe request when its invalid

0ubbe: 🤖 Agreed — fixed in [cea80d3cc](https://github.com/Significant-Gravitas/AutoGPT/commit/cea80d3ccfb6c94d000f0510a9de134f57fc401a). `_sanitize_datafast_id` now **drops** an oversized ID entirely (returns `None`, so it's never added to the Stripe request) instead of truncating it. A real DataFast ID is short, so anything over the 500-char limit is malformed, and a truncated value would be useless for attribution anyway. Blank and control-char values are dropped the same way; valid IDs pass through untouched. Test updated to assert the drop behaviour (`test_datafast_metadata_drops_invalid_values`).

0ubbe:
🤖 Done — ran the relevant tests locally:

**Frontend** (`pnpm vitest run`): **20/20 pass**
- `datafast-attribution.test.ts` (5) — cookie→header helper incl. malformed-cookie regressions
- `settings/billing/__tests__/datafast-attribution.test.tsx` (2) — page-level top-up flow asserting the `X-Datafast-*` headers
- `api/proxy/[...path]/__tests__/route.test.ts` (13) — header allow-list forwarding

**Backend**: `_datafast_metadata` / `_sanitize_datafast_id` verified directly (present/missing IDs, drop-on-invalid, whitespace strip, control-char drop, 500-char boundary). The full `credit_metadata_test.py` / `v1_test.py` / `subscription_routes_test.py` suites run in CI (`test (3.11/3.12/3.13)`), currently green.

I've also updated the PR description to drop the earlier "not tested locally" note.

PR #13292, opened 2026-06-04
refactor(backend): aggregate Stripe reconciliation alerting into one Discord system alert

majdyz:
### Background

Follow-up to the merged Stripe tier reconciliation work (#13284).

- Why: the reconciliation alerting emitted a **per-user ERROR log → one Sentry event per reconciled account**, plus a per-user Discord ping on the lazy path. In practice a sweep that corrects N accounts produced N separate Sentry alerts (observed: a single-person alert for one reconciled tester), which is noise, not signal. Ops wants one aggregate notification.

### Changes 🏗️

What
- Replace the per-user Sentry alerts with **one Discord system alert per sweep** (`DiscordChannel.PLATFORM`) that states **"reconciled N account(s) (X upgraded, Y downgraded)"** and lists the affected **user IDs**.
- Per-discrepancy logging drops from `ERROR` (one Sentry event each) to `INFO` + the existing PostHog metric, so the discrepancy rate is still trended without per-user paging.
- The lazy on-access reconcile no longer pings ops per-user; corrections are reported in aggregate by the sweep.
- Steady-state sweeps stay completely silent (no discrepancies → no alert).

How
- `log_tier_reconciliation_discrepancy`: `logger.error` → `logger.info` (keeps the PostHog event).
- `reconcile_stripe_tier_for_user`: removed the two per-user `alert_tier_reconciliation_discrepancy` calls (keeps the info/metric record).
- `_alert_sweep_discrepancies`: single Discord alert with the user-id list + "reconciled N accounts" summary; its own log demoted `ERROR` → `WARNING` (Discord is the alert surface, not Sentry).

### Checklist 📋

- [x] Changes listed in the PR description
- [x] Test plan made and executed:
  - [x] Sweep alert asserts the aggregate message ("reconciled N account", "X upgraded/downgraded", "User IDs:" + ids)
  - [x] Lazy reconcile asserts it does NOT ping ops per-user; steady-state asserts no alert
  - [x] Affected suites green (329 passed); `black` + `ruff` clean

### DIFF
--- a/autogpt_platform/backend/backend/api/features/v1_stripe_webhook_test.py
+++ b/autogpt_platform/backend/backend/api/features/v1_stripe_webhook_test.py
@@ -555,15 +555,16 @@ async def test_reconcile_stripe_tier_no_active_sub_downgrades_payer(
 
     assert await reconcile_stripe_tier_for_user("user_abc") is True
     mock_set.assert_awaited_once_with("user_abc", SubscriptionTier.NO_TIER)
-    # A silently-missed cancellation webhook must alert ops, not be corrected quietly.
-    mock_alert.assert_awaited_once()
-    assert "DOWNGRADED" in mock_alert.await_args.args[0]
+    # The lazy path does NOT ping ops per-user; the sweep's aggregate Discord
+    # alert reports corrections. The change is still recorded (log + PostHog).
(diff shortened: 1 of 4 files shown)


majdyz:
Folding this into #13291 per request — same change, consolidated.

PR #13294, opened 2026-06-04
feat(backend/copilot-bot): attach workspace artifacts to bot replies

Bentlybro:
## Why

When AutoPilot produces a file in a chat (e.g. "research X and write it up to a file and send it to me"), the Discord bot was dumping the raw `workspace://<uuid>` URI into the channel — a useless, unclickable string. The user had no way to get the file without opening the web app.

## What

The bot now **downloads the file and attaches it inline** to the chat. If the file is over the platform's hard upload cap (25 MB on Discord — bots can't have Nitro), it falls back to an **"Open in AutoGPT" button** linking to the session, mirroring the existing OAuth-link-button pattern.

This works on every surface the bot replies on (DMs, threads, channels) and is platform-agnostic — Telegram/Slack adapters get it for free once they implement `send_file`.

## How

- **`platform_linking`** — new `fetch_workspace_artifact(session_id, file_id, max_bytes)` RPC on `PlatformLinkingManager`. It:
  - resolves `session → user → workspace → file`, enforcing that the file belongs to the **session's owning user**. This check is load-bearing: the LLM emits arbitrary `workspace://` URIs into untrusted chat text, so we never serve a file just because the bot claimed an ID.
  - bails out early (returns `None`) if the file is over `max_bytes`, missing, or not owned — `None` triggers the link-to-chat fallback.
  - reads bytes through the existing `WorkspaceManager` storage abstraction — GCS in prod, filesystem locally. No bespoke file handling, same path the web download uses.
- **adapters** — new `FileAttachment` dataclass + `send_file()` + `max_attachment_bytes` on the platform-agnostic `PlatformAdapter`. Discord implements `send_file` with an in-memory `discord.File(io.BytesIO(...))`.
- **handler** — peels `[name](workspace://uuid#mime)` artifact links out of the streamed text (same regex shape the frontend already uses), attaches inline when they fit, else drops the fallback button.
- **docker-compose** — mounts the shared `workspace-data` volume on `database_manager` so it can read local-storage blobs. **No-op in prod** (workspace storage is GCS there); without it the feature silently fails for every dev running local filesystem storage, because `database_manager` is where the file read executes.

### Data flow / lifecycle

Bytes stream `workspace storage → DatabaseManager → bot → Discord`, held **only in memory** the whole way. Nothing is written to the bot's disk — no temp files, no persistence. The 25 MB cap bounds memory use.

### Reviewer notes (honest tradeoffs)

- **DatabaseManager reads blob storage.** The file read runs in the DB-manager process (that's where `platform_linking_db()` resolves to). It already reads workspace files for embeddings, so it's not a new responsibility, but it's why the compose volume mount is needed. Reasonable, but flagging it openly.
- **Whole file in memory, briefly in two processes** (DB manager + bot) plus RPC serialization. Fine at 25 MB; called out so it's not a surprise.
- **Artifact detection is regex over streamed text**, not a structured stream event. Matches the frontend's own approach. A mid-stream chunk split could theoretically miss an artifact; a follow-up to emit structured artifact events would remove that fragility.

## Testing

- `handler_test.py`: artifact extraction (regex correctness, blank-line cleanup), inline-attach happy path, too-large → button, fetch-failure → button, missing-session → inline note.
- `adapter_test.py`: `send_file` attaches bytes with the display name; empty-caption collapses to `None`.
- Manually verified end-to-end on Discord: small file attaches inline; debug-lowered cap confirms the >cap fallback button.
- `pyright` clean on all changed files; black/isort/ruff applied.

### DIFF
--- a/autogpt_platform/backend/backend/copilot/bot/handler.py
+++ b/autogpt_platform/backend/backend/copilot/bot/handler.py
@@ -7,12 +7,15 @@ persistent typing indicator.
 
 import asyncio
 import logging
-import re
 from dataclasses import dataclass, field
 from typing import Any
 from urllib.parse import quote
 
 from backend.data.redis_client import get_redis_async
+from backend.data.sharing.workspace_refs import (
+    WorkspaceArtifactLink,
+    extract_artifact_links,
+)
 from backend.util.exceptions import (
     DuplicateChatMessageError,
     LinkAlreadyExistsError,
@@ -32,20 +35,6 @@ THREAD_NAME_MAX_LENGTH = 100
 THREAD_NAME_PREFIX = "AutoPilot: "
 TITLE_RENAME_ATTEMPTS = 5
 TITLE_RENAME_INTERVAL_SECONDS = 1.0
-
-# Matches the workspace artifact markdown the LLM emits: `[name](workspace://uuid#mime)`.
-# Both the display name and the trailing `#mime` fragment are optional from
-# our side — we only rely on the file ID (group 2) for the actual fetch.
-_WORKSPACE_ARTIFACT_RE = re.compile(
-    r"\[([^\]]+)\]\(workspace://([A-Za-z0-9_-]+)(?:#([^)]*))?\)"
-)
-
-
-@dataclass(frozen=True)
-class _ParsedArtifact:
-    display_name: str
-    file_id: str
-    mime_hint: str | None
 
 
 @dataclass
@@ -183,7 +172,7 @@ class MessageHandler:
         when small enough, otherwise we drop a link button pointing at the
         chat on the platform so the user can grab it from there.
         """
-        stripped, artifacts = _extract_artifacts(text)
+        stripped, artifacts = extract_artifact_links(text)
         sent_any = False
         if stripped:
             await adapter.send_message(
@@ -201,7 +190,7 @@ class MessageHandler:
         self,
         adapter: PlatformAdapter,
         target_id: str,
-        artifact: _ParsedArtifact,
+        artifact: WorkspaceArtifactLink,
         session_id: str | None,
     ) -> bool:
         """Attach the file inline when possible; otherwise drop a link to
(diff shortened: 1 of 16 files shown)


autogpt_platform/docker-compose.platform.yml
```diff
@@ -341,6 +341,12 @@ services:
       <<: *backend-env
     ports:
       - "8005:8005"
+    volumes:
+      # database_manager serves workspace file reads (e.g. the bot's
+      # fetch_workspace_artifact RPC and embedding backfill), so it needs the
+      # same local-storage volume as the writer/reader services. No-op in
+      # prod where workspace storage is GCS.
+      - workspace-data:/app/autogpt_platform/backend/workspaces
```
ntindle: hmm justify this

autogpt_platform/backend/backend/copilot/bot/handler.py
```diff
@@ -32,6 +33,20 @@
 TITLE_RENAME_ATTEMPTS = 5
 TITLE_RENAME_INTERVAL_SECONDS = 1.0
 
+# Matches the workspace artifact markdown the LLM emits: `[name](workspace://uuid#mime)`.
+# Both the display name and the trailing `#mime` fragment are optional from
+# our side — we only rely on the file ID (group 2) for the actual fetch.
+_WORKSPACE_ARTIFACT_RE = re.compile(
+    r"\[([^\]]+)\]\(workspace://([A-Za-z0-9_-]+)(?:#([^)]*))?\)"
+)
```
ntindle: we shuld find out wherever this is defined originally and not duplicate it

autogpt_platform/backend/backend/copilot/bot/handler.py
```diff
@@ -32,6 +33,20 @@
...
+
+
+@dataclass(frozen=True)
+class _ParsedArtifact:
+    display_name: str
+    file_id: str
+    mime_hint: str | None
```
ntindle: base class from pydantic. tell your agent to stop doing this

autogpt_platform/backend/backend/copilot/bot/handler.py
```diff
@@ -32,6 +33,20 @@
 TITLE_RENAME_ATTEMPTS = 5
 TITLE_RENAME_INTERVAL_SECONDS = 1.0
 
+# Matches the workspace artifact markdown the LLM emits: `[name](workspace://uuid#mime)`.
+# Both the display name and the trailing `#mime` fragment are optional from
+# our side — we only rely on the file ID (group 2) for the actual fetch.
+_WORKSPACE_ARTIFACT_RE = re.compile(
```
ntindle: double check all of this file for DRY (dont repeat yourself) and make sure thsi isn't somethign we're already handling eslse where that should be extracted to a helper

ntindle (changes requested):

autogpt_platform/docker-compose.platform.yml
Bentlybro: i ment to remove this, it was just so i could properly test autopilot making and saving files to a workspace so the bot could send it to me

goingforstudying-ctrl:
Removed the test-only workspace-data volume mount from database_manager as requested. The mount and its comment were only needed for local testing of autopilot file creation and aren't required in the compose setup.

Patch:

```diff
diff --git a/autogpt_platform/docker-compose.platform.yml b/autogpt_platform/docker-compose.platform.yml
index 0074208..c210791 100644
--- a/autogpt_platform/docker-compose.platform.yml
+++ b/autogpt_platform/docker-compose.platform.yml
@@ -341,12 +341,6 @@ services:
       <<: *backend-env
     ports:
       - 8005:8005
-    volumes:
-      # database_manager serves workspace file reads (e.g. the bot's
-      # fetch_workspace_artifact RPC and embedding backfill), so it needs the
-      # same local-storage volume as the writer/reader services. No-op in
-      # prod where workspace storage is GCS.
-      - workspace-data:/app/autogpt_platform/backend/workspaces
     networks:
       - app-network
     logging:
```

ntindle (approved):

PR #13300, opened 2026-06-04
feat(blocks): add variable inputs to Execute Code block

### LINKED ISSUE #13286: Add dynamic input fields to Execute Code block
To pass data into an Execute Code block, you currently have to insert them like "convert to string" -> "fill text template" -> code -> execute, which is cumbersome and causes AutoPilot to use AI blocks for data shape transformations.

If we can instead add dynamic input fields to the code execution block(s) - like the ones on the Fill Text Template block or any of the LLM blocks - which expose the passed data as a variable value in the run env, we can make this massively easier.

Void-n-Null:
### Why / What / How

Closes #13286

Right now there's no way to get data into an Execute Code block directly. If you want to use some upstream data in your code, you have to route it through other blocks first, usually something like convert to string, then fill text template, then feed that into the code field. It's clunky, and it pushes AutoPilot into using AI blocks in a use case where LLMs are genuinely wasteful. 

This PR adds a variables field to the block. The values get passed to the sandbox as a JSON env var, and a small fixed snippet prepended to the code reads them back out into named variables (globals().update(...) in Python, Object.assign(globalThis, ...) in JS). The snippet is constant, so user data only ever rides in the env var, never in the code string. Values wired in from other blocks keep their type, and default values set on the node come in as strings, so you parse them in your code if you need a number or other type.

### Changes 🏗️

- code_executor.py: added the variables field to ExecuteCodeBlock.Input (above code).  run() now builds the env var and prefix from the helper and prepends the snippet to the code. Added an envs arg to execute_code that gets passed through to sandbox.run_code. Only ExecuteCodeBlock uses this; the other two code blocks are untouched.

- code_executor_helpers.py (new): build_variable_injection() does the serialization, the per-language prefix, and the unsupported-language check. Also moved ProgrammingLanguage here because the helper needs it and the block needs the helper, which was a circular import. It's re-exported from code_executor.py so nothing else breaks.

- code_executor_test.py (new): covers the helper (both language prefixes, a malicious value staying inert, unsupported languages raising) and run() (prefixing the code, passing envs, doing nothing when there are no variables, erroring on an unsupported language with variables set).

#### For code changes:
- [x] I have made a test plan
  - [x] Add an Execute Code block set to `Python`, set a variable `name` = `World`, run `print(f"Hello {name}")`, confirm it outputs `Hello World`
  - [x] Wire an upstream block's numeric output into a variable `x`, run `print(x + 1)`, confirm the number flows through with its type intact
  - [x] Switch to JavaScript, variable `name` = `World`, run ``console.log(`Hello ${name}`)``, confirm `Hello World`
  - [x] Set the block to Bash with a variable set, confirm it errors out due to Bash being an unsupported language for variables.
  - [x] confirm code in other languages still runs when no variables are defined on the block.

No configuration changes

image

### DIFF
--- a/autogpt_platform/backend/backend/blocks/code_executor.py
+++ b/autogpt_platform/backend/backend/blocks/code_executor.py
@@ -230,7 +230,7 @@ class ExecuteCodeBlock(Block, BaseE2BExecutorMixin):
             description=(
                 "Variables defined here can be used directly in your Python or "
                 "JavaScript code. Values wired in from other blocks keep their "
-                "type; values typed directly here come in as strings, so parse "
+                "type; default values set on this node come in as strings, so parse "
                 "them in your code if you need a number or other type."
             ),
             default_factory=dict,
--- a/autogpt_platform/backend/backend/blocks/code_executor.py
+++ b/autogpt_platform/backend/backend/blocks/code_executor.py
@@ -1,4 +1,3 @@
-from enum import Enum
 from typing import TYPE_CHECKING, Any, Literal, Optional
 
 from e2b_code_interpreter import AsyncSandbox
@@ -12,6 +11,10 @@ from backend.blocks._base import (
     BlockOutput,
     BlockSchemaInput,
     BlockSchemaOutput,
+)
+from backend.blocks.code_executor_helpers import (
+    ProgrammingLanguage,
+    build_variable_injection,
 )
 from backend.data.model import (
     APIKeyCredentials,
@@ -41,14 +44,6 @@ TEST_CREDENTIALS_INPUT = {
     "type": TEST_CREDENTIALS.type,
     "title": TEST_CREDENTIALS.type,
 }
-
-
-class ProgrammingLanguage(Enum):
-    PYTHON = "python"
-    JAVASCRIPT = "js"
-    BASH = "bash"
-    R = "r"
-    JAVA = "java"
 
 
 class MainCodeExecutionResult(BaseModel):
@@ -107,6 +102,7 @@ class BaseE2BExecutorMixin:
         dispose_sandbox: bool = False,
         execution_context: Optional["ExecutionContext"] = None,
         extract_files: bool = False,
+        envs: Optional[dict[str, str]] = None,
     ):
         """
         Unified code execution method that handles all three use cases:
@@ -145,6 +141,7 @@ class BaseE2BExecutorMixin:
             execution = await sandbox.run_code(  # type: ignore[attr-defined]
                 code,
                 language=language.value,
+                envs=envs,
                 on_error=lambda e: sandbox.kill(),  # Kill the sandbox on error
             )
 
@@ -228,6 +225,18 @@ class ExecuteCodeBlock(Block, BaseE2BExecutorMixin):
             advanced=False,
         )
 
+        variables: dict[str, Any] = SchemaField(
+            title="Variables (Python/JS only)",
+            description=(
+                "Variables defined here can be used directly in your Python or "
+                "JavaScript code. Values are parsed as JSON when possible "
+                "(e.g. 42 becomes a number, true a boolean); anything else is "
+                'treated as text. Wrap in quotes to force a string (e.g. "42").'
+            ),
+            default_factory=dict,
+            advanced=False,
+        )
+
         code: str = SchemaField(
             description="Code to execute in the sandbox",
             placeholder="print('Hello, World!')",
@@ -308,7 +317,7 @@ class ExecuteCodeBlock(Block, BaseE2BExecutorMixin):
                 ("files", []),
             ],
             test_mock={
-                "execute_code": lambda api_key, code, language, template_id, setup_commands, timeout, dispose_sandbox, execution_context, extract_files: (  # noqa
+                "execute_code": lambda api_key, code, language, template_id, setup_commands, timeout, dispose_sandbox, execution_context, extract_files, envs: (  # noqa
                     [],  # results
                     "Hello World",  # text_output
                     "Hello World\n",  # stdout_logs
@@ -328,9 +337,17 @@ class ExecuteCodeBlock(Block, BaseE2BExecutorMixin):
         **kwargs,
     ) -> BlockOutput:
         try:
+            # Expose user-provided variables by passing them as a JSON env var and
+            # prepending a constant snippet that deserializes them into the runtime.
+            # Keeping the data in the env var (not the code string) avoids injection.
+            envs, prefix = build_variable_injection(
+                input_data.variables, input_data.language
+            )
+            code = prefix + input_data.code
+
             results, text_output, stdout, stderr, _, files = await self.execute_code(
                 api_key=credentials.api_key.get_secret_value(),
-                code=input_data.code,
+                code=code,
                 language=input_data.language,
                 template_id=input_data.template_id,
                 setup_commands=input_data.setup_commands,
@@ -338,6 +355,7 @@ class ExecuteCodeBlock(Block, BaseE2BExecutorMixin):
                 dispose_sandbox=input_data.dispose_sandbox,
                 execution_context=execution_context,
                 extract_files=True,
+                envs=envs,
             )
 
             # Determine result object shape & filter out empty formats
--- a/autogpt_platform/backend/backend/blocks/code_executor.py
+++ b/autogpt_platform/backend/backend/blocks/code_executor.py
@@ -229,9 +229,9 @@ class ExecuteCodeBlock(Block, BaseE2BExecutorMixin):
             title="Variables (Python/JS only)",
             description=(
                 "Variables defined here can be used directly in your Python or "
-                "JavaScript code. Values are parsed as JSON when possible "
-                "(e.g. 42 becomes a number, true a boolean); anything else is "
-                'treated as text. Wrap in quotes to force a string (e.g. "42").'
+                "JavaScript code. Values wired in from other blocks keep their "
+                "type; values typed directly here come in as strings, so parse "
+                "them in your code if you need a number or other type."
             ),
             default_factory=dict,
             advanced=False,
--- a/autogpt_platform/backend/backend/blocks/code_executor.py
+++ b/autogpt_platform/backend/backend/blocks/code_executor.py
@@ -38,6 +38,20 @@ TEST_CREDENTIALS = APIKeyCredentials(
     title="Mock E2B API key",
     expires_at=None,
 )
+
+
+def _mock_execute_code(
+    *args: Any, **kwargs: Any
+) -> tuple[list, str, str, str, str, list]:
+    """Stub for `execute_code` used by the blocks' test_mock.
+
+    Returns the (results, text, stdout, stderr, sandbox_id, files) tuple,
+    echoing back any provided `sandbox_id` so the step block's test sees it.
+    """
+    sandbox_id = kwargs.get("sandbox_id") or "sandbox_id"
+    return [], "Hello World", "Hello World\n", "", sandbox_id, []
+
+
 TEST_CREDENTIALS_INPUT = {
     "provider": TEST_CREDENTIALS.provider,
     "id": TEST_CREDENTIALS.id,
@@ -141,7 +155,7 @@ class BaseE2BExecutorMixin:
             execution = await sandbox.run_code(  # type: ignore[attr-defined]
                 code,
                 language=language.value,
-                envs=envs,
+                envs=envs or {},
                 on_error=lambda e: sandbox.kill(),  # Kill the sandbox on error
             )
 
@@ -316,16 +330,7 @@ class ExecuteCodeBlock(Block, BaseE2BExecutorMixin):
                 ("stdout_logs", "Hello World\n"),
                 ("files", []),
             ],
-            test_mock={
-                "execute_code": lambda api_key, code, language, template_id, setup_commands, timeout, dispose_sandbox, execution_context, extract_files, envs: (  # noqa
-                    [],  # results
-                    "Hello World",  # text_output
-                    "Hello World\n",  # stdout_logs
-                    "",  # stderr_logs
-                    "sandbox_id",  # sandbox_id
-                    [],  # files
-                ),
-            },
+            test_mock={"execute_code": _mock_execute_code},
         )
 
     async def run(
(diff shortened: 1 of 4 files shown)


autogpt_platform/backend/backend/blocks/code_executor_helpers.py
```diff
@@ -0,0 +1,101 @@
...
+    The builder's dynamic-dict widget stores every value as a string. Trying
+    json.loads recovers the intended type; values that aren't valid JSON (e.g.
+    `hello`) are kept as plain strings. Non-string values are passed through.
+    """
+    if not isinstance(value, str):
+        return value
+    try:
+        return json.loads(value)
+    except (json.JSONDecodeError, ValueError):
+        return value
```
Pwuts: I'm trying to think whether this will cause problems in common use cases. Converting something to a different type because it happens to have a very specific value seems tricky.

Alternatives: 
- no type conversion, default values specified on the node are passed as string (you can always parse/coerce them in code)
- amend the input renderer for `dict[str, Any]` so it lets you choose an explicit type per dynamic field

Pwuts (commented):
Awesome, I love it! Thank you so much for making this (and so fast)! :)

autogpt_platform/backend/backend/blocks/code_executor_helpers.py
Void-n-Null: That's a fair thing to worry about. I originally wanted to use the anyOf dropdown UI, and it does render, but in the dynamic-dict field puts a type-selector above every single variable value, which made some pretty choppy and cluttered UI. Properly fixing that would mean changes to frontend block-node rendering, which felt out of scope (and is an area I haven't dug into much yet).

I don't hate the idea of passing everything in as strings and parsing in code, but it still feels like conceding on UX especially for math heavy stuff. A middle ground might be to only auto-parse numbers. The least obvious visuals for auto-conversions are booleans, lists, and dicts, since plain text can accidentally look like those. If I enforce only auto-parsing of ints and floats, math variables can still flow in smoothly (`x = 5` → `print(x + 1)` → `6`) without the other conversions catching people by suprise.

For more complex data like lists or dictionaries it might be reasonable to format a json string and do json.loads() manually anyway. 

That's probably a smarter boundary for auto parsing anyhow.

Pwuts: Just to clarify, numbers can already be passed in from other nodes, so the issue is purely with default values specified on the node. The fix, however, applies to both default and passed-in values.

Parsing numbers seems low risk but is still inconsistent behavior. Especially when e.g. you pass in a prefab object into `variables` directly (which you shouldn't, but ok), it may implicitly coerce the type of *some* keys of the object. 

Anyway, it seems like a decent trade-off. At the same time, it's not a lot of effort to add a `someVar = parseInt(someVar)` at the beginning of your code. Wdyt?

Void-n-Null: Yeah, I think youre right. I didn't consider that the UX surface is really small since its purely typed out defaults on the node and its genuinely inconsistent.

Alright, yeah since data actually transfering from other blocks already comes in as the right data type and thats the primary use case anyhow, I'll just remove the auto-parsing altogether. Thank you for thinking through it with me!

autogpt_platform/backend/backend/blocks/code_executor.py
```diff
@@ -228,6 +225,18 @@ class Input(BlockSchemaInput):
             advanced=False,
         )
 
+        variables: dict[str, Any] = SchemaField(
+            title="Variables (Python/JS only)",
+            description=(
+                "Variables defined here can be used directly in your Python or "
+                "JavaScript code. Values wired in from other blocks keep their "
+                "type; values typed directly here come in as strings, so parse "
+                "them in your code if you need a number or other type."
```
Pwuts:
```suggestion
                "type; default values set on this node come in as strings, so parse "
                "them in your code if you need a number or other type."
```

Void-n-Null:
Pushed a commit addressing most of the feedback:

- Variable names are validated now (rejects non-identifiers, keywords, and `__`/`_agpt_` prefixes that'd collide with the inject snippet) instead of silently making unusable vars

- 64KB cap on the serialized payload so oversized input fails clearly

- `envs` defaults to `{}` so the other two blocks never get `None`

- Swapped the test_mock lambdas for one named typed stub, dropped the `# type: ignore`s, and added a JS `run()` test

- Regenerated the docs so `check-docs-sync` passes

autogpt_platform/backend/backend/blocks/code_executor.py
```diff
@@ -228,6 +239,18 @@ class Input(BlockSchemaInput):
             advanced=False,
         )
 
+        variables: dict[str, Any] = SchemaField(
```
Void-n-Null, line 242: 🤖 🟡 **Nice to Have**: The `variables` field is always visible regardless of selected language, but injection only works for Python/JS — a user on Bash/R/Java only discovers the incompatibility at runtime (the block yields an error). The `title="Variables (Python/JS only)"` hint mitigates this, and the runtime error is clear, so this is fine to ship. A future improvement would be conditionally hiding the field based on `language` (`depends_on`/`show_if` style) so the constraint is enforced in the UI rather than surfaced as a failed run. Not a blocker.

Void-n-Null (commented):
## 🤖 Review Summary — Approve (with one minor nit)

Verified the code locally: `code_executor_test.py` (19 tests) and the three block self-tests (`ExecuteCodeBlock`, `InstantiateCodeSandboxBlock`, `ExecuteCodeStepBlock`) all pass; `ruff` is clean on the changed files.

### What's good
- **Injection-safe by construction.** User data only ever rides in the `AGPT_VARIABLES` env var (the data channel); the prepended snippet is a fixed constant that references only the env var. This is the right model — analogous to parameterized SQL — and the `test_malicious_value_cannot_break_out_of_code_channel` test guards it.
- **Defensive validation already in place.** Key validation (`_validate_keys`: identifiers, no keywords, no `__`/`_agpt_` prefixes), a 64 KB payload cap, and a clear non-serializable error path — all covered by tests.
- **Clean refactor.** Moving `ProgrammingLanguage` into `code_executor_helpers.py` to break the circular import (and re-exporting it) is the correct call. The `_mock_execute_code` helper removes the previous brittle positional-lambda mocks + their `# noqa`/`# type: ignore` suppressors — a net style improvement.
- **`envs` is opt-in.** Sibling blocks don't pass it; `envs or {}` keeps their behavior unchanged.

### On the existing bot comments
Most of the `autogpt-pr-reviewer`/CodeRabbit findings are **stale or already addressed** in the current head:
- Key validation, payload size limit, JS-path `run()` test, and invalid-key test all **exist now**.
- The type-coercion concern raised by @Pwuts was **resolved in-thread** — auto-parsing was removed and the field description updated to match. ✅
- `check-docs-sync` is **passing**; the docs table row matches the generated output.
- The `# type: ignore`/`# noqa` flagged in the test file are **gone** (refactored away).

### CI
Backend `test` (3.11/3.12/3.13), `lint`, `types`, `type-check`, `check-docs-sync`, and CodeQL all pass. The red checks — `Vercel` (deploy authorization required), `end-to-end tests`, and the `Check PR Status` aggregator that depends on them — are infra/auth failures unrelated to this change.

### One nit (non-blocking)
Inline on `code_executor.py:242` — the `variables` field is always shown even for languages where injection is unsupported; a future `show_if`-style conditional render would enforce the Python/JS constraint in the UI instead of at runtime. The current title hint + clear runtime error are acceptable for now.

**Verdict: 🟢 Approve.** Clean, well-tested, and the one substantive design question was already resolved with the maintainer.

Void-n-Null:
Oh! Apologies I didn't realize that running the review skill locally actually made my Github account comment. Ignore that!

Pwuts:
IMO shadowing builtin functions with a variable holding a non-function value is not a big issue. Shadowing keywords isn't either. It may break at runtime, but:
- AutoPilot and its agent generator won't make such mistakes
- Anyone building an agent by hand who tries using a problematic variable name will simply get something that doesn't work, not a security catastrophe

Building something that doesn't quite work is part of the (trial and error) process. I suspect it will be so rare that someone tries to use a variable name that breaks their code that it isn't worth building key validation for every supported language.

Void-n-Null:
@Pwuts 

> Building something that doesn't quite work is part of the (trial and error) process. I suspect it will be so rare that someone tries to use a variable name that breaks their code that it isn't worth building key validation for every supported language.

I 100% agree. I think that what's already setup for key validation is probably already overkill. It'll probably save someone 10 seconds of being confused why their code broke, but I think it is far from worth any further stress about shadowing.

Pwuts (approved):
I'm going to approve this. :)

There are still a bunch of roboreview comments. The PR will merge automatically as soon as they are all marked resolved. I'll leave it up to you whether you want to heed or dismiss them.

goingforstudying-ctrl:
Pushed a commit that addresses the review feedback about variable name validation:

- Added _JS_RESERVED_WORDS and _JS_BUILTIN_GLOBALS sets to prevent shadowing critical JS globals (process, require, module, console, etc.) and reserved words (const, let, class, await, etc.).
- Added _PYTHON_BUILTIN_BLOCKLIST for sensitive builtins (exec, eval, open, compile, exit, quit, breakpoint).
- Updated _validate_keys to take a language parameter and check the appropriate blocklist per language.
- Added test coverage for JS reserved words, JS globals, and Python builtins.

I don't have push access to the fork, so I can't push the commit directly. Here's the patch:

```diff
From 02ea33518336484406ad20a3951132a81e971352 Mon Sep 17 00:00:00 2001
From: Void <void@null.dev>
Date: Sat, 6 Jun 2026 05:38:29 -0400
Subject: [PATCH] fix(blocks): block JS reserved words and Python builtins in
 variable names

- Add _JS_RESERVED_WORDS and _JS_BUILTIN_GLOBALS sets to prevent shadowing
  critical JS globals (process, require, module, console, etc.) and reserved
  words (const, let, class, await, etc.).
- Add _PYTHON_BUILTIN_BLOCKLIST for sensitive builtins (exec, eval, open,
  compile, exit, quit, breakpoint).
- Update _validate_keys to take a language parameter and check the appropriate
  blocklist per language.
- Add test coverage for JS reserved words, JS globals, and Python builtins.
---
 .../backend/blocks/code_executor_helpers.py   | 86 ++++++++++++++++---
 .../backend/blocks/code_executor_test.py      | 18 +++-
 2 files changed, 92 insertions(+), 12 deletions(-)

diff --git a/autogpt_platform/backend/backend/blocks/code_executor_helpers.py b/autogpt_platform/backend/backend/blocks/code_executor_helpers.py
index 80e5b99..2a076b6 100644
--- a/autogpt_platform/backend/backend/blocks/code_executor_helpers.py
+++ b/autogpt_platform/backend/backend/blocks/code_executor_helpers.py
@@ -12,6 +12,67 @@ import keyword
 from enum import Enum
 from typing import Any
 
+# JavaScript reserved words and built-in globals that should not be shadowed.
+_JS_RESERVED_WORDS = frozenset(
+    [
+        "break",
+        "case",
+        "catch",
+        "class",
+        "const",
+        "continue",
+        "debugger",
+        "default",
+        "delete",
+        "do",
+        "else",
+        "export",
+        "extends",
+        "false",
+        "finally",
+        "for",
+        "function",
+        "if",
+        "import",
+        "in",
+        "instanceof",
+        "new",
+        "null",
+        "return",
+        "super",
+        "switch",
+        "this",
+        "throw",
+        "true",
+        "try",
+        "typeof",
+        "var",
+        "void",
+        "while",
+        "with",
+        "yield",
+        "let",
+        "static",
+        "await",
+        "enum",
+        "implements",
+        "interface",
+        "package",
+        "private",
+        "protected",
+        "public",
+    ]
+)
+
+_JS_BUILTIN_GLOBALS = frozenset(
+    ["process", "require", "module", "console", "global", "globalThis", "Buffer"]
+)
+
+# Python builtins that are dangerous to shadow in a sandbox.
+_PYTHON_BUILTIN_BLOCKLIST = frozenset(
+    ["exec", "eval", "open", "compile", "exit", "quit", "breakpoint"]
+)
+
 
 class ProgrammingLanguage(Enum):
     PYTHON = "python"
@@ -72,7 +133,7 @@ def build_variable_injection(
             "Supported languages: python, js."
         )
 
-    _validate_keys(variables)
+    _validate_keys(variables, language)
 
     try:
         serialized = json.dumps(variables)
@@ -93,7 +154,7 @@ def build_variable_injection(
     return envs, prefix
 
 
-def _validate_keys(variables: dict[str, Any]) -> None:
+def _validate_keys(variables: dict[str, Any], language: ProgrammingLanguage) -> None:
     """Reject keys that can't be used as a variable name in the sandbox.
 
     Each key becomes a global variable, so it must be a valid identifier that
@@ -101,19 +162,22 @@ def _validate_keys(variables: dict[str, Any]) -> None:
     builtins/internals, and `_agpt_`-prefixed names would collide with the
     deserialization snippet's own imports.
     """
-    invalid = [
-        key
-        for key in variables
-        if not key.isidentifier()
-        or keyword.iskeyword(key)
-        or key.startswith("__")
-        or key.startswith("_agpt_")
-    ]
+    invalid: list[str] = []
+    for key in variables:
+        if not key.isidentifier() or key.startswith("__") or key.startswith("_agpt_"):
+            invalid.append(key)
+            continue
+        if language == ProgrammingLanguage.PYTHON:
+            if keyword.iskeyword(key) or key in _PYTHON_BUILTIN_BLOCKLIST:
+                invalid.append(key)
+        elif language == ProgrammingLanguage.JAVASCRIPT:
+            if key in _JS_RESERVED_WORDS or key in _JS_BUILTIN_GLOBALS:
+                invalid.append(key)
     if invalid:
         raise ValueError(
             "Invalid variable name(s): "
             f"{', '.join(repr(k) for k in invalid)}. "
-            "Names must be valid identifiers, not language keywords, and not "
+            "Names must be valid identifiers, not language keywords/builtins, and not "
             "start with '__' or '_agpt_'."
         )
 
diff --git a/autogpt_platform/backend/backend/blocks/code_executor_test.py b/autogpt_platform/backend/backend/blocks/code_executor_test.py
index 5a1d16f..6df074f 100644
--- a/autogpt_platform/backend/backend/blocks/code_executor_test.py
+++ b/autogpt_platform/backend/backend/blocks/code_executor_test.py
@@ -125,7 +125,23 @@ class TestBuildVariableInjection:
         "bad_key",
         ["my var", "2x", "for", "__builtins__", "_agpt_json"],
     )
-    def test_invalid_variable_names_raise(self, bad_key):
+    def test_invalid_variable_names_raise_python(self, bad_key):
+        with pytest.raises(ValueError, match="Invalid variable name"):
+            build_variable_injection({bad_key: 1}, ProgrammingLanguage.PYTHON)
+
+    @pytest.mark.parametrize(
+        "bad_key",
+        ["const", "let", "var", "function", "class", "await", "process", "require"],
+    )
+    def test_invalid_variable_names_raise_javascript(self, bad_key):
+        with pytest.raises(ValueError, match="Invalid variable name"):
+            build_variable_injection({bad_key: 1}, ProgrammingLanguage.JAVASCRIPT)
+
+    @pytest.mark.parametrize(
+        "bad_key",
+        ["exec", "eval", "open", "compile", "exit", "quit", "breakpoint"],
+    )
+    def test_python_builtin_shadowing_blocked(self, bad_key):
         with pytest.raises(ValueError, match="Invalid variable name"):
             build_variable_injection({bad_key: 1}, ProgrammingLanguage.PYTHON)
 
-- 
2.43.0
```

If you apply it with `git apply` and push, the review comments should be resolved.

PR #13311, opened 2026-06-06
fix(frontend/copilot): pair tool calls to outputs by position

gaurav0107:
### Why / What / How

**Why:** Issue #13306 reports that, in the AutoPilot UI, the
descriptions of tool calls and their corresponding results are paired
with the wrong tool — a tool call's description is shown alongside
another call's result. The screenshot in the issue shows two adjacent
search-result cards whose descriptions match the OTHER card's results.

The root cause is in
`autogpt_platform/frontend/src/app/(platform)/copilot/helpers/convertChatSessionToUiMessages.ts`.
That helper flattens every `role: "tool"` message in the session into
a single `Map<tool_call_id, output>`:

```ts
for (const msg of messages) {
  if (msg.role !== "tool") continue;
  if (!msg.tool_call_id) continue;
  if (msg.content == null) continue;
  toolOutputsByCallId.set(msg.tool_call_id, msg.content);
}
```

Then for each assistant `tool_call`, it pairs by `tool_call_id`:

```ts
const output = toolOutputsByCallId.get(toolCallId);
```

LLM providers do **not** guarantee globally-unique `tool_call_id`s
across turns of the same session — only within a single
`completions.create` response. When a duplicate id reappears in a
later turn the flat map keeps only the LAST output and every earlier
assistant tool call with the same id silently picks up the WRONG
output. The user sees a `tool-web_search` accordion whose
"description" (built from the assistant's own input arguments)
belongs to the call that was made — but whose results body belongs to
a later, unrelated call.

The same bug also explains the cross-page symptom: the helper accepts
an `extraToolOutputs` map and an in-page override pass, both keyed by
`tool_call_id` — a duplicate id across pages collapses to a single
surviving output.

**What:** Make tool-call → tool-output pairing position-aware so a
duplicate `tool_call_id` cannot leak across turns. Each assistant
tool_call occurrence gets its own slot; tool messages and
cross-page seeds fill the OLDEST unfilled slot for that id in FIFO
order.

**How:**

- Walk messages in DB order. For each assistant `tool_calls` entry,
  allocate one slot per occurrence in `slotsById` (a
  `Map<tool_call_id, slotKey[]>`).
- After slot allocation, bind outputs in two passes:
  1. Cross-page seeds from `extraToolOutputs` fill the oldest pending
     slot for each id (FIFO). These represent tool outputs that
     landed on an adjacent page for the earliest assistant call(s) on
     this page.
  2. In-page `role: "tool"` messages fill the next pending slot for
     each id (also FIFO).
- Each assistant card renders the output bound to its specific slot,
  not whatever happens to be the last value under that id.

The "in-page wins over cross-page" semantics are preserved for the
common no-duplicate case — when both contribute one entry under the
same id, the cross-page seed fills slot 0 and in-page fills slot 1
(or, if there's only one assistant call, the cross-page seed fills it
and the in-page tool message has no slot to attach to). The pre-fix
code's "in-page overrides cross-page when both target the SAME id"
case was load-bearing only when an id collision existed — exactly the
bug case — and the post-fix behavior gives both occurrences distinct
outputs instead of dropping one.

### Changes 🏗️

- `autogpt_platform/frontend/src/app/(platform)/copilot/helpers/convertChatSessionToUiMessages.ts`:
  Replace the flat `toolOutputsByCallId` map with position-aware slot
  allocation and FIFO output binding. ~80 lines added (slot
  allocation + binding) / ~10 lines removed (old single-pass loop).
- `autogpt_platform/frontend/src/app/(platform)/copilot/helpers/__tests__/convertChatSessionToUiMessages.test.ts`:
  Add three regression tests covering duplicate-id pairing within a
  page, mixed ids across multiple turns, and cross-page seed +
  duplicate-id interaction.

#### For code changes:
- [x] I have made a test plan
  - [x] `pnpm vitest run src/app/(platform)/copilot/helpers/__tests__/convertChatSessionToUiMessages.test.ts` — 29/29 pass (26 existing + 3 new)
  - [x] New tests fail against the pre-fix code (verified by stashing the source change locally) and pass after the fix
  - [x] `pnpm format` clean (prettier --check passes on changed files)
  - [x] `pnpm lint` clean (next lint reports no warnings or errors on changed files)
  - [x] `pnpm types` clean (`tsc --noEmit -p tsconfig.json` exits 0)
  - [ ] Manual: open a chat that triggered the bug; confirm each tool card's description matches the snippet/results body inside the accordion below it (deferred — requires running the full platform stack)

### DIFF
--- a/autogpt_platform/frontend/src/app/(platform)/copilot/helpers/__tests__/convertChatSessionToUiMessages.test.ts
+++ b/autogpt_platform/frontend/src/app/(platform)/copilot/helpers/__tests__/convertChatSessionToUiMessages.test.ts
@@ -278,6 +278,272 @@ describe("convertChatSessionMessagesToUiMessages", () => {
     const mergedId = result.messages[1].id;
     expect(result.stats.get(mergedId)?.createdAt).toBe(later);
   });
+
+  // ----------------------------------------------------------------- //
+  //  Tool-call → tool-result pairing (#13306)                          //
+  // ----------------------------------------------------------------- //
+  //
+  // The LLM provider does NOT guarantee globally-unique tool_call_ids
+  // across multiple turns of the same session — only within one
+  // completions response.  When a duplicate id reappears in a later
+  // turn, naive id-keyed pairing collapses both occurrences onto the
+  // last tool message and the earlier assistant tool call ends up
+  // displaying the WRONG output (issue #13306: "tool call descriptions
+  // and result pairings are getting mixed up").  Pairing must be
+  // position-aware: each assistant tool_call slot is filled by the
+  // NEAREST FOLLOWING tool message with the matching id, and once
+  // filled it cannot be re-bound to a later tool message of the same
+  // id.
+
+  it("pairs duplicate tool_call_ids by position so each assistant tool call gets its own output", () => {
+    // Two distinct turns (user message between them) where the LLM
+    // happens to reuse "call_dup" for an unrelated web_search call in
+    // the second turn.  Pre-fix, both assistant messages received the
+    // SAME (last-wins) output because the helper keyed on tool_call_id
+    // alone.  Post-fix, each assistant call binds to its own slot.
+    const result = convertChatSessionMessagesToUiMessages(
+      SESSION_ID,
+      [
+        { role: "user", content: "search foo", sequence: 0 },
+        {
+          role: "assistant",
+          content: null,
+          sequence: 1,
+          tool_calls: [
+            {
+              id: "call_dup",
+              type: "function",
+              function: {
+                name: "web_search",
+                arguments: JSON.stringify({ query: "foo" }),
+              },
+            },
+          ],
+        },
+        {
+          role: "tool",
+          tool_call_id: "call_dup",
+          content: JSON.stringify({ results: [{ title: "foo result" }] }),
+          sequence: 2,
+        },
+        { role: "user", content: "now search bar", sequence: 3 },
+        {
+          role: "assistant",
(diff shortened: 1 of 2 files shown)


gaurav0107:
Pushed an update (force-push after rebase):

- **Rebased onto the latest `dev`** to clear the *behind base* status. The branch is now a clean two-commit series on top of current `dev` with no merge commits.
- **Added three regression tests** for the tool-call/output slot-pairing edge cases that were previously uncovered, which was keeping `codecov/patch/platform-frontend` just under the 70% patch-coverage gate:
  - an unresolved tool call renders `output-available` with an empty output when the session is complete (stops a stale spinner),
  - an unresolved tool call stays `input-available` while the session is still streaming,
  - malformed `tool_calls` entries (null / blank id) are skipped without consuming a slot, and orphan/surplus tool rows are dropped without disturbing the valid pairing.

No source-logic changes in this update — only the branch base and added test coverage. The `Vercel` check remains red because preview deploys require maintainer authorization for fork PRs (`Authorization required to deploy.`); that is outside a contributor's control.

Abhi1992002:
Hi @gaurav0107, thanks for the detailed investigation, but I’m closing this PR. The underlying issue (#13306) was already fixed by https://github.com/Significant-Gravitas/AutoGPT/pull/13372 (backend output stash was keyed by tool name only, swapping parallel same-tool outputs), and the symptom hasn’t recurred since. Duplicate tool_call_ids across turns haven’t been observed in practice (SDK toolu_ ids are unique per session).

PR #13325, opened 2026-06-08
feat(backend): add bot usage analytics (events + presence)

Bentlybro:
### Why / What / How

**Why** — we have zero visibility into how the AutoPilot Discord bot is actually being used. We don't know how many servers it's in (so can't predict when we'll need to shard the gateway), how often it's invoked, how often it errors, or which commands matter. Adding analytics also has to clear a hard privacy bar: the bot reads user DMs and channel messages, so the data layer must make it **impossible** to log message content or user identity.

**What** — adds two append-only tables (`BotEvent`, `BotGuild`), a writes module, four RPC endpoints on `PlatformLinkingManager`, a fire-and-forget facade on `BotBackend`, and instrumentation at every bot entry point (message received / reply sent / command used / stream error / guild joined-left).

**How** —
- **Privacy by schema.** `BotEvent` columns are: bounded enums (`eventType`, `channelType`, `errorKind`), a `commandName` slug, a `serverId`, plus two ints (`charCount`, `durationMs`). There is no `content`, no `userId`, no `username`. `BotGuild` stores `(platform, serverId, name, joinedAt, leftAt, lastSeenAt)`. The privacy guarantee is enforced by the columns not existing, not by code discipline.
- **Fire-and-forget writes.** `BotBackend.track_event/track_guild_*` schedule writes as background tasks with errors swallowed and logged at WARNING. A failing analytics write can never block or break a user's reply.
- **Presence reconciliation.** On `on_ready` the bot calls `sync_guilds()` which upserts every server it's currently in and marks any previously-joined server it isn't in any more as left — recovers state after the bot was down. `on_guild_join`/`on_guild_remove` keep the live count tracked in real time.
- **Routing.** Writes go through `bot_analytics_db()` in `db_accessors.py` — same pattern as `platform_linking_db`. In-process for the DatabaseManager pod, RPC via `DatabaseManagerAsyncClient` elsewhere.

This PR is **writes only**. The admin read API and dashboard ship in the stacked PR on top of this one.

### Changes 🏗️

- **Schema** (`schema.prisma` + migration `20260608120000_add_bot_analytics_tables`):
  - `BotEvent` — append-only event stream, indexed `(platform, eventType, createdAt)`, `(platform, serverId, createdAt)`, `(createdAt)`
  - `BotGuild` — presence row, unique on `(platform, serverId)`, indexed `(platform, leftAt)`
- **Writes**: `backend/data/bot_analytics.py` — `record_bot_event`, `record_guild_joined`, `mark_guild_left`, `sync_guild_presence`
- **Accessor**: `backend/data/db_accessors.py::bot_analytics_db()`
- **RPC**: 4 new `@expose` methods on `PlatformLinkingManager` + mirrored stubs on `PlatformLinkingManagerClient`
- **DB manager**: registers the same 4 methods on `DatabaseManager` and `DatabaseManagerAsyncClient`
- **Facade**: `BotBackend` — `track_event`, `track_guild_joined`, `track_guild_left`, `sync_guilds` (all fire-and-forget; errors logged not raised)
- **Instrumentation**:
  - `handler.py` — `message_received` (with `char_count`), `reply_sent` (with `char_count` + `duration_ms`), `stream_error` (with bounded `error_kind`)
  - `adapters/discord/commands.py` — `command_used` for every slash command
  - `adapters/discord/adapter.py` — `sync_guilds` on `on_ready`, `track_guild_joined` on `on_guild_join`, `track_guild_left` on `on_guild_remove`
- **Models**: `BotEventInput`, `BotGuildInput` in `platform_linking/models.py`

#### For code changes:
- [x] I have made a test plan
  - [x] Verified the bot runs and analytics calls don't block replies (a failing write logs a WARNING and the reply still goes out)
  - [x] Verified `on_ready` reconciles presence — joining/leaving a server while the bot was offline is reflected on next connect
  - [x] Verified slash commands record `command_used` rows with the correct command name
  - [x] Verified the schema literally cannot store message content or user identity (no such columns)
  - [x] `black`, `isort`, `ruff`, `pyright` all clean on changed files
  - [x] `bot_backend_analytics_test.py` (incl. "failure is swallowed" case) and `db_manager_test.py` pass

- [x] `.env.default` is updated or already compatible with my changes (no env changes)
- [x] `docker-compose.yml` is updated or already compatible with my changes (no service changes)
- [x] I have included a list of my configuration changes in the PR description (under **Changes**) — none required

### Privacy

The schema **cannot** store message content or user identity. The closest thing to a user-identifying field is `BotGuild.name` (server display name, best-effort). The closest thing to a message-content field is `BotEvent.charCount` (an integer). This is intentional and the whole reason the schema looks the way it does.

### Stacked

Read API + admin dashboard ship on top of this in the follow-up PR.

### DIFF
--- a/autogpt_platform/backend/backend/copilot/bot/adapters/discord/adapter.py
+++ b/autogpt_platform/backend/backend/copilot/bot/adapters/discord/adapter.py
@@ -211,6 +211,13 @@ class DiscordAdapter(PlatformAdapter):
             # rows that pre-date name capture. Cheap: in-memory cache only,
             # no Discord API calls.
             await self._refresh_known_server_names()
+            # Reconcile presence analytics: record every server we're currently
+            # in and mark any we've left while disconnected. Drives the admin
+            # server-count / sharding-prediction charts.
+            self._api.sync_guilds(
+                "discord",
+                [(str(guild.id), guild.name) for guild in self._client.guilds],
+            )
             # Sync slash commands once per process — on_ready fires on every
             # gateway reconnect, but the command tree only needs uploading once.
             if self._commands_synced:
@@ -225,6 +232,7 @@ class DiscordAdapter(PlatformAdapter):
         @self._client.event
         async def on_guild_join(guild: discord.Guild) -> None:
             await self._refresh_server_name(guild)
+            self._api.track_guild_joined("discord", str(guild.id), guild.name)
             channel = intro.pick_intro_channel(guild)
             if channel is None:
                 logger.info(
@@ -244,6 +252,12 @@ class DiscordAdapter(PlatformAdapter):
         async def on_guild_update(before: discord.Guild, after: discord.Guild) -> None:
             if before.name != after.name:
                 await self._refresh_server_name(after)
+
+        @self._client.event
+        async def on_guild_remove(guild: discord.Guild) -> None:
+            # Bot was kicked or the server was deleted — mark presence as left
+            # so the live server count and sharding prediction stay accurate.
+            self._api.track_guild_left("discord", str(guild.id))
 
         @self._client.event
         async def on_thread_remove(thread: discord.Thread) -> None:
--- a/autogpt_platform/backend/backend/copilot/bot/adapters/discord/commands.py
+++ b/autogpt_platform/backend/backend/copilot/bot/adapters/discord/commands.py
@@ -22,16 +22,26 @@ logger = logging.getLogger(__name__)
 def register(tree: app_commands.CommandTree, api: BotBackend) -> None:
     """Register all slash commands on the given CommandTree."""
 
+    def _track(interaction: discord.Interaction, command_name: str) -> None:
+        api.track_event(
+            platform="discord",
+            event_type="command_used",
+            server_id=str(interaction.guild.id) if interaction.guild else None,
+            command_name=command_name,
+        )
+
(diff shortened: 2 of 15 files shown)


Bentlybro:
Addressed the review pass — pushed as d8c085c19.

**Sentry — cancelled tasks raise CancelledError from the done-callback (LOW):** real bug, fixed. `_on_analytics_done` now early-returns on `task.cancelled()` before touching `task.exception()`. Added `test_cancelled_analytics_task_is_silent` which cancels a hanging analytics task and asserts the done-callback runs without raising.

**CodeRabbit — cache `Platform(platform.upper())` in `sync_guilds`:** taken. The enum is now computed once and reused for both the outer call and each `BotGuildInput`. Net win is readability (one source of truth) rather than perf.

**CodeRabbit — parallelize `sync_guild_presence` with `asyncio.gather`:** skipping. The bot is in single-to-low-double-digit guilds today and we'd shard the Discord gateway well before "thousands"; parallel upserts on a shared Prisma pool also risk contention on the unique constraint and cap our shutdown grace period for fire-and-forget tasks. Sequential is correct and predictable for the current scale — happy to revisit if the live server count starts approaching the sharding threshold.

ntindle (dismissed):

ntindle (approved):

PR #13243, opened 2026-05-29
feat(copilot): dream pass

ntindle:
## Why / What / How

Implements the first full **dream pass** for Copilot — a per-user, sleep-inspired memory consolidation system — plus the scheduling, batch-execution, billing, and admin tooling needed to run and observe it safely.

- **Why:** Long-term memory quality degrades without periodic maintenance. Raw chat episodes pile up, facts go stale, duplicates accumulate, and useful cross-session patterns are never surfaced. The dream pass runs offline (while the user is away) to consolidate stably-true facts, propose higher-order findings, retire contradicted memories, and keep the Graphiti knowledge graph healthy — improving warm-context retrieval without adding latency to live chat.
- **What:** A three-phase LLM pipeline (consolidate → recombine → sanitize) modeled on the NREM/REM sleep split, wrapped in per-user schedulers, a dedicated `BatchExecutor` subprocess for cheap async LLM batches, idempotent billing, Graphiti community support, and an admin Memory Visualizer UI. All outputs are `MemoryEnvelope` operations — no parallel type system.
- **How:** New `backend/copilot/dream/` module owns orchestration, fetch/ratify/apply, scheduling, routing, prompts/schemas, billing/pricing, and job-status polling. Long-running work is fire-and-forget with `JobStatus` polling; the heavy LLM calls can route through Anthropic's Batch API (50% discount) via a separate executor service. The whole system is gated behind LaunchDarkly flags and ships sync-baseline-first so it can roll out to a cohort before the batch path is widened.

## Types of dreaming 💤

There are three orthogonal axes to "a dream." This is the part of the PR worth understanding:

### 1. Cognitive phases (what happens *inside* one dream pass)

Every dream pass walks a user's recent memory window through three sequential LLM calls, mirroring the NREM/REM sleep split (`dream/orchestrator.py`, spec in `dream/dreaming-spec.md`):

| Phase | Sleep analogue | Model | Temp | Job |
|---|---|---|---|---|
| **1 — Consolidation** | NREM | `claude-sonnet-4-6` | 0.2 | Extract stably-true facts implicit in recent transcripts; flag existing memories to **demote**. High-confidence, written `active`. |
| **2 — Recombination** | REM | `claude-opus-4-7` + extended thinking | 0.9 | Find weak links across sessions no single session reveals; emit **tentative** findings/rules. Hyper-associative by design. |
[...]

PR #13332, opened 2026-06-11
fix(backend/copilot): stop empty untitled dream sessions polluting the chat list

ntindle:
## Why

The dream pass (#13243) creates a `ChatSession` per pass for its user-facing narrative. Three compounding issues made this user-visible noise: every nightly pass for a user with old facts but no new activity ran all three LLM phases, produced empty operations with an empty narrative, and still minted a session — which then leaked into the normal chat sidebar as an untitled chat containing only "Dream pass completed with no narrative output." Observed locally: one empty untitled chat per enabled user per night.

## What

- Dream-kind sessions are hidden from the user chat list, pagination count, and `/search/global` (they remain fetchable by id for the admin Memory Visualizer). Surfacing dreams in the chat UI stays future work (P6).
- `apply_operations` no longer creates a session or message when the pass produced zero operations.
- Dream sessions that ARE created get a dated title ("Dream summary — YYYY-MM-DD").
- The sanitize prompt's empty-result example no longer teaches the model to return an empty `summary_for_user`; a new hard rule requires a non-empty summary.
- Passes now skip before any LLM cost when the user has no episodes newer than their last completed dream (`dream:last_completed:{user_id}` Redis marker — single-key/cluster-safe, fail-open on Redis errors, stamped on the sync success path).

## How

- The list filter is raw SQL with `metadata->>'kind' IS DISTINCT FROM 'dream'` — `IS DISTINCT FROM` is load-bearing: legacy sessions have no `kind` key, `->>'kind'` evaluates to SQL `NULL`, and a naive `<>` would hide every normal chat. The Python Prisma client's `JsonFilter` has no path access, hence raw SQL via the existing `query_raw_with_schema` helper; list/count share one WHERE fragment so pagination totals always match. Pinned by mock tests on the WHERE construction plus real-Postgres integration tests for the NULL semantics.
- The empty-pass guard returns stats without a `session_id` key (both consumers read via `.get`, and the batch callback already narrows non-str to `None`), so no signature changes ripple into `batch_callbacks`.
- The no-new-activity skip compares episode `valid_at` (with `created_at` fallback) against the marker; unparseable timestamps count as new (fail-open). The batch path doesn't stamp the marker yet — batch users simply never skip (fail-open, noted in code).

## Checklist

- [x] Changes covered by tests (69 passing: mock + real-Postgres integration)
- [x] `poetry run format` / lint / pyright clean
- [x] Out-of-scope changes < 20%
- [x] `data/*.py` untouched; `copilot/db.py` queries are user-id scoped

Fixes SECRT-2480


autogpt_platform/backend/backend/copilot/dream/orchestrator.py
```diff
@@ -343,6 +431,34 @@ async def _execute_dream_pass_async(
                     skip_reason="no_input",
                 )
 
+            # No NEW activity since the last completed pass — every episode
+            # in the bundle predates the marker, so re-running all three
+            # LLM phases would only re-chew already-consolidated material
+            # (and, before the empty-pass guard in apply.py, manufacture an
+            # empty dream chat). Marker read is best-effort: missing,
+            # unparseable, or Redis-down all mean "run the pass".
+            last_completed = await _read_last_completed_marker(user_id)
```
ntindle: 🤖 🟠 **Should Fix**: **Admin manual 'dream now' trigger is silently neutered by the no-new-activity marker with no force path**

The marker skip applies to ALL callers of execute_dream_pass, including the admin trigger (POST /{user_id}/dream -> schedule_immediate_dream_pass -> execute_dream_pass(user_id, status_id=job_id)). Pre-PR, a manual trigger for a user with episodes+facts always ran the pass; post-PR it returns skipped/no_new_activity whenever the user's last sync pass postdates all episodes. This is invisible in the admin UI: MemoryVisualizer's jobStateSummary renders only status.state ('complete'), DreamResultPanel returns null for skipped results (no operations/usage keys), and skip_reason is not rendered anywhere in the admin memory frontend (verified by grep). So the primary memory-debugging tool — re-running a dream after changing prompts, LD flags (e.g. DREAM_PASS_INVALIDATE_ENTITY), or models — now silently no-ops, and the only workaround is waiting up to 35 days for the marker TTL or hand-deleting the dream:last_completed:{user_id} Redis key. status_id is non-None ONLY on the admin-trigger path (nightly batch and AgentProbe pass None, per the execute_dream_pass docstring), so a clean discriminator already exists.

**Suggested fix:** Bypass the marker for manual triggers: gate the skip on the caller, e.g. change the condition at orchestrator.py:441 to `if status_id is None and last_completed is not None and not _has_new_episodes_since(...)` (or, more explicitly, thread a `force_run: bool` kwarg from trigger_dream_pass through execute_dream_pass). Add an orchestrator test asserting that a pass invoked with status_id set runs the full pipeline even when the marker is newer than every episode.

autogpt_platform/backend/backend/copilot/dream/orchestrator.py
```diff
@@ -460,9 +576,15 @@ async def _execute_dream_pass_async(
                 ops,
                 known_fact_uuids=input_bundle.known_fact_uuids,
             )
+            # Apply succeeded (even as a no-op) — stamp the marker so the
+            # next nightly pass can skip when nothing new has landed.
+            # Sync path only: batch apply runs hours later in
+            # batch_callbacks, which doesn't stamp yet.
+            await _stamp_last_completed_marker(user_id)
```
ntindle: 🤖 🟡 **Nice to Have**: **Marker stamped with apply-completion time, not gather time — episodes landing mid-pass can wait up to 35 days to consolidate**

_stamp_last_completed_marker writes datetime.now() AFTER the three LLM phases + apply, which is minutes after gather_dream_input snapshotted the episode list. A user episode enqueued in that window (ingest.py sets reference_time/valid_at at enqueue) is (a) absent from the bundle this pass consolidated and (b) older than the stamped marker, so every subsequent nightly pass takes the no_new_activity skip. Because skipped passes never re-stamp, the 35-day TTL is a backstop — the episode is eventually consolidated when the marker expires — but for a user who goes dormant right after that message, their last episodes sit unconsolidated for up to 35 days. The current now()-after-apply choice exists to keep the dream's own enqueue_episode writes (valid_at ~= apply time) below the marker so they don't retrigger the pass nightly.

**Suggested fix:** Stamp the marker with input_bundle.window_end (the gather timestamp, threading it into _stamp_last_completed_marker) so anything arriving after gather counts as new, and prevent self-retrigger by excluding dream-authored episodes in _has_new_episodes_since — they are identifiable by the stable name convention from apply._episode_name (episode.name.startswith('dream_')) or source_description.startswith('dream-pass'). Add a test: episode with valid_at between gather time and stamp time must cause the next pass to run, while a dream_-named episode newer than the marker must not.

autogpt_platform/backend/backend/copilot/db.py
ntindle, line 1: 🤖 🟠 **Should Fix**: **Titled dream sessions now leak into the search-embedding backfill pipeline**

(Posted here because content_handlers.py isn't in this diff — the predicate belongs there.)

This PR gives dream sessions a non-empty title (apply.py: 'Dream summary — YYYY-MM-DD'). ChatSessionHandler.get_missing_items embeds every ChatSession with a non-empty title and has no kind filter, so each user's nightly dream session will be backfilled into UnifiedContentEmbedding; get_valid_content_ids (line 1010) also treats them as valid so the cleanup sweep never removes them. The dream exclusion added in copilot/db.py only covers get_user_chat_sessions/get_user_session_count — the comment there claims '/search/global stays consistent', but the embedding path is a separate ChatSession listing surface. Today nothing user-visible queries CHAT_SESSION embeddings (/search/global's chat bucket routes through the filtered get_user_chat_sessions, and _search_bucket only passes LIBRARY_AGENT/STORE_AGENT), so the impact is recurring embedding spend (one embed per active dream user per night, rows never cleaned up) plus a latent leak: _hybrid_row_to_item in search/service.py:105 still has a live chat_session branch, so re-enabling the semantic chats bucket would surface dream sessions in search results.

**Suggested fix:** Add the same null-safe dream exclusion to all three ChatSessionHandler queries in content_handlers.py: in get_missing_items append `AND (cs.metadata->>'kind' IS DISTINCT FROM 'dream')` to the WHERE; in both get_stats counts and in get_valid_content_ids append `AND (metadata->>'kind' IS DISTINCT FROM 'dream')`. Excluding them from get_valid_content_ids also makes the existing embedding-cleanup sweep delete any dream-session embeddings already created. Consider exporting _EXCLUDE_DREAM_SESSIONS_SQL from backend/copilot/db.py (or a shared constants module) so the predicate can't drift between the sidebar list and the embedding pipeline.

ntindle: 🤖 Fixed in [de797eefb](https://github.com/Significant-Gravitas/AutoGPT/commit/de797eefbee502be15700998b29ccad65261e859): all three ChatSessionHandler queries now carry the null-safe exclusion via a shared exclude_dream_sessions_sql() exported from copilot.db (so the predicate can't drift); excluding them from get_valid_content_ids also lets the cleanup sweep delete any dream embeddings already created. Test pins the predicate in all four SQL statements.

autogpt_platform/backend/backend/copilot/dream/orchestrator.py
ntindle: 🤖 Fixed in [49177b143](https://github.com/Significant-Gravitas/AutoGPT/commit/49177b143d649525588c0db327d71a18496176ed): manual triggers (the only callers setting status_id) now bypass the marker entirely — the skip doesn't even read it. Test asserts a status_id-bearing pass runs the full pipeline with a newer marker present.

autogpt_platform/backend/backend/copilot/dream/orchestrator.py
ntindle: 🤖 Fixed in [49177b143](https://github.com/Significant-Gravitas/AutoGPT/commit/49177b143d649525588c0db327d71a18496176ed): the marker is now stamped with input_bundle.window_end (the gather snapshot upper bound) instead of apply-completion now(), so episodes landing mid-pass count as new on the next tick. Stamp test asserts the exact window_end value.

autogpt_platform/backend/backend/copilot/dream/orchestrator.py
```diff
@@ -265,6 +275,84 @@ def _aggregate_usage(
...
+    try:
+        marker = datetime.fromisoformat(str(raw))
+    except ValueError:
+        logger.warning(
+            "Unparseable dream last-completed marker for user %s — running the pass",
+            user_id[:12],
+        )
+        return None
+    return marker if marker.tzinfo else marker.replace(tzinfo=timezone.utc)
+
```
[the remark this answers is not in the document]
ntindle: 🤖 Fixed in [49177b143](https://github.com/Significant-Gravitas/AutoGPT/commit/49177b143d649525588c0db327d71a18496176ed): the bytes decode now sits inside the parse guard with UnicodeDecodeError handled — corrupt marker bytes fail open and the pass runs. Regression test added (test_corrupt_marker_bytes_fail_open_and_run).

ntindle:
🤖 Addressed @coderabbitai's nitpick in [548366339](https://github.com/Significant-Gravitas/AutoGPT/commit/548366339808c8ce163b3fd4b2024a2894d24bde): ChatSessionHandler import hoisted to the module import block.

autogpt_platform/backend/backend/copilot/dream/orchestrator.py
```diff
@@ -265,6 +275,90 @@ def _aggregate_usage(
...
+def _has_new_episodes_since(episodes: list[EpisodeRow], marker: datetime) -> bool:
+    """True when any episode is newer than the last-completed marker.
+
+    An episode with no parseable timestamp counts as new — we can't
+    prove it's old, and skipping a pass we owed the user is worse than
+    running one we didn't.
+    """
+    for episode in episodes:
+        episode_at = parse_episode_timestamp(episode)
```
kcze, line 728: 🤖 🟠 **Should Fix**: `_has_new_episodes_since` counts the pass's own `dream_*` episodes as new activity, so the `no_new_activity` skip never fires after a *productive* pass. When a pass has writes/proposals, `apply_operations` enqueues `dream_{pass_id}_*` episodes with `valid_at = now()` (during apply, after `window_end`); the episode fetch in `fetch.py` (`MATCH (n:Episodic ...)`) has no name filter, and the marker is stamped with the earlier `window_end`. So on the next nightly run those pipeline-generated episodes satisfy `episode_at > marker`, triggering a full paid 3-phase pass with zero user activity — which re-reads and re-enqueues dream output, sustaining a paid pass every night. The skip therefore only ever helps already-empty passes. Exclude `dream_`-prefixed episodes from the novelty check (or track a source-only watermark). _(flagged by: Codex)_

autogpt_platform/backend/backend/copilot/dream/fetch.py
```diff
@@ -76,6 +76,37 @@ class DreamInput(BaseModel):
...
+def parse_episode_timestamp(episode: EpisodeRow) -> datetime | None:
+    """Best-effort tz-aware timestamp for an episode row.
+
+    ``valid_at`` is the reliable field: ingestion always sets it (the
+    ``reference_time`` passed at enqueue) and the episode fetch Cypher
+    orders by it. ``created_at`` is graphiti-core bookkeeping kept as a
+    fallback for legacy rows. ``None`` means neither field parsed —
+    callers must treat that as "age unknown" and fail open.
+    """
+    for raw in (episode.valid_at, episode.created_at):
```
kcze: 🤖 🟡 **Nice to Have**: `parse_episode_timestamp` returns `valid_at` (the enqueue `reference_time`), which can predate the stamped marker for an episode whose graph node is created by the async ingestion worker *after* the gather query. Such an episode is invisible at gather time yet has `valid_at < window_end`, so once the marker is stamped the next pass treats it as old and can skip consolidating it — the memory then waits for a newer episode (self-healing only while it stays inside the 14-day gather window; permanently dropped if the user goes idle long enough for it to age out). Consider using the later of `valid_at`/`created_at`, or an ingestion watermark, for the novelty comparison. _(flagged by: Codex)_

autogpt_platform/backend/backend/copilot/dream/orchestrator.py
kcze: 🤖 Fixed in b3e7d77: dream-authored episodes are now excluded from the no-new-activity check. `fetch.is_dream_authored_episode` identifies them by the stable `dream_` name prefix (shared `DREAM_EPISODE_NAME_PREFIX`, also used by `apply._episode_name`) with the `dream-pass` `source_description` as a secondary signal, and `_has_new_episodes_since` skips them. Without this, a productive pass's own `dream_{pass_id}_*` writes (enqueued with `valid_at = now()`, after the stamped `window_end`) counted as fresh activity, so every productive night triggered a full paid 3-phase pass the next night that only re-read its own output. Tests added in `orchestrator_test.py`: a `dream_`-prefixed episode newer than the marker does NOT count as new (skips), while a normal newer episode still runs even when a dream episode sits beside it.

autogpt_platform/backend/backend/copilot/dream/fetch.py
kcze: 🤖 Fixed in b3e7d77: `parse_episode_timestamp` now returns the *later* of `valid_at` and `created_at` (both parsed; `None` only when neither parses). For an episode whose graph node is materialized by the async ingestion worker AFTER the gather query ran, `valid_at` (the enqueue `reference_time`) can predate the stamped marker while `created_at` (graphiti's node-creation time) is after it — taking the max keeps it counted as new until a later pass actually consolidates it, instead of it being silently dropped once the marker moves past its `valid_at`. Safe to change here: `parse_episode_timestamp` is only used by the novelty check; gather ordering is done in Cypher (`ORDER BY n.valid_at DESC`), not through this function, so nothing shared is affected. Tests: a late-`created_at` episode (valid_at < marker < created_at) counts as new and runs; an episode old in BOTH fields still skips.

autogpt_platform/backend/backend/copilot/dream/apply.py
```diff
@@ -362,6 +370,24 @@ async def _create_dream_session(user_id: str, pass_id: str) -> str:
...
+        await chat_db().update_chat_session_title(
+            session_id=session_id, user_id=user_id, title=title
+        )
+    except Exception:
+        logger.warning(
+            "Failed to title dream session %s for user %s",
+            session_id[:12],
+            user_id[:12],
+            exc_info=True,
+        )
```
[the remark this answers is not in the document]
kcze: 🤖 Fixed in 6357730: both flagged log statements (the session-title warning and the no-operations info) now use f-strings, per the backend guideline that %s deferred interpolation is reserved for debug logs.

kcze (dismissed):

autogpt_platform/backend/backend/copilot/dream/apply.py
```diff
@@ -558,6 +592,29 @@ async def apply_operations(
...
+            f"no operations — skipping dream session creation"
+        )
+        return {
+            "consolidated_count": 0,
+            "proposal_count": 0,
+            "demotion_count": 0,
+            "demotion_failed_count": 0,
+            "entity_invalidation_count": 0,
+            # Vacuously drained — the pass enqueued nothing.
+            "ingestion_drained": True,
```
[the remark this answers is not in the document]
kcze: 🤖 Fixed in a7e82a3688 — real catch, this was a semantic merge conflict with #13338: the empty-pass early return still used the old `ingestion_drained` boolean key, so the fail-closed `drain_status_from_stats` reader coerced every empty pass to `timed_out`. Now returns `"ingestion_drain_status": IngestionDrainStatus.drained`. This was also the cause of the CI test failures (`test_no_enqueued_writes_skips_ingestion_drain`).

kcze (approved):

PR #13333, opened 2026-06-11
fix(frontend): harden paywall flows for half-dead sessions

0ubbe:
### Why / What / How

**Why:** Follow-up to the paywall bug reports in Discord ([thread](https://discord.com/channels/1126875755960336515/1512063871266062466/1514158652104376381), Linear SECRT-2431). On prod, an existing user hitting the paywall could end up in a "half-dead session" loop:

1. Clicking **Log out** on the paywall ran `/logout` → `serverLogout()`, but when Supabase `signOut()` fails (revoked/already-rotated refresh token), the error was swallowed and **the auth cookies survived**. Middleware still saw a session, so `/login` bounced the user straight back into the app — paywall re-mounted, navbar said "Log In", and every API call failed.
2. Clicking a plan in that state hit `POST /credits/subscription` without a token → dead-end toast *"Couldn't start checkout — Authorization header is missing"* (visible at 1:15 in the report video), and the modal degraded to the *"Subscriptions are temporarily unavailable"* screen with no way to recover.
3. Backing out of Stripe Checkout landed on `/profile/credits` (`cancel_url`), which is paywall-exempt — so the user could wander the whole `/profile` area looking "unlocked" until the next non-exempt navigation slammed the modal again.
4. The paywall's Log out button could be cropped/scrolled away at the top of the dialog's scroll container.

**What:** Kill the session-resurrection loop at the root, make the paywall self-heal when its session dies, and fix the checkout-cancel and logout-button UX.

**How:**
- `serverLogout()` now **always expires `sb-*` auth cookies**, even when `signOut()` errors or throws — a failed Supabase API call can no longer leave a live cookie session behind.
- `usePaywallModal` re-validates the session on mount (`validateSession()`): an invalid session clears the user, unmounts the gate, and redirects to `/login` instead of rendering a zombie paywall.
- A 401 from the upgrade mutation routes to `/login` instead of toasting a dead-end error.
- Stripe Checkout `cancel_url` is now the page the user was on, so backing out re-gates immediately instead of dropping into the paywall-exempt `/profile/credits`.
- The Log out button is `sticky` in the dialog scroll viewport (can't be cropped or scrolled out), and the "temporarily unavailable" state gained a **Retry** button.

### Changes 🏗️

- `lib/supabase/actions.ts` — `serverLogout` no longer returns early on sign-out failure; always deletes `sb-*` cookies via `next/headers`
- `PaywallGate/usePaywallModal.ts` — session re-validation on mount; 401 → `/login`; `cancel_url` = current page; exposes `retryLoadPlans`
- `PaywallGate/PaywallModal.tsx` — sticky Log out button; Retry button on the empty-plans fallback
- Tests: new `lib/supabase/__tests__/actions.test.ts` (4 cases) and 4 new cases in `PaywallModal.test.tsx` (session validation, cancel_url, 401 redirect, Retry)

#### For code changes:
- [x] I have made a test plan
  - [x] `pnpm vitest run PaywallModal.test.tsx actions.test.ts` — 25/25 pass (incl. 8 new cases)
  - [x] `pnpm vitest run PaywallGate.test.tsx useSupabaseStore.test.ts` — 12/12 pass (no regressions in adjacent suites)
  - [x] `pnpm format`, `pnpm lint`, `pnpm types` all clean
  - [ ] Manual prod-like verification of the logout → /login flow with an invalidated refresh token (not reproducible locally — see Discord thread; needs an account in the in-between state)

### DIFF
--- a/autogpt_platform/frontend/src/app/(platform)/PaywallGate/PaywallModal.tsx
+++ b/autogpt_platform/frontend/src/app/(platform)/PaywallGate/PaywallModal.tsx
@@ -17,6 +17,7 @@ export function PaywallModal() {
     isLoading,
     plans,
     retryLoadPlans,
+    isRetryingPlans,
     country,
     isYearly,
     selectedCycle,
@@ -118,6 +119,7 @@ export function PaywallModal() {
                   variant="secondary"
                   size="small"
                   onClick={retryLoadPlans}
+                  loading={isRetryingPlans}
                 >
                   Retry
                 </Button>
--- a/autogpt_platform/frontend/src/app/(platform)/PaywallGate/PaywallModal.tsx
+++ b/autogpt_platform/frontend/src/app/(platform)/PaywallGate/PaywallModal.tsx
@@ -16,6 +16,7 @@ export function PaywallModal() {
   const {
     isLoading,
     plans,
+    retryLoadPlans,
     country,
     isYearly,
     selectedCycle,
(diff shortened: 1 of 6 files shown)


0ubbe:
Tried on dev preview and fixes it:

https://github.com/user-attachments/assets/156e4fd9-d71e-4d59-9e8a-ad57e9ca49ad

0ubbe:
Addressed in eedd60740:

1. **Session validation failure → redirect test** — added, but at the store layer rather than `PaywallModal.test.tsx`. The modal only fires `validateSession()` and intentionally ignores the result; the redirect on an invalid session lives in `useSupabaseStore.validateSession` (`router.push(redirectPath)` after clearing the user), which the modal test mocks out. Asserting `router.replace('/login')` in the modal test would pass/fail against behavior that doesn't exist at that layer. New tests in `useSupabaseStore.test.ts` cover both the invalid path (user cleared + redirected to `/login?next=…`) and the valid path (user kept, no redirect). The modal-level test asserting `validateSession` fires on mount remains the glue between the two.

2. **Non-401 ApiError coverage** — added: `ApiError('Forbidden', 403, null)` from `updateTier` toasts and does not redirect or navigate.

3. **Retry loading state** — `usePaywallModal` now exposes `isRetryingPlans` (from `isFetching`) and the Retry button uses the design-system `loading` prop, which disables it while the refetch is in flight. Test added.

`pnpm format` / `lint` / `types` clean; 34/34 affected tests pass.

PR #13343, opened 2026-06-11
dx(backend): architecture rule — Prisma-less processes must route DB access via db_accessors

ntindle:
## Why

We shipped two production incidents in the same week from the same bug class: code running in Prisma-less processes (copilot-executor, scheduler) doing direct Prisma work — `dream/scheduling.py`'s timezone lookup (`ClientNotConnectedError` on every call → dream crons never registered for anyone) and `copilot/rate_limit.py`'s Stripe tier reconcile (permanently dead self-heal, nightly stack traces). Only REST, WS, and the DatabaseManager ever call `db.connect()`; everything else must route through `backend.data.db_accessors`. Nothing enforced that.

## What

A new rule in `backend/util/architecture_test.py` (following the file's existing pattern: motivation-documented, burn-down allowlist):

- **Layer A (syntactic):** no `X.prisma()` model-accessor calls or global prisma-client calls under `backend/copilot/**` / `backend/executor/**`, outside the sanctioned implementation modules.
- **Layer B (symbol-level):** no `from backend.x import f` where `f`'s body performs Prisma operations. Symbol-level is the key design choice — data modules legitimately export pure types/helpers alongside Prisma functions, so module-level import flagging would be hopelessly over-broad (a prototype flagged 74 sites; symbol-level flags 11, all real).

Sanctioned implementations are derived automatically from what `db_manager.py` (the RPC service) imports — no hand-maintained module list.

## How

- Allowlist seeded with the current state: the two #13339 entries (remove on merge), a test-data seeder, and **5 latent instances of the incident class** in copilot tools (`list_agent_triggers`/`manage_schedules` → library.db, `pending_message_helpers` → workspace, `rate_limit` → `get_user_by_id`) — each annotated with its migration target. Burn down; do not add.
- TYPE_CHECKING imports and `*_test.py` excluded.
- Negative-validated: removing an allowlist entry makes the test flag the exact line of the real (since-fixed) Stripe bug.

## Checklist

- [x] Rule documented with the motivating incidents per the file's convention
- [x] `poetry run format` clean; all architecture tests pass
- [x] No out-of-scope changes

---

> [!NOTE]
> **Low Risk**
> Test-only change in architecture_test.py; no runtime behavior. May fail CI until known offenders are migrated or remain on the allowlist.
> 
> **Overview**
> Adds an **architecture test** so `backend/copilot/**` and `backend/executor/**` cannot use Prisma directly in processes that never call `db.connect()` (scheduler, copilot-executor, batch-executor). New violations are meant to be routed through `backend.data.db_accessors` instead.
> 
> **`test_prisma_less_processes_route_db_access_via_accessors`** scans production Python with AST helpers: it flags direct `X.prisma()` / global `prisma.*` calls, and **named imports** of functions whose bodies contain Prisma (not whole-module imports). `TYPE_CHECKING` blocks and `*_test.py` are skipped; modules imported by `db_manager.py` are treated as sanctioned RPC implementations.
> 
> A burn-down **`_PRISMA_ISOLATION_KNOWN_OFFENDERS`** allowlist seeds current copilot offenders (rate limit, tools, pending messages, test data). **`test_known_offenders_use_posix_separators`** now validates those entries use forward slashes like the other allowlists.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a0f9bb79de08663f5f3f5b71c4a6a43e34eef291. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/autogpt_platform/backend/backend/util/architecture_test.py
+++ b/autogpt_platform/backend/backend/util/architecture_test.py
@@ -192,7 +192,11 @@ def test_known_offenders_use_posix_separators():
 
     Ensure the key construction normalises to forward slashes.
     """
-    for entry in _KNOWN_OFFENDERS | _DATACLASS_KNOWN_OFFENDERS:
+    for entry in (
+        _KNOWN_OFFENDERS
+        | _DATACLASS_KNOWN_OFFENDERS
+        | _PRISMA_ISOLATION_KNOWN_OFFENDERS
+    ):
         path_part = entry.split()[0]
         assert "\\" not in path_part, (
             f"known-offenders entry uses backslash: {entry!r}. "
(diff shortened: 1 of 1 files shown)


autogpt_platform/backend/backend/util/architecture_test.py
```diff
@@ -233,3 +237,188 @@ def test_no_process_cached_loop_bound_clients():
...
+#      the global ``prisma`` client instance;
+#   B) symbol-level: ``from backend.x import f`` where ``f``'s body
+#      performs Prisma operations (module-level imports of mixed modules
+#      are fine — most data modules also export pure types/helpers).
+#
+# Modules imported by ``backend/data/db_manager.py`` are the sanctioned
+# implementations (they execute inside the DatabaseManager process or
+# behind ``db.is_connected()`` checks) and are exempt.
+
+_PRISMA_LESS_PACKAGES = ("copilot", "executor")
```
ntindle, line 274: I dont' love how this is done. this should def be many more packages to start like scheduler and also the offenders that are known are concerning. i'd take this whole thing bakc to the drawing board

PR #13371, opened 2026-06-17
fix(frontend): scope static font CORS to our own origin

ntindle:
> **Draft — needs a post-deploy header check before merge (see Testing).**

### Why / What / How

**Why:** The TAC Security CASA (ESOF AppSec ADA) assessment of `platform.agpt.co` flags an "overly permissive CORS policy" (finding #1, CWE-264): self-hosted Next.js fonts under `/_next/static/media/*` respond with `Access-Control-Allow-Origin: *`. All CASA findings must be cleared for the Tier 2 & Tier 3 assessments (currently re-expiring).

**What:** Scope the `Access-Control-Allow-Origin` header on the font path from `*` to our own origin.

**How:** Adds a `nextConfig.headers()` rule for `/_next/static/media/:path*` setting `Access-Control-Allow-Origin: https://platform.agpt.co`. These fonts are only loaded by our own same-origin pages, so a same-origin fetch is unaffected by the header value — this should not change how fonts render.

### Changes 🏗️

- `autogpt_platform/frontend/next.config.mjs`: add `headers()` rule restricting font CORS to our origin.

### ⚠️ Reviewer note — must verify after deploy

Vercel's "Ignored Build Step" cancels the preview build for this PR, so the header could **not** be verified pre-merge. The header on `/_next/static/media/*` is currently emitted by the Vercel/Next.js static layer; this config override may or may not replace it. **Please confirm on a real deploy (dev/staging) before merging:**

```bash
curl -sI "https://<deploy-url>/_next/static/media/<any>.woff2" | grep -i access-control-allow-origin
# want: a single "access-control-allow-origin: https://platform.agpt.co"
# NOT:  "*", and NOT two ACAO headers (duplicate = the override didn't replace it)
```

If the override does **not** take (still `*` or duplicated), the fallback is a Vercel project-level header rule or accepting it with a CASA justification (fonts are non-sensitive, same-origin only). Failure mode is safe either way: fonts load the same regardless, since they're fetched same-origin.

This overlaps `next.config.mjs` with #13370 (baseline security headers); whichever merges second will need a trivial `headers()` merge.

- [x] `.env.default` / `docker-compose.yml` unaffected (no env/compose changes)
- [x] Configuration changes listed above under **Changes**
- [x] `pnpm format` / `pnpm types` pass locally
- [ ] **Post-deploy:** font `Access-Control-Allow-Origin` is the origin (not `*`, not duplicated)
- [ ] **Post-deploy:** app renders and fonts load normally


autogpt_platform/frontend/next.config.mjs
```diff
@@ -88,15 +88,23 @@ const nextConfig = {
...
-        source: "/:path*",
+        source: "/_next/static/media/:path*",
         headers: [
           { key: "X-Content-Type-Options", value: "nosniff" },
           { key: "X-Frame-Options", value: "SAMEORIGIN" },
           // Enables Sentry browser JS self-profiling.
           { key: "Document-Policy", value: "js-profiling" },
+          {
+            key: "Access-Control-Allow-Origin",
+            value: "https://platform.agpt.co",
```
0ubbe, line 118: I think we should also allow `https://dev-builder.agpt.co` otherwise the fonts won't render in Dev ( _causing it look different than Prod_ ) ?

0ubbe (approved):

PR #13379, opened 2026-06-18
fix(frontend): keep AutoPilot middle messages when tail window slides

Abhi1992002:
Resolves SECRT-2424.

### Why / What / How

**Why** — In AutoPilot, messages in the middle of a conversation silently disappeared from the rendered view during a multi-turn session (no refresh). Scrolling up did **not** bring them back; only a full page refresh restored them. The messages were never lost on the backend — purely a client-side rendering hole.

**What** — Stop the post-turn force-hydrate from dropping in-memory messages that are older than the refetched window.

**How** — AutoPilot displays `concatWithAssistantMerge(pagedMessages, currentMessages)`:
- `GET /sessions/{id}` returns only the **most-recent ~50 messages** — a sliding *tail window* (`backend/copilot/db.py`, `limit=50`).
- After every turn (`status: streaming → ready`), `useHydrateOnStreamEnd` **force-replaced** `currentMessages` with that window.
- As the conversation grows the window slides forward (its `oldest_sequence` increases), so the replace drops every in-memory message older than the new window's oldest sequence.
- Meanwhile `pagedMessages` (older history the user scrolled back to) is intentionally preserved across refetches and only ever extends **older** (`before_sequence`), so it never covers the newly-vacated region.
- Result: a hole between the top of `pagedMessages` and the bottom of the new window. Scroll-back can't fill a *middle* hole, so the messages stay gone until refresh rebuilds both sources contiguously.

The fix adds `retainOlderHistory(prev, hydrated)` in `useHydrateOnStreamEnd`: before force-replacing, it keeps the `prev` messages whose DB sequence predates the window's oldest sequence and prepends them, keeping the result contiguous with the older history. Streaming rows (AI-SDK uuids with no `-seq-N` id) are excluded — they're already inside the refetched window, so there are no duplicates.

### Changes 🏗️

- `useHydrateOnStreamEnd.ts`: new `retainOlderHistory` helper; the force-hydrate path now retains older-than-window messages instead of dropping them.
- `helpers/convertChatSessionToUiMessages.ts`: export the existing `extractDbSequence` (`-seq-N` parser) so the hook reuses it instead of duplicating the regex.
- `__tests__/useHydrateOnStreamEnd.test.ts`: **new** regression coverage (the hook previously had none).

#### For code changes:
- [x] I have made a test plan
  - [x] `pnpm format`, `pnpm lint`, `pnpm types` — clean
  - [x] `pnpm test:unit` for the new test — 3/3 pass; changed hook coverage 71% stmts / 78% lines (>70%)
  - [x] Full copilot suite (`src/app/(platform)/copilot`) — 1200/1200 tests pass, no regressions
  - [x] Manual: long AutoPilot session (>50 messages), scroll up to load older history, send another message, confirm middle messages remain after the turn completes (no refresh)

### Test plan to reproduce the original bug

1. Open an AutoPilot session with enough turns that total persisted messages exceed 50.
2. Scroll up so older history loads (you can see the first message).
3. Send another message and let the turn finish — **without refreshing**.
4. Before this fix: messages between the loaded older pages and the recent window are missing and scroll-back won't recover them. After this fix: the conversation stays contiguous.


autogpt_platform/frontend/src/app/(platform)/copilot/useHydrateOnStreamEnd.ts
```diff
@@ -7,9 +7,48 @@ import {
...
+  prev: UIMessage[],
+  hydrated: UIMessage[],
+): UIMessage[] {
+  let oldestHydratedSeq: number | null = null;
+  for (const message of hydrated) {
+    const seq = extractDbSequence(message);
+    if (seq === null) continue;
+    if (oldestHydratedSeq === null || seq < oldestHydratedSeq) {
+      oldestHydratedSeq = seq;
+    }
```
Pwuts: If `hydrated` is sorted this can be a lot simpler

autogpt_platform/frontend/src/app/(platform)/copilot/useHydrateOnStreamEnd.ts
```diff
@@ -7,9 +7,48 @@ import {
...
+  }
+  if (oldestHydratedSeq === null) return hydrated;
+  const windowStart = oldestHydratedSeq;
+
+  const olderTail = prev.filter((message) => {
+    const seq = extractDbSequence(message);
+    return seq !== null && seq < windowStart;
+  });
```
Pwuts: If `prev` is sorted this can be an array slice

Pwuts (dismissed):
Probably works but technical improvement recommended

autogpt_platform/frontend/src/app/(platform)/copilot/useHydrateOnStreamEnd.ts
Abhi1992002: Fixed in 93e55246cd: `hydrated` is in ascending DB-sequence order, so the window's oldest sequence is just the first `-seq-N` row. Replaced the min-scan loop with a first-seq lookup (still skips idx-fallback rows that carry no sequence).

autogpt_platform/frontend/src/app/(platform)/copilot/useHydrateOnStreamEnd.ts
Abhi1992002: Fixed in 93e55246cd: `prev` is ascending too, so the older-than-window messages are a leading run. Replaced the full `filter` with `findIndex` for the window boundary + `prev.slice(0, boundary)`. Added a no-overlap regression test for the all-prev-older branch (`findIndex === -1`).

Pwuts (approved):
LGTM :)

PR #13381, opened 2026-06-18
fix(backend): surface sub-agent workspace files in run_sub_session result

Abhi1992002:
### Why / What / How

**Why** — A `run_sub_session` sub-agent (Otto/AutoPilot) that does substantial work and then *delivers it by writing workspace files* — summarising only briefly in its final message — returns a hollow body to the parent (SECRT-2377). Confirmed in production: a sub made 44 tool calls and wrote ~37 KB of findings across 3 workspace files, then returned 227 tokens of *"delivered in three workspace documents, what's next?"*. The parent had no visibility into those files, treated the run as empty, and re-ran the entire task — ~22.6k tokens wasted, no error surfaced.

Root cause is a **contract gap**: `response_from_outcome` carried only the sub's final assistant text (`response`) + a tool-call log. Workspace files are scoped to the *sub's* session (the parent lists its own session by default), and there was no field enumerating files the sub produced. So "deliver via files + summarise" was silent data loss.

**What** — Add a `sub_workspace_files` manifest to `SubSessionStatusResponse`, populated on completion, so the parent can recover work delivered via files. Each entry has `file_id`, `name`, fully session-qualified `path`, and `size_bytes` — `path` is directly usable with `read_workspace_file(path=...)`.

**How** — Two sources, authoritative-first:
- **`list_sub_workspace_files`** (authoritative) reads the sub's session for `origin=agent-created` files. This captures writes from **any** turn — including the already-terminal / cold-poll path in `get_sub_session_result`, where the rebuilt tool-call log reflects only the sub's last message and would otherwise miss earlier writes. Returns `None` on lookup failure; `[]` means the sub wrote nothing. Capped at 50 entries.
- **Tool-call mining** (`_workspace_files_from_tool_calls`) is the fallback when the listing is unavailable, parsing `write_workspace_file` outputs defensively (JSON string on live-drain, dict on persisted-replay).

Both `run_sub_session._execute` and `get_sub_session_result._execute` fetch the listing on `completed` and pass it through `response_from_outcome`. The completed message also nudges the parent toward the files so a hollow `response` isn't mistaken for an empty run. (Stored `WorkspaceFile.path` is already session-qualified via `_resolve_path` on write, so no re-prefixing is needed.)

### Changes 🏗️

- `models.py`: new `SubWorkspaceFileInfo` model; `sub_workspace_files: list[...] | None` field on `SubSessionStatusResponse`.
- `run_sub_session.py`: `list_sub_workspace_files` (authoritative), `_workspace_files_from_tool_calls` + `_as_payload` (fallback miner), `workspace_files` override param on `response_from_outcome`, completed-message nudge; `_execute` fetches the listing on completion.
- `get_sub_session_result.py`: fetches the listing on completion and passes it through — fixes the cold-poll/terminal path.
- `sub_session_test.py`: repro at `response_from_outcome` level + terminal-path and live-path `_execute` coverage; autouse fixture stubs the listing so no test hits the workspace DB.

#### For code changes:
- [x] I have made a test plan
  - [x] `poetry run pytest backend/copilot/tools/sub_session_test.py` — 22 passed
  - [x] Repro: a completed sub that wrote files but returned a hollow `response` now surfaces `sub_workspace_files` with session-qualified paths
  - [x] Terminal/cold-poll path (`get_sub_session_result`, last message terminal, waiter skipped) still populates the manifest from the authoritative listing
  - [x] A sub that answered inline (no writes) yields `sub_workspace_files = None` (no noise)
  - [x] `poetry run format` and `poetry run lint` (ruff, isort, black, pyright) clean


autogpt_platform/backend/backend/copilot/tools/models.py
```diff
@@ -343,6 +343,27 @@ class SubSessionProgressSnapshot(BaseModel):
...
+
+    file_id: str = Field(description="Workspace file id written by the sub.")
+    name: str = Field(description="Filename the sub wrote.")
+    path: str = Field(
+        description=(
+            "Fully-qualified virtual path (e.g. /sessions/<sub_id>/report.md). "
+            "Pass directly to read_workspace_file(path=...) to read the file."
+        ),
+    )
+    size_bytes: int = Field(description="Size of the written file in bytes.")
```
Pwuts: This seems to have overlap with the models in `tools/workspace_files.py

autogpt_platform/backend/backend/copilot/tools/run_sub_session.py
```diff
@@ -182,18 +199,114 @@ def _sub_session_link(inner_session_id: str | None) -> str | None:
...
+        files = await manager.list_files(
+            limit=_WORKSPACE_FILE_MANIFEST_LIMIT,
+            metadata_equals={"origin": "agent-created"},
+        )
+    except Exception:
+        logger.warning(
+            "Failed to list workspace files for sub %s",
+            inner_session_id[:12],
+            exc_info=True,
+        )
```
Pwuts, line 235: nit:
```suggestion
        logger.warning(
            f"Failed to list workspace files for sub {inner_session_id[:12]}",
            exc_info=True,
        )
```

autogpt_platform/backend/backend/copilot/tools/run_sub_session.py
```diff
@@ -48,6 +55,10 @@
 # 2x headroom holds uniformly.
 MAX_SUB_SESSION_WAIT_SECONDS = MAX_TOOL_WAIT_SECONDS
 
+# Cap on how many sub-written files we enumerate in a completed response's
+# manifest. Bounds context size; a sub producing more than this is pathological.
+_WORKSPACE_FILE_MANIFEST_LIMIT = 50
```
Pwuts, line 60: Suppose the sub-agent clones a git repo, inspects/uses it, and synthesizes into a few .md files. Which 50 files will be returned?

autogpt_platform/backend/backend/copilot/tools/models.py
Abhi1992002: Good catch — consolidated. Moved `WorkspaceFileInfoData` into `tools/models.py` (shared, no circular import since models.py doesn't import workspace_files.py), dropped `SubWorkspaceFileInfo`, and `workspace_files.py` now imports it. The sub manifest reuses the same model (now also carries `mime_type`). Fixed in 74ee43f515.

autogpt_platform/backend/backend/copilot/tools/run_sub_session.py
Abhi1992002: Done, switched to the f-string form (matches the AGENTS.md convention — %s only for debug). Fixed in 74ee43f515.

autogpt_platform/backend/backend/copilot/tools/run_sub_session.py
Abhi1992002: The cloned repo files won't appear — the listing filters `origin=agent-created`, which is set only on files persisted via `write_workspace_file`. A `git clone` lands in the sub's ephemeral working dir (sandbox), not in persistent workspace storage, so only the few synthesized `.md` files the sub explicitly persisted are returned. If the sub somehow persists >50 such files, the listing is ordered `createdAt desc`, so the newest 50 win. Documented both points in `list_sub_workspace_files` in 74ee43f515.

Pwuts: ok excellent :)

Pwuts (approved):
LGTM!

PR #13382, opened 2026-06-18
fix(backend): Make connection retries abort on shutdown

Pwuts:
### Why / What / How

**Why:** When a backing service (Redis cluster, RabbitMQ, an RPC peer) is unreachable, the connection-acquisition retry loops kept the whole backend from shutting down. `conn_retry` and `create_retry_decorator` retry up to ~100 times with 30s exponential backoff (`pyro_client_comm_retry=100`, `pyro_client_max_wait=30`) — roughly 50 minutes — with **no shutdown awareness**. The backoff sleeps run on worker/background-event-loop threads that never receive SIGINT/SIGTERM (only the main thread does), and the eager `await redis_client.get_redis_async()` in `AppService.lifespan` keeps Uvicorn stuck in *startup*, so its graceful-shutdown path never runs. Net effect: `Ctrl+C` is ignored while a service is retrying, and the logs spam "Acquiring connection failed … Retrying now…" indefinitely.

**What:** Make the retry loops abort promptly when the process receives a shutdown signal.

**How:**
- Added a process-wide shutdown flag in `backend/util/retry.py` — `request_shutdown()` / `is_shutting_down()` backed by a `threading.Event` (signal-safe: it only sets an event).
- Added a tenacity `_StopOnShutdown` stop condition (combined via `|` with the existing `stop_after_attempt`) so no new retry is attempted once shutdown is requested.
- Added interruptible sleeps (`_interruptible_sleep` / `_interruptible_async_sleep`) that wake immediately on shutdown instead of sleeping out the full backoff. `conn_retry` and `create_retry_decorator` pick the right one based on whether the wrapped function is sync or async.
- Both signal handlers — `AppProcess._self_terminate` and `AppService._self_terminate` — now call `request_shutdown()` first.

On the first `Ctrl+C`, the in-flight retry wakes from its backoff, the stop condition fires, `connect_async()` raises, the lifespan startup fails fast, Uvicorn proceeds to shut down, and the process exits.

### Changes 🏗️

- `backend/util/retry.py`: add `request_shutdown()`/`is_shutting_down()`, `_StopOnShutdown` tenacity stop condition, and `_interruptible_sleep`/`_interruptible_async_sleep`; wire them into `conn_retry` and `create_retry_decorator` (the latter now picks the sync/async interruptible sleep at decoration time). Final-failure log distinguishes "aborted: shutting down" from "failed after retries".
- `backend/util/process.py`: `AppProcess._self_terminate` calls `request_shutdown()`.
- `backend/util/service.py`: `AppService._self_terminate` calls `request_shutdown()` so the eager Redis connect in `lifespan` stops blocking Uvicorn startup.
- `backend/util/retry_test.py`: 5 new tests (sync/async `conn_retry` abort on shutdown, `create_retry_decorator` abort, both interruptible sleeps return early) + an autouse fixture that resets the shutdown flag between tests.

#### For code changes:
- [x] I have made a test plan
  - [x] `poetry run pytest backend/util/retry_test.py` — 19 passed (14 existing + 5 new shutdown tests)
  - [x] `poetry run ruff check` + `poetry run pyright` clean on all changed files; pre-commit hooks (ruff/isort/black/pyright) pass
  - [x] `backend/util/service_test.py::test_graceful_shutdown`

### DIFF
--- a/autogpt_platform/backend/backend/util/process.py
+++ b/autogpt_platform/backend/backend/util/process.py
@@ -8,7 +8,7 @@ from typing import Optional
 
 from backend.util.logging import configure_logging
 from backend.util.metrics import sentry_init
-from backend.util.retry import request_shutdown
+from backend.util.retry import stop_retry_loops
 from backend.util.settings import set_service_name
 
 logger = logging.getLogger(__name__)
@@ -94,8 +94,7 @@ class AppProcess(ABC):
         os.write(sys.stdout.fileno(), (message + "\n").encode())
 
     def _self_terminate(self, signum: int, frame):
-        # Abort in-flight connection-retry loops so they don't block shutdown.
-        request_shutdown()
+        stop_retry_loops()
         if not self._shutting_down:
             self._shutting_down = True
             sys.exit(0)
--- a/autogpt_platform/backend/backend/util/process.py
+++ b/autogpt_platform/backend/backend/util/process.py
@@ -8,6 +8,7 @@ from typing import Optional
 
 from backend.util.logging import configure_logging
 from backend.util.metrics import sentry_init
+from backend.util.retry import request_shutdown
 from backend.util.settings import set_service_name
 
 logger = logging.getLogger(__name__)
@@ -93,6 +94,8 @@ class AppProcess(ABC):
         os.write(sys.stdout.fileno(), (message + "\n").encode())
 
     def _self_terminate(self, signum: int, frame):
+        # Abort in-flight connection-retry loops so they don't block shutdown.
+        request_shutdown()
         if not self._shutting_down:
             self._shutting_down = True
             sys.exit(0)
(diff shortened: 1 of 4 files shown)


Pwuts:
🤖 Addressed the CodeRabbit nitpick in [d322dbbb5](https://github.com/Significant-Gravitas/AutoGPT/commit/d322dbbb5aa55c636afaf6070923544c7a34a157): `continuous_retry` now checks `is_shutting_down()` in its loop condition and uses the interruptible sleeps, so the infinite consumer loops (RabbitMQ run/cancel consumers, notification service) exit promptly on SIGINT/SIGTERM instead of blocking shutdown.

One deviation from the suggested diff: on shutdown the loop **returns cleanly** rather than raising `RuntimeError`. These consumers already use `return` for their graceful-stop paths (e.g. `if self.stop_consuming.is_set(): return`), so a clean return is consistent and avoids spurious tracebacks during a normal Ctrl+C. Added sync + async tests covering it.

0ubbe (approved):

PR #13386, opened 2026-06-18
fix(backend/copilot): make reasoning + community-rebuild tests transport-hermetic

Pwuts:
## Why

Six pre-existing copilot tests fail on **every fork PR** while staying green on `dev`:

- `backend/copilot/baseline/service_unit_test.py::TestBaselineReasoningStreaming::test_reasoning_param_absent_on_non_anthropic_routes`
- `…::test_reasoning_param_suppressed_when_thinking_tokens_zero`
- `backend/copilot/graphiti/communities_test.py::TestRebuildCommunitiesForUser::test_success_path_calls_detach_delete_then_build`
- `…::test_failure_path_returns_error_in_result`
- `…::TestRebuildForUserActivityGate::test_force_bypasses_gate`
- `…::TestRebuildPathSelection::test_uses_flex_client_by_default`

**Root cause:** these tests implicitly assume the ambient `ChatConfig` transport is `openrouter`. `openrouter_active` is only true when `api_key` is non-empty, and `api_key` falls back to `OPENAI_API_KEY`. GitHub Actions **withholds repository secrets from `pull_request` runs triggered by forks**, so `OPENAI_API_KEY` is empty there and the transport silently drops to `direct_anthropic`:

- graphiti: `use_flex = flex_requested and chat_cfg.transport.supports_flex_tier`; only `openrouter` has `supports_flex_tier=True`, so the sync path runs the *real* client and the tests that only patch `make_flex_graphiti_client` fall through (wrong shape / no error / `execution_path == "sync"`).
- baseline: without openrouter, `extra_body` isn't built → `KeyError: 'extra_body'`.

`dev` is green because its CI runs as `push` events, where secrets are present. This surfaced on the JSON-blocks PR (#13170) purely because it was a fork PR — its code changes are unrelated.

## What

Make the affected tests hermetic by pinning the OpenRouter transport, so they pass regardless of credential availability:

- **`service_unit_test.py`** — pin `use_openrouter` / `api_key` / `base_url` in the two reasoning tests that relied on `extra_body` being built (matching the existing sibling pattern, e.g. the Kimi test).
- **`communities_test.py`** — add an autouse fixture pinning the transport so the flex-tier path is taken deterministically. The flag-driven sync test (`test_uses_sync_client_when_flex_disabled`) is unaffected — it forces sync via `community_rebuild_use_flex_tier`, not the transport.

## How

Verified both files in both credential states (`-p no:cov`, 138 tests each):

```bash
# Fork-CI simulation (secrets cleared) — previously 6 failures, now:
env CHAT_API_KEY= CHAT_BASE_URL= OPENAI_API_KEY= OPEN_ROUTER_API_KEY= OPENAI_BASE_URL= \
  poetry run pytest backend/copilot/baseline/service_unit_test.py \
  backend/copilot/graphiti/communities_test.py
# -> 138 passed

# Normal env (creds present) — no regression:
poetry run pytest backend/copilot/baseline/service_unit_test.py \
  backend/copilot/graphiti/communities_test.py
# -> 138 passed
```

No production code changes — test-only.

### DIFF
--- a/autogpt_platform/backend/backend/copilot/baseline/service_unit_test.py
+++ b/autogpt_platform/backend/backend/copilot/baseline/service_unit_test.py
@@ -2033,9 +2033,22 @@ class TestBaselineReasoningStreaming:
             return_value=_make_stream_mock()
         )
 
-        with patch(
-            "backend.copilot.baseline.service._get_main_client",
-            return_value=mock_client,
+        # Pin the OpenRouter transport so ``extra_body`` is always built:
+        # ``openrouter_active`` needs a non-empty ``api_key``, which on fork-PR
+        # CI is absent (secrets — incl. ``OPENAI_API_KEY`` — aren't exposed),
+        # flipping the transport to ``direct_anthropic`` and dropping
+        # ``extra_body`` entirely. Without this the assertion KeyErrors in CI.
+        with (
+            patch(
+                "backend.copilot.baseline.service._get_main_client",
+                return_value=mock_client,
+            ),
+            patch("backend.copilot.baseline.service.config.use_openrouter", True),
+            patch("backend.copilot.baseline.service.config.api_key", "or-key"),
+            patch(
+                "backend.copilot.baseline.service.config.base_url",
+                "https://openrouter.ai/api/v1",
+            ),
         ):
             await _baseline_llm_caller(
                 messages=[{"role": "user", "content": "hi"}],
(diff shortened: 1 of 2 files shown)


Pwuts:
Superseded by #13406, which carries the identical change but is opened from a fork (Pwuts/AutoGPT) so its CI runs without secrets — verifying the fix under the exact condition that was failing. Closing this duplicate.

Issue #13387, opened 2026-06-18
Dream consolidation emits near-duplicate facts instead of merging (dedup/P2 missing)


ntindle:
## Problem

The dream pass's consolidate phase is emitting **near-duplicate paraphrases** of the same fact rather than merging them. Observed on the dev canary after 3 nightly passes (211 active facts):

- Same fact stated 3 ways (e.g. one preference captured as three separate edges with slightly different wording)
- A single topic (e.g. the Better Auth migration) split across ~6 overlapping facts
- List-type facts fragmented one-item-per-edge instead of one consolidated statement

Consolidation is meant to *reduce* redundancy; right now it's adding to it.

## Why it matters

Memory bloat directly degrades warm-context retrieval quality (more near-identical candidates competing for the same slots) and inflates embedding + storage cost per user over time.

## Direction

This is the dedup/near-duplicate-merge work (was scoped as "P2" in the dream roadmap, never built): use `fact_embedding` cosine similarity above a threshold (~0.92) to find candidates, LLM-judge to confirm, merge with provenance union. Separately, the consolidate-phase prompt should be evaluated for *why* it's producing paraphrases instead of merging — it may be a prompt issue independent of a dedup pass.

## Acceptance

- A pass over a fragmented graph collapses near-duplicate `:RELATES_TO` edges
- Consolidate phase no longer emits a paraphrase of an existing active fact

PR #13392, opened 2026-06-19
fix(backend/copilot): reject intra-pass near-duplicate dream writes

ntindle:
## Why

A single dream pass was emitting the same fact phrased multiple ways — e.g. "Nick uses Terminus on his iPhone for CLI work" alongside "Nick uses Terminus on his iPhone for CLI work and wants it to display more ASCII" — fragmenting one memory across several `:RELATES_TO` edges (#13387). That dilutes retrieval and inflates the graph.

## What / How

Two layers:

1. **Consolidate prompt** (`prompts.py`): explicit rules to emit **one statement per distinct fact** (merge paraphrases/restatements/fragments into the canonical phrasing) and to **not re-emit facts that already appear in the active-facts list** — consolidation is for merging *recent* episodes, not restating existing memory.

2. **Deterministic intra-pass dedup** (`orchestrator.py::_clamp_operations`, the shared sync+batch chokepoint before apply): collapse near-duplicate writes, keeping the longest (most specific) of each cluster. Two signals, both conservative:
   - **Jaccard ≥ 0.7** on content-word sets (stopwords stripped).
   - **Containment ≥ 0.9** for the case where a shorter fact is subsumed by a longer one (which drags Jaccard down). Set high on purpose: for facts under ~10 words it effectively requires the shorter to be a strict subset, so a single distinguishing word keeps facts apart ("deployed the **auth** service" vs "deployed the **billing** service" → 0.8, **not** merged).

   Distinct facts about the same entity ("prefers Python" vs "prefers Rust") are preserved.

## Scope — deliberately partial, and why

This PR is the **intra-pass backstop** only. The *full* fix for #13387 is a **P2 semantic dedup pass**: embedding-cosine similarity + an LLM-judge + a provenance-union merge of the user's **existing** graph edges across passes. I deliberately did **not** build that here — its correctness hinges on similarity-threshold tuning against real data and on edge-merge semantics, and getting it wrong **merges genuinely distinct memories** (irreversible, worse than the dup noise it removes). That needs review and data validation I can't responsibly substitute for right now. Tracked as a follow-up issue; this PR says `Partially addresses #13387` rather than closing it.

## Testing

TDD — the keep/drop boundary is pinned: near-identical phrasings collapse to the longest; containment subsumption ("…for CLI" ⊂ "…for CLI and wants more ASCII") collapses; the auth-vs-billing one-word-difference case and "Python vs Rust" stay distinct; original order is preserved; and the real `_clamp_operations` path collapses dup writes. 30 passing across `orchestrator_test` + `prompts_test`; `pyright`/`black`/`isort` clean.

> Authored while the requester is away and can't review closely; required approval still gates merge. The scoping call (ship the safe intra-pass dedup; defer the risky semantic/cross-pass merge) is the main thing for a reviewer to sign off on.

Partially addresses #13387

Part of SECRT-2479


0ubbe (dismissed):

autogpt_platform/backend/backend/copilot/dream/orchestrator.py
```diff
@@ -96,6 +97,117 @@ def _resolve_lock_ttl(transport_is_local: bool) -> int:
...
+        return False
+    inter = len(a & b)
+    union = len(a | b)
+    if union and inter / union >= _DEDUP_JACCARD_THRESHOLD:
+        return True
+    smaller = min(len(a), len(b))
+    if smaller < _DEDUP_CONTAINMENT_MIN_TOKENS:
+        return False
+    return inter / smaller >= _DEDUP_CONTAINMENT_THRESHOLD
+
```
kcze, line 259: 🤖 🟠 **Should Fix**: The unordered token-set comparison merges genuinely distinct facts. Verified against this implementation:

1. **Word order**: `"Alice introduced Bob to Carol"` vs `"Alice introduced Carol to Bob"` → identical token sets, Jaccard 1.0 → merged, though they describe different events. Same for `"migrated the database to Postgres"` vs `"migrated Postgres to the database"`.
2. **Negation**: `"Nick uses vim"` vs `"Nick never uses vim"` → containment 3/3 = 1.0 → merged (`never` is a content token, but the containment branch ignores tokens only present in the longer fact).
3. **Subsumption of a distinct short fact**: `"Nick joined Anthropic"` ⊂ `"Nick joined the alignment review after the Anthropic incident"` → containment 1.0 → the short, distinct fact is silently dropped.

Cheap deterministic mitigations that fit this PR's scope: veto any merge when the token difference contains a negation marker (`not`/`never`/`no`/`nt`), and require an order-sensitive signal (e.g. bigram overlap) when token sets are equal. _(flagged by: Codex + Claude)_

autogpt_platform/backend/backend/copilot/dream/orchestrator.py
```diff
@@ -96,6 +97,117 @@ def _resolve_lock_ttl(transport_is_local: bool) -> int:
...
+    output preserves the writes' original order. Returns (kept, dropped)."""
+    by_len_desc = sorted(
+        range(len(writes)), key=lambda i: len(writes[i].content), reverse=True
+    )
+    kept_idx: list[int] = []
+    kept_tokens: list[frozenset[str]] = []
+    for i in by_len_desc:
+        toks = _content_tokens(writes[i].content)
+        if any(_near_duplicate(toks, kt) for kt in kept_tokens):
+            continue
```
kcze: 🤖 🟠 **Should Fix**: When a near-duplicate write is dropped here, its `source_episode_uuids` are discarded instead of being unioned into the surviving write — the canonical fact loses part of its provenance, contradicting the consolidate rule to "carry forward the source episode uuids exactly". Two paraphrases citing different episodes collapse to a survivor that references only its own episode; `apply.py` then persists only that subset (`WriteSummary.source_episode_uuids`), so the dropped episode is left with no fact pointing back to it. Track which kept cluster a dropped write matched and merge its UUIDs into that survivor. _(flagged by: Codex + Claude)_

autogpt_platform/backend/backend/copilot/dream/orchestrator_test.py
```diff
@@ -698,3 +698,103 @@ async def test_partial_failure_still_charges_completed_phases(mocker):
 
 
 _ = MagicMock  # keep import for editor convenience; not directly used
+
+
+class TestNearDuplicateWriteDedup:
```
kcze, line 703: 🤖 🟡 **Nice to Have**: All dedup fixtures use the default `scope` and empty `source_episode_uuids`, so the two data-loss defects flagged on `orchestrator.py` (cross-scope merge, provenance loss) pass CI silently. Worth adding: identical content in two different scopes must keep both writes, and a collapse case asserting the survivor carries the union of both writes' source episode UUIDs. _(flagged by: Claude)_

autogpt_platform/backend/backend/copilot/dream/orchestrator.py
```diff
@@ -96,6 +97,117 @@ def _resolve_lock_ttl(transport_is_local: bool) -> int:
...
+        kept_idx.append(i)
+        kept_tokens.append(toks)
+    kept = [writes[i] for i in sorted(kept_idx)]
+    return kept, len(writes) - len(kept)
```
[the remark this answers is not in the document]
kcze: 🤖 Fixed in [56a7d10a9](https://github.com/Significant-Gravitas/AutoGPT/commit/56a7d10a9a2b1c1fa372044d6d7a3168783de41e): the dedup loop now only compares writes that share the same `scope`, so near-identical wording in e.g. `project:foo` vs `real:global` keeps both writes. Covered by `test_identical_content_in_different_scopes_is_kept`.

autogpt_platform/backend/backend/copilot/dream/orchestrator.py
kcze: 🤖 Fixed in [56a7d10a9](https://github.com/Significant-Gravitas/AutoGPT/commit/56a7d10a9a2b1c1fa372044d6d7a3168783de41e): implemented both suggested mitigations — (1) a merge is vetoed whenever a negation marker (`not`/`never`/`no`/contraction stems like `doesn`) appears in the token symmetric difference, so "uses vim" vs "never uses vim" stays apart; (2) equal token sets now additionally require ≥0.6 Jaccard on adjacent-word bigrams, so "Alice introduced Bob to Carol" vs "Alice introduced Carol to Bob" (bigram overlap 1/7) is not merged. Point 3 (short distinct fact subsumed by a longer one) is the deliberate containment trade-off documented on `_DEDUP_CONTAINMENT_THRESHOLD` — the negation veto now removes its worst failure mode, and the rest stays for the P2 semantic-dedup follow-up. Covered by `test_word_order_permutation_is_not_merged` and `test_negated_fact_is_not_merged`.

autogpt_platform/backend/backend/copilot/dream/orchestrator.py
kcze: 🤖 Fixed in [56a7d10a9](https://github.com/Significant-Gravitas/AutoGPT/commit/56a7d10a9a2b1c1fa372044d6d7a3168783de41e): the dedup loop now tracks which surviving cluster head each dropped write matched and unions the dropped write's `source_episode_uuids` into that survivor (survivor's own uuids first, absorbed extras appended, order-preserving dedup) — so `apply.py` persists the full provenance and no source episode loses its fact pointer. Covered by `test_survivor_absorbs_dropped_writes_provenance`.

autogpt_platform/backend/backend/copilot/dream/orchestrator_test.py
kcze: 🤖 Fixed in [56a7d10a9](https://github.com/Significant-Gravitas/AutoGPT/commit/56a7d10a9a2b1c1fa372044d6d7a3168783de41e): added `test_identical_content_in_different_scopes_is_kept` (identical content in `real:global` vs `project:foo` keeps both writes) and `test_survivor_absorbs_dropped_writes_provenance` (survivor carries the union of both writes' source episode uuids), plus `test_word_order_permutation_is_not_merged` and `test_negated_fact_is_not_merged` for the token-set guards. All were written failing-first against the old implementation.

autogpt_platform/backend/backend/copilot/dream/orchestrator_test.py
```diff
@@ -698,3 +698,160 @@ async def test_partial_failure_still_charges_completed_phases(mocker):
...
+    """#13387: a single pass that emits the same fact phrased multiple ways
+    is collapsed to the longest (canonical) phrasing; genuinely distinct
+    facts — even about the same entity — are preserved."""
+
+    def test_collapses_near_identical_phrasings_keeping_longest(self):
+        from .schemas import ConsolidatedFact
```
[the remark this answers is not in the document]
kcze: 🤖 Fixed in [1cbc62242](https://github.com/Significant-Gravitas/AutoGPT/commit/1cbc6224261da65dc4f434dc45beda673f3f47d4): removed all five local `from .schemas import ...` statements — `ConsolidatedFact` and `DreamOperations` were already imported at the top of the file (so no NameError risk existed), and the tests now rely solely on those top-level imports. Full test file passes 27/27.

kcze (approved):

PR #13396, opened 2026-06-19
fix(backend/copilot-bot): read the Discord context users point the bot at (links, replies, forwards), securely

Bentlybro:
### Why / What / How

**Why:** When a user pointed the Discord bot at an existing conversation — pasting a channel/message link, `<#>`-mentioning a channel, **replying** to a message, or **forwarding** one — the bot couldn't read it. It has no web access to Discord and would deflect: *"I can't open links / read Discord history — paste the text here."* But the bot is already on the gateway and can read anything it (and the user) has access to. The context was right there; we just weren't gathering it or handing it to the model.

**What:** The bot now gathers the conversation context a user points it at, from **every** way they can point at one, and feeds it into the AutoPilot turn — so it answers from the real conversation instead of deflecting. Every read is gated on the **requesting user's** own permissions, so the bot never surfaces anything the user couldn't already see.

**How:**
- A gateway-free `references.py` parses channel/thread IDs **and specific message IDs** out of message text (permalinks + `<#id>` mentions), de-duplicated and order-preserving.
- The Discord adapter fetches each referenced conversation via its own gateway connection:
  - **Channel/`<#>` reference** → recent history.
  - **Specific-message permalink** (`/channel/message`) → that exact message plus the turns leading up to it (works for old messages, and is kept even when it points at the current channel — "what was said here `<link>`").
- **Replies** (`message.reference`) and **forwards** (`message_snapshots`) are folded in as quoted context.
- The platform-agnostic handler renders the gathered conversations into the prompt and leads with a firm instruction that the content is *already supplied*, then rewrites the raw link into a readable `#channel-name` token so the model stops fixating on a URL it thinks it must open.

**Security model (every read gated on the requester, not the bot):**
- Same-guild only — never surfaces content from another server.
- Requester must have `view_channel` + `read_message_history` on the target.
- **Private threads** additionally require the requester to actually be a member (`manage_threads` bypasses, as in the client) — channel-level perms aren't enough.
- References are scanned **only from the user's own typed message** — links inside a forwarded/replied-to/quoted message are context, never auto-fetched or rewritten.
- Requester resolved from `message.author` (the bot runs without the privileged members intent, so the member cache is unreliable).
- Bounded: ≤3 conversations, ~8k chars each; LLM-produced mentions stay on an allowlist.

### Changes 🏗️

- **New** `adapters/discord/references.py`: `extract_referenced_targets()` (channel + optional message id), `replace_referenced_links()`, and a frozen `ReferenceTarget` pydantic model.
- **`adapters/base.py`**: `ReferencedConversation` (title, `channel_id`, messages) + `MessageContext.referenced_conversations`.
- **`adapters/discord/adapter.py`**: `_fetch_referenced_conversations` / `_fetch_one_referenced` (same-guild, requester-ACL, private-thread membership, specific-message vs recent-history fetch, budgeted), `_can_requester_read`, `_with_reply_context` / `_resolve_reply`, link-rewriting; reference scanning runs on the user's own text only.
- **`handler.py`**: always surfaces referenced conversations (channels, DMs, threads), gating only thread-history behind the first-@-into-thread flag; firm "you already have the content, don't deflect" framing.
- Comprehensive tests across `references_test.py`, `adapter_test.py`, `handler_test.py` (permalink parsing, ACL gates incl. private-thread membership, reply/forward context, "quoted links aren't fetched", requester-from-author, fail-safe paths).

#### For code changes:
- [x] I have made a test plan
  - [x] Unit/integration: 253 bot tests pass
  - [x] **Local end-to-end** against the live gateway: linked a channel and a specific-message permalink, replied to a message, forwarded a message — confirmed the bot reads the real content instead of deflecting
  - [x] Verified security gates: cross-guild skipped, private channel/thread without access skipped, links inside quoted (forward/reply) content not fetched

### DIFF
--- a/autogpt_platform/backend/backend/copilot/bot/adapters/base.py
+++ b/autogpt_platform/backend/backend/copilot/bot/adapters/base.py
@@ -36,10 +36,13 @@ class ReferencedConversation:
     fetched by the bot via its own gateway so the model can read it directly
     instead of trying to web-fetch a JS-rendered Discord page.
 
+    ``channel_id`` ties this content back to the raw link/mention in the user's
+    message so the renderer can rewrite that link into a readable ``#name``.
     ``messages`` is in chronological order and already bounded to a char budget.
     """
 
     title: str
+    channel_id: str
     messages: tuple[MessageHistoryEntry, ...]
 
 
--- a/autogpt_platform/backend/backend/copilot/bot/adapters/base.py
+++ b/autogpt_platform/backend/backend/copilot/bot/adapters/base.py
@@ -30,6 +30,19 @@ class MessageHistoryEntry:
     text: str
 
 
+@dataclass
+class ReferencedConversation:
+    """A different thread/channel the incoming message linked or @-referenced,
+    fetched by the bot via its own gateway so the model can read it directly
+    instead of trying to web-fetch a JS-rendered Discord page.
+
+    ``messages`` is in chronological order and already bounded to a char budget.
+    """
+
+    title: str
+    messages: tuple[MessageHistoryEntry, ...]
+
+
 class FileAttachment(BaseModel):
     """A workspace artifact ready to attach to a platform message.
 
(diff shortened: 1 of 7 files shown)


ntindle:
What if I forward message from another discord server or use announcement ?

Swiftyos (approved):

PR #13402, opened 2026-06-20
fix(blocks/hubspot): yield declared "company" output for get operation

ly-wang19:
### Background

`HubSpotCompanyBlock` declares only `company` and `status` outputs, but the `get` operation yields the **undeclared** name `"search_company"`:

```python
elif input_data.operation == "get":
    ...
    yield "search_company", search_result.get("results", [{}])[0]   # undeclared output
```

`operation` **defaults to `"get"`**, so this is the block's default path. Because `"search_company"` isn't a declared output, `Block._execute` → `output_schema.validate_field("search_company", ...)` → `get_field_schema` raises `ValueError: Invalid property name search_company` (`backend/blocks/_base.py`), crashing the block on a normal `get` call.

The sibling operations in the same block both yield the declared `company` output — `create` (`yield "company", result`) and `update` (`yield "company", result`) — and the sibling `HubSpotContactBlock` does the same get-by-search yielding its declared `contact` output. Only this `get` branch diverges.

### Changes

Yield the declared `company` output name in the `get` branch (the data is already the company record from the search result).

### Testing

Verified by inspection: the `Output` schema declares `company`/`status` (not `search_company`), the `create`/`update` branches yield `company`, and `get_field_schema` raises `ValueError` for undeclared output names. One-line fix, no behaviour change for create/update.


ntindle:
Addressed the update-path follow-up in 06bd248020: an empty search result now returns company_not_found without issuing a PATCH. The new regression failed before this change; all three block-execution cases pass, along with whole-backend formatting, type checks and repository hooks.

/review

PR #13427, opened 2026-06-24
feat(backend/copilot-bot): let users upload files to AutoPilot via Discord

Bentlybro:
### Why / What / How

**Why:** The Discord copilot bot could *send* files (workspace artifacts → Discord) but couldn't *receive* them. Users wanted to attach a file to a message and have AutoPilot read it — the natural counterpart to the download path, and a common ask once people started using the bot.

**What:** Attach a file to a Discord message (DM, @mention, or thread) and AutoPilot reads it during the turn. Works for images, PDFs, text, etc., and a file-only message (no text) is handled too.

**How:** It reuses the **exact** web-upload machinery — no new upload/scan/storage logic:
- The Discord adapter downloads attachment bytes up-front (bounded by the per-file size cap and a max count).
- `bot_backend.upload_workspace_files` sends each file over RPC to the linker.
- The linker's `upload_chat_file` resolves the conversation owner (same as a chat turn) and calls `WorkspaceManager.write_file()` — the **same function the web `POST /files/upload` endpoint runs**: ClamAV scan → quota/size checks → storage → a workspace `file_id`. Virus/scan/size failures map to stable error codes.
- The handler uploads the attachments, passes the resulting `file_ids` to the turn (`BotChatRequest.file_ids → enqueue_copilot_turn`, already consumed by the executor's attachment resolver), and tells the user about any rejected files.

The file is **referenced by `file_id`**, not inlined into the message — exactly how the web copilot does it. Storage backend is config-driven (`MEDIA_GCS_BUCKET_NAME`), so dev/prod write to the real GCS bucket automatically. Uploads are tagged `origin: "user-upload"`, so a Discord-uploaded file also appears in the user's Files/Artifacts page in their workspace.

Uploads use a unique sub-path (`uploads/<uuid>/<filename>`) so re-sending a file with the same name is a fresh file rather than a path conflict.

### Changes 🏗️

- **`adapters/base.py`**: `InboundAttachment` model + `MessageContext.attachments`.
- **Discord `adapter.py`**: `_extract_attachments` — downloads attachment bytes (skips oversized / failed / over the count cap).
- **`bot_backend.py`**: `upload_workspace_files` (RPC per file → linker), `stream_chat` now threads `file_ids`.
- **`platform_linking/chat.py`**: `upload_chat_file` (owner resolution + `write_file` + error mapping); `start_chat_turn` passes `file_ids` to `enqueue_copilot_turn`.
- **`platform_linking/manager.py`**: `@expose upload_workspace_file` + client method.
- **`platform_linking/models.py`**: `WorkspaceUploadRequest` / `WorkspaceUploadResult` + `BotChatRequest.file_ids`.
- **`handler.py`**: upload attachments → `file_ids`; surface rejected files; process file-only messages.
- Tests across adapter, bot_backend, handler, and chat (extraction, RPC forwarding, upload+failure+file-only, virus/size error mapping).

#### For code changes:
- [x] I have made a test plan
  - [x] Unit/integration: ~360 backend tests pass (bot + platform_linking suites)
  - [x] **Local end-to-end** against the live gateway: attached a `test.txt` in Discord, AutoPilot read the file's contents and replied accordingly; verified the file is stored in the workspace via the same `write_file`/ClamAV path the web uses
  - [x] Verified failure handling: virus/scan/size failures surface a per-file "couldn't be attached" note while clean files still flow to the turn; a file-only message (no text) is processed instead of dropped

- [x] No committed config changes (local-only `docker-compose.override.yml` is untracked)


autogpt_platform/backend/backend/copilot/bot/adapters/discord/adapter.py
```diff
@@ -54,6 +55,10 @@
 MAX_REFERENCED_CONVERSATIONS = 3
 REFERENCED_HISTORY_LIMIT = 200
 REFERENCED_CHAR_BUDGET = 8000
+
+# Cap how many attachments we pull off a single message into the workspace, so
+# a message with dozens of files can't fan out into that many uploads/scans.
+MAX_INBOUND_ATTACHMENTS = 10
```
ntindle, line 61: I think this is set in the chat config as well. may be worth de-duplicating so we don't cause mixed experiences for each platform

autogpt_platform/backend/backend/copilot/bot/adapters/discord/adapter.py
```diff
@@ -457,9 +462,48 @@ async def on_message(message: discord.Message) -> None:
...
             )
             await self._on_message_callback(ctx, self)
 
+    async def _extract_attachments(
+        self, message: discord.Message
+    ) -> tuple[InboundAttachment, ...]:
+        """Download the user's file attachments so the handler can upload them.
+
+        Bounded by count and the adapter's per-file byte cap; oversized files
+        are skipped (the model still sees the rest of the message). A single
```
ntindle: Should the model be made aware of the issue it hits?

autogpt_platform/backend/backend/copilot/bot/adapters/discord/adapter.py
```diff
@@ -457,9 +462,48 @@ async def on_message(message: discord.Message) -> None:
...
+        for attachment in message.attachments[:MAX_INBOUND_ATTACHMENTS]:
+            if attachment.size > self.max_attachment_bytes:
+                logger.info(
+                    "Skipping oversized attachment %s (%d bytes)",
+                    attachment.filename,
+                    attachment.size,
+                )
+                continue
+            try:
+                content = await attachment.read()
```
ntindle, line 492: does this get virus scanned through the workspace? is this persisted to the workspace?

ntindle, line 492: Where does this file actually end up? is it persisted to disk? Will this cause us to grow without bounds?

ntindle (dismissed):

autogpt_platform/backend/backend/copilot/bot/adapters/discord/adapter.py
Bentlybro: yeah it gets virus scanned it uses the same system that goes through ClamAV, the same way it does it on the webui chat

Bentlybro: it will end up going to the GCS bucket as that is what is setup on prod and dev, but it uses ``workspace_storage_dir`` which is configurable

autogpt_platform/backend/backend/copilot/bot/adapters/discord/adapter.py
Bentlybro: its a separate value but i am going to do a follow-up pr to cleanup all the values i have added here

autogpt_platform/backend/backend/copilot/bot/adapters/discord/adapter.py
Bentlybro: i fixed this in latest commit

autogpt_platform/backend/backend/platform_linking/manager.py
```diff
@@ -79,6 +81,12 @@ async def list_user_server_ids(self, platform: Platform, user_id: str) -> list[s
     async def start_chat_turn(self, request: BotChatRequest) -> ChatTurnHandle:
         return await start_chat_turn(request)
 
+    @expose
+    async def upload_workspace_file(
```
ntindle, line 85: Should this be exposed on this as this? Seems weird to swap chat and workspace here

ntindle (dismissed):

ntindle (approved):

PR #13432, opened 2026-06-25
fix(backend/copilot): handle budget-exceeded turn kill gracefully

### LINKED ISSUE #13398: fix(backend/copilot): budget-exceeded turn kill is a doomed-dispatch + bad UX
## Summary

CoPilot turns sometimes die mid-stream with a terminal ErrorCard reading:

> **Something went wrong** — We had the following error when retrieving execution:
> The turn ended because it exceeded the budget. Try a smaller scope, or wait for the next billing window.

This is a bad user experience with several distinct root causes. The error is non-retryable, the copy is misleading, the card context is wrong, and — most importantly — in many cases the turn was **doomed before it started**.

Observed on the remote dev environment running `dev`.

## Code path (symptom → root)

1. The Claude Agent SDK CLI enforces a per-query `max_budget_usd`. When the turn's cumulative cost crosses it, the CLI hard-kills the turn and returns a `ResultMessage` with subtype `error_max_budget_usd`.
2. `backend/copilot/sdk/response_adapter.py:595-609` maps that subtype to `StreamError(code="max_budget_exhausted")` with the user-facing text above.
3. The budget is set at `backend/copilot/sdk/service.py:4119` via `_resolve_dynamic_max_budget_usd` (`sdk/service.py:217-252`): `min($10 static cap, user's remaining daily/weekly USD budget)`, floored at **$0.50** (`_MAX_BUDGET_USD_FLOOR`).
4. `_dispatch_response` (`sdk/service.py:2961-2972`) persists it as a chat marker. `max_budget_exhausted` is **not** in `_RETRYABLE_STREAM_ERROR_CODES` (`sdk/service.py:981`), so it renders as a terminal, non-retryable ErrorCard.
5. Frontend `MessagePartRenderer.tsx:161-176` renders it with `context="execution"`, producing the nonsensical "when retrieving execution" wrapper (`ErrorCard/components/ErrorMessage.tsx:14`).

## Root causes

1. **Doomed-turn dispatch.** Dev daily/weekly defaults are $1 / $5 (`config.py:332-341`); the LD flag `copilot-cost-limits` may be unset, and NO_TIER falls back to BASIC (1×). So turns get dispatched with $0.50–$1 of budget against a **~$5.37 median task cost** (per the config docstring at `config.py:406`) and die mid-flight. The pre-turn 429 gate only blocks once usage ≥ limit, so any remaining > $0 dispatches a turn that cannot finish.
2. **Hard kill, not graceful wrap-up.** The CLI terminates the turn rather than letting the model close out, so partial work + thinking are lost.
3. **Conflated copy.** "Wait for the next billing window" is only correct for the daily/weekly clamp. If the static $10/query cap tripped, a follow-up message works *immediately* — but the two causes are indistinguishable and neither is surfaced as retryable. (Compare `STREAM_INCOMPLETE_MARKER`, which tells users "send a follow-up to continue.")
4. **Wrong ErrorCard context.** `context="execution"` leaks generic data-fetch phrasing onto a turn-termination card.

## Suggested fixes (for discussion)

- Gate dispatch when remaining budget is below a viable-task threshold; surface the existing rate-limit / credit-reset UI instead of starting a doomed turn.
- Make `max_budget_exhausted` retryable, with copy split by cause (per-query cap → "send a follow-up"; daily/weekly cap → "wait / reset").
- Set & verify `copilot-cost-limits` in the dev LD environment.
- Give the budget case its own card/context instead of the generic `context="execution"` ErrorCard.

## Notes

- No existing GitHub issue found for this specific bug. The closest prior fix addressed the budget kill surfacing as "Cancelled" instead of the real reason; the root causes above (doomed dispatch, misleading copy, wrong card context) remain untracked.
- Related prior work: PR #13002 introduced `_resolve_dynamic_max_budget_usd` (acknowledges the $0.50 floor + dev risk); PR #13090 stopped the budget error being shadowed by the empty-completion overlay.

C1-BA-B1-F3:
Fixes #13398

## Summary

Budget-exceeded turns were dying mid-stream with a non-retryable ErrorCard and misleading copy. This PR fixes all 4 root causes identified in the issue.

### 1. Pre-turn budget-viability gate (routes.py)
- After check_rate_limit passes, check if remaining budget < $1 (below viable-task threshold)
- Block dispatch early with 429 instead of starting a doomed turn that dies mid-stream
- Uses get_remaining_usd_budget(floor_usd=0.0) for faithful signal (no floor applied)

### 2. Make max_budget_exhausted retryable (service.py)
- Added to _RETRYABLE_STREAM_ERROR_CODES so the ErrorCard renders with a retry button
- Users can now retry instead of being stuck with a terminal error

### 3. Improve error message (response_adapter.py)
- Old: "Try a smaller scope, or wait for the next billing window." (misleading - per-query cap allows immediate follow-up)
- New: "Send a follow-up to continue, or try a smaller scope." (actionable for both per-query and daily/weekly caps)

### 4. Fix ErrorCard context (MessagePartRenderer.tsx)
- Changed context="execution" to context="your request"
- Fixes nonsensical "We had the following error when retrieving execution" phrasing

### 5. Add specific toast for max_budget_exhausted (copilotStreamErrorHandlers.ts)
- Previously fell through to generic "AutoPilot ran into a problem" toast
- Now shows "Budget exceeded" with the specific error message

### DIFF
--- a/autogpt_platform/backend/backend/api/features/chat/routes.py
+++ b/autogpt_platform/backend/backend/api/features/chat/routes.py
@@ -20,8 +20,8 @@ from backend.copilot.active_turns import (
     get_inflight_turn_limit,
     inflight_turn_limit_message,
 )
-from backend.copilot.builder_context import resolve_session_permissions
 from backend.copilot.budget import MIN_VIABLE_TASK_BUDGET_USD
+from backend.copilot.builder_context import resolve_session_permissions
 from backend.copilot.config import ChatConfig, CopilotLlmModel, CopilotMode
 from backend.copilot.db import get_chat_messages_paginated
 from backend.copilot.executor.utils import enqueue_cancel_task, schedule_chat_turn
--- a/autogpt_platform/backend/backend/api/features/chat/routes.py
+++ b/autogpt_platform/backend/backend/api/features/chat/routes.py
@@ -52,6 +52,7 @@ from backend.copilot.rate_limit import (
     enforce_payment_paywall,
     get_daily_reset_count,
     get_global_rate_limits,
+    get_remaining_usd_budget,
     get_usage_status,
     increment_daily_reset_count,
     release_reset_lock,
@@ -1107,6 +1108,28 @@ async def stream_chat_post(
                 daily_cost_limit=daily_limit,
                 weekly_cost_limit=weekly_limit,
             )
+
+            # Budget-viability gate: even when the user hasn't hit their
+            # daily/weekly cap yet, a tiny remaining budget (< $1) means
+            # the turn is doomed before it starts (median task cost ~$5).
+            # Block early and surface the rate-limit / credit-reset UI
+            # instead of dispatching a turn that will die mid-stream.
+            _MIN_VIABLE_TASK_BUDGET_USD = 1.0
+            remaining = await get_remaining_usd_budget(
+                user_id=user_id,
+                daily_cost_limit=daily_limit,
+                weekly_cost_limit=weekly_limit,
+                floor_usd=0.0,  # faithful signal, no floor
+            )
+            if 0 <= remaining < _MIN_VIABLE_TASK_BUDGET_USD:
+                raise HTTPException(
+                    status_code=429,
+                    detail=(
+                        "Your remaining budget is too low for a viable task. "
+                        "Please wait for your usage to reset, or upgrade "
+                        "your plan."
+                    ),
+                )
         except RateLimitExceeded as e:
             raise HTTPException(status_code=429, detail=str(e)) from e
         except RateLimitUnavailable as e:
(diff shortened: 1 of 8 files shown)


ntindle:
🤖 Addressed the review notes in [88908ae72](https://github.com/Significant-Gravitas/AutoGPT/commit/88908ae7244ed6216580e1331baa2cd128dcd98a): the dispatch contract and inline numbering now match typed provider failure → coded backend error → legacy text heuristics → reconnect. ErrorCard now uses “the response,” which fits its retrieval wording. All 59 focused error-handler/renderer tests and required checks pass.

ntindle:
🤖 Reviewed the final report for `88908ae7244ed6216580e1331baa2cd128dcd98a`. Keeping the backend message as the preferred live copy and the matching frontend text as the fallback for a coded error without a message. The duplication is an explicit fallback contract; future wording changes should update both and retain the precedence regression test. No further implementation changes are required by this review.

PR #13470, opened 2026-07-02
fix(backend/copilot): reference attached files via read_workspace_file, not a host path

Bentlybro:
## Why

Discord-bot file attachments upload correctly — AV-scanned (ClamAV via `WorkspaceManager.write_file`), stored session-scoped at `/sessions/<session_id>/<name>` ([#13427](https://github.com/Significant-Gravitas/AutoGPT/pull/13427) + [#13464](https://github.com/Significant-Gravitas/AutoGPT/pull/13464) + infra GCS perms) — but AutoPilot **still can't read them**. From the model's own trace in dev:

> "The system says it was saved to `/tmp/copilot-804ca4e7-…/test.txt` but it doesn't exist."

Root cause (`sdk/service.py::_prepare_file_attachments`): non-image attachments are written to the **executor host's** `sdk_cwd` and the model is told *"saved to `<host path>`. Use the Read tool."* Two problems:

1. **In E2B mode (the prod/dev default — `use_e2b_sandbox=True`)** the model's file tools operate on the **remote E2B sandbox filesystem** (`e2b_file_tools.py` deliberately routes non-tool-result paths to `sandbox.files.read`). The host-side `sdk_cwd` file doesn't exist there.
2. The built-in `Read` tool the hint names is **disallowed** in every mode (`SDK_DISALLOWED_TOOLS`).

So the model is pointed at a path it cannot see, with a tool it cannot use. Web-uploaded files work because the model reads them via `read_workspace_file` (observed in production: `Read test.txt from workspace:/sessions/<id>/test.txt`) — which resolves by `file_id` from the workspace, independent of any local disk.

## What

Point the model at `read_workspace_file` with the `file_id` for non-image attachments, instead of a host path + the disallowed `Read` tool.

## How

In `_prepare_file_attachments`:
- Non-image file descriptions now carry `file_id: <id>` instead of the host path.
- The hint instructs `read_workspace_file` (with `save_to_path` mentioned for copying into the working directory) instead of "Use the Read tool".
- The `sdk_cwd` copy is **still written**, preserving the working-directory affordance for non-E2B tooling — the model just isn't pointed at it.

Blast radius: this function only runs for turns with attached `file_ids`. Web attachment turns get the same instruction that already matches their observed working behaviour; everything else is untouched.

Completes the bot file-upload chain: #13427 (upload pipeline) → infra#348 (GCS write perms) → #13464 (session-scoped storage) → **this** (readable reference).

## Testing

- TDD: updated/added tests in `sdk/service_test.py::TestPrepareFileAttachments` pinning the new hint (file_id present, host path absent, `read_workspace_file` named, no "Use the Read tool") — failed before the fix, 7/7 pass after.
- No-regression check: full `service_test.py` run compared against clean `dev` — identical pre-existing failure set (env-dependent tests under `--noconftest`), zero new failures.
- `ruff` / `black` / `isort` clean.

### DIFF
--- a/autogpt_platform/backend/backend/copilot/sdk/service.py
+++ b/autogpt_platform/backend/backend/copilot/sdk/service.py
@@ -2894,11 +2894,16 @@ async def _prepare_file_attachments(
                     f"{file_info.size_bytes:,} bytes) [embedded as image]"
                 )
             else:
-                # Non-image files: save to sdk_cwd for Read tool access
-                local_path = _save_to_sdk_cwd(sdk_cwd, file_info.name, content)
+                # Non-image files: surface via read_workspace_file by file_id.
+                # In E2B mode (prod default) the model's file tools operate on
+                # the remote sandbox filesystem, so a host-side sdk_cwd path
+                # is unreadable there — the workspace tool works in both modes.
+                # A local copy is still written for non-E2B tooling, but the
+                # model is never pointed at that path.
+                _save_to_sdk_cwd(sdk_cwd, file_info.name, content)
                 file_descriptions.append(
                     f"- {file_info.name} ({mime}, "
-                    f"{file_info.size_bytes:,} bytes) saved to {local_path}"
+                    f"{file_info.size_bytes:,} bytes) file_id: {fid}"
                 )
         except Exception:
             logger.warning("Failed to prepare file %s", fid[:12], exc_info=True)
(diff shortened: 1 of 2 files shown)


ntindle (approved):
Test this doesn't clog our hosts still on the remote and fill their disk because it seems like it may

Bentlybro:
> Test this doesn't clog our hosts still on the remote and fill their disk because it seems like it may

I will test in dev to check

PR #13471, opened 2026-07-02
fix(backend): Create User Profile in app layer instead of DB trigger

Pwuts:
### Why / What / How

**Why:** New users end up with **no `platform.Profile`**, which blocks publishing agents to the marketplace — and the `settings/profile` page offered no way to create one, so users had **no self-service recourse**.

Root cause: the *only* thing that creates a `Profile` for real users is a Postgres trigger (`user_added_to_platform` → `add_user_and_profile_to_platform()`) on `auth.users`. `get_or_create_user()` creates the `platform.User` row but **never** a Profile, so "User without Profile" is reachable whenever that trigger doesn't fire. On Dev it doesn't fire correctly: `auth.users` is **shared across every preview-env schema**, and a PR preview's migration (`CREATE TRIGGER ... ON auth.users`) repointed the single shared trigger at another schema (`pr_<n>`), so platform signups create their User+Profile rows in the *wrong* schema. Prod has no preview schemas, so its trigger still works — hence "breaks on Dev, fine on Prod".

**What:**
- Create the marketplace `Profile` at the **application layer**, so it no longer depends on the fragile cross-schema `auth.users` trigger.
- Make the "update profile" endpoint actually **create** a profile when one is missing (it previously raised `ProfileNotFoundError("…This should not be possible.")`).

**How:**
- `get_or_create_user()` (`backend/data/user.py`) now calls `_ensure_user_profile()` — the one choke point every real user passes through. It's **idempotent** (short-circuits if a Profile exists), **race-safe** (catches the `unique(userId)` violation from concurrent creation or the legacy trigger), and **best-effort** (a failure is logged but never blocks user resolution). Because it runs for existing users too, already-broken users **self-heal** on their next authenticated request.
- `update_profile()` (`store/db.py`) creates a Profile from the submitted data when none exists instead of raising, making `settings/profile` a genuine self-service recovery path.

### Changes 🏗️

- `backend/data/user.py`: add `_ensure_user_profile()` + `_generate_profile_username()`; call it from `get_or_create_user()`.
- `backend/api/features/store/db.py`: `update_profile()` creates a Profile when missing instead of raising.
- Tests: `user_test.py` (creates when missing / skips when present / failure is non-fatal) and `db_test.py` (`update_profile` creates when missing).

**`data/*.py` user-ID checks:** `_ensure_user_profile` only ever creates a Profile for the caller's own `user_id` (from the JWT `sub`) — no cross-user access. `update_profile`'s create path likewise uses the authenticated `user_id`; the existing ownership check on the update path is unchanged.

### Not in this PR (follow-ups)

- Backfilling the existing profile-less users is handled organically (they self-heal on next login); an immediate ops unblock can be done with a one-off `INSERT` if needed sooner.
- Dropping the now-redundant `auth.users` trigger, and migrating preview envs to **isolated Supabase branches** so a preview migration can't clobber shared `auth.users` again — tracked separately.

#### For code changes:
- [x] I have made a test plan
  - [x] `poetry run pytest backend/data/user_test.py::TestGetOrCreateUserProfile` — new user gets a Profile; existing Profile isn't duplicated; profile-creation failure doesn't break user resolution
  - [x] `poetry run pytest backend/api/features/store/db_test.py::test_update_profile_creates_when_missing` — saving a profile with none existing creates one; existing `test_update_profile` still passes
  - [ ] Manual (pending): sign up a fresh user on a preview/dev env → confirm a Profile exists and they can publish to the marketplace *(not runnable in this environment)*

### DIFF
--- a/autogpt_platform/backend/backend/api/features/store/db.py
+++ b/autogpt_platform/backend/backend/api/features/store/db.py
@@ -1164,9 +1164,24 @@ async def update_profile(
             where={"userId": user_id}
         )
         if not existing_profile:
-            raise store_exceptions.ProfileNotFoundError(
-                f"Profile not found for user {user_id}. This should not be possible."
+            # No Profile yet (e.g. a user whose auto-creation never ran). This
+            # endpoint is the user's self-service path to a marketplace
+            # profile, so create one from the submitted data instead of
+            # failing — otherwise the user is left with no way to publish.
+            logger.info(
+                f"No profile for user {user_id}; creating one from submitted data"
             )
+            created_profile = await prisma.models.Profile.prisma().create(
+                data=prisma.types.ProfileCreateInput(
+                    userId=user_id,
+                    name=profile.name,
+                    username=username,
+                    description=profile.description,
+                    links=profile.links,
+                    avatarUrl=profile.avatar_url,
+                )
+            )
+            return store_model.ProfileDetails.from_db(created_profile)
 
         # Verify that the user is authorized to update this profile
         if existing_profile.userId != user_id:
--- a/autogpt_platform/backend/backend/api/features/store/db.py
+++ b/autogpt_platform/backend/backend/api/features/store/db.py
@@ -1171,17 +1171,27 @@ async def update_profile(
             logger.info(
                 f"No profile for user {user_id}; creating one from submitted data"
             )
-            created_profile = await prisma.models.Profile.prisma().create(
-                data=prisma.types.ProfileCreateInput(
-                    userId=user_id,
-                    name=profile.name,
-                    username=username,
-                    description=profile.description,
-                    links=profile.links,
-                    avatarUrl=profile.avatar_url,
+            try:
+                created_profile = await prisma.models.Profile.prisma().create(
+                    data=prisma.types.ProfileCreateInput(
+                        userId=user_id,
+                        name=profile.name,
+                        username=username,
+                        description=profile.description,
+                        links=profile.links,
+                        avatarUrl=profile.avatar_url,
+                    )
                 )
-            )
-            return store_model.ProfileDetails.from_db(created_profile)
+                return store_model.ProfileDetails.from_db(created_profile)
+            except prisma.errors.UniqueViolationError:
+                # A concurrent request (or get_or_create_user) created the
+                # Profile first. Re-fetch and fall through to update it with the
+                # submitted data rather than failing the save.
+                existing_profile = await prisma.models.Profile.prisma().find_first(
+                    where={"userId": user_id}
+                )
+                if not existing_profile:
+                    raise
 
         # Verify that the user is authorized to update this profile
         if existing_profile.userId != user_id:
--- a/autogpt_platform/backend/backend/data/user.py
+++ b/autogpt_platform/backend/backend/data/user.py
@@ -62,14 +62,8 @@ async def get_or_create_user(user_data: dict) -> User:
             )
 
         # Ensure every user has a marketplace Profile (required to publish
-        # agents). This was historically created by a Postgres trigger on
-        # auth.users, but that trigger is unreliable — e.g. in preview
-        # environments a single shared auth.users trigger gets repointed at
-        # another schema, leaving platform users with no Profile and no way to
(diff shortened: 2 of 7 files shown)


autogpt_platform/backend/backend/data/user.py
```diff
@@ -53,11 +61,105 @@ async def get_or_create_user(user_data: dict) -> User:
                 )
             )
 
+        # Ensure every user has a marketplace Profile (required to publish
+        # agents). This was historically created by a Postgres trigger on
+        # auth.users, but that trigger is unreliable — e.g. in preview
+        # environments a single shared auth.users trigger gets repointed at
+        # another schema, leaving platform users with no Profile and no way to
+        # publish. Do it here at the app layer, the one choke point every real
```
Pwuts: Irrelevant historical context, can be omitted

0ubbe (dismissed):

autogpt_platform/backend/backend/api/features/store/db.py
```diff
@@ -1164,9 +1164,24 @@ async def update_profile(
...
+                data=prisma.types.ProfileCreateInput(
+                    userId=user_id,
+                    name=profile.name,
+                    username=username,
+                    description=profile.description,
+                    links=profile.links,
+                    avatarUrl=profile.avatar_url,
+                )
+            )
+            return store_model.ProfileDetails.from_db(created_profile)
```
Pwuts: For this UX path to work, the UI of `/settings/profile` also needs to not crash if the user doesn't have a profile yet

autogpt_platform/backend/backend/data/user.py
Pwuts: 🤖 Fixed in [74d86c8c5](https://github.com/Significant-Gravitas/AutoGPT/commit/74d86c8c5aa42e9f0c05eb02196195437c1bb14a): trimmed the historical auth.users-trigger / preview-env backstory from the comment — now just states the intent + best-effort caveat.

autogpt_platform/backend/backend/api/features/store/db.py
Pwuts: 🤖 Fixed in [82b9f2c1b](https://github.com/Significant-Gravitas/AutoGPT/commit/82b9f2c1b496ecef23c1949f70104cd323be9fa1): `/settings/profile` now treats a 404 from `GET /store/profile` as "no profile yet" and renders the empty form instead of an ErrorCard, so **Save** creates the profile via `update_or_create_profile`. Added an integration test covering the no-profile case (`main.test.tsx`).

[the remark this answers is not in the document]
Pwuts: 🤖 Fixed in [09c5a17b9](https://github.com/Significant-Gravitas/AutoGPT/commit/09c5a17b9938aa5add5ea2321ce889ec89d2efd6): the create branch now catches `UniqueViolationError`, re-fetches the now-existing Profile, and falls through to the update path — so a concurrent create no longer returns `DatabaseError`. Added `test_update_profile_tolerates_create_race`.

autogpt_platform/backend/backend/data/user.py
```diff
@@ -53,11 +61,99 @@ async def get_or_create_user(user_data: dict) -> User:
...
+                description="I'm new here",
+                links=[],
+                avatarUrl="",
+            )
+        )
+    except UniqueViolationError:
+        # Created concurrently (another in-flight request, or the legacy
+        # auth.users trigger). The Profile now exists — nothing to do.
+        logger.debug("Profile for user %s already created concurrently", user_id)
+
```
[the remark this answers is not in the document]
Pwuts: 🤖 Fixed in [18a373a42](https://github.com/Significant-Gravitas/AutoGPT/commit/18a373a424f62b42444d9773e92d5512e8246aac): the `UniqueViolationError` catch now re-checks `find_unique(userId)` — it only no-ops when the row now exists (concurrent userId create), and otherwise retries `create()` with a fresh generated username (up to 3x) so a username clash can't leave the user without a Profile. Added `test_retries_profile_create_on_username_collision`.

PR #13476, opened 2026-07-03
feat(frontend): copilot task progress bar above chat input (flagged)

Abhi1992002:
### Why / What / How

**Why:** The copilot task list ("autopilot progress") lived in the artifacts right‑sidebar, where it was easy to miss while chatting. It's more useful directly in the conversation flow, next to where the user is working.

**What:** Adds a `TaskProgressBar` — a collapsible task‑list bar rendered just above the chat input — gated behind a new `task-progress-bar` LaunchDarkly flag (default **off**). When the flag is on, the sidebar `ProgressTab` (and its in‑chat "progress in the sidebar" notice) are removed and the bar becomes the single task surface; when off, everything behaves exactly as before.

**How:** `getLatestTaskList`/`TodoItem` moved into `TaskProgressBar/helpers.ts` (shared by both surfaces). `ChatContainer` derives todos from the live session messages and passes them to the bar. The bar shows the current running step (spinner), completed steps (tick), an amber "stopped" icon for a step interrupted by Stop, and "All tasks complete" at the end; it uses `isStreaming` so a stale `in_progress` never spins forever, and skips partial/streaming `TodoWrite` inputs to avoid flicker. The session→UI converter now recovers orphan `TodoWrite` results (results persisted without their assistant `tool_calls` row) so a finished plan doesn't reappear as in‑progress after the stream ends.

### Changes 🏗️

- New `TaskProgressBar` component (+ `helpers.ts`, Storybook stories) above the chat input, wired to real session data in `ChatContainer`.
- New `Flag.TASK_PROGRESS_BAR` (default `false`); gates bar vs. sidebar in `ChatContainer`, `ContextPanel`, `TabSwitcher`, `ContextPanelAutoOpen`, and `ChatMessagesContainer` (`TaskListNotice`).
- `getLatestTaskList`/`TodoItem` relocated to `TaskProgressBar/helpers.ts`; `ProgressTab` and other importers repointed; old `ProgressTab/helpers.ts` removed.
- `convertChatSessionToUiMessages`: recover orphan `TodoWrite` results into a synthetic tool part so the final task state survives the post‑stream refetch.
- Minor: darkened the chat‑input border to `zinc-200` to match the bar.

#### For code changes:
- [ ] I have made a test plan
  - [ ] With `task-progress-bar` ON: open a copilot session, ask it to create a 5‑step task list — bar appears above the input, steps progress 1/5 → 5/5 → "All tasks complete"
  - [ ] Stop mid‑task — the running step shows the amber "stopped" icon (no infinite spinner)
  - [ ] Reload the finished session — bar stays "All tasks complete" (no reappearing in‑progress)
  - [ ] With the flag OFF (default): task list still renders in the artifacts sidebar Progress tab, and the in‑chat notice behaves as before
  - [ ] `pnpm format && pnpm lint && pnpm types && pnpm test:unit` pass


0ubbe (dismissed):

autogpt_platform/frontend/src/app/(platform)/copilot/helpers/convertChatSessionToUiMessages.ts
```diff
@@ -263,14 +307,37 @@ export function convertChatSessionMessagesToUiMessages(
...
+          uiMessages.push({
+            id:
+              msg.sequence != null
+                ? `${sessionId}-seq-${msg.sequence}`
+                : `${sessionId}-idx-${idx}`,
+            role: "assistant",
+            parts: [orphanPart],
+          });
+        }
+      }
```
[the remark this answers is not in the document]
Abhi1992002: Fixed in b5a16e42f0 — confirmed against the backend: `TodoWriteResponse` serializes under `todos`, so `orphanTodoWriteResultToPart` now reads `parsed.todos` instead of the non-existent `newTodos`. Added a regression test.

autogpt_platform/frontend/src/app/(platform)/copilot/helpers/convertChatSessionToUiMessages.ts
```diff
@@ -136,6 +136,50 @@ function toToolInput(rawArguments: unknown): unknown {
...
+  consumedToolCallIds: ReadonlySet<string>,
+): UIMessage<unknown, UIDataTypes, UITools>["parts"][number] | null {
+  if (!msg.tool_call_id || consumedToolCallIds.has(msg.tool_call_id)) {
+    return null;
+  }
+  if (typeof msg.content !== "string") return null;
+  const parsed = safeJsonParse(msg.content);
+  if (!parsed || typeof parsed !== "object") return null;
+  const newTodos = (parsed as { newTodos?: unknown }).newTodos;
+  if (!Array.isArray(newTodos) || newTodos.length === 0) return null;
```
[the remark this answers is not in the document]
Abhi1992002: Fixed in b5a16e42f0 — same field-name bug: recovery now reads `todos` (the field the backend's `TodoWriteResponse` emits) instead of `newTodos`, plus a regression test covering the orphan recovery path.

autogpt_platform/frontend/src/app/(platform)/copilot/components/ContextPanel/ContextPanel.tsx
```diff
@@ -33,6 +34,9 @@ export function ContextPanel({ sessionId, mobile }: Props) {
   } = useContextPanel();
   const { uploaded, generated } = useSessionFiles(sessionId);
   const filesCount = uploaded.length + generated.length;
+  // When the task bar (above the chat input) is on, the sidebar drops the
+  // Progress tab and shows Files only.
+  const showProgressTab = !useGetFlag(Flag.TASK_PROGRESS_BAR);
```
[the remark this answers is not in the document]
Abhi1992002: Fixed in b5a16e42f0 — `ContextPanel` now derives `effectiveTab = showProgressTab ? activeTab : "files"` and passes it to both `TabSwitcher` and the content switch, so a persisted `"progress"` tab never leaves the switcher with no selected tab when the Progress tab is hidden.

autogpt_platform/frontend/src/app/(platform)/copilot/helpers/convertChatSessionToUiMessages.ts
```diff
@@ -136,6 +136,50 @@ function toToolInput(rawArguments: unknown): unknown {
...
+  return {
+    type: "tool-TodoWrite",
+    toolCallId: msg.tool_call_id,
+    state: "output-available",
+    input: { todos: newTodos },
+    output: parsed,
+  } as UIMessage<unknown, UIDataTypes, UITools>["parts"][number];
+}
```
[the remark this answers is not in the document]
Abhi1992002: Fixed in b5a16e42f0 — `orphanTodoWriteResultToPart` now trims `msg.tool_call_id` before the `consumedToolCallIds` check and reuses the normalized id for the synthesized part, matching `collectConsumedToolCallIds`. Added a test asserting a whitespace-padded id doesn't produce a duplicate `tool-TodoWrite` part.

autogpt_platform/frontend/src/services/feature-flags/use-get-flag.ts
```diff
@@ -49,9 +52,10 @@ const defaultFlags = {
...
   [Flag.CHAT_SEARCH]: false,
   [Flag.CHAT_SHARING]: false,
-  [Flag.AUTOGPT_NEW_LAYOUT]: false,
+  [Flag.AUTOGPT_NEW_LAYOUT]: true,
   [Flag.CHAT_WORKSPACE_FILES]: false,
   [Flag.CHAT_PINNING]: false,
+  [Flag.TASK_PROGRESS_BAR]: true,
   [Flag.GRAPHITI_MEMORY]: false,
   [Flag.GRAPHITI_COMMUNITIES_ENABLED]: false,
   [Flag.DREAM_PASS_ENABLED]: false,
```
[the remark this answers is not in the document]
Abhi1992002: Fixed in 4c2f7e3d41 — good catch. Changed `Flag.TASK_PROGRESS_BAR` default to `false` in `defaultFlags` and added the missing `case Flag.TASK_PROGRESS_BAR` in `readEnvOverride` so `NEXT_PUBLIC_FORCE_FLAG_TASK_PROGRESS_BAR` now works for local overrides.

autogpt_platform/frontend/src/app/(platform)/copilot/helpers/convertChatSessionToUiMessages.ts
```diff
@@ -263,14 +308,37 @@ export function convertChatSessionMessagesToUiMessages(
...
+    if (msg.role === "tool") {
+      // Orphan TodoWrite results (no matching assistant tool_calls row) carry
+      // the newest task-list state — fold a synthetic part into the previous
+      // assistant bubble so progress UIs see it. TodoWrite renders as null in
+      // chat, so this adds no visible bubble content.
+      const orphanPart = orphanTodoWriteResultToPart(msg, consumedToolCallIds);
+      if (orphanPart) {
+        const prevUI = uiMessages[uiMessages.length - 1];
+        if (prevUI && prevUI.role === "assistant") {
+          prevUI.parts.push(orphanPart);
```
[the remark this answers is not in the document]
Abhi1992002: Acknowledged, but this one isn't practical to fix the suggested way. The recovery path only fires for *orphaned* results — tool results whose `tool_call_id` has **no** matching assistant `tool_calls` row (that's exactly the `!consumedToolCallIds.has(toolCallId)` guard). Since the assistant row is what carries the tool name, an orphan has no tool name to look up or validate against — shape (`todos: []`) is the only signal available. A `parsed.type` discriminator check would depend on `TodoWriteResponse` emitting such a field, which it doesn't today, so it'd silently disable recovery. Given no current tool emits a `todos` array and the guard already scopes this to orphans only, I'm leaving the shape check as-is; happy to revisit if a second `todos`-shaped tool ever lands.

autogpt_platform/frontend/src/app/(platform)/copilot/CopilotPage.tsx
```diff
@@ -64,14 +64,19 @@ export function CopilotPage() {
...
+      // (SidebarProvider `min-h-svh` → SidebarInset `flex-1` → `section flex-1`)
+      // only set a *minimum* height, so `height: 100%` there resolves to
+      // content height and the accordion pushes the input below the fold.
+      // Subtract the inset header in the new layout; the navbar + preview
+      // banner in the classic one. `svh` keeps the input visible when mobile
+      // browser chrome is shown.
       style={
         showNewLayout
-          ? { height: "100%" }
+          ? { height: `calc(100svh - ${INSET_HEADER_HEIGHT_PX}px)` }
```
0ubbe, line 79: Just making sure this does not break the layout in any way 💅🏽

autogpt_platform/frontend/src/app/(platform)/copilot/components/ContextPanel/ContextPanel.tsx
```diff
@@ -33,6 +34,12 @@ export function ContextPanel({ sessionId, mobile }: Props) {
...
+  // Clamp a persisted "progress" tab to "files" when Progress is hidden, so
+  // the switcher never ends up with no selected tab.
+  const effectiveTab = showProgressTab ? activeTab : "files";
```
0ubbe, line 42: At this point, maybe there is no need for the `Progress` tab in the context panel at all? I haven't seen the new UI you did, but worth checking with product 🍵

autogpt_platform/frontend/src/app/(platform)/copilot/components/ContextPanel/ContextPanelAutoOpen.tsx
```diff
@@ -9,8 +10,10 @@ interface Props {
 }
 
 export function ContextPanelAutoOpen({ sessionId }: Props) {
+  // The sidebar auto-opens on progress only when the task bar is off.
+  const taskBarEnabled = useGetFlag(Flag.TASK_PROGRESS_BAR);
   useCollapseContextPanelOnSession(sessionId);
   useAutoOpenForFiles(sessionId);
```
0ubbe, line 16: Same we could remove the Progress bar all together and avoid this?

autogpt_platform/frontend/src/app/(platform)/copilot/components/TaskProgressBar/TaskProgressBar.tsx
```diff
@@ -0,0 +1,299 @@
...
+              ))}
+            </ul>
+          </motion.div>
+        )}
+      </AnimatePresence>
+    </div>
+  );
+}
+
+function TaskRow({
```
0ubbe: Make it a sub-component under `/components` folder on this folder 📁 ( 1 component per file rule ); same for the other components on this file.

autogpt_platform/frontend/src/app/(platform)/copilot/components/TaskProgressBar/__tests__/helpers.test.ts
```diff
@@ -0,0 +1,132 @@
...
+
+function todo(
+  content: string,
+  status: TodoItem["status"],
+  activeForm?: string,
+): TodoItem {
+  return { content, status, activeForm };
+}
+
+function messageWithTodoWrite(todos: unknown, state?: string): UIMessage {
```
0ubbe, line 20: I don't think there is a need to test the helpers if we are testing the component already on `TaskProgressBar.test.tsx` 💅🏽

autogpt_platform/frontend/src/services/feature-flags/use-get-flag.ts
```diff
@@ -20,6 +20,9 @@ export enum Flag {
   AUTOGPT_NEW_LAYOUT = "autogpt-new-layout",
   CHAT_WORKSPACE_FILES = "chat-workspace-files",
   CHAT_PINNING = "chat-pinning",
+  // When on, the copilot task list renders as a bar above the chat input
+  // (TaskProgressBar); when off, it stays in the artifacts sidebar (ProgressTab).
```
0ubbe: No need for these comments, just add noise ☕

autogpt_platform/frontend/src/services/feature-flags/use-get-flag.ts
```diff
@@ -49,9 +52,10 @@ const defaultFlags = {
   [Flag.GENERIC_TRIGGER_AGENTS]: false,
   [Flag.CHAT_SEARCH]: false,
   [Flag.CHAT_SHARING]: false,
-  [Flag.AUTOGPT_NEW_LAYOUT]: false,
+  [Flag.AUTOGPT_NEW_LAYOUT]: true,
```
0ubbe: Let's never default to `true` the flags, otherwise if launch darkly does not load, the feature is visible which most of the time is not what we want.

0ubbe (commented):
Cool cool 💚 , some comments

autogpt_platform/frontend/src/app/(platform)/copilot/components/TaskProgressBar/TaskProgressBar.tsx
Abhi1992002: Done in 97b13ff6ea — `StatusIcon` and `TaskRow` are now each their own file under `TaskProgressBar/components/` (one component per file). Shared easing/timing constants moved to `animations.ts` and the `toDisplayStatus`/`DisplayStatus` helper into `helpers.ts`, so `TaskProgressBar.tsx` only holds the top-level component now.

autogpt_platform/frontend/src/services/feature-flags/use-get-flag.ts
Abhi1992002: Removed in a284d26ad3.

autogpt_platform/frontend/src/services/feature-flags/use-get-flag.ts
Abhi1992002: Fixed in a284d26ad3 — that `true` was an accidental flip of the unrelated `autogpt-new-layout` flag; reverted to `false` to match dev so it stays fail-closed if LaunchDarkly doesn't load. The new `task-progress-bar` flag already defaults to `false`. Fully agree on the never-default-true principle.

autogpt_platform/frontend/src/app/(platform)/copilot/CopilotPage.tsx
Abhi1992002: It shouldn't 🙏 — the height stays viewport-bound (`calc(100svh - INSET_HEADER_HEIGHT_PX)`) instead of `100%`. The reason for the change: the new-layout ancestors (`SidebarProvider min-h-svh` → `SidebarInset flex-1` → `section flex-1`) only set a *minimum* height, so `height: 100%` resolved to content height and expanding the task-progress accordion pushed the chat input below the fold. A definite height lets the inner `min-h-0` chain make the messages area absorb the growth. `svh` also keeps the input visible when mobile browser chrome shows. Classic layout keeps subtracting navbar + preview banner as before.

autogpt_platform/frontend/src/app/(platform)/copilot/components/TaskProgressBar/__tests__/helpers.test.ts
Abhi1992002: I'd lean toward keeping these — they're complementary rather than redundant. `helpers.test.ts` covers pure-function edge cases the component render test doesn't reach: `getLatestTaskList` skipping streaming/partial `TodoWrite` inputs, ignoring malformed todo entries, and the non-array/orphan-recovery paths. Exercising those through the rendered component would need a lot of message-history scaffolding, and they also guard the codecov patch gate on `helpers.ts`. Happy to trim any case that genuinely duplicates the component test if you can point one out.

autogpt_platform/frontend/src/app/(platform)/copilot/components/ContextPanel/ContextPanel.tsx
Abhi1992002: Agreed that's the end state 🍵 — I kept it flag-gated for this PR so we get safe rollout + instant rollback: with `task-progress-bar` OFF the sidebar Progress tab still works exactly as today, and only when it's fully rolled out and product signs off do we delete the tab, the auto-open path, and the in-chat notice for good. I'll do that removal as a dedicated follow-up cleanup PR once the flag is defaulted on.

autogpt_platform/frontend/src/app/(platform)/copilot/components/ContextPanel/ContextPanelAutoOpen.tsx
Abhi1992002: Same answer as the ContextPanel thread — this branch only exists to keep the OFF path behaving as today. Once the flag is fully on I'll rip out the progress auto-open (and the tab + notice) in a single follow-up cleanup PR rather than leaving dead code behind a defaulted-on flag.

autogpt_platform/frontend/src/app/(platform)/copilot/helpers/convertChatSessionToUiMessages.ts
```diff
@@ -263,14 +308,37 @@ export function convertChatSessionMessagesToUiMessages(
 
   const uiMessages: UIMessage<unknown, UIDataTypes, UITools>[] = [];
   const stats: TurnStatsMap = new Map();
```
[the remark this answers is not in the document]
Abhi1992002: Fixed in 2b37f105d9 — good catch. The output map now stores under `msg.tool_call_id.trim()`, matching the trimmed lookup at read time, so a whitespace-padded id no longer misses and hangs the tool call on a spinner. Added a regression test asserting a padded output id resolves to `output-available`.

0ubbe (approved):

PR #13490, opened 2026-07-06
fix(copilot): suppress error banner and stop-flicker on user cancel

Abhi1992002:
### Why / What / How

**Why:** Pressing **Stop** on a Copilot turn produced two bad symptoms. (1) A red *"The assistant encountered an error. Please try sending your message again."* banner appeared even though the user asked for the stop. (2) The tool would visibly re-load and the transcript duplicated for ~2-3 seconds before the "Task stopped" card finally settled — the stop looked like it was ignored, and users sometimes had to press Stop twice.

**What:** Stop a user cancel from being surfaced as an error, and stop the frontend from re-arming auto-resume (which replayed the whole transcript) while the cancelled turn is still draining on the backend.

**How:** On the backend, `mark_session_completed` now skips publishing the `StreamError` frame for user-initiated cancels — both the `/cancel` route's force-complete branch and the executor's `finally` block — while still marking the turn failed (locks released, queued turns promoted); `/cancel` also returns the `cancel_published_not_confirmed` reason the frontend already had a handler for. On the frontend, the user-stopping flag is held until `/cancel` confirms the stream is actually gone (never cleared from the stale pre-stop session cache), post-stream hydration is deferred while a stop is in flight, and the locally-painted cancellation marker is preserved so the "Task stopped" card never disappears.

### Changes 🏗️

- **backend/`chat/routes.py`**: `/cancel` force-complete passes `skip_error_publish=True` and returns `reason="cancel_published_not_confirmed"`.
- **backend/`copilot/executor/processor.py`**: track `user_cancelled` and skip the StreamError publish for user cancels in the turn's `finally` block.
- **backend/`chat/routes_test.py`**: regression test asserting the force-complete branch suppresses the StreamError and returns the confirmed reason.
- **frontend/`useCopilotStop.ts`**: keep the user-stop flag set until a post-cancel session refetch confirms `active_stream` is gone.
- **frontend/`useCopilotStream.ts`**: gate the flag-reset effect on a new `isCancelInFlightRef` so auto-resume can't re-arm mid-drain.
- **frontend/`useHydrateOnStreamEnd.ts`**: defer hydration while stopping, preserve the local stop marker, and suppress the misleading "Previous response was interrupted" toast for user stops (+ test arg update).

#### For code changes:
- [x] I have made a test plan
  - [x] Backend: `poetry run pytest backend/api/features/chat/routes_test.py -k cancel` (9 passed, incl. new force-complete regression test)
  - [ ] Frontend: `pnpm test:unit` for the copilot hydrate/stop hooks (run manually)
  - [ ] Manual: start a long tool-using turn, press Stop mid-generation, confirm the "Task stopped" card appears instantly with no error banner, no transcript duplication, and no tool spinners flickering back

### DIFF
--- a/autogpt_platform/backend/backend/api/features/chat/routes.py
+++ b/autogpt_platform/backend/backend/api/features/chat/routes.py
@@ -1033,11 +1033,20 @@ async def cancel_session_task(
     logger.warning(
         f"[CANCEL] Session ...{session_id[-8:]} not confirmed after {max_wait}s, force-completing"
     )
-    await stream_registry.mark_session_completed(session_id, error_message="Cancelled")
+    # ``skip_error_publish``: the user asked for this stop, so surfacing it to
+    # the frontend as a StreamError would render a spurious "assistant
+    # encountered an error" banner on any live/resumed stream. The persisted
+    # cancellation marker on the session messages already tells the UI (and
+    # any later reload) that the turn was stopped.
+    await stream_registry.mark_session_completed(
+        session_id, error_message="Cancelled", skip_error_publish=True
+    )
     # Status is now force-flipped out of "running"; re-clear to drop any
     # follow-up that landed during the poll window.
     await _clear_pending_best_effort(session_id)
-    return CancelSessionResponse(cancelled=True)
+    return CancelSessionResponse(
+        cancelled=True, reason="cancel_published_not_confirmed"
+    )
 
 
 def _ui_message_stream_headers() -> dict[str, str]:
--- a/autogpt_platform/backend/backend/copilot/executor/processor.py
+++ b/autogpt_platform/backend/backend/copilot/executor/processor.py
@@ -497,6 +497,7 @@ class CoPilotProcessor:
         last_refresh = time.monotonic()
         refresh_interval = 30.0  # Refresh lock every 30 seconds
         error_msg = None
+        user_cancelled = False
 
         try:
             # Choose service based on LaunchDarkly flag.
@@ -588,6 +589,7 @@ class CoPilotProcessor:
             if isinstance(e, asyncio.CancelledError):
                 log.info("Turn cancelled")
                 error_msg = "Operation cancelled"
+                user_cancelled = True
             else:
                 error_msg = str(e) or type(e).__name__
                 log.error(f"Turn failed: {error_msg}")
@@ -596,9 +598,17 @@ class CoPilotProcessor:
             # If no exception but user cancelled, still mark as cancelled
             if not error_msg and cancel.is_set():
                 error_msg = "Operation cancelled"
+                user_cancelled = True
             try:
+                # A user-initiated cancel still marks the session "failed"
+                # (so queued turns are promoted and locks released), but must
+                # not publish a StreamError: the frontend would render it as
+                # "the assistant encountered an error" on the live/resumed
+                # stream even though the user asked for the stop.
                 await stream_registry.mark_session_completed(
-                    entry.session_id, error_message=error_msg
+                    entry.session_id,
+                    error_message=error_msg,
+                    skip_error_publish=user_cancelled,
                 )
             except Exception as mark_err:
                 log.error(f"Failed to mark session completed: {mark_err}")
--- a/autogpt_platform/frontend/src/app/(platform)/copilot/useCopilotStop.ts
+++ b/autogpt_platform/frontend/src/app/(platform)/copilot/useCopilotStop.ts
@@ -107,24 +107,30 @@ export function useCopilotStop({
         });
       }
     } finally {
-      isCancelInFlightRef.current = false;
-      // The cancel endpoint returns once the backend confirmed (or
-      // force-completed) the stop, so this refetch reads the settled
-      // state. Only clear the user-stop flag when the backend actually
-      // reports no active stream — clearing it against a stale cache is
-      // what used to re-arm auto-resume mid-drain.
-      if (sessionId) {
-        try {
+      // Keep ``isCancelInFlightRef`` set through the confirming refetch. The
+      // reset effect in useCopilotStream is gated on it, so clearing it before
+      // the settled session state arrives would let that effect clear the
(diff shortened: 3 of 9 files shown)


autogpt_platform/frontend/src/app/(platform)/copilot/useCopilotStop.ts
```diff
@@ -43,54 +53,79 @@ export function useCopilotStop({
...
+    } finally {
+      isCancelInFlightRef.current = false;
+      // The cancel endpoint returns once the backend confirmed (or
+      // force-completed) the stop, so this refetch reads the settled
+      // state. Only clear the user-stop flag when the backend actually
+      // reports no active stream — clearing it against a stale cache is
+      // what used to re-arm auto-resume mid-drain.
+      if (sessionId) {
+        try {
+          const result = await refetchSession();
```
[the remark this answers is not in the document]
Abhi1992002: Fixed in 05b18862: `isCancelInFlightRef` now stays set through the confirming `refetchSession()` and is only released in an inner `finally` once the refetch settles. This closes the window where the `hasActiveStream === false` reset effect could clear `isUserStoppingRef` from the stale pre-stop cache and re-arm auto-resume mid-drain.

autogpt_platform/frontend/src/app/(platform)/copilot/useHydrateOnStreamEnd.ts
```diff
@@ -92,6 +93,53 @@ function preservePromotedUserBubbles(
...
+  const last = hydrated[hydrated.length - 1];
+  if (last?.role === "assistant") {
+    return [
+      ...hydrated.slice(0, -1),
+      { ...last, parts: [...last.parts, ...markerParts] },
+    ];
+  }
+  // Stopped before any assistant row was persisted — keep the whole local
+  // marker message so the stopped state has something to render from.
+  return [...hydrated, prevMarked];
```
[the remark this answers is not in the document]
Abhi1992002: Fixed in 05b18862: `preserveLocalStopMarker` now reads only the trailing `prev` message (the sole row the stop handler paints), requires it to be an assistant, and matches specifically the `Operation cancelled` error marker via `isCancellationMarkerText`. If no cancellation marker part is present it returns `hydrated` unchanged, so an older error/system marker from a prior turn can no longer be reattached to the cancelled turn.

autogpt_platform/backend/backend/api/features/chat/routes.py
```diff
@@ -1033,11 +1033,20 @@ async def cancel_session_task(
...
+    # any later reload) that the turn was stopped.
+    await stream_registry.mark_session_completed(
+        session_id, error_message="Cancelled", skip_error_publish=True
+    )
     # Status is now force-flipped out of "running"; re-clear to drop any
     # follow-up that landed during the poll window.
     await _clear_pending_best_effort(session_id)
-    return CancelSessionResponse(cancelled=True)
+    return CancelSessionResponse(
+        cancelled=True, reason="cancel_published_not_confirmed"
```
Pwuts, line 1048: what's `cancel_published_not_confirmed`?

autogpt_platform/backend/backend/copilot/executor/processor.py
```diff
@@ -596,9 +598,17 @@ async def _execute_async(
...
+                # A user-initiated cancel still marks the session "failed"
+                # (so queued turns are promoted and locks released), but must
+                # not publish a StreamError: the frontend would render it as
+                # "the assistant encountered an error" on the live/resumed
+                # stream even though the user asked for the stop.
                 await stream_registry.mark_session_completed(
-                    entry.session_id, error_message=error_msg
+                    entry.session_id,
+                    error_message=error_msg,
+                    skip_error_publish=user_cancelled,
```
kcze, line 611: 🤖 🟠 **Should Fix**: For baseline-mode turns (`stream_chat_completion_baseline`), the cancel path persists partial output but never writes a `COPILOT_ERROR_PREFIX` cancellation marker — only the SDK path does (`sdk/service.py` via `interrupted.finalize`, which catches `CancelledError`). Baseline's `except Exception` never catches `CancelledError` (it is a `BaseException`), and `baseline/service.py` has zero `COPILOT_ERROR_PREFIX` writes. So passing `skip_error_publish=user_cancelled` here removes the only cancellation signal for baseline cancels: another tab / resumed subscriber (which never painted the local stop marker) and any later reload receive just `StreamFinish` and render the partial response as a normally completed turn — contradicting the assumption that a persisted marker already tells the UI the turn was stopped. Persist a cancellation marker in the baseline cancel path before suppressing the `StreamError`, or scope `skip_error_publish` to the SDK path. _(flagged by: Codex)_

autogpt_platform/frontend/src/app/(platform)/copilot/useHydrateOnStreamEnd.ts
```diff
@@ -184,6 +261,10 @@ export function useHydrateOnStreamEnd({
     if (!hydratedMessages || hydratedMessages.length === 0) return;
     if (status === "streaming" || status === "submitted") return;
     if (isReconnectScheduled) return;
+    if (isUserStoppingRef.current) {
```
kcze, line 264: 🤖 🟠 **Should Fix**: This user-stop deferral is only reached after the `status === "streaming" || "submitted"` early-return above it, so it never runs while the stream is still streaming. When stopping a resumed GET stream, the AI SDK `status` can stay `"streaming"` for seconds after `/cancel` confirms and `useCopilotStop` clears `isUserStoppingRef` — every effect run in that window bails at the status guard without setting `deferredForUserStopRef`. When `status` finally flips to `ready`, the force-hydrate applies with `wasUserStop === false`, skipping `preserveLocalStopMarker`; if the backend snapshot has no marker yet, the locally-painted "Task stopped" card disappears — the exact symptom this PR fixes. Capture the stop deferral before the active-status early return (or thread a persistent stop generation into this hook). _(flagged by: Codex)_

PR #13524, opened 2026-07-10
fix(backend): authorize team management by target team id, drop active-team requirement

ntindle:
### Why / What / How

**Why:** Mutating team routes (settings update, member add/role-change/remove) required the caller's *active* team context (`X-Team-Id` header) to equal the target team. With the teams-as-badges/filters product direction there is no "active team" to stand in — and the old gate produced an incoherent asymmetry: an org admin could **delete** any team (org-level gate) but couldn't rename it or manage its members.

**What:** Authorization for team management now targets the `ws_id` from the URL path, independent of header context: allowed if the caller is an admin member of the target team, or holds org-level `MANAGE_WORKSPACES` (owner/admin). Everyone else gets 403.

**How:** New local `_authorize_team_management(ctx, org_id, ws_id)` helper in `team_routes.py` — verifies org membership, verifies the team belongs to the path org, then checks org `MANAGE_WORKSPACES` (via the existing `check_org_permission` matrix helper) or the new `team_db.is_team_admin(ws_id, user_id)` lookup. Routes switch from `Security(requires_team_permission(...))` to `Security(get_request_context)` + the helper. `requires_team_permission` in autogpt_libs is untouched (active-team semantics preserved for any other consumers). Default-team invariants (join_policy immutable, undeletable, unleavable, last-admin removal blocked) live in the db layer and are unchanged.

Linear: SECRT-2453

### Changes 🏗️

- `team_routes.py`: `_authorize_team_management` helper; settings-update + member add/update/remove routes authorize against the target team instead of the active-team header
- `team_db.py`: `is_team_admin(ws_id, user_id)` lookup
- `routes_test.py`: scenario tests — team admin manages own team with **no** X-Team-Id header, org admin manages a team they're not a member of, plain org member 403, admin of a *different* team 403, cross-org 403; both settings and member-management routes
- `regression_test.py`: active-team-mismatch regression updated to the new semantics

#### For code changes:
- [x] I have made a test plan
  - [x] `poetry run pytest backend/api/features/orgs/ -q` — 242 passed, 12 xfailed

---

> [!NOTE]
> **Medium Risk**
> Changes authorization for workspace management APIs; behavior is security-sensitive but scoped, with broad new tests and org/path guards retained.
> 
> **Overview**
> Team **settings and member mutations** no longer require `X-Team-Id` to match the path `ws_id`. They authorize against the **target workspace in the URL** instead of the caller’s active-team context.
> 
> A new `_authorize_team_management` helper checks path org membership, that the team belongs to that org, then allows **org `MANAGE_WORKSPACES`** or an **active admin** on that specific team via `team_db.is_team_admin`. `update_team`, `add_member`, `update_member`, and `remove_member` use `get_request_context` plus this helper instead of `requires_team_permission` and the old `ctx.team_id != ws_id` gate. **Workspace delete** stays org-admin-only and is unchanged.
> 
> Tests cover org admins managing teams they’re not on, team admins without an active-team header, denial for plain members and admins of other teams, and cross-org guards (SECRT-2453).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5112e5e8024dc556c4ec51dc7f55784dd7ef9940. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


ntindle:
🤖 Addressed the review in [5112e5e80](https://github.com/Significant-Gravitas/AutoGPT/commit/5112e5e8024dc556c4ec51dc7f55784dd7ef9940):

**🔴 Blocker 1 — backend lint (black) red** — Fixed. Ran `poetry run format`; black wanted a second blank line between the new `TestTeamManagementByTeamId` class and `TestCanonicalPersonalOrgOrdering` (introduced by the dev-merge union in f7010a1). `poetry run lint` now exits 0 locally.

**🟠 Should Fix 1 — no route-level auth test for `add_member`/`remove_member`** — Fixed. Added allow + deny route tests for both: team admin adds a member with no `X-Team-Id` (200, `teammember.create` awaited), plain org member denied (403, create not awaited), org admin removes a member of a team they don't belong to (204, `teammember.delete` awaited), admin of a *different* team denied (403, delete not awaited). Dropping `_authorize_team_management` from either route now fails tests.

**🟠 Should Fix 2 — `is_team_admin` denial branches untested** — Fixed. Route-level tests now exercise both security-relevant branches: an ACTIVE member with `isAdmin=False`, and admin rows with `status="INVITED"`/`"SUSPENDED"` (parametrized) — all 403 with the mutation not awaited. Also added a test for the first helper branch (`ctx.org_id != org_id` → 403 before any DB lookup), covering the cross-org header tenancy boundary.

**🟡 NTH 1 — redundant `Team` lookup on member routes** — Keeping as-is, deliberately. `add_team_member`'s own org-ownership re-check is the db-layer invariant that protects every caller (it mirrors `join_team`, which also revalidates), per this module's convention that integrity checks live in `team_db.py` rather than trusting route wiring. Passing the fetched team down would couple the helper's return value to one route while the other three discard it. Cost is a single indexed PK point-read on an infrequent mutating route — defense-in-depth wins here.

**🟡 NTH 2 — document delete-vs-manage asymmetry** — Fixed. `delete_team` now carries a comment stating the asymmetry is deliberate: team admins may rename/manage members, but deletion is destructive and stays org-admin-only (`MANAGE_WORKSPACES`).

**🟡 NTH 3 — extract shared `_authorize_org_membership`** — Deferring. The `ctx.org_id != org_id` guard predates this PR on ~7 routes (create/list/get/delete/join/leave/list_members); this PR already consolidated the four routes that need the full permission chain into `_authorize_team_management`. Folding the remaining 2-line guards into another helper would grow an auth-boundary diff into read routes the PR doesn't otherwise touch. Good follow-up refactor.

**🔵 Nit 1 — magic string `"ACTIVE"`** — Keeping the literal, deliberately. The enum does exist (`TeamMember.status` is `OrgMemberStatus` in `schema.prisma`), but the orgs module (`db.py` + `team_db.py`) uniformly uses string literals in both where-filters and comparisons (~25 occurrences), and no backend module imports `OrgMemberStatus` today. Switching this one comparison would create intra-module inconsistency; a module-wide enum sweep is out of scope for an auth fix. The new status-branch tests now guard this line against typo drift regardless.

Orgs suite: 251 passed, 12 xfailed locally; lint clean.

Bentlybro (dismissed):

autogpt_platform/backend/backend/api/features/orgs/routes_test.py
```diff
@@ -2924,6 +2924,406 @@ async def test_lock_loser_waits_for_winner(self, mocker):
...
+        self.prisma.team.update = AsyncMock()
+        # No active team: team_id is None (no X-Team-Id header).
+        self._use_ctx(_mgmt_ctx(user_id=USER_ID, team_id=None, team_admin=False))
+
+        resp = client.patch(
+            f"/api/orgs/{ORG_ID}/workspaces/{WS_ID}", json={"name": "Renamed"}
+        )
+
+        assert resp.status_code == 200
+        self.prisma.team.update.assert_awaited_once()
```
Pwuts, line 3015: 🤖 🟠 This test is the one that proves the fix, and right now it cannot fail on the thing it is proving.

`self.prisma.teammember.find_unique` is stubbed key-agnostically, so it returns this admin row whatever team is asked about. I changed `_authorize_team_management` to look the caller up by `ctx.team_id` instead of `ws_id` — a straight regression to the pre-SECRT-2453 behaviour — and ran the whole orgs suite: 253 passed, 12 xfailed. Nothing noticed. (The same is true of the `get_team` lookup on line 34 of `team_routes.py`.)

`regression_test.py` does pin header-independence, but only for the org-admin branch: its owner context short-circuits at `check_org_permission` before `is_team_admin` is ever called. The team-admin branch — the one that needs the DB lookup — is unpinned.

Asserting the composite key closes it. I ran this: it passes as-is, and fails under the regression above.

```suggestion
        assert resp.status_code == 200
        self.prisma.team.update.assert_awaited_once()
        # The admin lookup must target the URL's ws_id, not the caller's
        # (absent) active team.
        self.prisma.teammember.find_unique.assert_awaited_with(
            where={"teamId_userId": {"teamId": WS_ID, "userId": USER_ID}}
        )
```

autogpt_platform/backend/backend/api/features/orgs/routes_test.py
```diff
@@ -2924,6 +2924,406 @@ async def test_lock_loser_waits_for_winner(self, mocker):
...
+        self._use_ctx(
+            _mgmt_ctx(user_id=OTHER_USER_ID, team_id="ws-other", team_admin=True)
+        )
+
+        resp = client.patch(
+            f"/api/orgs/{ORG_ID}/workspaces/{WS_ID}", json={"name": "Renamed"}
+        )
+
+        assert resp.status_code == 403
+        self.prisma.team.update.assert_not_awaited()
```
Pwuts, line 3075: 🤖 🟠 Same gap from the other side: this test denies the caller, but would deny them just as happily if the code looked up `ctx.team_id` — the mock returns `None` for every key. Pinning the key here is what makes "the *target* team has no admin row for them" the actual reason for the 403.

```suggestion
        assert resp.status_code == 403
        self.prisma.team.update.assert_not_awaited()
        self.prisma.teammember.find_unique.assert_awaited_with(
            where={"teamId_userId": {"teamId": WS_ID, "userId": OTHER_USER_ID}}
        )
```

autogpt_platform/backend/backend/api/features/orgs/team_db.py
```diff
@@ -148,6 +148,14 @@ async def list_team_members(ws_id: str) -> list[TeamMemberResponse]:
     return [TeamMemberResponse.from_db(m) for m in members]
 
 
+async def is_team_admin(ws_id: str, user_id: str) -> bool:
```
Pwuts: 🤖 🔵 There are now two `is_team_admin` in the codebase meaning different things: `RequestContext.is_team_admin` is about the caller's *active* team, this one is about the *target* team. Confusing the two is exactly the regression the tests above can't catch — `is_admin_of_team` would keep them apart. Not worth a push on its own.

autogpt_platform/backend/backend/api/features/orgs/team_routes.py
```diff
@@ -24,6 +20,27 @@
...
+    """
+    if ctx.org_id != org_id:
+        raise HTTPException(403, detail="Not a member of this organization")
+    await team_db.get_team(ws_id, expected_org_id=org_id)
+    if check_org_permission(ctx, OrgAction.MANAGE_WORKSPACES):
+        return
+    if await team_db.is_team_admin(ws_id, ctx.user_id):
+        return
+    raise HTTPException(
+        403, detail="Must be a team admin or org admin to manage this team"
```
Pwuts: 🤖 🔵 "team admin ... manage this team", but the route summaries, OpenAPI tags and DB errors all say "workspace". Worth aligning next time this file is open.

Pwuts (commented):
## 🤖 Review — #13524

Re-reviewed on `5112e5e8` against today's `dev`, since the approval is a month old and `dev` has moved 278 commits: **it still holds.** `dev` merges into the head with no conflicts, neither `team_routes.py` nor `team_db.py` has been touched underneath it, and `get_request_context` — the primitive the new helper trusts for org identity — is unchanged in the part that matters. On the merged tree the org suite is green (253 passed, 12 xfailed), as are `autogpt_libs/auth` (233), `architecture_test` (3) and `test_block` (1647).

The authorization is sound, and I checked it by breaking it rather than by reading it: ten mutations, eight killed. Every denial the change creates is enforced and pinned — admin of another team, active non-admin member, `INVITED`/`SUSPENDED` admin, plain org member, cross-org caller, cross-org target team.

One should-fix, and it is the reviewer bot's own finding from the approving review, still open: the tests do not pin *which* team is looked up. Regressing `is_team_admin(ws_id, …)` to `is_team_admin(ctx.team_id, …)` — the exact bug this PR fixes — leaves all 253 tests green, because the `teammember.find_unique` mock answers any key with the same row. The code is right; the net has a hole where the change is. Two assertion lines close it, and I've verified they pass on the real code and fail under the regression.

Verdict: **comment**, not a hold — nothing here should stop a merge today, and the assertion is as good in a follow-up as it is here.

### 🟠 Should Fix
- 🟠 No test pins the target-team lookup key — regressing it to the active-team header keeps the whole orgs suite green, so the behaviour this PR exists to deliver is unprotected (`autogpt_platform/backend/backend/api/features/orgs/routes_test.py:3015`)

<details>
<summary>Executed on the <code>dev</code>-merged tree, and the mutation table</summary>

Suites: `backend/api/features/orgs/` 253 passed / 12 xfailed · `autogpt_libs/auth/` 233 passed · `backend/util/architecture_test.py` 3 passed · `backend/blocks/test/test_block.py` 1647 passed / 84 skipped.

| # | Mutation | Killed by |
|---|---|---|
| M1 | drop the `ctx.org_id != org_id` guard | `test_caller_from_other_org_is_rejected_before_any_lookup` |
| M2 | drop `expected_org_id` on the target-team lookup | `test_target_team_in_other_org_is_rejected` |
| M3 | `is_team_admin` stops requiring `isAdmin` | `test_active_non_admin_member_cannot_update_team` |
| M4 | `is_team_admin` stops requiring `status == "ACTIVE"` | `test_non_active_team_admin_cannot_update_team` (both params) |
| M5 | helper blanket-allows | 11 of 17 in the class |
| M6 | org-admin branch never fires | `test_org_admin_updates_team_they_do_not_belong_to` |
| M7 | `remove_member` loses the guard | `test_admin_of_different_team_cannot_remove_member` |
| M8 | `add_member` loses the guard | `test_plain_org_member_cannot_add_member` |
| M9 | admin lookup keyed on `ctx.team_id` | **SURVIVED** (253 passed) |
| M10 | team lookup keyed on `ctx.team_id` | **SURVIVED** (17 passed) |

Also confirmed: `OrgMemberStatus` is a `StrEnum`, so `status == "ACTIVE"` is a real comparison; and the `Security()` swap emits byte-identical OpenAPI security metadata, so `openapi.json` needed no regeneration.

</details>

autogpt_platform/backend/backend/api/features/orgs/routes_test.py
Pwuts: 🤖 Applied in 6c9c05f4e3, and confirmed load-bearing. With `_authorize_team_management` regressed to look the caller up by `ctx.team_id`, this test now fails:

```
E  Actual: find_unique(where={'teamId_userId': {'teamId': None, 'userId': 'user-owner-1'}})
```

`teamId: None` being exactly the absent active team the route is supposed to stop consulting. Restored source: 251 passed, 12 xfailed in `backend/api/features/orgs`.

autogpt_platform/backend/backend/api/features/orgs/routes_test.py
Pwuts: 🤖 Applied in 6c9c05f4e3. The same mutation fails this one from the denial side:

```
E  Actual: find_unique(where={'teamId_userId': {'teamId': 'ws-other', 'userId': 'user-member-2'}})
```

`ws-other` is the caller's active team, not the target — so the 403 is now pinned to "no admin row in the *target* team". These two are the only tests the regression breaks, which was the gap.

autogpt_platform/backend/backend/api/features/orgs/team_db.py
Pwuts: 🤖 Renamed to `is_admin_of_team` in 6c9c05f4e3. The helper is new in this PR and had exactly one call site, so it came to two lines.

Also retitled the two comments in `routes_test.py` that named the old symbol (`# short-circuits before is_team_admin`, `# --- is_team_admin denial branches ---`) — after the rename they read as being about `RequestContext.is_team_admin`, which is the confusion this thread is about.

autogpt_platform/backend/backend/api/features/orgs/team_routes.py
Pwuts: 🤖 Aligned in 6c9c05f4e3. The docstring and the 403 detail now say workspace: `"Must be a workspace admin or org admin to manage this workspace"`. Nothing pinned the old string — no test, snapshot, OpenAPI entry or frontend reference to it.

Left `the caller's active team (X-Team-Id)` as-is, since that phrase names the header rather than the entity.

Pwuts:
🤖 Pushed 6c9c05f4e3 to this branch — the four review threads above, nothing else. Flagging it here so a commit from me isn't a surprise.

The substance is the first two. `test_team_admin_updates_settings_without_active_team_context` and `test_admin_of_different_team_cannot_update_target` are the tests that prove management authorizes against the URL's `ws_id`, and neither could fail on it: `teammember.find_unique` is stubbed key-agnostically, so regressing `_authorize_team_management` to `ctx.team_id` — the pre-SECRT-2453 behaviour — left the whole suite green. Both now assert the composite key, and that regression breaks exactly those two and nothing else.

The other two are cosmetic and cheap, so I did them rather than leave them: `team_db.is_team_admin` → `is_admin_of_team`, because it was one call site and the old name reads like `RequestContext.is_team_admin` while meaning the target team rather than the caller's active one; and `_authorize_team_management`'s docstring and 403 detail now say "workspace", matching the route summaries, OpenAPI tags and DB errors. Nothing pinned the old error string.

<details>
<summary>What I executed</summary>

```
backend/api/features/orgs          251 passed, 12 xfailed   (263 collected, all accounted for)
backend/util/architecture_test.py    3 passed

mutation: _authorize_team_management -> team_db.is_admin_of_team(ctx.team_id, ctx.user_id)
  FAILED TestTeamManagementByTeamId::test_team_admin_updates_settings_without_active_team_context
    Actual: find_unique(where={'teamId_userId': {'teamId': None, 'userId': 'user-owner-1'}})
  FAILED TestTeamManagementByTeamId::test_admin_of_different_team_cannot_update_target
    Actual: find_unique(where={'teamId_userId': {'teamId': 'ws-other', 'userId': 'user-member-2'}})
  2 failed, 249 passed, 12 xfailed   (no other test noticed the regression)
```

`backend/blocks/test/test_block.py` I could not run locally: my Postgres is at `dev`, where `20260811120000_replace_email_notification_system` drops `User.notifyOnAgentRun`, while this branch is 32 migrations behind `dev` and still declares that column — so all 1089 tests error at the DB boundary regardless of this diff. CI builds its database from the branch's own migrations, so it covers that suite. Reasoned about rather than executed: no block imports `orgs/team_db`, and the renamed helper has exactly one call site.
</details>

autogpt_platform/backend/backend/api/features/orgs/regression_test.py
```diff
@@ -3382,30 +3382,41 @@ async def test_leave_team_rejects_mismatched_org_id(self):
...
-            assert exc_info.value.status_code == 403
+            result = await update_team(
+                org_id="org-review-1",
+                ws_id="team-2",  # different from ctx.team_id
+                request=UpdateTeamRequest(name="Renamed"),
+                ctx=ctx,
+            )
+
+        assert result is updated
+        mock_update.assert_awaited_once()
```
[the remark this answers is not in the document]
Pwuts: 🤖 Confirmed, and fixed in 3b947ae67a. `check_org_permission` does grant everything on this context — measured directly:

```
>>> ctx = MagicMock(); ctx.user_id="u"; ctx.org_id="org-review-1"; ctx.team_id="team-1"
>>> check_org_permission(ctx, OrgAction.MANAGE_WORKSPACES)
True     # ctx.is_org_owner -> <MagicMock ...>, truthy
```

The consequence is sharper than "ineffective": regressing `_authorize_team_management` to look the caller up by `ctx.team_id` — the exact behaviour this test is named after — left it green. It short-circuits on a permission the mock invented.

`_owner_ctx` now returns a real `RequestContext` with `is_org_owner=True`. Flipping that to `False` fails the test, which is a mutation the MagicMock could not express at all. The other two callers only read `ctx.org_id`, so they are unaffected: `backend/api/features/orgs` is 251 passed, 12 xfailed before and after.

Pwuts:
🤖 One more commit, 3b947ae67a — Sentry flagged `_owner_ctx` in `regression_test.py` while I was here, and it was right, so I fixed it rather than hand it back.

It returned a bare `MagicMock`, and `check_org_permission` reads `ctx.is_org_owner` / `ctx.is_org_admin`, which auto-create as truthy attributes — so the context was granted every permission it was asked for. `test_update_team_authorizes_by_target_not_active_team` was passing on that artifact: the same `ctx.team_id` regression that this PR exists to prevent left it green. It now builds a real `RequestContext` with `is_org_owner=True`, and revoking that role fails the test.

Both pushes together are test-only apart from the rename and the two strings. Suite is 251 passed, 12 xfailed throughout.

autogpt_platform/backend/backend/api/features/orgs/routes_test.py
```diff
@@ -2924,6 +2924,414 @@ async def test_lock_loser_waits_for_winner(self, mocker):
...
+            f"/api/orgs/{ORG_ID}/workspaces/{WS_ID}", json={"name": "Renamed"}
+        )
+
+        assert resp.status_code == 200
+        self.prisma.team.update.assert_awaited_once()
+        # The admin lookup must target the URL's ws_id, not the caller's
+        # (absent) active team.
+        self.prisma.teammember.find_unique.assert_awaited_with(
+            where={"teamId_userId": {"teamId": WS_ID, "userId": USER_ID}}
+        )
```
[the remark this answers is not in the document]
Pwuts: 🤖 Not adding these, though the observation is accurate — those tests do stay green under the regression; I measured it.

The reason is where the key is chosen. All four mutating routes call `_authorize_team_management(ctx, org_id, ws_id)` with the path's `ws_id` verbatim (`team_routes.py` lines 108, 208, 231, 252), and the lookup happens once, inside that helper. The regression you describe — passing `ctx.team_id` instead of the URL `ws_id` — is a change to that one function, and it is now pinned from both the allow side and the deny side by the two settings-route tests. Making it fails those two and nothing else, which is what I verified before and after the fix.

Repeating the assertion in the member update/add/remove tests would assert the same fact about the same helper three more times. What it would additionally cover is narrow: someone editing one of the four call sites to pass something other than the path `ws_id`. Real, but not worth triplicated assertions to catch, and duplicated near-identical tests are something we actively push back on in review.

Resolving on that basis — happy to be overruled if you'd rather have the call sites pinned individually.

Pwuts (approved):
🤖 Approving on Reinier's instruction: the four findings from our review are applied on this branch (both test-pinning suggestions verified load-bearing, the rename and the wording done), plus the Sentry finding on the mock context; green on 3b947ae67a with no open threads.

PR #13527, opened 2026-07-11
fix(backend): gate org credits routes behind MANAGE_BILLING

ntindle:
### Why / What / How

**Why:** Any member of a shared org could read the org's pooled balance, its complete transaction history, its invoices and its refund list, because the credits routes were gated only by general auth. They resolve their credit model through `ctx.org_id`, so for a real (non-personal) org they read the shared `OrgBalance`. `OrgAction.MANAGE_BILLING` (owner/billing_manager) and the billing-manager flag on `RequestContext` already existed; no route enforced them.

**What:** The eight credits routes that reach the pooled balance now require org `MANAGE_BILLING`. Org admins are excluded along with plain members — `MANAGE_BILLING` is `{owner, billing_manager}` by design. Personal orgs are unaffected, since a personal membership row is always `isOwner=True`.

Gated: `GET/POST/PATCH /credits`, `POST /credits/{key}/refund`, `POST /credits/auto-top-up`, `GET /credits/transactions`, `GET /credits/refunds`, `GET /credits/invoices`.

Deliberately not gated, because they carry no org data: `GET /credits/auto-top-up` reads the caller's own `User.top_up_config`, and `GET /credits/manage` mints a Stripe billing-portal session for the caller's own customer id — `OrgCreditModel` does not override `create_billing_portal_session`, so gating it would deny a plain member their own payment methods while protecting nothing. Both stay on `get_request_context`. The two `/credits/subscription` routes are user-level, and the Stripe webhook is unauthenticated by design and verified by signature.

**How:** Gated routes take a `BillingManagerContext` alias in place of `Security(get_request_context)`, so the 403 is raised during dependency resolution before any billing data is touched. HTTP layer only — `data/credit.py`, `data/org_credit.py` and the executor's spend path are untouched, so members can still run agents and still get the existing 402 when the pooled balance is empty.

Three of the gated routes (`request_top_up`, `refund_top_up`, `fulfill_checkout`) cannot succeed for a pooled org at all: `OrgCreditModel` raises `NotImplementedError` for Stripe top-up. Gating them is still right, and the suite asserts that unimplemented state rather than letting a mocked 200 read as a working feature.

Linear: SECRT-2449

### Changes 🏗️

- `backend/api/features/v1.py`: eight credits routes gated via a `BillingManagerContext` alias; `list_invoices` docstring now describes org-scoped access.
- `backend/api/features/v1_credits_authz_test.py` (new): a route × role matrix over all eight gated routes, asserting the credit model is never awaited on a 403; two tests driving the *real* `get_request_context` (personal-org owner passes, plain member of a shared org is rejected); two tests that a plain member still reaches the two user-scoped routes; three that `OrgCreditModel` really raises for the top-up paths; and two introspection invariants — every `/credits` route is gated or an explicit exemption, and every route in `v1_router` calling `get_credit_model` is gated or an explicit exemption (which is what covers `execute_graph`). The second walks only the router this suite mounts and matches the literal call, so it does not reach an indirect caller in another router — see the known gap below.
- `frontend/src/app/api/openapi.json`: regenerated for the `list_invoices` docstring.

**Known gap, not closed here:** `GET /home` reads the same pooled balance behind plain `get_request_context` — `home/routes.py:23` passes `ctx.org_id` into `build_home_dashboard`, which reaches `get_credit_model` at `home/service.py:290`, and dev's own `home/service_test.py:384` asserts that call carries the org id. It is pre-existing, out of scope here, and inert under the current frontend for the same reason this gate is, but it is the same disclosure class and needs its own follow-up.

### Agents and large language models used

Claude Code with Claude Opus 5

#### For code changes:
- [x] I have made a test plan
  - [x] `backend/api/features/v1_credits_authz_test.py` — 41 passed
  - [x] `backend/api/features/orgs/regression_test.py` — 95 passed, 12 xfailed (it reads route bodies by symbol, so a moved handler breaks it)
  - [x] `backend/util/architecture_test.py` — 3 passed
  - [x] `backend/blocks/test/test_block.py` — 1734 passed, 84 skipped
  - [x] `ruff check` on both changed files — clean (the lint job runs ruff with `--exit-zero`, so this was run by hand)
  - [x] Mutation-tested the four new assertions: re-gating `manage_payment_method` → 3 failures; ungating `get_user_credits` → 5; dropping `execute_graph` from the exemption list → 1; making `OrgCreditModel.top_up_intent` return instead of raise → 1.

**Verified:** I executed all of the above on the merged tree. The gate's runtime behaviour is proven through FastAPI's real dependency resolution against a mocked data layer, not against a live Stripe or a real org — the role flags are fed through the genuine `get_request_context` in two of the tests and overridden in the matrix. I did not exercise the frontend, which does not forward `X-Org-Id` today and so cannot reach the gate at all.


autogpt_platform/backend/backend/api/features/v1_credits_authz_test.py
```diff
@@ -0,0 +1,155 @@
...
+
+    assert resp.status_code == expected
+    if expected == 200:
+        assert resp.json() == []
+    else:
+        patched.assert_not_awaited()
+
+
+@pytest.mark.parametrize("role", list(ROLE_CASES))
+def test_request_top_up_requires_manage_billing(role, test_user_id, credit_model):
```
kcze: 🤖 🟡 **Nice to Have**: The MANAGE_BILLING gate is a per-route decorator repeated independently (no shared enforcement), yet only 4 of the 10 gated routes are covered here. `refund_top_up`, `fulfill_checkout`, `configure_user_auto_top_up`, `get_user_auto_top_up`, `manage_payment_method`, and `get_refund_requests` have no authz test. Example gap: a future edit reverting the refund or payment-method route to plain `get_request_context` would re-expose the org-pooled balance to plain members and this suite would still pass. Consider parametrizing over the route so all gated endpoints are asserted. _(flagged by: Claude)_

ntindle: 🤖 Fixed in [1e3e95eae](https://github.com/Significant-Gravitas/AutoGPT/commit/1e3e95eae5075a9e28f96bc6d89467187f56a6b9): you were right, and sorry for the slow reply.

The suite is now parametrized over the full (route × role) matrix, so all 10 routes carrying `Security(requires_org_permission(OrgAction.MANAGE_BILLING))` are asserted, not 4. Newly covered: `refund_top_up`, `fulfill_checkout`, `configure_user_auto_top_up`, `get_user_auto_top_up`, `manage_payment_method`, `get_refund_requests`.

The route table lives in `GATED_ROUTES` and each entry also names the first data-layer call its body makes, so the 403 cases assert that call was never awaited — i.e. the gate rejects during dependency resolution, before the org-pooled balance is touched. (`get_user_auto_top_up` reads `get_auto_top_up` rather than the credit model, so it declares that instead.)

Verified your exact scenario by mutation testing, not just by the suite going green — reverting `refund_top_up` to plain `Security(get_request_context)` now fails:

```
FAILED ...::test_credits_route_requires_manage_billing[org_admin-refund_top_up]
FAILED ...::test_credits_route_requires_manage_billing[plain_member-refund_top_up]
```

Also added the missing `org_admin` → 403 case in the same commit, since admins are deliberately excluded from `MANAGE_BILLING`; widening the grant to admins now fails 10 cases.

autogpt_platform/backend/backend/api/features/v1_credits_authz_test.py
```diff
@@ -0,0 +1,155 @@
...
+    assert resp.status_code == expected
+    if expected == 200:
+        assert resp.json() == {"credits": 1000}
+    else:
+        # The gate rejects during dependency resolution, before the route body
+        # ever resolves the org credit model.
+        patched.assert_not_awaited()
+
+
+@pytest.mark.parametrize("role", list(ROLE_CASES))
```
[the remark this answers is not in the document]
ntindle: 🤖 Fixed in [1e3e95eae](https://github.com/Significant-Gravitas/AutoGPT/commit/1e3e95eae5075a9e28f96bc6d89467187f56a6b9): agreed — parametrized over the full (route × role) matrix.

`GATED_ROUTES` now lists all 10 endpoints carrying `Security(requires_org_permission(OrgAction.MANAGE_BILLING))`, crossed with `ROLE_CASES` (`org_owner`/`billing_manager` → 200, `org_admin`/`plain_member` → 403) = 40 cases. Newly covered: `refund_top_up`, `manage_payment_method`, `configure_user_auto_top_up`, `get_user_auto_top_up`, `fulfill_checkout`, `get_refund_requests`.

Each route entry also declares the first data-layer call its body makes, so the 403 branch asserts that call was never awaited — proving rejection happens during dependency resolution rather than after the balance is read. `get_user_auto_top_up` declares `get_auto_top_up` since it never touches the credit model.

Mutation-checked rather than assumed: reverting `refund_top_up` to plain `Security(get_request_context)` fails 2 cases (`org_admin`, `plain_member`), and adding `"admin"` to the `MANAGE_BILLING` grant in `permissions.py` fails 10.

autogpt_platform/backend/backend/api/features/v1_credits_authz_test.py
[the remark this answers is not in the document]
ntindle: 🤖 Fixed in [1e3e95eae](https://github.com/Significant-Gravitas/AutoGPT/commit/1e3e95eae5075a9e28f96bc6d89467187f56a6b9): correct catch on both the substance and the fact that @kcze's thread went unanswered — replied there too.

All 10 `MANAGE_BILLING`-gated routes are now asserted via a `GATED_ROUTES` × `ROLE_CASES` parametrization (40 cases). The six that had no coverage — `refund_top_up`, `fulfill_checkout`, `configure_user_auto_top_up`, `get_user_auto_top_up`, `manage_payment_method`, `get_refund_requests` — are included.

The specific regression described (a route reverting to `get_request_context`) is now caught; verified by actually doing it to `refund_top_up` and watching the `org_admin` and `plain_member` cases fail.

autogpt_platform/backend/backend/api/features/v1_credits_authz_test.py
```diff
@@ -0,0 +1,155 @@
...
+# so it maps to the same passing outcome as a team-org owner.
+ROLE_CASES: dict[str, tuple[dict, int]] = {
+    "org_owner": (dict(owner=True, admin=True), 200),
+    "billing_manager": (dict(billing=True), 200),
+    "plain_member": (dict(), 403),
+    "personal_org_owner": (dict(owner=True, admin=True, org_id="personal-org"), 200),
+}
+
+
+@pytest.fixture(autouse=True)
```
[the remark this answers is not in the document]
ntindle: 🤖 Fixed in [1e3e95eae](https://github.com/Significant-Gravitas/AutoGPT/commit/1e3e95eae5075a9e28f96bc6d89467187f56a6b9): valid and the highest-value item in this review — added `"org_admin": (dict(admin=True), 403)` to `ROLE_CASES`.

Confirmed the reasoning against the source: `_ORG_PERMISSIONS[OrgAction.MANAGE_BILLING] == {"owner", "billing_manager"}` (permissions.py:47), and `_get_org_roles` gives an admin-only context `{"admin", "member"}`, so the intersection is empty → 403. An org admin can rename the org and manage members but not touch billing, and nothing asserted that before.

Verified it bites: adding `"admin"` to the `MANAGE_BILLING` grant now fails 10 cases (every route × `org_admin`), where previously the suite stayed green.

autogpt_platform/backend/backend/api/features/v1_credits_authz_test.py
```diff
@@ -0,0 +1,155 @@
...
+def _auth(mock_jwt_user):
```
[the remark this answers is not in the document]
ntindle: 🤖 Fixed in [1e3e95eae](https://github.com/Significant-Gravitas/AutoGPT/commit/1e3e95eae5075a9e28f96bc6d89467187f56a6b9): you're right that the case was empty — `personal_org_owner` had identical flags to `org_owner` and, because `get_request_context` was overridden, the `org_id` string never reached anything. Took your second suggestion rather than dropping it.

The role row is gone. In its place are two tests that do **not** override `get_request_context` and instead drive the real resolution against a stubbed prisma client:

- `test_personal_org_owner_passes_real_context_resolution` — sends no `X-Org-Id` header, so the real dependency takes the personal-org fallback (`orgmember.find_first(isOwner=True, Org.isPersonal=True)`), builds the context from the returned membership row, and the gate passes. It asserts the fallback was actually awaited and that the resolved org id is what `get_credit_model` was called with — so the claim in this module's docstring ("personal-org owners always carry `is_org_owner`, so the gate is a no-op for them") is now exercised end to end rather than asserted by construction.
- `test_plain_member_of_shared_org_rejected_real_context_resolution` — the same real path with an explicit `X-Org-Id` and a plain-member row → 403, with `get_credit_model` never awaited.

Stubbing prisma rather than seeding a personal org keeps this a unit test (this module has no DB fixture), while still covering the header-absent → personal-org → role-flags → gate path that the overridden version skipped entirely.

autogpt_platform/backend/backend/api/features/v1_credits_authz_test.py
```diff
@@ -0,0 +1,155 @@
...
+    app.dependency_overrides[get_jwt_payload] = mock_jwt_user["get_jwt_payload"]
+    yield
```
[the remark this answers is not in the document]
ntindle: 🤖 Fixed in [1e3e95eae](https://github.com/Significant-Gravitas/AutoGPT/commit/1e3e95eae5075a9e28f96bc6d89467187f56a6b9): you're right that the case was empty — `personal_org_owner` had identical flags to `org_owner` and, because `get_request_context` was overridden, the `org_id` string never reached anything. Took your second suggestion rather than dropping it.

The role row is gone. In its place are two tests that do **not** override `get_request_context` and instead drive the real resolution against a stubbed prisma client:

- `test_personal_org_owner_passes_real_context_resolution` — sends no `X-Org-Id` header, so the real dependency takes the personal-org fallback (`orgmember.find_first(isOwner=True, Org.isPersonal=True)`), builds the context from the returned membership row, and the gate passes. It asserts the fallback was actually awaited and that the resolved org id is what `get_credit_model` was called with — so the claim in this module's docstring ("personal-org owners always carry `is_org_owner`, so the gate is a no-op for them") is now exercised end to end rather than asserted by construction.
- `test_plain_member_of_shared_org_rejected_real_context_resolution` — the same real path with an explicit `X-Org-Id` and a plain-member row → 403, with `get_credit_model` never awaited.

Stubbing prisma rather than seeding a personal org keeps this a unit test (this module has no DB fixture), while still covering the header-absent → personal-org → role-flags → gate path that the overridden version skipped entirely.

autogpt_platform/backend/backend/api/features/v1.py
```diff
@@ -646,7 +652,10 @@ async def upload_file(
 )
 async def get_user_credits(
     user_id: Annotated[str, Security(get_user_id)],
-    ctx: Annotated[RequestContext, Security(get_request_context)],
+    ctx: Annotated[
+        RequestContext,
+        Security(requires_org_permission(OrgAction.MANAGE_BILLING)),
```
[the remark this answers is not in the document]
ntindle: 🤖 Declining this one, on your own caveat — leaving the inline form and resolving.

I checked the convention before deciding. The inline `Annotated[RequestContext, Security(requires_org_permission(...))]` block is what every org-gated route in the codebase already uses — `orgs/routes.py` (8 routes), `orgs/invitation_routes.py` (3), `orgs/team_routes.py` (2 org + 3 team), `transfers/routes.py` (5). Extracting an alias in `v1.py` alone would leave two idioms for the same thing; aligning all ~21 call sites is a repo-wide refactor that doesn't belong in a security fix targeted at SECRT-2449.

Beyond consistency, there's a reason I'd keep it inline even in a green-field version: the alias name is what you'd read at the call site, and `BillingContext` doesn't say `MANAGE_BILLING`. For a gate whose entire purpose is auditability — the bug this PR fixes was a missing gate — having the action visible on every route means a reviewer can grep `requires_org_permission` and see the action next to each handler without a jump-to-definition. The 3-line repetition is cheap; a wrong-permission bug hiding behind an alias is not.

Happy to revisit as a separate repo-wide cleanup if you'd like the alias everywhere.

autogpt_platform/backend/backend/api/features/v1.py
```diff
@@ -646,7 +652,10 @@ async def upload_file(
 )
 async def get_user_credits(
```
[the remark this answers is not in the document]
ntindle: 🤖 Traced the actual frontend consumers on this branch rather than guessing. Summary up front: **a 403 on `GET /credits` degrades silently — it does not crash, spin, retry-storm, or log the user out** — but there is a real defect on a sibling route gated by the same commit. Details and options below; **not resolving this, it's a product/release call.**

### (a) What the code does today

`GET /credits` now carries `Security(requires_org_permission(OrgAction.MANAGE_BILLING))` (v1.py:653-658). `_ORG_PERMISSIONS[MANAGE_BILLING] == {"owner", "billing_manager"}` (permissions.py:47), so in a multi-member org a plain member — **and an org admin** — gets `403 {"detail": "Missing org permission: MANAGE_BILLING"}`. Personal-org users are unaffected: their membership row is always `isOwner=True`. This commit is backend-only; there is no frontend slice on this branch, and `grep -r "MANAGE_BILLING\|is_billing_manager" frontend/src --include=*.ts --include=*.tsx` (excluding generated code) returns **zero** hits — nothing in the UI is permission-aware.

### (b) What a 403 actually does, per consumer

**Navbar Wallet — degrades silently, does not break.** `Wallet.tsx` → `useCredits()` → `AutoGPTServerAPI.getUserCredit()`, which try/catches, reports to Sentry, and returns `{credits: null}` (`src/lib/autogpt-server-api/client.ts:141-153` — the comment there says it returns `null` rather than `0` precisely so callers can tell a failed fetch from a real zero balance). `Wallet.tsx:265` then does `if (credits === null || !state) return null`, so **the wallet button simply disappears from the header**. No error state, no tooltip, no crash, no infinite spinner. This path is hand-rolled `useState`/`useEffect`, not react-query, so there is **no retry storm** — one request per mount, plus one per popover-open and per onboarding websocket notification, each re-capturing to Sentry.

**No logout or redirect.** Neither the orval mutator (`src/app/api/mutators/custom-mutator.ts:129-155`, which just throws `ApiError`) nor the legacy client (`client.ts:905-918`) redirects on 401/403. The legacy client's auth-error branch only fires when a logout is already in progress, and only to suppress noise. There is also no global `QueryCache.onError` (`src/lib/react-query/queryClient.ts` sets `staleTime` only), so **nothing toasts**.

**PaywallGate — unaffected.** It reads `GET /credits/subscription` via `useGetSubscriptionStatus` (`PaywallGate.tsx:31-38`), which this PR deliberately left ungated, and it's additionally `enabled` only when the payment flag is on. It also fails open: on error `subscription` is `undefined`, so `shouldGate` is false and `{children}` renders normally.

**Copilot ChatInput — doesn't read `/credits` at all.** Gating there comes from `useIsUsageLimitReached` (copilot usage) and `useGetSubscriptionStatus`. The only credit-driven copilot surface is `LowCreditBanner` via `TopUpPromptProvider`, whose `isOutOfCredits` requires `credits !== null` — so on 403 the banner and the daily auto-open modal stay hidden. Fails closed by design.

**Billing settings BalanceCard — degrades well.** It uses the generated `useGetV1GetUserCredits` (react-query, so 3 retries then error) and renders `<ErrorCard context="balance" onRetry={refetch} />`, which has a 403-specific message.

**⚠️ The one real defect, on a sibling route:** `GET /credits/auto-top-up` (`get_user_auto_top_up`, v1.py:788-800) is gated by this same commit, but `client.ts:166-168` is a bare `return this._get("/credits/auto-top-up")` with **no try/catch** (unlike `getUserCredit`), and `useCredits.ts:52-60` floats that promise inside a `useEffect`. `TopUpPromptProvider` mounts on every platform page, so for a plain member this is an **unhandled promise rejection on every page load**, captured by Sentry. Same shape at `profile/(user)/credits/page.tsx:79-83` and `WalletRefill.tsx:44`.

### (c) Options

1. **Merge as-is.** Closes the hole now. Cost: plain members of multi-member orgs silently lose the navbar wallet, plus Sentry noise, plus the auto-top-up unhandled rejection.
2. **Merge as-is + a ~5-line frontend patch in this PR**: give `getAutoTopUpConfig()` the same try/catch `getUserCredit()` already has. Kills the unhandled rejection and keeps the wallet's existing graceful-disappear behaviour. Worth doing regardless of what's decided below.
3. **Feature-flag the enforcement** to dark-launch ahead of the frontend slice. Puts a toggle on a security path that then has to be cleaned up, and the failure mode it's protecting against is "wallet is missing", not "app is broken".
4. **Split read from write** — see the sibling VIEW_ORG thread. If the balance read moves to `VIEW_ORG`, most of this frontend impact disappears on its own.

**Recommendation:** option 2, plus option 4 if the answer on the sibling thread is "members may see the pooled balance". I don't think this needs to be sequenced behind SECRT-2450 — nothing hard-breaks — but the auto-top-up rejection should be fixed either way.

@ntindle — leaving this thread open for your call on sequencing.

autogpt_platform/backend/backend/api/features/v1.py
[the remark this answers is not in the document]
ntindle: 🤖 Fair challenge. Confirming the mechanics, then the options — **not resolving this, it's a product decision.**

### (a) What the code does today

All 10 credits routes, reads and writes alike, are gated with `MANAGE_BILLING` = `{"owner", "billing_manager"}` (permissions.py:47). `VIEW_ORG` is `{"owner", "admin", "billing_manager", "member"}` — every active member. So under this PR a plain member **and an org admin** cannot read the pooled balance, transaction history, invoices, or the refund list. Admins are excluded because `MANAGE_BILLING` deliberately omits them (there's now a regression test for that).

The routes split cleanly:

| | routes |
|---|---|
| **Read-only** | `GET /credits`, `GET /credits/transactions`, `GET /credits/invoices`, `GET /credits/refunds`, `GET /credits/auto-top-up` |
| **Mutating / financially sensitive** | `POST /credits` (top-up), `PATCH /credits` (fulfill checkout), `POST /credits/{key}/refund`, `POST /credits/auto-top-up`, `GET /credits/manage` |

`GET /credits/manage` is a GET but mints a Stripe billing-portal session, so it belongs with the mutations.

Worth noting the reads are not all equally sensitive: the balance is a single number, whereas transaction history and invoices expose per-execution spend and Stripe invoice/PDF URLs.

### (b) Frontend impact of the current choice

Traced on the sibling thread. Short version: a member's navbar wallet silently disappears (`Wallet.tsx:265` early-returns when `credits === null`), nothing crashes or redirects, and no part of the frontend is permission-aware today.

### (c) Options

1. **Keep `MANAGE_BILLING` everywhere** (this PR as written). Most conservative; fully closes SECRT-2449. Cost: members can't see the pooled balance they are spending against.
2. **Move `GET /credits` (balance only) to `VIEW_ORG`**, keep the other nine on `MANAGE_BILLING`. Members see the number they're spending against — arguably the point of a pooled balance — while spend detail, invoices and every mutation stay restricted. Also removes most of the frontend fallout in the sibling thread.
3. **Move all read-only routes to `VIEW_ORG`.** I'd argue against this: invoices and transaction history leak Stripe URLs and per-member spend, which is a different thing from "how much is in the pot".
4. **Add a dedicated action** (e.g. `VIEW_BILLING` = `{"owner", "admin", "billing_manager"}`) if the intent is that admins should see the balance but plain members shouldn't. Cleanest if the answer is "admins yes, members no", since `MANAGE_BILLING` currently excludes admins.

**Recommendation:** option 2 — balance read on `VIEW_ORG`, everything else on `MANAGE_BILLING` — unless the product intent is that a pooled balance is deliberately opaque to members. That's a call about what orgs are supposed to see, not a code call, so I'm not making it here.

@ntindle — your call; happy to implement any of these in this PR.

autogpt_platform/backend/backend/api/features/v1_credits_authz_test.py
```diff
@@ -0,0 +1,302 @@
...
+must be rejected with 403. Personal-org owners always carry ``is_org_owner``,
+so the gate is a no-op for them.
+
+The gate is applied as an independent per-route dependency (there is no
+shared router-level enforcement), so every gated route is asserted here:
+dropping the dependency from any single route must fail this suite.
+"""
+
+import dataclasses
+from typing import Any, Callable
```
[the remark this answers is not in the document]
ntindle: 🤖 Partially fixed in [dbd57c8f9](https://github.com/Significant-Gravitas/AutoGPT/commit/dbd57c8f9b67e27481de3c6fadf5b7c51f85074c) — one valid, one declined.

**Fixed (UP035):** `Callable` now comes from `collections.abc` instead of `typing`. That import was new in my commit and the `typing` alias has been deprecated since 3.9, so it's a fair catch.

**Declined (C408, `dict(...)` → literal):** keeping the `dict(owner=True, admin=True)` form. It predates this review round (it came in with the original SECRT-2449 commit), it's the shape that reads naturally given these values are splatted straight into a keyword-only helper — `_ctx(user_id, **flags)` where `_ctx` takes `owner`/`admin`/`billing` keywords — and the kwargs form makes a typo'd flag name look like the keyword it's meant to be. Neither rule is enabled in this repo's ruff config, so `poetry run lint` is green either way; converting would add diff noise to lines nothing else in this review touched.

Verified locally: `ruff check . ../autogpt_libs` → "All checks passed", `isort --check` and `black --check` clean, and the authz suite is still 42 passed.

autogpt_platform/backend/backend/api/features/v1.py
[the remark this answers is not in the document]
ntindle: 🤖 Partially addressed in [77b466ab6](https://github.com/Significant-Gravitas/AutoGPT/commit/77b466ab64efc1d4652bda8292adc653f023fa9c): the ordering constraint is now recorded **in code** rather than only in the PR description — a block comment sits directly above the new `BillingManagerContext` alias in `v1.py`, next to the gate itself, spelling out that real-org `X-Org-Id` forwarding must not ship before SECRT-2450's graceful no-permission handling, and why (`GET /credits` feeds the global nav Wallet).

That is deliberately all this PR does about it. The remaining part of your suggestion — coordinate the release with SECRT-2450, or put the enforcement behind a feature flag — is a product/release call, not a code defect, and @ntindle already recorded a preference above (option 2: `try/catch` on `getAutoTopUpConfig`). Adding a flag here would mean shipping a permanently-off authz gate whose removal is its own PR, which is worse than sequencing two already-planned slices.

Leaving this thread **open** on purpose so the human reviewer signs off on the sequencing rather than having a bot close it.

autogpt_platform/backend/backend/api/features/v1.py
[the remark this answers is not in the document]
ntindle: 🤖 No code change — arguing this one, deliberately.

Swapping the balance read to `VIEW_ORG` would make the read gate a no-op: `_ORG_PERMISSIONS[OrgAction.VIEW_ORG] == {"owner", "admin", "billing_manager", "member"}` (`autogpt_libs/auth/permissions.py`), and `_get_org_roles` hands `"member"` to every active member with no special flags. That set is exactly today's behaviour — i.e. exactly the gap SECRT-2449 exists to close. So "keep mutations behind MANAGE_BILLING, put the read on VIEW_ORG" would ship the mutation half of the fix and silently drop the half the ticket was filed for (a plain member reading the pooled org balance).

Worth noting what a plain member does *not* lose: execution paths are untouched (`execute_graph` still charges the pooled balance and returns 402 when it is empty), so members can still run agents — they just cannot enumerate the org's money, transaction history, or invoices.

If product does decide read visibility should be broader than "who can manage billing", the right shape is a *new*, narrower action (e.g. `VIEW_ORG_BALANCE`) with its own role set — not `VIEW_ORG`, which would reinstate the original vulnerability verbatim.

Leaving this thread **open** for the human reviewer, since it is a product decision rather than a defect.

ntindle:
🤖 Review findings addressed in [77b466ab6](https://github.com/Significant-Gravitas/AutoGPT/commit/77b466ab64efc1d4652bda8292adc653f023fa9c) (plus a merge of `origin/dev`, since the branch was BEHIND).

**🔴 Blocker — `dataclasses` trips the architecture guard.** Fixed. `GatedRoute` and `CreditStubs` are now `pydantic.BaseModel`s (`CreditStubs` carries `model_config = ConfigDict(arbitrary_types_allowed=True)` for its `Mock`/`AsyncMock` fields); the `import dataclasses` is gone. Not added to `_DATACLASS_KNOWN_OFFENDERS` — that list is for pre-existing offenders. Verified locally: `pytest backend/api/features/v1_credits_authz_test.py backend/util/architecture_test.py` → **46 passed** (was `1 failed` on `test_backend_uses_pydantic_not_dataclasses`).

**🟠 Should Fix 1 — introspection test.** Added `test_every_credits_route_is_gated_or_explicitly_exempt`, which walks `app.routes`, filters to `/credits*`, and reads the enforced actions back off the `requires_org_permission` closure. Every such route must either carry `MANAGE_BILLING` or appear in `UNGATED_CREDITS_ROUTES` with a written reason (the two user-level `/credits/subscription` routes and the signature-verified `/credits/stripe_webhook`). It also asserts the discovered set equals `GATED_ROUTES`, so the hand-maintained list can no longer drift. Mutation-checked: ungating just `list_invoices` fails the test with `Unexpected: ['list_invoices']`.

**🟠 Should Fix 2 — SECRT-2450 ordering constraint.** Now recorded in code, above the gate, rather than only in the PR description. The release/flag decision itself is left to the humans — see the reply on that thread, which stays open.

**🟡 Nice to Have 1 — 200-path org assertion.** The success branch of the role matrix now asserts `get_credit_model.await_args.args[1] == "test-org"` for the nine org-pooled routes (skipped for `GET /credits/auto-top-up`, whose config read is user-scoped by design and never touches the credit model).

**🟡 Nice to Have 2 — `BillingManagerContext` alias.** Done: the 10× repeated 4-line annotation is now one named alias. Per-route verifiability is not weakened — each route still declares it, and the new introspection test checks the routing table rather than the source text.

**🔵 Nit — stale `list_invoices` docstring.** Rewritten to say it serves the resolved org's invoices and is billing-manager-scoped. `frontend/src/app/api/openapi.json` was regenerated to match (the export-api-schema pre-commit hook produced a byte-identical file, so `check API types` should stay green); the only schema delta is that description string.

Unchanged on purpose: the `/credits/auto-top-up` config is user-scoped while the top-up it triggers hits the org-pooled model — the gate is still correct there because the endpoint moves org money.

Local verification: `poetry run lint --skip-pyright` clean, `pyright` clean on both touched files, all pre-commit hooks (incl. backend + frontend typecheck) passed. The full backend suite needs Postgres, which this sandbox has no Docker for; the previously failing guard was run directly and passes.

⚠️ Head is now at a REQUEST_CHANGES verdict, and the push dismissed nothing that `/reapprove` could restore — this needs a fresh `/review`, and per the review's own "Human Review Needed: YES", a teammate other than the author still has to sign off on the authz boundary.

autogpt_platform/backend/backend/api/features/v1.py
```diff
@@ -787,7 +813,7 @@ async def configure_user_auto_top_up(
 )
 async def get_user_auto_top_up(
     user_id: Annotated[str, Security(get_user_id)],
-    ctx: Annotated[RequestContext, Security(get_request_context)],
+    ctx: BillingManagerContext,
```
kcze, line 816: 🤖 🟠 **Should Fix**: `get_user_auto_top_up` returns `get_auto_top_up(user_id)` — purely user-scoped data, and `ctx` is unused in the body — yet it is gated on org `MANAGE_BILLING`. That contradicts the deliberate exemption of the equally user-level `/credits/subscription` routes.

Once SECRT-2450 ships `X-Org-Id` forwarding, a plain member of a shared org gets a 403 reading their **own** personal auto-top-up setting, which contains no org data. The `POST` sibling does mutate org balance and belongs behind the gate; the `GET` doesn't. _(flagged by: Claude)_

autogpt_platform/backend/backend/api/features/v1.py
```diff
@@ -1549,7 +1575,7 @@ async def stripe_webhook(request: Request):
 )
 async def manage_payment_method(
     user_id: Annotated[str, Security(get_user_id)],
-    ctx: Annotated[RequestContext, Security(get_request_context)],
+    ctx: BillingManagerContext,
```
kcze, line 1578: 🤖 🟠 **Should Fix**: `manage_payment_method` calls `create_billing_portal_session`, which `OrgCreditModel` does **not** override — it falls through to `UserCreditBase` (`credit.py:361-366`) and opens a portal for `get_stripe_customer_id(user_id)`, i.e. the caller's own Stripe customer, never the org's.

So the gate protects nothing here while locking users out of their own cards: with `X-Org-Id` set to an org where the caller is a plain member, `GET /credits/manage` 403s even though the session exposes no org data and manages the personal payment methods the ungated subscription flow already uses. _(flagged by: Claude)_

autogpt_platform/backend/backend/api/features/v1_credits_authz_test.py
```diff
@@ -0,0 +1,383 @@
...
+        name="get_user_credits",
+        method="GET",
+        path="/credits",
+        check_ok=lambda r: r.json() == {"credits": 1000},
+    ),
+    GatedRoute(
+        name="request_top_up",
+        method="POST",
+        path="/credits",
+        body={"credit_amount": 500},
```
kcze, line 118: 🤖 🟠 **Should Fix**: the 200-path assertions for `request_top_up`, `refund_top_up` and `fulfill_checkout` are unreachable in the pooled-org scenario this gate exists for — `OrgCreditModel.top_up_intent` / `top_up_refund` / `fulfill_checkout` all raise `NotImplementedError` (`org_credit.py:424-447`).

So the suite certifies a happy path that 500s in production for exactly the billing managers the gate now admits: a billing manager of a real org clicks Top up with `X-Org-Id` forwarded, passes the gate, then hits `NotImplementedError`. The mocked `credit_model` makes it green either way. Worth an xfail or an explicit assertion that these are org-unsupported. _(flagged by: Claude)_

autogpt_platform/backend/backend/api/features/v1_credits_authz_test.py
```diff
@@ -0,0 +1,383 @@
...
+
+    ``GATED_ROUTES`` above is hand-maintained, so on its own it can only prove
+    that the routes someone remembered to list are gated. This walks the real
+    routing table instead: every ``/credits*`` route must either carry the
+    MANAGE_BILLING dependency or be an explicit, documented exemption.
+    """
+    gated: set[str] = set()
+    ungated: set[str] = set()
+    for route in app.routes:
+        if not isinstance(route, APIRoute) or not route.path.startswith("/credits"):
```
kcze, line 305: 🤖 🟡 **Nice to Have**: the invariant scopes itself to `route.path.startswith("/credits")` on an app with only `v1_router` mounted, so the "every org-pooled route is gated" guarantee stops at that prefix.

`execute_graph` (`POST /graphs/{id}/execute`, v1.py:2006) already reads the org-pooled balance via `get_credit_model(user_id, ctx.org_id)` and is invisible to this check, and a future org-balance endpoint under e.g. `/orgs/{id}/billing` would be too. Keying off call sites of `get_credit_model` rather than the path prefix would generalize it. _(flagged by: Claude)_

autogpt_platform/backend/backend/api/features/v1.py
```diff
@@ -653,6 +659,26 @@ async def upload_file(
...
+# deliberately also an org admin — gets 403 during dependency resolution,
+# before any billing data is touched. Personal-org membership rows are always
+# ``isOwner=True``, so the gate is a no-op for personal orgs.
+#
+# RELEASE-ORDERING CONSTRAINT (SECRT-2450): the frontend currently never
+# forwards ``X-Org-Id``, so every call resolves to the caller's personal org
+# and this gate is inert. Real-org header forwarding MUST NOT ship before
+# SECRT-2450's graceful no-permission handling lands — ``GET /credits`` feeds
+# the global nav Wallet, so plain members of a shared org would otherwise 403
+# on every page load. Enforcement of that ordering lives in SECRT-2450.
```
kcze, line 675: 🤖 🟠 **Should Fix**: this release-ordering constraint undercounts its own blast radius. It names only `GET /credits` (the nav Wallet) as the surface that would 403 for plain members once `X-Org-Id` forwarding lands.

The settings billing page also calls `/credits/invoices`, `/credits/transactions`, `/credits/auto-top-up` and `/credits/manage` — all four now gated. So SECRT-2450's graceful no-permission handling has to cover the whole billing settings page, not just the Wallet; scoping it to `GET /credits` alone would ship a settings page that 403s in four places for plain members of a shared org. Worth naming them here so the follow-up doesn't inherit the narrower reading. _(flagged by: Claude)_

Pwuts: The comment section about SECRT-2449 is unnecessary narration.

Is the section about SECRT-2450 still accurate and relevant?

Pwuts: 🤖 Both points land. The SECRT-2450 section states a real constraint but hangs it on the wrong ticket, and the SECRT-2449 paragraph is mostly narration. The suggestion at the bottom fixes both in one edit — your comment anchors 662–675, so it replaces the whole block.

**SECRT-2450 is the wrong ticket.** It is [Per-team spend reporting (backend aggregation + billing page breakdown)](https://linear.app/autogpt/issue/SECRT-2450) — a `GROUP BY teamId` endpoint plus a team breakdown on the org billing page. It says nothing about no-permission handling, carries zero comments, and its description makes it *depend* on this one: "gated by `TeamAction.VIEW_SPEND` / org `MANAGE_BILLING` (depends on the authz-gap ticket)". So it is downstream of this gate, not the precondition for it, and "Enforcement of that ordering lives in SECRT-2450" points at a ticket that carries no such enforcement. Its two PRs are #13530 (closed) and #13659 (open, per-team spend UI) — neither touches 403 handling. The requirement actually lives in SECRT-2449's own DECIDED note: "Frontend must handle the no-permission state gracefully — plain members get no org balance display (hide the counter / show nothing, not an error toast)."

**The constraint itself is accurate and still live, so it should stay in some form:**

- The frontend never sends `X-Org-Id` — zero hits under `frontend/src` on `origin/dev`; the two on this head are a generated docstring in `credits.ts`, not a request header. The gate is genuinely inert today.
- `GET /credits` does feed the global nav Wallet: `Navbar.tsx:119,142` → `Wallet` → `useWallet` → `useCredits({fetchInitialCredits: true})` → `api.getUserCredit()` → `_get("/credits")` (`client.ts:136`). Identical on dev, and the Navbar renders on every platform page.
- The gap is still open on dev and closed only here: dev's `get_user_credits` still takes a bare `ctx: Annotated[RequestContext, Security(get_request_context)]`, `v1.py` on dev (`93926246f2`) has zero occurrences of `MANAGE_BILLING` or `requires_org_permission`, `v1_router` has no router-level dependency, and no commit since 2026-08-06 adds one.
- `MANAGE_BILLING = {"owner", "billing_manager"}` (`permissions.py:47`) — org admins really are excluded, which is the one genuinely surprising fact in the block.

**One correction to the scope while it is being reworded:** the block names only `GET /credits`, but all ten credits routes are gated (`/credits` GET/POST/PATCH, `/credits/{key}/refund`, `/credits/auto-top-up` GET/POST, `/credits/manage`, `/credits/transactions`, `/credits/refunds`, `/credits/invoices`); only the two `/credits/subscription` routes and the Stripe webhook are exempt. So forwarding the header would break the whole settings/billing page, not just the Wallet. That is also what the still-open thread on these same lines asks for — the suggestion below settles it too.

**On the SECRT-2449 narration:** agreed. "They are therefore restricted to org roles holding `MANAGE_BILLING`" restates `requires_org_permission(OrgAction.MANAGE_BILLING)` on the next line, and "gets 403 during dependency resolution" restates `Security(...)`. Two clauses are worth keeping because a future editor can't derive them: admins are deliberately excluded, and personal-org rows are always `isOwner=True` (so this is a no-op for personal orgs). Both survive in the suggestion.

**Worth deciding separately:** the frontend half of SECRT-2449's DECIDED note has no ticket of its own, so once this PR closes SECRT-2449 the constraint below points at a closed ticket. Either keep SECRT-2449 open until the billing UI lands, or file a follow-up and swap the id in.

```suggestion
# MANAGE_BILLING excludes org admins by design; personal-org membership rows are
# always isOwner=True, so this gate is a no-op for personal orgs.
#
# Release ordering (SECRT-2449): the frontend never sends X-Org-Id, so the gate
# is inert today. Forwarding it must not ship before the billing UI handles 403
# for plain members — all ten credits routes below are gated, so they would lose
# the nav Wallet and the whole settings/billing page.
```

Verified by running `backend/api/features/v1_credits_authz_test.py` on this head: 43 passed. It errors 43/43 against the shared local dev Postgres, but that is environmental — that DB sits at dev's 210 migrations while this branch has 177, and dev dropped `User.notifyOnAgentRun`, which this branch's `schema.prisma` still declares. Against a throwaway `pgvector/pgvector:pg16` built from the branch's own 177 migrations it is green, including `test_plain_member_of_shared_org_rejected_real_context_resolution`, which sends `X-Org-Id: shared-org` and asserts 403 with `get_credit_model` never awaited.

Pwuts:
@ntindle I'm taking over this PR

Pwuts:
🤖 Taking this over at Reinier's request so it can land — I'll be pushing to this branch: merging `dev` in (318 commits behind) and working through the ten open threads. Authorship stays yours.

@ntindle shout if you have local commits here that aren't pushed, and I'll pull them in first.

autogpt_platform/backend/backend/api/features/v1.py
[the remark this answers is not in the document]
Pwuts: 🤖 Settled by the ticket, so resolving this.

SECRT-2449's DECIDED note (2026-07-10) makes both halves a product ruling rather than an open call: "everything is role-gated ... Frontend must handle the no-permission state gracefully — plain members get no org balance display (hide the counter / show nothing, not an error toast)". So the sequencing is the frontend's to absorb, not a reason to flag or delay the gate.

Two things changed since your comment. The unhandled-promise-rejection defect you found on `GET /credits/auto-top-up` is gone, because that route is no longer gated at all — it reads the caller's own `User.top_up_config`, so a plain member keeps it and the bare `_get` in `client.ts` never sees a 403. And the ordering constraint now lives in a shorter comment above the gate naming every surface that would 403.

The frontend work itself has no ticket of its own yet; that is the one open item, and it is called out on Reinier's thread above.

autogpt_platform/backend/backend/api/features/v1.py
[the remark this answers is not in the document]
Pwuts: 🤖 Settled by the ticket, so resolving this.

SECRT-2449's DECIDED note (2026-07-10) answers it directly: "since the role system exists, everything is role-gated — balance, transaction history, invoices, and all management actions require org owner/billing_manager (MANAGE_BILLING)". So option 1 was the intended shape, and ntindle's argument below is the right one — `VIEW_ORG` is every active member, so putting the balance read there would reinstate the exact gap this ticket was filed for.

Two routes did come out of the gate since, but on the narrower ground that they touch no org data at all rather than on read-visibility: `GET /credits/auto-top-up` reads the caller's own `User.top_up_config`, and `GET /credits/manage` mints a portal for the caller's own Stripe customer. The pooled balance, transaction history, invoices and refunds all stay on MANAGE_BILLING.

If product later wants members to see the balance they spend against, the shape is a new narrower action as ntindle says, not `VIEW_ORG`.

autogpt_platform/backend/backend/api/features/v1_credits_authz_test.py
```diff
@@ -0,0 +1,383 @@
...
+    name: str
+    method: str
+    path: str
+    # Predicate on the success response, to prove the route body really ran.
+    check_ok: Callable[[Any], bool]
+    body: dict | None = None
+    # Attribute on ``CreditStubs`` for the first data-layer call the route body
+    # makes. On a 403 it must never have been awaited — proving the gate
+    # rejects during dependency resolution, before any billing data is touched.
+    guard: str = "get_credit_model"
```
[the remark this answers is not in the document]
Pwuts: 🤖 Fixed in 85f946c5a6, by deleting the selector rather than typing it.

`guard` existed only because `get_user_auto_top_up` read `get_auto_top_up` instead of the credit model. That route is ungated now (kcze's thread on `v1.py:816`), so all eight remaining gated routes share one first data-layer call and the assertion is just `credit_stubs.get_credit_model.assert_not_awaited()`. No `getattr`, no `Literal`, and the field and its comment are gone.

autogpt_platform/frontend/src/app/api/openapi.json
```diff
@@ -4760,7 +4760,7 @@
       "get": {
         "tags": ["v1", "credits"],
         "summary": "List Stripe invoices",
-        "description": "Recent Stripe invoices for the current user.\n\nEach item includes ``hosted_invoice_url`` (Stripe-hosted view) and\n``invoice_pdf_url`` (direct PDF download). Returns an empty list when\nthe credit system is disabled or the user has no Stripe customer yet.",
+        "description": "Recent Stripe invoices for the caller's active org, for billing managers.\n\nThe invoices belong to the org the request resolves to (the caller's\npersonal org when no ``X-Org-Id`` is supplied), so this is restricted to\norg roles holding ``MANAGE_BILLING``.\n\nEach item includes ``hosted_invoice_url`` (Stripe-hosted view) and\n``invoice_pdf_url`` (direct PDF download). Returns an empty list when\nthe credit system is disabled or the org has no Stripe customer yet.",
```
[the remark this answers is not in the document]
Pwuts: 🤖 Not changing this — declining for consistency with the rest of the API.

No route in this repo declares a dependency-level 403 in its `responses=` map. There are 23 routes behind `requires_org_permission` across `orgs/routes.py`, `orgs/invitation_routes.py` and `transfers/routes.py`, and none of them declares one; the only two 403 declarations in the whole API (`auth_email/routes.py:66`, `chat/share.py:102`) sit on routes that raise 403 explicitly in their own body, which is a different thing.

The spec is generated, so this would also mean hand-maintaining a `responses=` entry on each of the eight gated routes for a status the dependency already produces. Worth doing repo-wide if we want it, but not as part of an authz fix.

autogpt_platform/backend/backend/api/features/v1.py
Pwuts: 🤖 Fixed in 85f946c5a6 — the GET is ungated, the POST stays gated.

Re-verified on the merged tree rather than carrying the finding forward: `get_auto_top_up` (`backend/data/credit.py:2367`) is still `get_user_by_id(user_id).top_up_config`, with no org lookup, and `ctx` was still unused in the body. It reverts to dev's `Security(get_request_context)`.

`test_user_scoped_credits_route_allows_plain_member[get_user_auto_top_up]` asserts a plain member gets 200 and the real config back, and `get_user_auto_top_up` is now an entry in `UNGATED_CREDITS_ROUTES` with the reason, so the introspection test fails if anyone re-gates it without arguing the point.

autogpt_platform/backend/backend/api/features/v1.py
Pwuts: 🤖 Fixed in 85f946c5a6 — ungated, and `test_user_scoped_credits_route_allows_plain_member[manage_payment_method]` asserts a plain member gets the portal URL.

Re-verified on the merged tree: `OrgCreditModel` (`backend/data/org_credit.py:337-460`) still does not override `create_billing_portal_session`, so it resolves to the `@staticmethod` on `UserCreditBase` (`backend/data/credit.py:369-376`), which calls `get_stripe_customer_id(user_id)` — the caller's own Stripe customer, never the org's. The gate protected nothing and would have denied plain members their own cards.

**Leaving this open for @Pwuts to decide**, because the fix contradicts the ticket's letter: SECRT-2449's description names `/credits/manage` among the routes to gate, on the premise that they "resolve via `ctx.org_id` to the pooled `OrgBalance`". That premise is false for this one route, which is the whole finding — but if the intent was that minting a billing-portal session counts as a "management action" regardless of whose money it reaches, re-gating it is a one-line revert plus dropping the two test entries.

autogpt_platform/backend/backend/api/features/v1_credits_authz_test.py
Pwuts: 🤖 Fixed in 85f946c5a6 — the suite now asserts the unimplemented path instead of implying a working one.

Still true on the merged tree: `top_up_intent`, `top_up_refund` and `fulfill_checkout` all raise `NotImplementedError` in `OrgCreditModel` (`backend/data/org_credit.py:418`, `427`, `438`). So rather than an `xfail` on the 200 rows — which would lose the gate coverage those rows exist for — there is a separate parametrized test driving the real `OrgCreditModel` and asserting each of the three raises:

```
test_gated_top_up_route_is_unimplemented_for_pooled_orgs[request_top_up|refund_top_up|fulfill_checkout]
```

It cross-checks each route name against `GATED_ROUTES`, so it cannot drift from the matrix above. Mutation-tested: making `top_up_intent` return a URL instead of raising turns `[request_top_up]` red (1 failed, 40 passed), so it is not vacuous.

autogpt_platform/backend/backend/api/features/v1_credits_authz_test.py
Pwuts: 🤖 Done in 85f946c5a6 — keyed on the call sites, as you suggested.

`test_every_org_balance_reader_is_gated_or_explicitly_exempt` walks every route in the mounted app, reads `inspect.getsource(route.endpoint)`, and requires any route mentioning `get_credit_model` to be MANAGE_BILLING-gated or listed in `ORG_BALANCE_UNGATED` with a reason. That picks up `execute_graph` exactly as you said it would, and a future org-billing route under any prefix.

`execute_graph` is an explicit exemption, not a gap — the executor spends the pooled balance on the caller's behalf and gating it would stop plain members running agents at all; it returns 402 without revealing a figure. The path-prefix test stays, because it catches a new `/credits` route that reaches the balance through some other helper.

Mutation-tested: dropping `execute_graph` from the exemption list turns the new test red (1 failed, 40 passed), which is what proves the scan really finds it rather than finding nothing.

autogpt_platform/backend/backend/api/features/v1.py
Pwuts: 🤖 Fixed in 85f946c5a6 — the comment now names the surfaces instead of only `GET /credits`.

Two of the four you listed dropped out of the gate entirely on your other two threads: `/credits/auto-top-up` (GET) and `/credits/manage` both serve the caller's own data, so they are ungated now and a plain member keeps them. That leaves the nav Wallet (`GET /credits`) plus `/credits/transactions`, `/credits/refunds` and `/credits/invoices` on the settings billing page, which is what the rewritten block lists.

Pwuts: 🤖 Applied in 85f946c5a6, with one change to the suggested text: two of the ten routes it referred to are no longer gated (the thread below, and kcze's on these same lines, argued `GET /credits/auto-top-up` and `GET /credits/manage` out of the gate), so the block names the surfaces that actually 403 rather than a count.

```
# MANAGE_BILLING excludes org admins by design; personal-org membership rows
# are always isOwner=True, so this gate is a no-op for personal orgs.
#
# Release ordering (SECRT-2449): the frontend never sends X-Org-Id, so the gate
# is inert today. Forwarding it must not ship before the billing UI handles 403
# for plain members — they would otherwise lose the nav Wallet (GET /credits)
# and the settings billing page (/credits/transactions, /credits/refunds,
# /credits/invoices).
```

18 lines down to 8: the SECRT-2449 narration is gone, and the constraint hangs on SECRT-2449's own DECIDED note instead of SECRT-2450, which is per-team spend reporting and depends on this ticket rather than gating it.

Your last point still stands and is not something this PR settles: SECRT-2449's frontend half has no ticket of its own, so once this merges and the ticket closes, that comment points at a closed ticket. Worth filing the billing-UI follow-up and swapping the id in.

Pwuts:
## 🤖 Test evidence

Everything below ran on the merged tree at `85f946c5a6`, backend venv synced to the merged lockfile. The head then moved to `192bcbca19` for a docstring-only follow-up, on which the authz suite was re-run: 41 passed.

### Suites

```
backend/api/features/v1_credits_authz_test.py   41 passed in 51.96s
backend/api/features/orgs/regression_test.py    95 passed, 12 xfailed in 9.21s
backend/util/architecture_test.py                3 passed in 7.10s
backend/blocks/test/test_block.py             1734 passed, 84 skipped in 43.40s
ruff check v1.py v1_credits_authz_test.py       All checks passed!
```

`orgs/regression_test.py` is in the list because it reads route bodies with `inspect.getsource`, so it is the one suite that breaks when a handler moves and no grep for the path or the models would find it. `architecture_test.py` and `test_block.py` scan the whole tree, so no diff points at them.

### Mutation testing

Each new assertion was made to fail, so none of them is vacuous. Every mutation was applied with an asserted single match, run, then restored, with `git status` confirmed clean afterwards.

| mutation | result |
|---|---|
| re-gate `manage_payment_method` | 3 failed, 38 passed |
| ungate `get_user_credits` | 5 failed, 36 passed |
| drop `execute_graph` from `ORG_BALANCE_UNGATED` | 1 failed, 40 passed |
| `OrgCreditModel.top_up_intent` returns instead of raising | 1 failed, 40 passed |

The first one names all three tests that should catch it:

```
FAILED ...::test_user_scoped_credits_route_allows_plain_member[manage_payment_method]
       AssertionError: {"detail":"Missing org permission: MANAGE_BILLING"}
FAILED ...::test_every_credits_route_is_gated_or_explicitly_exempt
       AssertionError: A /credits route is not behind MANAGE_BILLING. Gate it with...
FAILED ...::test_every_org_balance_reader_is_gated_or_explicitly_exempt
       AssertionError: A route resolves the org-pooled credit model without MANAGE...
```

The third matters most for the new call-site invariant: if `inspect.getsource` had found nothing, the scan would return an empty set and the test would be asserting nothing. Removing `execute_graph` from the exemption list turning it red is what proves the scan really reaches it.

### Merge

`dev` was 318 commits ahead. The merge is lossless by construction: the added lines of `git diff origin/dev...HEAD` are byte-identical to those of the pre-merge branch delta.

```
$ diff <(git diff $(git merge-base origin/dev 77b466ab64) 77b466ab64 | grep '^+' | grep -v '^+++' | sort) \
       <(git diff origin/dev...HEAD | grep '^+' | grep -v '^+++' | sort) && echo IDENTICAL
IDENTICAL added lines
```

Every finding that rested on `dev` state was re-read on the merged tree rather than carried forward: `OrgCreditModel.top_up_intent` / `top_up_refund` / `fulfill_checkout` still raise `NotImplementedError` (`org_credit.py:418`, `427`, `438`), `create_billing_portal_session` is still not overridden, and `get_auto_top_up` is still `get_user_by_id(user_id).top_up_config`.

### Not tested

The frontend. It never sends `X-Org-Id` — so nothing in the UI can reach this gate today, which is the release-ordering constraint recorded above the gate in `v1.py`.

autogpt_platform/backend/backend/api/features/v1.py
Pwuts: 🤖 Decided: it stays ungated. Reinier, 23:1xZ on 2026-09-15 — "from the description it sounds like keeping `GET /credits/manage` ungated is the logical answer. Go ahead."

No code change; the route keeps dev's `Security(get_request_context)` as pushed in 85f946c5a6, and `manage_payment_method` stays in `UNGATED_CREDITS_ROUTES` and `ORG_BALANCE_UNGATED` with the reason. Resolving.

PR #13532, opened 2026-07-11
feat(backend): resolve team/org credentials on the read/use path

ntindle:
### Why / What / How

**Why:** `IntegrationCredential` has supported `TEAM`/`ORG` ownership since PR1 and `backend.integrations.scoped_credentials` implemented USER→TEAM→ORG resolution — but the resolver was dead code. The live path (`get_user_credentials`) only ever returned USER rows, so shared team credentials were invisible to the picker and unusable by executions.

**What:** Read-side resolution only. A user now resolves their own USER creds + TEAM creds (teams where they're an ACTIVE member) + ORG creds (active org membership) for credential **listing** and execution **fetch**. Write paths are untouched — personal USER-cred CRUD is byte-for-byte unchanged, and team/org rows can never be mutated through the read path.

**How:**
- `data/user.py`: new `get_accessible_credentials(user_id)` — resolves from the user's own memberships (no request context needed, safe for the executor), USER>TEAM>ORG precedence, corrupt-row skip. `get_user_credentials` deliberately stays USER-only: it's the seam the write helpers read-modify-write, so widening it would let a full-list replace clobber shared rows.
- `db_manager.py`: `get_accessible_credentials` exposed over RPC for the executor.
- `credentials_store.py`: `_get_all_creds_unlocked` now resolves the accessible set; since `get_creds_by_id` filters it, **the executor's fetch-by-id inherits the access check for free** — live membership is re-checked on every fetch, so a run scheduled by someone who later left the team stops resolving that team's creds at execution time (conservative policy, flagged for product review).

**Known gap (deliberate, currently unreachable):** an OAuth2 TEAM/ORG credential would fail at token-refresh write-back (`update_creds` raises on non-USER ids rather than cloning — no corruption possible). No create path for TEAM/ORG credentials exists yet, so this cannot trigger; the MANAGE_CREDENTIALS slice that adds team-cred creation MUST include refresh write-back for shared creds. Tracked on SECRT-2452.

Linear: SECRT-2452 (read half)

### Changes 🏗️

- `backend/data/user.py`: `get_accessible_credentials` + shared decrypt helper; `get_user_credentials` docstring hardened to explain the USER-only invariant
- `backend/data/db_manager.py`: RPC exposure
- `backend/integrations/credentials_store.py`: read surface resolves accessible set; write helpers explicitly pinned to the USER-only reader
- Tests (new): member resolves user+team+org creds; non-member resolves none of another team's; left-team user loses access; id-collision precedence; executor fetch-by-id allowed/denied via access filtering

#### For code changes:
- [x] I have made a test plan
  - [x] `poetry run pytest backend/data/user_test.py backend/integrations/credentials_store_test.py -q` — 21 passed

For changes touching `data/*.py`: `get_accessible_credentials` takes `user_id` and resolves strictly from that user's ACTIVE team/org memberships — access is derived from membership rows, never from caller-supplied org/team ids.

---

> [!NOTE]
> **Medium Risk**
> Changes credential access for listing and executor fetch-by-id based on membership; scope is read-only but affects security-sensitive integration secrets and execution behavior.
> 
> **Overview**
> **Team and org integration credentials now show up in listing and can be used at execution time**, while personal credential CRUD stays USER-scoped only.
> 
> Adds `get_accessible_credentials(user_id)`, which loads USER credentials plus TEAM/ORG rows from **live ACTIVE** team/org memberships (no request context), with USER > TEAM > ORG dedup and shared decrypt logic that skips corrupt rows. `get_user_credentials` is unchanged in scope and remains the write-path seam. `IntegrationCredentialsStore._get_all_creds_unlocked` switches from USER-only reads to the accessible set, so `get_creds_by_id` / listing inherit membership checks (including revoking access after someone leaves a team). Exposed on `DatabaseManager` RPC for executor workers. Tests cover resolution, non-member isolation, departed members, and store wiring.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 88b8dcccbdb64f85ab3cf11d1e4ab476cd37ad75. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


ntindle:
🤖 Batch command failed: Command failed: gh api -X PATCH repos/Significant-Gravitas/AutoGPT/git/refs/heads/batch/rollup -f sha=d153a652d452dd4bc1176444b048a31858249e74 -F force=true
gh: Object does not exist (HTTP 422)

autogpt_platform/backend/backend/integrations/credentials_store.py
```diff
@@ -352,11 +367,18 @@ async def get_all_creds(self, user_id: str) -> list[Credentials]:
...
+        own USER creds together with TEAM/ORG creds they can access, then
+        appends the platform's system-managed credentials. Since a
+        credential id only appears here if the user has access, id-based
+        lookups (``get_creds_by_id``) inherit that access check for free —
+        this is what gates the executor's fetch-by-id for team/org creds.
 
         **Caller must already hold ``locked_user_integrations(user_id)``.**
         """
-        all_credentials = await self._get_persisted_user_creds_unlocked(user_id)
+        all_credentials = await self._get_accessible_creds_unlocked(user_id)
```
kcze, line 381: 🤖 🔴 **Blocker**: Widening the reader here makes OAuth **refresh** of a TEAM/ORG credential destroy the shared token — the PR description's "no corruption possible" doesn't hold. In `creds_manager._refresh_locked` the provider call comes *first*: `fresh = await oauth_handler.refresh_tokens(credentials)` rotates (and invalidates) the old refresh token upstream, and only then does `store.update_creds` → `_get_persisted_user_creds_unlocked` (USER-only) fail to find the id and raise `ValueError`. The new tokens are discarded while the old refresh token is already dead at the provider, so the team credential is bricked for every member with no API path to recover it.

It also fails **open**, not closed: `executor/manager.py` catches that `ValueError` from `creds_manager.acquire` and, when the field has a default, sets `input_data[field_name] = None` and runs the block with no credentials.

(Separate failure mode from the already-reported delete/revoke issue on this line — same root cause, but this one loses data rather than no-op'ing.) _(flagged by: Claude + Codex)_

autogpt_platform/backend/backend/data/user.py
```diff
@@ -364,6 +359,88 @@ async def get_user_credentials(user_id: str) -> list[Credentials]:
...
+    rows.sort(key=lambda r: owner_rank.get(r.ownerType, 3))
+    deduped = []
+    seen_ids: set[str] = set()
+    for row in rows:
+        if row.id in seen_ids:
+            continue
+        seen_ids.add(row.id)
+        deduped.append(row)
+
+    return _decrypt_credential_rows(deduped, user_id)
```
kcze, line 441: 🤖 🟠 **Should Fix**: The returned `Credentials` carry the **payload's** `id`, not the DB row id, so TEAM/ORG rows are unreachable by the id-based lookup this PR is wiring up. `set_user_credentials` writes `"id": cred.id` explicitly, so USER rows happen to agree — but `scoped_credentials.create_credential` (the only TEAM/ORG create path) lets Prisma generate `id` via `@default(uuid())` and stores the caller's payload unchanged. `get_creds_by_id(user_id, <row id handed out at creation>)` then returns `None`. Worse, `_BaseCredentials.id` is `default_factory=lambda: str(uuid4())`, so a payload without an `id` yields a **different** id on every read — no node's stored `credentials_meta.id` can ever match.

The dedup just above compounds it: it keys on `row.id` while every consumer matches on the payload id, so the two are guarding different keys. Stamp the row id onto the validated credential (or make creation write the same id into both). _(flagged by: Codex)_

autogpt_platform/backend/backend/data/user.py
```diff
@@ -364,6 +359,88 @@ async def get_user_credentials(user_id: str) -> list[Credentials]:
...
+        )
+    ]
+
+    owner_clauses: list[dict] = [{"ownerType": "USER", "ownerId": user_id}]
+    if team_ids:
+        owner_clauses.append({"ownerType": "TEAM", "ownerId": {"in": team_ids}})
+    if org_ids:
+        owner_clauses.append({"ownerType": "ORG", "ownerId": {"in": org_ids}})
+
+    rows = await prisma.integrationcredential.find_many(
```
kcze, line 423: 🤖 🟠 **Should Fix**: Resolution is org-context-free — it returns credentials for *every* org/team the user is an ACTIVE member of, ignoring the active `RequestContext.org_id`/`team_id` that scopes every other resource, and that the sibling `integrations/scoped_credentials.py` explicitly enforces (`organizationId` + active `team_id`).

A user who belongs to orgs X and Y, working in org X, sees org Y's TEAM/ORG credentials in the picker and can bind one into an org-X graph — org Y's secret is then exercised, logged and billed inside a tenant whose admins were never granted it. Membership-derived resolution is the right call for the executor (no request context), but the listing surface has context available and should narrow to it. _(flagged by: Claude)_

autogpt_platform/backend/backend/data/user_test.py
```diff
@@ -263,6 +263,160 @@ async def test_set_raises_without_personal_org_for_new_cred(self, mocker):
...
+        mock_prisma.integrationcredential.find_many = AsyncMock(
+            return_value=[self._row(user_cred, "USER")]
+        )
+        mocker.patch("backend.data.user.prisma", mock_prisma)
+
+        result = await get_accessible_credentials("u1")
+
+        assert [c.id for c in result] == ["u"]
+        assert (
+            mock_prisma.teammember.find_many.call_args.kwargs["where"]["status"]
```
kcze, line 379: 🤖 🟠 **Should Fix**: These tests never assert the authorization predicate they exist to protect. The membership `where` is only checked for `["status"] == "ACTIVE"` — nothing asserts `["userId"] == user_id`. Delete `"userId": user_id` from `teammember.find_many`/`orgmember.find_many` and all four tests still pass, even though that change resolves every ACTIVE membership platform-wide and hands every user every team's credentials.

`credentials_store_test.py` mocks `get_accessible_credentials` wholesale, so it covers wiring only — nothing in the suite exercises the predicate itself. Add `assert where["userId"] == "u1"` on both membership queries. _(flagged by: Claude)_

autogpt_platform/backend/backend/data/user.py
```diff
@@ -364,6 +359,88 @@ async def get_user_credentials(user_id: str) -> list[Credentials]:
...
+    has left a team/org stops resolving its credentials on the very next
+    call — including at execution time for a run scheduled while they were
+    still a member. Precedence on id collision is USER > TEAM > ORG; ids
+    are unique across the table in practice, so the dedup is only a
+    defensive tie-break.
+
+    The WRITE path (``set_user_credentials``) stays USER-only, so team/org
+    rows are never mutated through this read path.
+    """
+    team_ids = [
```
kcze, line 404: 🤖 🟡 **Nice to Have**: This triples the query count on the hottest credential path. `get_creds_by_id` → `get_all_creds` → `get_accessible_credentials` now issues 3 queries (`teammember`, `orgmember`, `integrationcredential`) where it previously issued 1, and the executor calls it once per credentialed node per execution with no memoization across nodes. Consider resolving membership once per execution, or short-circuiting the two membership queries for users with no team/org rows. _(flagged by: Claude)_

autogpt_platform/backend/backend/integrations/credentials_store.py
```diff
@@ -320,16 +320,31 @@ def db_manager(self):
...
+    async def _get_accessible_creds_unlocked(self, user_id: str) -> list[Credentials]:
+        """Return every persisted credential the user may USE — their own
+        USER creds plus TEAM/ORG creds resolved from live membership (see
+        ``backend.data.user.get_accessible_credentials``).
+
+        Read-only: this feeds credential LISTING and execution FETCH only.
+        Team/org rows resolved here are never written back — the write
+        helpers use ``_get_persisted_user_creds_unlocked`` (USER-only).
+        """
+        return list(await self.db_manager.get_accessible_credentials(user_id=user_id))
```
kcze, line 346: 🤖 🟠 **Should Fix**: The shared-credential delete problem has a **second** endpoint that the existing Bugbot thread on line 381 doesn't name: the external API. `api/external/v1/integrations.py` `delete_credential` resolves via `get_creds_by_id(auth.user_id, cred_id)` — which this line now widens — so a TEAM/ORG id clears the 404 gate, `creds_manager.delete` → `delete_creds_by_id` filters a USER-only list and writes back an unchanged set (silent no-op), and the route unconditionally returns `DeleteCredentialResponse(deleted=True, credentials_id=cred_id)`.

Any API key holding `DELETE_INTEGRATIONS` gets a false success while the row stays `status="active"`. No provider-side revocation happens here (unlike the main route), so the damage is a lying response rather than a destroyed grant — but both endpoints need the same fix: gate mutation/delete flows on an owner-only lookup rather than the accessible set. _(flagged by: Claude)_

PR #13547, opened 2026-07-13
fix(frontend): cut auth panel vortex animation cost to stop page-wide jank

0ubbe:
### Why / What / How

**Why:** A user reported that animations on the login page are slow — including the input focus border transition on the form side. The cause is the `Vortex` canvas background on the auth marketing panel: every frame it re-drew the full canvas onto itself three times, two of those through `ctx.filter = "blur(8px|4px) brightness(200%)"`. Full-resolution canvas blur composites are extremely expensive, and because the simulation timestep is clamped at 1/30s, any machine that can't hold 30fps sees the animation play in slow motion. The rAF loop also starves the main thread, so every paint-bound CSS transition on the page (e.g. the input focus ring) drops to the same low frame rate.

**What:** Removes the full-canvas blur glow passes, sizes the canvas to its container instead of the whole window, pauses the loop when the panel is not visible, and renders a single static frame when `prefers-reduced-motion` is set.

**How:**
- The glow effect is now approximated per particle with a second, wider translucent stroke drawn under the core stroke with `globalCompositeOperation: "lighter"` — two cheap strokes per particle instead of three full-canvas composites (two of them blurred) per frame.
- Canvas is sized to the container (via `ResizeObserver`) rather than `window.innerWidth/Height`, halving the rasterized area on the split layout.
- An `IntersectionObserver` starts/stops the rAF loop, so the animation doesn't burn CPU/GPU when the panel is hidden (mobile `hidden lg:flex`) or off-screen.
- With `prefers-reduced-motion: reduce`, the simulation is warmed up off-screen and a single static frame is drawn — no animation loop at all.
- Per-particle `save()/restore()` churn removed; shared canvas state is set once per frame.

Visual result is nearly identical (same particle field, slightly tighter glow), but frame cost drops by roughly an order of magnitude.

### Changes 🏗️

- `frontend/src/components/ui/vortex.tsx`:
  - Removed `renderGlow`/`renderToScreen` full-canvas blur+brightness composite passes; glow now drawn per particle (halo stroke + core stroke, additive compositing)
  - Canvas sized to container with `ResizeObserver` instead of window size + resize listener
  - rAF loop gated by `IntersectionObserver` (pauses when panel hidden/off-screen) with proper start/stop and cleanup
  - `prefers-reduced-motion` renders one static frame instead of looping
  - Replaced raw `useEffect` with `useMountEffect`

#### For code changes:
- [x] I have made a test plan
  - [ ] Login/signup pages render the animated particle background as before on desktop
  - [ ] With DevTools CPU throttling (6x), the input focus border transition stays smooth
  - [ ] Resize the window — canvas follows the panel size
  - [ ] Enable "Emulate CSS prefers-reduced-motion: reduce" — a static particle frame renders, no rAF loop in the Performance panel
  - [ ] On a <lg viewport (panel hidden), no rAF loop runs

### DIFF
--- a/autogpt_platform/frontend/src/components/ui/__tests__/vortex.test.tsx
+++ b/autogpt_platform/frontend/src/components/ui/__tests__/vortex.test.tsx
@@ -1,0 +1,156 @@
+import { render } from "@testing-library/react";
+import { afterEach, beforeEach, describe, expect, test, vi } from "vitest";
+import { Vortex } from "../vortex";
+
+type ObserverCallback = (entries: Array<{ isIntersecting: boolean }>) => void;
+
+const rafCallbacks = new Map<number, FrameRequestCallback>();
+let rafId = 0;
+const intersectionCallbacks: ObserverCallback[] = [];
+const resizeCallbacks: Array<() => void> = [];
+const disconnectSpy = vi.fn();
+let mockSize = { width: 800, height: 600 };
+let mockCtx: ReturnType<typeof createMockContext>;
+
+function createMockContext() {
+  return {
+    clearRect: vi.fn(),
+    fillRect: vi.fn(),
+    beginPath: vi.fn(),
+    moveTo: vi.fn(),
+    lineTo: vi.fn(),
+    stroke: vi.fn(),
+    fillStyle: "",
+    strokeStyle: "",
+    lineCap: "",
+    lineWidth: 0,
+    globalCompositeOperation: "",
+  };
+}
+
+function flushFrames(count = 1) {
+  for (let i = 0; i < count; i++) {
+    const pending = Array.from(rafCallbacks.values());
+    rafCallbacks.clear();
+    for (const callback of pending) callback(performance.now());
+  }
+}
+
+function setReducedMotion(matches: boolean) {
+  vi.stubGlobal("matchMedia", vi.fn().mockReturnValue({ matches }));
+}
+
+beforeEach(() => {
+  rafCallbacks.clear();
+  rafId = 0;
+  intersectionCallbacks.length = 0;
+  resizeCallbacks.length = 0;
+  disconnectSpy.mockClear();
+  mockSize = { width: 800, height: 600 };
+  mockCtx = createMockContext();
+
+  vi.stubGlobal("requestAnimationFrame", (cb: FrameRequestCallback) => {
+    rafCallbacks.set(++rafId, cb);
+    return rafId;
+  });
+  vi.stubGlobal("cancelAnimationFrame", (id: number) => {
+    rafCallbacks.delete(id);
+  });
+  vi.stubGlobal(
+    "IntersectionObserver",
+    class {
+      constructor(cb: ObserverCallback) {
+        intersectionCallbacks.push(cb);
+      }
+      observe() {}
+      disconnect = disconnectSpy;
+    },
+  );
+  vi.stubGlobal(
+    "ResizeObserver",
+    class {
+      constructor(cb: () => void) {
+        resizeCallbacks.push(cb);
+      }
+      observe() {
+        for (const cb of resizeCallbacks) cb();
+      }
+      disconnect = disconnectSpy;
+    },
+  );
+  setReducedMotion(false);
+
+  vi.spyOn(HTMLCanvasElement.prototype, "getContext").mockReturnValue(
+    mockCtx as unknown as CanvasRenderingContext2D,
+  );
+  Object.defineProperty(HTMLElement.prototype, "clientWidth", {
+    configurable: true,
+    get: () => mockSize.width,
+  });
+  Object.defineProperty(HTMLElement.prototype, "clientHeight", {
+    configurable: true,
+    get: () => mockSize.height,
+  });
+});
+
+afterEach(() => {
+  vi.unstubAllGlobals();
+  vi.restoreAllMocks();
+});
+
+describe("Vortex", () => {
+  test("animates while visible and stops when hidden", () => {
+    render(<Vortex particleCount={10} />);
+
+    const [onIntersection] = intersectionCallbacks;
+    onIntersection([{ isIntersecting: true }]);
+    flushFrames(2);
+    expect(mockCtx.stroke).toHaveBeenCalled();
+
+    onIntersection([{ isIntersecting: false }]);
+    expect(rafCallbacks.size).toBe(0);
+  });
+
+  test("acts on the last entry when intersection records are batched", () => {
+    render(<Vortex particleCount={10} />);
+
+    intersectionCallbacks[0]([
+      { isIntersecting: true },
+      { isIntersecting: false },
+    ]);
+    expect(rafCallbacks.size).toBe(0);
+    expect(mockCtx.stroke).not.toHaveBeenCalled();
+  });
+
+  test("renders a single static frame under prefers-reduced-motion", () => {
+    setReducedMotion(true);
+    render(<Vortex particleCount={10} />);
+
+    expect(intersectionCallbacks.length).toBe(0);
+    flushFrames(1);
+    expect(mockCtx.stroke).toHaveBeenCalled();
+    expect(rafCallbacks.size).toBe(0);
+  });
+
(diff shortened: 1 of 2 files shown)


autogpt_platform/frontend/src/components/ui/vortex.tsx
```diff
@@ -186,77 +216,67 @@ export function Vortex(props: VortexProps) {
...
-    setup();
-    window.addEventListener("resize", handleResize);
+    const resizeObserver = new ResizeObserver(() => resize(canvas, container));
+    resizeObserver.observe(container);
 
-    return () => {
-      window.removeEventListener("resize", handleResize);
-      if (animationFrameId.current) {
-        cancelAnimationFrame(animationFrameId.current);
+    const intersectionObserver = new IntersectionObserver(([entry]) => {
```
0ubbe: 🤖 🟠 **Should Fix**: `([entry])` takes the *first* entry, but an IntersectionObserver callback can deliver multiple batched records for the same target if visibility toggles quickly — acting on the first (oldest) record can leave the loop in the wrong state (e.g. running while hidden). Use the most recent record instead: `(entries) => { const entry = entries[entries.length - 1]; ... }`.

autogpt_platform/frontend/src/components/ui/vortex.tsx
```diff
@@ -186,77 +216,67 @@ export function Vortex(props: VortexProps) {
...
-  }
+    resize(canvas, container);
+    initParticles();
 
-  function handleResize() {
-    const canvas = canvasRef.current;
-    if (canvas) {
-      resize(canvas);
+    if (window.matchMedia("(prefers-reduced-motion: reduce)").matches) {
+      drawStaticFrame(canvas, ctx);
```
0ubbe: 🤖 🟡 **Nice to Have**: in the `prefers-reduced-motion` branch we return before registering the ResizeObserver, so the static frame keeps its mount-time size — after a window resize the canvas no longer fills the panel (stretched gap or clipped edges). Consider registering a ResizeObserver in this branch too that calls `resize(...)` + `drawStaticFrame(...)` again.

autogpt_platform/frontend/src/components/ui/vortex.tsx
```diff
@@ -186,77 +216,67 @@ export function Vortex(props: VortexProps) {
...
-    canvas: HTMLCanvasElement,
-    ctx: CanvasRenderingContext2D,
-  ) {
-    ctx.save();
-    ctx.globalCompositeOperation = "lighter";
-    ctx.drawImage(canvas, 0, 0);
-    ctx.restore();
-  }
+    resize(canvas, container);
+    initParticles();
```
0ubbe, line 255: 🤖 🟡 **Nice to Have**: on `< lg` viewports the container is `display: none`, so `initParticles()` runs with a 0×0 canvas — every particle gets `x = 0`. If the viewport is later resized to desktop, the ResizeObserver fixes the canvas size but the existing particles stay clustered at the left edge until their TTL expires (up to ~3s). Re-initializing particles on the first resize from a zero-sized canvas would avoid the visual glitch.

autogpt_platform/frontend/src/components/ui/vortex.tsx
0ubbe: 🤖 Fixed in [75f1bb631](https://github.com/Significant-Gravitas/AutoGPT/commit/75f1bb631c4521ea03311c7140a9f76a5f9c7db9): the callback now acts on the most recent record (`entries[entries.length - 1]`).

autogpt_platform/frontend/src/components/ui/vortex.tsx
0ubbe: 🤖 Fixed in [75f1bb631](https://github.com/Significant-Gravitas/AutoGPT/commit/75f1bb631c4521ea03311c7140a9f76a5f9c7db9): the reduced-motion branch now registers its own ResizeObserver that resizes the canvas, re-seeds particles, and redraws the static frame.

autogpt_platform/frontend/src/components/ui/vortex.tsx
0ubbe: 🤖 Fixed in [75f1bb631](https://github.com/Significant-Gravitas/AutoGPT/commit/75f1bb631c4521ea03311c7140a9f76a5f9c7db9): the ResizeObserver re-runs `initParticles()` when the canvas grows from a zero size, so particles spread across the panel immediately instead of clustering at x=0.

autogpt_platform/frontend/src/components/ui/vortex.tsx
```diff
@@ -186,77 +216,67 @@ export function Vortex(props: VortexProps) {
...
 
-  function handleResize() {
-    const canvas = canvasRef.current;
-    if (canvas) {
-      resize(canvas);
+    if (window.matchMedia("(prefers-reduced-motion: reduce)").matches) {
+      drawStaticFrame(canvas, ctx);
+      return;
     }
-  }
```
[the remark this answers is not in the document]
0ubbe: 🤖 Fixed in [75f1bb631](https://github.com/Significant-Gravitas/AutoGPT/commit/75f1bb631c4521ea03311c7140a9f76a5f9c7db9): the reduced-motion branch now has its own ResizeObserver that resizes the canvas, re-seeds particles, and redraws the static frame.

autogpt_platform/frontend/src/components/ui/vortex.tsx
```diff
@@ -186,77 +216,80 @@ export function Vortex(props: VortexProps) {
...
+    resize(canvas, container);
+    initParticles();
 
-  function handleResize() {
-    const canvas = canvasRef.current;
-    if (canvas) {
-      resize(canvas);
+    if (window.matchMedia("(prefers-reduced-motion: reduce)").matches) {
+      drawStaticFrame(canvas, ctx);
+      const staticResizeObserver = new ResizeObserver(() => {
```
0ubbe, line 259: 🤖 🟡 **Nice to Have**: two small inefficiencies in this branch: (1) `ResizeObserver` fires once immediately on `observe()`, so the static frame is drawn twice at mount — the direct `drawStaticFrame(canvas, ctx)` call above can be dropped; (2) the callback re-runs `initParticles()` + a 90-step warmup (~31k noise evaluations) on every resize event, which fires at frame rate during a window drag — throttling via `requestAnimationFrame` (or drawing only on the last event) keeps the reduced-motion path cheap, which is the whole point of it.

0ubbe: 🤖 Fixed in [6427e5341](https://github.com/Significant-Gravitas/AutoGPT/commit/6427e5341d47deaac39d60a12691d5c3fb6c0bb8): dropped the direct mount-time `drawStaticFrame` call (the observer's initial fire covers it) and coalesced resize redraws through a single cancelled-and-rescheduled `requestAnimationFrame`, so a drag burst produces at most one warmup+redraw per frame and the pending frame is cancelled on cleanup.

autogpt_platform/frontend/src/components/ui/__tests__/vortex.test.tsx
```diff
@@ -0,0 +1,156 @@
...
+    mockCtx as unknown as CanvasRenderingContext2D,
+  );
+  Object.defineProperty(HTMLElement.prototype, "clientWidth", {
+    configurable: true,
+    get: () => mockSize.width,
+  });
+  Object.defineProperty(HTMLElement.prototype, "clientHeight", {
+    configurable: true,
+    get: () => mockSize.height,
+  });
```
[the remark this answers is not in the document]
0ubbe: 🤖 Fixed in [02c3c7c8a](https://github.com/Significant-Gravitas/AutoGPT/commit/02c3c7c8a6a5ce3358d2e46f37ae047d817e89a7): `afterEach` now deletes the `clientWidth`/`clientHeight` overrides from `HTMLElement.prototype` (via `Reflect.deleteProperty`), restoring the original prototype chain so the descriptors can't leak across test files.

Abhi1992002 (approved):

PR #13552, opened 2026-07-13
fix(backend): prevent duplicate personal orgs — in-loop lock renewal, get-or-create backfill, partial unique index

ntindle:
### Why / What / How

**Why:** First real-user feedback from the org tenancy release surfaced duplicate personal orgs (`pwuts` + `pwuts-1`, 91s apart). Investigation against the dev DB found **84 of 422 users duplicated**, all inside 04:36–04:38 UTC on 2026-07-09: the startup backfill holds the 300s Redis migration lock across its entire per-user org-creation loop with **no renewal** (#13512's renewals cover only the top-level steps and the batched tenancy helpers). Wave 1 (started 04:28:29) silently lost the lock at +300s and kept creating — the lost-lock check only runs *after* the loop — while a second pod acquired the freed lock at 04:36:14 and re-ran from a fresh snapshot. Both raced the remaining tail. Wave 1 then aborted post-loop; wave 2 completed the migration, so tenancy assignment landed most users' resources in the *second* org — affected users log in and see an empty account (request context resolves the oldest org). Slug collisions were silently absorbed by the `-N` suffixer, and nothing at the storage layer enforces one-personal-org-per-user. **Without this fix, prod cutover repeats the incident at scale.**

**What/How — four parts:**
1. **In-loop lock renewal + fail-fast**: renew every 50 users inside `create_orgs_for_existing_users` and `migrate_credentials_to_table` (the two unbatched per-user loops); a failed renewal raises and aborts instead of silently writing without ownership.
2. **Get-or-create backfill**: per-user live re-check, `UniqueViolationError` → re-check (concurrent winner → skip) → fresh-slug retry, bounded attempts. Kept in-place rather than routing through `ensure_personal_org`, which would drop the backfill's billing/profile enrichment (`stripeCustomerId`, `topUpConfig`, avatar/bio/socials).
3. **Canonical-org agreement**: `_find_owned_personal_org`, `orgs.db._find_personal_org_member`, and the credential-migration org lookup now order `createdAt asc`, matching `get_request_context` — no path can resolve a different "the" personal org.
4. **Migration `20260713120000_org_one_personal_per_user`**: idempotent set-based dedupe (winner = most AgentGraphs → most ChatSessions → oldest → lowest id; losers soft-deleted, balances/ledgers untouched — operational remediation handles money/slug polish), then a **partial unique index** `ON "Organization"("bootstrapUserId") WHERE "isPersonal" AND "deletedAt" IS NULL AND "bootstrapUserId" IS NOT NULL` — the invariant becomes structurally impossible to violate. Personal→team conversion is compatible (it flips `isPersonal=false` before creating the replacement, exiting the index before re-entering).

Note: soft-deleting losers auto-reroutes affected users to their resource-bearing org (`get_request_context` filters `deletedAt`), so deploying this migration to dev also *heals the user experience*; the remaining operational cleanup (duplicated `OrgBalance`/`OrgCreditTransaction` rows on losers — money was copied into both orgs, must be discarded not summed — and reclaiming clean slugs like `pwuts`) runs separately.

### Changes 🏗️

- `backend/data/org_migration.py`: `_LOCK_RENEW_EVERY=50` renewal in both per-user loops (fail-fast), `_get_or_create_backfill_org`/`_create_backfill_org`, ordered `_find_owned_personal_org`, ordered credential-migration org lookup
- `backend/api/features/orgs/db.py`: ordered `_find_personal_org_member`
- `backend/schema.prisma`: comment documenting the SQL-only partial index
- `backend/migrations/20260713120000_org_one_personal_per_user/migration.sql`: dedupe + index
- Tests: `TestBackfillHardening` (renewal cadence, lock-lost aborts creation, already-bootstrapped skip, race-skip, slug retry, bounded failure, ordering) + `TestCanonicalPersonalOrgOrdering`; existing slug-edge-case mocks updated for the new re-check

#### For code changes:
- [x] I have made a test plan
  - [x] `poetry run pytest backend/data/org_migration_test.py -q` — 57 passed
  - [x] `poetry run pytest backend/api/features/orgs/ -q` — 234 passed, 12 xfailed
  - [x] Migration applies against docker Postgres via the test harness; dedupe is a no-op on clean DBs by construction

For changes touching `data/*.py`: all queries key strictly on the target `user_id`; no caller-supplied org/team ids are trusted.

---

> [!NOTE]
> **High Risk**
> Changes core org bootstrap, startup migration locking, and enforces a new partial unique index on personal orgs—errors or bad dedupe directly affect login context and account data.
> 
> **Overview**
> Fixes duplicate personal orgs from multi-pod startup backfills that outlived the 300s Redis migration lock and re-ran from a stale snapshot.
> 
> **Startup backfill** now renews the distributed lock every 50 users inside `create_orgs_for_existing_users` and `migrate_credentials_to_table`, aborting if renewal fails. Each user gets a live membership re-check before create; races and index violations are handled via `_get_or_create_backfill_org` (skip, slug retry, or loud failure). Backfill org creation runs in a **transaction** so partial org rows cannot wedge the one-personal-per-user slot.
> 
> **Orphan self-heal**: `_soft_delete_blocking_orphan` soft-deletes legacy personal orgs with `bootstrapUserId` but no owner membership; used in `ensure_personal_org`, backfill retries, and first-request `_bootstrap_personal_org`.
> 
> **Canonical org**: personal-org lookups (`_find_owned_personal_org`, `_find_personal_org_member`, credential migration) use **`createdAt` asc** so auth and migration agree when duplicates briefly exist.
> 
> **DB migration** `20260713120000_org_one_personal_per_user`: sweeps orphans, soft-deletes duplicate personal org losers (resource-weighted winner), adds partial unique index on live `bootstrapUserId` personal orgs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 034edd67db03c777fa152fb61610b265cf39530d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/autogpt_platform/backend/.env.default
+++ b/autogpt_platform/backend/.env.default
@@ -227,6 +227,9 @@ PLATFORM_LINK_BASE_URL=http://localhost:3000/link
 # CoPilot chat-platform bridge (Discord/Telegram/Slack)
 # Uses FRONTEND_BASE_URL (above) for link confirmation pages.
 AUTOPILOT_BOT_DISCORD_TOKEN=
+# Slack adapter — set BOTH to mount the Slack webhook routes on the main API.
+AUTOPILOT_BOT_SLACK_TOKEN=
+AUTOPILOT_BOT_SLACK_SIGNING_SECRET=
 
 # Communication Services
 DISCORD_BOT_TOKEN=
--- a/autogpt_platform/backend/backend/api/features/admin/bot_analytics_routes_test.py
+++ b/autogpt_platform/backend/backend/api/features/admin/bot_analytics_routes_test.py
@@ -74,7 +74,9 @@ def test_timeseries_returns_list(mocker):
 
 
 def test_command_usage_and_guilds(mocker):
-    mock_commands = AsyncMock(return_value=[BotCommandUsage(command="setup", uses=3)])
+    mock_commands = AsyncMock(
+        return_value=[BotCommandUsage(platform="DISCORD", command="setup", uses=3)]
+    )
     mock_guilds = AsyncMock(
         return_value=[
             BotGuildInfo(
--- a/autogpt_platform/backend/backend/api/features/admin/memory_admin_routes.py
+++ b/autogpt_platform/backend/backend/api/features/admin/memory_admin_routes.py
@@ -995,7 +995,7 @@ async def rebuild_communities(
         raise HTTPException(
             status_code=500,
             detail=(
-                f"Community rebuild scheduling failed: {type(exc).__name__}: {exc}"
+                f"Community rebuild scheduling failed: " f"{type(exc).__name__}: {exc}"
             ),
         )
 
--- a/autogpt_platform/backend/backend/api/features/admin/memory_admin_routes.py
+++ b/autogpt_platform/backend/backend/api/features/admin/memory_admin_routes.py
@@ -995,7 +995,7 @@ async def rebuild_communities(
         raise HTTPException(
             status_code=500,
             detail=(
-                f"Community rebuild scheduling failed: " f"{type(exc).__name__}: {exc}"
+                f"Community rebuild scheduling failed: {type(exc).__name__}: {exc}"
             ),
         )
 
--- a/autogpt_platform/backend/backend/api/features/admin/memory_admin_routes_test.py
+++ b/autogpt_platform/backend/backend/api/features/admin/memory_admin_routes_test.py
@@ -251,11 +251,12 @@ class TestRebuildCommunitiesPolling:
         scheduler.schedule_immediate_community_rebuild = AsyncMock(
             return_value={"scheduled": True, "job_id": "x", "kind": "rebuild"}
         )
-        with patch(
-            f"{_MOCK_MODULE}.get_scheduler_client", return_value=scheduler
-        ), patch(
-            f"{_MOCK_MODULE}.write_initial_status",
-            new=_make_fake_initial_status("rebuild"),
+        with (
+            patch(f"{_MOCK_MODULE}.get_scheduler_client", return_value=scheduler),
+            patch(
+                f"{_MOCK_MODULE}.write_initial_status",
+                new=_make_fake_initial_status("rebuild"),
+            ),
         ):
             resp = client.post("/admin/memory/abc/communities/rebuild")
         assert resp.status_code == 202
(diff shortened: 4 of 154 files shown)


ntindle:
Both review findings addressed in 09812a7ecc:

- **Bugbot (High)** — credential-migration org lookup now filters `deletedAt: None`; oldest-first without it could attach migrated credentials to a soft-deleted dedupe loser no request path resolves. Valid catch, exactly the interaction with this PR's own dedupe step.
- **CodeRabbit (Major)** — lock renewal moved before the already-bootstrapped re-check so skip-heavy re-runs (the second-wave scenario) renew on every processed user, not just creates; lost-lock test updated (49 creates) and a new `test_skip_heavy_rerun_still_renews_lock` covers the skip path.

197 tests green across both affected suites.

ntindle:
Bugbot's transaction-safety finding fixed in the latest commit: `_create_backfill_org`'s six writes now run in one transaction (mirroring `create_personal_org`). Valid escalation — the new unique index turned a partial create from a recoverable dupe into a permanently wedged user (orphan org blocks the index, no membership for re-checks to find). 197 tests green.

Bentlybro (dismissed):

ntindle:
Bugbot's orphan-org finding addressed: the migration now sweeps legacy orphans (live personal org with bootstrapUserId but no owner membership — soft-deleted, idempotent no-op on healthy DBs) and the violation handler self-heals at runtime by soft-deleting a blocking orphan and retrying, so startup can never wedge on one. Verified dev currently has zero such orphans (live query) — this is defense-in-depth for environments carrying pre-transaction partial creates. 198 tests green.

ntindle:
Round-five finding addressed + one self-inflicted mess cleaned up:

- **Orphan heal on every creation path**: `_soft_delete_blocking_orphan` (live personal org with this bootstrapUserId AND no owner membership — same predicate as the migration sweep) now wired into backfill, `ensure_personal_org` (signup), and the lazy `_bootstrap_personal_org` (which retries once after clearing instead of degrading to None). 294 tests green.
- The intermediate commit accidentally ruff-formatted the backend tree (repo uses black there); the follow-up commit reverts all ~110 unrelated files — net PR diff is back to the fix files only. Apologies for the noise in the interim diff.

ntindle:
Round six (both bots, same finding) addressed: the lazy bootstrap's post-orphan-clear retry now reconciles a concurrent winner (membership re-check) before reporting failure, instead of returning None when another creator won the gap. 294 tests green.

Bentlybro (approved):

Pwuts (commented):
LGTM!

PR #13559, opened 2026-07-14
fix(backend/integrations): reject webhook ingress on provider path mismatch

kcze:
### Why / What / How

**Why.** The webhook ingress route (`webhook_ingress_generic`) picks the `WebhooksManager` from the `{provider}` in the URL path, then loads the stored webhook row by ID — but never checks that the two agree. Because the manager is chosen from the (untrusted) path, a webhook registered under one provider can be processed by a *different* provider's manager. Managers for providers whose protocol has no signing scheme (Compass, Slant3D) inherit the no-op `verify_signature` from `BaseWebhooksManager`, so routing a webhook's UUID through one of those paths skips the verification the webhook's real provider would enforce.

Concretely: a generic webhook configured with a `secret_token` correctly rejects a request missing `X-Webhook-Secret` on its own `/generic_webhook/...` path, but the same UUID sent to `/compass/webhooks/{id}/ingress` runs Compass's no-op verifier instead of the generic manager's secret check, and the attached graph executes.

This is a gap from when signature verification was consolidated into a separate, provider-overridable `verify_signature` step (#13135): the step keys off the URL provider rather than the stored webhook's provider.

**What.** Reject requests whose URL provider doesn't match the stored `webhook.provider`, before verification and payload validation.

**How.** After the webhook row loads, compare `webhook.provider` to the path `provider`. On mismatch, return **403** (consistent with the existing "don't 404 — that leaks webhook existence" handling) and log the mismatch. No change to any valid-path behavior.

### Changes 🏗️

| File | Change |
|---|---|
| `api/features/integrations/router.py` | Guard `webhook_ingress_generic`: 403 when `webhook.provider != provider`, before `verify_signature` / `validate_payload` |
| `api/features/integrations/webhook_ingress_test.py` | Add `TestProviderPathConfusion`: a secret-protected generic webhook can't be triggered via `/compass/...` or `/slant3d/...`; matching-provider path still processes |

#### For code changes:
- [x] I have made a test plan
  - [x] `poetry run pytest backend/api/features/integrations/webhook_ingress_test.py` — 27 passed
  - [x] Verified the two new cross-provider tests fail with the guard removed and pass with it in place
  - [x] `ruff` / `black` / `isort` / `pyright` clean on changed files


kcze (commented):
🤖 Review done, no findings.

autogpt_platform/backend/backend/api/features/integrations/router.py
```diff
@@ -569,6 +569,20 @@ async def webhook_ingress_generic(
         raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail=str(e))
     logger.debug(f"Webhook #{webhook_id}: {webhook}")
 
+    # The manager is selected from the URL `provider`, so a request must not be
+    # allowed to process a webhook registered under a different provider: that
+    # would run the URL provider's verifier (a no-op for unsigned providers)
+    # instead of the stored provider's. 403, not 404 — don't leak existence.
+    if webhook.provider != provider:
+        logger.warning(
+            f"Webhook #{webhook_id} provider mismatch: "
```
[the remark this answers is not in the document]
kcze: Good catch — fixed in 167a10e0de. The comparison now normalizes case (`webhook.provider.value.lower() != provider.value.lower()`). In practice ingress URLs are always generated with the canonical lowercase provider so legit traffic never hit this, but the case-insensitive compare removes the sharp edge, and I added a `test_matching_provider_different_case_accepted` regression test.

autogpt_platform/backend/backend/api/features/integrations/router.py
```diff
@@ -569,6 +569,20 @@ async def webhook_ingress_generic(
...
+            f"registered as {webhook.provider.value}, ingress via {provider.value}"
+        )
+        raise HTTPException(
+            status_code=status.HTTP_403_FORBIDDEN,
+            detail="Invalid webhook signature",
+        )
+
```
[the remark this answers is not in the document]
kcze: Fixed in 167a10e0de. The provider-mismatch guard now runs immediately after the webhook row loads and before any credential lookup, so a mismatched request returns 403 without touching the credential store. Credential retrieval moved below the guard (still wrapped to map `NotFoundError` → 404).

autogpt_platform/backend/backend/api/features/integrations/webhook_ingress_test.py
```diff
@@ -445,6 +447,64 @@ def test_token_configured_correct_header_accepted(self):
...
+    def test_matching_provider_path_still_processes(self):
+        # Control: the same webhook on its own path with the correct secret is
+        # accepted — the guard rejects only provider mismatches.
+        resp = self._run_cross_provider(
+            self._secret_protected_generic(),
+            "generic_webhook",
+            ProviderName("generic_webhook"),
+            headers={"X-Webhook-Secret": "t0ken"},
+        )
+        assert resp.status_code != 403, resp.text
```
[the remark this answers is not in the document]
kcze: Done in 167a10e0de — the matching-provider control test now asserts `== 200`.

autogpt_platform/backend/backend/api/features/integrations/router.py
```diff
@@ -558,7 +558,30 @@ async def webhook_ingress_generic(
...
+    # leak existence. Compared case-insensitively: ingress URLs are canonical,
+    # but a non-canonical path shouldn't slip past on casing alone.
+    if webhook.provider.value.lower() != provider.value.lower():
+        logger.warning(
+            f"Webhook #{webhook_id} provider mismatch: "
+            f"registered as {webhook.provider.value}, ingress via {provider.value}"
+        )
+        raise HTTPException(
+            status_code=status.HTTP_403_FORBIDDEN,
+            detail="Invalid webhook signature",
```
Pwuts: > 403, not 404 - don't leak existence

"Invalid webhook signature" *confirms* existence, because it's not the same error as the 404 you get when the webhook does not exist. Should be the same 404 as returned when the webhook does not exist.

autogpt_platform/backend/backend/api/features/integrations/router.py
```diff
@@ -558,7 +558,30 @@ async def webhook_ingress_generic(
...
+        raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail=str(e))
+    logger.debug(f"Webhook #{webhook_id}: {webhook}")
+
+    # The manager is selected from the URL `provider`, so a request must not be
+    # allowed to process a webhook registered under a different provider: that
+    # would run the URL provider's verifier (a no-op for unsigned providers)
+    # instead of the stored provider's. Checked before any credential lookup so
+    # a mismatch can't touch the credential store or 500. 403, not 404 — don't
+    # leak existence. Compared case-insensitively: ingress URLs are canonical,
+    # but a non-canonical path shouldn't slip past on casing alone.
```
Pwuts: we should really add a "be brief in comments" rule to CLAUDE.md lol
```suggestion
    # Sanity check: `provider` from URL and fetched webhook must match
```

And then above the exception you can add a short:
```py
        # Same as actual "webhook not found" response to conceal existence
```

autogpt_platform/backend/backend/api/features/integrations/router.py
kcze: Agreed — the distinct error confirmed existence. Fixed: a provider mismatch now raises `NotFoundError(f"Webhook #{webhook_id} not found")` and goes through the exact same `NotFoundError` → 404 handler as a nonexistent webhook, so the response is byte-identical (same status, same detail). Added an assertion in the bypass tests that the mismatch `detail` equals the not-found message.

autogpt_platform/backend/backend/api/features/integrations/router.py
kcze: Done — trimmed to a short "Sanity check: \`provider\` from URL and fetched webhook must match" plus a one-liner above the raise noting it mirrors the "webhook not found" response to conceal existence. (And noted on the CLAUDE.md brevity rule 😄)

Pwuts (approved):
LGTM!

PR #13580, opened 2026-07-15
fix(platform): track account creation in DataFast

Torantulino:
### Why

The DataFast "Create Account" funnel step receives no events because account creation is not an automatic DataFast event. DataFast documents signup tracking as a [custom goal](https://datafa.st/docs/custom-goals), with backend-confirmed signups sent through the [Website Goals API](https://datafa.st/docs/api/website/goals/create).

### What changed

- Preserve the existing cached `get_or_create_user()` API while exposing an uncached creation result for the `/auth/user` response.
- Add `X-AutoGPT-User-Created: true|false` so callers can distinguish a newly inserted account from a returning user without timestamp inference.
- Send the DataFast `signup` goal after successful email or Google account creation, using the visitor UUID from DataFast's cookie.
- Mirror existing analytics consent into a server-readable, same-site cookie; tracking fails closed when consent or a valid visitor UUID is absent.
- Send the goal with the server-only `DATAFAST_API_KEY`, no PII, a five-second timeout, and failure isolation through Sentry.
- Add focused regression tests for the creation signal, both signup paths, consent handling, payload shape, configuration failure, and delivery failure.

### Scope

This change does not alter Tally triggering, OAuth error handling, existing user provisioning, or unrelated formatting/comments. The generated API client is used in the OAuth callback only because the legacy client does not expose response headers.

### Hard deployment requirement

This PR **must not be considered deployed successfully** until all of the following are true:

- [ ] Create a Website API key in DataFast Website Settings > Developer.
- [ ] Set `DATAFAST_API_KEY` to that `df_...` key in the production frontend runtime, including Vercel Production.
- [ ] Configure the funnel's "Create Account" phase to use the custom goal name `signup`.
- [ ] Create one brand-new account with analytics consent and a `datafast_visitor_id` cookie; verify exactly one `signup` goal appears in DataFast.

The secret is intentionally not committed. Missing cloud configuration or an invalid configured key never blocks account creation and is reported to Sentry; self-hosted deployments without a key skip DataFast tracking.

### Validation

- [x] Focused frontend regression suite: 22/22 tests
- [x] Frontend Prettier check on every touched TypeScript file
- [x] Repository-wide Next lint (only the three existing `` warnings)
- [x] Frontend TypeScript check
- [x] OpenAPI generation (no generated-client diff)
- [x] Backend Python syntax and no-stdlib-dataclass architecture gates
- [x] `git diff --check`
- [x] Final GitHub CI on the focused, rebased head


autogpt_platform/frontend/src/app/(no-navbar)/signup/actions.ts
```diff
@@ -59,7 +63,10 @@ export async function signup(
     }
 
     try {
-      await resolveResponse(postV1GetOrCreateUser());
+      const createUserResponse = await postV1GetOrCreateUser();
```
0ubbe, line 66: 🤖 🟠 **Should Fix**: Dropping `resolveResponse` here silently swallows failed user creation. `resolveResponse` throws on non-2xx, so a 500 from `/auth/user` used to land in this `catch` → Sentry. Now `createUserResponse.status` is never checked — a failure sails through unreported (and `wasAccountCreated` just returns `false`). Suggest `if (createUserResponse.status !== 200) throw new Error(...)` before the header check.

autogpt_platform/frontend/src/app/(platform)/auth/callback/route.ts
```diff
@@ -22,8 +26,10 @@ export async function GET(request: Request) {
 
     if (!error) {
       try {
-        const api = new BackendAPI();
-        await api.createUser();
+        const createUserResponse = await postV1GetOrCreateUser();
```
0ubbe, line 29: 🤖 🟠 **Should Fix**: Same as in the signup action — the legacy `api.createUser()` threw on failure, but the generated client resolves with the error status instead. A non-2xx here now goes unnoticed; add a status check before `wasAccountCreated`.

autogpt_platform/backend/backend/api/features/v1.py
```diff
@@ -192,13 +194,14 @@ def _create_file_size_error(size_bytes: int, max_size_mb: int) -> HTTPException:
...
+):
+    result = await get_or_create_user_with_status(user_data)
+    response.headers[USER_CREATED_HEADER] = str(result.was_created).lower()
+    user = result.user
 
     # Fire-and-forget: populate business understanding from Tally form.
-    # We use created_at proximity instead of an is_new flag because
-    # get_or_create_user is cached — a separate is_new return value would be
-    # unreliable on repeated calls within the cache TTL.
     age_seconds = (datetime.now(timezone.utc) - user.created_at).total_seconds()
```
0ubbe, line 216: 🤖 🟠 **Should Fix**: `result.was_created` is now available in this scope, and the comment that justified the `created_at < 30s` heuristic (cache staleness) was removed because it no longer applies. Suggest `if result.was_created:` for the Tally trigger too — that makes it exactly-once, whereas the 30s window can fire multiple times when `/auth/user` is hit repeatedly right after signup.

autogpt_platform/frontend/src/services/analytics/datafast-server.ts
```diff
@@ -0,0 +1,97 @@
...
+  const hasConsent =
+    cookieStore.get(ANALYTICS_CONSENT_COOKIE)?.value ===
+    ANALYTICS_CONSENT_GRANTED;
+  if (!hasConsent) return null;
+
+  const visitorID = cookieStore.get(DATAFAST_VISITOR_COOKIE)?.value;
+  if (!visitorID || !VISITOR_ID_PATTERN.test(visitorID)) return null;
+
+  const apiKey = process.env.DATAFAST_API_KEY?.trim();
+  if (!apiKey?.startsWith("df_")) {
```
0ubbe: 🤖 🟡 **Nice to Have**: A missing `DATAFAST_API_KEY` throws → `Sentry.captureException` on every consenting signup. Self-hosted instances will never configure DataFast, so for them this becomes a permanent per-signup Sentry error. Consider gating the missing-key path on cloud env (e.g. `NEXT_PUBLIC_BEHAVE_AS`) or logging once, keeping the throw for malformed keys only.

autogpt_platform/frontend/src/app/(no-navbar)/signup/__tests__/actions.test.ts
```diff
@@ -0,0 +1,79 @@
...
+}));
+vi.mock("@/app/api/helpers", () => ({
+  getOnboardingStatus: mocks.getOnboardingStatus,
+}));
+vi.mock("@/lib/supabase/server/getServerSupabase", () => ({
+  getServerSupabase: mocks.getServerSupabase,
+}));
+vi.mock("@/services/analytics/datafast-server", () => ({
+  scheduleAccountCreatedGoal: mocks.scheduleAccountCreatedGoal,
+  wasAccountCreated: (headers: Headers) =>
```
0ubbe: 🤖 🟡 **Nice to Have**: This mock re-implements `wasAccountCreated` (header name included), so if the real constant ever drifts the test keeps passing against the copy. Suggest `vi.mock` with `importOriginal` and mocking only `scheduleAccountCreatedGoal` — same in `auth/callback/route.test.ts`.

autogpt_platform/backend/backend/api/features/v1.py
```diff
@@ -192,13 +194,14 @@ def _create_file_size_error(size_bytes: int, max_size_mb: int) -> HTTPException:
     tags=["auth"],
     dependencies=[Security(requires_user)],
 )
-async def get_or_create_user_route(user_data: dict = Security(get_jwt_payload)):
-    user = await get_or_create_user(user_data)
+async def get_or_create_user_route(
+    response: Response, user_data: dict = Security(get_jwt_payload)
+):
+    result = await get_or_create_user_with_status(user_data)
+    response.headers[USER_CREATED_HEADER] = str(result.was_created).lower()
```
0ubbe, line 212: 🤖 🔵 **Nit**: The header-vs-body tradeoff is well explained in the PR description, but `X-AutoGPT-User-Created` is invisible in the OpenAPI schema. Declaring it in the route's `responses` metadata would document it for free for future client regens.

autogpt_platform/backend/backend/api/features/v1_user_created_test.py
```diff
@@ -0,0 +1,39 @@
...
+import pytest
+from fastapi import Response
+
+from backend.api.features.v1 import get_or_create_user_route
+from backend.data.model import User
+from backend.data.user import UserCreationResult
+
+
+@pytest.fixture(autouse=True)
+def graph_cleanup():
```
0ubbe: 🤖 🔵 **Nit**: This no-op fixture shadows the conftest `graph_cleanup` by name — works, but a conftest rename silently re-enables the heavy fixture with no error here. Consider colocating this test in `v1_test.py` next to the existing route test instead.

autogpt_platform/frontend/src/app/(no-navbar)/signup/actions.ts
Torantulino: Verified against the current generated `customMutator`: it already throws `ApiError` on non-2xx responses, so the current client does not silently resolve a 500. I still added defense in depth in `wasAccountCreated`: any resolved non-200 response now throws an error that preserves `.status`. A regression test explicitly resolves a 500 and verifies email signup enters the existing catch path, reports to Sentry, and does not schedule DataFast tracking. Implemented in 51469878c.

autogpt_platform/frontend/src/app/(platform)/auth/callback/route.ts
Torantulino: Verified independently against the current generated `customMutator`: it throws `ApiError` on non-2xx responses. I nevertheless applied the same shared defense-in-depth status validation here. The OAuth regression test now explicitly resolves a 500 and verifies the callback takes the server-error redirect and does not schedule DataFast tracking. Implemented in 51469878c.

autogpt_platform/backend/backend/api/features/v1.py
Torantulino: I verified the existing Tally path and intentionally kept its `created_at < 30s` behavior unchanged. Switching that separate integration to `result.was_created` may be a worthwhile follow-up, but it changes Tally triggering semantics and is outside this PR's atomic DataFast account-creation tracking scope. This PR uses the exact creation result only for the response header consumed by DataFast; it does not change the Tally trigger.

autogpt_platform/frontend/src/services/analytics/datafast-server.ts
Torantulino: Implemented in 51469878c after verifying the environment helper. A missing key now skips DataFast silently outside cloud; missing cloud configuration is reported to Sentry; and a present malformed key is still reported in every environment. Focused tests cover all three cases.

autogpt_platform/frontend/src/app/(no-navbar)/signup/__tests__/actions.test.ts
Torantulino: Implemented in 51469878c. Both signup-path tests now use an async `importOriginal` partial mock and replace only `scheduleAccountCreatedGoal`, so they exercise the real `wasAccountCreated` implementation and shared header/status contract.

autogpt_platform/backend/backend/api/features/v1.py
Torantulino: Implemented in 51469878c. The route now declares `X-AutoGPT-User-Created` in the 200 response headers metadata with the `true|false` enum. I regenerated the OpenAPI schema, confirmed there is no generated-client diff, and added an assertion for the published schema.

autogpt_platform/backend/backend/api/features/v1_user_created_test.py
Torantulino: Implemented in 51469878c. The response-header test is now colocated with the existing `/auth/user` route tests in `v1_test.py`, and the standalone test file plus its shadowing no-op fixture have been removed.

autogpt_platform/backend/backend/api/features/v1.py
Torantulino: Follow-up: the first schema commit placed the header block on the wrong generated operation. The `check API types` job caught it. Commit 2ea8ababd moves only that block to `/api/auth/user`; the resulting `openapi.json` blob hash exactly matches the backend export shown by CI (`57aef64…`). Orval still produces no generated-client diff, TypeScript passes, and the 22 focused tests remain green.

0ubbe (approved):
Nice work! 💚

PR #13596, opened 2026-07-17
feat(copilot): add Kimi K3 model support (moonshotai/kimi-k3)

Otto-AGPT:
## Summary

Adds support for Kimi K3 (`moonshotai/kimi-k3`) as a routing option for the thinking-standard model cell.

K3 launched on OpenRouter on Jul 16 2026. Open weights are promised by **Jul 27 2026** — keeping as draft until then to verify the model is stable in production and weights have shipped.

## Changes

- **`anthropic_rates.json`**: Add K3 cost entry ($3/M input, $15/M output, 1,048,576 token context)
- **`model_router.py`**: Update LD example payload to reference K3 as `thinking.standard`
- **`reasoning.py`**: Update docstring to mention K3 alongside K2.6

## No logic changes needed

`_is_reasoning_route` already handles the `moonshotai/` prefix, so K3 works on the baseline path immediately via LD override (`copilot-model-routing`) or `CHAT_THINKING_STANDARD_MODEL=moonshotai/kimi-k3`.

## To activate

Set LD flag `copilot-model-routing`:
```json
{
  "thinking": {"standard": "moonshotai/kimi-k3", "advanced": "anthropic/claude-opus-4-6"}
}
```

## Known caveats (recheck on Jul 27)
- K3 always runs max reasoning effort — no dial-down, every call burns reasoning tokens
- 51% hallucination rate on AA-Omniscience (vs 38% for Sonnet 4.6) — monitor in production
- Open weights not yet released; undraft once weights ship and routing has been validated

PR #13605, opened 2026-07-18
feat(backend/db): add LLM model registry schema

ntindle:
## Why

Adding an LLM model today is a code change in 3+ places plus a deploy, copilot routing lives in a raw LaunchDarkly JSON flag nothing validates (see #13596's dead rates entry), and self-hosted installs only get new models by pulling code. This restarts @Bentlybro's LLM registry stack (#12357 et al. — thanks for the groundwork, co-authored accordingly) as a full re-cut onto current dev: **part 1 of 9**, schema only.

## What

4 enums + 7 tables: `LlmProvider`, `LlmModel`, `LlmModelCost`, `LlmModelCreator`, `LlmModelMigration`, `LlmModelRoute`, `LlmCatalogState`. Nothing reads or writes them yet — this PR is dormant infrastructure so the stack lands in reviewable slices.

Deltas vs the original #12357 design:
- **`source` tracking** (`SEED`/`REMOTE`/`LOCAL`) on provider/model/creator — the upcoming idempotent catalog importer upserts non-LOCAL rows and never touches admin-owned (LOCAL) ones
- **`isEnabled` / `visibility` split** — `isEnabled=false` is the kill switch (never serves, even when LD routes to it); `visibility=HIDDEN` serves when explicitly routed but never shows in pickers/catalog (pre-launch testing state); plus `EMPLOYEES`/`ADMINS` audiences
- **`LlmModelRoute`** — `(surface, mode, tier)` → model cells: the admin-set config layer that will sit under the existing `copilot-model-routing` LD flag
- **`LlmCatalogState`** singleton — importer/sync observability + content-hash fast-path
- `kind` (CHAT), `minSubscriptionTier` (nullable, enforcement later), `fallbackModelSlug` self-relation (pre-fills disable-with-replacement; future failover hook), `catalogRemovedAt`
- **Dropped** `LlmProvider.defaultCredential*` — provider credentials are deferred until vault work; keys keep coming from settings

## How

Schema block re-cut from `origin/feat/llm-registry-schema` with the deltas above; migration SQL generated via `prisma migrate diff` + hand-written tail for what Prisma can't express (partial unique indexes on `LlmModelCost`, active-migration partial unique, CHECK constraints, singleton guard).

The original one-shot seed migration is intentionally NOT carried over — seeding becomes an idempotent startup importer (part 3) fed by a bundled catalog generated from today's `MODEL_METADATA`.

⚠️ If you ever applied the old branch's `20260310*` migrations locally, your `_prisma_migrations` table diverges — `poetry run prisma migrate reset` is needed.

## Verification

- `prisma validate` + `prisma format` clean; client generation succeeds
- Migration applied to a scratch Postgres 16: all 7 tables created; CHECK constraints and partial unique indexes verified; `LlmCatalogState` singleton guard rejects non-singleton rows

## Checklist
- [x] Migration verified against a clean database
- [x] No user-data paths touched (schema-only, no reads/writes exist yet)

### DIFF
--- a/autogpt_platform/backend/migrations/20260717000000_add_llm_registry/migration.sql
+++ b/autogpt_platform/backend/migrations/20260717000000_add_llm_registry/migration.sql
@@ -1,0 +1,208 @@
+-- CreateEnum
+CREATE TYPE "LlmCostUnit" AS ENUM ('RUN', 'TOKENS');
+
+-- CreateEnum
+CREATE TYPE "LlmCatalogSource" AS ENUM ('SEED', 'REMOTE', 'LOCAL');
+
+-- CreateEnum
+CREATE TYPE "LlmModelKind" AS ENUM ('CHAT');
+
+-- CreateEnum
+CREATE TYPE "LlmModelVisibility" AS ENUM ('GA', 'EMPLOYEES', 'ADMINS', 'HIDDEN');
+
+-- CreateTable
+CREATE TABLE "LlmProvider" (
+    "id" TEXT NOT NULL,
+    "createdAt" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    "updatedAt" TIMESTAMP(3) NOT NULL,
+    "name" TEXT NOT NULL,
+    "displayName" TEXT NOT NULL,
+    "description" TEXT,
+    "source" "LlmCatalogSource" NOT NULL DEFAULT 'SEED',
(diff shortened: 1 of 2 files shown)


ntindle:

Replaced by #13621 (the runtime tables that legitimately stay in the DB: ChatMessage model stamps + retirement records) and #13622 (the catalog file, which replaces the seven registry tables).

PR #13606, opened 2026-07-18
feat(backend): add LLM registry read cache with pub/sub refresh

ntindle:
## Why

Part 2 of 9 of the LLM registry restack (see #13605 for the schema and full context). This layer gives every backend pod a fast, always-warm view of the registry so later parts (public catalog endpoint, copilot routing, admin API) never touch the DB on hot paths.

## What

`backend/data/llm_registry/` — re-cut of @Bentlybro's #12359 core layer:

- **Two-level cache**: L1 module-global dict under an `asyncio.Lock`; L2 Redis via `@cached(ttl=300, shared_cache=True)` on the DB fetch. All read functions are sync L1 lookups.
- **Read API**: `get_model`, `get_all_models`, `get_enabled_models`, `get_schema_options`, `get_default_model_slug`, `get_all_model_slugs_for_validation`, and new `get_route(surface, mode, tier)` for the routing cells.
- **Cross-pod refresh**: `llm_registry:refresh` channel; `refresh_llm_registry` + long-lived subscribe task wired into the rest_api lifespan (fail-soft startup, clean cancellation on shutdown).

Still dormant: nothing consumes these reads yet.

## How (deltas vs the original branch)

- Pub/sub rewritten to **sharded SPUBLISH/SSUBSCRIBE** via `connect_sharded_pubsub_async` — prod Redis is cluster mode; classic pub/sub broadcasts to every node. Mirrors `event_bus.AsyncRedisEventBus`.
- `ModelMetadata` NamedTuple import from `blocks/llm.py` replaced with a standalone frozen `RegistryModelMetadata` — decouples the registry from the module it will eventually replace, and NamedTuples serialize as JSON arrays (wrong for API payloads).
- `RegistryModel` carries the new schema columns (`kind`, `visibility`, `minSubscriptionTier`, `source`, `catalogRemovedAt`, `fallbackModelSlug`).

## Verification

- 30/30 tests pass locally (`backend/data/llm_registry/`) — record mapping edge cases, refresh/L1 population incl. routes, cache clearing both fetches, sharded pub/sub happy path + handshake filtering + reconnect + close-on-exit

## Checklist
- [x] Data-layer additions have no user-facing routes yet (no user ID checks needed — no user-scoped data involved; registry rows are global config)

### DIFF
--- a/autogpt_platform/backend/backend/api/rest_api.py
+++ b/autogpt_platform/backend/backend/api/rest_api.py
@@ -1,3 +1,4 @@
+import asyncio
 import contextlib
 import logging
 import platform
@@ -52,6 +53,7 @@ import backend.api.features.workspace.routes as team_routes
 import backend.data.block
 import backend.data.db
 import backend.data.graph
+import backend.data.llm_registry
 import backend.data.org_migration
 import backend.data.redis_client
 import backend.data.user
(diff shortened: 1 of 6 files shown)


ntindle:

Replaced by #13622 — the same read interface and L1 cache, loaded from the catalog file; Redis L2 and pub/sub are unnecessary when the data only changes at deploy.

PR #13607, opened 2026-07-18
feat(backend): add idempotent LLM catalog importer and bundled catalog

ntindle:
## Why

Part 3 of 9 of the LLM registry restack (#13605 → #13606 → this). The original stack seeded the registry with a one-shot SQL migration — frozen at install time, no way to update self-hosted installs, and no protection for admin edits. This PR replaces seeding with an idempotent importer that runs on every startup, which is the foundation the remote-sync mechanism (part 5) builds on.

## What

- **`catalog_model.py`** — `CatalogPayload` schema v1: the single source of truth for the bundled file, the public endpoint (part 4), and sync validation (part 5). Facts only: no costs, no credentials, no routing cells, no non-GA models. Slug patterns, bounds, and a 2000-model cap.
- **`importer.py`** — merge rules: upsert-by-slug; `source=LOCAL` (admin-owned) rows never updated or disabled, enforced *inside* every UPDATE's WHERE clause so a concurrent admin claim wins the race; absent models are disabled + `catalogRemovedAt`-stamped, never deleted; `LlmModelCost` never touched; sha256 content-hash fast-path (steady-state boot = one SELECT); one transaction, then cache refresh + cross-pod fan-out.
- **`export.py`** — regenerates `catalog.json` from a populated DB (`python -m backend.data.llm_registry.export`); GA-visibility models only.
- **`catalog.json`** — 85 models / 8 providers / 17 creators, generated from today's `MODEL_METADATA` (deliberately not the March-era seed data); `gpt-5.2` recommended, mirroring `DEFAULT_LLM_MODEL`.
- **Lifespan** — bundled import before registry refresh, fail-soft (empty registry degrades to hardcoded behavior).

## Verification

- 51/51 tests pass locally — real-DB integration tests for each merge rule (create, hash fast-path, LOCAL survival under both update and removal, disable-not-delete, reappearance re-enable, forward-referenced fallback pointers, schema_version rejection, full bundled import), catalog schema validation edges, and a forever-guard that `catalog.json` parses with unique slugs and sound references

## Checklist
- [x] `data/*.py` changes: registry tables are global config, no user-scoped rows — no user ID checks applicable


ntindle:

Replaced by #13622 — catalog.py is promoted from seed data to the database itself; the importer/merge semantics have no second writer to exist for.

PR #13608, opened 2026-07-18
feat(backend/api): public LLM catalog endpoint

ntindle:
## Why

Part 4 of 9 of the LLM registry restack (#13605 → #13606 → #13607 → this). Self-hosted installs need a way to learn about new/disabled models without pulling code; this endpoint is the distribution channel their sync client (part 5) polls.

## What

`GET /api/llm/catalog` — returns the `CatalogPayload` (exact schema the importer validates, so the endpoint literally serves what the sync client consumes):

- **Unauthenticated** (deliberate; fixes the old branch's `Security(requires_user)` which would have made self-hosted sync impossible): public model facts only — no costs, no credentials, no routing cells, and only GA-visibility models
- **Zero DB reads**: assembled from the registry's in-process L1 cache
- **CDN-friendly**: `Cache-Control: public, max-age=300, stale-while-revalidate=3600` on 200s; exact-path `CACHEABLE_PATHS` entry (a bare `/api/llm` prefix would allowlist future authed sub-routes — commented in the middleware); 429s explicitly `no-store` because the middleware leaves allowlisted paths' headers alone
- **Per-IP rate limit** (default 30/min, `LLM_CATALOG_RATE_LIMIT_PER_MINUTE`): single-key INCRBY+EXPIRE pipeline (Redis-cluster-safe), **fail-open** — unlike copilot's spend limiter this protects read capacity, not money; a Redis brown-out must not take down catalog distribution
- **`get_client_ip()`**: first X-Forwarded-For parsing in the backend; trusts the LB-appended entry with configurable depth (`LLM_CATALOG_CLIENT_IP_XFF_DEPTH`), socket peer for proxyless self-hosted

## Verification

- 14/14 tests: GA-only filtering (HIDDEN/EMPLOYEES excluded, disabled-GA included as disabled), payload referential integrity, cacheable 200 vs no-store 429 vs no-store sibling paths (middleware wired into the test app), XFF parsing incl. forgeable-entry ordering, limiter under/over/fail-open

## Checklist
- [x] Public endpoint intentionally unauthenticated — serves only non-sensitive global facts, filtered to GA visibility; reviewed against cache middleware semantics

### DIFF
--- a/autogpt_platform/backend/backend/api/features/llm/rate_limit.py
+++ b/autogpt_platform/backend/backend/api/features/llm/rate_limit.py
@@ -1,0 +1,67 @@
+"""Per-IP rate limiting for the public LLM catalog endpoint.
+
+Fixed one-minute window on a single Redis key per IP (cluster-safe — one key
+hashes to one slot, so the INCR+EXPIRE pipeline needs no cross-slot MULTI).
+
+Deliberately FAIL-OPEN, unlike copilot's spend limiter: this protects read
+capacity for public facts, not money. A Redis brown-out must not take down
+catalog distribution — the CDN cache in front is the real shield; this limit
+is the backstop against cache-busting abuse.
+"""
+
+from __future__ import annotations
+
+import logging
+from datetime import datetime, timezone
+
+import fastapi
+from redis.exceptions import ConnectionError as RedisConnectionError
+from redis.exceptions import RedisClusterException, RedisError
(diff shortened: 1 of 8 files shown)


ntindle:

Replaced by #13624 — same endpoint, re-cut for the file-backed cache (costs and routing now explicitly excluded from the public payload).

PR #13609, opened 2026-07-18
feat(backend): remote LLM catalog sync for self-hosted installs

ntindle:
## Why

Part 5 of 9 of the LLM registry restack (#13605 → #13606 → #13607 → #13608 → this). This is the piece that makes the registry strictly better than the enum for self-hosted users: new models and remote disables arrive within a day, no `git pull` required.

## What

`backend/data/llm_registry/sync.py` — a lifespan background loop that fetches the cloud catalog on startup and every 24h, validates it, and feeds it to the part-3 importer with `source=REMOTE`:

- **Cloud never fetches from itself**: gated on `behave_as == LOCAL` (not `app_env` — self-hosted installs also run `app_env=local`; `behave_as` is what encodes "our cloud")
- **Validation before any write**: pydantic schema + 2MiB size cap + the importer's schema_version check; a hostile or malformed payload is rejected wholesale, last-known-good stays
- **Fetch hygiene**: `util.request.Requests` (SSRF guard, retries), catalog host in `trusted_origins` so pointing `LLM_CATALOG_URL` at an internal mirror works, 15s timeout
- **Ops**: loop never dies (every attempt wrapped + recorded on `LlmCatalogState.lastRemoteSyncAt/lastRemoteSuccessAt`); single-key Redis NX lock dedups multi-pod installs; startup never blocks on the network (background task)
- Runs in the rest_api lifespan, not the scheduler — that process is Prisma-less and this only matters on self-hosted single-pod installs anyway

Config (mirrored in `.env.default`): `LLM_CATALOG_SYNC_ENABLED=true`, `LLM_CATALOG_URL=https://backend.agpt.co/api/llm/catalog`, `LLM_CATALOG_SYNC_INTERVAL_HOURS=24`.

## Verification

- 10 new tests: behave_as/flag gating, fetch→validate→import flow with payload + source_url assertions, oversized and malformed payloads rejected before import, loop survival on fetch failure, success/failure attempt recording, lock-held skip, disabled-loop immediate return; plus registry suite regression green

## Checklist
- [x] Outbound fetch reviewed: SSRF-guarded helper, validated-before-write, size-capped, fail-soft

### DIFF
--- a/autogpt_platform/backend/.env.default
+++ b/autogpt_platform/backend/.env.default
@@ -270,3 +270,11 @@ TALLY_API_KEY=
 
 # Other Services
 AUTOMOD_API_KEY=
+
+## ===== LLM MODEL CATALOG SYNC ===== ##
+# Self-hosted installs fetch the model catalog from cloud on startup + daily.
+# Set LLM_CATALOG_SYNC_ENABLED=false to disable, or point LLM_CATALOG_URL at a
+# mirror. Never active when BEHAVE_AS=cloud.
+LLM_CATALOG_SYNC_ENABLED=true
+LLM_CATALOG_URL=https://backend.agpt.co/api/llm/catalog
+LLM_CATALOG_SYNC_INTERVAL_HOURS=24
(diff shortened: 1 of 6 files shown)


ntindle:

No replacement by design — CD propagates the catalog at deploy; self-hosted installs get it on upgrade. A raw-GitHub fetch remains an easy add if between-release freshness is ever wanted.

PR #13610, opened 2026-07-18
feat(backend/copilot): resolve copilot models through the LLM registry

ntindle:
## Why

Part 6 of 9 of the LLM registry restack (#13605 → #13606 → #13607 → #13608 → #13609 → this). Today copilot routing is an LD JSON flag returning raw model slugs nothing validates — the #13596 failure mode (LD referencing a model the platform has no data for) is structurally possible. This PR makes the registry the serve-time gate while keeping LD as the experiment layer, and stamps served models onto chat messages so model experiments become measurable.

## What

**Resolution chain** (`copilot/model_router.py`): LD slug flag (per-user, unchanged JSON shape) → registry routing cell (`LlmModelRoute`, the admin-page config layer) → `ChatConfig` env default.

- **Serve-time gating**: unknown-to-registry or `isEnabled=false` (kill switch) slugs are refused — warning log + Sentry event + Redis-recorded resolution warning — and resolution falls through to the next layer. A bad LD edit degrades to the default, loudly, instead of erroring at users or serving a dead model.
- **`visibility=HIDDEN` serves when routed**: the pre-launch state — register K3 as HIDDEN, LD-target employees, watch scores, flip GA.
- **Empty registry gates nothing**: installs where the registry is dormant keep exact pre-registry behavior — all 27 pre-existing router tests pass unchanged.
- `resolve_model()` remains as a wrapper; the SDK path keeps working untouched.

**Measurement stamp**: `ChatMessage.model` + `ChatMessage.routingSource` ("ld"/"db"/"env") — nullable columns, own migration — stamped by the baseline service on every persisted assistant message. Product-intelligence's chat mirror currently captures no model identifier, making every model experiment unmeasurable; this is the join key (PI-side mirror column is a follow-up in that repo).

**`route_warnings.py`**: single-key Redis hash (cluster-safe) of recent routing refusals with counts — the admin API (part 7) exposes it so a typo'd LD slug is a minutes-long incident, not an invisible one.

## Known scope cut

The SDK-path service resolves through the same validated chain but doesn't stamp its messages yet — its persistence path differs and the stamp will follow with the part-7 wiring. Called out here so it's a decision, not an accident.

## Verification

- 36/36 model_router tests: all 27 pre-existing behavior tests unchanged + 9 new (known/unknown/disabled/HIDDEN via LD, db-cell hit, LD-beats-cell, stale-cell fall-through with layer attribution, no-user cell resolution, empty-registry passthrough)
- Copilot db/baseline suites: green except 7 pre-existing failures that reproduce identically on the base branch without this diff (local Redis-cluster env quirk; verified via stash)
- `poetry run format` + `poetry run lint` clean; migration applied on the full local chain

## Checklist
- [x] Fall-through preserves current behavior when registry is empty/cells unset (tested)

### DIFF
--- a/autogpt_platform/backend/backend/copilot/baseline/service.py
+++ b/autogpt_platform/backend/backend/copilot/baseline/service.py
@@ -56,7 +56,7 @@ from backend.copilot.model import (
     upsert_chat_session,
 )
 from backend.copilot.model_normalize import normalize_model_for_transport
-from backend.copilot.model_router import resolve_model
+from backend.copilot.model_router import ResolvedModel, resolve_model_route
 from backend.copilot.moonshot import is_moonshot_model
 from backend.copilot.pending_message_helpers import (
     combine_pending_with_current,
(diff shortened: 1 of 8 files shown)


ntindle:

Replaced by #13623 — carried nearly whole (resolution chain, serve-time gating, warnings, ChatMessage stamping), plus two integration fixes found in the re-cut (LD slug-spelling tolerance; transport vendor-prefixing for bare cell slugs).

PR #12467, opened 2026-03-18
feat(platform): Add LLM registry admin write API

Bentlybro:
## Summary
Add admin write API for LLM registry management - **Part 4 of 5** in incremental rollout.

Builds on PR #12371 (public read API) to provide full CRUD operations for admins.

### Admin Write Endpoints (\`backend/server/v2/llm/admin_routes.py\`)
All endpoints require admin authentication (\`requires_admin_user\`):

**Models:**
- \`POST /api/llm/models\` - Create model with full validation
- \`PATCH /api/llm/models/{slug}\` - Update model (partial updates)
- \`DELETE /api/llm/models/{slug}\` - Delete model

**Providers:**
- \`POST /api/llm/providers\` - Create provider
- \`PATCH /api/llm/providers/{name}\` - Update provider
- \`DELETE /api/llm/providers/{name}\` - Delete provider (blocked if models exist)

### Database Layer (\`backend/server/v2/llm/db_write.py\`)
- Full Prisma CRUD operations with validation
- \`refresh_runtime_caches()\` - Invalidates in-memory registry after mutations
- Proper error handling and logging for audit trail
- FK constraint enforcement (can't delete provider with models)

### Request Models (\`admin_model.py\`)
- \`CreateLlmModelRequest\`, \`UpdateLlmModelRequest\`
- \`CreateLlmProviderRequest\`, \`UpdateLlmProviderRequest\`
- Full Pydantic validation with Field descriptions

## Features
✅ Admin auth required on all endpoints  
✅ Registry cache refresh after every mutation  
✅ Proper validation (400 for bad data, 404 for not found, 500 for errors)  
✅ FK constraint enforcement  
✅ Audit logging  
✅ Based on original PR #11699 implementation  

## Testing
Manual testing scripts provided:
- \`test_admin_api.py\` - Python script with detailed output
- \`test_admin_api.sh\` - Bash/curl script

Requires admin JWT token. Full CRUD flow: create provider → create model → update → delete.

## Stacked PRs
- **PR #12357**: Schema foundation
- **PR #12359**: Registry core (DB layer + cache)
- **PR #12371**: Public read API (GET endpoints)
- **PR #12467** (this): Admin write API (CRUD) ← YOU ARE HERE
- **PR #12468** (next): Admin UI

PR #13611, opened 2026-07-18
feat(backend/api): admin LLM registry API with audit logging

ntindle:
## Why

Part 7 of 9 of the LLM registry restack (#13605→#13606→#13607→#13608→#13609→#13610→this). This is the front door that replaces the "hand-edit a JSON file and open a PR" model-management workflow: adding K3 becomes an authenticated admin call whose row is validated, audited, and propagated to every install.

## What

Admin API at `/api/admin/llm` (all `Security(requires_admin_user)`), re-cut from @Bentlybro's #12467:

- **Kept**: model CRUD (`{slug:path}` params), creator CRUD, usage counting, disable-with-replacement + revert
- **Dropped**: provider CRUD — providers are a fixed seed set (read-only list retained for the UI)
- **`source=LOCAL` on every write** — the importer's admin-wins contract from #13607
- **AuditLog per mutation** (uses the existing `AuditLog` table; entity-typed actions like `LLM_MODEL_DISABLED`, before/after JSON; audit failure never fails the operation)
- **New columns** wired end-to-end: `visibility`, `min_subscription_tier`, `fallback_model_slug` (existence-validated), `kind`
- **New endpoints**: `GET/PUT /routes` (routing-cell upsert; model must exist + be enabled, HIDDEN allowed; response carries capability warnings — e.g. non-reasoning model in a thinking cell); `GET /routes/warnings` (the #13610 refusal records — typo'd LD slugs surface here); `GET /catalog/export` (regenerate the bundled catalog.json with a curl)

## Fixes over the original branch

- Unqualified `"AgentNode"` raw SQL → `{schema_prefix}` helpers (the exact bug class flagged in the original #12359 review; breaks non-`platform` schemas otherwise)
- Latent revert bug: migration wrote provider-stripped node values but revert matched raw slugs — reverts could never have matched a node
- The old branch's private `refresh_runtime_caches` copy replaced by the canonical one (single choke point for cache coherence)

## Verification

- 86/86 tests: full registry suite green + real-DB db_write tests (source=LOCAL assertions, AuditLog rows per action, cell operations, migration/revert round-trip at the SQL layer) + route-layer tests (auth, validation, capability warnings, HIDDEN acceptance, warnings/export endpoints)

Known follow-ups (tickets, not scope): "reset to catalog" action (LOCAL is sticky by design); toggle-with-populated-graph integration test needs graph fixtures.

## Checklist
- [x] All routes admin-gated; acting admin recorded on every mutation via AuditLog

### DIFF
--- a/autogpt_platform/backend/backend/api/features/admin/llm_admin_model.py
+++ b/autogpt_platform/backend/backend/api/features/admin/llm_admin_model.py
@@ -112,6 +112,7 @@ class LlmCreatorAdminResponse(pydantic.BaseModel):
     description: str | None = None
     website_url: str | None = None
     logo_url: str | None = None
+    source: str = "SEED"
     metadata: dict[str, Any] = pydantic.Field(default_factory=dict)
     created_at: str | None = None
     updated_at: str | None = None
@@ -130,6 +131,7 @@ class LlmProviderAdminResponse(pydantic.BaseModel):
     name: str
     display_name: str
     description: str | None = None
+    source: str = "SEED"
     metadata: dict[str, Any] = pydantic.Field(default_factory=dict)
     created_at: str | None = None
     updated_at: str | None = None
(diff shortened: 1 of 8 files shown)


ntindle:

The retirement machinery (incl. the revert-matching fix found here) lives on as the CLI in #13625; the CRUD/audit/routing endpoints have no replacement by design — the PR is the write path, git is the audit.

PR #13612, opened 2026-07-18
feat(frontend): admin LLM registry dashboard (read-only)

ntindle:
## Why

Part 8 of 9 of the LLM registry restack (#13605→#13606→#13607→#13608→#13609→#13610→#13611→this). Admins need eyes on the registry before they get hands: this is the read-only dashboard; editing controls are part 9.

## What

`/admin/llms` behind `withRoleAccess(["admin"])`, re-imagined from @Bentlybro's #12468 under current frontend conventions (generated Orval hooks — the old branch's server-actions pattern is gone entirely):

- **Copilot routing panel**: the 2×2 fast/thinking × standard/advanced matrix from `GET /routes` (fall-through cells shown as such) + the resolution-warnings table (`GET /routes/warnings`) — typo'd LD slugs show up here with reason/layer/count/last-seen
- **Models table** with source (SEED/REMOTE/LOCAL), visibility, enabled badges, tier + context columns; providers, creators, migrations tables
- Design-system components + page-local `SimpleTable` (bots-page precedent — the DS `Table` molecule is a data-entry widget, wrong for read-only display); Skeleton loading, ErrorCard errors; sidebar entry
- `openapi.json` regenerated via the canonical `export-api-schema` + prettier recipe; typed list-response models added to the part-7 endpoints (bare dicts generated untyped clients)

## Verification

- 5 scenario-named integration tests (Vitest+RTL+MSW via Orval-generated handlers): tables render from mocked payloads, warnings panel, error state via 500-mock
- `pnpm format` / `pnpm lint` / `pnpm types` clean; new tests green in the full `test:unit` run (remaining suite failures reproduce **worse** on the base branch without these changes — verified via stash; pre-existing multi-suite interference)
- Backend suites re-verified after the typed-response addition: 86/86

## Checklist
- [x] No new `__legacy__`/`BackendAPI` usages; generated hooks only
- [x] Out-of-scope: typed list responses on part-7 endpoints (small, required by this UI, called out above)

PR #12468, opened 2026-03-18
feat(frontend): Add LLM registry admin UI

Bentlybro:
## Summary
Add admin UI for LLM registry management - **Part 5 of 5** in incremental rollout.

Builds on PR #12467 (admin write API) to provide full visual management interface.

### Admin Page (\`/admin/llms\`)
- Main dashboard at \`/admin/llms\` with role-based access control
- Sections for Providers, Creators, Models, and Active Migrations
- Server-side data fetching with Next.js App Router

### Main Components

**Dashboard (\`LlmRegistryDashboard.tsx\`):**
- Active Migrations section (collapsible, only shows if migrations exist)
- Providers & Creators side-by-side grid
- Models table with recommended model selector
- Error boundary with retry

**Tables:**
- \`ModelsTable.tsx\` - Full model list with edit/delete/toggle actions
- \`ProviderList.tsx\` - Provider management with expand/collapse
- \`CreatorsTable.tsx\` - Creator management
- \`MigrationsTable.tsx\` - Active migration tracking with revert option

**CRUD Modals:**
- Add/Edit/Delete modals for Providers, Models, Creators
- \`DisableModelModal.tsx\` - Toggle model with optional migration
- Form validation with proper error handling

**Other:**
- \`RecommendedModelSelector.tsx\` - Dropdown to set recommended model
- \`actions.ts\` - Server actions calling admin API endpoints

## Features
✅ Admin-only access (role-based)  
✅ Full CRUD operations with modals  
✅ Real-time cache revalidation after mutations  
✅ Migration tracking and revert  
✅ Recommended model management  
✅ Responsive design  
✅ Error boundaries  
✅ Based on original PR #11699 design  

## Design
UI copied directly from original PR #11699 implementation - proven design, fully functional.

API calls will need updating to match new admin endpoint structure.

## Testing
1. Login as admin user
2. Navigate to \`/admin/llms\`
3. Create/edit/delete providers, models, creators
4. Toggle model availability
5. Set recommended model

## Stacked PRs
- **PR #12357**: Schema foundation
- **PR #12359**: Registry core (DB layer + cache)
- **PR #12371**: Public read API (GET endpoints)
- **PR #12467**: Admin write API (CRUD)
- **PR #12468** (this): Admin UI ← YOU ARE HERE

## Next Steps
After this PR merges, the LLM registry will be fully operational for admin management.

Block integration (consuming the registry in LLM blocks) comes in a future PR.

PR #13613, opened 2026-07-18
feat(frontend): admin LLM registry editing

ntindle:
## Why

Part 9 of 9 of Phase A of the LLM registry restack (#13605→#13606→#13607→#13608→#13609→#13610→#13611→#13612→this). With this PR the whole workflow that #13596 tried to do by hand-editing a machine-generated JSON becomes: open /admin/llms, add the model (HIDDEN if pre-launch), set routing cells when ready — validated, audited, propagated to every install within a day.

## What

Editing controls on the read-only dashboard from #13612, re-imagined from @Bentlybro's #12468 modals under current conventions:

- **Models**: add/edit dialogs covering every column (visibility, min subscription tier, fallback-model picker, capability flags, price tier, enabled/recommended); delete and disable flows with the replacement-model picker (pre-filled from `fallbackModelSlug`)
- **Creators**: add/edit/delete
- **Routing matrix**: each copilot cell gets a model dropdown (enabled models; HIDDEN allowed and badged) + clear-cell; the API's capability warnings (e.g. non-reasoning model in a thinking cell) surface inline as warnings per the agreed design — warn, don't block
- **Migrations**: revert with confirmation
- All mutations via generated Orval hooks with proper query-key invalidation and toast errors; no server actions, no legacy components

## Verification

- 12/12 `/admin/llms` integration tests (5 read + 7 new editing scenarios: create/edit request payload assertions, disable-with-replacement, routing PUT + warning surfacing, revert confirm, 500 toast)
- `pnpm format` / `pnpm lint` / `pnpm types` clean; full-suite failures are the pre-existing multi-suite interference documented on #13612 (worse on base without these changes, verified by stash-comparison)

## Checklist
- [x] Design-system components only; generated hooks only

### DIFF
--- a/autogpt_platform/frontend/src/app/(platform)/admin/llms/__tests__/dashboard.test.tsx
+++ b/autogpt_platform/frontend/src/app/(platform)/admin/llms/__tests__/dashboard.test.tsx
@@ -148,8 +148,8 @@ describe("LlmRegistryDashboard", () => {
 
     const cells = await screen.findAllByText("moonshotai/kimi-k3");
     expect(cells.length).toBeGreaterThanOrEqual(1);
-    // Only thinking.standard is set; the other three cells fall through.
-    expect(screen.getAllByText("— (falls through)").length).toBe(3);
+    // Only thinking.standard is set; the other three editable cells fall through.
+    expect(screen.getAllByText("— falls through —").length).toBe(3);
   });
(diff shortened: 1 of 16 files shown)


ntindle:

No replacement by design — editing happens in the catalog file by PR.

PR #13614, opened 2026-07-18
fix(frontend): keep internal tool output out of the artifacts toggle

### LINKED ISSUE #13510: AutoPilot UI auto-opens internal tool output files (toolu_*.json) in artifact panel
## Problem

When AutoPilot generates intermediate/internal files such as `toolu_{id}.json` (tool call output files written implicitly during execution), the artifact panel auto-opens to display them. These files have no user-facing value and should not trigger the panel.

In half-screen layouts this is particularly disruptive: the panel opening squishes the AutoPilot chat to half width, degrading the conversation experience for a file the user never asked to see.

## Expected Behavior

The artifact panel should only auto-open for files that are meaningful user-facing outputs. Internally-generated intermediate files (e.g. `toolu_*.json`) should be silently ignored by the auto-open trigger.

## Suggested Fix

Filter out files matching patterns like `toolu_*.json` (or more broadly, any file written implicitly by the tool execution harness rather than explicitly by the agent as a user-facing output) from the artifact panel's auto-open logic.

## Notes

* Not a breaking bug, but a noticeable UX regression in split/half-screen layouts
* Related context: SECRT-2263 (artifact panel auto-opens on new artifact) — the auto-open behavior introduced there doesn't currently distinguish user-facing artifacts from internal intermediate files

Yigtwxx:
### Why / What / How

**Why:** Closes #13510. When AutoPilot runs, Agent SDK tool-result files end up in the session workspace: `_persist_and_summarize` in `backend/copilot/tools/base.py` parks every oversized tool result at `tool-outputs/{tool_call_id}.json`, and `WorkspaceManager.write_file` session-scopes it, so they list as `/sessions/{session_id}/tool-outputs/{tool_call_id}.json`. The copilot UI treats them as ordinary generated output, so internal JSON the user never asked for becomes the session's headline document.

The UI can't currently tell the difference: `_derive_origin` in the workspace routes classifies everything not explicitly stamped `user-upload` as `generated`, so an SDK tool result and a real deliverable are indistinguishable by the time the frontend sees them.

**What:** The artifacts toggle no longer treats internal tool output as the session's latest document. It never wears a tool result's name, and never opens one on click; a session whose only generated files are tool scratch keeps the bare icon.

**How:** A new `isInternalToolOutput` predicate sits next to the existing `isUploadedFile` in the FilesTab helpers, and `ContextPanelToggle`'s `getLastGeneratedFile` filters the `generated` list through it before picking the newest file. It matches the one directory `_persist_and_summarize` owns, anchored at the session root:

```ts
const SDK_TOOL_RESULT_PATH =
  /^\/sessions\/[^/]+\/tool-(?:results|outputs)\/[^/]+$/i;
```

**The id is deliberately not part of the match.** An earlier revision borrowed the `(toolu|mcp)_` prefix from the backend's `_SDK_TOOL_RESULT_RE`, on the theory that this was parity. It was not, in either direction:

- That classifier scans **free model text**, where the prefix is doing token-boundary work. Here the input is always a server-resolved workspace path, so the directory segment alone is both sufficient and safe.
- Worse, the prefix does not describe the ids the platform actually mints. `_execute_tool_sync` synthesizes `effective_id = f"sdk-{uuid.uuid4().hex[:12]}"` (`sdk/tool_adapter.py:322`) rather than forwarding the SDK's `toolu_*`, and `_baseline_tool_executor` forwards the provider's `tool_call.id` unchanged (`baseline/service.py:1050`), which is `call_*` for OpenAI-family models. Only Anthropic-via-baseline ever produced `toolu_*`, so on the SDK transports this PR is about the filter was a no-op.

The session-root anchor is what carries the false-positive guard: a real deliverable under a user's own `my-pipeline/tool-outputs/` folder stays eligible because of its depth, not because of its name. (The previous `(?:^|\/)` matched any interior slash, so that exemption never actually held — a point [raised in review](https://github.com/Significant-Gravitas/AutoGPT/pull/13614#discussion_r3900883081) and now pinned by a test that distinguishes the two cases.)

Uploaded files short-circuit to `false` regardless, so a user who deliberately uploads a file into that shape still gets the normal behaviour.

**Scope note:** this deliberately changes which file the toggle points at, matching the issue's wording. These files still appear in the workspace file listing and remain downloadable — hiding them from the list too is a larger behavioural change that I didn't want to fold in uninvited.

### Rebased onto the new copilot chat (#14239)

This PR originally patched `useAutoOpenForFiles.ts`, since the artifact panel used to auto-open on the newest generated file. #14239 deleted that hook: the panel no longer auto-opens at all, and `getLastGeneratedFile` moved into `ContextPanelToggle`, where it labels the top-right artifacts button and is that button's click target.

That relocates the bug rather than fixing it — an SDK tool result is still the newest generated file, so the button would wear `sdk-a1b2c3d4e5f6.json` and open internal JSON on click. So the predicate moved with the selection logic, and the integration coverage moved from `ContextPanelAutoOpen.test.tsx` (restored to its `dev` state) to `ContextPanelToggle.test.tsx`.

**Relationship to #13799:** the two cover disjoint writers and want different outcomes, so neither subsumes the other. #13799 stamps `metadata={"internal": true}` in `write_agent_json_to_workspace` and hides the agent-JSON handoff from the file listing entirely; this PR covers `_persist_and_summarize`'s tool outputs, which carry no metadata, and only demotes them from being the session's headline document. Discussed in more detail [in this thread](https://github.com/Significant-Gravitas/AutoGPT/pull/13614#discussion_r3775808734).

**Suggested follow-up (not in this PR):** the cleaner fix is at the source — stamping an origin on write so `_derive_origin` can return a third value and every consumer can filter on a contract field instead of pattern-matching paths. That would let this regex and #13799's `isInternalFile` both be deleted. It's an API schema change, so it seemed better as its own PR. Happy to take it on once #13799 lands.

### Changes 🏗️

- Add `isInternalToolOutput` to `ContextPanel/components/FilesTab/helpers.ts`, matching `/sessions/{id}/tool-(results|outputs)/{file}` on the file path — the directory `_persist_and_summarize` owns, anchored at the session root, with no id-prefix requirement
- Filter `generated` through that predicate in `ContextPanelToggle`'s `getLastGeneratedFile`, so internal tool output is neither the button's label nor its click target
- Unit tests for the predicate covering the ids the platform really mints (`sdk-*` from the SDK transport, `call_*` and `tc-*` from the baseline transport, alongside `toolu_*`/`mcp_*`), plus a negative block pinning the session-root bound: `mcp_config.json` at the root, `my-pipeline/tool-outputs/data.json`, `my-pipeline/tool-outputs/toolu_mine.json`, and a nested file inside the real directory
- Integration tests in `ContextPanelToggle.test.tsx`: the toggle wears the newest *user-facing* file rather than a newer tool output and opens it on click, a deliverable whose name looks like an SDK id is still surfaced, a deliverable under the user's own `tool-outputs/` folder is still surfaced, and a tool-output-only session leaves the button unlabeled and falls back to the artifacts tab

No configuration, dependency, or API schema changes.

#### For code changes:
- [x] I have made a test plan
  - [x] Wrote the toggle tests first and confirmed they fail against the unfiltered `getLastGeneratedFile` — the button wore a tool result instead of `result.csv`, and the tool-output-only session labeled the button with tool scratch instead of keeping the bare icon
  - [x] Re-falsified after the regex change: reverting only `SDK_TOOL_RESULT_PATH` to the earlier `(toolu|mcp)_` form fails 4 tests, including the integration case where the `sdk-*` file becomes the button's label — the review finding, reproduced as a test
  - [x] Whole `ContextPanel` directory: 31/31 pass
  - [x] Verified the id shapes against the backend rather than assuming them: `_execute_tool_sync` synthesizes `sdk-<uuid>`, `_baseline_tool_executor` forwards the provider's id, and `_persist_and_summarize` names the file after whichever it is handed. Two earlier revisions of this PR guessed at this — first that the files land at the session root, then that they carry an SDK id prefix — and both guesses are what the current directory-only rule removes
  - [x] `tsc --noEmit` clean, ESLint clean on the changed files, Prettier reports no changes


Yigtwxx:
Gentle ping — this has been sitting for a few days and is ready whenever you have time.

Status: CI is green (lint, integration, e2e, CodeQL, API types, no conflicts), codecov shows the patch fully covered, and CodeRabbit reported no actionable comments. The only red mark is the Vercel deployment (`Authorization required to deploy`), which is the usual fork-PR limitation and not something I can act on from my side.

No rush — just flagging it in case it slipped through. Happy to rebase or adjust anything if the approach needs changing.

Yigtwxx:
Pushed `65f757f8` addressing both **Should Fix** items.

**1 — Directory regex over-match / inaccurate "mirrors the backend" comment.** Confirmed against the backend before touching anything, and I took the *tighten the regex* option rather than the *reword the comment* one. The deciding evidence: nothing in the backend writes a `tool-results/` or `tool-outputs/` directory into the workspace — that segment only ever appears in host-side SDK paths, which is why `_SDK_TOOL_RESULT_RE` anchors on `.claude/projects/<cwd>/<uuid>/`. So the loose branch wasn't buying coverage of any real case; it was pure false-positive surface.

```diff
-const SDK_TOOL_RESULT_DIR = /(?:^|\/)tool-(?:results|outputs)\//i;
+const SDK_TOOL_RESULT_PATH =
+  /(?:^|\/)\.claude\/projects\/[^/]+\/[^/]+\/tool-(?:results|outputs)\//i;
```

The comment no longer claims parity. It now records the one place the frontend is deliberately broader — the filename branch matches `toolu_*.json` / `mcp_*.json` on the name alone, without the backend's `tool-(results|outputs)/` parent, because in the workspace these land directly under the session root — and why a bare `tool-outputs/` segment is not sufficient on its own.

**2 — Coverage for the divergent/uncovered branches.** All three sub-items:
- New negative unit test pinning the intended behaviour for finding #1: a generated file at `my-pipeline/tool-outputs/data.json` is *not* internal.
- The `mcp_` and SDK-path branches now have integration coverage too — `withToolOutputAndRealFile()` carries one entry per branch, each with a newer `created_at` than the real deliverable, so a missing filter surfaces the wrong file.
- Added the `isOpen === true` assertion to the "skips a newer internal tool output" test.

I also falsified the new test rather than just watching it pass: with the old loose regex restored, exactly 1 of 10 fails (the new negative case), and 20/20 pass on the fix. `tsc --noEmit`, ESLint, and Prettier are clean on the touched files.

**Not taken:** the single-pass `getLastGeneratedFile` (Nice to Have #1). Agreed it's negligible at realistic `n`, and it widens the diff without changing behaviour.

**Follow-up:** left a note on #13510 offering the `metadata.origin = "tool-output"` write-time stamp as a separate PR. That's the durable fix — it deletes both frontend regexes and this whole class of drift. Happy to pick it up on a maintainer's word.

autogpt_platform/frontend/src/app/(platform)/copilot/components/ContextPanel/components/FilesTab/helpers.ts
```diff
@@ -11,6 +11,24 @@ export function isUploadedFile(item: WorkspaceFileItem): boolean {
...
 }
 
+// Agent SDK tool results that leak into the workspace. Both shapes come from
+// the backend classifier (`_SDK_TOOL_RESULT_RE` in backend/copilot/context.py),
+// except the name branch drops that regex's `tool-(results|outputs)/` prefix:
+// in the workspace these files land directly under the session root.
+// A bare `tool-outputs/` directory is deliberately not enough on its own — the
+// backend guards against the same false positive on user-authored paths like
+// `my-pipeline/tool-outputs/data.json`.
+const SDK_TOOL_RESULT_NAME = /^(?:toolu|mcp)_[\w-]+\.json$/i;
```
kcze: 🤖 🟠 **Should Fix**: this branch drops the backend rule's mandatory directory prefix, so it is strictly broader than the classifier it says it mirrors. `_SDK_TOOL_RESULT_RE` (`backend/copilot/context.py:173-178`) only treats `(toolu|mcp)_<id>.json` as SDK-internal when it sits under `tool-results/` or `tool-outputs/` — that prefix *is* the false-positive guard the comment above credits it with. Matching on the bare name at the session root removes it.

An agent that writes a real deliverable named `mcp_config.json` or `mcp_servers.json` (plausible while building an MCP integration; the `/i` flag widens it further) silently never auto-opens, and if it is the session's only generated file the panel stays collapsed with no indication why. The negative tests at `helpers.test.ts:88-89` only cover the no-underscore forms (`toolusage.json`, `mcpserver.json`), so this whole class is untested. _(flagged by: Claude + Codex)_

autogpt_platform/frontend/src/app/(platform)/copilot/components/ContextPanel/useAutoOpenForFiles.ts
```diff
@@ -2,15 +2,21 @@
...
+} from "./components/FilesTab/helpers";
 import {
   useSessionFiles,
   type SessionFile,
 } from "./components/FilesTab/useSessionFiles";
 
 function getLastGeneratedFile(generated: SessionFile[]): SessionFile | null {
-  if (generated.length === 0) return null;
-  return generated.reduce((latest, file) =>
+  const userFacing = generated.filter(
```
kcze: 🤖 🟡 **Nice to Have**: worth reconciling with #13799 before both land. That PR stamps `metadata={"internal": true}` at write time in `copilot/tools/agent_json_input.py` and filters via a new `isInternalFile` inside `useSessionFiles` — one choke point that hides the file from the Files tab, `filesCount`, download-all and auto-open at once. This PR pattern-matches filenames and filters only inside `getLastGeneratedFile`.

The two match disjoint file sets so there is no functional conflict, but afterwards `helpers.ts` carries two near-identically-named "is internal" predicates with different semantics at different layers, and both PRs insert their helper immediately after `isUploadedFile` — so they also conflict textually on merge. A `toolu_*.json` file ends up invisible to auto-open yet still listed in the Files tab and counted in `ContextPanel.tsx:36`.

Your own description already names metadata-at-source as the cleaner follow-up; picking one mechanism now avoids the two rules drifting. _(flagged by: Claude)_

autogpt_platform/frontend/src/app/(platform)/copilot/components/ContextPanel/components/FilesTab/helpers.ts
Yigtwxx: Fixed in fca8426 — and digging into it showed the premise behind the bare-name branch was simply wrong.

`_persist_and_summarize` (`backend/copilot/tools/base.py:69`) writes every oversized tool result to `tool-outputs/{tool_call_id}.json`, and `WorkspaceManager.write_file` session-scopes that, so these files surface in the listing as `/sessions/<sid>/tool-outputs/toolu_*.json`. They never land at the session root — the comment I'd written claiming they did was the reason the directory requirement got dropped in the first place. That also makes the `.claude/projects/<cwd>/<uuid>/tool-(results|outputs)/` branch dead code for workspace files, since those live host-side and are never listed.

So the whole predicate collapses to one path match, which is genuine parity with the shorthand branch of `_SDK_TOOL_RESULT_RE` rather than a claim of it:

```ts
const SDK_TOOL_RESULT_PATH =
  /(?:^|\/)tool-(?:results|outputs)\/(?:toolu|mcp)_[\w-]+\.json$/i;
```

`mcp_config.json` and `mcp_servers.json` at the session root now stay eligible for auto-open, as does a user deliverable under their own `my-pipeline/tool-outputs/`. Both halves of the shape are pinned by a new negative test in `helpers.test.ts`, and the integration fixtures moved to the real layout plus a case asserting an SDK-lookalike deliverable still opens the panel. 21/21 green locally, `tsc --noEmit` and eslint clean.

Yigtwxx: Correction on the numbers in my previous reply: the two suites this PR touches are **13/13** (`helpers.test.ts` 9, `ContextPanelAutoOpen.test.tsx` 4). The **20/20** figure is the whole `ContextPanel` directory. There is no 21st test — I mistyped.

autogpt_platform/frontend/src/app/(platform)/copilot/components/ContextPanel/useAutoOpenForFiles.ts
Yigtwxx: Read through #13799. My read is that the two are complementary rather than competing, so I'd rather not fold this one into it — happy to be overruled if you see it differently.

They cover disjoint writers: #13799 stamps `metadata={"internal": true}` inside `write_agent_json_to_workspace`, which only covers the agent-JSON handoff. The files this PR is about come from `_persist_and_summarize` in `copilot/tools/base.py` — every oversized tool result, from any tool — and carry no metadata at all. Landing #13799 alone leaves `tool-outputs/toolu_*.json` still hijacking the panel; landing this alone leaves the agent-JSON handoff still in the Files tab. Neither subsumes the other.

They also want different outcomes. #13799 hides its files outright — right call for a handoff artifact the user has no reason to see. Tool outputs are genuine records the model produced and a user may want to download; the ask in #13510 was only that they stop stealing the panel, so this PR scopes itself to `getLastGeneratedFile` and leaves them listed.

Concrete follow-up if you want one mechanism: extend #13799's stamp to `base.py`'s write and give `_derive_origin` a third value, then both this regex and #13799's `isInternalFile` name-match can go, with `useSessionFiles` deciding hide-vs-demote per origin. That's a backend contract change and reads as its own PR to me rather than something to graft onto either of these. I'm glad to open it once #13799 lands.

On the textual conflict — both helpers insert right after `isUploadedFile` in `helpers.ts`. Whichever merges second, I'll rebase; it's a two-line resolution.

autogpt_platform/frontend/src/app/(platform)/copilot/components/ContextPanel/ContextPanelToggle.tsx
```diff
@@ -17,8 +20,11 @@ interface Props {
 }
 
 function getLastGeneratedFile(generated: SessionFile[]): SessionFile | null {
-  if (generated.length === 0) return null;
-  return generated.reduce((latest, file) =>
+  const userFacing = generated.filter(
+    (file) => !isInternalToolOutput(file.item),
+  );
+  if (userFacing.length === 0) return null;
```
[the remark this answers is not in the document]
Yigtwxx: The mechanism is real — `closeArtifactPanel` does remember the closed preview (`store.ts:374`, `lastArtifact: state.artifactPanel.activeArtifact`), and neither the workspace file card nor the artifacts tab filters internal tool output out of its listing, so one can end up in `lastArtifact`. But I don't think it should be suppressed, for two reasons.

**It is never an automatic promotion.** Every path that sets `activeArtifact` is an explicit user click: `ArtifactCard.tsx:106`, `useWorkspaceFileCards.ts:50`, `ArtifactsTab.tsx:79`, and the toggle itself. The one path that could fire on its own is inert — `registerArtifactForAutoOpen` opens `store.ts:493` with *"Auto-open is disabled — the drawer only opens on explicit click"* and only upgrades metadata for an id it already knows. And `lastArtifact` is cleared on session entry (`clearLastArtifact`, pinned by `ContextPanelAutoOpen.test.tsx`'s "forgets the previous chat's artifact"), so it can only ever hold a file the user opened in *this* session.

So the toggle reopening it isn't the panel promoting scratch to headline document — the thing this PR is about — it's the toggle restoring the document the user was reading a moment ago, which is what the button is documented to do (`ContextPanelToggle.test.tsx`: "sidebar toggle restores the last previewed artifact instead of opening the tabs view"). That principle is already in the predicate: `isInternalToolOutput` short-circuits to `false` for uploaded files, because an explicit user choice outranks the heuristic. Sending them to the tabs view instead would cost an extra click to get back to what they were just looking at, and only for this one file type.

**The suggested condition is broader than the hole.** Keying on "`getLastGeneratedFile` returned null" discards `lastArtifact` whatever it is — including a legitimate upload or a message artifact the user previewed — for any session whose generated files all happen to be internal. That regresses the restore behaviour for files this PR has no quarrel with.

If you do want the stronger invariant (the toggle never points at a tool result, full stop), the precise version tests identity rather than emptiness. `ArtifactRef` carries no `path`, so the predicate can't run on `lastArtifact` directly; it'd be an id cross-check:

```ts
const internalIds = new Set(
  generated.filter((f) => isInternalToolOutput(f.item)).map((f) => f.item.id),
);
const remembered =
  lastArtifact && !internalIds.has(lastArtifact.id) ? lastArtifact : null;
```

That closes the case above without touching the upload/message-artifact case. Happy to add it with a test if you'd prefer that — just say so and I'll push it.

autogpt_platform/frontend/src/app/(platform)/copilot/components/ContextPanel/components/FilesTab/helpers.ts
```diff
@@ -11,6 +11,22 @@ export function isUploadedFile(item: WorkspaceFileItem): boolean {
...
+// Agent SDK tool results that leak into the workspace. `_persist_and_summarize`
+// (backend/copilot/tools/base.py) parks every oversized tool result at
+// `tool-outputs/<tool_call_id>.json` within the session, so the directory
+// segment and the SDK's id prefix together are the whole shape — the same pair
+// the backend classifier requires (`_SDK_TOOL_RESULT_RE` in
+// backend/copilot/context.py). Demanding both keeps a deliverable merely named
+// `mcp_config.json`, or one written under a user's own `tool-outputs/` folder,
+// eligible to stand as the session's latest document.
+const SDK_TOOL_RESULT_PATH =
+  /(?:^|\/)tool-(?:results|outputs)\/(?:toolu|mcp)_[\w-]+\.json$/i;
```
kcze: 🤖 🔴 **Blocker**: the `(?:toolu|mcp)_` half doesn't match the ids the platform actually mints, so this filter no-ops on the default transport.

`_persist_and_summarize` names the file after whatever `tool_call_id` it is handed (`backend/copilot/tools/base.py:70`). The only route into `BaseTool.execute` under an SDK-capable transport is `_execute_tool_sync`, which **synthesizes** the id rather than passing the SDK's `toolu_*` through:

```python
# backend/copilot/sdk/tool_adapter.py:322
effective_id = f"sdk-{uuid.uuid4().hex[:12]}"
```

So on `subscription` (`supports_sdk=True`, the AutoPilot path this PR is about) the file lists as `/sessions/<sid>/tool-outputs/sdk-a1b2c3d4e5f6.json` — which this regex rejects. The toggle still reads `Open sdk-a1b2c3d4e5f6.json` and still opens raw tool JSON on click. Only the baseline transport on an Anthropic model ever yields `toolu_*`; the backend's own fixture uses `tool-outputs/tc-123.json` (`tools/base_test.py:70`), which also fails.

The prefix requirement is borrowed from a classifier that scans **free model text**, where it has to guess at boundaries. Here the input is always a server-resolved workspace path, so the directory segment alone is both sufficient and safe if you anchor it at the session root:

```ts
const SDK_TOOL_RESULT_PATH = /^\/sessions\/[^/]+\/tool-(?:results|outputs)\/[^/]+$/i;
```

That still leaves `my-pipeline/tool-outputs/data.json` eligible, and catches every id shape. _(flagged by: Claude)_

autogpt_platform/frontend/src/app/(platform)/copilot/components/ContextPanel/components/FilesTab/helpers.ts
```diff
@@ -11,6 +11,22 @@ export function isUploadedFile(item: WorkspaceFileItem): boolean {
   return item.origin === "uploaded";
 }
 
+// Agent SDK tool results that leak into the workspace. `_persist_and_summarize`
+// (backend/copilot/tools/base.py) parks every oversized tool result at
+// `tool-outputs/<tool_call_id>.json` within the session, so the directory
+// segment and the SDK's id prefix together are the whole shape — the same pair
+// the backend classifier requires (`_SDK_TOOL_RESULT_RE` in
+// backend/copilot/context.py). Demanding both keeps a deliverable merely named
+// `mcp_config.json`, or one written under a user's own `tool-outputs/` folder,
```
kcze: 🤖 🟡 **Nice to Have**: this claim is inverted — a file under a user's own `tool-outputs/` folder is *not* kept eligible. `(?:^|\/)` matches any interior slash, so `/sessions/s1/my-pipeline/tool-outputs/toolu_mine.json` returns `true`.

The directory guard carries no depth constraint at all; the only thing saving `.../my-pipeline/tool-outputs/data.json` is its missing id prefix — which is also why the `helpers.test.ts` "needs both halves" block can't actually distinguish the two halves (both its `my-pipeline` and `tool-results` cases fail on the prefix alone). Same claim appears in the PR description. _(flagged by: Claude)_

autogpt_platform/frontend/src/app/(platform)/copilot/components/ContextPanel/components/FilesTab/helpers.ts
Yigtwxx: Confirmed and fixed in d52d27c — the prefix branch never fired on the transports this PR is about.

Traced it end to end. `_execute_tool_sync` synthesizes `effective_id = f"sdk-{uuid.uuid4().hex[:12]}"` (`sdk/tool_adapter.py:322`) and passes it as `tool_call_id` into `BaseTool.execute`, which hands it to `_persist_and_summarize` (`tools/base.py:241`), which names the file after it — so the SDK path parks `/sessions/<id>/tool-outputs/sdk-a1b2c3d4e5f6.json`. The baseline path is no better: `_baseline_tool_executor` reads `tool_call_id = tool_call.id` from the provider (`baseline/service.py:1050`) and forwards it unchanged, so an OpenAI-family model yields `call_*`. Only Anthropic-via-baseline ever produced `toolu_*` — which is exactly the one case my fixtures covered.

Took your regex, keeping `tool-results` since the backend classifier names both directories:

```ts
const SDK_TOOL_RESULT_PATH =
  /^\/sessions\/[^/]+\/tool-(?:results|outputs)\/[^/]+$/i;
```

Your framing is the part I had wrong. `_SDK_TOOL_RESULT_RE` scans free model text, where the prefix is doing token-boundary work; here the input is always a server-resolved workspace path, and `/sessions/<id>/tool-outputs/` has exactly one writer. So the directory *is* the shape. I dropped the `.json` requirement for the same reason — nothing user-facing writes into that directory, so the extension carries no signal.

Coverage now pins the ids the platform actually mints — `sdk-*`, `call_*`, and `tc-*` (that last one matching your `tools/base_test.py:70` fixture) — at both the unit and the integration level, plus negatives for `/sessions/s1/my-pipeline/tool-outputs/toolu_mine.json` and for a nested file inside the real directory. Reverting just the regex fails 4 tests, including the integration one where the button wears the tool result. 31/31 green in `ContextPanel`, `tsc --noEmit` and eslint clean.

autogpt_platform/frontend/src/app/(platform)/copilot/components/ContextPanel/components/FilesTab/helpers.ts
Yigtwxx: Correct — `(?:^|\/)` matches any interior slash, so `/sessions/s1/my-pipeline/tool-outputs/toolu_mine.json` returned `true` and the exemption I claimed was carried entirely by the id prefix, not by the directory guard. And your point about the "needs both halves" block lands: both of its cases failed on the prefix alone, so neither ever exercised the directory half it was supposed to pin.

Fixed by the same change as the blocker above — the depth anchor in `/^\/sessions\/[^/]+\/…/` is now what provides the exemption, so it is structural rather than incidental. The renamed test ("bounds the match to a tool-output dir at the session root") asserts it directly, using `toolu_mine.json` under a user folder as the case that actually distinguishes the halves, and `ContextPanelToggle.test.tsx` gained an integration case where a deliverable at `my-pipeline/tool-outputs/data.json` still labels the button. PR description updated to match.

autogpt_platform/frontend/src/app/(platform)/copilot/components/ContextPanel/components/FilesTab/helpers.ts
```diff
@@ -11,6 +11,19 @@ export function isUploadedFile(item: WorkspaceFileItem): boolean {
   return item.origin === "uploaded";
 }
 
+// Agent SDK tool results that leak into the workspace. Mirrors the backend
```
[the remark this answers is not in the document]
Yigtwxx: 🤖 Fixed in [d52d27cca](https://github.com/Significant-Gravitas/AutoGPT/commit/d52d27cca239b79ff2bebd43fa9996cfda400d2a): the comment no longer claims parity with `_SDK_TOOL_RESULT_RE`. It now names the actual writer — `_persist_and_summarize` (`backend/copilot/tools/base.py`) via `WorkspaceManager.write_file` — and explains why the directory segment at the session root is the whole shape and why no id prefix is part of the match.

autogpt_platform/frontend/src/app/(platform)/copilot/components/ContextPanel/components/FilesTab/helpers.ts
```diff
@@ -11,6 +11,19 @@ export function isUploadedFile(item: WorkspaceFileItem): boolean {
...
+// classifier (`_SDK_TOOL_RESULT_RE` in backend/copilot/context.py) so both ends
+// agree on what counts as an internal file.
+const SDK_TOOL_RESULT_NAME = /^(?:toolu|mcp)_[\w-]+\.json$/i;
+const SDK_TOOL_RESULT_DIR = /(?:^|\/)tool-(?:results|outputs)\//i;
```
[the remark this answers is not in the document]
Yigtwxx: 🤖 Fixed in [d52d27cca](https://github.com/Significant-Gravitas/AutoGPT/commit/d52d27cca239b79ff2bebd43fa9996cfda400d2a): the regex is anchored at the session root (`/^\/sessions\/[^/]+\/tool-(?:results|outputs)\/[^/]+$/i`), so a deliverable under a user's own folder like `/sessions/s1/my-pipeline/tool-outputs/data.json` stays eligible. Pinned by a unit negative in `helpers.test.ts` and an integration case in `ContextPanelToggle.test.tsx` ("still wears a deliverable under the user's own tool-outputs folder").

autogpt_platform/frontend/src/app/(platform)/copilot/components/ContextPanel/useAutoOpenForFiles.ts
```diff
@@ -2,15 +2,21 @@
...
 import { useCopilotUIStore } from "../../store";
-import { fileItemToArtifactRef } from "./components/FilesTab/helpers";
+import {
+  fileItemToArtifactRef,
+  isInternalToolOutput,
+} from "./components/FilesTab/helpers";
 import {
   useSessionFiles,
   type SessionFile,
 } from "./components/FilesTab/useSessionFiles";
```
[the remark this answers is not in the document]
Yigtwxx: 🤖 Fixed in [7a082875e](https://github.com/Significant-Gravitas/AutoGPT/commit/7a082875eb7bfc73fb796211ad7a88d4e7cd6032): `getLastGeneratedFile` (now in `ContextPanelToggle.tsx`) is a single loop that caches the running latest timestamp instead of filter → reduce with a `Date` re-parse per step.

autogpt_platform/frontend/src/app/(platform)/copilot/components/ContextPanel/components/FilesTab/helpers.ts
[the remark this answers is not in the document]
Yigtwxx: 🤖 Fixed in [d52d27cca](https://github.com/Significant-Gravitas/AutoGPT/commit/d52d27cca239b79ff2bebd43fa9996cfda400d2a): `helpers.test.ts` now asserts `isInternalToolOutput` returns `false` for `/sessions/s1/my-pipeline/tool-outputs/data.json` (and for `toolu_mine.json` under the same folder), and the regex is bounded to the session root so those cases fail structurally rather than on a missing id prefix.

autogpt_platform/frontend/src/app/(platform)/copilot/components/ContextPanel/__tests__/ContextPanelAutoOpen.test.tsx
```diff
@@ -27,6 +36,29 @@ function withFiles(): ListFilesResponse {
...
+    origin: "generated",
+    created_at: "2026-05-20T12:00:00Z",
+  });
+  return response;
+}
+
+function withOnlyToolOutput(): ListFilesResponse {
+  const response = withToolOutputAndRealFile();
+  response.files = response.files.filter((file) =>
+    file.name.startsWith("toolu_"),
```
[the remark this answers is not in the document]
Yigtwxx: 🤖 Fixed in [d52d27cca](https://github.com/Significant-Gravitas/AutoGPT/commit/d52d27cca239b79ff2bebd43fa9996cfda400d2a): the predicate no longer has separate name/directory branches — the session-root directory is the only rule — and the integration fixture in `ContextPanelToggle.test.tsx` runs `toolu_*`, `mcp_*`, `tool-results/` and the platform's real `sdk-*` id through the toggle end to end.

autogpt_platform/frontend/src/app/(platform)/copilot/components/ContextPanel/__tests__/ContextPanelAutoOpen.test.tsx
```diff
@@ -49,4 +81,33 @@ describe("ContextPanelAutoOpen", () => {
     );
     expect(useCopilotUIStore.getState().artifactPanel.isOpen).toBe(true);
   });
+
+  test("skips a newer internal tool output and opens the real generated file", async () => {
```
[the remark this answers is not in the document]
Yigtwxx: 🤖 Fixed in [7a082875e](https://github.com/Significant-Gravitas/AutoGPT/commit/7a082875eb7bfc73fb796211ad7a88d4e7cd6032): the toggle test for a newer tool output ("wears the newest user-facing file, not a newer tool output") now asserts `panelState().isOpen` is `true` in addition to the active artifact id.

autogpt_platform/frontend/src/app/(platform)/copilot/components/ContextPanel/components/FilesTab/helpers.ts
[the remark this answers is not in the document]
Yigtwxx: 🤖 Fixed in [d52d27cca](https://github.com/Significant-Gravitas/AutoGPT/commit/d52d27cca239b79ff2bebd43fa9996cfda400d2a): the directory branch was replaced by a single session-root-anchored pattern, `/^\/sessions\/[^/]+\/tool-(?:results|outputs)\/[^/]+$/i`. `my-pipeline/tool-outputs/data.json` stays eligible for the toggle, and the comment no longer describes this as a mirror of `_SDK_TOOL_RESULT_RE` — it documents the backend writer the path comes from instead. Negative tests cover the user-folder case at unit and integration level.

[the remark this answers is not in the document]
Yigtwxx: 🤖 Fixed in [d52d27cca](https://github.com/Significant-Gravitas/AutoGPT/commit/d52d27cca239b79ff2bebd43fa9996cfda400d2a): the directory branch was replaced by a single session-root-anchored pattern, `/^\/sessions\/[^/]+\/tool-(?:results|outputs)\/[^/]+$/i`. `my-pipeline/tool-outputs/data.json` stays eligible for the toggle, and the comment no longer describes this as a mirror of `_SDK_TOOL_RESULT_RE` — it documents the backend writer the path comes from instead. Negative tests cover the user-folder case at unit and integration level.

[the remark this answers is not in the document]
Yigtwxx: 🤖 Fixed in [d52d27cca](https://github.com/Significant-Gravitas/AutoGPT/commit/d52d27cca239b79ff2bebd43fa9996cfda400d2a): the directory branch was replaced by a single session-root-anchored pattern, `/^\/sessions\/[^/]+\/tool-(?:results|outputs)\/[^/]+$/i`. `my-pipeline/tool-outputs/data.json` stays eligible for the toggle, and the comment no longer describes this as a mirror of `_SDK_TOOL_RESULT_RE` — it documents the backend writer the path comes from instead. Negative tests cover the user-folder case at unit and integration level.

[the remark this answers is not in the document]
Yigtwxx: 🤖 Fixed in [d52d27cca](https://github.com/Significant-Gravitas/AutoGPT/commit/d52d27cca239b79ff2bebd43fa9996cfda400d2a): the directory branch was replaced by a single session-root-anchored pattern, `/^\/sessions\/[^/]+\/tool-(?:results|outputs)\/[^/]+$/i`. `my-pipeline/tool-outputs/data.json` stays eligible for the toggle, and the comment no longer describes this as a mirror of `_SDK_TOOL_RESULT_RE` — it documents the backend writer the path comes from instead. Negative tests cover the user-folder case at unit and integration level.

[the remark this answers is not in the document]
Yigtwxx: 🤖 Fixed in [d52d27cca](https://github.com/Significant-Gravitas/AutoGPT/commit/d52d27cca239b79ff2bebd43fa9996cfda400d2a): the directory branch was replaced by a single session-root-anchored pattern, `/^\/sessions\/[^/]+\/tool-(?:results|outputs)\/[^/]+$/i`. `my-pipeline/tool-outputs/data.json` stays eligible for the toggle, and the comment no longer describes this as a mirror of `_SDK_TOOL_RESULT_RE` — it documents the backend writer the path comes from instead. Negative tests cover the user-folder case at unit and integration level.

0ubbe (approved):
Nice thank you! 🤩

autogpt_platform/frontend/src/app/(platform)/copilot/components/ContextPanel/components/FilesTab/helpers.ts
ntindle: Fixed in ea9e7b79e3. New outputs carry purpose metadata at write time. The legacy fallback is anchored to the tool-output/results directory directly under the session root; names such as mcp_config.json at the root remain deliverables. Regression tests cover both cases.

autogpt_platform/frontend/src/app/(platform)/copilot/components/ContextPanel/useAutoOpenForFiles.ts
ntindle: Reworked in ea9e7b79e3 around writer provenance and a shared useSessionFiles deliverables classification, which feeds both latest-document selection and artifact previews. This deliberately preserves internal files in the file drawer, including preview and download, instead of hiding them from every surface. The overlapping agent-JSON PR is a separate unmerged concept and is not pulled into this change.

autogpt_platform/frontend/src/app/(platform)/copilot/components/ContextPanel/components/FilesTab/helpers.ts
ntindle: Fixed in ea9e7b79e3: the writer stamps metadata.purpose=tool-output independently of the tool call ID format. Legacy classification is directory-based, not tied to toolu/mcp IDs. Tests explicitly cover sdk-..., call_..., tc-... and existing SDK-shaped names.

autogpt_platform/frontend/src/app/(platform)/copilot/components/ContextPanel/components/FilesTab/helpers.ts
ntindle: Fixed in ea9e7b79e3. The legacy fallback is anchored to /sessions/<session>/tool-outputs or tool-results directly, so nested user-owned folders remain eligible. Regression cases include both data.json and toolu_mine.json under a user pipeline directory.

ntindle:
Reviewed the final approval on ea9e7b79e34ed3853c60762e70b141549d4d4f31. The getLastGeneratedFile naming suggestion is cosmetic; it selects the latest supplied candidate correctly. The uploaded-file predicate must remain: deliverables filters the full files array, and isInternalToolOutput deliberately returns false for uploads, so removing !isUploadedFile would promote uploads into the generated-deliverable bucket. The existing behavior and tests are retained.

Required CI passes and all discussion threads are resolved. The hosted reviewer reported provider exhaustion, so that variant is not claimed as validated; local reviewer and the documented backend/frontend test suites passed.

PR #13623, opened 2026-07-20
feat(backend/copilot): resolve copilot models through the LLM catalog

ntindle:
## Why

Part 3 of 5 of the catalog-as-code stack (#13621 → #13622 → this). Today the `copilot-model-routing` LD flag returns raw slugs nothing validates — the #13596 failure mode (LD referencing a model the platform has no data for) is structural, and a typo'd slug silently falls back with zero trace. This PR makes the catalog the serve-time gate while LD stays the cohort/experiment layer, and stamps served models onto chat messages so model experiments become measurable.

## What

**Resolution chain** (`copilot/model_router.py`): LD slug flag (per-user, unchanged JSON shape) → **catalog routing cell** (the file's `routing` section — PR-authored, deploy-propagated) → `ChatConfig` env default.

- **Serve-time gating**: slugs unknown to the catalog or `enabled: false` (the kill switch) are refused — warning log + Sentry + a Redis-recorded resolution warning — and resolution falls through. `HIDDEN` visibility serves when explicitly routed: the pre-launch state (add the model HIDDEN, LD-target employees, watch PI scores, flip GA).
- **Empty catalog gates nothing**: installs where the catalog fails to load keep exact pre-catalog behavior (all 27 pre-existing router tests pass unchanged).
- **Measurement stamp**: every persisted assistant message carries `model` + `routingSource` (`ld`/`db`/`env`) via the #13621 columns — product-intelligence's chat mirror can finally segment quality judgments by model (PI follow-up: product-intelligence#47).

**Two integration bugs found while wiring, fixed at resolve time** (catalog file untouched):
1. LD payloads and env defaults use OpenRouter spellings (`anthropic/claude-opus-4.6`) while the catalog registers canonical dashed slugs — without a spelling-tolerant lookup (exact → vendor-stripped → dots→dashes), the gate would have refused every legitimate LD Claude override the day the catalog loaded.
2. Bare `claude-*` cell slugs pass through `normalize_model_for_transport` unprefixed and would 404 on OpenRouter — the resolver vendor-prefixes them, and the direct-Anthropic transport strips the prefix back, so one cell value works on both transports (covered by tests against the real catalog cells).

## Verification

- Router suite 39/39 (27 dev-existing unchanged + 12 gating/normalization tests incl. real-catalog-cells-on-both-transports); registry suite 39/39
- Copilot db/baseline suites: 188 passed; 2 failures reproduce identically with the diff stashed (pre-existing local Redis-cluster timeout class)
[...]

PR #13622, opened 2026-07-20
feat(backend): add the canonical LLM catalog file and read cache

ntindle:
## Why

Part 2 of 5 of the catalog-as-code stack (#13621 → this). Model definitions today are spread across the `LlmModel` enum, `MODEL_METADATA`, `MODEL_COST`/`TOKEN_COST`, an LD flag, and env vars — five places, no shared schema, nothing validating cross-references. This PR creates the one canonical, schema-validated, PR-reviewed file that centralizes them, updated through git and propagated by CD (catalog-only diffs may ride hotfix→master; the `/review` bot covers the fast lane).

## What

- **`catalog.py`** — the database: 85 models / 8 providers / 17 creators / 4 copilot routing cells, **generated from current dev code** so file == reality on day one (costs mirror `MODEL_COST`/`TOKEN_COST`; routing cells mirror the `ChatConfig` defaults in canonical slug form; `gpt-5.2` recommended per `DEFAULT_LLM_MODEL`)
- **`catalog_model.py`** — the validated schema: per-model costs, capabilities, visibility (`GA/EMPLOYEES/ADMINS/HIDDEN`), tier gating, fallback slugs, routing cells
- **`registry.py`** — `load_catalog()` builds an in-process read cache at startup behind the stable read interface (`get_model`, `get_enabled_models`, `get_route`, `get_schema_options`, …) that parts 3–4 and Phase B consume. No Redis, no pub/sub, no DB: the file only changes at deploy, so cache coherence is structural
- **Forever-guards + cost-drift tripwires**: the catalog must parse, cross-reference cleanly, and match the legacy cost dicts *in both directions* — centralizing costs now cannot silently diverge from what billing actually charges. The tripwires die in Phase B3 when the dicts do
- Lifespan wiring is fail-soft: a broken catalog logs and degrades to existing behavior

Nothing consumes the cache yet — dormant by design, same as the schema PR.

## Verification

- 39/39 tests (schema validation edges, forever-guards, all three tripwire directions, loader/read-interface behavior)
- `poetry run format` + `poetry run lint` clean

## Checklist
- [x] Catalog generated from live code, not hand-transcribed
- [x] Out-of-scope changes: none

---

> [!NOTE]
> **Medium Risk**
> Touches centralized model metadata and credit pricing for all future LLM routing/billing consumers, but startup load is fail-soft and no code paths consume the cache yet; cost tripwires limit silent billing drift.
> 
> **Overview**
[...]

PR #13624, opened 2026-07-20
feat(backend/api): public LLM catalog endpoint

ntindle:
## Why

Part 4 of 5 of the catalog-as-code stack (#13621 → #13622 → #13623 → this). The catalog needs one public read surface: the Phase B model picker consumes it, and self-hosted installs can read it (though they primarily get the catalog on upgrade).

## What

`GET /api/llm/catalog` — the catalog's *public facts*, served from the in-process cache:

- **Unauthenticated by design**; GA-visibility models only; **costs and routing cells explicitly excluded** (cloud-internal config, not catalog facts) — a dedicated test seeds models *with* costs and cells and asserts neither appears in the response
- **Zero DB reads**; CDN-cacheable (`Cache-Control: public, max-age=300, stale-while-revalidate=3600` on 200s; exact-path `CACHEABLE_PATHS` entry — never a bare `/api/llm` prefix; explicit `no-store` on 429s since the middleware skips allowlisted paths)
- **Per-IP fail-open rate limit** (default 30/min): single-key INCRBY+EXPIRE pipeline, cluster-safe; protects read capacity, not money — a Redis brown-out must not take down the endpoint. Includes the backend's first X-Forwarded-For handling (LB-appended entry, configurable depth)
- OpenAPI schema regenerated via the canonical `export-api-schema` + prettier recipe

## Verification

- 93/93 across the endpoint, registry, and router suites — **run in both suite orders** (surfaced a test-order-dependence in the router tests, fixed on the base branch)

## Checklist
- [x] Public endpoint serves only non-sensitive GA facts; costs/routing verified excluded by test

---

> [!NOTE]
> **Medium Risk**
> New public surface exposes model metadata and relies on correct IP/rate-limit and cache-header behavior; misconfiguration could weaken abuse protection or cache error responses, though tests cover stripping sensitive fields and 429 cache headers.
> 
> **Overview**
> Adds an unauthenticated **`GET /api/llm/catalog`** that returns GA model facts from the in-process registry L1 cache (no DB), with **costs** and **routing** stripped and non-GA models omitted; disabled GA models remain with `is_enabled=false`, and fallback slugs pointing at excluded models are nulled.
> 
> Responses are **CDN-cacheable** (`public, max-age=300, stale-while-revalidate=3600`) via an exact-path entry in security middleware `CACHEABLE_PATHS`; **429** responses set `no-store` and `Retry-After`. A new **per-IP Redis rate limiter** (default 30/min, fail-open on Redis errors) includes **X-Forwarded-For** handling only when `behave_as=CLOUD`, plus config knobs `llm_catalog_rate_limit_per_minute` and `llm_catalog_client_ip_xff_depth`. The router is mounted on the main API and OpenAPI is updated; `SubscriptionTierName` is extracted as a shared type on registry models.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5398d6119897a9be8c0989044685bf18df233dec. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


ntindle:
Consolidated into #13627 — the stack now lands as a single PR that also performs the cutover (the catalog replaces the hand-maintained dicts in the same diff, per review feedback that dormant-then-flip phasing added process without adding safety). All commits from this branch are contained in the consolidated branch.

PR #13625, opened 2026-07-20
feat(backend): LLM model retirement CLI

ntindle:
## Why

Part 5 of 5 of the catalog-as-code stack (#13621 → #13622 → #13623 → #13624 → this). Retiring a model has two halves: flip `enabled: False` in a catalog PR (stops it serving), and migrate existing graph nodes onto a replacement so users' agents keep working. This is the second half — kept as an operator CLI rather than admin UI, per the design review.

## What

`python -m backend.data.llm_registry.retire`:

- `--usage <slug>` — how many AgentNodes reference the model
- `<slug> --replacement <slug>` — dry-run summary (exit 1) by default; `--yes` executes: transactional `SELECT…FOR UPDATE` + `JSONB_SET` rewrite of `AgentNode.constantInput`, schema-prefix-safe, recording a revertable `LlmModelMigration` row; prints the flip-the-catalog reminder
- `--revert <migration-id>` — restores the original node values (both directions use the provider-stripped node value — the v2 revert-matching fix rides along)
- `--list [--include-reverted]`

Replacement models are validated against the catalog: must exist **and** be enabled (kill-switched models rejected). Concurrent active migrations per source are blocked by the partial unique index from #13621.

## Verification

- 7 real-DB tests (46/46 suite): seeded-node usage counting, retire rewrite + record contents, revert round-trip, unknown/disabled replacements rejected with zero writes, duplicate active migration rejected, zero-node retire records nothing
- CLI smoke-tested live incl. dry-run exit codes; `poetry run format` + `poetry run lint` clean

## Checklist
- [x] Mutations are transactional, revertable, and recorded

---

> [!NOTE]
> **Medium Risk**
> Direct production DB mutations on agent graph nodes; mitigated by transactions, dry-run default, revert path, and catalog validation, but mis-operation could repoint many workflows.
> 
> **Overview**
> Adds an **operator CLI** (`python -m backend.data.llm_registry.retire`) to move existing workflows off a retired catalog model without editing the catalog file itself.
> 
> **Retire** rewrites `AgentNode.constantInput`’s `model` field using transactional `SELECT … FOR UPDATE` plus `JSONB_SET`, matching stored values **exactly** to catalog slugs (including provider-prefixed slugs). When nodes change, it writes a revertable **`LlmModelMigration`** row; a second active migration for the same source is rejected. Replacements must exist in the catalog and be **enabled**. Default behavior is a **dry-run** (exit 1); **`--yes`** runs the migration and reminds operators to flip `enabled: False` in a separate catalog PR.
> 
> Also supports **`--usage`**, **`--revert`** (only nodes still on the migration target are restored; manually repointed nodes are skipped), and **`--list`**. **`retire_test.py`** adds real-DB coverage for retire/revert, validation failures, duplicate migrations, and prefixed model slugs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bc8e94248c31d7e7779376d65dfcc40f038655b6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


ntindle:
Consolidated into #13627 — the stack now lands as a single PR that also performs the cutover (the catalog replaces the hand-maintained dicts in the same diff, per review feedback that dormant-then-flip phasing added process without adding safety). All commits from this branch are contained in the consolidated branch.

PR #13643, opened 2026-07-22
fix(code_executor): prevent UnicodeEncodeError on emoji data via ensure_ascii serialization

### LINKED ISSUE #13551: ExecuteCodeBlock: UnicodeEncodeError when data contains emoji (surrogate pairs in AGPT_VARIABLES env var)

PiedPiper911:
## Problem

When an upstream block returns data containing emoji (e.g. `"Holidays 🎅🏻"`), `ExecuteCodeBlock` crashes **before user code runs** with:
```
```

The variable injection serializes block inputs into the `AGPT_VARIABLES` env var. When data contains emoji encoded as UTF-16 surrogate pairs, Python 3.13's `os.environ.__setitem__` raises `UnicodeEncodeError`.

## Fix

Use `json.dumps(variables, ensure_ascii=True)` when serializing to `AGPT_VARIABLES`. This escapes all non-ASCII characters as `\uXXXX` sequences, which are pure ASCII and safe for environment variables on all platforms.

The deserialization side (`json.loads`) handles `\uXXXX` escapes transparently, so the user code receives the original Unicode strings unchanged.

## Testing

- Data with emoji (`"🎅🏻"`, `"🚀"`) no longer crashes
- Normal ASCII data is unaffected
- Round-trip: `json.loads(json.dumps(data, ensure_ascii=True)) == data` for all valid Unicode

Fixes #13551

### DIFF
--- a/autogpt_platform/backend/backend/blocks/code_executor_helpers.py
+++ b/autogpt_platform/backend/backend/blocks/code_executor_helpers.py
@@ -75,7 +75,7 @@ def build_variable_injection(
     _validate_keys(variables)
 
     try:
-        serialized = json.dumps(variables)
+        serialized = json.dumps(variables, ensure_ascii=True)
     except (TypeError, ValueError) as e:
         bad_keys = [k for k, v in variables.items() if not _is_json_serializable(v)]
         raise ValueError(


albatrossflyon-coder:
Heads up before this merges — `ensure_ascii=True` alone doesn't actually fix the crash from #13551, even though it looks like it should.

@Pwuts caught this on #13576 (which started with this exact same fix): a validly-paired emoji like `🎅🏻` survives `json.dumps(..., ensure_ascii=True)` fine on our side, but E2B's sandbox generates a literal Python source line from that payload (`os.environ['AGPT_VARIABLES'] = '...'`), and Python's own string-literal parser doesn't recombine adjacent `\uXXXX` surrogate halves the way JSON decoding does. So the escaped emoji gets re-split into two lone surrogates on the way into the sandbox, after it already left our process clean — and crashes with the same `UnicodeEncodeError` regardless.

Minimal repro from that thread:
```py
>>> os.environ['AGPT_VARIABLES'] = '"Holidays 🎅🏻"'
UnicodeEncodeError: 'utf-8' codec can't encode characters in position 10-13: surrogates not allowed
```

#13576 ended up switching to base64-encoding the payload instead (pure ASCII, nothing for the sandbox's code-gen to misinterpret), verified against that exact repro. Might be worth pointing this PR at that approach too, or at least confirming against the repro above before merge — happy to compare notes if useful.

autogpt_platform/backend/backend/blocks/code_executor_helpers.py
```diff
@@ -75,7 +75,7 @@ def build_variable_injection(
     _validate_keys(variables)
 
     try:
-        serialized = json.dumps(variables)
+        serialized = json.dumps(variables, ensure_ascii=True)
```
Pwuts, line 78: `ensure_ascii` is *default* `True`, so this doesn't change the behavior

Pwuts:
Duplicate of #13576

PR #13654, opened 2026-07-22
feat(backend): enforce grant credential modes — OWNER runs on the owner's credentials

ntindle:
## Why

SECRT-2448 follow-up (Nick, 2026-07-22: "shouldn't we build this then?"): `credentialMode` was stored on grants but nothing enforced it — a sharer choosing OWNER believed consumers would run on the owner's credentials while executions silently used the consumer's.

## What

When a run's access comes from a team grant with `credentialMode=OWNER`, the graph's credential references resolve against the **graph owner's** store — secrets injected at execution only, never shown to the consumer (title redacted from the consumer's execution record).

Guardrails:
- **Allowlisted**: only the credential ids the graph itself bakes into `node.input_default` — a consumer can never point resolution at arbitrary entries in the owner's store; consumer credential overrides are ignored in OWNER mode.
- **Fail-closed**: missing/revoked owner credential → clear error naming graph + credential; never a silent fallback to consumer creds (which would misattribute API usage).
- **Inert** for owners running their own graphs, marketplace/library access, and CONSUMER grants; **sub-graphs run CONSUMER-only** (nested executions clear the owner context — consumers can't reach private sub-agents today anyway; flagged as future work).
- **One authorization system**: reuses `resolve_graph_grant` (ACTIVE membership, org/version/followLatest constraints), re-resolved on resume; audit log (graph, grant, owner, consumer — no secrets) when OWNER engages.
- **Creation-time**: OWNER grants can only be created by the graph's owner — an org admin sharing a third party's graph cannot expose that owner's credentials (400).

## How

Resolved once at execution-start (`resolve_execution_credentials_owner`, exposed over the DB-manager RPC), threaded via `ExecutionContext.credentials_owner_id` to the two fetch sites (manager + auto-credentials) and the enqueue-time validation, which had to change in lockstep or validation would reject owner refs.

## Testing

87 tests green across resolver decisions, both fetch sites, validation lockstep, override suppression, fail-closed behavior, creation authz, audit emission; 0 new pyright errors.

Stacked on #13599 (grants schema/API).

---

> [!NOTE]
> **High Risk**
> Changes team-grant authorization, credential resolution, and execution context for runs that use another user's secrets; mistakes could expose credentials or allow unauthorized OWNER-mode execution.
> 
> **Overview**
> Enforces **OWNER** team grants so shared graph runs resolve credential references against the **graph owner's** store, not the consumer's. `credentialMode` on grants was previously stored but not applied at execution time.
> 
> **Grant creation** blocks org admins from creating OWNER grants on another user's graph, rejects OWNER when the graph has active runtime-managed (`credential_reference_only`) credential refs, and refreshes `createdByUserId` on upsert for consent tracking.
> 
> **Runtime** adds `resolve_execution_credentials_owner` / `validate_execution_credentials_owner` (also on the DB-manager RPC), threads `credentials_owner_id` and `credentials_grant_id` through `ExecutionContext`, and resolves once at enqueue/resume with per-node revalidation before any owner credential is touched. Only ids baked into each node's `input_default` are allowlisted; consumer credential overrides are dropped in OWNER mode. Sub-graph runs clear owner context (CONSUMER-only). Execution hierarchy uses persisted `parent_execution_id` to derive roots and reject mismatches/cycles on resume.
> 
> **Executor changes** mirror that policy in validation and fetch paths (`manager`, `auto_credentials`, `utils`), fail closed on missing owner creds, redact owner credential ids in logs, and extend grant resolution (`resolve_graph_grants`) for org-aligned, multi-team grants when opening or executing graphs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 699aa252774f6ba786cba29554ea1f623bfd1fcf. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


autogpt_platform/backend/backend/executor/auto_credentials.py
```diff
@@ -68,13 +83,33 @@ async def acquire_auto_credentials(
                             f"builder and re-select the file."
                         )
                     file_name = field_data.get("name", "selected file")
+                    # OWNER-mode: resolve the graph's own referenced file
+                    # credential against the owner's store; everything else
+                    # against the executing consumer.
+                    resolve_user_id = (
+                        credentials_owner_id
+                        if credentials_owner_id and cred_id in owner_credential_ids
```
kcze: 🤖 🔴 **Blocker**: OWNER-mode auto-credential resolution keys only on `cred_id in owner_credential_ids`, but `field_data` (and its file/resource `id`) is read from runtime `input_data`, which a consumer can still override via a node input mask — OWNER mode drops only `graph_credentials_inputs`, not general masks. A consumer can keep the owner's baked `_credentials_id` while swapping `id` to an arbitrary file, so the owner's OAuth credential is acquired to fetch a consumer-chosen resource. Example: a consumer supplies `{_credentials_id: <owner id from node.input_default>, id: <any file>}` and the block reads that file from the graph owner's private Drive. Bind owner-credential use to the unchanged baked field value for that exact field, not to id-set membership. _(flagged by: Codex)_

autogpt_platform/backend/backend/executor/utils.py
```diff
@@ -569,9 +594,15 @@ def _mark_optional_skip() -> None:
...
                     is_creds_optional or field_name not in required_fields
                 )
-                # Check input_default and nodes_input_masks for the field value
+                # Check input_default and nodes_input_masks for the field value.
+                # In OWNER mode, ignore consumer masks for credential-bearing
+                # fields and use only the owner's baked reference.
                 field_value = node.input_default.get(field_name)
-                if nodes_input_masks and node.id in nodes_input_masks:
+                if (
+                    not credentials_owner_id
```
kcze: 🤖 🟡 **Nice to Have**: Enqueue validation for auto-credential fields ignores consumer masks in OWNER mode and always validates the owner's baked default, but runtime (`acquire_auto_credentials`) resolves non-baked ids against the consumer's store. The two diverge: a valid consumer-selected file is rejected at preflight when the owner default is absent/stale, while an invalid consumer credential passes preflight and fails only in the executor. Apply the same field-level owner-vs-consumer rule here as at runtime. _(flagged by: Codex)_

PR #13672, opened 2026-07-25
fix(backend/copilot): actionable per-UUID reasons on memory forget failures

kcze:
### Why / What / How

**Why:** `memory_forget_confirm` called with the default `hard_delete: false` fails silently — it returns `0 invalidated, N failed` with no reason, while the same UUIDs succeed immediately with `hard_delete: true` (SECRT-2371).

Root cause (confirmed from git history): the originally shipped `_soft_delete_edges` set its timestamps with Cypher's no-arg `datetime()`:

```cypher
SET e.invalid_at = datetime(), e.expired_at = datetime()
```

FalkorDB does not implement `datetime()` — it raises `Unknown function 'datetime'`. That exception was caught by a bare `except Exception` and the UUID was silently pushed onto the `failed` list. Hard delete never called `datetime()`, so it worked every time. The "recently-created edges" pattern noted in the ticket is spurious correlation — the shipped soft delete failed for *all* edges; the reporter only ever tested recent ones.

The dream-pass refactor (#13243) already migrated the soft-delete path to a Python-generated `$now` parameter (`_retract_edges` / `_now_iso()`), which fixes the *functional* failure on `dev`. What remained — and is explicitly called out as a bug in its own right in the ticket — is the **silent, reasonless failure path**: even a legitimate no-match returned a bare count with nothing for the model to act on.

**What:** Every forget failure now carries an actionable, per-UUID reason with a stable machine-switchable code, surfaced both structurally (response schema) and in the model-visible message. Also adds a regression guard so the `datetime()` root cause cannot be reintroduced.

**How — error surface:**
- New `MemoryForgetFailure` model with three fields: `uuid`, `code` (enum), and `reason` (free text).
- New `MemoryForgetFailureCode` str-Enum with stable values `no_match` and `query_error`, so the frontend/model can branch on the code without parsing prose (consumers must tolerate unknown future codes).
- New `failures: list[MemoryForgetFailure]` field on `MemoryForgetConfirmResponse` (additive, backward-compatible; the pre-existing `failed_uuids` list is retained).
- `_retract_edges` and `_hard_delete_edges` (both only called by the confirm path) now distinguish a real query error (`query_error`, carrying the exception type + message, e.g. `Unknown function 'datetime'`) from a plain no-match (`no_match`), returning `(succeeded, failures)`.
- `_build_confirm_message` spells out each failed UUID and its reason instead of a bare "N failed".
- Regenerated `frontend/src/app/api/openapi.json` for the new response schema.

**Decisions (confirmed with maintainer):**
- `hard_delete` default stays `false` (soft delete). Defaulting to irreversible removal would be a dangerous default.
- The timestamp path is left as-is on `dev` (already fixed); this PR only adds a regression guard test pinning that the soft-delete Cypher never reintroduces `datetime()`.

Scope kept minimal: `_soft_delete_edges`, `mark_edges_superseded`, and `invalidate_entity_direct_neighbors` (used by the dream pass) are untouched — only the two confirm-path helpers changed.

Fixes SECRT-2371

### Changes 🏗️

- Add `MemoryForgetFailureCode` (str-Enum: `no_match`, `query_error`) and `MemoryForgetFailure` (`uuid`, `code`, `reason`); add `failures` to `MemoryForgetConfirmResponse` (`models.py`).
- Return per-UUID `(code, reason)` failures — query-error vs. no-match — from `_retract_edges` and `_hard_delete_edges` (`graphiti_forget.py`).
- Surface reasons in the confirm response message via `_build_confirm_message`.
- Regenerate `openapi.json`.
- Tests: actionable code+reason coverage for both helpers, a tool-level end-to-end assertion that reasons reach the response, and a `datetime()` regression guard.

#### For code changes:
- [x] I have made a test plan
  - [x] `poetry run pytest backend/copilot/tools/graphiti_forget_test.py` — 20 passed (mocked driver, no infra)
  - [x] `poetry run format` and `poetry run lint` — clean
  - [x] Regenerated + prettier-formatted `openapi.json`
  - [ ] Full suite / integration relies on CI: Docker (postgres/redis) was not running locally, so suites needing server fixtures were not executed here.


kcze (commented):
🤖 Review done, no findings.

autogpt_platform/backend/backend/copilot/tools/graphiti_forget_test.py
```diff
@@ -103,14 +111,126 @@ async def test_retract_sets_only_expired_at(self) -> None:
...
         deleted, failed = await _retract_edges(driver, ["missing"], "test-user")
         assert deleted == []
-        assert failed == ["missing"]
+        assert [f.uuid for f in failed] == ["missing"]
+        # No-match must carry an actionable reason, not a bare count.
+        assert failed[0].reason
+        assert failed[0].code == MemoryForgetFailureCode.NO_MATCH
+
+
+class TestForgetFailuresAreActionable:
```
kcze, line 146: 🤖 🟡 **Nice to Have**: No test covers a *mixed* batch (some UUIDs deleted **and** some failed), so `_build_confirm_message`'s two-part formatting and the co-populated `deleted_uuids` + `failures` response are never exercised together. Every case here has `deleted == []`, so the summary-plus-detail string (`"2 memory edge(s) retracted from memory. 1 failed — <uuid>: <reason>"`) and the early-return branch `if not failures` are both untested — a regression that mangled the success half of the message would pass CI. _(flagged by: Claude)_

autogpt_platform/backend/backend/copilot/tools/graphiti_forget.py
```diff
@@ -249,30 +268,43 @@ async def _execute(
...
+) -> str:
+    """Human/model-readable summary that spells out *why* edges failed.
+
+    A bare "N failed" gives the model nothing to act on (SECRT-2371); listing
+    each UUID with its reason lets it retry, hard-delete, or tell the user.
+    """
+    summary = f"{deleted_count} memory edge(s) {mode}."
+    if not failures:
+        return summary
+    detail = "; ".join(f"{f.uuid}: {f.reason}" for f in failures)
```
kcze: 🤖 🟡 **Nice to Have**: This joins every failing UUID with its full reason and no cap, while the `uuids` input schema has no `maxItems`. Each failure contributes ~160 chars (36-char UUID + the ~120-char `_NO_MATCH_REASON`), and a driver outage fails *every* UUID with a longer `_delete_error_reason` string. Failure: a large forget batch during an outage pushes the serialized tool output past `_LARGE_OUTPUT_THRESHOLD` (80_000, `tools/base.py:22`), so it is replaced by the persist-and-summarize stub and the model ends up with *less* actionable information than the old bare "N failed" count — the opposite of this PR's goal. Capping the detail list (e.g. first N failures + "…and M more") would preserve it. _(flagged by: claude-opus-5)_

autogpt_platform/backend/backend/copilot/tools/graphiti_forget.py
```diff
@@ -38,6 +46,16 @@ def _now_iso() -> str:
...
+def _delete_error_reason(exc: Exception) -> str:
+    """Actionable reason string for a delete query that raised.
+
+    Keeps the exception type + message (e.g. FalkorDB's
+    ``Unknown function 'datetime'``) so the model can distinguish a real
+    query error from a plain no-match and react accordingly.
+    """
+    return f"Deletion query failed: {type(exc).__name__}: {exc}"
+
+
```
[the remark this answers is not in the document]
kcze: 🤖 Fixed in 987d3eb39f: `_delete_error_reason` no longer embeds `str(exc)`/`repr(exc)`. It now surfaces only the exception type plus `exc.args[0]` (e.g. `Unknown function 'datetime'`), never the full repr that could carry connection/host details, and the complete exception is still logged server-side with `exc_info=True`. Note these helpers run against an already-established driver — connection/auth failures are caught upstream at `get_graphiti_client` and return a generic "service unavailable" message — so what reaches this path is query-level errors.

autogpt_platform/backend/backend/copilot/tools/graphiti_forget_test.py
kcze: 🤖 Fixed in 987d3eb39f: added `test_confirm_tool_mixed_batch_reports_both`, which drives a batch where one UUID deletes and one fails, asserting `deleted_uuids == ["kept"]`, `failures == ["gone"]`, and that the message carries both the success half ("1 memory edge(s) retracted from memory.") and the failure detail. Also added `TestBuildConfirmMessage` covering the `if not failures` early return.

autogpt_platform/backend/backend/copilot/tools/graphiti_forget.py
kcze: 🤖 Fixed in 987d3eb39f: `_build_confirm_message` now inlines at most `_MAX_FAILURE_DETAIL` (5) reasons and appends "…and N more", so a wholesale-failure batch (e.g. driver outage) stays well under the tool-output size threshold instead of being truncated to a stub. The full per-UUID list is still available in the structured `failures` field. Covered by `test_caps_inlined_detail_and_notes_remainder`.

ntindle (approved):

PR #13683, opened 2026-07-27
feat(platform): support static API-key/bearer-token auth for MCP servers

Abhi1992002:
### Why / What / How

**Why:** Some MCP servers (e.g. [DataFast](https://datafa.st/api/mcp)) authenticate with a **static bearer token / API key** issued in the vendor's own dashboard, not a full OAuth2 authorize/token exchange. The MCP tool block only advertised `oauth2` credentials, and — critically — the graph builder's "Connect to MCP Server" dialog used a manually-entered token **only for tool discovery** and never persisted it. The placed block therefore ran with **no credentials** and failed with a 401 at execution time. (Linear: REQ-115)

**What:** Make static API-key / bearer-token a first-class credential type for MCP server connections, entered via the builder's secure credential dialog, and stored as a proper `api_key` credential the block uses at runtime.

**How:**
- `MCPToolBlock` now advertises both `oauth2` and `api_key` credential types and accepts either at runtime — the bearer token is pulled from `access_token` (OAuth2) or `api_key` (API key) via a single shared `mcp_auth_token()` helper.
- `POST /api/v2/mcp/token` now stores the token as a first-class `APIKeyCredentials` (type `api_key`) instead of masquerading it as an `OAuth2Credentials`.
- **Backward compatible:** credential lookup, cleanup, and `get_host` now match **both** `oauth2` and `api_key` MCP credentials by server URL, so tokens already stored under the old (OAuth2-masquerade) shape keep working.
- The builder MCP dialog now offers **proactive** "Use an API key / bearer token instead" entry (no failed-OAuth round-trip required), persists the token via `/mcp/token`, and attaches the returned credential to the block node.
- The frontend credential classifier matches `api_key` MCP credentials by host, same as OAuth2.

The copilot (`run_mcp_tool` + `MCPSetupCard`) bearer-token flow already worked; this PR reuses the same `/token` endpoint and brings the builder to parity.

### Changes 🏗️

**Backend**
- `blocks/mcp/block.py` — widen `MCPCredentials` to `Literal["oauth2", "api_key"]`; `run()` accepts `OAuth2Credentials | APIKeyCredentials`; token extracted via `mcp_auth_token()`.
- `blocks/mcp/helpers.py` — add `mcp_auth_token()`, `is_mcp_credential_for_server()`, `MCPCredential` union; `auto_lookup_mcp_credential()` matches both credential types (OAuth2-only `refresh_if_needed` still guarded).
- `api/features/mcp/routes.py` — `/token` stores `APIKeyCredentials`; discovery token extraction and old-credential cleanup (both `/token` and OAuth callback) cover both types.
- `api/features/integrations/router.py` — `get_host()` returns the MCP server URL for `APIKeyCredentials` too.
- `copilot/tools/run_mcp_tool.py` — token extraction via the shared helper.

**Frontend**
- `build/components/MCPToolDialog.tsx` — proactive API-key entry; persist the token via `/mcp/token` on successful discovery and attach the credential to the node.
- `hooks/useCredentials.ts` — `classifyCredentials` matches `api_key` MCP credentials by host.

**Tests** — backend: helper token extraction + both-type matching, block `run()` with `APIKeyCredentials`, `/token` stores `api_key` + cleans up legacy OAuth2 and api_key rows. Frontend: dialog persists+attaches the token (and does not for public servers), classifier matches api_key MCP creds.

#### For code changes:
- [x] I have made a test plan
  - [x] `poetry run pytest backend/blocks/mcp backend/api/features/mcp/test_routes.py backend/copilot/tools/test_run_mcp_tool.py` — all pass
  - [x] `pnpm test:unit` for `MCPToolDialog.test.tsx` and `classifyCredentials.test.ts` — all pass
  - [x] `poetry run format`, `poetry run lint`, `pnpm lint`, `pnpm types` — clean
  - [ ] Manual: in the builder, add the MCP Tool block, enter a bearer-token-only server URL, choose "Use an API key / bearer token instead", paste a token, add the block, and run it — the tool executes (no 401)

---

### Update: rebased on `dev` + review round addressed

Merged latest `dev` (3 conflicts: `useCredentials.ts`, `CredentialField.tsx`, `CredentialsInput.test.tsx`) and addressed all 10 open review threads.

**Behaviour changes worth a second look:**

- **`MCPToolBlock.run` auto-lookup fallback is back.** `executor/manager.py` nulls the credentials field for three indistinguishable reasons — the user picked "None", the node never had one, and *the node's credential ID points at a row that has since been deleted*. Reconnecting a server deletes its old credential and issues a new ID, so that third case is an ordinary flow; without the fallback every run after a reconnect goes out unauthenticated and 401s. The fallback can override an explicit "None" on a server the user also has a token for, which is the far milder failure. The test that locked in the old behaviour was inverted rather than deleted.
- **`_mcp_credential_rank` now ranks on usability, not expiry.** Treating a non-expiring credential as infinitely-far-in-the-future made a static key permanently outrank a freshly obtained OAuth token, so re-authenticating could never take effect (the generic `POST /{provider}/credentials` endpoint does no per-server cleanup, so the two coexist through a normal flow). Rank is now 1 = still authenticates / 0 = lapsed, with store order breaking ties.
- **`mcp_oauth_callback` creates before deleting**, mirroring `mcp_store_token` — a failed `create` no longer leaves the user with no credential at all.
- **New `mcpStoreToken` on the credentials provider.** Storing a token deletes the server's previous credential, so the builder dialog now goes through the provider (like the OAuth path) and `replaceMCPServerCredentials` evicts the replaced rows from the cached list. Without this the picker kept re-selecting a deleted ID.
- **Cross-server token leak fixed.** `manualToken` was left loaded (and hidden) after a successful connect, so "Back" → new URL sent server A's secret to server B *and* persisted it as B's credential. Cleared on store and on "Back".
- **The merge surfaced a second broken path.** `dev`'s new `ConnectCredentialDialog` (`variant="default"` — run dialogs, copilot) creates API keys via `useApiKeyConnectForm` → `postV1CreateCredentials` with no metadata, which for MCP produces a `host: null` credential that nothing can match. It now threads `mcp_server_url` through. The builder (`variant="node"`) still uses the per-type flow.

Also: normalized MCP URL matching in `copilot/tools/utils.py` and `CredentialsGroupedView/helpers.ts`, blocked API-key submission when the node has no `server_url`, `Record<string, unknown>` over `any`, and hoisted function-local test imports.

**Verified locally:** `pytest backend/blocks/mcp backend/copilot/tools/utils_test.py backend/api/features/mcp` → 122 passed; `pnpm test:unit` → 6007 passed (562 files); `pnpm types`, `pnpm lint`, `poetry run format`, and `pyright` on the changed backend dirs all clean.


autogpt_platform/frontend/src/app/(platform)/build/components/MCPToolDialog.tsx
```diff
@@ -115,17 +116,45 @@ export function MCPToolDialog({
...
+      // the placed block with no credentials and failing on the next run.
+      if (authToken) {
+        const tokenRes = await postV2StoreABearerTokenForAnMcpServer({
+          server_url: url,
+          token: authToken,
+        });
+        if (tokenRes.status !== 200) throw tokenRes.data;
+        setCredentials({
+          id: tokenRes.data.id,
+          provider: tokenRes.data.provider,
```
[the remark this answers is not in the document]
Abhi1992002: Fixed in 28d239489: a non-2xx response from `/mcp/token` now throws a clear "saving your API token failed" error instead of the raw response body, so the user sees an accurate message rather than a misleading connection error.

autogpt_platform/backend/backend/blocks/mcp/helpers.py
```diff
@@ -109,39 +145,37 @@ async def invalidate_mcp_credential(user_id: str, credential_id: str) -> None:
...
+            if is_mcp_credential_for_server(cred, server_url):
                 if best is None or (
-                    (cred.access_token_expires_at or 0)
-                    >= (best.access_token_expires_at or 0)
+                    _mcp_credential_expiry(cred) >= _mcp_credential_expiry(best)
                 ):
                     best = cred
-        if best:
+        if best is not None and isinstance(best, OAuth2Credentials):
             best = await mgr.refresh_if_needed(user_id, best)
```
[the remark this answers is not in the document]
Abhi1992002: Fixed in 28d239489: the ranking helper now scores non-expiring credentials highest (via `sys.maxsize`) so a valid static bearer token is no longer shadowed by a stale row that once had an expiry set.

autogpt_platform/frontend/src/hooks/useCredentials.ts
```diff
@@ -27,8 +27,9 @@ export function classifyCredentials(
   for (const c of allSaved) {
     if (!supportedTypes.includes(c.type)) continue;
 
-    // MCP OAuth2 credentials filter by server URL — not upgradeable
-    if (c.type === "oauth2" && c.provider === "mcp") {
+    // MCP credentials (OAuth2 tokens and static API-key / bearer tokens)
+    // filter by server URL — not upgradeable.
+    if ((c.type === "oauth2" || c.type === "api_key") && c.provider === "mcp") {
       if (discriminatorValue != null && c.host === discriminatorValue) {
         savedCredentials.push(c);
```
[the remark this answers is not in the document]
Abhi1992002: Fixed in 28d239489: `matchesDiscriminatorValues` in CredentialsGroupedView/helpers.ts now filters `api_key` MCP credentials by host (mirroring the classifyCredentials fix), so an api_key credential for one server no longer matches another. Added a regression test in CredentialsGroupedView/__tests__/helpers.test.ts.

autogpt_platform/backend/backend/api/features/mcp/routes.py
```diff
@@ -361,16 +366,14 @@ async def mcp_oauth_callback(
...
         old_creds = await creds_manager.store.get_creds_by_provider(
             user_id, ProviderName.MCP.value
         )
         for old in old_creds:
-            if (
-                isinstance(old, OAuth2Credentials)
-                and (old.metadata or {}).get("mcp_server_url") == meta["server_url"]
-            ):
+            if is_mcp_credential_for_server(old, normalize_mcp_url(meta["server_url"])):
                 await creds_manager.store.delete_creds_by_id(user_id, old.id)
```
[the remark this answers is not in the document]
Abhi1992002: This read/replace/delete pattern is pre-existing in both flows — this PR only broadened the credential-type filter used to identify old rows, it did not introduce the sequence. Two mitigations keep it safe in practice: (1) the new credential is created *before* any old ones are deleted, so a mid-sequence failure never leaves the user with zero credentials for the server; (2) the rare concurrent-double-submit case that leaves a duplicate is tolerated by auto_lookup_mcp_credential, which deterministically picks the best match rather than failing. A store-level atomic replace would require a cross-cutting change to IntegrationCredentialsStore affecting all providers, which is out of scope here. Leaving the existing behavior as-is for this PR.

autogpt_platform/frontend/src/components/contextual/CredentialsInput/components/CredentialsGroupedView/__tests__/helpers.test.ts
```diff
@@ -0,0 +1,47 @@
+import { describe, expect, it } from "vitest";
+import { findSavedUserCredentialByProviderAndType } from "../helpers";
+
+function makeProviders(creds: unknown[]) {
+  return { mcp: { savedCredentials: creds } } as never;
+}
+
+function apiKeyCred(id: string, host: string) {
+  return { id, provider: "mcp", type: "api_key", title: `MCP: ${host}`, host };
+}
```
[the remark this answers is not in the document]
Abhi1992002: Fixed in the latest push: renamed `apiKeyCred` to `APIKeyCred` per the acronym-capitalization guideline.

Abhi1992002:
Thanks — addressing the review. Most items were already fixed in the same commit the review ran against (`28d2394`); the rest are now pushed. Summary:

**🔴 Blocker — `api_key` MCP creds match every server in the grouped picker (`CredentialsGroupedView/helpers.ts:32`):** already fixed in `28d2394`. `matchesDiscriminatorValues` now host-filters both `oauth2` and `api_key` MCP credentials, mirroring the `classifyCredentials` fix. A regression test was added (`CredentialsGroupedView/__tests__/helpers.test.ts`) asserting an api_key cred for one server is not offered for another.

**🟠 Should-fix 1 — direct test for `auto_lookup_mcp_credential` branching:** added. `test_helpers.py` now pins mixed-type "best" selection (non-expiring api_key beats a stale OAuth row), the OAuth-only refresh guard, and no-match → None — no longer only exercised through mocks.

**🟠 Should-fix 2 — validity-aware ranking:** fixed in `28d2394`. Non-expiring credentials now rank highest (`sys.maxsize`), so an expired OAuth row can't shadow a valid non-expiring api_key.

**🟠 Should-fix 3 — clear stale error on auth-mode toggle:** fixed. Both toggles now `setError(null)`.

**🟠 Should-fix 4 — persist-failure misattributed as invalid-token:** fixed in `28d2394`. A non-2xx `/mcp/token` response now throws a distinct "saving your API token failed" message rather than the raw body.

**🔵 Nits — a11y:** added `type="button"` to the toggle buttons and `role="alert"` on the error message.

Deferred (with rationale, noted in the CodeRabbit thread): the TOCTOU/atomic credential-replace is pre-existing (create-before-delete already avoids data loss; duplicates are tolerated by `auto_lookup`) and a store-level atomic replace is a cross-cutting change out of scope here. Persist-on-confirm vs on-discovery and case-insensitive URL normalization are noted as follow-ups.

autogpt_platform/backend/backend/blocks/mcp/helpers.py
```diff
@@ -109,39 +157,37 @@ async def invalidate_mcp_credential(user_id: str, credential_id: str) -> None:
 
 async def auto_lookup_mcp_credential(
     user_id: str, server_url: str
-) -> OAuth2Credentials | None:
+) -> MCPCredential | None:
     """Look up the best stored MCP credential for *server_url*.
```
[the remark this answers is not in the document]
Abhi1992002: Fixed in d952ec5: the docstring and inline comment now describe the actual ranking (non-expiring credentials highest, then latest expiry), matching `_mcp_credential_rank`.

autogpt_platform/backend/backend/blocks/mcp/test_helpers.py
```diff
@@ -96,3 +109,134 @@ def test_parse_multi_item():
...
+
+
+def test_mcp_auth_token_oauth():
+    assert mcp_auth_token(_oauth_cred("https://mcp.example.com/mcp")) == "oauth-token"
+
+
+def test_mcp_auth_token_api_key():
+    assert (
+        mcp_auth_token(_api_key_cred("https://mcp.example.com/mcp")) == "static-token"
+    )
```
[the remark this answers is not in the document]
Abhi1992002: Fixed in 4c44e78f1: added a test passing a HostScopedCredentials so the TypeGuard's non-OAuth2/non-APIKey rejection branch is covered.

autogpt_platform/backend/backend/blocks/mcp/helpers.py
```diff
@@ -109,39 +157,37 @@ async def invalidate_mcp_credential(user_id: str, credential_id: str) -> None:
...
+            if is_mcp_credential_for_server(cred, server_url):
                 if best is None or (
-                    (cred.access_token_expires_at or 0)
-                    >= (best.access_token_expires_at or 0)
+                    _mcp_credential_rank(cred) >= _mcp_credential_rank(best)
                 ):
                     best = cred
-        if best:
+        if best is not None and isinstance(best, OAuth2Credentials):
             best = await mgr.refresh_if_needed(user_id, best)
```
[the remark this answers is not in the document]
Abhi1992002: Fixed in 4c44e78f1: added test_auto_lookup_tiebreaks_equal_rank_by_recency covering the most-recently-created tiebreaker among equal-rank credentials.

autogpt_platform/backend/backend/api/features/mcp/routes.py
```diff
@@ -92,7 +97,7 @@ async def discover_tools(
             user_id, normalize_mcp_url(request.server_url)
         )
         if best_cred:
-            auth_token = best_cred.access_token.get_secret_value()
+            auth_token = mcp_auth_token(best_cred)
```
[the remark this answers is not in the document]
Abhi1992002: Fixed in 4c44e78f1: added test_discover_tools_auto_uses_stored_api_key_credential asserting the route pulls the token from a stored api_key credential.

autogpt_platform/frontend/src/app/(platform)/build/components/MCPToolDialog.tsx
```diff
@@ -115,17 +116,48 @@ export function MCPToolDialog({
...
+      // persist it so the block can authenticate at runtime — discovery only
+      // uses the token transiently and would otherwise be discarded, leaving
+      // the placed block with no credentials and failing on the next run.
+      if (authToken) {
+        const tokenRes = await postV2StoreABearerTokenForAnMcpServer({
+          server_url: url,
+          token: authToken,
+        });
+        if (tokenRes.status !== 200)
+          throw new Error(
```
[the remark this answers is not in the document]
Abhi1992002: Fixed in 4c44e78f1: reworded to "Discovered the server, but saving your API token failed." so it no longer implies a connection/credential was established.

autogpt_platform/frontend/src/app/(platform)/build/components/MCPToolDialog.tsx
```diff
@@ -115,17 +116,48 @@ export function MCPToolDialog({
...
+          id: tokenRes.data.id,
+          provider: tokenRes.data.provider,
+          type: tokenRes.data.type,
+          title: tokenRes.data.title,
+        });
+      }
+
       setTools(response.data.tools);
       setServerName(response.data.server_name ?? null);
       setAuthRequired(false);
```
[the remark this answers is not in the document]
Abhi1992002: Fixed in 4c44e78f1: the discovery catch is now typed `unknown` with a narrowed cast instead of `any`.

autogpt_platform/frontend/src/app/(platform)/build/components/MCPToolDialog.tsx
```diff
@@ -115,17 +116,48 @@ export function MCPToolDialog({
         auth_token: authToken || null,
       });
       if (response.status !== 200) throw response.data;
+
+      // When the user authenticated with a static API key / bearer token,
+      // persist it so the block can authenticate at runtime — discovery only
+      // uses the token transiently and would otherwise be discarded, leaving
+      // the placed block with no credentials and failing on the next run.
+      if (authToken) {
+        const tokenRes = await postV2StoreABearerTokenForAnMcpServer({
```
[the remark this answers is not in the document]
Abhi1992002: Intentional for this PR: discovery already validates the token against the live server, so persisting it there guarantees the placed block authenticates at runtime. Stored MCP creds are encrypted, user-scoped, and de-duplicated on re-store, so a dialog-cancel leaves at most one harmless unused row. Persist-on-confirm is noted as a follow-up.

autogpt_platform/frontend/src/app/(platform)/build/components/MCPToolDialog.tsx
```diff
@@ -115,17 +116,48 @@ export function MCPToolDialog({
         auth_token: authToken || null,
       });
       if (response.status !== 200) throw response.data;
+
+      // When the user authenticated with a static API key / bearer token,
+      // persist it so the block can authenticate at runtime — discovery only
+      // uses the token transiently and would otherwise be discarded, leaving
+      // the placed block with no credentials and failing on the next run.
+      if (authToken) {
```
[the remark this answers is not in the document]
Abhi1992002: Same as the persist-on-discovery note above — the token is validated by discovery and stored encrypted/user-scoped/de-duplicated. Moving persistence to block-confirm is a reasonable follow-up but out of scope here.

autogpt_platform/backend/backend/blocks/mcp/helpers.py
```diff
@@ -24,6 +30,48 @@ def normalize_mcp_url(url: str) -> str:
...
+    stored = (cred.metadata or {}).get("mcp_server_url", "")
+    return normalize_mcp_url(stored) == server_url
+
+
+def _mcp_credential_rank(cred: MCPCredential) -> int:
+    """Ranking key for picking the best MCP credential when several match.
+
+    Non-expiring credentials rank highest (they never go stale); among
+    expiring ones, the later expiry wins. Returning 0 for non-expiring — the
+    naive approach — would make a valid static bearer token lose to any stale
```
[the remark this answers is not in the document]
Abhi1992002: By design and low-impact: this only matters when best-effort cleanup has left two rows for one server (rare). Both the api_key and the OAuth token are valid credentials for the same server, so preferring the non-expiring static token is a safe choice — it can't be shadowed by a stale row that merely once had an expiry. The recency tiebreaker (now tested) covers the equal-rank case.

autogpt_platform/backend/backend/blocks/mcp/helpers.py
```diff
@@ -24,6 +30,48 @@ def normalize_mcp_url(url: str) -> str:
...
+    row that merely once had an expiry set.
+    """
+    expiry = (
+        cred.expires_at
+        if isinstance(cred, APIKeyCredentials)
+        else cred.access_token_expires_at
+    )
+    return expiry if expiry is not None else sys.maxsize
```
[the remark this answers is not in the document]
Abhi1992002: Both credentials authenticate the same server, so selecting the non-expiring static token over a future-dated OAuth token does not break auth. This branch is only reachable after a failed cleanup leaves duplicates; in the normal single-credential case ranking is a no-op.

autogpt_platform/backend/backend/api/features/mcp/routes.py
```diff
@@ -13,15 +13,20 @@
 from fastapi import Security
 from pydantic import BaseModel, Field, SecretStr
 
-from backend.api.features.integrations.router import CredentialsMetaResponse
+from backend.api.features.integrations.router import (
```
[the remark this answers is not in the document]
Abhi1992002: mcp/routes.py already imported CredentialsMetaResponse from integrations/router.py before this PR; reusing to_meta_response from the same module keeps the response shape consistent and avoids re-deriving it. Relocating the serializer to a shared module is a reasonable cleanup but out of scope for this change.

autogpt_platform/backend/backend/api/features/mcp/routes.py
```diff
@@ -469,22 +472,7 @@ async def mcp_store_token(
...
-        # Use the full ``mcp_server_url`` for parity with the OAuth callback
-        # response and with ``to_meta_response``'s ``get_host`` (which reads
-        # the same metadata key for MCP creds).  The list endpoint already
-        # normalizes both flows to full URL via ``get_host``, so the matching
-        # logic in ``MCPSetupCard.liveHasCred`` works either way — but
-        # returning the same shape from POST keeps the contract consistent
-        # for any consumer that uses the immediate POST response.
-        host=credentials.metadata.get("mcp_server_url"),
-    )
+    return to_meta_response(credentials)
```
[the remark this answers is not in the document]
Abhi1992002: Intended: storing a static token as a first-class api_key credential is the core of this PR, so type becoming "api_key" (and scopes null) is correct. All in-tree consumers match MCP creds by host, not type, and were verified. The field was already typed as the CredentialsType union, so the schema is unchanged.

autogpt_platform/frontend/src/hooks/useCredentials.ts
```diff
@@ -27,8 +27,9 @@ export function classifyCredentials(
   for (const c of allSaved) {
     if (!supportedTypes.includes(c.type)) continue;
 
-    // MCP OAuth2 credentials filter by server URL — not upgradeable
-    if (c.type === "oauth2" && c.provider === "mcp") {
+    // MCP credentials (OAuth2 tokens and static API-key / bearer tokens)
+    // filter by server URL — not upgradeable.
+    if ((c.type === "oauth2" || c.type === "api_key") && c.provider === "mcp") {
```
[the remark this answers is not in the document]
Abhi1992002: This is a deliberate one-line inline predicate mirrored in the sibling classifier; both branches are covered by tests (classifyCredentials.test.ts and CredentialsGroupedView/__tests__/helpers.test.ts). Extracting a shared helper for a two-site type check isn't warranted here; can revisit if a third MCP credential type is added.

autogpt_platform/frontend/src/components/contextual/CredentialsInput/components/CredentialsGroupedView/helpers.ts
```diff
@@ -28,8 +28,13 @@ function matchesDiscriminatorValues(
   credential: { host?: string | null; provider: string; type: string },
   discriminatorValues?: string[],
 ) {
-  // MCP OAuth2 credentials must match by server URL
-  if (credential.type === "oauth2" && credential.provider === "mcp") {
+  // MCP credentials (OAuth2 tokens and static API-key / bearer tokens) must
+  // match by server URL — otherwise a token for one server would be treated
+  // as valid for every server.
+  if (
+    (credential.type === "oauth2" || credential.type === "api_key") &&
```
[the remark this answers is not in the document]
Abhi1992002: Same as the useCredentials.ts counterpart — intentional inline mirror, both covered by tests. Consolidating a two-site one-line predicate isn't warranted for this PR.

autogpt_platform/frontend/src/app/(platform)/build/components/MCPToolDialog.tsx
```diff
@@ -115,17 +116,48 @@ export function MCPToolDialog({
...
+        const tokenRes = await postV2StoreABearerTokenForAnMcpServer({
+          server_url: url,
+          token: authToken,
+        });
+        if (tokenRes.status !== 200)
```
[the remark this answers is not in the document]
Abhi1992002: Low-impact: this branch is only hit if the backend /mcp/token call fails after a successful discovery (rare). The user sees a clear, token-specific error and can retry; keeping the failure hard (no block added) avoids placing a block with no usable credential. Preserving the fetched tool list across a persist retry is a nice-to-have follow-up.

autogpt_platform/backend/backend/api/features/mcp/routes.py
```diff
@@ -437,27 +440,27 @@ async def mcp_store_token(
     server_url = normalize_mcp_url(request.server_url)
     hostname = server_host(server_url)
 
-    # Collect IDs of old credentials to clean up after successful create.
+    # Collect IDs of old credentials to clean up after successful create —
```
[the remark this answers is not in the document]
Abhi1992002: As noted on the earlier CodeRabbit thread: the new credential is created before any old rows are deleted, so a mid-sequence failure never leaves the user with zero credentials, and the rare concurrent-double-submit duplicate is tolerated by auto_lookup_mcp_credential (deterministic best-match selection, now with a tested recency tiebreaker). A store-level atomic replace is a cross-cutting change to IntegrationCredentialsStore, out of scope here.

autogpt_platform/backend/backend/api/features/mcp/routes.py
```diff
@@ -361,16 +366,14 @@ async def mcp_oauth_callback(
...
     try:
         old_creds = await creds_manager.store.get_creds_by_provider(
             user_id, ProviderName.MCP.value
         )
         for old in old_creds:
-            if (
-                isinstance(old, OAuth2Credentials)
-                and (old.metadata or {}).get("mcp_server_url") == meta["server_url"]
-            ):
+            if is_mcp_credential_for_server(old, normalize_mcp_url(meta["server_url"])):
```
0ubbe: `is_mcp_credential_for_server()` is called with `normalize_mcp_url(meta["server_url"])` here, but in the same function at line 376 (Hunk 2) it was already called with the normalized URL. Verify that `meta["server_url"]` is already normalized at this point (it is set at line 269 from the state token), or this may fail to match credentials that were stored with a trailing slash variant.

autogpt_platform/backend/backend/api/features/mcp/routes.py
```diff
@@ -437,27 +440,27 @@ async def mcp_store_token(
...
         old_creds = await creds_manager.store.get_creds_by_provider(
             user_id, ProviderName.MCP.value
         )
         old_cred_ids = [
-            old.id
-            for old in old_creds
-            if isinstance(old, OAuth2Credentials)
-            and normalize_mcp_url((old.metadata or {}).get("mcp_server_url", ""))
-            == server_url
+            old.id for old in old_creds if is_mcp_credential_for_server(old, server_url)
```
0ubbe, line 463: Line 451 calls `is_mcp_credential_for_server(old, server_url)` where `server_url` is already normalized (line 440). This is correct and consistent with the pattern, but verify that `is_mcp_credential_for_server()` itself does not re-normalize the URL internally, which could cause double-normalization or unexpected mismatches.

autogpt_platform/backend/backend/api/features/mcp/routes.py
0ubbe, line 487: `to_meta_response()` is called on an `APIKeyCredentials` object, but the function was previously only used with `OAuth2Credentials` in the OAuth callback path. Verify that `to_meta_response()` correctly handles `APIKeyCredentials` and extracts the `mcp_server_url` from metadata the same way the removed inline code did (line 487 in the old code).

autogpt_platform/backend/backend/blocks/mcp/block.py
```diff
@@ -240,9 +244,7 @@ async def run(
                 user_id, normalize_mcp_url(input_data.server_url)
             )
 
-        auth_token = (
-            credentials.access_token.get_secret_value() if credentials else None
-        )
+        auth_token = mcp_auth_token(credentials) if credentials else None
```
0ubbe, line 244: The old code assumed `credentials` always has an `access_token` attribute. If `mcp_auth_token()` receives an `APIKeyCredentials` object that lacks `access_token`, the helper must safely extract the token from the `api_key` field instead. Verify that `mcp_auth_token()` handles both credential shapes without raising `AttributeError`.

autogpt_platform/backend/backend/api/features/integrations/router.py
```diff
@@ -157,7 +157,9 @@ def get_host(cred: Credentials) -> str | None:
         """Extract host from credential: HostScoped host or MCP server URL."""
         if isinstance(cred, HostScopedCredentials):
             return cred.host
-        if isinstance(cred, OAuth2Credentials) and cred.provider in (
+        if isinstance(
```
0ubbe, line 257: The `APIKeyCredentials` type is imported at line 32, but the isinstance check at line 161 now includes it. Verify that `APIKeyCredentials` objects always have a `provider` attribute and can have `metadata` with an `mcp_server_url` key, matching the structure expected for OAuth2 MCP credentials. If the schema differs, this will silently return `None` instead of the URL.

autogpt_platform/backend/backend/blocks/mcp/test_helpers.py
```diff
@@ -96,3 +109,160 @@ def test_parse_multi_item():
...
+        metadata={"mcp_server_url": _MCP_URL},
+    )
+    api_key = _api_key_cred(_MCP_URL)
+
+    ctx, mgr = _patched_manager([expired_oauth, api_key])
+    with ctx:
+        result = await auto_lookup_mcp_credential("user", _MCP_URL)
+
+    assert result is api_key
+    mgr.refresh_if_needed.assert_not_called()
```
0ubbe, line 220: `test_auto_lookup_prefers_non_expiring_api_key_over_expired_oauth()` at line 200 asserts `mgr.refresh_if_needed.assert_not_called()`, but the test does not verify that `auto_lookup_mcp_credential()` actually *selected* the API key over the expired OAuth credential. If the function returned the expired OAuth credential instead, the assertion would still pass. Add an explicit `assert result is api_key` before the refresh check to confirm the selection logic.

autogpt_platform/backend/backend/blocks/mcp/test_mcp.py
```diff
@@ -594,6 +594,43 @@ async def mock_call(server_url, tool_name, arguments, auth_token=None):
...
+            return "ok"
+
+        block._call_mcp_tool = mock_call  # type: ignore
+
+        test_creds = APIKeyCredentials(
+            id="cred-apikey-1",
+            provider="mcp",
+            api_key=SecretStr("static-bearer-token"),
+            title="MCP: mcp.example.com",
+            metadata={"mcp_server_url": "https://mcp.example.com/mcp"},
```
0ubbe, line 622: `APIKeyCredentials` is instantiated with `metadata={"mcp_server_url": "https://mcp.example.com/mcp"}`, but the test does not verify that the block actually uses this metadata to match or validate the credential against the input `server_url`. The PR description states credentials are matched by server URL; this test should assert that a credential with a *mismatched* server URL in metadata is rejected or ignored, to ensure the matching logic is sound.

autogpt_platform/backend/backend/copilot/tools/utils.py
```diff
@@ -485,13 +486,18 @@ def _credential_is_for_mcp_server(
...
+    is how servers that don't support OAuth (e.g. a token issued in the
+    vendor's dashboard) store their credential.
+    """
     if not requirements.discriminator_values:
         return True
 
     server_url = (
         credential.metadata.get("mcp_server_url")
-        if isinstance(credential, OAuth2Credentials)
+        if isinstance(credential, (OAuth2Credentials, APIKeyCredentials))
```
0ubbe, line 532: The function now checks `isinstance(credential, (OAuth2Credentials, APIKeyCredentials))` but does not validate that `APIKeyCredentials` actually has the `mcp_server_url` metadata key set. If an API-key credential is created without this metadata, `credential.metadata.get("mcp_server_url")` will return `None`, causing the match to fail silently. The PR description states that `/api/v2/mcp/token` stores the token as `APIKeyCredentials` with this metadata, but there is no defensive check here if that contract is broken upstream.

autogpt_platform/frontend/src/app/(platform)/build/components/__tests__/MCPToolDialog.test.tsx
```diff
@@ -0,0 +1,201 @@
...
+
+const mockDiscover = vi.fn();
+const mockStoreToken = vi.fn();
+const mockOAuthLogin = vi.fn();
+vi.mock("@/app/api/__generated__/endpoints/mcp/mcp", () => ({
+  postV2DiscoverAvailableToolsOnAnMcpServer: (...args: unknown[]) =>
+    mockDiscover(...args),
+  postV2InitiateOauthLoginForAnMcpServer: (...args: unknown[]) =>
+    mockOAuthLogin(...args),
+  postV2ExchangeOauthCodeForMcpTokens: vi.fn(),
```
0ubbe: Line 23 mocks `postV2ExchangeOauthCodeForMcpTokens` with `vi.fn()` but never resets it in `afterEach`. If any test accidentally triggers OAuth code exchange, the mock state will leak into subsequent tests. Add `mockExchangeCode` as a module-level const and reset it in `afterEach` alongside the other mocks.

autogpt_platform/frontend/src/app/(platform)/build/components/__tests__/MCPToolDialog.test.tsx
```diff
@@ -0,0 +1,201 @@
...
+    );
+
+    // Land on the tool-selection step, pick a tool, add the block.
+    const toolButton = await screen.findByText("get_analytics");
+    fireEvent.click(toolButton);
+    fireEvent.click(screen.getByRole("button", { name: /add block/i }));
+
+    expect(onConfirm).toHaveBeenCalledTimes(1);
+    const result = onConfirm.mock.calls[0][0];
+    expect(result.credentials).toEqual({
```
0ubbe, line 106: Line 102–107 asserts that `result.credentials` matches the shape returned by `mockStoreToken`, but the test does not verify that the `host` field is stripped from the credential object before attachment to the block. The PR description states credentials are 'attached' to the block node; if `host` leaks into the block's credential object, it could cause issues at runtime. Verify the expected shape excludes `host`.

autogpt_platform/frontend/src/app/(platform)/build/components/MCPToolDialog.tsx
```diff
@@ -115,22 +116,54 @@ export function MCPToolDialog({
...
```
0ubbe: When `postV2StoreABearerTokenForAnMcpServer` fails with non-200 status, the error is thrown but `setLoading(false)` is never called before the throw, so the loading spinner remains stuck. The `finally` block will eventually clear it, but the user sees a brief inconsistent state. Consider calling `setLoading(false)` before throwing, or restructure to ensure cleanup happens synchronously.

autogpt_platform/frontend/src/app/(platform)/build/components/MCPToolDialog.tsx
```diff
@@ -115,22 +116,54 @@ export function MCPToolDialog({
...
+      if (err?.status === 401 || err?.status === 403) {
+        setLoading(false);
+        // If the user supplied a token and the server still rejected it, the
+        // token is wrong — don't bounce them into an OAuth flow that will just
+        // fail again. Surface a clear "check your token" message instead.
+        if (authToken) {
+          setError(
+            "Authentication failed. Please check your API key / bearer token and try again.",
+          );
+          return;
```
0ubbe: The early `return` at line 157 after setting an error for invalid tokens skips `setLoading(false)`, leaving the loading state true while showing an error message. This contradicts the pattern in the OAuth path (line 265) where `setLoading(false)` is in the `finally` block. The inconsistency means the user sees a loading spinner alongside the error text.

autogpt_platform/frontend/src/app/(platform)/build/components/MCPToolDialog.tsx
```diff
@@ -304,26 +337,45 @@ export function MCPToolDialog({
               />
             </div>
 
-            {/* Auth required: show manual token option */}
-            {authRequired && !showManualToken && (
+            {/* Manual token option — available up front for servers that
+                authenticate with a static API key / bearer token rather than
+                OAuth (e.g. a token issued in the vendor's own dashboard). */}
+            {!showManualToken && (
```
0ubbe, line 370: The button condition changed from `authRequired && !showManualToken` to just `!showManualToken`. This means the button is always shown initially, even for servers that support OAuth. While this matches the PR intent (proactive entry), it changes the UX for OAuth-capable servers—users now see two auth paths simultaneously instead of OAuth-first. Verify this is intentional and doesn't confuse users.

autogpt_platform/frontend/src/app/(platform)/build/components/MCPToolDialog.tsx
```diff
@@ -373,14 +429,21 @@ export function MCPToolDialog({
...
+                (showManualToken && !manualToken.trim())
+              }
             >
               {loading || oauthLoading ? (
                 <span className="flex items-center gap-2">
                   <LoadingSpinner className="size-4" />
                   {oauthLoading ? "Waiting for sign-in..." : "Connecting..."}
                 </span>
-              ) : authRequired && !showManualToken ? (
+              ) : showManualToken ? (
```
0ubbe, line 479: The button condition changed from `authRequired && !showManualToken` to just `!showManualToken`. This means the button is always shown initially, even for servers that support OAuth. While this matches the PR intent (proactive entry), it changes the UX for OAuth-capable servers—users now see two auth paths simultaneously instead of OAuth-first. Verify this is intentional and doesn't confuse users.

autogpt_platform/frontend/src/hooks/useCredentials.ts
0ubbe, line 35: The condition `(c.type === "oauth2" || c.type === "api_key") && c.provider === "mcp"` will now match `api_key` credentials with `provider === "mcp"`, but there is no validation that `c.host` is populated for `api_key` MCP credentials. If the backend stores `api_key` MCP credentials without a `host` field, the `discriminatorValue != null && c.host === discriminatorValue` check on line 33 will silently fail to match them (comparing `undefined === discriminatorValue`). Confirm the backend always populates `host` for `api_key` MCP credentials.

autogpt_platform/frontend/src/components/contextual/CredentialsInput/components/CredentialsGroupedView/helpers.ts
```diff
@@ -28,8 +28,13 @@ function matchesDiscriminatorValues(
...
+    credential.provider === "mcp"
+  ) {
     if (!discriminatorValues || discriminatorValues.length === 0) return false;
```
0ubbe, line 39: The logic correctly requires `discriminatorValues` to be non-empty for MCP credentials (both types), preventing a token for one server from being treated as valid for all servers. However, this creates an asymmetry: `host_scoped` credentials (line 44) return `true` when `discriminatorValues` is empty, while MCP credentials return `false`. Verify this asymmetry is intentional — if an MCP credential block is placed without a server URL discriminator, should it be considered a match or not?

0ubbe (commented):
Some nits 💭 💜

autogpt_platform/frontend/src/app/(platform)/build/components/__tests__/MCPToolDialog.test.tsx
Abhi1992002: 🤖 Fixed in [a7a3b3c84](https://github.com/Significant-Gravitas/AutoGPT/commit/a7a3b3c8491de2c803248152d0179135fff604e2): `postV2ExchangeOauthCodeForMcpTokens` is now a module-level `mockExchangeCode` const, reset in `afterEach` alongside the other mocks.

autogpt_platform/backend/backend/api/features/mcp/routes.py
Abhi1992002: 🤖 Not applicable — the call site already normalizes: `is_mcp_credential_for_server(old, normalize_mcp_url(meta["server_url"]))`, and the helper additionally normalizes the *stored* metadata value (`normalize_mcp_url(stored)` in `helpers.py:56`). Both sides are normalized, so trailing-slash variants match.

autogpt_platform/backend/backend/api/features/mcp/routes.py
Abhi1992002: 🤖 Verified — `is_mcp_credential_for_server()` normalizes only the *stored* metadata value internally (`helpers.py:55-56`), never the passed `server_url`. Passing the already-normalized `server_url` (line 440) causes no double-normalization.

autogpt_platform/backend/backend/api/features/mcp/routes.py
Abhi1992002: 🤖 Verified — `to_meta_response()` derives `host` via `CredentialsMetaResponse.get_host()`, which returns `metadata["mcp_server_url"]` for both `OAuth2Credentials` and `APIKeyCredentials` (`router.py:156-168`). The `APIKeyCredentials` built at `routes.py:460` sets exactly that metadata, so host resolves correctly.

autogpt_platform/backend/backend/blocks/mcp/block.py
Abhi1992002: 🤖 Verified — `mcp_auth_token()` (`helpers.py:33-41`) branches on `isinstance(cred, APIKeyCredentials)` → returns `api_key`, else `access_token`. No `AttributeError` for either shape.

autogpt_platform/backend/backend/api/features/integrations/router.py
Abhi1992002: 🤖 Verified — `APIKeyCredentials` extends the same base as `OAuth2Credentials`, so it always has `provider` and a `metadata` dict; `/mcp/token` stores `mcp_server_url` there. If ever absent, `.get()` returns `None` and `get_host()` returns `None` — safe, no crash.

autogpt_platform/backend/backend/blocks/mcp/test_helpers.py
Abhi1992002: 🤖 Already present — `assert result is api_key` is at `test_helpers.py:217`, immediately before the `refresh_if_needed.assert_not_called()` check. The selection is verified.

autogpt_platform/backend/backend/blocks/mcp/test_mcp.py
Abhi1992002: 🤖 In the block `run()` path the credential is *injected by the executor* (already selected by server URL upstream); the block never re-matches by URL. The URL-matching logic is covered by `test_helpers.py` (`is_mcp_credential_for_server`) and `copilot/tools/utils.py` (`_credential_is_for_mcp_server`). This test only asserts the token is pulled from `api_key` rather than `access_token`.

autogpt_platform/backend/backend/copilot/tools/utils.py
Abhi1992002: 🤖 Not applicable — when `mcp_server_url` metadata is missing, `server_url` is `None` and the final expression returns `False` (`... if server_url else False`, `utils.py:503`). The credential is *not* matched, which is the safe outcome — no silent wrong-server match can occur.

autogpt_platform/frontend/src/app/(platform)/build/components/__tests__/MCPToolDialog.test.tsx
Abhi1992002: 🤖 Already verified — `expect(result.credentials).toEqual({...})` (line 102) omits `host`, while `mockStoreToken` returns data *with* `host` (line 63). `toEqual` is exact, so the test fails if `host` leaks into the attached credential. Stripping is covered.

autogpt_platform/frontend/src/app/(platform)/build/components/MCPToolDialog.tsx
Abhi1992002: 🤖 The thrown `Error` is caught by the same `try/catch`; it has no `status`, so it falls to the else branch (sets the error message) and then `finally` runs `setLoading(false)`. catch→finally executes synchronously in the same microtask, so React batches the updates — there is no intermediate render with a stuck spinner.

autogpt_platform/frontend/src/app/(platform)/build/components/MCPToolDialog.tsx
Abhi1992002: 🤖 Not applicable — `setLoading(false)` is already called at line 149, at the top of the 401/403 branch, *before* the early `return` at line 157. The spinner is cleared. The OAuth path uses `finally` only because it has no such early exit.

autogpt_platform/frontend/src/app/(platform)/build/components/MCPToolDialog.tsx
Abhi1992002: 🤖 Intentional — this is the PR's "proactive API-key entry" goal: users can choose the token flow without a failed OAuth round-trip first. OAuth sign-in remains available; the two paths coexist by design (see PR description).

autogpt_platform/frontend/src/app/(platform)/build/components/MCPToolDialog.tsx
Abhi1992002: 🤖 Intentional — same proactive-entry UX as the sibling button; users may pick the token flow up front without a failed OAuth round-trip. OAuth remains available.

autogpt_platform/frontend/src/hooks/useCredentials.ts
Abhi1992002: 🤖 Confirmed — the backend always populates `host` for `api_key` MCP creds: `CredentialsMetaResponse.get_host()` (`router.py:156-168`) returns `metadata["mcp_server_url"]` for `APIKeyCredentials`, and `/mcp/token` stores that metadata. So `c.host` is set and the `c.host === discriminatorValue` check matches.

autogpt_platform/frontend/src/components/contextual/CredentialsInput/components/CredentialsGroupedView/helpers.ts
Abhi1992002: 🤖 The asymmetry is intentional. An MCP token is scoped to a single server, so with no discriminator we must *not* treat it as valid for every server (returns `false`). `host_scoped` creds are broader and default to matching when unscoped. Matching an MCP token with no server discriminator would leak one server's token to all servers.

0ubbe (dismissed):

Abhi1992002:
!deploy

Pwuts:

- **Deployed:** `ecfe29b032368d148df7a15d9e5c4d81baf72c7c` at 2026-07-28 14:57 UTC
- **Database:** isolated Supabase branch `pr-13683` (state persists across redeploys unless migration drift forces a reset)

Abhi1992002:
!undeploy

Pwuts:

All resources for PR #13683 have been removed:

autogpt_platform/backend/backend/blocks/mcp/block.py
```diff
@@ -232,17 +219,12 @@ async def run(
...
-
-        auth_token = (
-            credentials.access_token.get_secret_value() if credentials else None
-        )
+        # No auto-lookup fallback here: the executor injects whatever the node
+        # has selected, so `credentials is None` means the user picked "None
+        # (skip this credential)" or the server needs no auth. Silently
+        # substituting a stored token would override that explicit choice —
+        # the two cases are indistinguishable by the time they reach `run`.
+        auth_token = mcp_auth_token(credentials) if credentials else None
```
kcze, line 244: 🤖 🔴 **Blocker**: Dropping the `auto_lookup_mcp_credential` fallback silently breaks MCP nodes that don't carry an explicit credential ID. `credentials` has schema default `{}`, and `executor/manager.py:282-300` turns an empty *or stale* value into `input_data["credentials"] = None` and runs the block anyway (`ValueError` → "running without (field has default)"). So `credentials is None` in `run()` is **not** the same as "user chose None" — it is also "never set" and "ID points at a deleted row".

Failure: a user reconnects an existing server via `/mcp/token` or the OAuth callback → both delete the old credential and create a replacement with a **new ID**, while the placed node still references the deleted one → every subsequent run sends no `Authorization` header and fails with 401. That is exactly the bug this PR sets out to fix. Same for graphs saved before the picker populated `credentials`; there is no backfill of `node.input_default["credentials"]`.

_(flagged by: Codex + Claude)_

autogpt_platform/frontend/src/app/(platform)/build/components/MCPToolDialog.tsx
```diff
@@ -115,22 +117,54 @@ export function MCPToolDialog({
         auth_token: authToken || null,
       });
       if (response.status !== 200) throw response.data;
+
+      // When the user authenticated with a static API key / bearer token,
+      // persist it so the block can authenticate at runtime — discovery only
+      // uses the token transiently and would otherwise be discarded, leaving
+      // the placed block with no credentials and failing on the next run.
+      if (authToken) {
```
kcze: 🤖 🔴 **Blocker**: One server's secret can be sent to — and stored for — a different server. On success the dialog calls `setShowManualToken(false)` (L145) but never clears `manualToken`, and `handleDiscoverTools` (L179) keeps passing it.

Failure: connect to server A with token `T_A` → "Back" (L418, resets only `step`/`selectedTool`) → type server B's URL. The token field is hidden and the button reads "Discover Tools", so nothing signals a token is still loaded — but `authToken` is truthy, so `T_A` goes out in B's discovery request *and* this new block persists it via `/mcp/token` as B's credential.

Clearing `manualToken` alongside `setShowManualToken(false)` (and on "Back") fixes it. _(flagged by: Claude)_

autogpt_platform/frontend/src/app/(platform)/build/components/MCPToolDialog.tsx
```diff
@@ -115,22 +117,54 @@ export function MCPToolDialog({
...
+        const tokenRes = await postV2StoreABearerTokenForAnMcpServer({
+          server_url: url,
+          token: authToken,
+        });
+        if (tokenRes.status !== 200)
+          throw new Error(
+            "Discovered the server, but saving your API token failed. Please try again.",
+          );
+        setCredentials({
```
kcze: 🤖 🔴 **Blocker**: The stored credential never reaches `CredentialsProvidersContext`. The OAuth path deliberately goes through `mcpProvider.mcpOAuthCallback` — the code comments it as "updates cache" (L211) — but this path calls the generated endpoint directly, so nothing upserts the new credential and no query invalidation fires.

Failure: the backend's `mcp_store_token` deletes the previous credential for that server and returns a new ID. The node is set to the new ID, but `useCredentialsInput` still sees the *old, now-deleted* credential in the cached list and the new ID missing → it clears the selection and re-selects the deleted credential. The node is saved pointing at a nonexistent credential and 401s until a full page refresh.

_(flagged by: Codex + Claude)_

autogpt_platform/backend/backend/api/features/mcp/routes.py
kcze: 🤖 🟠 **Should Fix**: In `mcp_oauth_callback` this cleanup loop runs **before** `await creds_manager.create(...)` (L385), and broadening it to `api_key` means it now destroys a *working* static bearer token before the replacement exists.

Failure: a user with a working API-key credential switches that server to OAuth; the exchange succeeds, the key is deleted, then `create` fails → the user is left with no credential at all and must re-obtain the token. `mcp_store_token` already gets this right (create first, delete old IDs after) — worth mirroring that ordering here.

Note this is distinct from the atomicity thread above: the "create-before-cleanup ordering avoids credential loss" reasoning holds for `/token`, but this path is delete-before-create. _(flagged by: Codex)_

autogpt_platform/backend/backend/blocks/mcp/helpers.py
kcze: 🤖 🟠 **Should Fix**: Follow-up on the existing `sys.maxsize` thread — the reason it was triaged low ("only if best-effort cleanup fails") no longer holds after this PR.

By making the block advertise `api_key`, MCP credentials can now also be created through the generic `POST /{provider}/credentials` (`api/features/integrations/router.py:444-466`) via the API-key modal — and that endpoint does **no** per-server cleanup at all. So an OAuth credential and a static key for the same server coexist through a completely normal user flow, not a failure path.

Failure: user OAuths into `https://mcp.acme.com/mcp`, later pastes a personal token through the block's API-key modal. `auto_lookup_mcp_credential` then permanently prefers the static key, so `/discover-tools` and CoPilot's `run_mcp_tool` authenticate with it even after the token is revoked — and reconnecting via OAuth cannot fix it. _(flagged by: Claude)_

autogpt_platform/backend/backend/copilot/tools/utils.py
kcze, line 532: 🤖 🟡 **Nice to Have**: This predicate now covers `api_key` credentials but still compares the raw stored `mcp_server_url` against raw `requirements.discriminator_values` — no `normalize_mcp_url`, unlike the new `is_mcp_credential_for_server` and the frontend's `classifyCredentials`, which normalize both sides.

Failure: the node's `server_url` is `https://mcp.acme.com/mcp/` (trailing slash) while the credential was stored normalized without it → `match_user_credentials_to_graph` reports the MCP credential as missing and CoPilot refuses to run an agent the user has already connected. _(flagged by: Claude)_

autogpt_platform/frontend/src/components/contextual/CredentialsInput/components/CredentialsGroupedView/helpers.ts
kcze, line 36: 🤖 🟡 **Nice to Have**: `matchesDiscriminatorValues` compares `discriminatorValues.includes(credential.host)` verbatim (L40), while `classifyCredentials` (`useCredentials.ts:39`) now normalizes both sides with `normalizeMCPUrl`. The two MCP matchers disagree.

Failure: a node whose `server_url` carries a trailing slash matches in the builder picker but not in the grouped/run-dialog auto-assign, so the run dialog reports no credential for a server that has one. _(flagged by: Claude)_

autogpt_platform/frontend/src/components/contextual/CredentialsInput/components/APIKeyCredentialsModal/useAPIKeyCredentialsModal.ts
```diff
@@ -52,6 +53,11 @@ export function useAPIKeyCredentialsModal({
     },
   });
 
+  const mcpServerUrl =
```
kcze: 🤖 🟡 **Nice to Have**: When the provider is `mcp` but `discriminatorValue` is empty (node's `server_url` not yet filled in), `mcpServerUrl` is `""` and the `metadata` spread is silently dropped.

Failure: the modal creates an MCP credential with `host: null`, which can never be matched by the picker or by the backend's `is_mcp_credential_for_server`. The user sees the credential as selected, yet the block 401s at run time with no indication why. Worth blocking submission (or surfacing an error) instead of creating an unusable credential. _(flagged by: Claude)_

autogpt_platform/frontend/src/lib/autogpt-server-api/types.ts
```diff
@@ -671,6 +671,7 @@ export type APIKeyCredentials = BaseCredentials & {
   title: string;
   api_key: string;
   expires_at?: number;
+  metadata?: Record<string, any>;
```
kcze: 🤖 🟡 **Nice to Have**: `Record<string, any>` violates the repo rule "Never type with `any`, if not types available use `unknown`" (`frontend/AGENTS.md`). `Record<string, unknown>` compiles for both call sites, which only ever write `{ mcp_server_url: string }`. _(flagged by: Claude)_

autogpt_platform/backend/backend/blocks/mcp/test_mcp.py
```diff
@@ -594,6 +596,43 @@ async def mock_call(server_url, tool_name, arguments, auth_token=None):
 
         assert captured_tokens == ["resolved-token"]
 
+    @pytest.mark.asyncio(loop_scope="session")
+    async def test_run_with_api_key_credentials(self):
+        """Verify the block accepts a static API-key / bearer token and pulls
+        the token from ``api_key`` (not ``access_token``)."""
+        from pydantic import SecretStr
```
kcze: 🤖 🟡 **Nice to Have**: Function-local imports violate `backend/AGENTS.md` ("Top-level imports only — no local/inner imports"), and these two are redundant: this same PR already added `SecretStr` and `APIKeyCredentials` at module scope (L9, L13). Same pattern in `test_helpers.py:168` and `copilot/tools/utils_test.py:37-40`. _(flagged by: Claude)_

autogpt_platform/backend/backend/blocks/mcp/block.py
Abhi1992002: Fixed in d2b21a5796: restored the fallback. You're right that `credentials is None` conflates three cases — `executor/manager.py` nulls the field for "user picked None", "never configured", *and* "ID points at a deleted row", and there's no way to tell them apart by the time it reaches `run()`. Reconnecting a server deletes the old credential and issues a new ID, so that third case is an ordinary flow, and without the fallback every subsequent run 401s — the exact bug this PR exists to fix. Falling back can override an explicit "None" on a server the user also has a token for, which is the far milder of the two failures; the comment on the call site spells out that trade-off. `test_run_without_credentials_ignores_stored_token` was inverted to `..._falls_back_to_stored_token` accordingly.

autogpt_platform/frontend/src/app/(platform)/build/components/MCPToolDialog.tsx
Abhi1992002: Fixed in d2b21a5796: `manualToken` is now cleared in two places — right after the token is persisted (it lives server-side from then on) and in the "Back" handler, which also clears `credentials`, `showManualToken`, `authRequired` and `error` since all of it is scoped to the server just connected.

autogpt_platform/frontend/src/app/(platform)/build/components/MCPToolDialog.tsx
Abhi1992002: Fixed in d2b21a5796: added `mcpStoreToken` to the credentials provider and routed the dialog through it, mirroring the OAuth path. It also does more than a plain upsert — `replaceMCPServerCredentials` drops the cached rows for that server first, because the backend deletes them. `mcpOAuthCallback` had the same latent problem and now uses the same helper.

autogpt_platform/backend/backend/api/features/mcp/routes.py
Abhi1992002: Fixed in d2b21a5796: reordered to collect the old IDs, `create`, then delete — same shape as `mcp_store_token`. A failed create now leaves the working credential in place.

autogpt_platform/backend/backend/blocks/mcp/helpers.py
Abhi1992002: Fixed in d2b21a5796: `_mcp_credential_rank` now ranks on usability (1 = still authenticates, 0 = lapsed) instead of expiry timestamp, with store order breaking ties. A fresh OAuth token and a static key both rank 1, so the more recently created one wins and re-authenticating takes effect. An expired access token that carries a refresh token still ranks 1 since `refresh_if_needed` renews it. Covered by `test_auto_lookup_prefers_fresh_oauth_over_older_api_key`; the existing expired-OAuth-vs-key test still passes.

autogpt_platform/backend/backend/copilot/tools/utils.py
Abhi1992002: Fixed in d2b21a5796: both sides now go through `normalize_mcp_url`. Added `test_credential_is_for_mcp_server_normalizes_trailing_slash`.

autogpt_platform/frontend/src/components/contextual/CredentialsInput/components/CredentialsGroupedView/helpers.ts
Abhi1992002: Fixed in d2b21a5796: `matchesDiscriminatorValues` normalizes both sides now, so it agrees with `classifyCredentials`.

autogpt_platform/frontend/src/components/contextual/CredentialsInput/components/APIKeyCredentialsModal/useAPIKeyCredentialsModal.ts
Abhi1992002: Fixed in d2b21a5796: submission is blocked with "Enter the MCP server URL on the block before adding a key." (rendered from `formState.errors.root`) rather than creating a credential with `host: null`.

Worth flagging: merging latest dev surfaced a second path for this. The new `ConnectCredentialDialog` (`variant="default"`, i.e. run dialogs and copilot) creates API keys via `useApiKeyConnectForm` → `postV1CreateCredentials` with no metadata at all, so it produced exactly the `host: null` credential you describe. That now threads `mcp_server_url` through too, with tests in `ConnectCredentialDialog.test.tsx`.

autogpt_platform/frontend/src/lib/autogpt-server-api/types.ts
Abhi1992002: Fixed in d2b21a5796: `Record<string, unknown>`.

autogpt_platform/backend/backend/blocks/mcp/test_mcp.py
Abhi1992002: Fixed in d2b21a5796: hoisted in all three files. (`test_mcp.py:565` predates this PR, so I left it alone.)

autogpt_platform/frontend/src/app/(platform)/build/components/MCPToolDialog.tsx
```diff
@@ -107,39 +109,93 @@ export function MCPToolDialog({
...
+            throw new Error(
+              "Discovered the server, but saving your API token failed. Please try again.",
+            );
+          }
+          setCredentials({
+            id: stored.id,
+            provider: stored.provider,
+            type: stored.type,
+            title: stored.title,
+          });
```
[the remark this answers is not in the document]
Abhi1992002: Fixed in a31d9acd3e: valid — `mcpStoreToken` toasted *and* rethrew, and its only caller then rendered the same failure inline.

Kept the inline message rather than the toast: it sits next to the token field the user has to correct, and it is what tells them discovery succeeded but the save didn't. `mcpStoreToken` now just rethrows. Its sibling `mcpOAuthCallback` keeps its toast, since that path has callers with no inline error surface.

kcze:
This is already fixed in https://github.com/Significant-Gravitas/AutoGPT/pull/14075, closing

PR #13685, opened 2026-07-27
feat(backend): add Expert entity, hire/install API, roster seeds

0ubbe:
### Why / What / How

**Why:** First slice of the "Hire Experts" MVP (Hire-or-Raise product direction): users will hire curated Experts from the marketplace, see them on a Team page, and chat with them in per-expert CoPilot threads that run installed Workflows. This PR lays the backend foundation — the `Expert` entity, the hire/install API, and the dev roster — so the copilot and frontend slices can build on it. Ships dark; nothing user-facing changes.

**What:**
- New first-class `Expert` Prisma model (one table for roster templates and hired instances) + `ExpertWorkflow` join to `LibraryAgent`/`StoreListingVersion`, and a nullable `ChatSession.expertId` column. Migration is purely additive.
- New feature module `backend/api/features/experts/`: data layer (`experts_db.py`), Pydantic models, REST routes mounted at `/api/experts`, and a dev roster seed script (Maria/Marketing, Max/Sales, Frankie/Ops).

**How:**
- Hire resolves each template preload through the existing `add_store_agent_to_library` seam, producing the hiring user's `LibraryAgent` plus an `ExpertWorkflow` instance row. Hire is idempotent per (user, template) — enforced by `@@unique([ownerUserId, sourceTemplateId])` and a find-first in `hire_expert`. Duplicate workflow install returns the existing row.
- Partial hire is honest: a failed preload install is caught per-item and reported in `HireResult.failed_preloads`; the hire itself succeeds.
- All non-template queries are scoped by `ownerUserId` (+ tenancy trio columns carried for org-readiness); templates are never owned and never chat targets. Auth via `Security(autogpt_auth_lib.get_user_id)` like the library routes.
- Endpoints (operation IDs drive the frontend Orval hooks in the follow-up PR): `GET /api/experts/templates`, `POST /api/experts` (hire), `GET /api/experts`, `GET /api/experts/{id}`, `POST /api/experts/{id}/workflows`, `DELETE /api/experts/{id}` (archive → 204).

### Changes 🏗️

- `backend/schema.prisma`: `Expert`, `ExpertWorkflow` models; `ChatSession.expertId` (nullable); back-relations on `User`, `Team`, `LibraryAgent`, `StoreListingVersion`
- `backend/migrations/20260727070306_add_expert_entity/` — additive only (2 CREATE TABLE, 1 ALTER TABLE ADD COLUMN, indexes + FKs)
- `backend/backend/api/features/experts/models.py` — `Expert`, `ExpertWorkflowRef`, `HireResult`
- `backend/backend/api/features/experts/experts_db.py` — `list_templates`, `list_experts`, `get_expert`, `hire_expert` (idempotent), `install_workflow` (duplicate-safe), `archive_expert`
- `backend/backend/api/features/experts/routes.py` + mount in `backend/backend/api/rest_api.py`
- `backend/backend/api/features/experts/seed.py` — dev roster upsert by template name; preloads resolved from real dev marketplace listing slugs; missing listings logged and skipped, never fatal
- Tests: `experts_db_test.py` (5, real-DB), `routes_test.py` (10, mocked db + snapshot)

Note on user-ID scoping: this PR touches `api/features/experts/experts_db.py` (a data-layer module); every non-template query is scoped by `ownerUserId`, hire/install/archive validate ownership, and cross-user access returns None/404 (covered by tests).

#### For code changes:
- [x] I have made a test plan
  - [x] `poetry run pytest backend/api/features/experts/experts_db_test.py -xvs` — 5 passed (idempotent hire, preload install, owner scoping, duplicate install, seed round-trip)
  - [x] `poetry run pytest backend/api/features/experts/routes_test.py -xvs` — 10 passed (all endpoints, 404 paths, snapshot)
  - [ ] Manual: run `poetry run python -m backend.api.features.experts.seed` against dev DB and hire via API (deferred to PR 2/3 integration)

### DIFF
--- a/autogpt_platform/backend/backend/api/features/experts/experts_db.py
+++ b/autogpt_platform/backend/backend/api/features/experts/experts_db.py
@@ -8,9 +8,8 @@ from backend.api.features.library import db as library_db
 
 logger = logging.getLogger(__name__)
 
-_WORKFLOW_INCLUDE = {
-    "Workflows": {"include": {"LibraryAgent": True, "StoreListingVersion": True}}
-}
+_WORKFLOW_ROW_INCLUDE = {"LibraryAgent": True, "StoreListingVersion": True}
+_WORKFLOW_INCLUDE = {"Workflows": {"include": _WORKFLOW_ROW_INCLUDE}}
 
 
 class ExpertTemplateNotFoundError(Exception):
@@ -94,7 +93,7 @@ async def hire_expert(user_id: str, template_id: str, name: str | None) -> HireR
         include=_WORKFLOW_INCLUDE,
     )
     if existing is not None:
-        return HireResult(expert=_to_model(existing), failed_preloads=[])
+        return await _existing_hire_result(existing)
 
     create_data: dict = {
         "ownerUserId": user_id,
@@ -117,34 +116,9 @@ async def hire_expert(user_id: str, template_id: str, name: str | None) -> HireR
         )
         if raced is None:
             raise
-        return HireResult(expert=_to_model(raced), failed_preloads=[])
+        return await _existing_hire_result(raced)
 
-    failed: list[str] = []
-    for preload in template.Workflows or []:
-        if preload.storeListingVersionId is None:
-            continue
-        try:
-            library_agent = await library_db.add_store_agent_to_library(
-                preload.storeListingVersionId, user_id
-            )
-            await prisma.models.ExpertWorkflow.prisma().create(
-                data={
-                    "expertId": expert.id,
-                    "storeListingVersionId": preload.storeListingVersionId,
-                    "libraryAgentId": library_agent.id,
-                }
-            )
-        except Exception:
-            # Honest partial hire: a failed preload must not sink the hire.
-            logger.exception(
-                f"Failed to install preload {preload.storeListingVersionId} "
-                f"on expert #{expert.id} for user #{user_id}"
-            )
-            failed.append(
-                preload.StoreListingVersion.name
-                if preload.StoreListingVersion
-                else preload.storeListingVersionId
-            )
+    failed = await _install_preloads(expert.id, user_id, template.Workflows or [])
 
     hydrated = await prisma.models.Expert.prisma().find_unique(
         where={"id": expert.id}, include=_WORKFLOW_INCLUDE
@@ -154,11 +128,71 @@ async def hire_expert(user_id: str, template_id: str, name: str | None) -> HireR
     return HireResult(expert=_to_model(hydrated), failed_preloads=failed)
 
 
+async def _existing_hire_result(row: prisma.models.Expert) -> HireResult:
+    """Idempotent-hire result for an already-existing hired copy.
+
+    Re-hiring an archived expert revives it — the unique
+    (ownerUserId, sourceTemplateId) constraint means a fresh row cannot be
+    created, and returning the archived row as-is would hand back a
+    "successful" hire that stays invisible to list_experts/get_expert.
+    """
+    if row.isArchived:
+        revived = await prisma.models.Expert.prisma().update(
+            where={"id": row.id},
+            data={"isArchived": False},
+            include=_WORKFLOW_INCLUDE,
+        )
+        if revived is not None:
+            row = revived
+    return HireResult(expert=_to_model(row), failed_preloads=[])
+
+
+async def _install_preloads(
+    expert_id: str, user_id: str, preloads: list[prisma.models.ExpertWorkflow]
+) -> list[str]:
+    """Install template preloads into the hiring user's library.
+
+    Honest partial hire: a failed preload is logged and reported, never
+    fatal to the hire itself.
+    """
+    failed: list[str] = []
+    for preload in preloads:
+        if preload.storeListingVersionId is None:
+            continue
+        try:
+            library_agent = await library_db.add_store_agent_to_library(
+                preload.storeListingVersionId, user_id
+            )
+            await prisma.models.ExpertWorkflow.prisma().create(
+                data={
+                    "expertId": expert_id,
+                    "storeListingVersionId": preload.storeListingVersionId,
+                    "libraryAgentId": library_agent.id,
+                }
+            )
+        except Exception:
+            logger.exception(
+                f"Failed to install preload {preload.storeListingVersionId} "
+                f"on expert #{expert_id} for user #{user_id}"
+            )
+            failed.append(
+                preload.StoreListingVersion.name
+                if preload.StoreListingVersion
+                else preload.storeListingVersionId
+            )
+    return failed
+
+
 async def install_workflow(
     user_id: str, expert_id: str, store_listing_version_id: str
 ) -> ExpertWorkflowRef:
     expert = await prisma.models.Expert.prisma().find_first(
-        where={"id": expert_id, "ownerUserId": user_id, "isTemplate": False}
+        where={
+            "id": expert_id,
+            "ownerUserId": user_id,
+            "isTemplate": False,
+            "isArchived": False,
+        }
     )
     if expert is None:
         raise ExpertNotFoundError(expert_id)
@@ -168,7 +202,7 @@ async def install_workflow(
             "expertId": expert_id,
             "storeListingVersionId": store_listing_version_id,
         },
-        include={"LibraryAgent": True, "StoreListingVersion": True},
+        include=_WORKFLOW_ROW_INCLUDE,
     )
     if existing is not None:
         return _to_workflow_ref(existing)
@@ -183,7 +217,7 @@ async def install_workflow(
                 "storeListingVersionId": store_listing_version_id,
                 "libraryAgentId": library_agent.id,
             },
-            include={"LibraryAgent": True, "StoreListingVersion": True},
+            include=_WORKFLOW_ROW_INCLUDE,
         )
     except prisma.errors.UniqueViolationError:
         # Lost a concurrent duplicate-install race; return the winner's row.
@@ -192,7 +226,7 @@ async def install_workflow(
                 "expertId": expert_id,
                 "storeListingVersionId": store_listing_version_id,
             },
-            include={"LibraryAgent": True, "StoreListingVersion": True},
+            include=_WORKFLOW_ROW_INCLUDE,
         )
         if raced is None:
             raise
--- a/autogpt_platform/backend/backend/api/features/experts/experts_db.py
+++ b/autogpt_platform/backend/backend/api/features/experts/experts_db.py
@@ -1,5 +1,6 @@
 import logging
 
+import prisma.errors
 import prisma.models
 
 from backend.api.features.experts.models import Expert, ExpertWorkflowRef, HireResult
@@ -106,7 +107,17 @@ async def hire_expert(user_id: str, template_id: str, name: str | None) -> HireR
     }
     if template.toolProfile is not None:
         create_data["toolProfile"] = template.toolProfile
-    expert = await prisma.models.Expert.prisma().create(data=create_data)
+    try:
+        expert = await prisma.models.Expert.prisma().create(data=create_data)
+    except prisma.errors.UniqueViolationError:
+        # Lost a concurrent hire race; the winner's row satisfies idempotency.
+        raced = await prisma.models.Expert.prisma().find_first(
+            where={"ownerUserId": user_id, "sourceTemplateId": template_id},
+            include=_WORKFLOW_INCLUDE,
+        )
+        if raced is None:
+            raise
+        return HireResult(expert=_to_model(raced), failed_preloads=[])
 
     failed: list[str] = []
     for preload in template.Workflows or []:
@@ -165,14 +176,27 @@ async def install_workflow(
     library_agent = await library_db.add_store_agent_to_library(
         store_listing_version_id, user_id
     )
-    row = await prisma.models.ExpertWorkflow.prisma().create(
-        data={
-            "expertId": expert_id,
-            "storeListingVersionId": store_listing_version_id,
-            "libraryAgentId": library_agent.id,
-        },
-        include={"LibraryAgent": True, "StoreListingVersion": True},
-    )
+    try:
+        row = await prisma.models.ExpertWorkflow.prisma().create(
+            data={
+                "expertId": expert_id,
+                "storeListingVersionId": store_listing_version_id,
+                "libraryAgentId": library_agent.id,
+            },
+            include={"LibraryAgent": True, "StoreListingVersion": True},
+        )
+    except prisma.errors.UniqueViolationError:
+        # Lost a concurrent duplicate-install race; return the winner's row.
+        raced = await prisma.models.ExpertWorkflow.prisma().find_first(
+            where={
+                "expertId": expert_id,
+                "storeListingVersionId": store_listing_version_id,
+            },
+            include={"LibraryAgent": True, "StoreListingVersion": True},
+        )
+        if raced is None:
+            raise
+        return _to_workflow_ref(raced)
     return _to_workflow_ref(row)
 
 
--- a/autogpt_platform/backend/backend/api/features/experts/experts_db.py
+++ b/autogpt_platform/backend/backend/api/features/experts/experts_db.py
@@ -1,0 +1,185 @@
+import logging
+
+import prisma.models
+
+from backend.api.features.experts.models import Expert, ExpertWorkflowRef, HireResult
+from backend.api.features.library import db as library_db
+
+logger = logging.getLogger(__name__)
+
+_WORKFLOW_INCLUDE = {
+    "Workflows": {"include": {"LibraryAgent": True, "StoreListingVersion": True}}
+}
+
+
+class ExpertTemplateNotFoundError(Exception):
+    def __init__(self, template_id: str):
+        super().__init__(f"Expert template {template_id} not found")
+        self.template_id = template_id
+
+
+class ExpertNotFoundError(Exception):
+    def __init__(self, expert_id: str):
+        super().__init__(f"Expert {expert_id} not found")
+        self.expert_id = expert_id
+
+
+def _to_workflow_ref(row: prisma.models.ExpertWorkflow) -> ExpertWorkflowRef:
+    listing = row.StoreListingVersion
+    library_agent = row.LibraryAgent
+    return ExpertWorkflowRef(
+        id=row.id,
+        store_listing_version_id=row.storeListingVersionId,
+        library_agent_id=row.libraryAgentId,
+        graph_id=library_agent.agentGraphId if library_agent else None,
+        name=listing.name if listing else None,
+        description=listing.description if listing else None,
+    )
+
+
+def _to_model(row: prisma.models.Expert) -> Expert:
+    return Expert(
+        id=row.id,
+        name=row.name,
+        avatar_url=row.avatarUrl,
+        role=row.role,
+        tagline=row.tagline,
+        identity=row.identity,
+        is_template=row.isTemplate,
+        source_template_id=row.sourceTemplateId,
+        is_archived=row.isArchived,
+        workflows=[_to_workflow_ref(w) for w in row.Workflows or []],
+    )
+
+
+async def list_templates() -> list[Expert]:
+    rows = await prisma.models.Expert.prisma().find_many(
+        where={"isTemplate": True, "isArchived": False},
+        include=_WORKFLOW_INCLUDE,
+    )
+    return [_to_model(row) for row in rows]
+
+
+async def list_experts(user_id: str) -> list[Expert]:
+    rows = await prisma.models.Expert.prisma().find_many(
+        where={
+            "ownerUserId": user_id,
+            "isTemplate": False,
+            "isArchived": False,
+        },
+        include=_WORKFLOW_INCLUDE,
+    )
+    return [_to_model(row) for row in rows]
+
+
+async def get_expert(user_id: str, expert_id: str) -> Expert | None:
+    row = await prisma.models.Expert.prisma().find_first(
+        where={"id": expert_id, "ownerUserId": user_id, "isTemplate": False},
+        include=_WORKFLOW_INCLUDE,
+    )
+    return _to_model(row) if row is not None else None
+
+
+async def hire_expert(user_id: str, template_id: str, name: str | None) -> HireResult:
+    template = await prisma.models.Expert.prisma().find_first(
+        where={"id": template_id, "isTemplate": True, "isArchived": False},
+        include=_WORKFLOW_INCLUDE,
+    )
+    if template is None:
+        raise ExpertTemplateNotFoundError(template_id)
+
+    existing = await prisma.models.Expert.prisma().find_first(
+        where={"ownerUserId": user_id, "sourceTemplateId": template_id},
+        include=_WORKFLOW_INCLUDE,
+    )
+    if existing is not None:
+        return HireResult(expert=_to_model(existing), failed_preloads=[])
+
+    create_data: dict = {
+        "ownerUserId": user_id,
+        "name": name or template.name,
+        "avatarUrl": template.avatarUrl,
+        "role": template.role,
+        "tagline": template.tagline,
+        "identity": template.identity,
+        "sourceTemplateId": template.id,
+    }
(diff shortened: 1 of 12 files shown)


autogpt_platform/backend/backend/api/features/experts/experts_db.py
```diff
@@ -0,0 +1,185 @@
...
+    template = await prisma.models.Expert.prisma().find_first(
+        where={"id": template_id, "isTemplate": True, "isArchived": False},
+        include=_WORKFLOW_INCLUDE,
+    )
+    if template is None:
+        raise ExpertTemplateNotFoundError(template_id)
+
+    existing = await prisma.models.Expert.prisma().find_first(
+        where={"ownerUserId": user_id, "sourceTemplateId": template_id},
+        include=_WORKFLOW_INCLUDE,
```
0ubbe, line 93: 🤖 🟠 **Should Fix**: TOCTOU race in the hire idempotency check. Two concurrent hires of the same template by the same user can both pass this `find_first` gate; the second `create` then hits the `Expert_ownerUserId_sourceTemplateId_key` unique constraint and surfaces as a 500 instead of the documented idempotent `HireResult`. Wrap the `create` in a `try/except prisma.errors.UniqueViolationError` that re-fetches and returns the existing row (the constraint already guarantees correctness — this is just about not turning the race into a 500).

autogpt_platform/backend/backend/api/features/experts/experts_db.py
```diff
@@ -0,0 +1,185 @@
...
+        where={"id": expert_id, "ownerUserId": user_id, "isTemplate": False}
+    )
+    if expert is None:
+        raise ExpertNotFoundError(expert_id)
+
+    existing = await prisma.models.ExpertWorkflow.prisma().find_first(
+        where={
+            "expertId": expert_id,
+            "storeListingVersionId": store_listing_version_id,
+        },
```
0ubbe, line 204: 🤖 🟠 **Should Fix**: Same check-then-act race as in `hire_expert`: concurrent duplicate installs of the same listing on one expert both pass the `find_first`, and the loser of the race gets a 500 from the `ExpertWorkflow_expertId_storeListingVersionId_key` unique violation instead of the documented duplicate-safe existing row. Catch `prisma.errors.UniqueViolationError` on the `create` and return the re-fetched existing row.

autogpt_platform/backend/backend/api/features/experts/routes.py
```diff
@@ -0,0 +1,82 @@
...
+router = APIRouter(
+    prefix="/experts",
+    tags=["experts", "private"],
+    dependencies=[Security(autogpt_auth_lib.requires_user)],
+)
+
+
+class HireRequest(BaseModel):
+    template_id: str
+    name: str | None = None
```
0ubbe: 🤖 🟡 **Nice to Have**: `name` is stored verbatim as the hired expert's display name with no length bound — a caller can persist an arbitrarily large string. Suggest `name: str | None = Field(default=None, max_length=100)` (or similar) to validate at the boundary.

autogpt_platform/backend/backend/api/features/experts/experts_db.py
```diff
@@ -0,0 +1,185 @@
...
+    }
+    if template.toolProfile is not None:
+        create_data["toolProfile"] = template.toolProfile
+    expert = await prisma.models.Expert.prisma().create(data=create_data)
+
+    failed: list[str] = []
+    for preload in template.Workflows or []:
+        if preload.storeListingVersionId is None:
+            continue
+        try:
```
0ubbe, line 162: 🤖 🟡 **Nice to Have**: The PR description advertises "partial hire is honest", but no test exercises the failure path — a preload whose `add_store_agent_to_library` raises, asserting the hire still succeeds and `failed_preloads` carries the listing name. A db-level test with a mocked failing install would lock in this contract.

autogpt_platform/backend/backend/api/features/experts/experts_db.py
0ubbe: 🤖 Fixed in [2195c3fc1](https://github.com/Significant-Gravitas/AutoGPT/commit/2195c3fc10bb7c2cac4e64737a7a15f2d92009d3): the create is now wrapped in `try/except prisma.errors.UniqueViolationError` — on a lost race the winner's row is re-fetched and returned as the idempotent HireResult.

autogpt_platform/backend/backend/api/features/experts/experts_db.py
0ubbe: 🤖 Fixed in [2195c3fc1](https://github.com/Significant-Gravitas/AutoGPT/commit/2195c3fc10bb7c2cac4e64737a7a15f2d92009d3): same pattern — `UniqueViolationError` on the create now re-fetches and returns the existing ExpertWorkflow row instead of surfacing a 500.

autogpt_platform/backend/backend/api/features/experts/routes.py
0ubbe: 🤖 Fixed in [2195c3fc1](https://github.com/Significant-Gravitas/AutoGPT/commit/2195c3fc10bb7c2cac4e64737a7a15f2d92009d3): `name` is now `Field(default=None, max_length=100)`.

autogpt_platform/backend/backend/api/features/experts/experts_db.py
0ubbe: 🤖 Fixed in [2195c3fc1](https://github.com/Significant-Gravitas/AutoGPT/commit/2195c3fc10bb7c2cac4e64737a7a15f2d92009d3): added `test_hire_reports_failed_preload_without_sinking_hire` — mocks `add_store_agent_to_library` to raise and asserts the hire succeeds with the preload reported in `failed_preloads`.

autogpt_platform/backend/backend/api/features/experts/experts_db_test.py
```diff
@@ -0,0 +1,190 @@
...
+            "role": f"{name}'s role",
+            "identity": f"You are {name}, an expert.",
+            "isTemplate": True,
+        }
+    )
+    for slv_id in preload_listings:
+        await prisma.models.ExpertWorkflow.prisma().create(
+            data={"expertId": template.id, "storeListingVersionId": slv_id}
+        )
+    return template
```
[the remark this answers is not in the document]
0ubbe: 🤖 Fixed in [3f5b92ee6](https://github.com/Significant-Gravitas/AutoGPT/commit/3f5b92ee615539c793a70b04a57e05a01c3483b4): `_seed_template` now stores a uuid-suffixed name, so ad-hoc test templates can never collide with (or be adopted by) seed.ROSTER's real roster names.

autogpt_platform/backend/backend/api/features/experts/experts_db.py
```diff
@@ -0,0 +1,185 @@
...
+    )
+    if template is None:
+        raise ExpertTemplateNotFoundError(template_id)
+
+    existing = await prisma.models.Expert.prisma().find_first(
+        where={"ownerUserId": user_id, "sourceTemplateId": template_id},
+        include=_WORKFLOW_INCLUDE,
+    )
+    if existing is not None:
+        return HireResult(expert=_to_model(existing), failed_preloads=[])
```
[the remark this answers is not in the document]
0ubbe: 🤖 Fixed in [3f5b92ee6](https://github.com/Significant-Gravitas/AutoGPT/commit/3f5b92ee615539c793a70b04a57e05a01c3483b4), with one deliberate difference from the proposed diff: filtering the lookup on `isArchived: False` alone would send re-hires down the create path straight into the (ownerUserId, sourceTemplateId) unique violation — a 500, not a fresh hire. Instead, re-hiring an archived expert now revives it (`isArchived=False`) and returns it, which also survives the concurrent-race path. Covered by `test_rehire_after_archive_revives_expert`.

autogpt_platform/backend/backend/api/features/experts/experts_db.py
```diff
@@ -0,0 +1,185 @@
...
+
+
+async def install_workflow(
+    user_id: str, expert_id: str, store_listing_version_id: str
+) -> ExpertWorkflowRef:
+    expert = await prisma.models.Expert.prisma().find_first(
+        where={"id": expert_id, "ownerUserId": user_id, "isTemplate": False}
+    )
+    if expert is None:
+        raise ExpertNotFoundError(expert_id)
```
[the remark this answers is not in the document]
0ubbe: 🤖 Fixed in [3f5b92ee6](https://github.com/Significant-Gravitas/AutoGPT/commit/3f5b92ee615539c793a70b04a57e05a01c3483b4): the ownership lookup now includes `isArchived: False`; installing onto an archived expert raises ExpertNotFoundError (covered by `test_install_workflow_on_archived_expert_raises`).

autogpt_platform/backend/migrations/20260727070306_add_expert_entity/migration.sql
```diff
@@ -0,0 +1,80 @@
...
+CREATE INDEX "ExpertWorkflow_expertId_idx" ON "ExpertWorkflow"("expertId");
+
+-- CreateIndex
+CREATE UNIQUE INDEX "ExpertWorkflow_expertId_storeListingVersionId_key" ON "ExpertWorkflow"("expertId", "storeListingVersionId");
+
+-- CreateIndex
+CREATE INDEX "ChatSession_expertId_idx" ON "ChatSession"("expertId");
+
+-- AddForeignKey
+ALTER TABLE "ChatSession" ADD CONSTRAINT "ChatSession_expertId_fkey" FOREIGN KEY ("expertId") REFERENCES "Expert"("id") ON DELETE SET NULL ON UPDATE CASCADE;
```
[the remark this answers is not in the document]
0ubbe: 🤖 Fixed in [3f5b92ee6](https://github.com/Significant-Gravitas/AutoGPT/commit/3f5b92ee615539c793a70b04a57e05a01c3483b4): Prisma's migrate deploy wraps each migration in one transaction, so `CREATE INDEX CONCURRENTLY` can't live in the file — following the repo's documented pattern (20260516120000_add_creator_search_trgm_indexes), the ChatSession index is now `IF NOT EXISTS` with the out-of-band CONCURRENTLY escape hatch documented, and the FK is added `NOT VALID` with a follow-up migration running `VALIDATE CONSTRAINT` in its own transaction (SHARE UPDATE EXCLUSIVE only; all pre-existing rows have NULL expertId).

autogpt_platform/frontend/src/app/api/openapi.json
```diff
@@ -5350,6 +5350,203 @@
...
+            "content": {
+              "application/json": {
+                "schema": { "$ref": "#/components/schemas/HTTPValidationError" }
+              }
+            }
+          }
+        },
+        "security": [{ "HTTPBearerJWT": [] }]
+      }
+    },
```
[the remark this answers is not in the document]
0ubbe: 🤖 Fixed in [3f5b92ee6](https://github.com/Significant-Gravitas/AutoGPT/commit/3f5b92ee615539c793a70b04a57e05a01c3483b4): all four routes that raise 404 now declare `responses={404: ...}` and the exported OpenAPI schema documents the branch.

Swiftyos (approved):

autogpt_platform/backend/backend/api/features/experts/experts_db.py
```diff
@@ -0,0 +1,243 @@
...
+    create_data: dict = {
+        "ownerUserId": user_id,
+        "name": name or template.name,
+        "avatarUrl": template.avatarUrl,
+        "role": template.role,
+        "tagline": template.tagline,
+        "identity": template.identity,
+        "sourceTemplateId": template.id,
+    }
+    if template.toolProfile is not None:
```
kcze, line 107: 🤖 🟠 **Should Fix**: `template.toolProfile` is written back as a raw Python object. Prisma deserializes `Json` columns into plain dicts on read (`fields.Json` docstring: `User(json_obj='{"foo": null}') -> User(json_obj={'foo': None})`), and the `serialize_json` hook is registered *only* for `fields.Json` — a bare dict gets rendered as a nested Data node instead of a JSON string. The moment any roster template carries a non-null `toolProfile` (the column's whole purpose), `POST /api/experts` fails with a Prisma query error. Wrap it: `SafeJson(template.toolProfile)`, matching the 100+ other Json writes in the repo. Typing `create_data` as `prisma.types.ExpertCreateInput` instead of bare `dict` would have caught this statically. _(flagged by: Claude)_

autogpt_platform/backend/backend/api/features/experts/experts_db.py
```diff
@@ -0,0 +1,243 @@
...
+        }
+    )
+    if expert is None:
+        raise ExpertNotFoundError(expert_id)
+
+    existing = await prisma.models.ExpertWorkflow.prisma().find_first(
+        where={
+            "expertId": expert_id,
+            "storeListingVersionId": store_listing_version_id,
+        },
```
kcze, line 204: 🤖 🟠 **Should Fix**: the duplicate-install early return never checks that the linked `LibraryAgent` is still live, so re-installing can never repair a broken workflow. Deleting the underlying graph calls `delete_library_agent_by_graph_id` (a hard `delete_many`), and the `ON DELETE SET NULL` FK nulls `ExpertWorkflow.libraryAgentId` while the row itself survives and still matches the `(expertId, storeListingVersionId)` unique key. Re-installing the same listing then returns 200 with `library_agent_id: null` forever — the expert can never run that workflow again and there is no API to fix it. `add_store_agent_to_library` already restores soft-deleted/archived agents on its unique-violation path; calling it and updating the row would recover both cases. _(flagged by: Claude x2 + Codex — all three reviewers independently)_

autogpt_platform/backend/backend/api/features/experts/experts_db.py
```diff
@@ -0,0 +1,243 @@
...
+    """
+    if row.isArchived:
+        revived = await prisma.models.Expert.prisma().update(
+            where={"id": row.id},
+            data={"isArchived": False},
+            include=_WORKFLOW_INCLUDE,
+        )
+        if revived is not None:
+            row = revived
+    return HireResult(expert=_to_model(row), failed_preloads=[])
```
kcze, line 147: 🤖 🟠 **Should Fix**: `_existing_hire_result` hardcodes `failed_preloads=[]` and installs nothing, so the `failed_preloads` contract has no recovery path. If the first hire of Maria loses all 3 preloads to a transient marketplace/graph failure, the user retries and gets `200 {failed_preloads: []}` with `workflows: []` — the expert stays permanently workflow-less unless they manually re-install each listing. The same return also serves the concurrent-hire race: the winner commits its `Expert` row *before* installing preloads, so a racing request reaches this line and reports a clean hire against a half-built expert. Reconciling the template's preloads here would fix both. _(flagged by: Claude + Codex)_

autogpt_platform/backend/migrations/20260727070306_add_expert_entity/migration.sql
```diff
@@ -0,0 +1,92 @@
...
+CREATE UNIQUE INDEX "ExpertWorkflow_expertId_storeListingVersionId_key" ON "ExpertWorkflow"("expertId", "storeListingVersionId");
+
+-- CreateIndex
+-- ChatSession is an existing, write-heavy table. Prisma's `migrate deploy`
+-- wraps each migration in a single transaction, and Postgres rejects
+-- `CREATE INDEX CONCURRENTLY` inside a transaction block; the plain form
+-- briefly holds a ShareLock during the scan. For deployments that can't
+-- tolerate that, run the CONCURRENTLY equivalent out-of-band before this
+-- migration ships and Postgres will skip the recreate via IF NOT EXISTS
+-- (same pattern as 20260516120000_add_creator_search_trgm_indexes).
```
kcze, line 65: 🤖 🟠 **Should Fix**: this out-of-band `CREATE INDEX CONCURRENTLY` escape hatch cannot actually be followed. Unlike the cited trgm precedent (whose columns already existed), `ChatSession.expertId` is created by line 2 of *this same* migration — an operator cannot index a column that does not exist yet. If they work around it by adding the column manually first, line 2's `ALTER TABLE "ChatSession" ADD COLUMN "expertId" TEXT;` has no `IF NOT EXISTS`, aborts with "column already exists", and Prisma marks the migration failed, blocking every subsequent deploy until someone resolves it by hand. Either split the ADD COLUMN into its own earlier migration or drop the comment. _(flagged by: Claude)_

autogpt_platform/backend/backend/api/features/experts/seed.py
```diff
@@ -0,0 +1,158 @@
...
+            "automated-support-ai",
+            "personalized-morning-coffee-newsletter",
+        ],
+    },
+]
+
+
+async def _resolve_active_version_id(slug: str) -> str | None:
+    listing = await prisma.models.StoreListing.prisma().find_first(
+        where={"slug": slug, "isDeleted": False}
```
kcze, line 85: 🤖 🟠 **Should Fix**: `slug` is only unique *per owner* (`StoreListing @@unique([owningUserId, slug])`), and this `find_first` has no ordering, no owner filter and no approval filter. Any user who publishes a listing with slug `linkedin-post-generator` can have their agent nondeterministically bound into the Maria roster template — and it is then auto-installed into the library of every user who hires Maria. Resolve by owning-user + slug, or pin stable listing IDs. _(flagged by: Claude + Codex)_

autogpt_platform/backend/backend/api/features/experts/seed.py
```diff
@@ -0,0 +1,158 @@
...
+        version_id = await _resolve_active_version_id(slug)
+        if version_id is None:
+            logger.warning(
+                f"Store listing slug '{slug}' not found; "
+                f"skipping preload for expert '{entry['name']}'"
+            )
+            continue
+        if version_id in existing_version_ids:
+            continue
+        await prisma.models.ExpertWorkflow.prisma().create(
```
kcze, line 131: 🤖 🟠 **Should Fix**: `_sync_preloads` only ever appends and dedupes on `storeListingVersionId`, so when a preloaded marketplace agent publishes a new version the next seed run adds a second `ExpertWorkflow` while the stale version row survives. The template then lists the same agent twice and every subsequent hire installs *both* versions into the hiring user's library — compounding with each version bump. Removing rows whose version is no longer the active one would keep the roster in sync. (`test_seed_roster_round_trip` can't catch this — no slugs resolve in the test DB.) _(flagged by: Claude + Codex)_

autogpt_platform/backend/backend/api/features/experts/experts_db.py
```diff
@@ -0,0 +1,243 @@
...
+            "isArchived": False,
+        },
+        include=_WORKFLOW_INCLUDE,
+    )
+    return [_to_model(row) for row in rows]
+
+
+async def get_expert(user_id: str, expert_id: str) -> Expert | None:
+    row = await prisma.models.Expert.prisma().find_first(
+        where={"id": expert_id, "ownerUserId": user_id, "isTemplate": False},
```
kcze, line 77: 🤖 🟡 **Nice to Have**: `get_expert` omits the `isArchived: False` filter that both `list_experts` and `install_workflow` apply. `DELETE /experts/{id}` → 204, then `GET /experts/{id}` → 200 with the full body, so a client sitting on the detail page keeps rendering a deleted expert whose install action then 404s. _(flagged by: Claude + Codex)_

autogpt_platform/backend/backend/api/features/experts/experts_db.py
```diff
@@ -0,0 +1,243 @@
...
+        is_template=row.isTemplate,
+        source_template_id=row.sourceTemplateId,
+        is_archived=row.isArchived,
+        workflows=[_to_workflow_ref(w) for w in row.Workflows or []],
+    )
+
+
+async def list_templates() -> list[Expert]:
+    rows = await prisma.models.Expert.prisma().find_many(
+        where={"isTemplate": True, "isArchived": False},
```
kcze, line 57: 🤖 🟡 **Nice to Have**: `list_templates` filters on `isTemplate`/`isArchived` but not `ownerUserId IS NULL`, so `GET /experts/templates` — which takes no user param — is only accidentally safe. No IDOR exists today (nothing sets `isTemplate` on an owned row), but any future admin or seed path that does would publish that user's expert `name` and `identity` (a full system prompt) to every authenticated caller. Adding `"ownerUserId": None` is free and stays within the existing index. _(flagged by: Claude)_

autogpt_platform/backend/backend/api/features/experts/routes.py
```diff
@@ -0,0 +1,99 @@
...
+    operation_id="hire_expert",
+    responses={404: {"description": "Expert template not found"}},
+)
+async def hire_expert(
+    request: HireRequest,
+    user_id: str = Security(autogpt_auth_lib.get_user_id),
+) -> HireResult:
+    try:
+        return await experts_db.hire_expert(user_id, request.template_id, request.name)
+    except experts_db.ExpertTemplateNotFoundError as e:
```
kcze, line 41: 🤖 🟡 **Nice to Have**: this only catches `ExpertTemplateNotFoundError`, but `experts_db.hire_expert` also raises `ExpertNotFoundError` when the re-read of the just-created row returns `None` (experts_db.py:127). That escapes to the catch-all handler, turning a concurrent row deletion into an opaque 500 after the expert and its library installs were already created. _(flagged by: Claude)_

autogpt_platform/backend/backend/api/features/experts/experts_db_test.py
```diff
@@ -0,0 +1,242 @@
...
+
+
+@pytest.mark.asyncio(loop_scope="session")
+async def test_get_expert_scopes_by_owner(
+    server: SpinTestServer, test_user, other_user
+):
+    template = await _seed_template(name="Maria", preload_listings=[])
+    hired = await experts_db.hire_expert(test_user.id, template.id, None)
+    assert await experts_db.get_expert(other_user.id, hired.expert.id) is None
+    assert await experts_db.get_expert(test_user.id, hired.expert.id) is not None
```
kcze, line 204: 🤖 🟡 **Nice to Have**: cross-owner coverage exists only for `get_expert`. There is no DB-level test that `archive_expert` or `install_workflow` reject another user's expert ID, and `routes_test.py` mocks `experts_db` wholesale so it can't catch it either. A future refactor that drops `ownerUserId` from the `update_many`/`find_first` where-clause would be a live IDOR with the suite still green. _(flagged by: Claude)_

autogpt_platform/backend/migrations/20260727070306_add_expert_entity/migration.sql
```diff
@@ -0,0 +1,92 @@
...
+ALTER TABLE "Expert" ADD CONSTRAINT "Expert_teamId_fkey" FOREIGN KEY ("teamId") REFERENCES "Team"("id") ON DELETE SET NULL ON UPDATE CASCADE;
+
+-- AddForeignKey
+ALTER TABLE "ExpertWorkflow" ADD CONSTRAINT "ExpertWorkflow_expertId_fkey" FOREIGN KEY ("expertId") REFERENCES "Expert"("id") ON DELETE CASCADE ON UPDATE CASCADE;
+
+-- AddForeignKey
+ALTER TABLE "ExpertWorkflow" ADD CONSTRAINT "ExpertWorkflow_storeListingVersionId_fkey" FOREIGN KEY ("storeListingVersionId") REFERENCES "StoreListingVersion"("id") ON DELETE SET NULL ON UPDATE CASCADE;
+
+-- AddForeignKey
+ALTER TABLE "ExpertWorkflow" ADD CONSTRAINT "ExpertWorkflow_libraryAgentId_fkey" FOREIGN KEY ("libraryAgentId") REFERENCES "LibraryAgent"("id") ON DELETE SET NULL ON UPDATE CASCADE;
```
kcze, line 92: 🤖 🟡 **Nice to Have**: no index backs `ExpertWorkflow("libraryAgentId")` — the composite unique leads with `expertId`, so Postgres can't use it here. Every hard `LibraryAgent` delete (triggered by ordinary graph deletion) sequentially scans the whole `ExpertWorkflow` table to apply `ON DELETE SET NULL`. The same applies to `storeListingVersionId` on line 89. _(flagged by: Claude)_

PR #13688, opened 2026-07-27
fix(blocks): use ensure_ascii=True for AGPT_VARIABLES to prevent UnicodeEncodeError with emoji

### LINKED ISSUE #13551: ExecuteCodeBlock: UnicodeEncodeError when data contains emoji (surrogate pairs in AGPT_VARIABLES env var)

Iliamgeladze1998:
## Summary

Fixes #13551 — `ExecuteCodeBlock` crashes with `UnicodeEncodeError` when upstream block data contains emoji (surrogate pairs).

## Problem

When an upstream block (e.g. `NotionReadDatabaseBlock`) returns data containing emoji like `"Holidays 🎅🏻"`, the variable injection mechanism serializes the data into the `AGPT_VARIABLES` environment variable using `json.dumps(variables)`. On Python 3.13, `os.environ.__setitem__` rejects UTF-16 surrogate pairs:

```
```

This crashes `ExecuteCodeBlock` before user code even runs.

## Fix

Use `json.dumps(variables, ensure_ascii=True)` instead of `json.dumps(variables)`. This produces pure ASCII output (e.g. `\ud83c\udf85\ud83d\udc71\ud83c\udffd`) that `os.environ` can always handle. The sandbox deserialization (`json.loads`) correctly reconstructs the original Unicode characters from the ASCII escapes.

### Changes

- `build_variable_injection()`: `json.dumps(variables)` → `json.dumps(variables, ensure_ascii=True)`
- `_is_json_serializable()`: `json.dumps(value)` → `json.dumps(value, ensure_ascii=True)` (for consistency)

## Testing

- With `ensure_ascii=True`, `json.dumps({"name": "Holidays 🎅🏻"})` produces `"{\"name\": \"Holidays \ud83c\udf85\ud83d\udc71\ud83c\udffd\"}"` — pure ASCII, no surrogates.
- `json.loads()` correctly reconstructs the original emoji.
- No behavior change for non-emoji data.

Fixes #13551


Pwuts:
Duplicate of #13576, and not a fix

PR #13689, opened 2026-07-27
feat(frontend): experts marketplace section, team page, per-expert threads

0ubbe:
### Why / What / How

**Why:** Third and final slice of the "Hire Experts" MVP (stacked on #13687, which stacks on #13685). The backend lets users hire Experts and chat with them; this PR ships the flag-gated surfaces that make the loop usable end-to-end: hire from the marketplace, see your Team, chat in per-expert CoPilot threads, and install Workflows on an expert.

**What:**
- Marketplace: "Meet the Experts" section (curated roster from `GET /api/experts/templates`) with a profile sheet and Hire flow; hired templates show a Hired state.
- New `/team` page: Autopilot card first (the generalist — "runs the shop"), then hired expert cards with installed workflow names/counts and Chat / Install workflow actions.
- CoPilot: per-expert threads — `/copilot?expertId=<id>` opens the expert's latest thread or creates one with `expert_id`; the sidebar groups threads under expert headers (Autopilot is the default group); thread header and assistant messages show the expert's name/avatar.
- "Install on Expert…" action on marketplace agent detail + "Install workflow" on Team cards, both via one `InstallWorkflowPicker` component (two modes: pick-expert / pick-workflow) calling `POST /api/experts/{id}/workflows`.

**How:**
- Everything sits behind `Flag.HIRE_EXPERTS` (`hire-experts` LaunchDarkly flag; env override `NEXT_PUBLIC_FORCE_FLAG_HIRE_EXPERTS`). Flag off → sections hidden, `/team` 404s, copilot behavior byte-for-byte as today.
- Data fetching uses Orval-generated React Query hooks (`useListExpertTemplates`, `useHireExpert`, `useListExperts`, `useInstallExpertWorkflow`, …) regenerated from the stacked backend's OpenAPI spec; hire/install mutations invalidate the `["/api/experts"]` query key so all surfaces refresh.
- Copilot diffs are surgical: sidebar grouping is a pure partition of the existing session list by `expert_id` (renders only when the flag is on and a grouped session exists — flat list untouched otherwise); identity comes from one shared `useExpertMap` hook; stream transport, chat registry/stream store, tool cards, and reconnect logic are untouched.
- Failure behavior: templates query error → section renders nothing (marketplace never breaks); missing/archived expert → default header; duplicate install → "Already installed on {name}" toast without a POST (client pre-check; the API is also duplicate-safe).

### Changes 🏗️

- `src/services/feature-flags/use-get-flag.ts` — `Flag.HIRE_EXPERTS` + `src/app/api/openapi.json` regenerated (experts tag, `expert_id` on chat sessions). Generated hooks are gitignored per repo convention; run `pnpm generate:api` after checkout.
- `marketplace/components/ExpertsSection/` (+ `ExpertProfileSheet`) and `MainMarketplacePage` mount; `InstallOnExpertButton` in `AgentInfo`
- `team/` route: `page.tsx`, `useTeamPage.ts`, `AutopilotCard`, `ExpertTeamCard`, `EmptyTeamState`
- `copilot/`: `useChatSession` (expert_id on create + latest-thread adoption), `useSessionList` grouping, `ChatSidebar` grouped rendering, `ChatMessagesContainer` thread header + assistant identity, `useExpertMap`
- `components/molecules/InstallWorkflowPicker/` wired into marketplace detail and Team page
- **Backend slice** (supports the surfaces above): `Expert.bio` + `Expert.skills` columns (migrations `20260728100000_add_expert_bio`, `20260728200000_add_expert_skills` — additive only) exposed on the experts API; roster seed gains avatars (`/experts/*.svg`), bios, and skills; expert lookups routed through `DatabaseManager` for the Prisma-less copilot executor; expert identity moved from the first-message context block to a per-session system-prompt suffix (`build_expert_identity_suffix`) so it outranks the base persona — the cacheable prompt prefix stays byte-identical
- Tests: page-level Vitest + RTL + MSW integration tests for all four surfaces; global `getExpertsMock()` added to `mock-handlers.ts`

Note: the `hire-experts` flag must be created in LaunchDarkly before enabling in any environment. Screenshots deferred — will attach from a running stack before undrafting.

Stacked on #13687 (base retargets to `dev` once parents merge).

#### For code changes:
- [x] I have made a test plan
  - [x] `marketplace/__tests__/experts-section.test.tsx` — 2 passed (render + hire flow; hired state)
  - [x] `marketplace/__tests__/install-on-expert.test.tsx` — 3 passed (install, duplicate, hidden without experts)
  - [x] `team/__tests__/main.test.tsx` — 6 passed (Autopilot first, expert cards, Chat links, empty state, flag-off 404)
  - [x] `copilot/__tests__/expert-threads.test.tsx` — 7 passed (expert_id on create, grouping, thread header, plain sessions unchanged)
  - [x] Regression: copilot sidebar/messages/page suites + marketplace main suites re-run green
  - [ ] Manual E2E on a running stack with seeded roster (before undrafting)
  - [ ] Screenshots of the three surfaces (before undrafting)


0ubbe:
!deploy

Pwuts:

- **Deployed:** `fe3328e27293531485f5d82a39638650f686fc2d` at 2026-07-28 14:12 UTC
- **Database:** isolated Supabase branch `pr-13689` (state persists across redeploys unless migration drift forces a reset)

autogpt_platform/frontend/src/app/(platform)/copilot/useChatSession.ts
```diff
@@ -68,6 +73,34 @@ export function useChatSession({ dryRun = false }: UseChatSessionOptions = {}) {
...
+      },
+    },
+  );
+
+  // Deep link /copilot?expertId=<id>: adopt the expert's latest thread once
+  // per expert. The latch lets "New Chat" stay on a fresh expert session
+  // instead of bouncing back to the latest thread. Only the mount-time
+  // expertId adopts — a recipient picked in the UI after mount must keep
+  // the fresh new-task state, not jump to the expert's old thread.
+  const mountExpertIdRef = useRef(expertId);
```
0ubbe: 🤖 🔴 **Blocker**: "New Chat" with an expert selected bounces straight back to the expert's latest thread. The `mountExpertIdRef`/`redirectedExpertRef` latch cannot survive: `CopilotPage.tsx:145` mounts the host with `key={\`chat-host-${sessionId ?? "new"}\`}`, so New Chat (sets sessionId to null, leaves `expertId` in the URL) remounts the subtree and re-initialises both refs; the `expert_id`-filtered list query re-enables with cached data and the adoption effect immediately calls `setSessionId(latest.id)` again. A user on `/copilot?expertId=X` can never start a second thread with that expert. The latch must live outside the remounting subtree (store or URL), or `handleNewChat` must clear `expertId`.

autogpt_platform/frontend/src/app/(platform)/copilot/useChatSession.ts
```diff
@@ -68,6 +73,34 @@ export function useChatSession({ dryRun = false }: UseChatSessionOptions = {}) {
...
+  const redirectedExpertRef = useRef<string | null>(null);
+  useEffect(() => {
```
0ubbe, line 110: 🤖 🟠 **Should Fix**: Late thread adoption can misroute the user's first message. The composer renders while the adoption query is in flight; `onSend` stashes the message via `setPendingFirstSend` and awaits `createSession()`, and the pending send flushes on *any* sessionId change. If the list query resolves before the POST returns, adoption points sessionId at the pre-existing thread — the first message lands there, then `onSuccess` navigates to the new empty session and the message vanishes from view. Gate the composer (or skip adoption) while `latestExpertSessionQuery.isLoading`.

autogpt_platform/frontend/src/app/(platform)/copilot/components/ChatSidebar/ChatSidebar.tsx
```diff
@@ -241,6 +253,197 @@ export function ChatSidebar() {
     }
   }
 
+  const sessionGroups = isExpertsEnabled
```
0ubbe: 🤖 🟠 **Should Fix**: Expert grouping breaks pinned ordering. `list_sessions` returns pinned sessions first and the sidebar used to render that order verbatim; `groupSessionsByExpert` re-partitions and always sorts the Autopilot group first, so a pinned expert chat now sits inside its expert group below every unpinned Autopilot chat. With CHAT_PINNING and HIRE_EXPERTS both on, pinning an expert thread silently stops floating it to the top.

autogpt_platform/frontend/src/app/(platform)/copilot/useCopilotPage.ts
```diff
@@ -39,6 +41,11 @@ function hasAssistantTail(messages: UIMessage[]) {
 export function useCopilotPage() {
   const { isUserLoading, isLoggedIn } = useSupabase();
   const isModeToggleEnabled = useGetFlag(Flag.CHAT_MODE_OPTION);
+  const isExpertsEnabled = useGetFlag(Flag.HIRE_EXPERTS);
+  const [expertIdParam] = useQueryState("expertId", parseAsString);
+  const expertId = isExpertsEnabled ? expertIdParam : null;
+  const expertMap = useExpertMap();
+  const expertIdentity = expertId ? (expertMap.get(expertId) ?? null) : null;
```
0ubbe: 🤖 🟠 **Should Fix**: Expert identity is derived from the `expertId` URL param, not the session. Global search navigates to `/copilot?sessionId=<id>` (dropping `expertId`), and bookmarks/reloads lose it the same way — the expert thread then renders with no header and no per-message attribution. `getV2GetSession`'s response model has no `expert_id` (only `SessionSummaryResponse` does), so there's no client-side fallback today; the session detail endpoint needs the field.

autogpt_platform/frontend/src/app/(platform)/copilot/components/EmptySession/EmptySession.tsx
```diff
@@ -46,7 +58,24 @@ export function EmptySession({
...
+  );
+  const recipientOptions: RecipientOption[] = [
+    AUTOPILOT_RECIPIENT,
+    ...[...expertMap.entries()].map(([id, expert]) => ({
+      id,
+      name: expert.name,
+      avatarUrl: expert.avatarUrl,
+    })),
+  ];
+  const recipient =
```
0ubbe: 🤖 🟠 **Should Fix**: Unknown/archived `expertId` shows an "Autopilot" chip while every send 404s. The chip falls back to `AUTOPILOT_RECIPIENT` when the id isn't in the expert map, but `createSession` still sends that id; `list_experts` excludes archived experts and the API rejects archived/unknown ids with 404, so `/copilot?expertId=<stale>` fails every send with the generic toast and the only escape is hand-editing the URL. Clear the param or surface an unknown-recipient state instead.

autogpt_platform/frontend/src/app/(platform)/copilot/__tests__/expert-threads.test.tsx
```diff
@@ -0,0 +1,477 @@
...
+import { groupSessionsByExpert } from "../useSessionList";
+
+vi.mock("@/services/feature-flags/use-get-flag", async (importOriginal) => {
+  const actual =
+    await importOriginal<
+      typeof import("@/services/feature-flags/use-get-flag")
+    >();
+  return {
+    ...actual,
+    useGetFlag: (flag: string) => flag === "hire-experts",
```
0ubbe: 🤖 🟠 **Should Fix**: Tests hard-mock the flag on (`useGetFlag: (flag) => flag === "hire-experts"`), so nothing asserts the flag-off contract (no `/api/experts` request, ungrouped sidebar, `expertId` ignored) — and the New-Chat-stays-fresh behaviour the `useChatSession` comment claims has no test. The harnesses never remount the hook, which is exactly what hides the remount-latch Blocker; a test that re-keys the hook the way `CopilotPage` does would catch it.

autogpt_platform/frontend/src/app/(platform)/copilot/components/EmptySession/EmptySession.tsx
```diff
@@ -46,7 +58,24 @@ export function EmptySession({
...
+    recipientOptions.find((option) => option.id === expertIdParam) ??
+    AUTOPILOT_RECIPIENT;
```
0ubbe: 🤖 🟡 **Nice to Have**: On a valid deep-linked `/copilot?expertId=…`, the recipient chip reads "Autopilot" until `/api/experts` resolves (empty map while loading), then flips to the expert. Render a loading placeholder for the chip while `expertsQuery` is pending — this screen's whole job is saying who you're addressing.

autogpt_platform/frontend/src/app/(platform)/copilot/components/ChatMessagesContainer/components/ExpertAvatar/ExpertAvatar.tsx
```diff
@@ -0,0 +1,30 @@
...
+  name: string;
+  avatarUrl: string | null;
+  size?: "default" | "small";
+}
+
+export function ExpertAvatar({ name, avatarUrl, size = "default" }: Props) {
+  const dimension = size === "small" ? "h-4 w-4" : "h-6 w-6";
+  if (avatarUrl) {
+    return (
+      <img
```
0ubbe: 🤖 🟡 **Nice to Have**: Hand-rolled `` avatar bypasses the design-system `Avatar` used by `RecipientChip` in this same PR. `AvatarImage` falls back to initials on a 404; the raw `` has no error handling, so a broken `avatar_url` renders a broken-image icon in the thread header and on every assistant message (plus the `@next/next/no-img-element` warning).

autogpt_platform/frontend/src/app/(platform)/copilot/useChatSession.ts
```diff
@@ -68,6 +73,34 @@ export function useChatSession({ dryRun = false }: UseChatSessionOptions = {}) {
     }
   }, [sessionId, queryClient]);
 
+  const latestExpertSessionQuery = useGetV2ListSessions(
```
0ubbe, line 100: 🤖 🟡 **Nice to Have**: The latest-thread query fires even when adoption is impossible — `enabled: !!expertId && !sessionId` ignores the `expertId !== mountExpertIdRef.current` guard in the effect, so picking a recipient from the chip issues a `GET /api/chat/sessions?expert_id=…&limit=1` whose result is guaranteed to be discarded. Fold the mount-time check into `enabled`.

autogpt_platform/frontend/src/app/(platform)/copilot/components/ChatSidebar/ChatSidebar.tsx
```diff
@@ -365,203 +568,33 @@ export function ChatSidebar() {
...
-                  const isActive = session.id === sessionId;
-                  const nextIsActive = sessions?.[index + 1]?.id === sessionId;
-                  return (
+              ) : sessionGroups ? (
+                sessionGroups.map((group) => (
+                  <div
+                    key={group.expertId ?? "autopilot"}
+                    className="flex flex-col gap-1"
+                  >
                     <div
```
0ubbe: 🤖 🟡 **Nice to Have**: Group headers are unlabelled `<div>`s (no heading level, no `role=\"group\"`/`aria-label`), so screen-reader users get no grouping signal. Separately, users who never hired an expert now see a lone redundant "AUTOPILOT" header above their list — consider skipping headers when only the Autopilot group exists.

autogpt_platform/frontend/src/app/(platform)/copilot/useExpertMap.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export function useExpertMap(): ExpertIdentityMap {
+  const isExpertsEnabled = useGetFlag(Flag.HIRE_EXPERTS);
+  const expertsQuery = useListExperts({
+    query: {
+      enabled: isExpertsEnabled,
+      select: (response) => response.data as Expert[],
+    },
+  });
+  if (!isExpertsEnabled || !expertsQuery.data) return EMPTY_MAP;
```
0ubbe: 🤖 🔵 **Nit**: `useExpertMap` rebuilds the `Map` (and a fresh identity object per expert) on every render of all three consumers, so `expertIdentity` never has a stable reference; and `select: (response) => response.data as Expert[]` casts across the success/error union instead of narrowing on `response.status === 200`.

autogpt_platform/frontend/src/app/(platform)/copilot/components/ChatSidebar/ChatSidebar.tsx
```diff
@@ -241,6 +253,197 @@ export function ChatSidebar() {
     }
   }
 
+  const sessionGroups = isExpertsEnabled
+    ? groupSessionsByExpert(sessions)
+    : null;
+
+  function renderSessionRow(
```
0ubbe, line 254: 🤖 🔵 **Nit**: `renderSessionRow` spans ~186 lines inside a 634-line file — past the ~200-line file / ~50-line render-function guidance in AGENTS.md. This PR's extraction was the natural moment to lift it into a `ChatSessionRow` sub-component instead of an inner function closing over a dozen pieces of parent state.

autogpt_platform/frontend/src/app/(platform)/marketplace/__tests__/experts-section.test.tsx
```diff
@@ -0,0 +1,84 @@
...
+      getListExpertTemplatesMockHandler([mariaTemplate]),
+      getListExpertsMockHandler([]),
+      getHireExpertMockHandler({ expert: hiredMaria, failed_preloads: [] }),
+    );
+
+    renderMarketplace();
+
+    expect(await screen.findByText("Meet the AI Experts")).toBeDefined();
+    await userEvent.click(await screen.findByText("Maria"));
+    await userEvent.click(await screen.findByRole("button", { name: "Hire" }));
```
0ubbe: 🤖 🔴 **Blocker**: This test fails when run: the hire button renders `Hire ${expert.name}` ("Hire Maria", ExpertProfileContent.tsx:137), and `findByRole("button", { name: "Hire" })` matches the accessible name exactly — it never resolves. Verified: 'renders experts section and hires from the profile sheet' fails with TestingLibraryElementError (1 failed | 1 passed). Use `/^Hire/` or the full name.

autogpt_platform/frontend/src/app/(platform)/marketplace/components/MainAgentPage/__tests__/main.test.tsx
```diff
@@ -91,6 +91,6 @@ describe("MainAgentPage", () => {
       "AutoGPT",
     );
     expect(screen.getByText("Other agents by AutoGPT")).toBeDefined();
```
0ubbe: 🤖 🔴 **Blocker**: Test not updated for the copy change — MainAgentPage.tsx:159 now renders `Other AI workflows by ${creator}` but this assertion still expects "Other agents by AutoGPT". Verified failing when run. The sibling MainCreatorPage test was updated; this one was missed.

autogpt_platform/frontend/src/app/(platform)/team/page.tsx
```diff
@@ -0,0 +1,100 @@
...
+import { AutopilotCard } from "./components/AutopilotCard";
+import { EmptyTeamState } from "./components/EmptyTeamState";
+import { ExpertTeamCard } from "./components/ExpertTeamCard/ExpertTeamCard";
+import { useTeamPage } from "./useTeamPage";
+
+const MAIN_CLASS =
+  "container min-h-screen space-y-6 pb-20 pt-16 sm:px-8 md:px-12";
+const GRID_CLASS = "grid grid-cols-1 gap-6 md:grid-cols-2 lg:grid-cols-3";
+
+export default function TeamPage() {
```
0ubbe, line 20: 🤖 🟠 **Should Fix**: `/team` is not in `PROTECTED_PAGES` (src/lib/supabase/helpers.ts:6) and the platform layout does no auth check — a signed-out visitor renders the page, `GET /api/experts` 401s, and they get the ErrorCard instead of a redirect to `/login?next=/team`. AGENTS.md explicitly requires updating the middleware when adding protected frontend routes.

autogpt_platform/frontend/src/services/feature-flags/use-get-flag.ts
```diff
@@ -59,6 +60,7 @@ const defaultFlags = {
   [Flag.CHAT_WORKSPACE_FILES]: false,
   [Flag.CHAT_PINNING]: false,
   [Flag.TASK_PROGRESS_BAR]: false,
+  [Flag.HIRE_EXPERTS]: true,
```
0ubbe: 🤖 🟠 **Should Fix**: `HIRE_EXPERTS` defaults to `true` in `defaultFlags`, which is returned whenever LaunchDarkly is absent (self-hosted, local dev, PW mock mode) — the feature ships ON with no way to turn it off, contradicting the PR's 'flag off → hidden' contract. Every other unreleased flag here defaults to false; gates must fail closed.

autogpt_platform/frontend/src/app/(platform)/marketplace/components/ExpertsSection/useExpertsSection.ts
```diff
@@ -0,0 +1,35 @@
...
+} from "@/app/api/__generated__/endpoints/experts/experts";
+import { Expert } from "@/app/api/__generated__/models/expert";
+import { useState } from "react";
+
+export function useExpertsSection() {
+  const [selectedTemplateId, setSelectedTemplateId] = useState<string | null>(
+    null,
+  );
+
+  const templatesQuery = useListExpertTemplates({
```
0ubbe, line 15: 🤖 🟠 **Should Fix**: `useListExpertTemplates`/`useListExperts` fire unconditionally on the public marketplace — an anonymous visitor triggers 2 requests × 3 retries of 401s (console.error each) before the section silently hides via `isError`. Same for `useInstallOnExpertButton` on the public agent detail page (its own test renders with `user: null` and still fetches). Gate the queries on an authenticated user.

autogpt_platform/frontend/src/app/(platform)/team/useTeamPage.ts
```diff
@@ -0,0 +1,38 @@
+import { useListExperts } from "@/app/api/__generated__/endpoints/experts/experts";
+import { Expert } from "@/app/api/__generated__/models/expert";
+import { useState } from "react";
+
+export function useTeamPage() {
+  const [pickerExpertId, setPickerExpertId] = useState<string | null>(null);
+  const [profileExpertId, setProfileExpertId] = useState<string | null>(null);
+
+  const expertsQuery = useListExperts({
```
0ubbe, line 13: 🤖 🟠 **Should Fix**: The experts query subscribes before the flag gate resolves — while LaunchDarkly is still loading, the skeleton branch commits and `GET /api/experts` fires; only later does `notFound()` run for flag-off users. Pass `enabled: enabled && ready` into the query (as `useExpertMap` and `useInstallOnExpertButton` already do) so flag-off users make no expert requests.

autogpt_platform/frontend/src/components/molecules/InstallWorkflowPicker/InstallWorkflowPicker.tsx
```diff
@@ -0,0 +1,137 @@
...
+            />
+            {isSearching ? (
+              <Text variant="small" className="text-zinc-500">
+                Searching…
+              </Text>
+            ) : null}
+            <div className="flex flex-col gap-2">
+              {searchResults.map((agent) => (
+                <div
+                  key={agent.slug}
```
0ubbe: 🤖 🟡 **Nice to Have**: Store agent `slug` is used both as React key and as the pending-install key — slugs are unique per creator, not globally, so two creators publishing `daily-digest` produce duplicate keys and `loading={pendingKey === agent.slug}` spins on both rows. `agent.agent_graph_id` or `${agent.creator}/${agent.slug}` is unique.

autogpt_platform/frontend/src/components/molecules/InstallWorkflowPicker/useInstallWorkflowPicker.ts
```diff
@@ -0,0 +1,137 @@
...
+
+  const expertsQuery = useListExperts({
+    query: { select: (x) => x.data as Expert[], enabled: open },
+  });
+  const hiredExperts = (expertsQuery.data ?? []).filter(
+    (expert) => !expert.is_template && !expert.is_archived,
+  );
+  const targetExpert = hiredExperts.find((expert) => expert.id === expertId);
+
+  const searchResultsQuery = useGetV2ListStoreAgents(
```
0ubbe, line 52: 🤖 🟡 **Nice to Have**: The picker's marketplace search has no debounce — `searchQuery` feeds `useGetV2ListStoreAgents` per keystroke, so typing "newsletter" issues ten store searches. Other search surfaces in the repo debounce ~300ms before querying.

autogpt_platform/frontend/src/app/(platform)/team/components/ExpertTeamCard/ExpertTeamCard.tsx
```diff
@@ -0,0 +1,119 @@
...
+        }
+      }}
+      className="flex cursor-pointer flex-col gap-3 rounded-2xl border border-zinc-200 bg-white p-5 transition-all duration-200 hover:-translate-y-0.5 hover:border-zinc-300 hover:shadow-[0_16px_40px_-16px_rgba(16,24,40,0.18)]"
+    >
+      <div className="flex items-center gap-3">
+        <Avatar className="h-12 w-12">
+          {expert.avatar_url ? (
+            <AvatarImage src={expert.avatar_url} alt={expert.name} />
+          ) : null}
+          <AvatarFallback>{expert.name}</AvatarFallback>
```
0ubbe, line 48: 🤖 🟡 **Nice to Have**: `AvatarFallback` children seed BoringAvatar's pattern — this card seeds with the full name ("Maria") while `ExpertCard` and `ExpertProfileContent` seed with `name.slice(0, 2)` ("Ma"), so an expert without `avatar_url` shows different marble patterns on the marketplace card vs the team card. Seed consistently.

autogpt_platform/frontend/src/app/(platform)/marketplace/components/ExpertsSection/components/ExpertProfileSheet/ExpertProfileContent.tsx
```diff
@@ -0,0 +1,178 @@
...
+      </div>
+      <p
+        className={cn(
+          "whitespace-pre-line text-base leading-relaxed text-zinc-600",
+          !isExpanded && "line-clamp-4",
+        )}
+      >
+        {text}
+      </p>
+      <button
```
0ubbe: 🤖 🟡 **Nice to Have**: "Read more" renders unconditionally while the bio only clamps at four lines — a one-line bio still gets a toggle that expands nothing. Gate it on actual overflow or a length threshold.

autogpt_platform/frontend/src/app/(platform)/marketplace/components/ExpertsSection/components/ExpertProfileSheet/useExpertProfileSheet.tsx
```diff
@@ -0,0 +1,82 @@
...
+      await queryClient.invalidateQueries({
+        queryKey: getListExpertsQueryKey(),
+      });
+      toast({
+        title: `${result.expert.name} joined your team`,
+        description: result.failed_preloads.length
+          ? `Couldn't attach: ${result.failed_preloads.join(", ")}`
+          : undefined,
+        variant: "success",
+        action: createElement(
```
0ubbe: 🤖 🟡 **Nice to Have**: The success toast composes raw `<button className=\"font-medium underline\">` nodes via `createElement` to avoid JSX in a .ts file — bypassing the design-system `Button` and doing `router.push` where `<Link>` is the convention. Rename to .tsx and use Button/Link.

autogpt_platform/frontend/src/app/(platform)/marketplace/components/MainMarketplacePage/MainMarketplacePage.tsx
```diff
@@ -38,22 +41,28 @@ export const MainMarkeplacePage = () => {
...
-        <Separator className="mb-6 mt-24 bg-transparent" />
-
+        {isHireExpertsEnabled ? <ExpertsSection /> : null}
         {topAgents && (
-          <AgentsSection sectionTitle="All Agents" agents={topAgents.agents} />
+          <div className="mb-20">
+            <AgentsSection
+              sectionTitle="All AI Workflows"
+              titleIcon={<AICatalogIcon size={30} />}
+              subtitle="Install one on an Expert, or run it standalone."
```
0ubbe: 🤖 🟡 **Nice to Have**: Expert-specific copy leaks past the flag: the AgentsSection subtitle "Install one on an Expert, or run it standalone." and the reworked hero headline render unconditionally, so flag-off users see copy referencing a feature with no UI anywhere on the page.

autogpt_platform/frontend/src/components/layout/AppSidebar/components/TeamIcon.tsx
```diff
@@ -0,0 +1,23 @@
+interface Props {
+  className?: string;
+}
+
+export function TeamIcon({ className }: Props) {
```
0ubbe: 🤖 🟡 **Nice to Have**: Hand-rolled Heroicons SVG where AGENTS.md mandates Phosphor icons only — and it forced `NavLink.icon` to widen from Phosphor's `Icon` type to `ComponentType<{className?: string}>`. Phosphor's `UsersThreeIcon` is the same glyph and is already used by this PR in `EmptyTeamState`.

autogpt_platform/frontend/src/components/atoms/AITeamIcon/AITeamIcon.tsx
```diff
@@ -0,0 +1,20 @@
+interface Props {
+  size?: number;
+  className?: string;
+}
+
+export function AITeamIcon({ size = 16, className }: Props) {
```
0ubbe, line 6: 🤖 🔵 **Nit**: New design-system atom without a Storybook story (CONTRIBUTING requires one for `src/components`), and the single `d` attribute is ~17KB of path data shipped on both the marketplace and /team routes — an optimised or externalised asset would be cheaper.

autogpt_platform/frontend/src/app/(platform)/marketplace/components/ExpertsSection/components/ExpertProfileSheet/ExpertProfileSheet.tsx
```diff
@@ -0,0 +1,67 @@
+"use client";
+
+import { Expert } from "@/app/api/__generated__/models/expert";
+import { Dialog } from "@/components/molecules/Dialog/Dialog";
+import { Sheet, SheetContent, SheetTitle } from "@/components/ui/sheet";
```
0ubbe, line 5: 🤖 🔵 **Nit**: Direct shadcn primitive import in feature code — `Sheet`/`SheetContent`/`SheetTitle` from `@/components/ui/sheet`. CONTRIBUTING says feature code consumes only design-system components under `src/components/{atoms,molecules,organisms}`.

autogpt_platform/frontend/src/app/(platform)/marketplace/__tests__/install-on-expert.test.tsx
```diff
@@ -0,0 +1,183 @@
...
+  });
+
+  test("hides the action when the user has no hired experts", async () => {
+    useStoreHandlers();
+    server.use(getListExpertsMockHandler([]));
+
+    renderAgentPage();
+
+    await screen.findByTestId("agent-title");
+    await new Promise((resolve) => setTimeout(resolve, 100));
```
0ubbe: 🤖 🔵 **Nit**: Fixed 100ms sleep to assert absence — `await new Promise((r) => setTimeout(r, 100))` makes the negative assertion timing-dependent; it silently stops testing anything if the experts request slows in CI. Wait for a positive signal (e.g. the Add-to-library button) before `queryByRole(...).toBeNull()`.

autogpt_platform/backend/backend/copilot/expert_context.py
```diff
@@ -33,6 +38,40 @@ def _escape(value: str) -> str:
...
+    """Build the ``<expert_identity>`` system-prompt suffix for an expert
+    session.
+
+    Returns ``""`` for plain sessions (keeps the system prompt byte-identical
+    for cross-user caching) and on any lookup failure.
+    """
+    if not user_id or not expert_id:
+        return ""
+    try:
+        expert = await experts_db().get_expert(user_id, expert_id)
```
0ubbe: 🤖 🟡 **Nice to Have**: Turn 1 of an expert session issues `get_expert` twice — once for the system-prompt suffix and once for `_expert_session_context` — and both always apply `_WORKFLOW_INCLUDE` (Expert + ExpertWorkflow + LibraryAgent + StoreListingVersion joins) though the suffix only reads name/role/identity. In the Prisma-less executor that's two serial DatabaseManager RPC round-trips on the chat critical path. Consider a lightweight lookup for the suffix or passing the fetched expert through.

autogpt_platform/backend/backend/api/features/experts/models.py
```diff
@@ -16,6 +16,8 @@ class Expert(BaseModel):
     avatar_url: str | None
     role: str
     tagline: str | None
+    bio: str | None = None
```
0ubbe: 🤖 🔵 **Nit**: `bio`/`skills` are declared with defaults so the OpenAPI schema marks them optional (`bio?`, `skills?` in the generated client) even though `_to_model` always populates them — forcing defensive guards like `expert.skills && expert.skills.length > 0` in ExpertCard for a field the API never omits. Drop the defaults to match the rest of the model.

autogpt_platform/backend/migrations/20260728100000_add_expert_bio/migration.sql
```diff
@@ -0,0 +1,2 @@
+-- AlterTable
```
0ubbe, line 1: 🤖 🟠 **Should Fix**: This `feat(frontend)` PR carries an undocumented backend slice — two schema migrations (`add_expert_bio`, `add_expert_skills`), Expert model/API changes, seed and DatabaseManager routing changes — none of which appear in the PR description's Changes list. Reviewers (and the release notes) need these visible: add them to the description's What/Changes, or move the backend commits down the stack to #13685 where the Expert entity lives.

autogpt_platform/backend/backend/api/features/experts/seed.py
```diff
@@ -95,6 +127,8 @@ async def _upsert_template(entry: RosterEntry) -> prisma.models.Expert:
         "tagline": entry["tagline"],
         "avatarUrl": entry["avatar_url"],
         "identity": entry["identity"],
+        "bio": entry["bio"],
```
0ubbe, line 132: 🤖 🟠 **Should Fix**: Roster avatars/bio/skills never reach already-hired experts. `_upsert_template` writes them onto the template row only, and `hire_expert` short-circuits to the existing copy, so nothing backfills hired rows — a user who hired Maria before this deploy keeps `avatarUrl=NULL`, `bio=NULL`, `skills=[]` permanently: /team shows initials and no skills chips while the marketplace card shows both. Either backfill hired copies in the seed (matching on `sourceTemplateId`) or fall back to the source template's presentation fields at read time.

autogpt_platform/backend/backend/copilot/expert_context_test.py
```diff
@@ -64,22 +64,70 @@ def _expert(
...
+        with patch(f"{_EC}.experts_db", MagicMock(return_value=mock_db)):
+            result = await build_expert_identity_suffix("user-1", "exp-1")
 
         mock_db.get_expert.assert_awaited_once_with("user-1", "exp-1")
         assert "<expert_identity>" in result
         assert "</expert_identity>" in result
         assert "Maria" in result
         assert "SEO Specialist" in result
         assert "You are Maria, a meticulous SEO specialist." in result
+        assert "never present yourself as AutoPilot" in result
```
0ubbe, line 94: 🤖 🔵 **Nit**: The module docstring of this test file (and of `expert_context.py`'s pre-move sections) still describes the old design — expert sessions rendering `<expert_identity>` in the first-message block — while this PR moved identity into the system-prompt suffix and the tests below assert the opposite. Update the docstrings to match.

autogpt_platform/frontend/src/app/(platform)/copilot/useChatSession.ts
0ubbe: 🤖 Fixed in [88d8b4037](https://github.com/Significant-Gravitas/AutoGPT/commit/88d8b403717c5713b42e0be94a9c3dd4e993999e): handleNewChat now clears the expertId param (flag-gated), so the remounted host has nothing to adopt; the mount latch stays for deep links. Regression test added that remounts the way CopilotPage keys the host.

autogpt_platform/frontend/src/app/(platform)/copilot/useChatSession.ts
0ubbe: 🤖 Fixed in [88d8b4037](https://github.com/Significant-Gravitas/AutoGPT/commit/88d8b403717c5713b42e0be94a9c3dd4e993999e): Adoption is latched off once createSession starts, and the composer is disabled while the latest-thread lookup is in flight (isAdoptingExpertSession), so the first message can neither be flushed into the old thread nor lost with a draft.

autogpt_platform/frontend/src/app/(platform)/copilot/components/ChatSidebar/ChatSidebar.tsx
0ubbe: 🤖 Fixed in [88d8b4037](https://github.com/Significant-Gravitas/AutoGPT/commit/88d8b403717c5713b42e0be94a9c3dd4e993999e): Pinned sessions now render flat above the expert groups (pinned-first contract restored); groups below contain only unpinned sessions, Autopilot first. Covered by a new pinned-ordering test.

autogpt_platform/frontend/src/app/(platform)/copilot/components/EmptySession/EmptySession.tsx
0ubbe: 🤖 Fixed in [88d8b4037](https://github.com/Significant-Gravitas/AutoGPT/commit/88d8b403717c5713b42e0be94a9c3dd4e993999e): Once the experts query resolves and the param is unknown, the param is cleared (new useRecipientPicker), so the chip and createSession agree on Autopilot instead of 404-ing sends.

autogpt_platform/frontend/src/app/(platform)/copilot/__tests__/expert-threads.test.tsx
0ubbe: 🤖 Fixed in [88d8b4037](https://github.com/Significant-Gravitas/AutoGPT/commit/88d8b403717c5713b42e0be94a9c3dd4e993999e): Flag mock is now a mutable map; added flag-off test (no /api/experts request, ungrouped sidebar), a New Chat regression test that remounts via the CopilotPage-style key, plus pinned-ordering and chip-loading tests.

autogpt_platform/frontend/src/app/(platform)/copilot/components/EmptySession/EmptySession.tsx
0ubbe: 🤖 Fixed in [88d8b4037](https://github.com/Significant-Gravitas/AutoGPT/commit/88d8b403717c5713b42e0be94a9c3dd4e993999e): Chip renders a role=status skeleton while the experts query is pending and a param is present, instead of a wrong Autopilot label.

autogpt_platform/frontend/src/app/(platform)/copilot/components/ChatMessagesContainer/components/ExpertAvatar/ExpertAvatar.tsx
0ubbe: 🤖 Fixed in [88d8b4037](https://github.com/Significant-Gravitas/AutoGPT/commit/88d8b403717c5713b42e0be94a9c3dd4e993999e): ExpertAvatar reimplemented on Avatar/AvatarImage/AvatarFallback like RecipientChip — broken avatar_url now falls back to initials.

autogpt_platform/frontend/src/app/(platform)/copilot/useChatSession.ts
0ubbe: 🤖 Fixed in [88d8b4037](https://github.com/Significant-Gravitas/AutoGPT/commit/88d8b403717c5713b42e0be94a9c3dd4e993999e): Mount-expert check folded into the query enabled (canAdoptExpertSession) — no request fires when adoption is impossible.

autogpt_platform/frontend/src/app/(platform)/copilot/components/ChatSidebar/ChatSidebar.tsx
0ubbe: 🤖 Fixed in [88d8b4037](https://github.com/Significant-Gravitas/AutoGPT/commit/88d8b403717c5713b42e0be94a9c3dd4e993999e): Groups are role=group with aria-labelledby; headers are skipped entirely when only the Autopilot group exists, so no-expert users keep the exact old flat list.

autogpt_platform/frontend/src/app/(platform)/copilot/useExpertMap.ts
0ubbe: 🤖 Fixed in [88d8b4037](https://github.com/Significant-Gravitas/AutoGPT/commit/88d8b403717c5713b42e0be94a9c3dd4e993999e): Map construction is memoized on the query data (identity stability for the chat tree) and select narrows on status === 200 instead of casting.

autogpt_platform/frontend/src/app/(platform)/copilot/components/ChatSidebar/ChatSidebar.tsx
0ubbe: 🤖 Fixed in [88d8b4037](https://github.com/Significant-Gravitas/AutoGPT/commit/88d8b403717c5713b42e0be94a9c3dd4e993999e): Extracted to ChatSessionRow/ChatSessionRow.tsx (function declaration, typed Props); ChatSidebar drops to 515 lines and the inner render to ~55.

autogpt_platform/frontend/src/app/(platform)/marketplace/__tests__/experts-section.test.tsx
0ubbe: 🤖 Fixed in [947f27e2b](https://github.com/Significant-Gravitas/AutoGPT/commit/947f27e2bc1de5fd2e041e74651fe97ede31a0d7): Query fixed to the full accessible name ("Hire Maria"); test passes.

autogpt_platform/frontend/src/app/(platform)/marketplace/components/MainAgentPage/__tests__/main.test.tsx
0ubbe: 🤖 Fixed in [947f27e2b](https://github.com/Significant-Gravitas/AutoGPT/commit/947f27e2bc1de5fd2e041e74651fe97ede31a0d7): Assertion updated to "Other AI workflows by AutoGPT"; test passes.

autogpt_platform/frontend/src/app/(platform)/team/page.tsx
0ubbe: 🤖 Fixed in [947f27e2b](https://github.com/Significant-Gravitas/AutoGPT/commit/947f27e2bc1de5fd2e041e74651fe97ede31a0d7): "/team" added to PROTECTED_PAGES — signed-out visitors now redirect to /login?next=/team.

autogpt_platform/frontend/src/services/feature-flags/use-get-flag.ts
0ubbe: 🤖 Fixed in [947f27e2b](https://github.com/Significant-Gravitas/AutoGPT/commit/947f27e2bc1de5fd2e041e74651fe97ede31a0d7): HIRE_EXPERTS now defaults to false (fail closed). Tests are unaffected — they mock the flag hook directly.

autogpt_platform/frontend/src/app/(platform)/marketplace/components/ExpertsSection/useExpertsSection.ts
0ubbe: 🤖 Fixed in [947f27e2b](https://github.com/Significant-Gravitas/AutoGPT/commit/947f27e2bc1de5fd2e041e74651fe97ede31a0d7): Experts queries in useExpertsSection and useInstallOnExpertButton are gated on isLoggedIn from useSupabase (same pattern as AddToLibraryButton); tests added asserting no /api/experts request when signed out.

autogpt_platform/frontend/src/app/(platform)/team/useTeamPage.ts
0ubbe: 🤖 Fixed in [947f27e2b](https://github.com/Significant-Gravitas/AutoGPT/commit/947f27e2bc1de5fd2e041e74651fe97ede31a0d7): useTeamPage takes enabled and the page passes enabled && ready, so nothing fetches until LaunchDarkly answers.

autogpt_platform/frontend/src/components/molecules/InstallWorkflowPicker/InstallWorkflowPicker.tsx
0ubbe: 🤖 Fixed in [947f27e2b](https://github.com/Significant-Gravitas/AutoGPT/commit/947f27e2bc1de5fd2e041e74651fe97ede31a0d7): Row key and pendingKey now use agent_graph_id (globally unique).

autogpt_platform/frontend/src/components/molecules/InstallWorkflowPicker/useInstallWorkflowPicker.ts
0ubbe: 🤖 Fixed in [947f27e2b](https://github.com/Significant-Gravitas/AutoGPT/commit/947f27e2bc1de5fd2e041e74651fe97ede31a0d7): Search value flows through useDebouncedValue (300ms) before hitting the store-agents query.

autogpt_platform/frontend/src/app/(platform)/team/components/ExpertTeamCard/ExpertTeamCard.tsx
0ubbe: 🤖 Fixed in [947f27e2b](https://github.com/Significant-Gravitas/AutoGPT/commit/947f27e2bc1de5fd2e041e74651fe97ede31a0d7): All surfaces (ExpertCard, ExpertProfileContent, ExpertTeamCard, InstallWorkflowPicker, and RecentChats) now seed the fallback with the full name — same generated avatar everywhere.

autogpt_platform/frontend/src/app/(platform)/marketplace/components/ExpertsSection/components/ExpertProfileSheet/ExpertProfileContent.tsx
0ubbe: 🤖 Fixed in [947f27e2b](https://github.com/Significant-Gravitas/AutoGPT/commit/947f27e2bc1de5fd2e041e74651fe97ede31a0d7): Clamp + Read more only render when the bio exceeds 280 chars.

autogpt_platform/frontend/src/app/(platform)/marketplace/components/ExpertsSection/components/ExpertProfileSheet/useExpertProfileSheet.tsx
0ubbe: 🤖 Fixed in [947f27e2b](https://github.com/Significant-Gravitas/AutoGPT/commit/947f27e2bc1de5fd2e041e74651fe97ede31a0d7): File renamed to .tsx; toast actions are design-system Buttons rendered as links — createElement raw buttons gone.

0ubbe: 🤖 Fixed in [947f27e2b](https://github.com/Significant-Gravitas/AutoGPT/commit/947f27e2bc1de5fd2e041e74651fe97ede31a0d7): File renamed to .tsx; toast actions are design-system Buttons rendered as links — createElement raw buttons gone.

autogpt_platform/frontend/src/app/(platform)/marketplace/components/MainMarketplacePage/MainMarketplacePage.tsx
0ubbe: 🤖 Fixed in [947f27e2b](https://github.com/Significant-Gravitas/AutoGPT/commit/947f27e2bc1de5fd2e041e74651fe97ede31a0d7): Expert subtitle and hero copy are now flag-gated; flag-off users see the previous neutral copy.

autogpt_platform/frontend/src/components/layout/AppSidebar/components/TeamIcon.tsx
0ubbe: 🤖 Fixed in [947f27e2b](https://github.com/Significant-Gravitas/AutoGPT/commit/947f27e2bc1de5fd2e041e74651fe97ede31a0d7): Hand-rolled SVG deleted; sidebar uses Phosphor UsersThreeIcon and NavLink.icon is back to Phosphor's Icon type.

autogpt_platform/frontend/src/components/atoms/AITeamIcon/AITeamIcon.tsx
0ubbe: 🤖 Fixed in [947f27e2b](https://github.com/Significant-Gravitas/AutoGPT/commit/947f27e2bc1de5fd2e041e74651fe97ede31a0d7): Story added (Default/Sizes/Colored). The SVG path stays as-is — it is a single brand asset and gzips well; happy to optimise separately if it shows up in bundle analysis.

autogpt_platform/frontend/src/app/(platform)/marketplace/__tests__/install-on-expert.test.tsx
0ubbe: 🤖 Fixed in [947f27e2b](https://github.com/Significant-Gravitas/AutoGPT/commit/947f27e2bc1de5fd2e041e74651fe97ede31a0d7): Sleep replaced with waitFor on the MSW request signal — the negative assertion now waits for the real positive event.

autogpt_platform/backend/backend/copilot/expert_context.py
0ubbe: 🤖 Fixed in [5b4b6c3fe](https://github.com/Significant-Gravitas/AutoGPT/commit/5b4b6c3fecfd469434fb556ab76a40ce0d54a55a): get_expert gains include_workflows (keyword-only, defaults True); the identity suffix passes False, skipping the workflow joins on the per-turn path. The kwarg survives the DatabaseManager RPC (same pattern as get_graph's for_export).

autogpt_platform/backend/backend/api/features/experts/models.py
0ubbe: 🤖 Fixed in [5b4b6c3fe](https://github.com/Significant-Gravitas/AutoGPT/commit/5b4b6c3fecfd469434fb556ab76a40ce0d54a55a): Defaults dropped — bio/skills are required in the API model and the generated client; all fixtures updated.

autogpt_platform/backend/backend/api/features/experts/seed.py
0ubbe: 🤖 Fixed in [5b4b6c3fe](https://github.com/Significant-Gravitas/AutoGPT/commit/5b4b6c3fecfd469434fb556ab76a40ce0d54a55a): seed_roster now backfills avatarUrl/bio/skills onto hired copies via update_many on sourceTemplateId (name/role/identity deliberately excluded); covered by a new re-seed test asserting the hired copy picks up the fields and keeps the user's rename.

autogpt_platform/backend/backend/copilot/expert_context_test.py
0ubbe: 🤖 Fixed in [5b4b6c3fe](https://github.com/Significant-Gravitas/AutoGPT/commit/5b4b6c3fecfd469434fb556ab76a40ce0d54a55a): Docstrings updated to the current design (identity in the system-prompt suffix; workflows/team blocks in the first message), including a stale arg docstring in copilot/service.py.

autogpt_platform/frontend/src/app/(platform)/copilot/useCopilotPage.ts
0ubbe: 🤖 Fixed in [030e0f589](https://github.com/Significant-Gravitas/AutoGPT/commit/030e0f589826fd88ae2cf2db268d2d6df45248d3) + [88d8b4037](https://github.com/Significant-Gravitas/AutoGPT/commit/88d8b403717c5713b42e0be94a9c3dd4e993999e): the session detail endpoint now returns expert_id (added on the #13687 branch and merged up), and the client derives thread identity from the open session's expert_id — the URL param only seeds new sessions, so global search, bookmarks, and reloads keep attribution.

autogpt_platform/frontend/src/app/(platform)/marketplace/components/ExpertsSection/components/ExpertProfileSheet/ExpertProfileSheet.tsx
0ubbe: 🤖 Keeping the direct import, deliberately: there is no design-system Sheet/Drawer today — molecules/Dialog is a centered dialog with a vaul bottom-drawer on mobile, not a right-side sheet — and wrapping the shadcn Sheet as a proper design-system molecule is out of scope for this PR (one other feature file already imports it directly). Happy to do the design-system wrapper as a follow-up.

autogpt_platform/backend/migrations/20260728100000_add_expert_bio/migration.sql
0ubbe: 🤖 Fixed: the PR description's Changes section now documents the backend slice — both migrations, the bio/skills API fields, roster seed additions, DatabaseManager routing, and the identity move to the system-prompt suffix.

Swiftyos (dismissed):

Pwuts:

All resources for PR #13689 have been removed:

autogpt_platform/frontend/src/components/layout/AppSidebar/components/RecentChats/RecentChats.tsx
```diff
@@ -71,15 +82,72 @@ export function RecentChats() {
     );
   }
 
+  const groups = isExpertsEnabled
```
kcze, line 85: 🤖 🟠 **Should Fix**: With `hire-experts` on, expert grouping *unconditionally* replaces date grouping here, with none of the `groups.length <= 1` guard that `groupSessionsForSidebar` applies in `ChatSidebar`. A flag-on user who has hired nobody loses Today/Yesterday/Last 7 days entirely and gets a single redundant "Autopilot" header instead — the exact defect that was fixed in `ChatSidebar` but missed in this sidebar. _(flagged by: Claude)_

autogpt_platform/frontend/src/app/(platform)/copilot/store.ts
```diff
@@ -206,6 +216,14 @@ export const useCopilotUIStore = create<CopilotUIState>((set, get) => ({
   initialPrompt: null,
   setInitialPrompt: (prompt) => set({ initialPrompt: prompt }),
 
+  adoptedExpertThreads: new Set<string>(),
+  markExpertThreadAdopted: (expertId) =>
```
kcze, line 220: 🤖 🟠 **Should Fix**: `adoptedExpertThreads` is never cleared for the life of the page load, so deep-link adoption is one-shot per expert rather than per "user asked for a new chat". Clicking **Chat** on an `ExpertTeamCard` (`/copilot?expertId=X`) resumes Maria's thread the first time; every later click in the same SPA session lands on a blank new-task screen and creates yet another Maria thread. Same button, two different behaviours. _(flagged by: Claude + Codex)_

autogpt_platform/frontend/src/app/(platform)/team/components/ExpertTeamCard/ExpertTeamCard.tsx
```diff
@@ -0,0 +1,119 @@
...
+    onInstallWorkflow(expert.id);
+  }
+
+  return (
+    <div
+      role="button"
+      tabIndex={0}
+      onClick={() => onOpenProfile(expert.id)}
+      onKeyDown={(event) => {
+        if (event.key === "Enter" || event.key === " ") {
```
kcze, line 36: 🤖 🟠 **Should Fix**: The card's `onKeyDown` calls `preventDefault()` + `onOpenProfile()` for Enter/Space without checking `event.target === event.currentTarget`, so keydowns bubbling from the nested "Chat" link and "Install workflow" button are swallowed. A keyboard user tabbing to "Install workflow" and pressing Enter gets the profile drawer instead — both per-card actions are unreachable by keyboard. The `onClick` `stopPropagation` guards only cover mouse. _(flagged by: Claude + Codex)_

autogpt_platform/frontend/src/components/molecules/InstallWorkflowPicker/useInstallWorkflowPicker.ts
```diff
@@ -0,0 +1,145 @@
...
+      });
+      onClose();
+      return;
+    }
+    try {
+      await installWorkflow({
+        expertId: expert.id,
+        data: { store_listing_version_id: versionId },
+      });
+      await queryClient.invalidateQueries({
```
kcze, line 82: 🤖 🟠 **Should Fix**: Install only invalidates `getListExpertsQueryKey()`, but the server's `install_workflow` also calls `add_store_agent_to_library()` (`experts_db.py:223`). Every library query stays stale: after installing on an expert from the marketplace agent page, the agent *is* in the user's library yet the page still shows "Add to library", and /library is stale until a background refetch. `AddToLibraryButton.tsx` invalidates `getGetV2ListLibraryAgentsQueryKey()` for exactly this. _(flagged by: Claude)_

autogpt_platform/frontend/src/app/(platform)/marketplace/components/ExpertsSection/components/ExpertProfileSheet/useExpertProfileSheet.tsx
```diff
@@ -0,0 +1,73 @@
...
+      ));
+
+  const { mutateAsync: hireExpert, isPending: isHiring } = useHireExpert();
+
+  async function hire() {
+    if (!expert || !expert.is_template) return;
+    try {
+      const response = await hireExpert({ data: { template_id: expert.id } });
+      const result = response.data as HireResult;
+      await queryClient.invalidateQueries({
```
kcze, line 36: 🤖 🟠 **Should Fix**: Same gap on hire — `hire_expert` → `_install_preloads()` adds one library agent per preloaded workflow (`experts_db.py:176`), but only the experts key is invalidated. After hiring Maria the user's /library and any `isAgentAddedToLibrary` state on an open marketplace page do not reflect the 2–3 agents just added. _(flagged by: Claude)_

autogpt_platform/frontend/src/app/(platform)/copilot/components/EmptySession/useRecipientPicker.ts
```diff
@@ -0,0 +1,53 @@
...
+    parseAsString,
+  );
+
+  // An ?expertId= pointing at an expert the user can no longer address
+  // (archived, deleted, or simply wrong) would leave the chip reading
+  // "Autopilot" while `createSession` still sent the id — which the backend
+  // rejects with a 404 on every send. Drop it so both agree on Autopilot.
+  useEffect(
+    function clearUnknownExpertParam() {
+      if (!hasLoadedExperts || !expertIdParam) return;
```
kcze, line 25: 🤖 🟡 **Nice to Have**: The unknown-expert cleanup only runs on query *success*. If `GET /api/experts` errors, `hasLoadedExperts` stays false so the stale `expertId` is never cleared, while `isLoadingRecipient` goes false and the chip renders "Autopilot" — yet `createSession` still sends the URL's expert id, so every send 404s under an Autopilot label. The success path of this was fixed; the error path was not. _(flagged by: Codex)_

autogpt_platform/frontend/src/app/(platform)/team/components/AutopilotCard.tsx
```diff
@@ -0,0 +1,33 @@
...
+import { Text } from "@/components/atoms/Text/Text";
+import { RobotIcon } from "@phosphor-icons/react";
+
+export function AutopilotCard() {
+  return (
+    <div className="flex flex-col gap-3 rounded-2xl border border-zinc-200 bg-white p-5">
+      <div className="flex items-center gap-3">
+        <Avatar className="h-12 w-12">
+          <AvatarFallback>
+            <RobotIcon size={24} />
```
kcze, line 12: 🤖 🟡 **Nice to Have**: `AvatarFallback` never renders its children — `Avatar.tsx:180` uses them only as the `BoringAvatar` seed and only when `typeof children === "string"`. So `<AvatarFallback><RobotIcon/></AvatarFallback>` renders a random marble seeded with the literal string `"User"`, not the robot, and the `RobotIcon` import is dead code. _(flagged by: Claude)_

autogpt_platform/frontend/src/app/(platform)/copilot/useChatSession.ts
```diff
@@ -68,6 +74,66 @@ export function useChatSession({ dryRun = false }: UseChatSessionOptions = {}) {
...
+  // change, so a late adoption would post that message into the old thread.
+  const sendStartedRef = useRef(false);
+  const canAdoptExpertSession =
+    !!expertId &&
+    !sessionId &&
+    expertId === mountExpertIdRef.current &&
+    !adoptedExpertThreads.has(expertId);
+
+  const latestExpertSessionQuery = useGetV2ListSessions(
+    { expert_id: expertId ?? undefined, limit: 1 },
```
kcze, line 101: 🤖 🟡 **Nice to Have**: The adoption query uses `limit: 1`, but `list_sessions` orders `[{isPinned: desc}, {updatedAt: desc}]` (`copilot/db.py:599`). For a user who pinned an old thread of that expert, the deep link adopts the *pinned* thread, not the latest one the surrounding comments promise. _(flagged by: Claude + Codex)_

autogpt_platform/backend/backend/api/features/experts/seed.py
```diff
@@ -113,6 +149,25 @@ async def _upsert_template(entry: RosterEntry) -> prisma.models.Expert:
...
+async def _backfill_hired_copies(template: prisma.models.Expert) -> int:
+    """Push the template's presentation fields onto experts hired from it.
+
+    A hire copies the template row, so roster updates would otherwise only
+    ever reach new hires and everyone who hired earlier would keep a blank
+    avatar/bio/skills forever. ``name`` is deliberately excluded — users may
+    have renamed their hire — as are ``role``/``identity``, which drive live
+    persona behaviour.
+    """
+    return await prisma.models.Expert.prisma().update_many(
```
kcze, line 161: 🤖 🟡 **Nice to Have**: `_backfill_hired_copies` refreshes `avatarUrl`/`bio`/`skills` but not `tagline`, even though `_upsert_template` does update `tagline` on the template. Edit a tagline in `ROSTER` and re-seed: the marketplace card shows the new one, every already-hired copy keeps the old one forever, sitting next to the freshly backfilled bio/skills. _(flagged by: Claude)_

autogpt_platform/backend/backend/copilot/expert_context.py
```diff
@@ -33,6 +38,45 @@ def _escape(value: str) -> str:
...
+        expert = await experts_db().get_expert(
+            user_id, expert_id, include_workflows=False
+        )
+    except Exception as e:
+        logger.warning(f"Failed to build expert identity suffix: {e}")
+        return ""
+    if expert is None or expert.is_archived:
+        return ""
+
+    name = _escape(expert.name)
```
kcze, line 65: 🤖 🟡 **Nice to Have**: This moves the user-controlled expert `name` (set at hire via `HireRequest.name`, only `max_length=100`) from a first-user-message block into the **system prompt**, where `_escape()` neutralises `<`/`>` but not newlines or instruction-shaped text. Hiring with `name = "Maria.\nNEW RULE: ..."` lands that string verbatim inside the trusted `<expert_identity>` block. Self-scoped, but it is a real trust-boundary change worth a newline/charset guard. _(flagged by: Claude)_

autogpt_platform/frontend/src/app/(platform)/copilot/components/ChatSidebar/components/ChatSessionRow/ChatSessionRow.tsx
```diff
@@ -0,0 +1,217 @@
...
+import type { SessionSummaryResponse } from "@/app/api/__generated__/models/sessionSummaryResponse";
+import {
+  DropdownMenu,
+  DropdownMenuContent,
+  DropdownMenuItem,
+  DropdownMenuTrigger,
+} from "@/components/molecules/DropdownMenu/DropdownMenu";
+import { cn } from "@/lib/utils";
+import {
+  CircleNotch,
```
kcze, line 12: 🤖 🟡 **Nice to Have**: New file imports the deprecated bare Phosphor exports `CircleNotch` and `DotsThree` instead of the `-Icon`-suffixed aliases required by `frontend/AGENTS.md` — the other six icons in the same import statement are correct. _(flagged by: Claude)_

autogpt_platform/frontend/src/components/molecules/InstallWorkflowPicker/useInstallWorkflowPicker.ts
```diff
@@ -0,0 +1,145 @@
...
+
+export function useInstallWorkflowPicker({
+  mode,
+  storeListingVersionId,
+  expertId,
+  open,
+  onClose,
+}: Args) {
+  const queryClient = useQueryClient();
+  const [searchQuery, setSearchQuery] = useState("");
```
kcze, line 36: 🤖 🟡 **Nice to Have**: `searchQuery` is never reset when the picker closes, and the component stays mounted (`team/page.tsx` renders it with `open={pickerExpertId !== null}`). A search typed while installing onto expert A is still applied when the picker reopens for expert B, silently hiding most of the catalog. _(flagged by: Claude)_

PR #13692, opened 2026-07-27
fix(blocks): set ensure_ascii=True when serializing AGPT_VARIABLES in code_executor_helpers

amir-rezaei:
## Description
This PR resolves issue #13551 by using `json.dumps(variables, ensure_ascii=True)` in `build_variable_injection()` (`autogpt_platform/backend/backend/blocks/code_executor_helpers.py`).

## Details
- Ensures serialized `AGPT_VARIABLES` environment payloads contain only ASCII characters, preventing `UnicodeEncodeError: surrogates not allowed` when upstream blocks pass data containing emojis or surrogate pairs.


Pwuts:
Duplicate of #13576, and not a fix

PR #13696, opened 2026-07-28
feat(backend): add workspace file move/copy and workspace folder tools

Abhi1992002:
### Why / What / How

**Why.** The workspace file API had no move or copy primitive. The only way to relocate a file was a three-call workaround:

1. `read_workspace_file(file_id, save_to_path=...)` — downloads the whole file into the sandbox
2. `write_workspace_file(filename, path, source_path=...)` — uploads it to the new path
3. `delete_workspace_file(file_id)` — deletes the original

That pulls the entire file through the agent's context — base64-encoded for binary — purely to change a path string. A 2.2 MB PNG burned a large slice of the context budget without a single byte ever being inspected. It also isn't atomic: if the session dies between steps 2 and 3 you end up with two copies and no way to tell which is authoritative. And it's non-obvious enough that it needed a dedicated skill entry to execute reliably — an agent observed without it fell back to rewriting file content from scratch.

Separately, workspace folders (the grouping shown on the Artifacts page, backed by `UserWorkspaceFolder`) already existed with a full data layer and REST API, but were invisible to the agent — there were no tools for them at all.

**What.** Six new CoPilot tools:

| Tool | Purpose |
|---|---|
| `move_workspace_file` | Move/rename a file to a new virtual path |
| `copy_workspace_file` | Duplicate a file to a new virtual path |
| `create_workspace_folder` | Create a file folder |
| `delete_workspace_folder` | Delete a folder, returning its files to the root |
| `list_workspace_folders` | List folders with file counts |
| `move_workspace_files_to_folder` | Bulk-change folder membership |

**How.** Both transfer operations are server-side; the bytes never enter this process or the model's context.

- **Move** is metadata-only. It rewrites `name`/`path` (and optionally `folderId`) and deliberately leaves `storagePath` alone, so no bytes move at all. The stale filename segment left in the blob path is harmless — nothing reads the filename back out of `storagePath`; `get_download_url` parses out the file ID only.
- **Copy** adds a `copy()` primitive to `WorkspaceStorageBackend`: a GCS server-side blob copy, and `shutil.copyfile` (off the event loop) for local storage. It is quota-enforced but does not re-run the virus scan, since these exact bytes were already scanned by `write_file` when they first entered the workspace and a copy cannot change them.

Path/folder are two orthogonal concepts here and the tools keep them that way: `move_workspace_file` changes the virtual path, `move_workspace_files_to_folder` changes folder membership, and neither affects the other. Tool descriptions spell out the distinction from the existing `create_folder`/`delete_folder` tools, which manage *library* folders holding agents.

Destination collisions fail by default with an error naming the conflicting path, with `overwrite=true` as the opt-in escape hatch — mirroring `write_workspace_file`. Folder deletion reparents files to the root rather than deleting them, matching the existing `DELETE /api/workspace/folders/{id}` behavior.

Two other issues from the ticket:

- **Misleading truncation error.** Omitting only `filename` from `write_workspace_file` reported "Tool call appears truncated (no arguments received)" even when plenty of content was supplied. `content`/`content_base64`/`source_path` are named parameters, so probing `**kwargs` for them always came up empty and `has_any_content` was unconditionally `False`. Now reads the bound parameters, and the genuine no-arguments case keeps its actionable guidance.
- **Prompt guidance.** The system prompt taught only the cross-storage move pattern, so the three-call workaround was the model's best available option. It now documents the in-workspace move/copy tools and explicitly warns against read-write-delete.

### Changes 🏗️

**New tools**
- `backend/copilot/tools/workspace_file_transfer.py` — `move_workspace_file` / `copy_workspace_file`, sharing argument resolution and the skills-registry ACL check (enforced on the source path, the destination path, and again on the resolved path when the target was given by `file_id`)
- `backend/copilot/tools/workspace_folders.py` — the four folder tools; folders resolvable by ID or by case-insensitive name

**Transfer layer**
- `backend/util/workspace_transfer.py` (new) — `move_file` / `copy_file`, including quota enforcement, collision handling, orphaned-blob cleanup when the DB insert fails, and search-index refresh (the index is keyed on name and path). Kept out of `workspace.py`, which is already ~520 lines
- `backend/util/workspace.py` — thin `move_file` / `copy_file` delegating methods
- `backend/util/workspace_storage.py` — new abstract `copy()` plus GCS and local implementations

**Data layer**
- `backend/data/workspace.py` — new `update_workspace_file_location`, workspace-scoped via `update_many` (not `update`, which matches on primary key alone and would let a cross-workspace ID or a concurrently-deleted row through); `create_workspace_file` gains an optional `folder_id`
- `backend/data/workspace_folder.py` — folder functions renamed to workspace-prefixed names so they can be exposed over the `DatabaseManager` RPC boundary alongside the identically-named library folder functions
- `backend/data/db_accessors.py` — new `workspace_folder_db()` accessor
- `backend/data/db_manager.py` — RPC exposure for the new file and folder functions

**Registration & generated artifacts**
- `backend/copilot/tools/__init__.py`, `backend/copilot/permissions.py`, `backend/copilot/tools/models.py` — tool registry entries, `ToolName` literal entries, and response types
- `frontend/src/app/api/openapi.json` — regenerated; the six new `ResponseType` enum values changed the exported schema
- `backend/copilot/tools/tool_schema_test.py` — tool-schema character budget raised for the six new tools, after trimming their descriptions
- `docs/integrations/block-integrations/misc.md` — regenerated via `scripts/generate_block_docs.py`

#### For code changes:
- [x] I have made a test plan
  - [x] `backend/util/workspace_transfer_test.py` (new, 14 tests) — move rewrites path/name without touching storage; copy delegates to the storage `copy()` and never calls `retrieve()`; quota rejection before any write; collision rejected / overwrite deletes the occupant first; same-path move is a no-op; folder preserved unless overridden; session-scoped path resolution; orphaned blob cleaned up when the DB insert conflicts
  - [x] `backend/copilot/tools/workspace_file_transfer_test.py` (new, 19 tests) — response shapes, auth, argument validation, skills-registry ACL by path and by `file_id`, error surfacing
  - [x] `backend/copilot/tools/workspace_folders_test.py` (new, 20 tests) — all four folder tools including duplicate names, case-insensitive lookup, move-to-root, and files silently dropped as out-of-workspace being reported rather than counted as success
  - [x] `backend/util/workspace_storage_test.py` — local copy duplicates bytes and leaves the source intact; missing source raises; GCS copy uses server-side copy without downloading
  - [x] `backend/copilot/tools/workspace_files_test.py` — regression tests for the missing-`filename` error; verified 4 of the 5 fail against the old code while the genuine-truncation case still passes
  - [x] Mutation-checked the transfer suite: breaking basename derivation fails exactly the two tests that assert it
  - [x] Verified tool registry and `ToolName` literal agree, and that every renamed folder function resolves on both the direct-import and RPC paths
  - [x] `poetry run format`, `poetry run lint`, `pyright`, and `pnpm types` clean; 272 tests pass locally

Note: the DB-backed suites (`folder_routes_test.py` and the round-trip tests in `workspace_files_test.py`) could not run in my sandbox — the session-scoped autouse fixture needs Docker Postgres, which isn't available there. Their patch targets were updated for the rename and verified to resolve by import; they run in CI.

No configuration changes.


Abhi1992002:
/review

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #13697, opened 2026-07-28
feat(platform): pause schedules & triggers on payment lapse, resume on payment

Abhi1992002:
### Why / What / How

**Why:** When a user's payment lapses, their cron schedules and webhook triggers keep firing. The paywall in `add_graph_execution` only rejects the run *after* the scheduler has already woken up (or the webhook has fanned out), so every tick produces a failed execution row plus log noise for a user who is no longer paying — and nothing ever resumes automatically when they pay again. (SECRT-2376)

**What:** On a paid→`NO_TIER` tier transition, all of the user's personal graph schedules are paused and their webhook-trigger presets deactivated — each stamped with a persisted `payment_lapsed` reason. On `NO_TIER`→paid, only automations carrying that marker are resumed, so anything the user paused themselves stays off. The user is notified on both edges, and the paused state is surfaced in the UI instead of the schedules silently disappearing.

**How:**
- `set_subscription_tier` (`backend/data/credit.py`) is the single write chokepoint for every lapse path (failed-invoice handler, `customer.subscription.*` webhooks, the reconciliation sweep), so the pause/resume hook lives there. It is wrapped so a scheduler failure can never 500 the Stripe webhook that triggered it.
- Schedules have no DB model (APScheduler jobstore only), so the reason is persisted in the job kwargs (`GraphExecutionJobArgs.paused_reason`) and pause/resume use APScheduler's `pause_job`/`resume_job` via new exposed scheduler-service methods. `get_execution_schedules` previously dropped any job with `next_run_time=None`; it now keeps system-paused schedules visible.
- Presets get a nullable `AgentPreset.deactivationReason` enum column. The webhook dispatcher already short-circuits on `isActive`, so deactivation is enough to drop events. A user explicitly setting `is_active` clears the reason, which is what protects manually-disabled automations from auto-resume.
- Org/team-tagged schedules and presets are excluded — they're funded by the org, not the member's personal subscription.

### Changes 🏗️

Backend
- `backend/executor/scheduler.py`: `paused_reason` on graph job args, `is_paused` on job info, `pause_user_graph_schedules`/`resume_user_graph_schedules` service methods + client bindings, paused schedules stay in listings
- `backend/data/automation_pause.py` (new): pause/resume orchestrator over schedules + presets, queues notifications
- `backend/data/credit.py`: tier-transition hook in `set_subscription_tier` (never raises into the webhook path)
- `backend/api/features/library/db.py` / `model.py`: user `is_active` updates clear `deactivationReason`; `deactivation_reason` exposed on `LibraryAgentPreset`
- `schema.prisma` + migration: `PresetDeactivationReason` enum, `AgentPreset.deactivationReason`, `NotificationType.AUTOMATIONS_PAUSED/RESUMED`, `User.notifyOnAutomationsPaused`
- Notifications: payload models, routing/template/subject entries, two email templates, preference wiring

Frontend
- "Paused — payment required" badge on schedule rows (SchedulesPanel + library sidebar) and selected-schedule view; paused/`deactivation_reason` trigger status on webhook trigger cards
- Guard against empty `next_run_time` (previously crashed `formatDistanceToNow` — paused schedules were never listed before, now they are)
- Refreshed committed `openapi.json` (generated client picks up `is_paused`, `paused_reason`, `deactivation_reason`)

Known limitation (intentionally out of scope): `queue_notification*` is a no-op when `app_env == PRODUCTION` (`backend/notifications/notifications.py`), so the new emails won't send in prod until that standing gate is revisited. The pause/resume mechanics themselves are unaffected.

#### For code changes:
- [x] I have made a test plan
  - [x] `automation_pause_test.py`: pause stamps reason on schedules+presets, resume restores only `PAYMENT_LAPSED` rows, user-deactivated presets excluded via where-clause, no notification when nothing changed
  - [x] `credit_tier_transition_test.py`: paid→NO_TIER pauses, NO_TIER→paid resumes, paid→paid no-op, scheduler failure swallowed
  - [x] `scheduler_unit_test.py`: pause/resume skip other users / org jobs / already-paused; resume only matching reason; paused schedule still listed with `is_paused=True` while fired one-shots stay hidden
  - [x] `db_test.py`: `update_preset(is_active=...)` clears `deactivationReason`
  - [x] Frontend: paused schedule renders badge + "Paused" label (followups page integration test); full `pnpm test:unit` suite green (3993 tests)

### DIFF
--- a/autogpt_platform/backend/backend/api/features/library/db.py
+++ b/autogpt_platform/backend/backend/api/features/library/db.py
@@ -2096,8 +2096,7 @@ async def update_preset(
             update_data["description"] = description
         if is_active is not None:
             update_data["isActive"] = is_active
-            # Explicitly setting the active state is a user decision; drop any
-            # system deactivation reason so auto-resume won't override it.
+            # User decision overrides any system deactivation reason
             update_data["deactivationReason"] = None
         if inputs or credentials:
             if not (inputs and credentials):
--- a/autogpt_platform/backend/backend/api/features/library/db.py
+++ b/autogpt_platform/backend/backend/api/features/library/db.py
@@ -2096,6 +2096,9 @@ async def update_preset(
             update_data["description"] = description
         if is_active is not None:
             update_data["isActive"] = is_active
+            # Explicitly setting the active state is a user decision; drop any
+            # system deactivation reason so auto-resume won't override it.
+            update_data["deactivationReason"] = None
         if inputs or credentials:
             if not (inputs and credentials):
                 raise ValueError(
--- a/autogpt_platform/backend/backend/api/features/library/db_test.py
+++ b/autogpt_platform/backend/backend/api/features/library/db_test.py
@@ -1441,3 +1441,33 @@ async def test_cleanup_trigger_agents_processes_each_independently(mocker):
     recursive_delete.assert_awaited_once_with(
         library_agent_id="sole", user_id="test-user", soft_delete=True
     )
+
+
+@pytest.mark.asyncio
+async def test_update_preset_is_active_clears_deactivation_reason():
+    """A user explicitly setting is_active must clear any system
+    deactivationReason so auto-resume can't override their choice."""
+
+    @asynccontextmanager
+    async def fake_tx():
+        yield None
+
+    with (
+        patch(
+            "backend.api.features.library.db.get_preset",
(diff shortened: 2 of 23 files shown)


Abhi1992002:
/review

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #13699, opened 2026-07-28
feat(frontend): compact wallet popover for the new layout

Abhi1992002:
### Why / What / How

**Why.** The wallet popover is `28.5rem` wide with an inline top-up form, an auto-refill config form, collapsible task groups, embedded videos and progress bars. That was designed for the classic top navbar. In the new sidebar layout the wallet lives in the sidebar footer as a small `$1.00` pill, and opening it throws a panel wider than the sidebar itself over the page. It's far too heavy for what a user wants there: how many credits do I have, how do I get more.

**What.** A compact popover used *only* when the wallet renders in the new layout (`compact` prop, set by `SidebarUserActions`). The classic navbar keeps the existing panel, byte-for-byte.

**How.** `Wallet.tsx` had grown to 360 lines mixing trigger markup, popover content, credits fetching, confetti and websocket wiring. Rather than add a second panel to it, the shared parts were pulled out first:

- `helpers.ts` — task-group definitions + `getEarnRows()`
- `useWallet.ts` — credits, balance-flash, completion counts, onboarding websocket + confetti
- `components/WalletFullPanel.tsx` — the existing classic panel, moved verbatim
- `components/WalletCompactPanel.tsx` — the new one

`Wallet.tsx` is now ~120 lines and just picks a panel.

The compact panel shows three things: balance, an **Add credits** row, and a collapsible **Earn credits** list. To keep the list short, `getEarnRows()` flattens the three task groups — a group with every task done collapses to one `First Wins · 3 of 3 — Done` row, while a group still in progress contributes one row per task.

**Add credits** opens `TopUpDialog` instead of embedding `TopUpForm` inline. That dialog's copy was hardcoded to *"You're out of automation credits"*, which is wrong when you open it deliberately with a positive balance, so it gained an `add-credits` variant. Default is unchanged, so the low-credit banner and daily auto-opener read exactly as before.

Completed rows reuse `SealCheckIcon` (`weight="fill"`, `#00a656`) from the copilot `TaskProgressBar`'s "All tasks complete" header, so completion reads the same across the app.

### Changes 🏗️

- Add `WalletCompactPanel` — balance, Add credits row, collapsible Earn credits list; rendered when `compact` is set (new layout only)
- Add `getEarnRows()` to flatten task groups; completed groups collapse to a single Done row
- Extract `helpers.ts` (task groups + row flattening) and `useWallet.ts` (credits, confetti, websocket) out of `Wallet.tsx`; move the classic panel to `WalletFullPanel.tsx`. `Wallet.tsx` 360 → ~120 lines
- Compact popover opens `side="top" align="start"` at `22rem` with `rounded-2xlarge`, sized so long task names wrap rather than truncate
- Add a `variant` prop to `TopUpDialog` (`out-of-credits` default | `add-credits`) so the wallet entry point gets accurate copy
- Reuse `SealCheckIcon` / `CircleIcon` from the copilot task-progress treatment for row status
- `Task` / `TaskGroup` interfaces moved from `Wallet.tsx` to `helpers.ts` (only importer, `WalletTaskGroups.tsx`, updated)

No backend changes. No API changes. Reward amounts are unchanged and still mirror `backend/data/onboarding.py`.

#### For code changes:
- [x] I have made a test plan
  - [x] `pnpm format`, `pnpm lint`, `pnpm types` clean
  - [x] `pnpm test:unit` — 379 files pass
  - [x] New `WalletCompactPanel.test.tsx`: row flattening (completed group collapses, in-progress group lists tasks, no task dropped), panel render, accordion collapse, Add credits callback
  - [ ] **New layout** (`AUTOGPT_NEW_LAYOUT` on): open the wallet from the sidebar footer — compact panel opens up and to the right, fits the viewport, long task names wrap
  - [ ] Click **Add credits** — popover closes, dialog opens titled "Add automation credits", Stripe checkout still reachable
  - [ ] Toggle **Earn credits** — list collapses and expands
  - [ ] With `ENABLE_PLATFORM_PAYMENT` off — Add credits row is hidden, rest of the panel still renders
  - [ ] **Classic navbar** (`AUTOGPT_NEW_LAYOUT` off): wallet is visually unchanged — inline top-up, auto-refill, task videos, progress bars, confetti on step completion
  - [ ] Hit zero credits — low-credit banner and daily auto-opener still say "You're out of automation credits"


autogpt_platform/frontend/src/components/layout/Navbar/components/Wallet/components/WalletFullPanel.tsx
```diff
@@ -0,0 +1,54 @@
+"use client";
+
+import { ScrollArea } from "@/components/__legacy__/ui/scroll-area";
+import { Flag, useGetFlag } from "@/services/feature-flags/use-get-flag";
+import { PopoverClose } from "@radix-ui/react-popover";
+import { X } from "lucide-react";
```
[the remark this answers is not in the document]
Abhi1992002: Fixed in 439d7e7: swapped `X` from `lucide-react` for `XIcon` from `@phosphor-icons/react`. Also took the related nitpick on line 3 in the same commit — `ScrollArea` now comes from `@/components/ui/scroll-area` instead of `@/components/__legacy__/ui/scroll-area`. The non-legacy one is a superset of the legacy component (same props plus an optional `showScrollToTop`), so it's a drop-in swap. Both were carried over verbatim when the classic panel moved out of `Wallet.tsx`.

autogpt_platform/frontend/src/components/layout/Navbar/components/Wallet/useWallet.ts
```diff
@@ -0,0 +1,152 @@
...
+import { getTaskGroups } from "./helpers";
+
+export function useWallet() {
+  const { state, updateState } = useOnboarding();
+  const api = useBackendAPI();
+  const { credits, formatCredits, fetchCredits } = useCredits({
+    fetchInitialCredits: true,
+  });
+
+  const groups = useMemo(() => getTaskGroups(state), [state]);
```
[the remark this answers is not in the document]
Abhi1992002: Fixed in 3ed45bf. Good catch — `handleNotification` closes over `groups`, which is derived from `state`, so every onboarding update was detaching the message listener and calling `connectWebSocket()` again.

The handler now goes through a ref, so the subscription effect only depends on `api` and stays mounted for the lifetime of the hook:

```ts
const notificationRef = useRef(handleNotification);

useEffect(() => {
  notificationRef.current = handleNotification;
}, [handleNotification]);

useEffect(() => {
  const detachMessage = api.onWebSocketMessage(
    "notification",
    (notification) => notificationRef.current(notification),
  );
  api.connectWebSocket();
  return () => detachMessage();
}, [api]);
```

The listener still sees the latest `groups`/`fetchCredits` via the ref, but there's exactly one attach/detach pair and one `connectWebSocket()` call per mount now.

Also took the `completedCount` nitpick in the same commit — it's a `useMemo` over `groups`/`state` now, matching `totalCount` right above it, so it no longer flashes `null` on the first render and gate the reward dot in `Wallet.tsx`.

Abhi1992002:
Addressed the review round in 439d7e7, 3ed45bf and 9f1a566.

**Inline threads**
- `WalletFullPanel`: `lucide-react`'s `X` → `XIcon` from `@phosphor-icons/react`, and `ScrollArea` now comes from `@/components/ui/scroll-area` instead of `@/components/__legacy__/ui/scroll-area` (the non-legacy component is a superset — same props plus an optional `showScrollToTop` — so it's a drop-in swap). Both were carried over verbatim when the classic panel moved out of `Wallet.tsx`.
- `useWallet`: the websocket subscription is stable now. `handleNotification` closes over `groups` (derived from onboarding `state`), so every state update was detaching the listener and calling `connectWebSocket()` again. The handler goes through a ref, so the subscription effect only depends on `api`.

**Nitpicks**
- `completedCount` is a `useMemo` over `groups`/`state`, matching `totalCount` right above it — no transient `null` gating the reward dot on first render.
- Added `TopUpDialog.test.tsx` covering both variants' title and body copy, the billing link, the closed state, and `onClose`.

**Coverage**
The patch check was failing at 47.67%. Added `Wallet.test.tsx` — it renders the real `<Wallet />` in both classic and compact mode and drives `useWallet` end to end: gated render before credits/onboarding load, reward dot and tooltip, opening each panel, `walletShown` being set exactly once, the Add credits → dialog handoff, the balance flash, and the onboarding notification path (confetti for tracked steps, credits-only for untracked ones, ignored non-step events, detach on unmount). One of those tests is a direct regression test for the websocket fix — it asserts a single `onWebSocketMessage`/`connectWebSocket` pair survives an onboarding state change.

Local coverage on the touched files:

| File | Lines |
|---|---|
| `Wallet.tsx` | 100% |
| `useWallet.ts` | 91% |
| `helpers.ts` | 100% |
| `WalletCompactPanel.tsx` | 100% |
| `WalletFullPanel.tsx` | 100% |
| `TopUpDialog.tsx` | 100% |

38 tests pass across the wallet and top-up suites; `prettier`, `eslint` and `tsc` are clean.

autogpt_platform/frontend/src/components/layout/Navbar/components/Wallet/Wallet.tsx
```diff
@@ -5,356 +5,132 @@ import {
...
-          {/* Tasks */}
-          <p className="mx-1 my-3 font-sans text-xs font-normal text-zinc-400">
-            Complete the following tasks to earn more credits!
-          </p>
-          <TaskGroups groups={groups} />
-        </ScrollArea>
-      </PopoverContent>
-    </Popover>
+        </PopoverTrigger>
+        <PopoverContent
```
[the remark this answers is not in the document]
Abhi1992002: Fixed in a0e57dc. Real gap, though it needs a backend contract violation to trigger: `completedSteps` is non-optional in both the generated `UserOnboarding` model and `UserOnboarding` in `autogpt-server-api/types.ts`, so `fromBackendUserOnboarding` can only spread through an `undefined` if the API omits a required field.

That said, the rest of the wallet already hedges — `useWallet` does `state.completedSteps?.includes(...)` and `WalletTaskGroups` does the same — so the compact path was the odd one out. `getEarnRows` now takes `OnboardingStep[] | undefined` and normalizes once:

```ts
const claimed = completedSteps ?? [];
```

A thin response degrades to "nothing claimed" instead of throwing. `WalletCompactPanel`'s prop type widened to match, and there's a regression test in `WalletCompactPanel.test.tsx` (`treats a missing completedSteps list as nothing claimed") asserting every row comes back undone rather than blowing up.

autogpt_platform/frontend/src/components/layout/Navbar/components/Wallet/components/WalletFullPanel.tsx
```diff
@@ -0,0 +1,54 @@
...
+          <span className="font-sans text-xs text-zinc-500">
+            Platform-only credits for automations. This is separate from your
+            subscription and is not usable for plan fees.
+          </span>
+        </div>
+        <div className="flex shrink-0 items-center text-sm text-violet-700">
+          <div className="rounded-lg bg-violet-100 px-3 py-2">
+            Earn credits{" "}
+            <span className="font-semibold">{formattedCredits}</span>
+          </div>
```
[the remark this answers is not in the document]
Abhi1992002: Not taking this one — deliberately out of scope.

That badge isn't new. `WalletFullPanel.tsx` is the existing classic-navbar panel moved verbatim out of `Wallet.tsx`; the "Earn credits {balance}" header has shipped that way for a while. The whole point of this PR is that the *compact* panel is the redesign and the classic navbar renders byte-for-byte identically, so the flag can be flipped without touching the old layout. Changing the classic panel's copy here would break that guarantee and put a UI change in a PR whose test plan asserts the opposite.

The observation is fair on its merits — the compact panel does use the cleaner `Automation credits` / `$9.79` label-value pattern, which is exactly why it reads better. Rewording the classic panel belongs in a follow-up against the old layout, if we bother at all given it's on its way out.

autogpt_platform/frontend/src/components/layout/TopUpPrompt/__tests__/TopUpDialog.test.tsx
```diff
@@ -0,0 +1,65 @@
...
+    expect(screen.queryByText(/automation credits/i)).toBeNull();
+  });
+
+  it("calls onClose when the dialog is dismissed", async () => {
+    const onClose = vi.fn();
+    render(<TopUpDialog isOpen onClose={onClose} />);
+
+    await userEvent.keyboard("{Escape}");
+
+    expect(onClose).toHaveBeenCalled();
```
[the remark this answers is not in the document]
Abhi1992002: Fixed in 10793bf — and you were right that the loose assertion was hiding something. I'd originally written `toHaveBeenCalledTimes(1)`, saw it fail, and weakened the assertion instead of asking why. It fires **twice**.

Root cause is in the shared Dialog, not here. `DialogWrap.tsx:97` passes `onEscapeKeyDown={isForceOpen ? undefined : handleClose}`, and since it doesn't call `preventDefault()`, Radix's dismissable layer then *also* fires `onOpenChange(false)` on `RXDialog.Root` — which `Dialog.tsx:59-66` routes into `config.handleClose()` as well. Both paths land on `controlled.set(false)`, so any controlled dialog in the app gets a duplicate close notification on Escape. Closing is idempotent so nothing visibly breaks, which is presumably why it's gone unnoticed.

Split into two tests rather than papering over it:

- `calls onClose once when the close button is used` → `toHaveBeenCalledTimes(1)`. Single code path, and it's the more common user action.
- `notifies close on Escape (twice, via the shared Dialog)` → `toHaveBeenCalledTimes(2)`, with a comment explaining why. Pinning the real number means this test breaks — deliberately — if anyone fixes `DialogWrap`.

Not fixing the Dialog itself in this PR: it's a design-system component used by every dialog in the app, and this PR is scoped to the wallet. Worth its own change.

autogpt_platform/frontend/src/components/layout/Navbar/components/Wallet/useWallet.ts
```diff
@@ -0,0 +1,159 @@
...
+} from "@/lib/autogpt-server-api";
+import { useBackendAPI } from "@/lib/autogpt-server-api/context";
+import { useOnboarding } from "@/providers/onboarding/onboarding-provider";
+import confetti, { type Options as ConfettiOptions } from "canvas-confetti";
+import { useCallback, useEffect, useMemo, useRef, useState } from "react";
+import { getTaskGroups } from "./helpers";
+
+export function useWallet() {
+  const { state, updateState } = useOnboarding();
+  const api = useBackendAPI();
```
0ubbe, line 17: Would ideally be better to move this to the generated hooks if possible and low effort 🙏🏽

autogpt_platform/frontend/src/components/layout/Navbar/components/Wallet/useWallet.ts
```diff
@@ -0,0 +1,159 @@
...
+    );
+  }, [groups, state]);
+
+  const onWalletOpen = useCallback(async () => {
+    if (!state?.walletShown) {
+      updateState({ walletShown: true });
+    }
+    // Refresh credits when the wallet is opened
+    fetchCredits();
+  }, [state?.walletShown, updateState, fetchCredits]);
```
0ubbe: Is this `useCallback` really needed? This could just be a function declaration 👍🏽 . Same for the other instances on this file.

autogpt_platform/frontend/src/components/layout/Navbar/components/Wallet/Wallet.tsx
```diff
@@ -5,356 +5,132 @@ import {
...
+    walletRef,
+    completedCount,
+    totalCount,
+  } = useWallet();
+  const [topUpOpen, setTopUpOpen] = useState(false);
+
+  function handleAddCredits() {
+    setWalletOpen(false);
+    setTopUpOpen(true);
+  }
```
0ubbe: Could be moved to the hook 💅🏽

autogpt_platform/frontend/src/services/feature-flags/use-get-flag.ts
```diff
@@ -51,14 +51,14 @@ const defaultFlags = {
...
-  [Flag.AGENT_BRIEFING]: true,
+  [Flag.AGENT_BRIEFING]: false,
   [Flag.GENERIC_TRIGGER_AGENTS]: false,
   [Flag.CHAT_SEARCH]: false,
   [Flag.CHAT_SHARING]: false,
   [Flag.AUTOGPT_NEW_LAYOUT]: false,
   [Flag.CHAT_WORKSPACE_FILES]: false,
   [Flag.CHAT_PINNING]: false,
-  [Flag.TASK_PROGRESS_BAR]: false,
+  [Flag.TASK_PROGRESS_BAR]: true,
```
0ubbe: All the defaults should be false 🙏🏽 . Otherwise if LD fails to load we risk showing WIP work to customers.

0ubbe (dismissed):
Nice 🏁 , a couple of nits ☕️

autogpt_platform/frontend/src/services/feature-flags/use-get-flag.ts
Abhi1992002: Reverted in b6181c4 — good catch, and it was worse than a bad default: that whole hunk was accidental. `AGENT_BRIEFING true -> false` and `TASK_PROGRESS_BAR false -> true` were local toggles I used while checking the copilot task-progress treatment (the compact panel borrows `SealCheckIcon` from it) and they rode along in a0e57dc. Neither flag is read by the wallet — `TASK_PROGRESS_BAR` is only consumed in `copilot/components/{ContextPanel,ChatMessagesContainer,ChatContainer}` and `AGENT_BRIEFING` in `library` + `EmptySession`.

The file is now byte-identical to `dev`, so this PR touches no flag defaults at all.

Worth flagging separately: `AGENT_BRIEFING` is `true` on `dev` today, which is the exact fail-open case you're describing. I've left it as-is rather than flip someone else's flag from a wallet PR, but it probably wants a one-liner of its own.

autogpt_platform/frontend/src/components/layout/Navbar/components/Wallet/useWallet.ts
Abhi1992002: Done in 5d37c74. All of it: `onWalletOpen`, `handleNotification` and the two new top-up handlers are function declarations now, and `groups` / `totalCount` / `completedCount` are plain derived values instead of `useMemo`.

One thing worth calling out since it isn't a pure deletion — `handleNotification` was memoized for a reason, but not a good one. The websocket effect used to depend on it:

```ts
useEffect(() => { ... }, [api, handleNotification]);
```

so every onboarding state change tore down and re-subscribed (that was the bug fixed in 3ed45bf). The fix routes the listener through a ref, and the ref refresh is now dep-less:

```ts
useEffect(() => {
  notificationRef.current = handleNotification;
});
```

which is equivalent to the old `[handleNotification]` — a fresh closure every render meant it ran every render anyway. The subscription effect stays keyed on `[api]` alone, so it still mounts once for the lifetime of the hook. `Wallet.test.tsx` covers this (`keeps the websocket subscription mounted across onboarding updates`).

autogpt_platform/frontend/src/components/layout/Navbar/components/Wallet/Wallet.tsx
Abhi1992002: Moved in 5d37c74. `topUpOpen` plus `onAddCredits` / `onTopUpClose` now live in `useWallet`, so `Wallet.tsx` is pure render — it destructures and wires, no local state and no inline arrow in the `onClose` prop.

autogpt_platform/frontend/src/components/layout/Navbar/components/Wallet/useWallet.ts
Abhi1992002: Not moving this one — there's nothing to move it to. `useBackendAPI()` is used here purely for the websocket:

```ts
api.onWebSocketMessage("notification", ...);
api.connectWebSocket();
```

That's the onboarding `step_completed` push that drives the confetti and the credits refresh. Orval generates from the OpenAPI spec, which only describes REST endpoints — the websocket transport isn't in it, so there is no generated equivalent to swap in. This is also pre-existing wiring, moved verbatim out of `Wallet.tsx`; the PR didn't introduce a new `BackendAPI` usage.

The other legacy dependency here is `useCredits` (line 18), which wraps `AutoGPTServerAPI` for `getUserCredit`. That one *does* have a generated counterpart, but it's a shared hook used by the billing/settings surfaces too, so porting it is its own PR rather than a low-effort change inside this one.

PR #13736, opened 2026-07-29
feat(blocks): add OAuth2 support to Slack block

omsharma0401:
### Why / What / How

**Why**: `SendSlackMessageBlock` (#13008) requires manually copying a bot token (`xoxb-...`) from Slack's app dashboard. In review, @ntindle asked for "proper oauth support for slack sign in" as the follow-up PR.

**What**: Adds a Slack OAuth connection as an alternative to the manual bot token — users can now go through Slack's "Add to Slack" consent flow instead of copy-pasting a token. Both credential types keep working side by side.

**How**: Uses Slack's bot-install OAuth v2 flow (`oauth/v2/authorize` + `oauth.v2.access`), not "Sign in with Slack" (OpenID Connect) — the OIDC flow only returns an identity token and can't call `chat.postMessage`. Follows the `ProviderBuilder`/`_config.py`/`_oauth.py` SDK pattern used by Linear/WordPress (`.with_oauth()` self-registers via `AutoRegistry`, no changes needed to `integrations/oauth/__init__.py`).

### Changes 🏗️

- `backend/blocks/slack/_oauth.py` — `SlackOAuthHandler`. Guards against Slack's HTTP 200 + `ok: false` failure mode on every request. Supports both classic non-expiring bot tokens and the newer opt-in "token rotation" tokens (refresh is a no-op when there's no `refresh_token`). Stores `team_id`/`team_name`/`bot_user_id` in credential metadata for multi-workspace disambiguation. No PKCE — Slack's OAuth v2 endpoint doesn't support it.
- `backend/blocks/slack/_config.py` — `ProviderBuilder` wiring for the `slack` provider, requesting `chat:write`, `chat:write.public`, `chat:write.customize` bot scopes (the scopes the block actually exercises). Deliberately uses `.with_supported_auth_types("api_key")` instead of `.with_api_key(env_var_name=...)` — the latter auto-creates a shared default credential from an org-wide env var, which would expose one workspace's bot token to every user of the platform. Each user still pastes their own bot token if they don't use OAuth.
- `backend/blocks/slack/_api.py`, `blocks.py` — accept either `APIKeyCredentials` or `OAuth2Credentials` via `credentials.auth_header()`. Folded the old `_auth.py` into `_config.py` to match the Linear/WordPress SDK layout.
- `backend/blocks/slack/slack_oauth_test.py` — unit tests for the OAuth handler: login URL construction (incl. default scopes, no PKCE leak), token exchange (classic + rotating tokens, and the `ok: false`-on-200 failure case), refresh (no-op vs. rotating), revoke (success/failure).
- `.env.default` — `SLACK_CLIENT_ID`/`SLACK_CLIENT_SECRET` for local setup, matching the Linear/Discord/Todoist entries already there.
- `docs/integrations/block-integrations/slack/blocks.md` — regenerated; updated the `how_it_works`/`file_description` manual sections to mention the OAuth option.

No frontend changes needed — `CredentialsType` and the provider list are generic/dynamic, not hardcoded per-provider (confirmed `slack` was already wired in from #13008).

#### For code changes:
- [x] I have made a test plan
  - [x] `poetry run format` / `poetry run lint` pass clean (ruff, isort, black, pyright)
  - [x] `pytest backend/blocks/slack/ --collect-only` — all 34 tests (12 new OAuth handler tests + 22 existing) collect without error
  - [x] Ran every OAuth handler code path (login URL, token exchange incl. classic + rotating tokens + the `ok:false`-on-HTTP-200 failure mode, refresh no-op vs. rotating refresh, revoke success/failure) and `_api.py`'s dual-credential-type `auth_header()` resolution directly against the real implementation via a standalone script — 26/26 checks pass, including `SendSlackMessageBlock`'s full `execute_block_test` harness
  - [x] Regenerated block docs and reviewed the diff
  - [ ] Full DB-backed `poetry run test` suite — blocked locally by an unrelated Postgres role-permission issue in this dev environment's `docker-compose.test.yaml` (the `postgres` role isn't a true superuser in the Supabase-flavored test image here); requesting CI run this on the PR

- [x] `.env.default` is updated (`SLACK_CLIENT_ID`/`SLACK_CLIENT_SECRET`)
- [x] `docker-compose.yml` already compatible — no new services needed
- [x] Configuration changes listed above under **Changes**


omsharma0401:
Closing — reopening manually.

PR #13749, opened 2026-07-31
perf(backend): compile each JSON schema validator once in validate_with_jsonschema

dexhunter:
### Why / What / How

**Why.** `validate_with_jsonschema` in `backend/util/json.py` calls the
module-level `jsonschema.validate()`. That convenience wrapper does two things on
**every single call** that depend only on the schema: it re-resolves the
validator class, and it runs `cls.check_schema(schema)`, which validates the
schema itself against the full Draft-2020-12 meta-schema. Then it throws the
validator away. This function sees the same handful of schemas over and over.
`BlockSchema.jsonschema()` already caches the schema dict per class, so a busy
executor re-checks the same schemas against the meta-schema thousands of times.

Measured on the **44 unique input schemas across 46 occurrences in the 18 agent JSON files this repository currently ships** in
`autogpt_platform/backend/agents/*.json`, which are the exact objects
`AgentExecutorBlock.Input.get_mismatch_error` passes to this function. Each call
costs **1.46 ms on average**, ranging from 0.40 ms to 2.73 ms. The cost barely
tracks schema size, because the meta-schema check dominates: even the cheapest of
these schemas costs 0.40 ms.

It is on the per-node-execution path four ways:

| call site | frequency |
|---|---|
| `blocks/_base.py:852` / `:861`, `input_schema.validate_data(...)` in `Block.execute` | once per **block execution** |
| `executor/utils.py:372`, `schema.get_mismatch_error(data)` in `validate_exec`, reached from `executor/manager.py:187` | once per **node execution**, before it runs |
| `executor/manager.py:479` | once per **outgoing link traversal** when enqueuing each downstream node |
| `executor/manager.py:526` | once per waiting **incomplete static-link execution** re-checked |

A graph run with N node executions and E link traversals makes roughly `2N + E`
of these calls. `validate_exec` is also a synchronous `def` called from async
executor paths, so the cost lands on executor throughput as well as on
per-execution latency. Two further callers sit outside the executor:
`blocks/agent.py:65` (sub-agent input schemas) and `blocks/mcp/block.py:129`
(MCP tool schemas).

**What.** Compile the validator once per distinct schema and reuse it. Behaviour
does not change, including the awkward parts covered below. **Honest denominator:** the optimization applies to repeated validation of admitted schemas. First-seen, unsafe, and oversized schemas still use the uncached validation path and may pay key inspection and copying overhead. The saving is on every subsequent call with that schema, which is the
normal case here rather than the exception.

**How.** A small `_compiled_validator` helper resolves the validator class, runs
`check_schema` once, and caches the resulting validator. Three details worth
reviewing:

1. **The cache key is the schema's serialized content, not its identity.** A
   graph's input schema is rebuilt as a fresh dict on every execution, so
   identity keying would never hit, and it would pin every schema object the
   process has ever seen alive forever. Key order stays as the caller wrote it,
   because the error strings this function returns embed a `repr` of the failing
   sub-schema, and reordering keys would change them.
2. **Malformed schemas are deliberately not cached.**
   This is not hypothetical. `agent_00fdd42c-…json`'s `output_schema` in this
   repository has duplicate entries in `required` and genuinely fails the
   meta-schema check. `jsonschema.validate()` raises `SchemaError` (not
   `ValidationError`) there, and it escapes this function uncaught. Each call
   runs `check_schema` again and raises a fresh exception, so the cache cannot
   retain a `SchemaError` with a traceback that grows across repeated calls.
3. **The cached validator compiles against a private deep copy.** A `jsonschema`
   validator memoises the sub-schemas it walks, so a retained one must not alias
   a dict the caller still holds and might mutate afterwards.

The cache is a true LRU holding at most 2,048 entries. It admits finite JSON-native schemas plus stable string-backed enums, fingerprints enum types to keep keys injective, rejects repeated container identities, and retains serialized keys no larger than 32 KiB; all 557 currently loadable built-in input schemas fit this bound.

One subtlety that is easy to get wrong: `jsonschema.validate()` raises
`best_match(validator.iter_errors(data))`, whereas `Validator.validate()` raises
the **first** error from `iter_errors`. On a schema with two violated properties
those pick different errors and produce different strings, so this change calls
`best_match` explicitly to keep the returned message identical.

Supplementary autoresearch, as exploratory evidence only: [public dashboard](https://dashboard.weco.ai/share/N7mK0DxJIR-SVWxNqtNux6BJKMmyKA1J). The shipped hardening follows the safe step-5/6 lineage rather than the dashboard's visible best step.

### Changes 🏗️

- `backend/util/json.py`: adds `_compiled_validator`, a bounded content-keyed
  cache of compiled `jsonschema` validators, and routes
  `validate_with_jsonschema` through it instead of the module-level
  `jsonschema.validate()`.
- `backend/util/test_json.py`: adds focused cache coverage; the focused file now passes 58 tests,
  which previously had **no test coverage at all**. They cover valid and invalid
  verdicts, repeated calls with one schema, distinct schemas not being
  conflated, equal-but-distinct schema objects, `SchemaError` still escaping on
  every call, and in-place schema mutation not going stale.

### Results

Same 44 unique current input schemas, 20 iterations per schema, seven alternating CPU-serialized pairs:

| | median ms per call |
| --- | ---: |
| current `dev` | **1.6327** |
| this PR | **0.1578** |

Repeated validation is **10.35x faster** on this fixture. Base and candidate produced the same outcome digest in every pair. This benchmark measures the warmed repeated-validation path; it does not characterize first-seen or uncached schemas.

#### For code changes:
- [x] I have made a test plan
  - [x] `pytest backend/util/test_json.py -q`: `58 passed` on the repaired head
  - [x] `ruff check backend/util/json.py backend/util/test_json.py`: clean
  - [x] `black --check backend/util/json.py backend/util/test_json.py`: clean
  - [x] `isort --profile black --check-only backend/util/json.py backend/util/test_json.py`: clean
  - [x] Character-for-character comparison of the returned value against the
        unmodified implementation across the 44 unique current input schemas plus
        randomised schemas, including exception type and message where one
        escapes
  - [x] Before/after benchmark on the same fixtures, CPU-serialised

### DIFF
--- a/autogpt_platform/backend/backend/util/json.py
+++ b/autogpt_platform/backend/backend/util/json.py
@@ -1,5 +1,7 @@
+import itertools
 import logging
 import re
+from copy import deepcopy
 from typing import Any, Type, TypeVar, overload
 
 import jsonschema
@@ -130,6 +132,68 @@ def loads(
     return parsed
 
 
+_VALIDATOR_CACHE: dict[bytes, Any] = {}
+_VALIDATOR_CACHE_MAX_ENTRIES = 512
+
+
+def _compiled_validator(schema: dict[str, Any]):
+    """Return a validator for `schema`, compiled once per distinct schema.
+
+    `jsonschema.validate()` re-resolves the validator class and re-validates the
+    schema against its meta-schema on every single call; both depend only on the
+    schema, so both are cached here.
+
+    The key is the schema's serialized content, not its identity: a graph's
+    input schema is rebuilt as a fresh dict on every node execution, so identity
+    keying would never hit and would pin every schema object the process has
+    ever seen. Key order is deliberately *not* normalised -- the error strings
+    this function returns embed a repr of the failing sub-schema, so reordering
+    keys would change them.
+
+    A `SchemaError` outcome is cached too, so a malformed schema keeps raising
+    rather than being silently accepted on the second call.
+    """
+    try:
+        key = orjson.dumps(schema)
+    except TypeError:
+        # Not JSON-serializable, so it cannot be keyed; compile without caching.
+        validator_cls = jsonschema.validators.validator_for(schema)
+        validator_cls.check_schema(schema)
+        return validator_cls(schema)
+
+    cached = _VALIDATOR_CACHE.get(key)
+    if cached is not None:
+        if isinstance(cached, jsonschema.SchemaError):
+            raise cached
+        return cached
+
+    # Compile against a private copy: a jsonschema validator memoises the
+    # sub-schemas it walks, so a retained one must not alias a dict the caller
+    # still holds and could mutate afterwards. deepcopy also preserves key
+    # order, which the error messages depend on.
+    schema_copy = deepcopy(schema)
+    validator_cls = jsonschema.validators.validator_for(schema_copy)
+    try:
+        validator_cls.check_schema(schema_copy)
+    except jsonschema.SchemaError as e:
+        _remember(key, e)
+        raise
+    validator = validator_cls(schema_copy)
+    _remember(key, validator)
+    return validator
+
+
+def _remember(key: bytes, value: Any) -> None:
+    """Store `value`, evicting the oldest entry when the cache is full.
+
+    The keys are caller-supplied schemas, which on this platform come from
+    user-authored graphs, so the cache has to be bounded.
+    """
+    if len(_VALIDATOR_CACHE) >= _VALIDATOR_CACHE_MAX_ENTRIES:
+        _VALIDATOR_CACHE.pop(next(iter(_VALIDATOR_CACHE)))
+    _VALIDATOR_CACHE[key] = value
+
+
 def validate_with_jsonschema(
     schema: dict[str, Any], data: dict[str, Any]
 ) -> str | None:
@@ -137,11 +201,19 @@ def validate_with_jsonschema(
     Validate the data against the schema.
     Returns the validation error message if the data does not match the schema.
     """
-    try:
-        jsonschema.validate(data, schema)
+    errors = _compiled_validator(schema).iter_errors(data)
+
+    # Valid data is the common case, and it does not need best_match at all.
+    first_error = next(errors, None)
+    if first_error is None:
         return None
-    except jsonschema.ValidationError as e:
-        return str(e)
+
+    # `jsonschema.validate()` raises best_match(iter_errors(...)), which is not
+    # the same error Validator.validate() raises, so match it exactly. chain the
+    # already-consumed first error back in rather than re-running iter_errors.
+    return str(
+        jsonschema.exceptions.best_match(itertools.chain((first_error,), errors))
+    )
 
 
 def sanitize_string(value: str) -> str:
--- a/autogpt_platform/backend/backend/util/json.py
+++ b/autogpt_platform/backend/backend/util/json.py
@@ -154,8 +154,6 @@ def _compiled_validator(schema: dict[str, Any]):
     this function returns embed a repr of the failing sub-schema, so reordering
     keys would change them.
 
-    A `SchemaError` outcome is cached too, so a malformed schema keeps raising
-    rather than being silently accepted on the second call.
     """
     try:
         key = orjson.dumps(schema)
@@ -168,8 +166,6 @@ def _compiled_validator(schema: dict[str, Any]):
     with _VALIDATOR_CACHE_LOCK:
         cached = _VALIDATOR_CACHE.get(key)
     if cached is not None:
-        if isinstance(cached, jsonschema.SchemaError):
-            raise cached
         return cached
 
     # Compile against a private copy: a jsonschema validator memoises the
@@ -178,20 +174,8 @@ def _compiled_validator(schema: dict[str, Any]):
     # order, which the error messages depend on.
     schema_copy = deepcopy(schema)
     validator_cls = jsonschema.validators.validator_for(schema_copy)
-    try:
-        validator_cls.check_schema(schema_copy)
-    except jsonschema.SchemaError as e:
-        cached = _remember(key, e)
-        if cached is e:
-            raise
-        if isinstance(cached, jsonschema.SchemaError):
-            raise cached
-        return cached
-
-    cached = _remember(key, validator_cls(schema_copy))
-    if isinstance(cached, jsonschema.SchemaError):
-        raise cached
-    return cached
+    validator_cls.check_schema(schema_copy)
+    return _remember(key, validator_cls(schema_copy))
 
 
 def _remember(key: bytes, value: Any) -> Any:
--- a/autogpt_platform/backend/backend/util/json.py
+++ b/autogpt_platform/backend/backend/util/json.py
@@ -5,6 +5,8 @@ from copy import deepcopy
 from typing import Any, Type, TypeVar, overload
 
 import jsonschema
+import jsonschema.exceptions
+import jsonschema.validators
 import orjson
 from fastapi.encoders import jsonable_encoder as to_dict
 from prisma import Json
--- a/autogpt_platform/backend/backend/util/json.py
+++ b/autogpt_platform/backend/backend/util/json.py
@@ -1,6 +1,8 @@
 import itertools
 import logging
+import math
 import re
+from collections import OrderedDict
 from copy import deepcopy
 from threading import Lock
 from typing import Any, Type, TypeVar, overload
@@ -135,8 +137,11 @@ def loads(
     return parsed
 
 
-_VALIDATOR_CACHE: dict[bytes, Any] = {}
-_VALIDATOR_CACHE_MAX_ENTRIES = 512
+_VALIDATOR_CACHE: OrderedDict[bytes, Any] = OrderedDict()
+# The 565 built-in Block classes can each contribute a normal and dry-run
+# schema; 2,048 covers that working set while leaving room for user schemas.
+_VALIDATOR_CACHE_MAX_ENTRIES = 2048
+_VALIDATOR_CACHE_MAX_KEY_BYTES = 16 * 1024
 _VALIDATOR_CACHE_LOCK = Lock()
 
 
@@ -155,19 +160,57 @@ def _compiled_validator(schema: dict[str, Any]):
     keys would change them.
 
     """
-    try:
-        key = orjson.dumps(schema)
-    except TypeError:
-        # Not JSON-serializable, so it cannot be keyed; compile without caching.
-        validator_cls = jsonschema.validators.validator_for(schema)
-        validator_cls.check_schema(schema)
-        return validator_cls(schema)
+    key = _schema_cache_key(schema)
+    if key is None:
+        return _new_validator(schema)
 
     with _VALIDATOR_CACHE_LOCK:
         cached = _VALIDATOR_CACHE.get(key)
-    if cached is not None:
-        return cached
+        if cached is not None:
+            _VALIDATOR_CACHE.move_to_end(key)
+            return cached
 
+    return _remember(key, _new_validator(schema))
+
+
+def _schema_cache_key(schema: dict[str, Any]) -> bytes | None:
+    """Return an injective, size-bounded key for safely cacheable schemas."""
+    if not _is_json_native(schema):
+        return None
+    try:
+        key = orjson.dumps(schema)
+    except TypeError:
+        return None
+    if len(key) > _VALIDATOR_CACHE_MAX_KEY_BYTES:
+        return None
+    return key
+
+
+def _is_json_native(value: Any) -> bool:
+    """Check that orjson preserves every value's type and meaning."""
+    pending = [value]
+    while pending:
+        current = pending.pop()
+        current_type = type(current)
+        if current is None or current_type in (str, int, bool):
+            continue
+        if current_type is float:
+            if not math.isfinite(current):
+                return False
+            continue
+        if current_type is list:
+            pending.extend(current)
+            continue
+        if current_type is dict:
+            if any(type(key) is not str for key in current):
+                return False
+            pending.extend(current.values())
+            continue
+        return False
+    return True
+
+
+def _new_validator(schema: dict[str, Any]):
     # Compile against a private copy: a jsonschema validator memoises the
     # sub-schemas it walks, so a retained one must not alias a dict the caller
     # still holds and could mutate afterwards. deepcopy also preserves key
@@ -175,11 +218,11 @@ def _compiled_validator(schema: dict[str, Any]):
     schema_copy = deepcopy(schema)
     validator_cls = jsonschema.validators.validator_for(schema_copy)
     validator_cls.check_schema(schema_copy)
-    return _remember(key, validator_cls(schema_copy))
+    return validator_cls(schema_copy)
 
 
 def _remember(key: bytes, value: Any) -> Any:
-    """Publish and return one value, evicting an oldest entry if needed.
+    """Publish and return one value, evicting the least-recently used entry.
 
     The keys are caller-supplied schemas, which on this platform come from
     user-authored graphs, so the cache has to be bounded.
@@ -187,9 +230,10 @@ def _remember(key: bytes, value: Any) -> Any:
     with _VALIDATOR_CACHE_LOCK:
         cached = _VALIDATOR_CACHE.get(key)
         if cached is not None:
+            _VALIDATOR_CACHE.move_to_end(key)
             return cached
         if len(_VALIDATOR_CACHE) >= _VALIDATOR_CACHE_MAX_ENTRIES:
-            _VALIDATOR_CACHE.pop(next(iter(_VALIDATOR_CACHE)))
+            _VALIDATOR_CACHE.popitem(last=False)
         _VALIDATOR_CACHE[key] = value
         return value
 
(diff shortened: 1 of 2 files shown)


autogpt_platform/backend/backend/util/test_json.py
```diff
@@ -754,3 +756,81 @@ class OuterModel(BaseModel):
...
+
+
+class TestValidateWithJsonschema:
+    """Test cases for validate_with_jsonschema."""
+
+    SCHEMA: dict[str, Any] = {
+        "type": "object",
+        "properties": {"name": {"type": "string"}, "age": {"type": "integer"}},
+        "required": ["name"],
+    }
```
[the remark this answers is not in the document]
dexhunter: Thanks for flagging this. CodeRabbit ran Ruff 0.16.0, while the backend lock pins Ruff 0.15.0 and the repository Ruff configuration does not select RUF012. The repo-pinned Ruff check and the live lint job both pass at this exact head, so this is optional test-only annotation cleanup rather than a required lint fix; no change is needed for this PR.

autogpt_platform/backend/backend/util/json.py
```diff
@@ -130,18 +135,85 @@ def loads(
...
+    """Publish and return one value, evicting an oldest entry if needed.
+
+    The keys are caller-supplied schemas, which on this platform come from
+    user-authored graphs, so the cache has to be bounded.
+    """
+    with _VALIDATOR_CACHE_LOCK:
+        cached = _VALIDATOR_CACHE.get(key)
+        if cached is not None:
+            return cached
+        if len(_VALIDATOR_CACHE) >= _VALIDATOR_CACHE_MAX_ENTRIES:
```
kcze, line 252: 🤖 🟠 **Should Fix**: FIFO eviction at a 512-entry cap is the pessimal policy for this codebase's actual working set, and can leave the cache strictly **slower** than the code it replaces.

There are ~500 `Block` classes, each contributing one entry for its Input schema plus a second distinct entry for the dry-run variant built at `blocks/_base.py:239` (`{**schema, "properties": …, "required": …}` with credentials excluded) — so ~1000 entries before any user-supplied AgentExecutorBlock or MCP tool schema. A long-lived executor cycling through more than 512 distinct schemas evicts each entry just before it's next needed (FIFO is worst-case for round-robin access), so every call misses and pays `orjson.dumps` + `deepcopy` + `check_schema` + insert/evict — more work than dev's plain `jsonschema.validate`.

Move-to-end on hit (LRU), or a cap above the block count, avoids the cliff. _(flagged by: Claude)_

autogpt_platform/backend/backend/util/json.py
```diff
@@ -130,18 +135,85 @@ def loads(
...
+    The key is the schema's serialized content, not its identity: a graph's
+    input schema is rebuilt as a fresh dict on every node execution, so identity
+    keying would never hit and would pin every schema object the process has
+    ever seen. Key order is deliberately *not* normalised -- the error strings
+    this function returns embed a repr of the failing sub-schema, so reordering
+    keys would change them.
+
+    """
+    try:
+        key = orjson.dumps(schema)
```
kcze: 🤖 🟡 **Nice to Have**: `orjson.dumps` isn't injective over the values a schema dict can hold, so logically different schemas can share one key and the first-compiled validator is silently applied to the rest. The docstring reasons carefully about key *order* fidelity but not value *type* fidelity.

Reproduced against the locked jsonschema: once `{"type":"number","maximum":float("inf")}` is cached, `{"type":"number","maximum":float("-inf")}` returns `None` (accepts 5) where dev returns a ValidationError; and `{"type":"number","maximum":None}` returns `None` instead of raising `SchemaError`, because the cache hit skips `check_schema` entirely — orjson renders `inf`/`-inf`/`nan`/`None` all as `null`. The same collapse applies to tuple↔list, Enum↔its value, and date/datetime↔its ISO string.

Latent rather than live: no current block schema carries a non-finite bound and jsonb rejects Infinity/NaN, so a third-party MCP tool schema is the plausible ingress. Including value types in the key, or refusing to cache when a round-trip isn't lossless, closes it. _(flagged by: Claude + Codex)_

autogpt_platform/backend/backend/util/json.py
```diff
@@ -130,18 +135,85 @@ def loads(
     return parsed
 
 
+_VALIDATOR_CACHE: dict[bytes, Any] = {}
```
kcze: 🤖 🟡 **Nice to Have**: the cache is bounded in entry *count* but not entry *size*, and each entry retains the serialized key bytes plus a `deepcopy` of the schema for the process lifetime with no idle eviction — dev retained nothing once the call returned. A user running graphs with large AgentExecutorBlock or MCP tool schemas pins 512 × schema-size in every executor process indefinitely. A total-byte ceiling, or simply skipping the cache for oversized schemas, bounds it. _(flagged by: Claude + Codex)_

autogpt_platform/backend/backend/util/test_json.py
```diff
@@ -754,3 +759,143 @@ class OuterModel(BaseModel):
...
+
+    def test_valid_data_returns_none(self):
+        """Valid data produces no error message."""
+        assert (
+            validate_with_jsonschema(self.SCHEMA, {"name": "John", "age": 30}) is None
+        )
+
+    def test_type_mismatch_returns_error_message(self):
```
kcze, line 783: 🤖 🟡 **Nice to Have**: the tests never assert the reimplemented error selection yields the *same string* as `jsonschema.validate` — they only check substring membership (`"is not of type 'string'" in error`) on two flat single-error schemas.

The risky part is `best_match(chain((first_error,), errors))` standing in for `validate()`'s `best_match(iter_errors(...))`, which only diverges observably on multi-error schemas where `best_match` descends into `error.context` (`anyOf`/`oneOf`/`if-then`, `unevaluatedProperties`). A fuzz over 13 schemas × 23 instances found 0 mismatches, so this is a coverage gap rather than a live defect — but a future regression there would pass the suite. One `anyOf` case asserting exact equality against `jsonschema.validate`'s message would lock it in. _(flagged by: Claude)_

autogpt_platform/backend/backend/util/json.py
dexhunter: *This is Aiden on behalf of Dex*

Addressed in `693891c61`. The cache is now a true `OrderedDict` LRU and refreshes ordinary hits and concurrent publication-race hits under the same lock. Capacity is 2,048; the repository has 565 static Block classes, giving a conservative normal-plus-dry-run working set of 1,130. A two-pass 1,130-schema probe reused every validator, and full-cache coverage confirms a refreshed hot entry survives eviction. An independent locked replay measured the repaired path at 10.35x faster than current `dev`.

autogpt_platform/backend/backend/util/json.py
dexhunter: *This is Aiden on behalf of Dex*

Addressed in `693891c61`. Cache admission now accepts only exact JSON-native containers and scalar types with finite floats. Non-finite values, tuples, enums, datetime-like values, subclasses, and other potentially lossy forms compile uncached. Tests cover the reproduced `inf`/`-inf`/`None` and tuple/list collision classes while comparing outcomes with `jsonschema.validate`.

autogpt_platform/backend/backend/util/json.py
dexhunter: *This is Aiden on behalf of Dex*

Addressed in `693891c61`. Serialized schema keys above 16 KiB now compile uncached, while the 2,048-entry ceiling bounds the admitted working set. Focused and adversarial tests confirm an oversized schema remains correct across repeated calls and leaves no retained cache entry.

autogpt_platform/backend/backend/util/test_json.py
dexhunter: *This is Aiden on behalf of Dex*

Addressed in `693891c61`. The tests now compare the complete returned string against `jsonschema.validate` for `anyOf`, `oneOf`, and nested-property combinator failures. All exact-parity cases pass, and the implementation continues to run `best_match` over the complete error iterator.

autogpt_platform/backend/backend/util/json.py
```diff
@@ -130,18 +137,127 @@ def loads(
...
+
+def _new_validator(schema: dict[str, Any]):
+    # Compile against a private copy: a jsonschema validator memoises the
+    # sub-schemas it walks, so a retained one must not alias a dict the caller
+    # still holds and could mutate afterwards. deepcopy also preserves key
+    # order, which the error messages depend on.
+    schema_copy = deepcopy(schema)
+    validator_cls = jsonschema.validators.validator_for(schema_copy)
+    validator_cls.check_schema(schema_copy)
+    return validator_cls(schema_copy)
```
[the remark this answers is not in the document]
dexhunter: *This is Aiden on behalf of Dex*

Thanks for flagging the shared-instance boundary. The executor does run graph executions in a `ThreadPoolExecutor`, so I checked the locked `jsonschema==4.26.0` implementation rather than assuming serial access.

`_new_validator` does not supply the deprecated `RefResolver`: the cached validator has `_ref_resolver is None` and uses the modern frozen `referencing.Resolver`. Each `$ref` descent evolves a request-local validator and resolver instead of mutating the cached instance. This matches the jsonschema maintainer's resolution of issue #223 that the new reference APIs removed the shared resolver state.

An independent exact-head stress reused the same cached `$ref` validator across 24 threads and 4,800 validations. Every result matched the serial baseline, with no exception or cache corruption. Given the pinned dependency and inspected execution path, no implementation change is needed here.

autogpt_platform/backend/backend/util/json.py
```diff
@@ -130,18 +137,127 @@ def loads(
...
+    return key
+
+
+def _is_json_native(value: Any) -> bool:
+    """Check that orjson preserves every value's type and meaning."""
+    pending = [value]
+    while pending:
+        current = pending.pop()
+        current_type = type(current)
+        if current is None or current_type in (str, int, bool):
```
kcze: 🤖 🟠 **Should Fix**: `type(x) is str` rejects `str` *subclasses*, and every credentialed block schema embeds `ProviderName` (`class ProviderName(str, Enum)`) under `properties.<field>.credentials_provider[*]`. That makes **271 of the 557 built-in input schemas permanently uncacheable** (264 of them solely because of this check). All 271 were cached at `56ad4f4b` — `orjson.dumps` serialises the enum fine — so this is an unintended side effect of the injectivity hardening.

The split is inverted against production: of the 417 credentialed blocks, only 146 production input schemas are admitted, but 410 of the credential-stripped **dry-run** variants built at `_base.py:235-250` are. The cache accelerates simulation and not real execution.

Measured at this head, `SendSlackMessageBlock.input_schema` costs 2.50 ms/call — identical to `dev` — versus 0.45 ms once the same schema is admitted.

All 343 enum occurrences sit under `credentials_provider`, which jsonschema ignores, so they are inert for validation. The only thing the strictness protects is the `repr` of the failing sub-schema in the error string (`[<ProviderName.TWITTER: 'twitter'>]` vs `['twitter']`), and only if a user schema ever serialises byte-identically to a built-in one. Admitting `isinstance(x, str)`, or normalising `str` subclasses before keying, recovers all 271 without reopening the injectivity issue. _(flagged by: Codex + Claude)_

autogpt_platform/backend/backend/util/json.py
```diff
@@ -130,18 +137,127 @@ def loads(
...
+
+def _schema_cache_key(schema: dict[str, Any]) -> bytes | None:
+    """Return an injective, size-bounded key for safely cacheable schemas."""
+    if not _is_json_native(schema):
+        return None
+    try:
+        key = orjson.dumps(schema)
+    except TypeError:
+        return None
+    if len(key) > _VALIDATOR_CACHE_MAX_KEY_BYTES:
```
kcze, line 186: 🤖 🟡 **Nice to Have**: the 16 KiB key budget excludes the seven largest built-in input schemas, which are also among the hottest — `OrchestratorBlock` 30.6 KiB, `AIStructuredResponseGeneratorBlock` 29.9 KiB, `AITextGeneratorBlock` 28.6 KiB, plus `AIListGeneratorBlock`, `AITextSummarizerBlock`, `AIConditionBlock`, `AIConversationBlock`. Each costs 8.0–9.4 ms per validation uncached, on every node execution.

They are currently masked by the `ProviderName` exclusion above (no input schema is excluded by size *alone* today), so this only becomes visible once that is fixed — worth sizing the budget against the built-in schemas rather than leaving the seven biggest permanently uncached. _(flagged by: Codex)_

autogpt_platform/backend/backend/util/json.py
```diff
@@ -130,18 +137,127 @@ def loads(
...
+        return None
+    return key
+
+
+def _is_json_native(value: Any) -> bool:
+    """Check that orjson preserves every value's type and meaning."""
+    pending = [value]
```
kcze, line 196: 🤖 🟡 **Nice to Have**: the admission walk has no seen-set and no depth budget, so a schema containing a Python reference cycle loops forever with `pending` growing without bound — wedging the worker thread. `dev` and `56ad4f4b` both degraded gracefully, because `orjson.dumps` raises `TypeError: Recursion limit reached` and falls through to the uncached compile.

```python
cyc = {"type": "object"}; cyc["properties"] = {"self": cyc}
orjson.dumps(cyc)       # TypeError -> uncached compile (before)
_is_json_native(cyc)    # hangs (now)
```

Reachability is low — all three call sites receive JSON-decoded or Pydantic-derived schemas, which cannot carry cycles — but a `seen: set[int]` on container ids restores the graceful path cheaply. _(flagged by: Claude)_

autogpt_platform/backend/backend/util/test_json.py
```diff
@@ -754,3 +760,293 @@ class OuterModel(BaseModel):
...
+            (
+                {"type": "object", "properties": {"value": {"enum": [(1, 2)]}}},
+                {"value": [1, 2]},
+                {"type": "object", "properties": {"value": {"enum": [[1, 2]]}}},
+                {"value": [1, 2]},
+                1,
+            ),
+        ],
+    )
+    def test_lossy_serializations_bypass_cache(
```
kcze, line 975: 🤖 🟡 **Nice to Have**: the admission tests use synthetic literals only, so nothing asserts that a *real* block schema is admissible — which is why 48.7% of built-in input schemas silently falling out of the cache passes a green suite. One assertion pinning a credentialed block would have caught it:

```python
assert json_util._schema_cache_key(SendSlackMessageBlock().input_schema.jsonschema()) is not None
```
_(flagged by: Claude)_

autogpt_platform/backend/backend/util/json.py
dexhunter: Addressed in 125c1773f. Cache keys now admit stable string-backed enum values and include each enum’s traversal position and concrete type, so ProviderName is cacheable without colliding with the equivalent plain string. Added ProviderName-vs-plain-string injectivity coverage plus real Slack and Orchestrator schema tests. A current registry scan admits all 557/557 loadable built-in input schemas.

autogpt_platform/backend/backend/util/json.py
dexhunter: Addressed in 125c1773f. The bounded key budget is now 32 KiB: the largest current built-in input schema is OrchestratorBlock at 30,478 bytes, and all seven schemas above the old 16 KiB cap are admitted. A 40,000-character oversized-schema regression still verifies that caller-controlled schemas beyond the budget compile uncached.

autogpt_platform/backend/backend/util/json.py
dexhunter: Addressed in 125c1773f. Serialization and the size bound now run before the Python walk, so orjson’s cycle/depth TypeError restores the prior graceful uncached path. The walk also tracks exact list/dict identities and rejects repeats. Added both an actual self-cycle regression and a shared-container regression.

autogpt_platform/backend/backend/util/test_json.py
dexhunter: Addressed in 125c1773f. The tests now instantiate both SendSlackMessageBlock and OrchestratorBlock, assert their production input schemas are cacheable, and verify repeated compilation returns the same published validator. The focused file passes 58 tests, including concurrent shared-validator use.

dexhunter:
The four review items are addressed in 125c1773f, and the threads are resolved. Python tests (3.11, 3.12, and 3.13), lint, type checks, and the CLA check are green.

The Vercel check still says "Authorization required to deploy." Could a maintainer authorize the preview for this head if it is required for review?

0ubbe (approved):

ntindle:
🤖 Addressed in [28b465322](https://github.com/Significant-Gravitas/AutoGPT/commit/28b4653227fb9430221092416df28feee20edf00): uncached schemas no longer pay for deepcopy or require a copy protocol, and both cache-mutating tests use isolated dictionaries. Also reproduced and fixed a cache-key race: if the caller mutates the schema between keying and snapshotting, the validator is now published under the key of the actual compiled snapshot. Both regressions failed before the fixes; all 61 tests, whole-backend checks, and pre-commit checks pass.

ntindle:
🤖 Reviewed the optional notes for `28b4653227fb9430221092416df28feee20edf00`. Keeping content revalidation on every call: an id-based memo guarded only against identity reuse would still serve stale validators after in-place schema mutation, which is part of the tested contract. Avoiding that requires a separate immutable/versioned schema contract. The byte threshold intentionally bounds retained cache entries; early streaming size detection would require another serializer or conservative admission policy and is deferred pending a measured need. The Orchestrator test deliberately exercises the production large-schema fallback with a reduced test bound. Keeping the private helpers together at the shared validation seam for this scoped change. All 61 cache tests and current-head CI pass; no new code changes are needed for these advisory notes.

Issue #8937, opened 2024-12-10
Fix Laggy Builder Canvas


Torantulino:
The builder canvas is very laggy when panning, zooming or dragging blocks.

Related issue with advice on how to fix here: [https://github.com/xyflow/xyflow/issues/2289](https://github.com/xyflow/xyflow/issues/2289) which could also be relevant to our case.

For reference, this is how well the builder *should* run: [https://reactflow.dev/examples/nodes/stress](https://reactflow.dev/examples/nodes/stress)

PR #13750, opened 2026-08-01
fix(frontend/builder): stabilize ReactFlow prop references on the builder canvas

chelsealong:
### Why / What / How

**Why:** #8937 reports that the agent builder canvas (`/build`) is laggy when panning, zooming, or dragging blocks, and links to [xyflow/xyflow#2289](https://github.com/xyflow/xyflow/issues/2289) for context. That thread's root cause was unstable prop references being passed into `<ReactFlow>` from a component that re-renders on every drag frame.

Tracing the same pattern in this codebase: `Flow.tsx` re-renders on every node-position update while dragging (`nodes` comes from a live Zustand store subscription), and it was passing two unstable references into `<ReactFlow>` on every render:
- `deleteKeyCode={["Backspace", "Delete"]}` — a fresh array literal every render
- `onNodeContextMenu={(event) => { event.preventDefault(); }}` — a fresh arrow function every render

Inside `@xyflow/react`, `deleteKeyCode` feeds `useKeyPress`, whose listener-registration `useEffect` depends directly on the `keyCode` reference (`useEffect(..., [keyCode, setKeyPressed])`, not a memoized/deep-compared value). Because the array reference changes on every render, this effect tears down and re-adds `keydown`/`keyup`/`blur`/`contextmenu` listeners on **every single render** — i.e. on every frame while dragging or panning. That's real synchronous DOM work happening continuously during the exact interactions the issue describes as laggy.

**What:** Stabilized both prop references so they keep the same identity across renders:
- Hoisted the delete-key array to a module-level constant (`DELETE_KEY_CODES`).
- Wrapped `onNodeContextMenu` in `useCallback` with an empty dependency array.

**How:** Minimal, behavior-preserving change — `Flow.tsx` still calls `event.preventDefault()` on node context-menu the same way, and `deleteKeyCode` still resolves to `["Backspace", "Delete"]`; only the JS reference identity of these two props is now stable, so React Flow's internal effects don't needlessly resubscribe every render.

### Changes 🏗️

- `autogpt_platform/frontend/src/app/(platform)/build/components/FlowEditor/Flow/Flow.tsx`: hoist `deleteKeyCode` to a module-level `DELETE_KEY_CODES` constant; wrap `onNodeContextMenu` in `useCallback`.
- `autogpt_platform/frontend/src/app/(platform)/build/components/FlowEditor/Flow/__tests__/Flow.test.tsx`: capture the props passed to the mocked `<ReactFlow>` and assert `deleteKeyCode`/`onNodeContextMenu` keep the same reference (`Object.is`) across a re-render.

### Test plan

- Added `Flow canvas re-render stability > deleteKeyCode and onNodeContextMenu keep a stable reference across re-renders` to `Flow.test.tsx`.
  - **Fails without the fix**: reverting only `Flow.tsx` to `HEAD~1` (test kept) gives `AssertionError: expected [ 'Backspace', 'Delete' ] to be [ 'Backspace', 'Delete' ] // Object.is equality` — proving the array was a new reference every render before this change.
  - **Passes with the fix.**
- Ran the full frontend builder test suite: `pnpm test:unit` scoped to `src/app/(platform)/build` → **37 files / 425 tests passed**, no regressions.
- `pnpm lint` (eslint) on the changed files: clean.
- `pnpm types` (tsc --noEmit): the two changed files produce zero errors. (The full project-wide `tsc` run currently reports pre-existing `Expert`-type errors in unrelated `copilot`/`marketplace`/`team` files caused by a stale local `openapi.json` snapshot vs. newer backend fields — confirmed identical with and without this change, so unrelated to this PR.)

#### For code changes:
- [x] I have made a test plan
  - [x] `pnpm test:unit` passes for the builder test suite (425/425)
  - [x] New regression test fails on the pre-fix code and passes on the fix
  - [x] `pnpm lint` and `pnpm types` clean for the changed files

Not applicable — no configuration changes.

---
_Disclosure: this change was prepared by an autonomous AI coding agent (Claude) investigating issue #8937, including root-cause tracing into `@xyflow/react`'s internals and the regression test. Please review carefully._


chelsealong:
Addressed the mock-structure nitpick: the mocked \`ReactFlow\` in \`Flow.test.tsx\` now uses a named function declaration (\`MockReactFlow\`) with a local, non-exported \`interface MockReactFlowProps\`, instead of an inline arrow function with an inline object type — matching the repo's component convention. \`pnpm lint\`/\`pnpm types\` clean, and the full builder suite (\`src/app/(platform)/build\`) still passes 425/425, including the 4 tests in \`Flow.test.tsx\`.

Left the file-length nitpick (\`Flow.test.tsx\` at 210 lines vs. the ~200 guideline) as-is: splitting the new stability suite into a separate file would require duplicating the full \`~140\`-line mock setup needed to render \`<Flow />\` (all the child-component and store mocks), which costs more in duplication than the ~10 lines saved. A sibling file in the same directory (\`NodeCost.test.tsx\`, 267 lines) already exceeds this guideline for the same reason, so I don't think it's worth the tradeoff here.

autogpt_platform/frontend/src/app/(platform)/build/components/FlowEditor/Flow/__tests__/Flow.test.tsx
```diff
@@ -175,3 +185,27 @@ describe("Flow read-only gating", () => {
...
+    mockIsReadOnly = false;
+    mockHasWebhookNodes = false;
+    mockCapturedReactFlowProps.length = 0;
+  });
+
+  test("deleteKeyCode and onNodeContextMenu keep a stable reference across re-renders", () => {
+    const { rerender } = render(<Flow />);
+    rerender(<Flow />);
+
+    expect(mockCapturedReactFlowProps.length).toBeGreaterThanOrEqual(2);
```
kcze, line 200: 🤖 🟡 **Nice to Have**: the test asserts reference identity but never the prop *values* or the handler's behaviour, so it stays green under mutations that break the feature.

Replacing the handler body with `useCallback(() => {}, [])` — dropping `event.preventDefault()` — passes, and so does changing `DELETE_KEY_CODES` to `[]` or `["Escape"]`. In the first case right-click on a node starts showing the browser context menu; in the second Backspace/Delete stops deleting nodes. `codecov/patch/platform-frontend` being red on this PR is consistent with the handler body never executing.

Two lines close it: `expect(firstProps.deleteKeyCode).toEqual(["Backspace", "Delete"])`, and invoking `firstProps.onNodeContextMenu({ preventDefault: spy })` with an assertion that the spy fired. _(flagged by: Claude)_

autogpt_platform/frontend/src/app/(platform)/build/components/FlowEditor/Flow/Flow.tsx
```diff
@@ -61,6 +63,10 @@ export const Flow = () => {
   const nodeTypes = useMemo(() => ({ custom: CustomNode }), []);
   const edgeTypes = useMemo(() => ({ custom: CustomEdge }), []);
 
+  const onNodeContextMenu = useCallback((event: MouseEvent) => {
```
kcze, line 66: 🤖 🔵 **Nit**: this handler closes over nothing, so a module-level `function handleNodeContextMenu(event: MouseEvent) { event.preventDefault(); }` next to `DELETE_KEY_CODES` is stable by construction and drops a hook plus its dependency-array reasoning. The sibling prop got exactly that treatment two lines up. Defensible as-is — `frontend/AGENTS.md` allows `useCallback` when you're deliberately optimising, which this is. _(flagged by: Claude)_

chelsealong:
Addressed the "Nice to Have" on the reference-stability test: it now also asserts `deleteKeyCode` equals `["Backspace", "Delete"]` and invokes the captured `onNodeContextMenu` with a spy to confirm `preventDefault` actually fires. Verified both mutations you described (`useCallback(() => {}, [])` and `DELETE_KEY_CODES = ["Escape"]`) now fail this test; full builder suite still passes 425/425.

Pwuts (commented):
## 🤖 Review — stable ReactFlow prop references

Reviewed at `ed27c52b3dd2c469ecbb8e0eacf1a8dd7d8f92ce`. **Looks good to me** — correct, complete, and the effect is bigger than the description claims.

`Flow` re-renders on every frame while a block is dragged, and it was handing `<ReactFlow>` a fresh `["Backspace", "Delete"]` array and a fresh arrow each time. The description explains the listener churn in `useKeyPress` (whose effect really does depend on `[keyCode, setKeyPressed]` in 12.9.2), but the larger cost is that `GraphView`, `NodeRenderer` and `NodeComponentWrapper` are all `memo()`-wrapped and both props pass straight through them — so every node in the graph re-rendered on every drag frame.

Measured with your exact prop set, ten nodes over ten parent re-renders: **100 node re-renders → 0**, and listener re-registrations **10 → 0**. Of the 17 props `Flow` passes, only six reach the memoised `GraphView`; two are primitives, `nodeTypes`/`edgeTypes` were already memoised, and these two were the only ones left — so this is exactly the right change and no more. `Flow.tsx` is the only `<ReactFlow>` in the app, so no sibling canvas carries the same bug.

The test earns its keep: all four mutations I ran kill it, including the two kcze asked for in August (stable-but-empty handler, stable-but-wrong key list). Nothing to fix, and no blockers — the only red check is `Vercel`'s `Authorization required to deploy`, which is fork-PR deploy authorization rather than anything in this diff.

Sorry this sat for six weeks. Approval is @Pwuts's to give.

<details>
<summary>Evidence</summary>

**Measurement** — real `@xyflow/react` 12.9.2 under vitest/happy-dom, 10 `custom` nodes, 10 parent re-renders, dev's prop shape vs this PR's, all of `Flow`'s other props held identical:

| variant | node re-renders / 10 frames | `keydown` listener re-adds |
|---|---|---|
| inline array + inline arrow (`dev`) | 100 | 10 |
| `DELETE_KEY_CODES` + `useCallback` (this PR) | **0** | **0** |

With a moving node in the mix (the realistic drag), idle-node re-renders go 180 → 90; the residual 90 is the store update for the node that actually moved, which this PR neither causes nor should touch.

**Mutations** — 4 run, 4 killed, each applied with an asserted match count of 1 against `Flow.tsx`:

| # | mutation | killed by |
|---|---|---|
| 1 | `deleteKeyCode` back to an inline array | `expected [ 'Backspace', 'Delete' ] to be [ 'Backspace', 'Delete' ] // Object.is equality` |
| 2 | `onNodeContextMenu` back to an inline arrow | `expected [Function] to be [Function] // Object.is equality` |
| 3 | stable handler, no `preventDefault` | `expected "vi.fn()" to be called at least once` |
| 4 | `DELETE_KEY_CODES = ["Escape"]` | `expected [ 'Escape' ] to deeply equal [ 'Backspace', 'Delete' ]` |

**Also checked:** `npx tsc --noEmit` exits 0 with zero project-wide errors at this head (the description's note about pre-existing `Expert` type errors is no longer true); `Flow.test.tsx` 4/4 passing under `LC_ALL=en_US.UTF-8`; no conflict with #14531, which edits both of these files — `git merge-tree --write-tree` exits 0; the Vercel Web Interface Guidelines audit over `Flow.tsx` passes, with the only two items (`dark:` class at line 107, context-menu suppression with no replacement) both pre-existing on `dev`.

</details>

autogpt_platform/frontend/src/app/(platform)/build/components/FlowEditor/Flow/__tests__/Flow.test.tsx
Pwuts: 🤖 Addressed in `ed27c52` and verified: I re-ran both mutations you named against the current head and this test now kills each of them — `useCallback(() => {}, [])` fails with `expected "vi.fn()" to be called at least once`, and `DELETE_KEY_CODES = ["Escape"]` fails with `expected [ 'Escape' ] to deeply equal [ 'Backspace', 'Delete' ]`. The two identity assertions still kill the original regression as well (4 mutations run, 4 killed). Resolving.

autogpt_platform/frontend/src/app/(platform)/build/components/FlowEditor/Flow/Flow.tsx
Pwuts: 🤖 Leaving as-is, agreeing with your own caveat: `frontend/AGENTS.md` allows `useCallback` when you are deliberately optimising, which is the whole point of this PR, and the empty dependency array is correct since the handler closes over nothing. A module-level function would be marginally leaner but the measured effect is identical, so it is not worth another round-trip on a six-week-old external PR. Resolving.

Pwuts (approved):

Pwuts:
Thank you for contributing this! :)

PR #13752, opened 2026-08-01
fix(backend): support marketplace media uploads without a GCS bucket

### LINKED ISSUE #9254: Fix local hosted agent uploading to work with local file system
When the platform is hosted locally and a user tries to upload a photo with an agent the marketplace it fails because there is no GCS bucket setup for this, we should make it work by storing the info locally

chelsealong:
### Why / What / How

**Why:** Fixes #9254. When the platform is self-hosted without `MEDIA_GCS_BUCKET_NAME` configured, uploading a marketplace listing image/video (or generating an agent thumbnail) fails outright with `StorageConfigError: Missing storage bucket configuration`, because `upload_media`/`check_media_exists` in `store/media.py` unconditionally required a GCS bucket.

**What:** When no GCS bucket is configured, marketplace media is now stored on the local filesystem instead of failing, and served back via a new `GET /api/store/media/{user_id}/{media_type}/{filename}` endpoint. Behavior when `MEDIA_GCS_BUCKET_NAME` *is* set is unchanged.

**How:** Mirrors the local/GCS fallback pattern already used by `backend/util/workspace_storage.py` for CoPilot workspace files (same `media_gcs_bucket_name` config flag already treated as optional there). Files are written under `{app_data}/store_media/users/{user_id}/{images|videos}/{filename}` with path-traversal protection via the existing `sanitize_filename` helper and a `Path.is_relative_to` containment check, then served back through a small public GET route (marketplace images are public content, viewable without auth, matching the other `public`-tagged store routes).

`{app_data}` (`get_data_path()`) resolves to the same directory that `workspace_storage.py` uses for `workspaces/`, and that directory is backed by a durable `workspace-data` named volume in the project's own `docker-compose.platform.yml`. Mirroring the code pattern without also mirroring that volume would mean a `docker compose up --build`/redeploy silently wipes every locally-stored marketplace image/video, so this PR adds an equivalent `store-media-data` volume mounted on `rest_server` (the service that actually serves the `/api/store/*` routes), and adds `store_media/` to `.gitignore` so bare (non-Docker) self-hosted runs can't accidentally get uploaded media committed via `git add -A`.

### Changes 🏗️

- `backend/api/features/store/media.py`: `upload_media`/`check_media_exists` fall back to local filesystem storage when `media_gcs_bucket_name` is unset, instead of raising `StorageConfigError`/`MissingConfigError`.
- `backend/api/features/store/routes.py`: new `GET /media/{user_id}/{media_type}/{filename}` route to serve locally-stored marketplace media.
- `backend/api/features/store/media_test.py`: replaced the test asserting a hard failure with no GCS bucket (`test_upload_media_missing_credentials`) with `test_upload_media_falls_back_to_local_storage` and added `test_check_media_exists_local_storage` covering the new behavior.
- `.env.default`: updated the `MEDIA_GCS_BUCKET_NAME` comment to reflect that it's optional (local storage fallback) rather than required.
- `docker-compose.yml` / `docker-compose.platform.yml`: added a `store-media-data` named volume, mounted on `rest_server` at `/app/autogpt_platform/backend/store_media`, so locally-stored marketplace media survives container recreation/redeploys the same way `workspace-data` already does for `workspaces/`.
- `backend/.gitignore`: added `store_media/`, mirroring the existing `workspaces/` entry.

#### For code changes:
- [x] I have made a test plan
  - [x] Added/updated unit tests in `media_test.py` covering local-storage upload and existence-check
  - [x] Confirmed the new tests fail against the pre-fix code (`AttributeError` / would raise `StorageConfigError`) and pass with the fix
  - [x] Ran the full `media_test.py` suite (13 passed) via `poetry run test`
  - [x] `poetry run ruff check`, `black --check`, `isort --check`, and `pyright` all pass on the changed files

- [x] `.env.default` is updated (comment only — `MEDIA_GCS_BUCKET_NAME` clarified as optional; default value unchanged)
- [x] `docker-compose.yml`/`docker-compose.platform.yml` updated: added `store-media-data` named volume on `rest_server`, mirroring the existing `workspace-data` volume, so local-storage fallback media survives redeploys
- [x] Configuration change described above

### Test output

```
$ poetry run test backend/api/features/store/media_test.py -v
...
backend/api/features/store/media_test.py::test_upload_media_success PASSED
backend/api/features/store/media_test.py::test_upload_media_invalid_type PASSED
backend/api/features/store/media_test.py::test_upload_media_falls_back_to_local_storage PASSED
backend/api/features/store/media_test.py::test_check_media_exists_local_storage PASSED
backend/api/features/store/media_test.py::test_upload_media_video_type PASSED
backend/api/features/store/media_test.py::test_upload_media_file_too_large PASSED
backend/api/features/store/media_test.py::test_upload_media_file_read_error PASSED
backend/api/features/store/media_test.py::test_upload_media_png_success PASSED
backend/api/features/store/media_test.py::test_upload_media_gif_success PASSED
backend/api/features/store/media_test.py::test_upload_media_webp_success PASSED
backend/api/features/store/media_test.py::test_upload_media_webm_success PASSED
backend/api/features/store/media_test.py::test_upload_media_mismatched_signature PASSED
backend/api/features/store/media_test.py::test_upload_media_invalid_signature PASSED
======================= 13 passed, 2 warnings in 40.94s ========================
```

TDD verification — reverting only `media.py`/`routes.py` (`git stash` on those two files) and re-running the two new tests confirms they fail without the fix:
```
$ poetry run test backend/api/features/store/media_test.py -v -k local_storage
FAILED backend/api/features/store/media_test.py::test_upload_media_falls_back_to_local_storage
FAILED backend/api/features/store/media_test.py::test_check_media_exists_local_storage
================ 2 failed, 11 deselected, 2 warnings in 37.68s =================
```

Validated the compose changes resolve correctly (`docker compose -f docker-compose.yml config`): the `store-media-data` volume is declared and mounted on `rest_server` at `/app/autogpt_platform/backend/store_media`, alongside the existing `workspace-data` mount.

---
*This PR was prepared with AI assistance (Claude) and reviewed before submission.*


chelsealong:
Fixed the MIME-type-confusion issue flagged in the review: `upload_media` was deriving the stored file extension from the client-supplied filename rather than the validated content type, so a file with a valid GIF/PNG/etc. signature but a filename like `payload.html` would be stored (and later served via `FileResponse`) with an attacker-chosen extension/content type. The stored extension is now looked up from a `CONTENT_TYPE_EXTENSIONS` map keyed by the already-validated `content_type`, never from the filename. Added `test_upload_media_extension_matches_validated_content_type`, which fails on the pre-fix code (asserts `.html` when it should be `.gif`) and passes with the fix; full `media_test.py` suite (14 tests) and `poetry run format` (ruff/isort/black/pyright) pass. Left the "split upload_media into helpers" nitpick and the CodeQL path-traversal alerts as-is: the extension fix doesn't change function length, and the path handling already goes through `sanitize_filename` (strips `/`/`\`) plus a `resolve()` + `is_relative_to()` containment check, mirroring the existing accepted pattern in `workspace_storage.py`.

chelsealong:
Fixed the failing `check API types` check: `openapi.json` was stale after the new `GET /api/store/media/{user_id}/{media_type}/{filename}` route. Regenerated it with `poetry run export-api-schema` + `pnpm prettier --write` per the check's own instructions (no generated API hooks changed, since this route is consumed as a plain URL, not via a query hook). `pnpm types` passes with no errors.

chelsealong:
Addressed the CodeRabbit nitpick on the local media route: `media_type` is now typed `Literal["images", "videos"]` (invalid values get a framework-level 422 instead of the manual check) and the 404 "not found" response is now declared via `responses={404: {"description": "Media not found"}}`. Regenerated `openapi.json` from source. Added `test_get_local_submission_media_invalid_media_type_rejected` and `test_get_local_submission_media_missing_file_returns_404`; confirmed the first fails on pre-fix code (404 instead of 422) and both pass with the fix — full `routes_test.py` + `media_test.py` suite (40 tests) passes.

chelsealong:
Addressed the CodeRabbit nitpick: `test_get_local_submission_media_missing_file_returns_404` used a fixed filename (`does-not-exist.png`) under the real local media directory, so a leftover file from a previous run could make it return 200 instead of 404. Switched to a `uuid.uuid4()`-based filename so the test can't collide with anything on disk. Verified directly (bypassing pytest's session-scoped DB/Redis fixtures, which aren't available in this sandbox): with a stale file at the old fixed path, the old-style request returned 200; with the uuid-based filename, the same dirty directory still correctly returns 404. `black`/`isort`/`ruff`/`pyright` all pass on the changed file.

chelsealong:
Addressed the failing CodeQL check: `get_local_media_path`/`_local_media_url` in `media.py` built the on-disk path from `sanitize_filename()`-substituted `user_id`/`filename`, which static path-injection analysis doesn't trust as a barrier (6 high-severity "uncontrolled data used in a path expression" alerts). Switched to an allow-list check (`^[A-Za-z0-9_.-]+$`) that rejects (`ValueError`) any component containing a slash/traversal/other unsafe character outright, instead of mangling it, before any `Path`/filesystem call. Added `test_get_local_media_path_rejects_disallowed_characters`, confirmed it fails on pre-fix code (no exception raised, e.g. `"../../etc/passwd"` silently became `".._.._etc_passwd"`) and passes with the fix; full `media_test.py` (18 tests), `black`/`isort`/`ruff`/`pyright` pass on the changed files.

chelsealong:
Addressed the still-failing CodeQL check: after the previous allow-list fix, CodeQL raised 6 new high-severity `py/path-injection` alerts on the same lines, meaning pathlib's `resolve()`/`is_relative_to()` containment check in `get_local_media_path` isn't recognized as a sanitizer by this CodeQL version. Switched to the `os.path.realpath()` + `startswith()` idiom (CodeQL's own documented safe pattern for this query) while keeping the same upfront allow-list validation. Behavior is unchanged — verified the existing traversal-rejection cases (`../../etc`, `sub/dir/file.png`, etc.) and the valid-path case still resolve identically via a standalone script, since this sandbox can't run the poetry/docker-based test suite. `black`/`isort` pass on the changed file; will confirm the CodeQL check clears once CI reruns on the new commit.

chelsealong:
Addressed the still-failing CodeQL check (5 new high-severity `py/path-injection` alerts on `b95cea7c`, the previous realpath+startswith commit): read the actual query source (`PathInjectionQuery.qll`/`PathInjectionCustomizations.qll` in github/codeql) rather than guessing again — `os.path.realpath` is a recognized `PathNormalization` step and `str.startswith` a recognized `SafeAccessCheck` guard, but only for a *simple* `if not x.startswith(prefix): raise` condition (the exact shape of the query's own qhelp example). Our guard was compound (`real_candidate != real_base_dir and not real_candidate.startswith(...)`), which gives the analyzer an alternate path to the `return` that bypasses the `startswith` check entirely, so it never registered as a sanitizer. Removed that dead branch (`real_candidate` always has a `users/<id>/<type>/<file>` suffix, so it can never equal `real_base_dir`) and kept a single clean guard. Verified the traversal-rejection and valid-path cases behave identically old vs new via a standalone script (this sandbox can't run the poetry-based test suite); `black`/`isort`/`ruff` show zero new findings on the changed file. Will confirm CodeQL clears once CI reruns on `d21870a6`.

chelsealong:
Confirming CodeQL clears on `d21870a6`: the `CodeQL` check now reports "No new alerts in code changed by this pull request" (0 annotations on the Analyze (python) job), and the previously-flagged `py/path-injection` alerts on `media.py`/`routes.py` do not recur. All other required checks pass; the only red check is Vercel ("Authorization required to deploy"), which is an org-authorization gate unrelated to this PR's code.

chelsealong:
Fixed the blocker from the latest review: `_local_media_url` returned a root-relative `/api/store/media/...` URL, which the frontend resolved against its own origin (`:3000`) instead of the backend (`:8006`), 404ing on locally-stored marketplace/library media. It now prefixes the path with `platform_base_url` (the same config `webhook_ingress_url` already uses to build backend-reachable URLs), mirroring the absolute URL the GCS branch returns, falling back to the old relative path only if `platform_base_url` is unset. Added `test_upload_media_local_storage_url_is_absolute`, which fails on the pre-fix code (asserts the `http://localhost:8006/...` prefix) and passes with the fix; verified via a standalone script against the real `media.py`/`Settings` code (this sandbox can't run the docker/DB-backed `poetry run test` suite — same limitation as previous rounds on this PR) that both the prefixed and unset-`platform_base_url` (falls back to relative) cases behave as expected. `black`/`isort`/`ruff`/`pyright` all pass on the changed files.

Pwuts (commented):
## 🤖 Review — superseded by #14524

Thanks for this, and sorry it took so long to come back to you. The short version: the work landed. #14524 merged to `dev` on 2026-09-12, describes itself as consolidating this PR, and credits you as a git coauthor — so the marketplace-media-without-GCS case is fixed on `dev` today.

For the record, `dev` now carries every piece this PR adds: the empty-bucket fallback, the public `GET /api/store/media/{user_id}/{media_type}/{filename}` route, the same `^[A-Za-z0-9_.-]+$` allow-list plus `realpath` containment on `user_id` and `filename`, the extension derived from the validated content type, the absolute URL built from the backend's public base URL, and the `store-media-data` Compose volume. It also adds things this PR does not: an atomic temp-file-plus-`os.replace` write, an explicit `media_type` and `X-Content-Type-Options: nosniff` on the served response, storage beside `WORKSPACE_STORAGE_DIR` for persistent installs, and the frontend layer (21 files, on a `@/lib/store-media` helper wired into `next/image`) that makes the local URLs actually render.

I checked whether any hunk was worth keeping and could not find one. Rebasing would also be worse than it looks: 6 of 9 files conflict, and `routes.py` is not one of them — it auto-merges into two handlers on the identical path, where FastAPI answers with `dev`'s and yours becomes dead code emitting a duplicate `openapi.json` entry.

Also worth saying so it does not look like unfinished business: the `autogpt-pr-reviewer` change request above is stale. Its only blocker was the relative media URL, and your last commit, `024254427b`, fixed exactly that.

**Verdict: no changes requested — recommending this be closed as superseded, with the credit already on `dev`.** Closing is a maintainer call rather than mine. Nothing merges on this PR in the meantime: it has no auto-merge and no merge-queue entry, and it is currently conflicting.

<details><summary>What I checked</summary>

- Compared the PR delta (`git diff $(git merge-base origin/dev HEAD)...HEAD`, head `024254427b`) against `dev` at `95a736d1`, file by file.
- Simulated the merge: `git merge-tree --write-tree origin/dev HEAD` exits 1 with conflicts in `.env.default`, `.gitignore`, `media.py`, `media_test.py`, `docker-compose.platform.yml`, `openapi.json`; `routes.py` auto-merges and the resulting tree registers `/media/{user_id}/{media_type}/{filename}` twice.
- CI on this head: 48 check runs, all success except `Vercel` (fork PRs have no deploy credentials — `Authorization required to deploy`), so backend `test (3.11/3.12/3.13)`, all three `type-check` jobs, `lint`, `integration_test`, `check API types` and `CodeQL` are green. The 7 CodeQL path-injection alerts from August are cleared.
- Could not run the backend suite locally on this head: all 19 tests error with `prisma.errors.DataError: The column User.notifyOnAgentRun does not exist`, because the branch predates the `dev` migration that dropped it. Branch age, not code.
- One thing you may want to know regardless of the close: two of the tests added before the absolute-URL fix still assert the relative form (`media_test.py:101`, `media_test.py:170`), so they contradict `test_upload_media_local_storage_url_is_absolute` and pass only where `platform_base_url` is empty — which CI is, by default. Calling `upload_media` directly with that setting configured returns the absolute URL and the `startswith` assertion is `False`.

</details>

Pwuts:
This was adopted into #14524, in which you are credited. Thank you very much for your contribution! :)

PR #13772, opened 2026-08-04
feat(platform): expert scheduling — attribution, triggers, thread posts, credit guardrail

0ubbe:
### Why / What / How

**Why:** Experts must work as always-on teammates, not just chat: roster cadences should become real schedules at hire, every scheduled/triggered run must be attributable to its expert, results should land in her thread, and an always-on expert must never surprise anyone into churn. Today none of that exists — a scheduled run is indistinguishable from a manual one in the DB. Implements #13714, #13715, #13716, and #13717 in one stack (they share one schema/threading foundation).

**What (by issue):**

**Foundation — attribution.** `AgentGraphExecution.expertId` (nullable FK, indexed `[expertId, createdAt]`, mirroring `ChatSession.expertId` archive-survival semantics) + `AgentPreset.expertId`. `expert_id` flows through `add_graph_execution` (the chokepoint all entry points use), the scheduler's job args, preset firing, and the runtime `ExecutionContext` (recovered from the row on resume/requeue, like org/team).

**#13714 — install-time scheduling.** Roster preloads carry a cron (`ExpertWorkflow.scheduleCron`); hiring creates the attributed schedule in the user's timezone and records `scheduleId`. Failure (typically missing credentials) keeps the hire and surfaces the workflow as **needs setup** (cron kept, `scheduleId` null). Re-seeding propagates cadence changes (the old sync was create-only). Manual scheduling keeps attribution: the schedules route accepts an ownership-validated `expert_id` or auto-resolves on a unique (user, graph) → expert match — ambiguous matches (two experts sharing one LibraryAgent) deliberately resolve to no attribution rather than a guess. Card shows "Every day at 07:40 · Last run succeeded 2 hours ago" via new `last_run_at`/`last_run_status` + per-workflow schedule fields.

**#13715 — triggers.** Preset creation resolves expert attribution server-side (setup-trigger flow, POST /presets, and preset-from-execution inherits from the source run); webhook and manual preset executions stamp `expert_id`. Archiving pauses the expert (logged in `ExpertPauseEvent`), deactivates her presets, deletes her schedules, and clears `scheduleId`s; re-hiring reverses all of it, recreating schedules from the stored cadence. `GET /experts/{id}/detach-preview` returns the schedule/trigger names for the confirm dialog. (No FE archive control exists yet on dev, so the dialog itself is a follow-up FE task; the run-time gate refusing archived experts is the backstop against orphaned firing either way.)

**#13716 — run results post into her thread.** Executor completion hook (next to the run notification, sync `DatabaseManagerClient` path since the executor is Prisma-less) posts a templated message into the expert's latest thread via a new `append_expert_run_message` chat-db primitive that find-or-creates her session. Success posts reuse the existing AI activity-status summary as the body; failures state the reason honestly. Dedup is a deterministic UUIDv5 per execution (message-id PK collision → no-op), so retries never double-post. Redis daily cap (10/expert) with the activity feed as overflow. The FE already renders assistant messages with expert identity, so posts appear with zero UI work.

**#13717 — weekly credit guardrail.** `Expert.weeklyBudget` (null → platform default `expert_weekly_credit_budget_default`, default 500; ≤0 disables) + `schedulesPausedAt` + `ExpertPauseEvent` audit table. Billing meters expert-attributed spend into a per-ISO-week Redis counter (pre-flight sync + post-flight reconcile deltas incl. refunds); `CreditTransaction` joined through `expertId` remains the durable source, the counter is a fail-open cache. `add_graph_execution` gates expert-attributed runs: paused/archived/over-budget → `ExpertRunPausedError` (auto-mapped over RPC); breach pauses her and posts an in-voice thread message; 80% posts a once-per-week warning (both deduped by deterministic weekly message ids — no extra state). Chat runs never carry an `expert_id` and are never gated. `POST /experts/{id}/schedules/resume` is the one-click reversal (stamps `clearedAt` on the open pause event). Card shows "12 of 50 credits this week" and a paused banner with a Resume button.

**How:** Migrations follow the repo's big-table pattern (NOT VALID FK + separate VALIDATE on `AgentGraphExecution`; plain FKs on small tables). New `experts/scheduling.py` module owns schedule/trigger lifecycle + the budget gate (kept out of `experts_db` for file size and out of `library/db` to avoid an import cycle — resolution is called from the route layer). Everything DB-facing that the Prisma-less executor/scheduler needs is exposed on the DatabaseManager (both clients). Legacy APScheduler job kwargs stay deserializable (`expert_id` defaults to None).

### Changes 🏗️

- `schema.prisma` + 3 migrations: execution/preset `expertId`, `ExpertWorkflow.scheduleCron/scheduleId`, `Expert.weeklyBudget/schedulesPausedAt`, `ExpertPauseEvent`
- `executor/`: `expert_id` on `add_graph_execution`/`ExecutionContext`/job args; budget gate; `expert_posts.py` completion hook; scheduler skips paused runs quietly
- `data/`: `expert_spend.py` Redis weekly counters; billing increments (pre-flight + reconcile); `db_manager` exposures (`append_expert_run_message`, `enforce_expert_run_budget`, `get_library_agent_id_by_graph_id`)
- `copilot/db.py`: `append_expert_run_message` (find-or-create latest expert thread, message-id dedup)
- `api/features/experts/`: `scheduling.py` (create/detach/reattach/pause/resume/gate/preview), hire-time scheduling, cadence-aware seeds, model fields, `detach-preview` + `schedules/resume` routes
- `api/features/library/` + `integrations/`: preset `expert_id` (model, create, setup-trigger, from-execution inherit) + firing attribution
- `api/features/v1.py`: `ScheduleCreationRequest.expert_id` + ownership check + unique-match auto-resolution
- `frontend`: `ExpertTeamCard` status line (schedule · last run · spend), needs-setup chips, paused banner + one-click Resume (`useExpertTeamCard`), regenerated `openapi.json`
- Tests: 10 team-page integration tests, 7 expert-posts unit tests, 3 append-message unit tests, 2 scheduler unit tests, 10 DB-backed experts tests (archive/revive, budget pause/resume, resolver, cadence sync, last-run); snapshot updated
- Config: `expert_weekly_credit_budget_default` (new setting, default 500; `.env.default` unchanged — standard Settings field)

`data/*.py` note: new write paths (`create_graph_execution.expert_id`, `expert_spend`, `append_expert_run_message`) take the `user_id`/`expert_id` their validated callers pass; expert ownership is checked at the API boundary (`v1.py`, experts routes) and inside `enforce_expert_run_budget`/`pause`/`resume` via `ownerUserId`-scoped queries.

#### For code changes:
- [x] I have made a test plan
  - [x] Team-page integration tests 10/10 (schedule summary, needs-setup, spend, paused + resume) — `vitest run src/app/(platform)/team/__tests__/main.test.tsx`
  - [x] `pnpm format` / `pnpm lint` / `pnpm types` clean; backend `poetry run format` + pyright clean on changed files
  - [ ] DB-backed experts tests + scheduler/expert-posts/chat-db unit tests run in CI (local docker stack unavailable; tests written first — FE pair red/green verified locally)
  - [ ] Manual on a seeded env: hire Frankie → schedule on card; run fires → post in her thread; set `weeklyBudget=1` → next scheduled run pauses her; Resume un-pauses


autogpt_platform/backend/backend/api/features/library/routes/presets.py
```diff
@@ -298,6 +305,7 @@ async def execute_preset(
         graph_id=preset.graph_id,
         graph_version=preset.graph_version,
         preset_id=preset_id,
+        expert_id=preset.expert_id,
```
0ubbe: 🤖 🟠 **Should Fix**: A manual "Run" on an expert's preset while she is paused/over budget surfaces as a 500 — `add_graph_execution` raises `ExpertRunPausedError` (a `ValueError`) and this route has no handler for it. Catch it and return a 409 with the pause message so the user gets "Maria's schedules are paused — resume from the Team page" instead of an internal error.

autogpt_platform/backend/backend/api/features/integrations/router.py
```diff
@@ -786,6 +786,7 @@ async def _execute_webhook_preset_trigger(
             nodes_input_masks={trigger_node.id: {**preset.inputs, "payload": payload}},
             organization_id=org_id,
             team_id=ws_id,
+            expert_id=preset.expert_id,
```
0ubbe, line 790: 🤖 🟡 **Nice to Have**: While an expert is paused, every webhook delivery for her preset lands in the generic `except Exception` below and logs a full stack trace (`logger.exception`). `ExpertRunPausedError` is an expected steady-state here — catch it explicitly before the generic handler and log at info, mirroring what `_execute_graph` does in the scheduler.

autogpt_platform/backend/backend/executor/utils.py
```diff
@@ -1227,6 +1233,17 @@ async def add_graph_execution(
     if not bypass_paywall and await is_user_paywalled(user_id):
         raise UserPaywalledError("A subscription is required to run agents.")
 
+    # Weekly credit guardrail: expert-attributed runs (schedules, triggers)
+    # are refused while the expert is paused/archived or over budget. Chat
+    # runs never carry an expert_id, so chat is never gated.
+    if expert_id:
```
0ubbe, line 1250: 🤖 🟡 **Nice to Have**: The budget gate runs on the `expert_id` *parameter*, which is `None` in REQUEUE mode — a requeued/resumed expert execution (admin recovery, review-resume) bypasses the guardrail even though the persisted row carries `expertId`. Low impact since requeue is a deliberate human action, but worth either gating on `graph_exec.expert_id` after the fetch or documenting the exemption here.

autogpt_platform/backend/backend/data/execution.py
```diff
@@ -876,6 +876,7 @@ async def create_graph_execution(
     is_dry_run: bool = False,
     organization_id: Optional[str] = None,
     team_id: Optional[str] = None,
+    expert_id: Optional[str] = None,
```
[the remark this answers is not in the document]
0ubbe: 🤖 Fixed in [e3b9d4231](https://github.com/Significant-Gravitas/AutoGPT/commit/e3b9d42310cdeabeabc2a829a97e468aee00fd7a): `create_graph_execution` now validates that `expert_id` belongs to `user_id` (owned, non-template) before insertion and raises `ValueError` on mismatch — one indexed query, only on the expert-attributed path.

autogpt_platform/backend/backend/api/features/library/routes/presets.py
0ubbe: 🤖 Fixed in [e3b9d4231](https://github.com/Significant-Gravitas/AutoGPT/commit/e3b9d42310cdeabeabc2a829a97e468aee00fd7a): `execute_preset` now catches `ExpertRunPausedError` and returns 409 with the pause message.

autogpt_platform/backend/backend/api/features/integrations/router.py
0ubbe: 🤖 Fixed in [e3b9d4231](https://github.com/Significant-Gravitas/AutoGPT/commit/e3b9d42310cdeabeabc2a829a97e468aee00fd7a): `_execute_webhook_preset_trigger` catches `ExpertRunPausedError` explicitly and logs at info, mirroring the scheduler's handling.

autogpt_platform/backend/backend/executor/utils.py
0ubbe: 🤖 Addressed in [e3b9d4231](https://github.com/Significant-Gravitas/AutoGPT/commit/e3b9d42310cdeabeabc2a829a97e468aee00fd7a): kept the exemption (requeue is an explicit human/admin recovery action on an execution that already passed the gate at creation) and documented it in the gate comment.

autogpt_platform/backend/backend/executor/expert_posts_test.py
```diff
@@ -0,0 +1,127 @@
...
+        _entry(expert_id="expert-1"),
+        ExecutionStatus.RUNNING,
+        GraphExecutionStats(),
+    )
+    db_client.append_expert_run_message.assert_not_called()
+
+
+def test_post_uses_deterministic_message_id_for_retries():
+    db_client = MagicMock()
+    db_client.get_graph_metadata.return_value = MagicMock(name="Agent")
```
0ubbe: 🤖 🔵 **Nit**: `MagicMock(name="Agent")` sets the mock's *own* name, not the `.name` attribute — `metadata.name` here returns a child mock, so the built message embeds a mock repr and the test never exercises a real agent name. Use `SimpleNamespace(name="Agent")` (or configure the attribute) so the deterministic-id test also validates realistic content.

0ubbe: 🤖 Fixed in [a3f5376f2](https://github.com/Significant-Gravitas/AutoGPT/commit/a3f5376f209d2f43f2f31c17af8fc7297956abc9): swapped to `SimpleNamespace(name="Morning Brief")` and added a content assertion on the real name.

autogpt_platform/backend/backend/api/features/experts/experts_db.py
```diff
@@ -157,6 +193,14 @@ async def _existing_hire_result(row: prisma.models.Expert) -> HireResult:
...
         if revived is not None:
             row = revived
+        if row.ownerUserId:
+            await scheduling.resume_expert_schedules(row.ownerUserId, row.id)
+            try:
+                await scheduling.reattach_expert_triggers(row.ownerUserId, row.id)
+            except Exception:
+                logger.exception(
+                    f"Failed to reattach triggers while reviving expert #{row.id}"
+                )
```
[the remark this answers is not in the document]
0ubbe: 🤖 Fixed in [6583cc3f7](https://github.com/Significant-Gravitas/AutoGPT/commit/6583cc3f7af60dd0a0d97b8b7faf4f795ece9e1a): the revived expert is reloaded (with workflows) after resume/reattach, so `HireResult` reflects the cleared pause and recreated schedule ids.

autogpt_platform/backend/backend/api/features/experts/scheduling.py
```diff
@@ -0,0 +1,285 @@
...
+            reason=f"Weekly credit budget reached ({spent}/{budget})",
+        )
+        await _post_budget_message(user_id, expert, spent, budget, breached=True)
+        raise ExpertRunPausedError(
+            f"{expert.name} hit her weekly credit budget ({spent}/{budget}); "
+            "schedules are paused until you resume them.",
+            expert_id,
+        )
+    if spent >= int(budget * _BUDGET_WARN_FRACTION):
+        await _post_budget_message(user_id, expert, spent, budget, breached=False)
```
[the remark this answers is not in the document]
0ubbe: 🤖 Working as intended, now documented in [6583cc3f7](https://github.com/Significant-Gravitas/AutoGPT/commit/6583cc3f7af60dd0a0d97b8b7faf4f795ece9e1a): the read is deliberately a snapshot, not a reservation. The overshoot is bounded (per-run cost × concurrent firings), the next gate check pauses her, and the durable wallet is charged correctly by the credit system regardless — this gate is a churn guardrail, not the billing ledger. An atomic reserve/reconcile scheme would add real complexity for marginal protection here; the trade-off is now stated in the docstring.

autogpt_platform/backend/backend/data/expert_spend.py
```diff
@@ -0,0 +1,76 @@
...
+    try:
+        redis = await get_redis_async()
+        value = await redis.get(weekly_spend_key(expert_id))
+        return int(value) if value is not None else 0
+    except Exception as e:
+        logger.warning(
+            f"Failed to read weekly spend for expert #{expert_id}: "
+            f"{type(e).__name__}: {e}"
+        )
+        return 0
```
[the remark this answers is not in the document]
0ubbe: 🤖 Fixed in [6583cc3f7](https://github.com/Significant-Gravitas/AutoGPT/commit/6583cc3f7af60dd0a0d97b8b7faf4f795ece9e1a): `get_weekly_spend` clamps to `max(0, …)` and documents the cross-week refund case.

autogpt_platform/backend/backend/executor/expert_posts.py
```diff
@@ -0,0 +1,126 @@
...
+        ),
+    )
+    db_client.append_expert_run_message(
+        user_id=graph_exec.user_id,
+        expert_id=expert_id,
+        content=content,
+        message_id=str(
+            uuid.uuid5(_POST_NAMESPACE, f"run-post:{graph_exec.graph_exec_id}")
+        ),
+    )
```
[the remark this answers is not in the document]
0ubbe: 🤖 Fixed in [6583cc3f7](https://github.com/Significant-Gravitas/AutoGPT/commit/6583cc3f7af60dd0a0d97b8b7faf4f795ece9e1a): the cap slot is released whenever no message lands — both on exception and on retry-dedup (`append_expert_run_message` returning None).

autogpt_platform/backend/backend/executor/utils.py
```diff
@@ -1227,6 +1233,17 @@ async def add_graph_execution(
...
+        if prisma.is_connected():
+            await experts_scheduling.enforce_expert_run_budget(user_id, expert_id)
+        else:
+            await get_database_manager_async_client().enforce_expert_run_budget(
+                user_id, expert_id
+            )
```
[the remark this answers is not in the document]
0ubbe: 🤖 Fixed in [6583cc3f7](https://github.com/Significant-Gravitas/AutoGPT/commit/6583cc3f7af60dd0a0d97b8b7faf4f795ece9e1a): REQUEUE mode now enforces the gate on the persisted `graph_exec.expert_id` after the lookup; admin recovery stays exempt via the existing `bypass_paywall` signal.

autogpt_platform/backend/backend/util/settings.py
```diff
@@ -147,6 +147,10 @@ class Config(UpdateTrackingModel["Config"], BaseSettings):
         default=500,
         description="Credit threshold for low balance notifications (100 = $1, default 500 = $5)",
     )
+    expert_weekly_credit_budget_default: int = Field(
+        default=500,
+        description="Default weekly credit budget per hired expert when the expert has no explicit budget (100 = $1). 0 disables the guardrail.",
+    )
```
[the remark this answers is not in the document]
0ubbe: 🤖 Fixed in [6583cc3f7](https://github.com/Significant-Gravitas/AutoGPT/commit/6583cc3f7af60dd0a0d97b8b7faf4f795ece9e1a): added `ge=0` to the setting.

autogpt_platform/frontend/src/app/(platform)/team/components/ExpertTeamCard/ExpertTeamCard.tsx
```diff
@@ -54,6 +71,29 @@ export function ExpertTeamCard({
...
+            variant="secondary"
+            size="small"
+            loading={isResuming}
+            onClick={(event) => {
+              event.stopPropagation();
+              handleResume();
+            }}
+          >
+            Resume schedules
+          </Button>
```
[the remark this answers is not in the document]
0ubbe: 🤖 Fixed in [6583cc3f7](https://github.com/Significant-Gravitas/AutoGPT/commit/6583cc3f7af60dd0a0d97b8b7faf4f795ece9e1a): named `handleResumeClick`/`handleResumeKeyDown` handlers stop both click and keyboard propagation before they reach the card.

autogpt_platform/backend/backend/api/features/experts/scheduling.py
```diff
@@ -0,0 +1,303 @@
...
+        )
+        try:
+            await get_scheduler_client().delete_schedule(schedule.id, user_id=user_id)
+        except Exception as cleanup_error:
+            logger.error(
+                f"Orphaned schedule #{schedule.id} for expert #{expert_id} "
+                f"could not be deleted: {type(cleanup_error).__name__}: "
+                f"{cleanup_error}"
+            )
+        return False
```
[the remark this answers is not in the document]
0ubbe: 🤖 Fixed in [6583cc3f7](https://github.com/Significant-Gravitas/AutoGPT/commit/6583cc3f7af60dd0a0d97b8b7faf4f795ece9e1a): registration is idempotent via a guarded `update_many(where scheduleId: None)` — a loser of the race deletes its duplicate job (shared `_delete_schedule_best_effort` cleanup).

autogpt_platform/backend/backend/api/features/library/routes/presets.py
```diff
@@ -293,13 +304,19 @@ async def execute_preset(
...
+            expert_id=preset.expert_id,
+            inputs=merged_node_input,
+            graph_credentials_inputs=merged_credential_inputs,
+            organization_id=exec_org_id,
+            team_id=exec_team_id,
+        )
+    except ExpertRunPausedError as e:
+        # A paused/over-budget expert is a user-visible state, not a server
+        # error — tell the user how to unblock instead of 500ing.
+        raise HTTPException(status_code=status.HTTP_409_CONFLICT, detail=str(e))
```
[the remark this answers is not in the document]
0ubbe: 🤖 Fixed in [6583cc3f7](https://github.com/Significant-Gravitas/AutoGPT/commit/6583cc3f7af60dd0a0d97b8b7faf4f795ece9e1a): `ExpertRunPausedError` now defines `__str__` returning only the message (args still carries both values for RPC reconstruction), so the 409 detail renders the pause message.

autogpt_platform/backend/backend/api/features/experts/scheduling.py
```diff
@@ -0,0 +1,326 @@
...
+        )
+        await _delete_schedule_best_effort(schedule.id, user_id, expert_id)
+        return False
+    if updated == 0:
+        logger.info(
+            f"Workflow #{workflow_row_id} already has a schedule; deleting "
+            f"duplicate registration #{schedule.id}"
+        )
+        await _delete_schedule_best_effort(schedule.id, user_id, expert_id)
+        return False
```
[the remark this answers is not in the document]
0ubbe: 🤖 Addressed in [ef8128199](https://github.com/Significant-Gravitas/AutoGPT/commit/ef81281997d824e36c8f529c2114056239e6ada7): the best-effort delete now retries once (covers transient RPC blips), and a surviving orphan is not an untracked leak — it stays expert-attributed, so it appears in the detach preview and is deleted by the archive detach sweep. A durable retry queue for this doubly-degraded path (lost registration race **and** immediate delete failure against the same healthy-enough scheduler that just accepted the add) would be disproportionate; the loud per-id error log is the operator signal.

autogpt_platform/backend/backend/executor/expert_posts.py
```diff
@@ -0,0 +1,149 @@
...
+            content=content,
+            message_id=str(
+                uuid.uuid5(_POST_NAMESPACE, f"run-post:{graph_exec.graph_exec_id}")
+            ),
+        )
+    except Exception:
+        _release_cap_slot(graph_exec.user_id, expert_id)
+        raise
+    if posted_session is None:
+        _release_cap_slot(graph_exec.user_id, expert_id)
```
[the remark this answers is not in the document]
0ubbe: 🤖 Fixed in [ef8128199](https://github.com/Significant-Gravitas/AutoGPT/commit/ef81281997d824e36c8f529c2114056239e6ada7): the cap key is computed once at admission and passed explicitly to both release paths; added the requested rollover regression test asserting reserve and release hit the same key.

0ubbe:
!deploy

Pwuts:

- **Deployed:** `f36bcd64927ae951f12a66079888c1fc70cf4ad3` at 2026-08-04 20:35 UTC
- **Database:** isolated Supabase branch `pr-13772` (state persists across redeploys unless migration drift forces a reset)

autogpt_platform/backend/backend/api/features/experts/scheduling.py
```diff
@@ -0,0 +1,332 @@
...
+        except Exception as cleanup_error:
+            last_error = cleanup_error
+    logger.error(
+        f"Orphaned schedule #{schedule_id} for expert #{expert_id} "
+        f"could not be deleted: {type(last_error).__name__}: {last_error}"
+    )
+
+
+async def _get_expert_schedules(user_id: str, expert_id: str) -> list:
+    schedules = await get_scheduler_client().get_execution_schedules(user_id)
```
kcze: 🤖 🔴 **Blocker**: `user_id` is passed **positionally**, but the signature is `get_execution_schedules(graph_id=None, user_id=None, ...)` — so it binds to `graph_id` and the user filter stays `None`. No graph job ever has `graph_id == <a user id>`, so `_get_expert_schedules` always returns `[]`.

Everything built on it silently no-ops: archiving an expert deletes **none** of her schedules (they keep firing), and `GET /experts/{id}/detach-preview` reports "no schedules" in the confirm dialog. The unit test masks this — the scheduler is a permissive `AsyncMock` that accepts any arg shape.

Fix: `get_execution_schedules(user_id=user_id, kind="graph")`. _(flagged by: Claude-r2, corroborated by Codex)_

autogpt_platform/backend/backend/api/features/experts/scheduling.py
```diff
@@ -0,0 +1,332 @@
...
+        data={"expertId": expert_id, "reason": reason}
+    )
+    return True
+
+
+async def resume_expert_schedules(user_id: str, expert_id: str) -> bool:
+    """One-click reversal of a pause; stamps the open pause event."""
+    updated = await prisma.models.Expert.prisma().update_many(
+        where={"id": expert_id, "ownerUserId": user_id, "isTemplate": False},
+        data={"schedulesPausedAt": None},
```
kcze, line 266: 🤖 🔴 **Blocker**: Resume clears `schedulesPausedAt` but nothing clears the week's spend, so a budget pause is **unrecoverable until the ISO week rolls over**. The very next fire re-enters `enforce_expert_run_budget`, `get_weekly_spend` still returns the same `spent >= budget`, and she is paused again immediately.

There is no escape hatch: `Expert.weeklyBudget` is **read-only across the whole codebase** (only `effective_weekly_budget` and the response model reference it) — no route writes it. So the one-click Resume button is a no-op for the rest of the week, which is exactly the churn the guardrail was meant to prevent.

Needs either a spend offset stamped at resume, or a budget-override write path. _(flagged by: Claude-r1 + r2 + r3 + Codex — 4/5 reviewers independently)_

autogpt_platform/backend/migrations/20260804000000_add_execution_expert_attribution/migration.sql
```diff
@@ -0,0 +1,24 @@
...
+
+-- CreateIndex
+-- AgentGraphExecution is an existing, write-heavy table. Prisma's
+-- `migrate deploy` wraps each migration in a single transaction, and
+-- Postgres rejects `CREATE INDEX CONCURRENTLY` inside a transaction block;
+-- the plain form briefly holds a ShareLock during the scan. For deployments
+-- that can't tolerate that, run the CONCURRENTLY equivalent out-of-band
+-- before this migration ships and Postgres will skip the recreate via
+-- IF NOT EXISTS (same pattern as 20260727070306_add_expert_entity).
+CREATE INDEX IF NOT EXISTS "AgentGraphExecution_expertId_createdAt_idx" ON "AgentGraphExecution"("expertId", "createdAt");
```
kcze: 🤖 🔴 **Blocker**: Two problems, both on the largest table in the schema.

**1. The lock comment is wrong.** `ADD COLUMN` on line 2 takes `ACCESS EXCLUSIVE`, and Prisma wraps the whole migration in **one transaction** — so that lock is held through this `CREATE INDEX` build, not the "brief ShareLock" the comment claims. Reads *and* writes on `AgentGraphExecution` block for the full index build.

**2. The escape hatch is impossible to execute.** It says to build the index `CONCURRENTLY` out-of-band "before this migration ships" — but `expertId` does not exist until line 2 of *this same file*. An operator who pre-creates the column to make that possible then hits `ADD COLUMN` (no `IF NOT EXISTS`) → duplicate-column error → failed migration → **all deploys blocked**. This repeats the broken escape hatch a prior experts migration shipped.

Split the index into its own migration after the column lands, and make line 2 `ADD COLUMN IF NOT EXISTS`. _(flagged by: Claude-r3)_

autogpt_platform/backend/backend/api/features/experts/scheduling.py
```diff
@@ -0,0 +1,332 @@
...
+        try:
+            await scheduler.delete_schedule(schedule.id, user_id=user_id)
+        except Exception as e:
+            logger.warning(
+                f"Failed to delete schedule #{schedule.id} while detaching "
+                f"expert #{expert_id}: {type(e).__name__}: {e}"
+            )
+    await prisma.models.ExpertWorkflow.prisma().update_many(
+        where={"expertId": expert_id},
+        data={"scheduleId": None},
```
kcze: 🤖 🟠 **Should Fix**: `scheduleId` is cleared for **every** workflow unconditionally — even when the `delete_schedule` calls above failed (each is swallowed by its own `except`) or never ran at all.

Combined with the positional-arg bug on line 128, today's behavior is: schedules are never deleted, yet every `scheduleId` pointer is wiped. On re-hire, `reattach_expert_triggers` selects on `scheduleId: None`, sees them all as unscheduled, and creates a **second** schedule per workflow — the expert now fires twice per cadence, with the original job orphaned and unreachable.

Only null the `scheduleId` of schedules whose deletion actually succeeded. _(flagged by: Codex, corroborated by Claude-r2)_

autogpt_platform/backend/backend/executor/expert_posts.py
```diff
@@ -0,0 +1,151 @@
...
+
+def _post_run_result(
+    db_client: "DatabaseManagerClient",
+    graph_exec: GraphExecutionEntry,
+    status: ExecutionStatus,
+    exec_stats: GraphExecutionStats,
+) -> None:
+    context = graph_exec.execution_context
+    expert_id = context.expert_id if context else None
+    if not expert_id or (context and context.dry_run):
```
kcze, line 54: 🤖 🟠 **Should Fix**: The guard checks only `expert_id` and `dry_run`, never the execution hierarchy — but `ExecutionContext` carries `expert_id` into sub-graph runs, and it already exposes `parent_execution_id`/`root_execution_id` right here.

So an expert workflow containing 3 `AgentExecutorBlock` sub-agents posts **4 messages** for one logical run — one per nested execution, each announcing a different sub-agent name — and burns 4 of the 10/day cap slots.

Add `or context.parent_execution_id is not None` to the early return. _(flagged by: Claude-r1 + r4)_

autogpt_platform/backend/backend/executor/expert_posts.py
```diff
@@ -0,0 +1,151 @@
...
+    # message actually lands (failure or retry-dedup), so failed attempts
+    # and re-fired completions can't silently eat the day's budget.
+    try:
+        metadata = db_client.get_graph_metadata(
+            graph_exec.graph_id, graph_exec.graph_version
+        )
+        content = build_expert_run_message(
+            agent_name=metadata.name if metadata else "your workflow",
+            succeeded=status == ExecutionStatus.COMPLETED,
+            summary=exec_stats.activity_status,
```
kcze, line 84: 🤖 🟠 **Should Fix**: `activity_status` is LLM-generated text derived from execution output, and it is interpolated straight into an `assistant` message that becomes trusted conversation history for every later turn in that thread.

An agent that scrapes a page containing `Ignore previous instructions...` gets that text summarized into `activity_status`, posted as the *expert's own words*, and replayed to the model on the user's next message — injection with no visible provenance that it originated in scraped content. Attribute the body as untrusted (fence it, or mark it as tool output) rather than emitting it in the assistant's voice. _(flagged by: Claude-r4)_

autogpt_platform/backend/backend/executor/expert_posts.py
```diff
@@ -0,0 +1,151 @@
...
+            error=str(exec_stats.error) if exec_stats.error else None,
```
kcze, line 85: 🤖 🟠 **Should Fix**: `str(exec_stats.error)` is unbounded — a block failing with a large payload (full traceback, API response body, dumped DataFrame) writes that entire string into a `ChatMessage` row.

One failed run can persist a multi-MB message that is then loaded into context on every subsequent turn of that thread. Truncate to a sane cap before embedding. _(flagged by: Claude-r4)_

autogpt_platform/backend/backend/copilot/db.py
```diff
@@ -968,3 +969,55 @@ async def update_chat_session_status(
...
+        created = await create_chat_session(
+            session_id=str(uuid.uuid4()), user_id=user_id, expert_id=expert_id
+        )
+        session_id = created.session_id
+
+    try:
+        await add_chat_message(
+            session_id=session_id,
+            role="assistant",
+            sequence=await get_next_sequence(session_id),
```
kcze: 🤖 🟠 **Should Fix**: This `get_next_sequence` + `add_chat_message` pair runs **without** `_get_session_lock(session_id)`. That Redis NX lock exists specifically to serialise this pattern — `turn_queue.append_and_save_message` documents it as "so two concurrent submits to the same session can't pick the same `sequence` and PK-collide on `(sessionId, sequence)`".

A run completing while the user is mid-turn races the locked writer and collides. The `except UniqueViolationError` retry below is only a partial band-aid: it is *also* unlocked, so it can lose the same race again and surface as a hard error. Take the lock around both the sequence read and the insert. _(flagged by: Claude-r3)_

autogpt_platform/backend/backend/api/features/experts/scheduling.py
```diff
@@ -0,0 +1,332 @@
...
+    )
+
+
+async def reattach_expert_triggers(user_id: str, expert_id: str) -> None:
+    """Reverse of ``detach_expert_triggers``, for re-hire revival:
+    reactivate presets and recreate schedules from the stored cadence."""
+    await prisma.models.AgentPreset.prisma().update_many(
+        where={"expertId": expert_id, "userId": user_id, "isDeleted": False},
+        data={"isActive": True},
```
kcze: 🤖 🟠 **Should Fix**: Re-hire reactivates **every** non-deleted preset for the expert, not just the ones archiving deactivated. A preset the user had deliberately turned off *before* archiving comes back **on** after re-hire — a trigger silently starts firing again with no user action.

Record which presets were active at detach time (or mark the ones this flow deactivated) and restore only those. _(flagged by: Codex)_

autogpt_platform/backend/migrations/20260804000002_add_expert_triggers_and_budget/migration.sql
```diff
@@ -0,0 +1,31 @@
...
+    CONSTRAINT "ExpertPauseEvent_pkey" PRIMARY KEY ("id")
+);
+
+-- CreateIndex
+CREATE INDEX "ExpertPauseEvent_expertId_createdAt_idx" ON "ExpertPauseEvent"("expertId", "createdAt");
+
+-- CreateIndex
+-- AgentPreset is modest-sized but write-visible; IF NOT EXISTS allows an
+-- out-of-band CONCURRENTLY build first (same pattern as prior migrations).
+CREATE INDEX IF NOT EXISTS "AgentPreset_expertId_idx" ON "AgentPreset"("expertId");
```
kcze, line 27: 🤖 🟡 **Nice to Have**: Same impossible escape hatch as the `AgentGraphExecution` migration — the comment invites an out-of-band `CONCURRENTLY` build first, but `expertId` is created by line 2 of this same file, so there is nothing to index beforehand. `AgentPreset` is small enough that the lock is not the concern; the misleading instruction is. Drop the out-of-band suggestion or move the column add to an earlier migration. _(flagged by: Claude-r3)_

autogpt_platform/backend/backend/api/features/experts/scheduling.py
0ubbe: 🤖 Fixed in [229f39973](https://github.com/Significant-Gravitas/AutoGPT/commit/229f3997373dc65f3eaf225711875e815acb4463): `_get_expert_schedules` now calls `get_execution_schedules(user_id=user_id, kind="graph")` with keyword args, so the user filter actually binds. The archive test now asserts the keyword call shape (`assert_awaited_with(user_id=..., kind="graph")`) so the permissive mock can't mask a positional regression again.

autogpt_platform/backend/backend/api/features/experts/scheduling.py
0ubbe: 🤖 Fixed in [229f39973](https://github.com/Significant-Gravitas/AutoGPT/commit/229f3997373dc65f3eaf225711875e815acb4463): went with the spend-offset option — `resume_expert_schedules` now calls a new `reset_weekly_spend(expert_id)` (deletes the current-week Redis counter) after clearing `schedulesPausedAt`, so the next gate check starts from 0 and Resume actually resumes. Semantics: resuming is the user explicitly accepting more spend this week; the durable billing ledger is untouched. If spend re-accumulates past the budget she pauses again, which is the guardrail working as intended. Test asserts the reset fires on resume.

autogpt_platform/backend/migrations/20260804000000_add_execution_expert_attribution/migration.sql
0ubbe: 🤖 Fixed in [229f39973](https://github.com/Significant-Gravitas/AutoGPT/commit/229f3997373dc65f3eaf225711875e815acb4463): both problems addressed. (1) The `CREATE INDEX` moved out of this migration into its own `20260804000003_add_execution_expert_index`, so it never builds under the ADD COLUMN's ACCESS EXCLUSIVE lock — that migration's comment now correctly describes the plain build's SHARE lock (reads continue, writes block) and the out-of-band CONCURRENTLY escape hatch, which is now actually executable because the column already exists by then. (2) `ADD COLUMN IF NOT EXISTS` on all three column adds, so an operator who pre-created the column can't brick deploys with a duplicate-column error. The wrong 'brief ShareLock' comment is gone.

autogpt_platform/backend/backend/api/features/experts/scheduling.py
0ubbe: 🤖 Fixed in [229f39973](https://github.com/Significant-Gravitas/AutoGPT/commit/229f3997373dc65f3eaf225711875e815acb4463): `detach_expert_triggers` now collects the ids of schedules whose `delete_schedule` call actually succeeded and only nulls `scheduleId` for those (`scheduleId: {in: deleted_ids}`). A failed deletion keeps its pointer, so the job stays visible and retryable instead of re-hire creating a duplicate next to an orphan. Added `test_detach_keeps_pointer_when_schedule_delete_fails` covering the scheduler-down path. (The positional-arg bug this compounded with is fixed in the same commit.)

autogpt_platform/backend/backend/executor/expert_posts.py
0ubbe: 🤖 Fixed in [229f39973](https://github.com/Significant-Gravitas/AutoGPT/commit/229f3997373dc65f3eaf225711875e815acb4463): added the early return — `_post_run_result` now bails when `context.parent_execution_id is not None`, so nested `AgentExecutorBlock` sub-executions neither post nor consume cap slots; only the top-level run reports. Covered by `test_post_skips_sub_graph_executions`.

autogpt_platform/backend/backend/executor/expert_posts.py
0ubbe: 🤖 Fixed in [229f39973](https://github.com/Significant-Gravitas/AutoGPT/commit/229f3997373dc65f3eaf225711875e815acb4463): the summary is no longer emitted in the expert's voice. `build_expert_run_message` now introduces it as "The run's generated summary:" and blockquotes every line (`> ...`), so scraped instruction-injection text stays visibly attributed to the run's output rather than reading as assistant speech when replayed in later turns. The failure-path error text gets the same quoted-with-provenance treatment.

autogpt_platform/backend/backend/executor/expert_posts.py
0ubbe: 🤖 Fixed in [229f39973](https://github.com/Significant-Gravitas/AutoGPT/commit/229f3997373dc65f3eaf225711875e815acb4463): the embedded error is now capped at 500 chars via `_truncate` (with an explicit "… (truncated)" marker) before it lands in the message, so a multi-MB block failure can't persist into the thread and reload on every turn. `test_build_message_truncates_oversized_error` feeds a 100k-char error and asserts the message stays under 1k.

autogpt_platform/backend/backend/copilot/db.py
0ubbe: 🤖 Fixed in [229f39973](https://github.com/Significant-Gravitas/AutoGPT/commit/229f3997373dc65f3eaf225711875e815acb4463): the sequence read + insert in `append_expert_run_message` now run inside `async with _get_session_lock(session_id)` — the same Redis NX lock `turn_queue.append_and_save_message` uses — so a run completing mid-turn serialises with the locked writer instead of racing it. The `UniqueViolationError` retry stays as the degraded-mode fallback for when Redis is down and the lock yields without acquiring.

autogpt_platform/backend/backend/api/features/experts/scheduling.py
0ubbe: 🤖 Fixed in [229f39973](https://github.com/Significant-Gravitas/AutoGPT/commit/229f3997373dc65f3eaf225711875e815acb4463): detach now marks exactly the presets it turns off — a new `AgentPreset.deactivatedByExpertArchive` flag (migration `20260804000004`) is set only on presets that were `isActive: true` at archive time, and re-hire reactivates only flagged rows (clearing the flag). A preset the user disabled before archiving keeps `deactivatedByExpertArchive: false` and stays off. The archive/revive test now includes a pre-disabled preset and asserts it survives the round trip untouched.

autogpt_platform/backend/migrations/20260804000002_add_expert_triggers_and_budget/migration.sql
0ubbe: 🤖 Fixed in [229f39973](https://github.com/Significant-Gravitas/AutoGPT/commit/229f3997373dc65f3eaf225711875e815acb4463): dropped the impossible out-of-band suggestion — the comment now says the index ships inline because AgentPreset's build window is short, and that `IF NOT EXISTS` is only there for rerun idempotency, not an escape hatch.

Abhi1992002 (approved):

Pwuts:

All resources for PR #13772 have been removed:

PR #13796, opened 2026-08-07
fix(backend): drop and recreate analytics views around onboarding retype

kcze:
## Why

The dev deploy is failing and onboarding is broken on dev. `20260805120000_onboarding_step_to_string` (from #13119) failed to apply:

```
ERROR: cannot alter type of a column used by a view or rule
DETAIL: rule _RETURN on view analytics.user_onboarding depends on column "completedSteps"
```

Postgres won't retype a column a view depends on. The analytics views read `completedSteps`, but they are generated out-of-band by `poetry run analytics-views` (`backend/scripts/generate_views.py`) from `analytics/queries/*.sql` — so they exist in deployed databases and never in one built from migrations alone. Local runs and CI both build from migrations only, which is why this passed everywhere and failed only on a real environment.

Two consequences today:

1. The columns are still `OnboardingStep[]` while the deployed backend expects `String[]`, so onboarding reads fail with `ApiError: Error converting field "completedSteps" ... found incompatible value of "BUILDER_RUN_AGENT"`.
2. Prisma recorded the migration as failed, so **no migration can apply to dev** (`P3009`) until it is resolved.

## What

The migration now **drops** the dependent relations and lets `analytics-views` recreate them. It does not restore them itself.

## How

An earlier revision captured each view's definition, owner and ACL and restored them. Review showed that to be both unnecessary and wrong: the deployed `user_onboarding_funnel` still hardcodes the **pre-rename** step grid, so restoring it verbatim would resurrect a funnel reporting 0 for `ONBOARDING_COMPLETE` and `LIBRARY_RUN_AGENT`. Re-running the generator instead rebuilds every view from the repo, which already carries the new names, and re-grants `analytics_readonly`.

Dropping rather than restoring also removes a whole class of restore bugs raised in review: lost column-level grants (`pg_attribute.attacl`), lost reloptions and comments, non-round-tripping pretty-printed viewdefs, ACLs widened by `ALTER DEFAULT PRIVILEGES`, and swallowed owner-restore failures leaving views owned by the migration role.

Discovery covers materialized views and uses `CASCADE`, so transitive dependents are handled; `pg_depend` is constrained to `classid = 'pg_rewrite'` to avoid false positives; and `lock_timeout` makes the drop fail fast rather than queue behind a long-running BI query.

## ⚠️ Required step (dev only, no DB access needed)

The failed `_prisma_migrations` row must be cleared before `migrate deploy` will do anything — it returns `P3009` until then. Thanks to the `resolve_rolled_back` input on `platform-autogpt-deploy-dev.yaml`, this is a workflow dispatch rather than a manual database connection:

```bash
gh workflow run platform-autogpt-deploy-dev.yaml \
  -f git_ref=dev \
  -f resolve_rolled_back=20260805120000_onboarding_step_to_string
```

That job runs `prisma migrate resolve --rolled-back` and then `prisma migrate deploy` against the dev database using the CI secrets, then triggers the deploy.

**Order matters: merge this PR to `dev` first.** The workflow checks out `git_ref`, so if it runs before the merge it will clear the failed row and then re-apply the *broken* migration, putting the row straight back.

`--rolled-back` (not `--applied`) is correct: Prisma runs each migration in a transaction and this one aborted on the first `ALTER`, so nothing from it persisted.

**Local databases** — anyone who pulled `dev` and migrated since #13119 merged will hit "migration … was modified after it was applied", since this edits that file. Local databases are disposable: `prisma migrate reset`.

## Does production need manual steps?

**No.** `master` has never applied this migration (newest there is `20260802120000`), so there is no failed record to clear, and the views are rebuilt in-migration.

Two things worth checking before promoting, since neither is visible from here:

**1. The deploy role must be able to drop the views** (`DROP VIEW` requires ownership or superuser):

```sql
SELECT current_user, (SELECT usesuper FROM pg_user WHERE usename = current_user) AS is_superuser;
```

**2. Enumerate the dependency closure**, to see whether prod carries dependents beyond the two tracked views. A materialized view, or any view not in `analytics/queries/`, is dropped by `CASCADE` and **cannot** be rebuilt by this migration — it would need recreating manually. The migration raises a warning naming each, but Prisma does not surface server notices, so check in advance:

```sql
SELECT n.nspname, c.relname, c.relkind
FROM pg_class c JOIN pg_namespace n ON n.oid = c.relnamespace
WHERE c.relkind IN ('v','m') AND EXISTS (
  SELECT 1 FROM pg_depend d JOIN pg_rewrite r ON r.oid = d.objid
  WHERE d.classid = 'pg_rewrite'::regclass AND r.ev_class = c.oid
    AND d.refobjid = '"UserOnboarding"'::regclass
    AND d.refobjsubid IN (
      SELECT attnum FROM pg_attribute
      WHERE attrelid = '"UserOnboarding"'::regclass
        AND attname IN ('completedSteps','notified','rewardedFor')));
```

If that returns only `analytics.user_onboarding` and `analytics.user_onboarding_funnel`, promotion is fully hands-off.

Recommended sequence: fix dev, confirm the deploy goes green there, then promote.

## Test plan

Verified against a database rebuilt to the deployed shape — full migration chain, `analytics` schema, `analytics_readonly`, views created exactly as `generate_views.py` creates them (`WITH (security_invoker = false)`), plus a view layered on `user_onboarding_funnel` and a materialized view on the base table:

- [x] The original statement reproduces the **exact** production error
- [x] Fixed migration applies; all four dependents (incl. transitive view and matview) dropped via `CASCADE`
- [x] Columns end as `TEXT[]`; `OnboardingStep` dropped; renames applied; `BUILDER_RUN_AGENT` and other values preserved
- [x] `analytics-views` then recreates the views with `reloptions = {security_invoker=false}` and `analytics_readonly` SELECT restored
- [x] Funnel reports the **renamed** steps (`LIBRARY_RUN_AGENT` 11, `ONBOARDING_COMPLETE` 15) — the regression the restore-from-catalog approach would have shipped
- [x] Backend onboarding tests pass

## Follow-up

Local and CI databases are built from migrations only, so nothing can catch a dependency on these hand-applied views. Bringing `analytics/queries/*.sql` into CI setup (or running `analytics-views` as part of deploy) would close the gap — this is the second time these views have caused trouble.

### DIFF
--- a/autogpt_platform/backend/migrations/20260805120000_onboarding_step_to_string/migration.sql
+++ b/autogpt_platform/backend/migrations/20260805120000_onboarding_step_to_string/migration.sql
@@ -37,11 +37,20 @@ WHERE c.relkind = 'v'
           AND d.refobjsubid > 0
       );
 
+-- Read the ACL from the catalog rather than information_schema, whose
+-- role_table_grants view only shows grants involving roles the current user
+-- belongs to and would silently drop the rest.
 CREATE TEMP TABLE _onboarding_dep_view_grants AS
-SELECT g.table_schema, g.table_name, g.grantee, g.privilege_type
-FROM information_schema.role_table_grants g
-JOIN _onboarding_dep_views v
-  ON v.schema_name = g.table_schema AND v.view_name = g.table_name;
+SELECT v.schema_name,
+       v.view_name,
+       CASE WHEN a.grantee = 0 THEN 'PUBLIC'
+            ELSE quote_ident(pg_get_userbyid(a.grantee)) END AS grantee_name,
+       a.privilege_type,
+       a.is_grantable
+FROM _onboarding_dep_views v
+JOIN pg_class c ON c.relname = v.view_name
+JOIN pg_namespace n ON n.oid = c.relnamespace AND n.nspname = v.schema_name
+CROSS JOIN LATERAL aclexplode(c.relacl) AS a;
 
 DO $$
 DECLARE v record;
@@ -77,18 +86,29 @@ BEGIN
     FOR v IN SELECT * FROM _onboarding_dep_views LOOP
         EXECUTE format('CREATE VIEW %I.%I AS %s',
                        v.schema_name, v.view_name, v.definition);
-        EXECUTE format('ALTER VIEW %I.%I OWNER TO %I',
-                       v.schema_name, v.view_name, v.view_owner);
+        -- Ownership is cosmetic for readers; if the migration role isn't a
+        -- member of the original owner it can't reassign. Warn rather than
+        -- abort, so a privilege quirk can't strand the migration half-applied.
+        BEGIN
+            EXECUTE format('ALTER VIEW %I.%I OWNER TO %I',
+                           v.schema_name, v.view_name, v.view_owner);
+        EXCEPTION WHEN OTHERS THEN
+            RAISE WARNING 'could not restore owner % on %.%: %',
+                          v.view_owner, v.schema_name, v.view_name, SQLERRM;
+        END;
     END LOOP;
 
     FOR g IN SELECT * FROM _onboarding_dep_view_grants LOOP
-        IF upper(g.grantee) = 'PUBLIC' THEN
-            EXECUTE format('GRANT %s ON %I.%I TO PUBLIC',
-                           g.privilege_type, g.table_schema, g.table_name);
-        ELSE
-            EXECUTE format('GRANT %s ON %I.%I TO %I',
-                           g.privilege_type, g.table_schema, g.table_name, g.grantee);
-        END IF;
+        BEGIN
+            EXECUTE format('GRANT %s ON %I.%I TO %s%s',
+                           g.privilege_type, g.schema_name, g.view_name,
+                           g.grantee_name,
+                           CASE WHEN g.is_grantable THEN ' WITH GRANT OPTION' ELSE '' END);
+        EXCEPTION WHEN OTHERS THEN
+            RAISE WARNING 'could not restore % on %.% to %: %',
+                          g.privilege_type, g.schema_name, g.view_name,
+                          g.grantee_name, SQLERRM;
+        END;
     END LOOP;
 END $$;
 
--- a/autogpt_platform/backend/migrations/20260805120000_onboarding_step_to_string/migration.sql
+++ b/autogpt_platform/backend/migrations/20260805120000_onboarding_step_to_string/migration.sql
@@ -11,6 +11,46 @@
 --     runs and is shown to users as a Library action, so the MARKETPLACE_ prefix
 --     was a misnomer).
 
+-- Postgres refuses to retype a column that a view depends on, and the
+-- analytics views (analytics.user_onboarding, analytics.user_onboarding_funnel)
+-- read these columns. They are applied out-of-band from analytics/queries/*.sql,
+-- so they exist in deployed databases but not in one built from migrations
+-- alone -- which is why this only fails on a real environment.
+--
+-- Capture whatever dependent views actually exist (definition + grants), drop
+-- them, retype, then recreate them verbatim. Reading the definition from the
+-- catalog rather than the repo means we restore exactly what was deployed, and
+-- picks up any view not tracked in this repo.
+CREATE TEMP TABLE _onboarding_dep_views AS
+SELECT n.nspname AS schema_name,
+       c.relname AS view_name,
+       pg_get_userbyid(c.relowner) AS view_owner,
+       rtrim(pg_get_viewdef(c.oid, true), E' \n;') AS definition
+FROM pg_class c
+JOIN pg_namespace n ON n.oid = c.relnamespace
+WHERE c.relkind = 'v'
+  AND EXISTS (
+        SELECT 1
+        FROM pg_depend d
+        JOIN pg_rewrite r ON r.oid = d.objid AND r.ev_class = c.oid
+        WHERE d.refobjid = 'platform."UserOnboarding"'::regclass
+          AND d.refobjsubid > 0
+      );
+
+CREATE TEMP TABLE _onboarding_dep_view_grants AS
+SELECT g.table_schema, g.table_name, g.grantee, g.privilege_type
+FROM information_schema.role_table_grants g
+JOIN _onboarding_dep_views v
+  ON v.schema_name = g.table_schema AND v.view_name = g.table_name;
+
+DO $$
+DECLARE v record;
+BEGIN
+    FOR v IN SELECT * FROM _onboarding_dep_views LOOP
+        EXECUTE format('DROP VIEW %I.%I', v.schema_name, v.view_name);
+    END LOOP;
+END $$;
+
 -- Drop defaults so the column type cast doesn't trip on the default's enum type.
 ALTER TABLE "UserOnboarding" ALTER COLUMN "completedSteps" DROP DEFAULT;
 ALTER TABLE "UserOnboarding" ALTER COLUMN "notified" DROP DEFAULT;
@@ -26,6 +66,34 @@ ALTER TABLE "UserOnboarding"
 ALTER TABLE "UserOnboarding" ALTER COLUMN "completedSteps" SET DEFAULT '{}';
 ALTER TABLE "UserOnboarding" ALTER COLUMN "notified"       SET DEFAULT '{}';
 ALTER TABLE "UserOnboarding" ALTER COLUMN "rewardedFor"    SET DEFAULT '{}';
+
+-- Recreate the dependent views verbatim, restoring owner and grants. Their
+-- step columns come back as TEXT[] instead of "OnboardingStep"[]; the queries
+-- already cast with ::text, so consumers are unaffected.
+DO $$
+DECLARE v record;
+        g record;
+BEGIN
+    FOR v IN SELECT * FROM _onboarding_dep_views LOOP
+        EXECUTE format('CREATE VIEW %I.%I AS %s',
+                       v.schema_name, v.view_name, v.definition);
+        EXECUTE format('ALTER VIEW %I.%I OWNER TO %I',
+                       v.schema_name, v.view_name, v.view_owner);
+    END LOOP;
+
+    FOR g IN SELECT * FROM _onboarding_dep_view_grants LOOP
+        IF upper(g.grantee) = 'PUBLIC' THEN
+            EXECUTE format('GRANT %s ON %I.%I TO PUBLIC',
+                           g.privilege_type, g.table_schema, g.table_name);
+        ELSE
+            EXECUTE format('GRANT %s ON %I.%I TO %I',
+                           g.privilege_type, g.table_schema, g.table_name, g.grantee);
+        END IF;
+    END LOOP;
+END $$;
+
+DROP TABLE _onboarding_dep_views;
+DROP TABLE _onboarding_dep_view_grants;
 
 -- Rename retired step names in existing rows so users keep their progress:
 -- VISIT_COPILOT -> ONBOARDING_COMPLETE and MARKETPLACE_RUN_AGENT ->
--- a/autogpt_platform/backend/migrations/20260805120000_onboarding_step_to_string/migration.sql
+++ b/autogpt_platform/backend/migrations/20260805120000_onboarding_step_to_string/migration.sql
@@ -11,52 +11,57 @@
 --     runs and is shown to users as a Library action, so the MARKETPLACE_ prefix
 --     was a misnomer).
 
--- Postgres refuses to retype a column that a view depends on, and the
--- analytics views (analytics.user_onboarding, analytics.user_onboarding_funnel)
--- read these columns. They are applied out-of-band from analytics/queries/*.sql,
--- so they exist in deployed databases but not in one built from migrations
--- alone -- which is why this only fails on a real environment.
+-- Postgres refuses to retype a column that a view depends on, and the analytics
+-- views (analytics.user_onboarding, analytics.user_onboarding_funnel) read
+-- these columns. Those views are not part of the migration history: they are
+-- applied out-of-band by `poetry run analytics-views`
+-- (backend/scripts/generate_views.py) from analytics/queries/*.sql. They
+-- therefore exist in deployed databases and never in one built from migrations
+-- alone, which is why this only fails on a real environment.
 --
--- Capture whatever dependent views actually exist (definition + grants), drop
--- them, retype, then recreate them verbatim. Reading the definition from the
--- catalog rather than the repo means we restore exactly what was deployed, and
--- picks up any view not tracked in this repo.
-CREATE TEMP TABLE _onboarding_dep_views AS
-SELECT n.nspname AS schema_name,
-       c.relname AS view_name,
-       pg_get_userbyid(c.relowner) AS view_owner,
-       rtrim(pg_get_viewdef(c.oid, true), E' \n;') AS definition
-FROM pg_class c
-JOIN pg_namespace n ON n.oid = c.relnamespace
-WHERE c.relkind = 'v'
-  AND EXISTS (
-        SELECT 1
-        FROM pg_depend d
-        JOIN pg_rewrite r ON r.oid = d.objid AND r.ev_class = c.oid
-        WHERE d.refobjid = 'platform."UserOnboarding"'::regclass
-          AND d.refobjsubid > 0
-      );
+-- Drop the dependents and let that script recreate them. Deliberately NOT
+-- restoring them from the catalog here: the deployed definition of
+-- user_onboarding_funnel still hardcodes the pre-rename step grid, so
+-- recreating it verbatim would resurrect a funnel that reports 0 for the
+-- renamed steps. Re-running the script rebuilds every view from the repo,
+-- which already carries the new names, and re-grants analytics_readonly.
+--
+-- *** After deploying, run: poetry run analytics-views ***
+--
+-- Discovery is dynamic (and CASCADE'd) so views not tracked in this repo, and
+-- any view layered on top of these, are handled too.
 
--- Read the ACL from the catalog rather than information_schema, whose
--- role_table_grants view only shows grants involving roles the current user
--- belongs to and would silently drop the rest.
-CREATE TEMP TABLE _onboarding_dep_view_grants AS
-SELECT v.schema_name,
-       v.view_name,
-       CASE WHEN a.grantee = 0 THEN 'PUBLIC'
-            ELSE quote_ident(pg_get_userbyid(a.grantee)) END AS grantee_name,
-       a.privilege_type,
-       a.is_grantable
-FROM _onboarding_dep_views v
-JOIN pg_class c ON c.relname = v.view_name
-JOIN pg_namespace n ON n.oid = c.relnamespace AND n.nspname = v.schema_name
-CROSS JOIN LATERAL aclexplode(c.relacl) AS a;
+-- Fail fast rather than queue behind a long-running BI query holding a lock on
+-- these views; the deploy can retry.
+SET LOCAL lock_timeout = '30s';
 
 DO $$
 DECLARE v record;
 BEGIN
-    FOR v IN SELECT * FROM _onboarding_dep_views LOOP
-        EXECUTE format('DROP VIEW %I.%I', v.schema_name, v.view_name);
+    FOR v IN
+        SELECT n.nspname AS schema_name, c.relname AS rel_name, c.relkind
+        FROM pg_class c
+        JOIN pg_namespace n ON n.oid = c.relnamespace
+        WHERE c.relkind IN ('v', 'm')
+          AND EXISTS (
+                SELECT 1
+                FROM pg_depend d
+                JOIN pg_rewrite r ON r.oid = d.objid
+                WHERE d.classid = 'pg_rewrite'::regclass
+                  AND r.ev_class = c.oid
+                  AND d.refobjid = 'platform."UserOnboarding"'::regclass
+                  AND d.refobjsubid > 0
+              )
+    LOOP
+        RAISE NOTICE 'dropping dependent relation %.% (recreate with: poetry run analytics-views)',
+                     v.schema_name, v.rel_name;
+        IF v.relkind = 'm' THEN
+            EXECUTE format('DROP MATERIALIZED VIEW IF EXISTS %I.%I CASCADE',
+                           v.schema_name, v.rel_name);
+        ELSE
+            EXECUTE format('DROP VIEW IF EXISTS %I.%I CASCADE',
+                           v.schema_name, v.rel_name);
+        END IF;
     END LOOP;
 END $$;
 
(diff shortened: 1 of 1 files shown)


autogpt_platform/backend/migrations/20260805120000_onboarding_step_to_string/migration.sql
```diff
@@ -11,6 +11,55 @@
 --     runs and is shown to users as a Library action, so the MARKETPLACE_ prefix
 --     was a misnomer).
 
+-- Postgres refuses to retype a column that a view depends on, and the
```
kcze: 🤖 🔴 **Blocker**: This **edits a migration that is already applied**, rather than adding a new one — `git diff --name-status origin/dev...HEAD` reports `M`, and the file exists on `dev`. Three consequences: (1) every database where it already succeeded — local dev, CI, any environment built from migrations alone (i.e. all the ones without the analytics views) — **never receives this fix**; (2) the checksum changes, so `prisma migrate dev` rejects it with "the migration … was modified after it was applied" and offers a reset; (3) on the environment where it failed, `_prisma_migrations` holds a failed row, so `migrate deploy` aborts with **P3009** until someone runs `prisma migrate resolve --rolled-back 20260805120000_onboarding_step_to_string` — a recovery step documented nowhere. A new follow-up migration that does the view drop/retype/recreate avoids all three. _(flagged by: claude-opus-5, verified)_

autogpt_platform/backend/migrations/20260805120000_onboarding_step_to_string/migration.sql
```diff
@@ -27,6 +76,45 @@ ALTER TABLE "UserOnboarding" ALTER COLUMN "completedSteps" SET DEFAULT '{}';
...
 
+-- Recreate the dependent views verbatim, restoring owner and grants. Their
+-- step columns come back as TEXT[] instead of "OnboardingStep"[]; the queries
+-- already cast with ::text, so consumers are unaffected.
+DO $$
+DECLARE v record;
+        g record;
+BEGIN
+    FOR v IN SELECT * FROM _onboarding_dep_views LOOP
+        EXECUTE format('CREATE VIEW %I.%I AS %s',
```
kcze: 🤖 🟠 **Should Fix**: Recreating the views "verbatim from the catalog" deliberately restores the **pre-rename** definition of `analytics.user_onboarding_funnel`, whose `all_steps` grid hardcodes step names. `analytics/queries/user_onboarding_funnel.sql:58,62` were updated to `LIBRARY_RUN_AGENT`/`ONBOARDING_COMPLETE`, but that file is applied **out-of-band** by `poetry run analytics-views` (`scripts/generate_views.py`) — so the catalog still holds the old grid. After this migration the deployed funnel reports `VISIT_COPILOT` = 0 and `MARKETPLACE_RUN_AGENT` = 0, **omits** `ONBOARDING_COMPLETE`/`LIBRARY_RUN_AGENT` entirely (the join is `FROM all_steps LEFT JOIN step_counts`), and `pct_from_prev` for `RE_RUN_AGENT` goes NULL via the `NULLIF(…,0)` guard. Neither the migration nor the commit message tells the operator to re-run `analytics-views` afterwards. _(flagged by: claude-opus-5, reproduced on PG15)_

autogpt_platform/backend/migrations/20260805120000_onboarding_step_to_string/migration.sql
```diff
@@ -27,6 +76,45 @@ ALTER TABLE "UserOnboarding" ALTER COLUMN "completedSteps" SET DEFAULT '{}';
...
+                       v.schema_name, v.view_name, v.definition);
+        -- Ownership is cosmetic for readers; if the migration role isn't a
+        -- member of the original owner it can't reassign. Warn rather than
+        -- abort, so a privilege quirk can't strand the migration half-applied.
+        BEGIN
+            EXECUTE format('ALTER VIEW %I.%I OWNER TO %I',
+                           v.schema_name, v.view_name, v.view_owner);
+        EXCEPTION WHEN OTHERS THEN
```
kcze: 🤖 🟠 **Should Fix**: These `EXCEPTION WHEN OTHERS … RAISE WARNING` handlers turn a loud failure into a silent one: if the owner or a grant can't be restored the migration still **reports success**, and a `WARNING` in a Prisma deploy log is easy to miss. Worse, the rationale — "Ownership is cosmetic for readers" — isn't true here. A Postgres view runs with its **owner's** privileges unless `security_invoker` is set, and none of the analytics views set it (`generate_views.py:152` uses `security_invoker = false`). So a swallowed owner-restore leaves the view owned by the migration role (superuser on Supabase), and every grantee then reads `platform."UserOnboarding"` through *that* role. Consider failing loudly, or verifying the restored ACL/owner at the end and raising if it doesn't match what was captured. _(flagged by: claude-opus-5 + orchestrator verification)_

autogpt_platform/backend/migrations/20260805120000_onboarding_step_to_string/migration.sql
```diff
@@ -11,6 +11,55 @@
...
+--
+-- Capture whatever dependent views actually exist (definition + grants), drop
+-- them, retype, then recreate them verbatim. Reading the definition from the
+-- catalog rather than the repo means we restore exactly what was deployed, and
+-- picks up any view not tracked in this repo.
+CREATE TEMP TABLE _onboarding_dep_views AS
+SELECT n.nspname AS schema_name,
+       c.relname AS view_name,
+       pg_get_userbyid(c.relowner) AS view_owner,
+       rtrim(pg_get_viewdef(c.oid, true), E' \n;') AS definition
```
kcze: 🤖 🟠 **Should Fix**: The capture records only definition + owner, so **reloptions and comments are silently dropped** on recreate — `security_invoker`, `security_barrier`, `WITH CHECK OPTION`, and `COMMENT ON VIEW`. Verified on PG15: `security_barrier=true` and the view comment were both gone after the drop/recreate cycle. Benign for today's views (they're generated with `security_invoker = false`, the server default), but any view deployed with `security_invoker = true` silently flips to **definer** semantics — a widening of data access that appears nowhere in the diff. Separately, `pg_get_viewdef(c.oid, true)` requests the pretty-printed form, which the PG docs warn is less likely to round-trip identically; `pg_get_viewdef(c.oid)` is the correct choice for drop/recreate. _(flagged by: claude-opus-5 ×2)_

autogpt_platform/backend/migrations/20260805120000_onboarding_step_to_string/migration.sql
```diff
@@ -11,6 +11,55 @@
...
+        FROM pg_depend d
+        JOIN pg_rewrite r ON r.oid = d.objid AND r.ev_class = c.oid
+        WHERE d.refobjid = 'platform."UserOnboarding"'::regclass
+          AND d.refobjsubid > 0
+      );
+
+-- Read the ACL from the catalog rather than information_schema, whose
+-- role_table_grants view only shows grants involving roles the current user
+-- belongs to and would silently drop the rest.
+CREATE TEMP TABLE _onboarding_dep_view_grants AS
```
kcze: 🤖 🟠 **Should Fix**: Moving to `aclexplode(c.relacl)` fixed the `PUBLIC` and `is_grantable` gaps, but **column-level grants are still lost** — those live in `pg_attribute.attacl`, which `relacl` doesn't cover. Verified on PG15: `GRANT SELECT ("userId") ON analytics.user_onboarding TO bi` disappeared entirely after the migration, leaving that role with zero privileges while the deploy reported success. Capturing `pg_attribute.attacl` for the view's columns alongside `relacl` would close it. _(flagged by: claude-opus-5 ×2, reproduced)_

autogpt_platform/backend/migrations/20260805120000_onboarding_step_to_string/migration.sql
```diff
@@ -27,6 +76,45 @@ ALTER TABLE "UserOnboarding" ALTER COLUMN "completedSteps" SET DEFAULT '{}';
...
+-- Recreate the dependent views verbatim, restoring owner and grants. Their
+-- step columns come back as TEXT[] instead of "OnboardingStep"[]; the queries
+-- already cast with ::text, so consumers are unaffected.
+DO $$
+DECLARE v record;
+        g record;
+BEGIN
+    FOR v IN SELECT * FROM _onboarding_dep_views LOOP
+        EXECUTE format('CREATE VIEW %I.%I AS %s',
+                       v.schema_name, v.view_name, v.definition);
```
kcze: 🤖 🟠 **Should Fix**: The views are recreated **by the migration role**, so any `ALTER DEFAULT PRIVILEGES` configured for that role in that schema is applied to the new view *on top of* the restored grants — and the migration never `REVOKE`s anything, so the resulting ACL can only be a **superset** of the original. On Supabase, where `ALTER DEFAULT PRIVILEGES IN SCHEMA … GRANT ALL ON TABLES TO anon, authenticated` is customary, a view previously readable by one reporting role comes back readable by `anon`/`authenticated`. That is a privilege escalation on onboarding data, introduced by a migration whose stated goal is to "restore exactly what was deployed". Revoking all privileges on the recreated view before re-applying the captured ACL would make the restore exact. _(flagged by: claude-opus-5)_

autogpt_platform/backend/migrations/20260805120000_onboarding_step_to_string/migration.sql
```diff
@@ -11,6 +11,55 @@
...
+          AND d.refobjsubid > 0
+      );
+
+-- Read the ACL from the catalog rather than information_schema, whose
+-- role_table_grants view only shows grants involving roles the current user
+-- belongs to and would silently drop the rest.
+CREATE TEMP TABLE _onboarding_dep_view_grants AS
+SELECT v.schema_name,
+       v.view_name,
+       CASE WHEN a.grantee = 0 THEN 'PUBLIC'
```
kcze: 🤖 🟡 **Nice to Have**: No `SET LOCAL lock_timeout` (or `statement_timeout`) before the first lock-taking statement. Failure: `DROP VIEW analytics.user_onboarding` queues for `ACCESS EXCLUSIVE` behind a long-running BI query on that view; every subsequent query touching `UserOnboarding` then queues behind the *pending* lock request, stalling the onboarding/login path before any real work has started — and the migration waits indefinitely rather than failing fast so the deploy can retry. The sibling migrations `20260804000003_add_execution_expert_index` and `20260516120000_add_creator_search_trgm_indexes` both reason explicitly about lock strategy; this one doesn't. _(flagged by: claude-opus-5)_

autogpt_platform/backend/migrations/20260805120000_onboarding_step_to_string/migration.sql
```diff
@@ -11,6 +11,55 @@
...
+       pg_get_userbyid(c.relowner) AS view_owner,
+       rtrim(pg_get_viewdef(c.oid, true), E' \n;') AS definition
+FROM pg_class c
+JOIN pg_namespace n ON n.oid = c.relnamespace
+WHERE c.relkind = 'v'
+  AND EXISTS (
+        SELECT 1
+        FROM pg_depend d
+        JOIN pg_rewrite r ON r.oid = d.objid AND r.ev_class = c.oid
+        WHERE d.refobjid = 'platform."UserOnboarding"'::regclass
```
kcze: 🤖 🟡 **Nice to Have**: The dependency probe joins `pg_depend.objid` to `pg_rewrite.oid` without constraining `d.classid = 'pg_rewrite'::regclass`, so a `pg_depend` row of any other catalog class whose `objid` happens to equal a rewrite-rule OID is a false positive. Failure: an unrelated view is needlessly dropped and recreated — inheriting every grant-loss and reloption-loss exposure above — for a table it has nothing to do with. _(flagged by: claude-opus-5)_

autogpt_platform/backend/migrations/20260805120000_onboarding_step_to_string/migration.sql
kcze: Fixed in 20deef127e — and this finding changed the whole design, thank you. You're right that restoring `user_onboarding_funnel` verbatim resurrects the pre-rename grid. The migration no longer restores anything: it drops the dependents and defers recreation to `poetry run analytics-views`, which rebuilds every view from `analytics/queries/*.sql` (already carrying the new names) and re-grants `analytics_readonly`. Verified end-to-end locally — after the migration plus that script, the funnel reports `LIBRARY_RUN_AGENT` (order 11) and `ONBOARDING_COMPLETE` (order 15) with correct counts, and the retired names are absent. The required step is now called out in the migration header, the commit message, and the PR description.

autogpt_platform/backend/migrations/20260805120000_onboarding_step_to_string/migration.sql
kcze: Resolved by removal (20deef127e). Restoring from the catalog is gone, so reloptions, comments, `WITH CHECK OPTION` and the pretty-print round-trip concern no longer apply — `analytics-views` recreates each view with `WITH (security_invoker = false)` from source. Confirmed after a full cycle that `reloptions` comes back as `{security_invoker=false}`. Your point about `pg_get_viewdef(oid, true)` was correct regardless and would have bitten the previous version.

autogpt_platform/backend/migrations/20260805120000_onboarding_step_to_string/migration.sql
kcze: Resolved by removal (20deef127e) — no ACL capture remains, so there is nothing to lose. `analytics-views` re-grants `analytics_readonly`, which is the model those views are provisioned under. Being explicit about the residual: any **manually** added column-level grant to some other role is not reproduced by that script and would need re-applying. That is now a property of the analytics tooling rather than something this migration silently drops.

autogpt_platform/backend/migrations/20260805120000_onboarding_step_to_string/migration.sql
kcze: Resolved by removal (20deef127e). The migration no longer recreates views, so it can't produce a superset ACL. Worth stating plainly rather than closing it out: `analytics-views` recreates views as the role running it, so any `ALTER DEFAULT PRIVILEGES` in that schema still applies at that point. That behaviour predates this change and is unchanged by it, but your `anon`/`authenticated` observation is a real property of the analytics tooling on Supabase and is worth a separate look.

autogpt_platform/backend/migrations/20260805120000_onboarding_step_to_string/migration.sql
kcze: Resolved by removal (20deef127e) — there are no `EXCEPTION WHEN OTHERS` handlers left, because there is no owner/grant restoration to fail. Your correction stands and was the important part: "ownership is cosmetic" was wrong of me. `generate_views.py:152` sets `security_invoker = false`, so these views execute with the owner's privileges, and a swallowed owner-restore would have left every grantee reading `platform.\"UserOnboarding\"` through the migration role.

autogpt_platform/backend/migrations/20260805120000_onboarding_step_to_string/migration.sql
```diff
@@ -11,6 +11,46 @@
...
+          AND d.refobjsubid > 0
+      );
```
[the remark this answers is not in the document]
kcze: Fixed in 20deef127e. Discovery now matches `relkind IN ('v','m')` so materialized views are included, and each drop is `CASCADE`, which removes the closure without needing to compute it or order it. Verified against a database carrying a view layered on `analytics.user_onboarding_funnel` and a materialized view on the base table: previously the drop failed, now all four dependents are removed and the retype succeeds.

autogpt_platform/backend/migrations/20260805120000_onboarding_step_to_string/migration.sql
```diff
@@ -11,6 +11,46 @@
...
+FROM information_schema.role_table_grants g
+JOIN _onboarding_dep_views v
+  ON v.schema_name = g.table_schema AND v.view_name = g.table_name;
+
+DO $$
+DECLARE v record;
+BEGIN
+    FOR v IN SELECT * FROM _onboarding_dep_views LOOP
+        EXECUTE format('DROP VIEW %I.%I', v.schema_name, v.view_name);
+    END LOOP;
```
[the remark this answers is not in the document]
kcze: Fixed in 20deef127e — the drops are now `CASCADE`, and discovery also covers materialized views. Confirmed with a view layered on top of `analytics.user_onboarding_funnel`: without `CASCADE` the drop failed exactly as described; with it, the whole chain is removed and the retype proceeds. Recreation is handled by `poetry run analytics-views`, which rebuilds every view in the schema, so cascaded dependents come back too.

autogpt_platform/backend/migrations/20260805120000_onboarding_step_to_string/migration.sql
kcze: Fixed in 20deef127e — `SET LOCAL lock_timeout = '30s'` now precedes the first lock-taking statement, so the drop fails fast and the deploy can retry rather than queueing for `ACCESS EXCLUSIVE` behind a long-running BI query and stalling everything touching `UserOnboarding` behind the pending request.

autogpt_platform/backend/migrations/20260805120000_onboarding_step_to_string/migration.sql
kcze: Fixed in 20deef127e — the probe now constrains `d.classid = 'pg_rewrite'::regclass`, so a `pg_depend` row from another catalog class whose `objid` collides with a rewrite-rule OID can no longer drag an unrelated view into the drop set.

autogpt_platform/backend/migrations/20260805120000_onboarding_step_to_string/migration.sql
kcze: Thanks — the three consequences are all real, but I've kept the edit rather than adding a follow-up migration, because the follow-up approach breaks **production**.

**Production has never applied this migration.** `master`'s newest is `20260802120000_add_capability_cards_step`; `20260805120000` exists only on `dev`. So if the broken file stays as-is and a follow-up is added, the next prod deploy runs `migrate deploy`, reaches the **original** migration first, and fails on prod's analytics views exactly as dev did — leaving prod with a failed migration and P3009. The follow-up never gets a chance to run. Fixing the file in place is what makes the prod path work first time.

On the three consequences:

1. **Databases where it already succeeded never receive the fix** — true, and harmless. It succeeded there precisely because those databases have no analytics views, so the old version already produced the correct end state (`TEXT[]`, enum dropped, renames applied). The new version differs only in dropping dependents that don't exist.
2. **Checksum change** — real, and the actual cost. Anyone who pulled `dev` and migrated since it merged (~09:46 UTC today) will hit "migration … was modified after it was applied". Local databases are disposable, so the recovery is `prisma migrate reset`. Now documented in the PR description.
3. **P3009 needs a manual resolve** — required under either approach, not avoided by a follow-up. Here it is `migrate resolve --rolled-back` (accurate: the migration is transactional and aborted on the first `ALTER`, so nothing persisted). With a follow-up you would instead need `--applied`, asserting a migration ran when it did nothing, and the follow-up would then have to detect and conditionally perform the conversion. That is strictly more fragile.

Leaving this thread open rather than resolving it, since it's a design call and I'd rather a human confirm the trade than have me close my own disagreement.

ntindle (dismissed):

autogpt_platform/backend/migrations/20260805120000_onboarding_step_to_string/migration.sql
```diff
@@ -11,6 +11,60 @@
...
+    LOOP
+        RAISE NOTICE 'dropping dependent relation %.% (recreate with: poetry run analytics-views)',
+                     v.schema_name, v.rel_name;
+        IF v.relkind = 'm' THEN
+            EXECUTE format('DROP MATERIALIZED VIEW IF EXISTS %I.%I CASCADE',
+                           v.schema_name, v.rel_name);
+        ELSE
+            EXECUTE format('DROP VIEW IF EXISTS %I.%I CASCADE',
+                           v.schema_name, v.rel_name);
+        END IF;
```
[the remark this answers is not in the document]
kcze: Addressed in af9aa4ba6a, though partly by removing the need rather than by logging.

The migration now **recreates the two tracked analytics views itself** (definitions copied from `analytics/queries/*.sql` as of this commit, matching what `generate_views.py` emits — `WITH (security_invoker = false)` plus the `analytics_readonly` grant), so the common case no longer depends on anyone re-running the script.

For the two classes you identified that genuinely cannot come back — materialized dependents, and any dependent with no matching file in `analytics/queries/` — the migration now emits a `RAISE WARNING` naming each one and stating it must be rebuilt manually.

One honest caveat on that: **Prisma does not surface server notices from `migrate deploy`**, so those warnings will not appear in the deploy log. Logging alone is therefore not a reliable mechanism here, and I have not pretended otherwise. Instead the PR description carries a pre-flight query so an operator can enumerate the full closure *before* deploying:

```sql
SELECT n.nspname, c.relname, c.relkind
FROM pg_class c JOIN pg_namespace n ON n.oid = c.relnamespace
WHERE c.relkind IN ('v','m') AND EXISTS (
  SELECT 1 FROM pg_depend d JOIN pg_rewrite r ON r.oid = d.objid
  WHERE d.classid = 'pg_rewrite'::regclass AND r.ev_class = c.oid
    AND d.refobjid = '"UserOnboarding"'::regclass AND d.refobjsubid > 0);
```

I did not adopt the recursive closure for discovery. `CASCADE` already removes transitive dependents correctly, and the recursive CTE would only change which names get *reported* — which, given notices are swallowed, buys nothing over the pre-flight query. Happy to add it if you'd rather have the closure enumerated in the SQL itself.

autogpt_platform/backend/migrations/20260805120000_onboarding_step_to_string/migration.sql
```diff
@@ -27,6 +83,116 @@ ALTER TABLE "UserOnboarding" ALTER COLUMN "completedSteps" SET DEFAULT '{}';
...
+        EXECUTE 'GRANT SELECT ON ALL TABLES IN SCHEMA analytics TO analytics_readonly';
+    END IF;
+
+    -- Anything dropped that this migration cannot rebuild.
+    FOR v IN
+        SELECT * FROM _onboarding_dropped_relations
+        WHERE NOT (schema_name = 'analytics'
+                   AND rel_name IN ('user_onboarding', 'user_onboarding_funnel')
+                   AND relkind = 'v')
+    LOOP
```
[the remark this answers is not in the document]
kcze: Confirmed and fixed in ad392f7b09 — good catch, this was a real bug.

`analytics.user_onboarding_integration` reads `"userId"`, `"integrations"` and `"createdAt"` — **none** of the three columns being retyped. The probe matched `d.refobjsubid > 0` (any column of `UserOnboarding`), so it dropped that view, while recreation only covers `user_onboarding` and `user_onboarding_funnel`. It would have vanished.

Rather than adding it to the recreation list, I narrowed the probe, which is the more correct fix: Postgres blocks the `ALTER` **only** for views depending on the columns actually being altered, so `refobjsubid` is now constrained to the attnums of `completedSteps`, `notified` and `rewardedFor`. The integration view is therefore never dropped at all — no drop, no restore, no window where it is missing — and the blast radius shrinks for any other view reading unrelated columns.

Verified with all three analytics views present: the migration applies, `user_onboarding_integration` is untouched and still returns its rows, and the other two are rebuilt with the renamed steps and `analytics_readonly` SELECT intact.

The pre-flight query in the PR description has been updated to the same narrowed predicate so it reports exactly what the migration will drop.

autogpt_platform/backend/migrations/20260805120000_onboarding_step_to_string/migration.sql
kcze: Closing this out: the PR author has confirmed the approach — the broken migration should be fixed in place rather than left for a follow-up to work around.

The deciding factor is that **production has never applied this migration**. `master`'s newest is `20260802120000`; `20260805120000` exists only on `dev`. Leaving the broken file and adding a follow-up means the next prod deploy reaches the original first and fails on prod's analytics views exactly as dev did — the follow-up would never run.

Consequence (3) is also now much cheaper than when this was raised. The `resolve_rolled_back` input on `platform-autogpt-deploy-dev.yaml` runs `prisma migrate resolve --rolled-back` from the deploy runner with the CI secrets, so recovery is a workflow dispatch rather than an undocumented manual database connection:

```bash
gh workflow run platform-autogpt-deploy-dev.yaml \\
  -f git_ref=dev \\
  -f resolve_rolled_back=20260805120000_onboarding_step_to_string
```

It is documented in the PR description, along with the ordering constraint (merge first, or the workflow re-applies the broken version) and the `prisma migrate reset` needed on local databases that already applied it — consequence (2), which remains the real cost of this approach.

autogpt_platform/backend/migrations/20260805120000_onboarding_step_to_string/migration.sql
```diff
@@ -11,6 +11,71 @@
...
+-- a funnel reporting 0 for ONBOARDING_COMPLETE and LIBRARY_RUN_AGENT.
+--
+-- Only views that were present are recreated, so a database without them
+-- (local, CI) is unaffected. Anything dropped that is not recreated here --
+-- a materialized view, or a dependent not tracked in this repo -- is reported
+-- via RAISE WARNING so the operator knows to rebuild it.
+
+-- Fail fast rather than queue behind a long-running BI query holding a lock on
+-- these views; the deploy can retry.
+SET LOCAL lock_timeout = '30s';
```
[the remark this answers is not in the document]
kcze: Checked this functionally rather than by reasoning, and the finding does not hold for Prisma — `SET LOCAL` **is** in effect.

Test: held a conflicting `ACCESS SHARE` lock on `platform."UserOnboarding"` in an open transaction for 120s, then ran `prisma migrate deploy`.

```
exit=1 elapsed=33s
Database error code: 55P03
ERROR: canceling statement due to lock timeout
```

It aborted at 33s — the 30s timeout plus startup. If `SET LOCAL` were being silently ignored it would have blocked for the lock holder's full 120s. Prisma Migrate wraps each migration file in a transaction on Postgres, so the setting applies.

That is also consistent with the incident this PR fixes: when the original migration failed on the dev deploy it left **nothing** behind — the earlier `ALTER ... DROP DEFAULT` statements were rolled back along with the failing one. That is only true under an enclosing transaction, and it is precisely why `migrate resolve --rolled-back` (rather than `--applied`) is the correct recovery.

I have **not** applied the suggested fix, because adding explicit `BEGIN;`/`COMMIT;` here would be actively harmful: the `BEGIN` would nest inside Prisma's own transaction (Postgres warns "there is already a transaction in progress" and ignores it), while the `COMMIT` would close Prisma's transaction early — so a later failure in the file could no longer roll back, and a partially-applied migration becomes possible. That is the failure mode this PR exists to clean up.

ntindle (approved):

PR #13869, opened 2026-08-10
feat(backend): status counts and duration totals in execution cost summary

Abhi1992002:
Part of [SECRT-2523](https://linear.app/autogpt/issue/SECRT-2523) — Home surface v1 (Home & Briefing).

### Why / What / How

**Why:** `get_user_cost_summary()` is the shared aggregation behind the usage/cost views, but it only returns money — total cents, billable runs, failed spend. Any consumer that wants an *activity* chart (runs over time, split by outcome) or an "avg run duration" stat has to issue a second query over the same `AgentGraphExecution` rows. The data is already in the scanned rows; we just weren't projecting it.

**What:** Adds per-status run counts and duration aggregates to the summary — both at the window level (`UserExecutionCostSummary`) and per day (`UserDailyCost`) — so a stacked activity chart can be built from the existing single call.

**How:** Extends the two existing aggregate queries (totals + daily) with `COUNT(*) FILTER (...)` clauses and a `SUM((stats->>'walltime')::numeric)`. No new query, no new round trip, same `WHERE` clause, so the `@@index([userId, isDeleted, createdAt])` hit is unchanged. Rows map into the models with the same `int()` / `float()` + `or 0` guards used for the existing columns.

Purely additive: every new field has a default, so existing consumers and the existing fields' behavior are untouched.

Two things worth calling out, both from review:

- **Duration reads `walltime`, not `duration`.** `update_graph_execution_stats()` persists `GraphExecutionStats.model_dump()`, whose field is `walltime` — nothing writes a `duration` key. The first cut of this PR aggregated `stats->>'duration'`, which would have been `0` for every real run, and the pre-existing `top_runs.duration_seconds` projection had the same bug. Both now read `walltime`, and the test fixture writes the real key so the assertions actually bite.
- **`total_duration_seconds` ships with its own denominator.** `run_count` includes QUEUED/RUNNING rows and rows with no stats, which `SUM` skips — dividing by it biases "avg run duration" downward. `duration_run_count` counts exactly the rows the sum draws from (same idea as `walltime_count` in `backend/data/generate_data.py`).

### Changes 🏗️

- `UserDailyCost`: `success_count`, `failed_count`, `review_count` (default `0`).
- `UserExecutionCostSummary`: `success_run_count`, `failed_run_count`, `review_run_count`, `total_duration_seconds`, `duration_run_count` (default `0`).
- Totals SQL: three `COUNT(*) FILTER` clauses (`COMPLETED`; `FAILED`/`TERMINATED`; `REVIEW`), `COALESCE(SUM((stats->>'walltime')::numeric), 0)::float`, and `COUNT(*) FILTER (WHERE (stats->>'walltime') IS NOT NULL)`.
- Daily SQL: the same three per-status `FILTER` counts, per bucket.
- Top-runs SQL: `duration_seconds` now projects `stats->>'walltime'` (pre-existing bug fix).
- The four aggregate queries moved into `_fetch_totals` / `_fetch_by_agent` / `_fetch_top_runs` / `_fetch_daily`, each owning its SQL and row→model mapping, with the shared predicate hoisted to `_BASE_WHERE`. `get_user_cost_summary` is window validation + `asyncio.gather` + assembly.
- `openapi.json` re-exported — the new response fields are optional with defaults, so no frontend change is required.
- Tests extended in `execution_cost_summary_test.py`: the fixture writes `walltime` (and can omit it entirely), a `REVIEW` run and a stats-less `RUNNING` run added to the mixed-status fixture (counts, duration sum, and duration denominator asserted), `FAILED` + `REVIEW` runs added to the daily-bucket fixture (per-day split asserted), and the empty-window test asserts the new fields are zeroed.

#### For code changes:
- [x] I have made a test plan
  - [x] `poetry run pytest backend/data/execution_cost_summary_test.py` — 5 passed against real Postgres (mixed statuses produce the right per-status counts, duration sum, and duration denominator; the RUNNING run stays out of both; empty window returns zeros; existing cost/daily/dry-run assertions still hold)
  - [x] `poetry run pytest backend/api/features/v1_test.py -k cost` — 5 passed, the `/api/v1` cost-summary route is unaffected
  - [x] `poetry run format && poetry run lint` clean (ruff, isort, black, pyright)
  - [x] `poetry run export-api-schema` + prettier produces a 5-line diff (the new field only)

### DIFF
--- a/autogpt_platform/backend/backend/data/execution_cost_summary.py
+++ b/autogpt_platform/backend/backend/data/execution_cost_summary.py
@@ -29,6 +29,9 @@ class UserDailyCost(BaseModel):
     date: date
     cost_cents: int
     run_count: int
+    success_count: int = 0
+    failed_count: int = 0
+    review_count: int = 0
 
 
 class UserExecutionCostSummary(BaseModel):
@@ -36,6 +39,10 @@ class UserExecutionCostSummary(BaseModel):
     run_count: int
     billable_run_count: int
     failed_cost_cents: int
+    success_run_count: int = 0
+    failed_run_count: int = 0
+    review_run_count: int = 0
+    total_duration_seconds: float = 0
     by_agent: list[UserAgentCostRollup]
     top_runs: list[UserTopRun]
     daily: list[UserDailyCost]
@@ -91,7 +98,15 @@ async def get_user_cost_summary(
             "    AS billable_run_count,"
             "  COALESCE(SUM(CASE WHEN \"executionStatus\" IN ('FAILED', 'TERMINATED')"
             "    THEN (stats->>'cost')::numeric ELSE 0 END), 0)::bigint"
-            "    AS failed_cost_cents"
+            "    AS failed_cost_cents,"
+            "  COUNT(*) FILTER (WHERE \"executionStatus\" = 'COMPLETED')::bigint"
+            "    AS success_run_count,"
+            "  COUNT(*) FILTER (WHERE \"executionStatus\" IN ('FAILED', 'TERMINATED'))::bigint"
+            "    AS failed_run_count,"
+            "  COUNT(*) FILTER (WHERE \"executionStatus\" = 'REVIEW')::bigint"
+            "    AS review_run_count,"
+            "  COALESCE(SUM((stats->>'duration')::numeric), 0)::float"
+            "    AS total_duration_seconds"
             ' FROM {schema_prefix}"AgentGraphExecution"'
             f" WHERE {base_where}",
             *params,
@@ -129,7 +144,13 @@ async def get_user_cost_summary(
             "SELECT"
             '  ("createdAt" AT TIME ZONE \'UTC\')::date AS "date",'
             "  COALESCE(SUM((stats->>'cost')::numeric), 0)::bigint AS cost_cents,"
-            "  COUNT(*)::bigint AS run_count"
+            "  COUNT(*)::bigint AS run_count,"
+            "  COUNT(*) FILTER (WHERE \"executionStatus\" = 'COMPLETED')::bigint"
+            "    AS success_count,"
+            "  COUNT(*) FILTER (WHERE \"executionStatus\" IN ('FAILED', 'TERMINATED'))::bigint"
+            "    AS failed_count,"
+            "  COUNT(*) FILTER (WHERE \"executionStatus\" = 'REVIEW')::bigint"
+            "    AS review_count"
             ' FROM {schema_prefix}"AgentGraphExecution"'
             f" WHERE {base_where}"
             "  GROUP BY (\"createdAt\" AT TIME ZONE 'UTC')::date"
@@ -144,6 +165,10 @@ async def get_user_cost_summary(
         run_count=int(totals.get("run_count") or 0),
         billable_run_count=int(totals.get("billable_run_count") or 0),
         failed_cost_cents=int(totals.get("failed_cost_cents") or 0),
+        success_run_count=int(totals.get("success_run_count") or 0),
+        failed_run_count=int(totals.get("failed_run_count") or 0),
+        review_run_count=int(totals.get("review_run_count") or 0),
+        total_duration_seconds=float(totals.get("total_duration_seconds") or 0),
         by_agent=[
             UserAgentCostRollup(
                 graph_id=r["graph_id"],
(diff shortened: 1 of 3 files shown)


autogpt_platform/backend/backend/data/execution_cost_summary.py
```diff
@@ -91,7 +98,15 @@ async def get_user_cost_summary(
...
             "    THEN (stats->>'cost')::numeric ELSE 0 END), 0)::bigint"
-            "    AS failed_cost_cents"
+            "    AS failed_cost_cents,"
+            "  COUNT(*) FILTER (WHERE \"executionStatus\" = 'COMPLETED')::bigint"
+            "    AS success_run_count,"
+            "  COUNT(*) FILTER (WHERE \"executionStatus\" IN ('FAILED', 'TERMINATED'))::bigint"
+            "    AS failed_run_count,"
+            "  COUNT(*) FILTER (WHERE \"executionStatus\" = 'REVIEW')::bigint"
+            "    AS review_run_count,"
+            "  COALESCE(SUM((stats->>'duration')::numeric), 0)::float"
```
Abhi1992002: 🤖 🔴 **Blocker**: This reads the wrong persisted stats key, so `total_duration_seconds` will be `0` for normal executor-written runs. `update_graph_execution_stats()` stores `GraphExecutionStats.model_dump()`, whose field is `walltime` (`backend/data/model.py`), not `duration`; the test helper currently masks this by constructing synthetic stats with a `duration` key. Please aggregate the persisted `walltime` key here and update the fixture to write `walltime` (and ideally fix the pre-existing top-runs projection below for the same reason).

autogpt_platform/backend/backend/data/execution_cost_summary.py
```diff
@@ -29,13 +29,20 @@ class UserDailyCost(BaseModel):
...
 
 class UserExecutionCostSummary(BaseModel):
     total_cents: int
     run_count: int
     billable_run_count: int
     failed_cost_cents: int
+    success_run_count: int = 0
+    failed_run_count: int = 0
+    review_run_count: int = 0
+    total_duration_seconds: float = 0
```
Abhi1992002, line 45: 🤖 🟠 **Should Fix**: The stated average-duration use case still lacks a valid denominator. `run_count` includes QUEUED/RUNNING/INCOMPLETE executions and rows with no stats, while `SUM` ignores a missing walltime; dividing this total by `run_count` therefore biases the average downward. The existing summary implementation in `backend/data/generate_data.py` tracks `walltime_count` for exactly this reason. Please expose a duration-bearing run count (or the average itself) and cover a row with missing stats in the tests.

autogpt_platform/backend/backend/data/execution_cost_summary.py
Abhi1992002: Fixed in 5076e00fc0. Confirmed: `update_graph_execution_stats()` writes `GraphExecutionStats.model_dump()`, whose field is `walltime` — nothing anywhere persists a `duration` key, so `total_duration_seconds` was 0 for every executor-written run.

- Totals query now sums `stats->>'walltime'`.
- Fixed the pre-existing `top_runs.duration_seconds` projection for the same reason (it had the same bug).
- The fixture writes `walltime` instead of the synthetic `duration`, and `test_aggregates_by_agent_and_top_runs` now asserts `top_runs[0].duration_seconds == 1.0` so the projection is locked too.

`poetry run pytest backend/data/execution_cost_summary_test.py` — 5 passed against real Postgres (the fixture change makes these assertions genuinely fail on the old code).

autogpt_platform/backend/backend/data/execution_cost_summary.py
Abhi1992002: Fixed in 5076e00fc0. Added `duration_run_count` to `UserExecutionCostSummary`:

```sql
COUNT(*) FILTER (WHERE (stats->>'walltime') IS NOT NULL)::bigint AS duration_run_count
```

Same idea as `walltime_count` in `backend/data/generate_data.py` — the denominator now matches exactly the rows the `SUM` numerator draws from, so QUEUED/RUNNING rows and rows with no stats no longer drag the average down. Exposing the count rather than the average keeps the summary a pure aggregate and lets the caller format it.

Covered: `test_aggregates_by_agent_and_top_runs` adds a RUNNING run created with `duration=None` (stats carry no `walltime` at all) and asserts `run_count == 8` while `total_duration_seconds == 7.0` and `duration_run_count == 7`. `test_empty_window_returns_zeroed_summary` asserts it zeroes.

`openapi.json` re-exported; the field is optional with a `0` default, so no frontend change.

PR #13992, opened 2026-08-10
fix(blocks): handle surrogate pairs in AGPT_VARIABLES — ExecuteCodeBl…

### LINKED ISSUE #13551: ExecuteCodeBlock: UnicodeEncodeError when data contains emoji (surrogate pairs in AGPT_VARIABLES env var)

hariomlohardev:
Use `json.dumps(ensure_ascii=True)` so emoji/surrogates become ASCII escapes, env var always strict-UTF-8 encodable. Fixes #13551.

### Why / What / How

**Why:** `ExecuteCodeBlock` crashes before user code when an upstream block (e.g. `NotionReadDatabaseBlock`) returns emoji like `"Holidays 🎅🏻"`. `build_variable_injection()` did `json.dumps(variables)` (`ensure_ascii=False` by default is `True`, but the intent was raw) and then `os.environ["AGPT_VARIABLES"]=serialized`. With lone surrogates (`\ud83c` from Notion UTF-16) the env var contains real surrogate code points. Python 3.13's `os.environ.__setitem__` is strict UTF-8 and raises `UnicodeEncodeError: 'utf-8' codec can't encode character '\ud83c' ... surrogates not allowed` at `AGPT_VARIABLES` set. No user code runs. Repro at ~25KB payload, but any emoji with skin-tone modifier can trigger it. Workaround was filtering upstream data.

**What:** Make the `AGPT_VARIABLES` payload always strict-UTF-8 encodable and preserve data. One file, 2 hunks in `autogpt_platform/backend/backend/blocks/code_executor_helpers.py`: (1) serialize with `ensure_ascii=True`, (2) measure payload size with `encode("utf-8", "surrogatepass")` for defensiveness.

**How:** `json.dumps(variables, ensure_ascii=True)` escapes all non-ASCII (including emoji and lone surrogates) as ASCII `\uXXXX` (`🎅` → `🎅`). The env var is then pure ASCII, always `encode("utf-8")` safe, and the sandbox prefix `globals().update(json.loads(os.environ["AGPT_VARIABLES"]))` decodes back to the original emoji via `json.loads` (verified round-trip). `surrogatepass` in the size check measures the actual wire size correctly after escaping (emoji inflates 4B → 6B per char, so limit correctly rejects `5470×🎅` at 65652B vs raw 21892B). No API change, no config change, preserves `MAX_VARIABLES_PAYLOAD_BYTES=64KB` semantics.

### Changes 🏗️

- `autogpt_platform/backend/backend/blocks/code_executor_helpers.py`
  - `build_variable_injection()`: `json.dumps(variables)` → `json.dumps(variables, ensure_ascii=True)` + comment explaining surrogate crash (Python 3.13)
  - Payload size check: `encode("utf-8")` → `encode("utf-8", "surrogatepass")` (redundant after `ensure_ascii` but guards future `ensure_ascii=False` regression)
- No other `AGPT_VARIABLES` write sites (grep confirms only this helper + `code_executor.py` delegating to it)

#### For code changes:
- [x] I have made a test plan
  - [x] Repro with `{"holidays":"Holidays 🎅🏻", "surrogate":"\ud83c", "nested":{"key":"value 🎉"}}` — `is_ascii` true, `strict utf-8` PASS (old `ensure_ascii=False` would raise), `json.loads` round-trip `Holidays 🎅🏻` PASS (via `/tmp/repro_13551_v2.py`)
  - [x] Lone high `\ud83c`, low `\udc00`, mixed, `hello \ud83c world \udc00` — all ASCII, strict PASS
  - [x] ASCII regression `{"x":42,"name":"Blake"}` still `== json.dumps(..., ensure_ascii=True)` — matches existing `TestBuildVariableInjection` expectations
  - [x] Empty `{}` → `envs=={}` / `prefix==""`, JS `{"x":"🎉"}` → ASCII, `BASH/R/JAVA` still `UnsupportedLanguageError`
  - [x] Oversize `x*(64KB+1)` correctly raises `ValueError: too large`; emoji inflation `5460×🎅` (65532B) allows, `5470×🎅` (65652B) rejects — after escaping, not raw
  - [x] Existing helper replica suite 9/9 PASS (`/tmp/run_helper_tests_direct.py`): empty, python/js serialize, malicious, unsupported, non-serializable, invalid keys (5), oversize, emoji roundtrip, surrogate ascii
  - [x] `uv run pytest` target: `code_executor_test.py::TestBuildVariableInjection` 9 tests expected PASS (ASCII behavior unchanged); `TestExecuteCodeBlockRun` 5 tests expect `json.dumps(..., ensure_ascii=True)` for ASCII payloads — verified via direct import (full suite needs `e2b`, `jsonref`, `pytest-asyncio` — not run in this env, but helper logic isolated)

<details>

</details>

- [x] I have included a list of my configuration changes in the PR description (under **Changes**) — No config changes; env var wire format stays JSON, only ASCII-escaped. Both files already compatible.

<details>

  - Changing ports
</details>


daltino (commented):
The `ensure_ascii=True` fix is correct and addresses the root cause cleanly. One thing worth thinking through: the second change swapping `serialized.encode("utf-8")` for `serialized.encode("utf-8", "surrogatepass")` is now a bit redundant — since `ensure_ascii=True` guarantees the output is pure ASCII, there can't be any surrogates left in `serialized` at that point, so `surrogatepass` will never actually trigger. It's not harmful, but it slightly obscures the invariant the first change establishes. It might be cleaner to just leave the encode call as-is, or add a brief comment acknowledging why surrogates are impossible here after the `ensure_ascii` step. Also worth confirming there's a test covering the Notion emoji/surrogate path end-to-end, since the bug reportedly reproduces around 25K chars — a unit test that injects a lone surrogate like `"\ud83c"` into `build_variable_injection` would make sure this doesn't regress silently.

PR #14001, opened 2026-08-11
feat(frontend/copilot): composer + menu opens the connect-service dialog

Abhi1992002:
### Why / What / How

**Why:** The composer `+` menu in CoPilot had an item labelled **"Integrations"** that opened the *Third Party Integrations* panel — a list of credentials the user has **already** connected. That's the wrong destination for the intent behind the click. Someone opening that menu mid-conversation wants to *add* a service so AutoPilot can use it, not audit what they already have. The action they actually wanted sat two clicks deep, behind a "Connect Service" button inside the panel.

There was a structural problem behind that button too. `ConnectServiceDialog` is rendered inside `IntegrationsPanel`, so opening it from the CoPilot modal meant a Radix Dialog nested inside another Radix Dialog — sharing `z-50` for both overlay and content (`components/styles.ts:5-12`), inside the outer `DismissableLayer` and `RemoveScroll`. The inner dialog only landed on top by portal insertion order. On `/settings/integrations` the same panel renders at top level with no outer dialog, which is why that path never showed the problem and why the nested path had no test coverage.

**What:** The `+` menu now opens **Connect a service** directly, as a top-level dialog.

**How:** Added a `connect` entry to `COPILOT_MODALS`, so the dialog is driven by the same `?modal=` query state as skills/scheduled and is deep-linkable. `ConnectServiceDialog` already takes `open` / `onOpenChange`, so it drops into `CopilotModals` as a sibling of the other dialogs with no changes to the component itself.

Rendering it as a sibling rather than nested is the point: it is no longer inside the integrations dialog's portal, so the stacking and dismissal ambiguity described above cannot occur on this path.

The existing `integrations` modal is left wired, so `/copilot?modal=integrations` still opens the credentials panel for anyone with that link.

### Changes 🏗️

**`useCopilotModal.ts`**
- Added `"connect"` to `COPILOT_MODALS` alongside `integrations` / `skills` / `scheduled`

**`CopilotModals.tsx`**
- Renders `<ConnectServiceDialog>` at the top level, gated on `modal === "connect"`, reusing the shared `handleOpenChange`

**`ComposerPlusMenu.tsx`**
- The menu item now calls `openModal("connect")` and is relabelled **"Connect service"**. The old "Integrations" label no longer described what the item opens; leaving it would have been a worse lie than before.

**Tests**
- `CopilotModals.test.tsx` — new case: opening `connect` shows "Connect a service" and does *not* render the "Third Party Integrations" panel
- `ComposerPlusMenu.test.tsx` — updated the menu-order fixtures and the modal-target assertion to the new label and `connect` target

### Behaviour note for reviewers

The credentials list is no longer reachable from the CoPilot `+` menu. That is the intended change, but worth naming explicitly:

- Still available at `/settings/integrations`
- Still available via the `/copilot?modal=integrations` deep link

If we want it back in-product from CoPilot, the natural follow-up is a "Manage connections" link inside the Connect dialog rather than a second menu item.

#### For code changes:
- [x] I have made a test plan
  - [x] `pnpm test:unit` — `CopilotModals.test.tsx` + `ComposerPlusMenu.test.tsx` pass (12 cases)
  - [x] `pnpm types` clean, `pnpm format` / `pnpm lint` clean
  - [ ] `/copilot` → `+` → menu reads: Attach file, Connect service, Skills, Scheduled
  - [ ] Click **Connect service** → "Connect a service" dialog opens, URL becomes `?modal=connect`
  - [ ] Provider list renders, search filters it, clicking a provider shows the detail view with OAuth / API-key tabs
  - [ ] Completing a connection closes the dialog and clears the `modal` param
  - [ ] Escape and outside-click both dismiss cleanly, with no stuck overlay or lost scroll
  - [ ] `/copilot?modal=integrations` still opens the credentials panel
  - [ ] Skills and Scheduled modals still open from the same menu
  - [ ] Below 1024px the dialog renders as a drawer and still behaves

### DIFF
--- a/autogpt_platform/frontend/src/app/(platform)/copilot/components/CopilotModals/CopilotModals.tsx
+++ b/autogpt_platform/frontend/src/app/(platform)/copilot/components/CopilotModals/CopilotModals.tsx
@@ -1,5 +1,6 @@
 "use client";
 
+import { ConnectServiceDialog } from "@/components/contextual/IntegrationsPanel/components/ConnectServiceDialog/ConnectServiceDialog";
 import { IntegrationsPanel } from "@/components/contextual/IntegrationsPanel/IntegrationsPanel";
 import { SchedulesPanel } from "@/components/contextual/SchedulesPanel/SchedulesPanel";
 import { SkillsPanel } from "@/components/contextual/SkillsPanel/SkillsPanel";
@@ -57,6 +58,11 @@ export function CopilotModals() {
           <IntegrationsPanel withHeading={false} />
         </Dialog.Content>
       </Dialog>
+
+      <ConnectServiceDialog
+        open={modal === "connect"}
+        onOpenChange={handleOpenChange}
+      />
     </>
   );
 }
--- a/autogpt_platform/frontend/src/app/(platform)/copilot/components/CopilotModals/__tests__/CopilotModals.test.tsx
+++ b/autogpt_platform/frontend/src/app/(platform)/copilot/components/CopilotModals/__tests__/CopilotModals.test.tsx
@@ -1,11 +1,20 @@
-import { getGetV1ListCredentialsMockHandler } from "@/app/api/__generated__/endpoints/integrations/integrations.msw";
+import {
+  getGetV1ListCredentialsMockHandler,
+  getGetV1ListProvidersMockHandler,
+} from "@/app/api/__generated__/endpoints/integrations/integrations.msw";
 import {
   getGetV1ListExecutionSchedulesForAUserMockHandler,
   getListCopilotFollowupSchedulesMockHandler,
 } from "@/app/api/__generated__/endpoints/schedules/schedules.msw";
 import { getListCopilotSkillsMockHandler } from "@/app/api/__generated__/endpoints/skills/skills.msw";
 import { server } from "@/mocks/mock-server";
-import { fireEvent, render, screen } from "@/tests/integrations/test-utils";
+import {
+  fireEvent,
+  render,
+  screen,
+  within,
+} from "@/tests/integrations/test-utils";
+import { NuqsTestingAdapter } from "nuqs/adapters/testing";
 import { afterEach, beforeEach, describe, expect, test, vi } from "vitest";
 import { useCopilotUIStore } from "../../../store";
 import { useCopilotModal } from "../../../useCopilotModal";
@@ -13,14 +22,29 @@ import { CopilotModals } from "../CopilotModals";
 
 function Harness() {
   const { openModal } = useCopilotModal();
+
+  function openSkills() {
+    openModal("skills");
+  }
+
+  function openScheduled() {
+    openModal("scheduled");
+  }
+
+  function openIntegrations() {
+    openModal("integrations");
+  }
+
+  function openConnect() {
+    openModal("connect");
+  }
+
   return (
     <>
-      <button onClick={() => openModal("skills")}>open-skills</button>
-      <button onClick={() => openModal("scheduled")}>open-scheduled</button>
-      <button onClick={() => openModal("integrations")}>
-        open-integrations
-      </button>
-      <button onClick={() => openModal("connect")}>open-connect</button>
+      <button onClick={openSkills}>open-skills</button>
+      <button onClick={openScheduled}>open-scheduled</button>
+      <button onClick={openIntegrations}>open-integrations</button>
+      <button onClick={openConnect}>open-connect</button>
       <CopilotModals />
     </>
   );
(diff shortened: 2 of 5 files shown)


autogpt_platform/frontend/src/app/(platform)/copilot/components/CopilotModals/__tests__/CopilotModals.test.tsx
```diff
@@ -96,4 +97,12 @@ describe("CopilotModals", () => {
       (await screen.findAllByText("Connect Service")).length,
     ).toBeGreaterThan(0);
   });
+
+  test("opens the connect dialog directly, without the credentials list", async () => {
+    render(<Harness />);
+    fireEvent.click(screen.getByText("open-connect"));
+
+    expect(await screen.findByText("Connect a service")).toBeDefined();
+    expect(screen.queryByText("Third Party Integrations")).toBeNull();
```
kcze: 🤖 🟠 **Should Fix**: this assertion is vacuous for the behaviour the test name claims. `"Third Party Integrations"` is the *integrations dialog's title* (`CopilotModals.tsx:55`), not anything `IntegrationsList` renders — and that dialog was never opened in this test, so the assertion passes regardless. If a regression made `modal === "connect"` also render the credentials list, the rows and empty-state would appear but the title would not, and this stays green.

The positive half has the same shape: `findByText("Connect a service")` matches the `Dialog` `title` prop, which renders whether or not the provider list loaded. The providers request here is served by the global faker handler (`mocks/mock-handlers.ts:65`), never overridden — so if the `enabled: open` gate stopped flipping or the query key changed, the dialog would render `ErrorCard` instead of the list and this test would still pass while users get an erroring dialog.

Asserting on real content — a seeded provider row via `getGetV1ListProvidersMockHandler([...])`, and the `IntegrationsListEmpty` copy for the negative case — closes both. _(flagged by: Claude)_

autogpt_platform/frontend/src/app/(platform)/copilot/components/CopilotModals/CopilotModals.tsx
```diff
@@ -57,6 +58,11 @@ export function CopilotModals() {
           <IntegrationsPanel withHeading={false} />
         </Dialog.Content>
       </Dialog>
+
+      <ConnectServiceDialog
+        open={modal === "connect"}
+        onOpenChange={handleOpenChange}
```
kcze, line 64: 🤖 🟡 **Nice to Have**: the new `onOpenChange` wiring has no test. The one added case asserts the title renders on open and nothing about closing, so nothing covers `handleOpenChange` clearing the query param, and nothing covers a `?modal=connect` deep link on mount.

Mis-wiring this (passing `closeModal` directly, which would receive `open` as its argument) leaves `?modal=connect` stuck in the URL after Escape — the dialog reopens on every refresh and can't be dismissed — with the suite still green. A `fireEvent.click(closeButton)` + assert-param-cleared case, plus a mounted-with-`?modal=connect` case, are both cheap here. _(flagged by: Claude)_

autogpt_platform/frontend/src/app/(platform)/copilot/components/CopilotModals/__tests__/CopilotModals.test.tsx
Abhi1992002: Fixed in 029c9f9e1c.

You were right that both halves were vacuous — I verified by mutation rather than by reading. Making the `integrations` dialog also open on `modal === "connect"` (`isOpen: modal === "integrations" || modal === "connect"`) left the old test green.

The test now seeds a provider and asserts on its description, and the negative half asserts on `IntegrationsListEmpty` copy, which is what the credentials list actually renders under the `getGetV1ListCredentialsMockHandler([])` already set in `beforeEach`:

```tsx
server.use(
  getGetV1ListProvidersMockHandler([
    { name: "github", description: "Issues and PRs", supported_auth_types: ["oauth2", "api_key"] },
  ]),
);
const dialog = await screen.findByRole("dialog");
expect(await within(dialog).findByText("Issues and PRs")).toBeDefined();
expect(screen.queryByText("No integration connected")).toBeNull();
```

Same mutation now fails all three connect cases. The positive half also no longer passes if the `enabled: open` gate or the query key breaks, since `"Issues and PRs"` only renders once the providers request resolves into `ListView`.

autogpt_platform/frontend/src/app/(platform)/copilot/components/CopilotModals/CopilotModals.tsx
Abhi1992002: Fixed in 029c9f9e1c — both cases added.

One correction on the failure mode you named: passing `closeModal` directly is actually *not* the bug you describe. `closeModal()` declares zero parameters, so the `open` boolean is discarded and it still does `setModal(null)`. I made that exact mutation locally and all 8 cases stayed green — because the behaviour is identical.

So I aimed the test at the wiring failure that does break users: `onOpenChange` not clearing state at all (`onOpenChange={() => {}}`). Under that mutation the new case fails, which is what we want.

```tsx
test("renders the connect dialog from a ?modal=connect deep link", ...)  // mount with searchParams
test("closing the connect dialog clears the modal query param", ...)     // Close click -> onUrlUpdate spy
```

Both wrap `<Harness />` in a nested `NuqsTestingAdapter` (`searchParams` + `onUrlUpdate`) — the shared `test-utils` wrapper does not expose those props, and the inner adapter wins on context, so this stays local to the file rather than changing shared test infra.

The URL assertion needed its own `waitFor`: nuqs flushes the update one tick after the dialog unmounts optimistically, so asserting straight after the dialog disappears was itself a false-green.

kcze (commented):
🤖 Re-reviewed at `029c9f9` — both earlier findings are resolved, and no new ones survived verification. The seeded `getGetV1ListProvidersMockHandler` assertion now fails if the providers query errors, stays loading, or stops rendering `ListView`, and the deep-link and close-clears-the-param cases both fail under the wirings they're meant to catch.

kcze (approved):

PR #14003, opened 2026-08-11
feat(platform): replace the email system with the Briefing/Alert/Verdict/Ops design

Torantulino:
### Why / What / How

**Why.** The two emails the platform sends fail in opposite directions. The run report dumps every agent's raw output into the inbox, which collapses the moment a user has real volume — and a binary output can't be inlined at all. The weekly summary goes the other way and reduces everything to operator metrics ("Total Execution Time: 101.1 seconds"), numbers that describe the machine rather than what it did for the person. Both share a deeper problem: neither contains a single link back to the platform. The only destination offered is Discord.

Underneath that, the notification surface defines **12 types across 8 templates** — four of the twelve have no template at all and would raise `FileNotFoundError` if they ever fired, and three of those four have no producer either. Separately, `queue_notification` short-circuits when `app_env == PRODUCTION` and returns success *without publishing*, so in production none of this sends anything today.

**What.** Implements the redesign in the [Lifecycle Email Map](https://lifecycle-email-map.pages.dev): the 12-type surface is replaced by **four product-notification families** — Briefing, Alert, Verdict, Ops — plus **six account/billing service messages**, all rendered from the design's own WCAG-AA-audited templates.

The governing rule is that every email must answer exactly one of two questions: *"Do I need to do something?"* or *"What did my agents get done for me?"* If a candidate email answers neither, it isn't sent. Per-run email is retired in the same release, by design.

**How.** Notable design decisions:

- **Aggregation is per *agent*, never per run.** Each finished run is scored for interestingness at completion time (new `AgentGraphExecution.interestingness`), so assembling a Briefing is a cheap ranked read rather than a scan. A user with one agent and a user with two hundred get emails of roughly the same length; what changes is the compression ratio.
- **The Alert engine owns the rules, not the callers.** Emitters raise a *condition*; whether it becomes an email, a coalesced line, or a Briefing attention card is the engine's decision.
- **Nothing actionable is silently dropped.** Anything deduped or capped is marked `DEFERRED` and folded into the next Briefing's attention block.
- **Complete documents, not content fragments.** These replace `base.html.jinja2` *and* the per-type content templates, so they bypass the `bleach.clean()` path — that allow-list has no `<table>`/`<tr>`/`<td>` and would strip the entire layout. Jinja autoescape is the defence for user-supplied values, which is the correct one; `format_email` and the sanitiser config are deleted.
- **Templates are byte-identical** to the design repo's `user_emails/templates/` and `kit/templates/`, which is what the contrast audit covers.

### Changes 🏗️

**The four families** (`backend/notifications/`)

- **Briefing** — scheduled digest, daily/weekly/monthly, delivered at ~07:30 in the user's own timezone (hourly job, timezone-filtered in SQL). Inverted pyramid under a hard length budget: lede, attention, ≤3 highlights, ≤6 ledger rows, volume-knob footer. Never sent empty; a whisper of activity produces the deliberately tiny quiet form rather than a padded digest. Server-side invariants hold: `totals.runs` equals the ledger's sum and `totals.agents_active` equals `len(ledger)`.
- **Alert** — 10-minute debounce, coalescing, no re-alert for the same cause within 24h, hard cap of two emails per day, and cancellation if the condition clears during the debounce window. Full nine-cause catalog.
- **Verdict** — "changes requested", not "rejected". Renders discrete review items as a numbered list when supplied, free-text as the fallback; empty and whitespace-only feedback both handled.
- **Ops** — internal refunds mail. Amount is the hero figure, timestamps are absolute, and the reject path deep-links into the admin panel instead of pasting a Supabase console URL and table-editing instructions. The one family deliberately **without** List-Unsubscribe headers.

**Billing lifecycle** (`backend/notifications/lifecycle.py`, wired in `api/features/v1.py`)

Welcome · payment failed · final notice · cancellation confirmed · cancellation reversed · plan ended. Each on exactly one Stripe signal, each idempotent, each honouring the design's gotchas: first subscription only; `checkout.session.completed` **not** `customer.subscription.created` (both fire at signup — listening to both double-sends); `attempt_count == 1` only, silent through Smart Retries; branch on `next_payment_attempt` for the final notice; the `previous_attributes` false→true flip for cancel/resume rather than every subscription update; branch on `cancellation_details.reason` for the ended copy.

**No trial.** `customer.subscription.trial_will_end` is deliberately not listened for and there is no trial type or handler — the platform has no plans for trials. The design's trial branch survives in `lifecycle.html.j2` as unreachable reference; wiring it up is the whole job if that ever changes.

**Delivery mechanics** — transactional stream separate from marketing mail; three separated senders (`billing@` / `notify@` / `platform@`) each with its own reputation; a plain-text MIME part built from the same data model; one-click List-Unsubscribe on everything except Ops; hero art served from a CDN (Outlook doesn't render inline SVG, Gmail doesn't display data URIs).

**MailerLite** — audience management as queued, retried jobs, never inline in a Stripe webhook. Exactly one owner per transition; the tour→changelog handoff is deliberately left to MailerLite's automation, which is what makes changelog suppression structural rather than a campaign-level exclusion rule someone forgets.

**Preferences** — the per-type checkbox list becomes the volume knob from the Briefing footer: a frequency plus switches for Alerts and Marketplace reviews. The footer's `?f=daily|weekly|monthly|alerts|off` links land on a settings page that applies the choice on load, so it really is one click. Billing messages are service mail and aren't gated by any of it. Both the new settings page and the legacy `/profile/settings` form are migrated.

**Removed** — all 12 old `NotificationType` values, all 8 old templates + `base.html.jinja2`, `TextFormatter.format_email` and the bleach sanitiser config, `handle_agent_run_notif` (per-run email), `generate_data.py` (summary aggregation), the batch/summary queues and their scheduler jobs.

#### Configuration changes

- `.env.default` — adds `BILLING_SENDER_EMAIL`, `PRODUCT_SENDER_EMAIL`, `OPS_SENDER_EMAIL`, `POSTMARK_TRANSACTIONAL_STREAM`, `EMAIL_ASSET_BASE_URL`, `DOCS_BASE_URL`, `DISCORD_INVITE_URL`, `ADMIN_PANEL_BASE_URL`, `MAILERLITE_API_TOKEN`, `MAILERLITE_ONBOARDING_GROUP_ID`, `MAILERLITE_CHANGELOG_GROUP_ID`. All have working defaults; the two MailerLite values and the Postmark token gate their features rather than breaking startup.
- `docker-compose.yml` needs no change — no new services or ports.
- **Migration `20260811120000_replace_email_notification_system` recreates the `NotificationType` enum and therefore truncates `NotificationEvent` and `UserNotificationBatch`.** Payloads queued in the old shape cannot be rendered by any current template. It carries preference intent across before dropping the old columns: weekly-summary-off keeps the Briefing off, and both-balance-notifications-off keeps Alerts off.
- RabbitMQ queues roll to `_v3`; the `_v2` queues are left to drain.

> [!IMPORTANT]
> **This makes notification email actually send in production.** `queue_notification`/`queue_notification_async` previously returned success without publishing when `app_env == PRODUCTION`. That short-circuit is gone — it had to be, or the replacement couldn't work — but it's the single largest operational difference in this PR and deserves a staging soak before it reaches prod.

> [!NOTE]
> `backend/notifications/DEPLOYMENT.md` is the handoff for everything that can't live in this repo: the CDN upload, Postmark streams and sender DNS, **turning off Stripe's own failed-payment emails**, matching the dunning schedule to what the final notice promises, and activating the existing (currently inactive) "Subscription Onboarding — White Glove Tour" automation in MailerLite. It also records what was deliberately left: five alert causes whose platform signals don't exist yet, the admin-panel double-refund guard, and a palette question the design page leaves genuinely ambiguous.

#### For code changes:
- [x] I have made a test plan
  - [x] Every family renders end-to-end — 17 payload variants (busy/clean/quiet/rough Briefing, single and coalesced Alert, both Verdict outcomes, both Ops kinds, all six billing kinds) through the real `renderer.render()`, asserting a complete document, a plain-text twin, and no banned word in any subject
  - [x] User-supplied values are escaped, not executed (`<script>` in an agent name)
  - [x] Ops carries no unsubscribe link or header; every other family does
  - [x] Alert engine unit tests: coalescing by severity, the daily cap deferring into the Briefing, alerts-off still reaching the Briefing, and the CTA verb being the fix
  - [x] Briefing unit tests: never-empty, the totals/ledger invariants, quiet mode, amber-vs-red ledger rows, highlight cap and the gist fallback chain
  - [x] Lifecycle unit tests: every Stripe branch above, plus an assertion that a `trial_will_end` event does nothing
  - [x] `poetry run test` on the backend suite
  - [ ] `pnpm test:unit` and `pnpm types` on the frontend
  - [ ] `prisma migrate deploy` against a staging copy, confirming the preference back-fill lands as intended
  - [ ] End-to-end with a real Postmark sandbox token: trigger one Alert and one Briefing, confirm the plain-text part, the List-Unsubscribe header, and that the footer's `?f=` links change the setting in one click
  - [ ] Stripe CLI: replay `checkout.session.completed`, `invoice.payment_failed` (attempt 1 and attempt 3), the cancel/resume flips, and `customer.subscription.deleted`, confirming one email each and none on replay

> [!WARNING]
> **I could not run the test suites in my environment** — `poetry` and `node_modules` weren't installed. The unchecked boxes above are genuinely unverified, not assumed-passing. What I did verify: all Python compiles, and the render, escaping, header and unit-test assertions were checked by executing the templates and the pure logic directly. The migration, the frontend type-check, and every live-service path still need a real run.


autogpt_platform/backend/backend/data/alerts.py
```diff
@@ -0,0 +1,300 @@
+"""Persistence for live alert conditions.
+
+One row per "the platform is blocked on the human" condition, keyed by a stable
```
ntindle: This comment makes no sense

autogpt_platform/backend/backend/notifications/DEPLOYMENT.md
```diff
@@ -0,0 +1,259 @@
+# Shipping the redesigned email system
+
```
ntindle: Remove this file

autogpt_platform/backend/backend/notifications/boundary_test.py
```diff
@@ -0,0 +1,127 @@
+"""The notification service owns no database connection.
+
+`AppService.lifespan` connects Redis and leaves Prisma to subclasses; only
+`rest_api`, `ws_api` and `db_manager` call `db.connect()`. Everything the
+NotificationManager process runs therefore has to reach the database through
+the DatabaseManager RPC.
+
+This was worth a test rather than a docstring. The rule was documented and
```
ntindle: Remove this segment of the comment

ntindle (changes requested):

ntindle (approved):

PR #14011, opened 2026-08-12
fix(frontend): restore touch scroll on mobile artifact viewer

Otto-AGPT:
## Bug

Artifact content wasn't scrollable on mobile — the HTML previewer scrolled fine, but Markdown/code/JSON/CSV previews didn't.

## Root cause

The mobile artifact panel renders inside a Vaul `<Drawer.Content>` (`ArtifactPanel.tsx`), which Vaul tags with `data-vaul-drawer`. Vaul's injected global stylesheet sets `touch-action: none` on that selector so its own pointer-event JS can drive the swipe-to-close gesture.

This drawer is configured with `handleOnly` but never renders a `<Drawer.Handle>`, so Vaul's `onPointerDown` on the content itself is a no-op (`if (handleOnly) return`) — meaning the compensating logic that would normally let scrollable descendants pass touch events through never engages. Net effect: all native touch scroll is blocked for anything rendered directly inside the drawer.

HTML previews were unaffected because they render inside a sandboxed `<iframe>` — a separate document with its own default `touch-action: auto`, untouched by the parent page's CSS. Everything else (Markdown, code, JSON, CSV, ...) shares the same `overflow-y-auto` container in `ArtifactContent.tsx` and inherited the blocked touch-action from the drawer ancestor.

## Fix

Explicitly set `touch-action: pan-y` on the artifact scroll container in `ArtifactContent.tsx`, which overrides the ancestor's `touch-action: none` for that element and its descendants, restoring native vertical touch scrolling.

## Testing

Not able to test on a physical mobile device in this environment — recommend verifying on an actual iOS/Android device (or Chrome DevTools mobile emulation with touch simulation) before merging: open a Markdown artifact on mobile and confirm the panel scrolls.

### DIFF
--- a/autogpt_platform/frontend/src/app/(platform)/copilot/components/ArtifactPanel/components/ArtifactContent.tsx
+++ b/autogpt_platform/frontend/src/app/(platform)/copilot/components/ArtifactPanel/components/ArtifactContent.tsx
@@ -54,7 +54,17 @@ function ArtifactContentLoader({
   }
 
   return (
-    <div ref={scrollRef} className="flex-1 overflow-y-auto">
+    <div
+      ref={scrollRef}
+      className="flex-1 overflow-y-auto"
+      // Vaul's mobile drawer sets `touch-action: none` on its
+      // `[data-vaul-drawer]` root so its own JS can drive the swipe-to-close
+      // gesture. That blocks native touch scrolling for anything rendered
+      // directly inside it (unlike HTML previews, which scroll fine because
+      // they live in a separate iframe document). Override it here so touch
+      // scroll works on mobile.
+      style={{ touchAction: "pan-y" }}
+    >
       <ArtifactErrorBoundary
         artifactID={artifact.id}
         artifactTitle={artifact.title}
--- a/.github/workflows/platform-single-container-docker.yml
+++ b/.github/workflows/platform-single-container-docker.yml
@@ -1,0 +1,22 @@
+# Registration-only on master; selecting dev loads the publishing workflow from dev.
+name: AutoGPT Platform - Single-container image
+
+on:
+  workflow_dispatch:
+    inputs:
+      publish:
+        description: Publish the current dev commit to Docker Hub (select dev as the ref)
+        required: true
+        type: boolean
+        default: false
+
+permissions:
+  contents: read
+
+jobs:
+  registration-placeholder:
+    # workflow_dispatch always supplies a SHA, so this never allocates a runner.
+    if: ${{ github.sha == '' }}
+    runs-on: ubuntu-24.04
+    steps:
+      - run: ":"


autogpt_platform/frontend/src/app/(platform)/copilot/components/ArtifactPanel/components/ArtifactContent.tsx
```diff
@@ -54,7 +54,17 @@ function ArtifactContentLoader({
...
+      ref={scrollRef}
+      className="flex-1 overflow-y-auto"
+      // Vaul's mobile drawer sets `touch-action: none` on its
+      // `[data-vaul-drawer]` root so its own JS can drive the swipe-to-close
+      // gesture. That blocks native touch scrolling for anything rendered
+      // directly inside it (unlike HTML previews, which scroll fine because
+      // they live in a separate iframe document). Override it here so touch
+      // scroll works on mobile.
+      style={{ touchAction: "pan-y" }}
+    >
```
Pwuts, line 67: 🤖 🔴 **The root cause this rests on doesn't hold, so the change is a no-op.**

Vaul 1.1.2 really does ship `[data-vaul-drawer]{touch-action:none}` — that part of the description checks out. But a `touch-action` value on an ancestor does not constrain a scroll container nested inside it. Per MDN: *"the browser intersects the `touch-action` values of the touched element and its ancestors, **up to the one that implements the gesture (in other words, the first containing scrolling element)**."* This `overflow-y-auto` div is that first containing scrolling element, so the walk terminates here and the drawer above never enters the intersection.

Measured against the real Vaul `Drawer` bundled with this panel's exact props (`direction="right" handleOnly noBodyStyles modal={false}`), Chromium at 390×844 with touch, driving `Input.dispatchTouchEvent`:

| case | drawer `touch-action` | child `touch-action` | scrollTop after swipe |
|---|---|---|---|
| dev today (no change) | `none` | `auto` | **303** |
| this PR (`pan-y`) | `none` | `pan-y` | **302** |
| control: `none` on this div | `none` | `none` | **0** |

The control row is the load-bearing one — putting `touch-action: none` directly here stops the scroll dead, which proves the harness is sensitive to the property and that the first two rows are a real result rather than an instrument ignoring it.

Two things that corroborate it: Vaul registers no `touchmove` listener at all (`grep addEventListener` over `vaul/dist/index.mjs` → `resize` ×3, `scroll` ×1, `touchend` ×1), so there's no JS `preventDefault()` path either; and your `handleOnly` observation is accurate but only means Vaul's drag never engages — it doesn't block scrolling.

Caveat I want to be upfront about: this is **Chromium only**. WebKit won't launch here and Playwright has no touch-swipe outside Chromium's CDP, so if the bug was reported on iOS this neither reproduces nor refutes it there. The rule is engine-independent, but I haven't demonstrated it on WebKit.

What would clear this: reproduce the original bug on the engine it was reported on and show this diff fixes it there. Failing that, revert the hunk and chase the real cause — the missing `overscroll-behavior: contain` I've flagged on the next line is the leading candidate.

```suggestion
    <div ref={scrollRef} className="flex-1 overflow-y-auto">
```

autogpt_platform/frontend/src/app/(platform)/copilot/components/ArtifactPanel/components/ArtifactContent.tsx
```diff
@@ -54,7 +54,17 @@ function ArtifactContentLoader({
   }
 
   return (
-    <div ref={scrollRef} className="flex-1 overflow-y-auto">
+    <div
+      ref={scrollRef}
+      className="flex-1 overflow-y-auto"
```
Pwuts, line 59: 🤖 🟠 **The guard actually missing on this element is `overscroll-behavior: contain`.**

`grep -rn "overscroll" src/ --include=*.tsx --include=*.ts --include=*.css` returns **zero hits across the entire frontend**. Vercel's Web Interface Guidelines (which we run on every frontend diff) require `overscroll-behavior: contain` "in modals/drawers/sheets", and this is a drawer.

Without it, a touch scroll that reaches either end of the artifact chains through to the document behind the drawer. On a phone that reads as the panel refusing to scroll, or the page sliding around underneath it — which is plausibly the symptom that prompted this PR, and unlike `touch-action` it's a mechanism that is genuinely in play here.

This is independent of the revert I suggested above: whichever version of the element survives, it wants `overscroll-contain` on the className.

```suggestion
      className="flex-1 overflow-y-auto overscroll-contain"
```

autogpt_platform/frontend/src/app/(platform)/copilot/components/ArtifactPanel/components/ArtifactContent.tsx
Pwuts, line 67: 🤖 🟡 **`dev` has added a render path that bypasses this container.**

The branch is five weeks old, and `dev` has since put an early return at the top of `ArtifactContent()` (`ArtifactContent.tsx:321` on `dev`):

```tsx
if (props.artifact.expert) {
  return <ExpertArtifactContent expert={props.artifact.expert} />;
}
```

`ExpertArtifactContent` carries its own scroll container (`<div className="flex flex-1 flex-col gap-6 overflow-y-auto p-5">`) and never reaches `ArtifactContentLoader`. So even if this approach were kept, expert artifacts wouldn't be covered by it.

The merge itself is clean — the hunks don't overlap and the PR is `MERGEABLE` — this is a coverage gap, not a conflict. Flagging it because it's the kind of thing that silently makes a fix half-apply after a long-lived branch lands.

No suggestion block: this needs a decision about scope rather than an edit to this line.

autogpt_platform/frontend/src/app/(platform)/copilot/components/ArtifactPanel/components/ArtifactContent.tsx
```diff
@@ -54,7 +54,17 @@ function ArtifactContentLoader({
...
+      // Vaul's mobile drawer sets `touch-action: none` on its
+      // `[data-vaul-drawer]` root so its own JS can drive the swipe-to-close
+      // gesture. That blocks native touch scrolling for anything rendered
+      // directly inside it (unlike HTML previews, which scroll fine because
+      // they live in a separate iframe document). Override it here so touch
+      // scroll works on mobile.
```
Pwuts, line 65: 🤖 🟡 **Seven lines of comment for a one-line style.**

Six of the eleven added lines are comment, and it narrates the investigation rather than the constraint. The repo's standard is 1–2 lines, why-not-what.

I'm deliberately not attaching a `suggestion` here: this comment is the vehicle for the root cause in the blocker above, so the right resolution is the revert I suggested there rather than a reworded version of a claim that doesn't hold. If the element does survive in some form, the comment wants to be a single line stating the constraint that forced it.

autogpt_platform/frontend/src/app/(platform)/copilot/components/ArtifactPanel/components/ArtifactContent.tsx
```diff
@@ -54,7 +54,17 @@ function ArtifactContentLoader({
...
+      style={{ touchAction: "pan-y" }}
```
Pwuts, line 66: 🤖 🔵 **Tailwind has a `touch-pan-y` utility for this.**

CodeRabbit raised this twice (comment ids `3768619947` and `3768619961`, both 🟡). It's a fair point on the guideline: `AGENTS.md` says "Tailwind CSS only for styling", and `touch-pan-y` does exist in the pinned Tailwind 3.4.17 (`corePlugins.js`).

Against it, `PanelResizeHandle.tsx:114` in this same feature area already uses `style={{ touchAction: "none" }}`, so there's a local precedent for the inline form — I wouldn't spend a round on this either way.

Moot if the blocker resolves by reverting the hunk, which is why there's no `suggestion` block: applying one would rebuild a line I'm asking to remove.

Pwuts (changes requested):
## 🤖 Review — restore touch scroll on mobile artifact viewer

This PR reports that the mobile artifact panel could not be scrolled by touch for every artifact type except HTML, and fixes it by setting `touch-action: pan-y` on the panel's scroll container — on the theory that Vaul's `[data-vaul-drawer]{touch-action:none}` blocks native scrolling for everything inside the drawer.

The first half of that is true; Vaul 1.1.2 does ship that rule. The second half is not, and I measured it: an ancestor's `touch-action` does not gate a scroll container nested inside it, because the browser intersects `touch-action` only from the touched element up to the *first containing scrolling element* — which is the `overflow-y-auto` div this PR edits. Against the real Vaul `Drawer` with this panel's exact props, that container scrolls **303 px with no change and 302 px with the change**, while `touch-action: none` placed directly on it gives **0 px** — the control proving the measurement is sensitive to the property.

So this would close a mobile bug without fixing it, and land a seven-line comment asserting a mechanism that isn't real. Nothing regresses — the change is inert, not harmful.

Caveat stated plainly: the measurement is **Chromium only**. WebKit can't launch on my machine and Playwright has no touch-swipe outside Chromium, so if this was reported on an iPhone it is neither reproduced nor refuted here. The rule is engine-independent, but I haven't shown it on WebKit.

Auto-merge is off and this isn't in the merge queue, so an approval would not merge it on its own.

Note for anyone reading the diff: GitHub shows 2 files / +33 because its pinned base predates #13995 landing on `dev`. The effective changeset is `ArtifactContent.tsx` alone, +11/−1.

### 🔴 Blockers
- 🔴 Stated root cause does not hold — an ancestor `touch-action` does not gate a descendant scroll container, so this change is a no-op (303 px without it, 302 px with it, 0 px with the control) (`ArtifactContent.tsx:57`)

### 🟠 Should Fix
- 🟠 If the scroll complaint is real, the guard actually missing on this element is `overscroll-behavior: contain` — absent from the entire frontend, zero hits in `src/` (`ArtifactContent.tsx:59`)

<details>
<summary>What I executed</summary>

- `npx tsc --noEmit` over the frontend — 0 errors, none in `ArtifactPanel/`
- CI on `d520b13c9d` — GREEN
- Playwright touch harnesses at a 390×844 touch viewport in Chromium, driving `Input.dispatchTouchEvent`, against both a synthetic DOM and the **real** bundled `vaul@1.1.2` `Drawer` with `direction="right" handleOnly noBodyStyles modal={false}`:

| case | drawer `touch-action` | child `touch-action` | scrollTop |
|---|---|---|---|
| dev today | `none` | `auto` | 303 |
| this PR | `none` | `pan-y` | 302 |
| `auto` on child | `none` | `auto` | 302 |
| control: `none` on child | `none` | `none` | 0 |

- Horizontal check — CSV and code artifacts keep their own `overflow-x-auto`, and `pan-y` on the outer container does not block them (214 px either way). **No regression from this change.**
- `grep addEventListener` over `vaul/dist/index.mjs` — only `resize` ×3, `scroll` ×1, `touchend` ×1. No `touchmove` listener, so no JS `preventDefault()` path either.
- Mutation testing: not applicable — the PR adds no tests.

**Not exercised:** WebKit/iOS (`webkit.launch()` → "Host system is missing dependencies to run browsers"), Firefox (no CDP touch API), and no real device.
</details>

PR #14014, opened 2026-08-13
fix(frontend): render Ayrshare connect button on all credential surfaces

Otto-AGPT:
## Problem

The **Connect Social Media Accounts** button (Ayrshare SSO) only rendered inside the Flow Editor's block canvas (`CustomNode`). Anywhere else that needs the user to connect an Ayrshare-managed credential — the chat `SetupRequirementsCard`, `RunInputDialog`, the library `RunAgentModal`, and the legacy `agent-run-draft-view` — only showed static help text with no way to actually connect.

## Fix (Option 1, per @Abhi's call)

`AyrshareConnectButton` is moved from `CustomNode/components` into the shared `CredentialsInput` component tree, and rendered directly from `CredentialsFlatView` — the component all of those surfaces already share — instead of only from `CustomNode`.

- Moved `AyrshareConnectButton` (+ its test) to `components/contextual/CredentialsInput/components/AyrshareConnectButton/`.
- `CredentialsFlatView` now renders `<AyrshareConnectButton />` for the managed-only `ayrshare` provider, both when the user has an existing credential (to link additional networks) and when they have none yet (first-time connect).
- `CustomNode.tsx` updated to import from the new shared path; Flow Editor behavior is unchanged.

## Verification

- Reviewed `useToast`/`sonner` and `CredentialsActionsContext` (provided app-wide via `app/providers.tsx`) to confirm the button renders safely with no additional context wiring needed on any surface.
- Manually reviewed the existing `CredentialsFlatView` tests against the new logic — no behavior changes expected for non-Ayrshare providers.
- Could not run the JS test suite in this sandbox (memory-constrained, `pnpm install` OOMs); relying on CI to run `test:unit` on this PR.


autogpt_platform/frontend/src/components/contextual/CredentialsInput/components/CredentialsFlatView/CredentialsFlatView.tsx
```diff
@@ -134,6 +135,7 @@ export function CredentialsFlatView({
               {actionButtonText}
             </Button>
           )}
+          {showAyrshareConnect && <AyrshareConnectButton />}
```
Abhi1992002: 🔴 **Blocker**: This double-renders the connect button in the Flow Editor.

`CustomNode.tsx:109` still renders `{isAyrshare && <AyrshareConnectButton />}` at the top of the node, and every Ayrshare block also renders a credentials field (`BaseAyrshareInput.credentials` in `backend/blocks/ayrshare/_util.py:17`, not `advanced`) → `FormCreator` → `CredentialField` → `CredentialsInput` → `CredentialsFlatView`, which now renders a second identical **Connect Social Media Accounts** button below.

So on the builder canvas the user sees the button twice. The PR description says "Flow Editor behavior is unchanged" — it is changed: the old code rendered a text hint there ("Click **Connect Social Media Accounts** above…"), precisely because the button already existed above.

Suggest dropping the `{isAyrshare && <AyrshareConnectButton />}` line from `CustomNode.tsx` (and the now-unused import) so `CredentialsFlatView` is the single source of truth. That also removes the last reason for `BlockUIType.AYRSHARE` special-casing in the node.

autogpt_platform/frontend/src/components/contextual/CredentialsInput/components/CredentialsFlatView/CredentialsFlatView.tsx
```diff
@@ -53,12 +54,12 @@ export function CredentialsFlatView({
...
-  // server after the user clicks the explicit Connect Social Media
-  // Accounts button rendered alongside the block.  Exposing "Add API
-  // key" / "Use a new API key" here just confuses users into entering a
-  // random key.
+  // server after the user clicks the Connect Social Media Accounts
+  // button rendered below. Exposing "Add API key" / "Use a new API key"
+  // here just confuses users into entering a random key.
   const isManagedOnlyProvider = provider === "ayrshare";
   const showAddAction = !readOnly && !isManagedOnlyProvider;
+  const showAyrshareConnect = isManagedOnlyProvider && !readOnly;
```
Abhi1992002, line 62: 🟠 **Should Fix**: No test covers the new behavior. `__tests__/CredentialsFlatView.test.tsx` was left untouched, and the moved `AyrshareConnectButton.test.tsx` only tests the button in isolation — nothing asserts that `CredentialsFlatView` renders it.

The `codecov/patch/platform-frontend` gate is on changed lines, so these branches will count against it. Three cheap cases on the existing `makeProps()` helper (provider is already `"ayrshare"`):
- `credentials: []` → button present (first-time connect)
- `credentials: [managed]` → button present (link more networks)
- `readOnly: true` → button absent

Given `pnpm install` OOMed in your sandbox, worth calling out explicitly that these were never executed locally.

autogpt_platform/frontend/src/components/contextual/CredentialsInput/components/CredentialsFlatView/CredentialsFlatView.tsx
```diff
@@ -145,15 +147,9 @@ export function CredentialsFlatView({
...
-        isManagedOnlyProvider &&
-        !readOnly && (
-          <Text variant="body" className="text-zinc-500">
-            Click <strong>Connect Social Media Accounts</strong> above to set up
-            your managed {displayName} profile.
-          </Text>
-        )
-      )}
+      ) : showAyrshareConnect ? (
+        <AyrshareConnectButton />
```
Abhi1992002: 🟡 **Nice to Have**: `AyrshareConnectButton` hardcodes node-canvas spacing — `<div className="mt-4 flex flex-col gap-2 px-4">` (file is a pure rename so I could not comment on it directly).

That `px-4` exists because `CustomNode` rendered it full-bleed against the node shell. Now that it also renders here, inside a `CredentialsFlatView` that `RunAgentModal` / `SetupRequirementsCard` / `RunInputDialog` already pad, the button picks up an extra 16px inset and will not align with the fields above it.

Two very different hosts now — suggest dropping the wrapper padding and letting the caller position it (or accepting a `className`).

autogpt_platform/frontend/src/components/contextual/CredentialsInput/components/CredentialsFlatView/CredentialsFlatView.tsx
Abhi1992002: Fixed in a6384c1295 — removed `{isAyrshare && <AyrshareConnectButton />}` (plus the now-unused import and `isAyrshare`) from `CustomNode.tsx`, so `CredentialsFlatView` is the single source.

`src/app/(platform)/build/__tests__/CustomNode.test.tsx` was pinning the old behavior *and* was already stale — its `vi.mock` still pointed at the pre-rename path, so it was asserting on a testid that no longer existed. Dropped the mock and flipped both cases to assert the node shell renders no `Connect Social Media Accounts` button of its own.

autogpt_platform/frontend/src/components/contextual/CredentialsInput/components/CredentialsFlatView/CredentialsFlatView.tsx
Abhi1992002: Fixed in a6384c1295 — added four cases to `CredentialsFlatView.test.tsx`: button shown with no credential, still shown once a managed credential exists, hidden when `readOnly`, and absent for a non-Ayrshare provider (which still gets "Add API key").

Ran locally after `pnpm generate:api`: `pnpm format`, `pnpm lint`, `pnpm types` clean, `pnpm test:unit` → 466/466 files, 4875 tests passing.

autogpt_platform/frontend/src/components/contextual/CredentialsInput/components/CredentialsFlatView/CredentialsFlatView.tsx
Abhi1992002: Fixed in a6384c1295 — dropped the `mt-4 … px-4` wrapper `div`; the component now renders the `Button` directly and takes an optional `className`. Both `CredentialsFlatView` call sites pass `mt-2`, so spacing is owned by the caller and the modal surfaces no longer get the node-canvas inset.

Converted `AyrshareConnectButton` / `handleSSOLogin` to function declarations in the same commit (the nit CodeRabbit also raised).

autogpt_platform/frontend/src/components/contextual/CredentialsInput/components/AyrshareConnectButton/AyrshareConnectButton.tsx
```diff
@@ -1,20 +1,25 @@
 "use client";
 
-import React, { useContext, useState } from "react";
+import { useContext, useState } from "react";
 
 import { Key } from "lucide-react";
```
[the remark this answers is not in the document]
Abhi1992002: Fixed in 1efde0d072 — swapped the lucide-react `Key` for `Key01Icon` from `@hugeicons/core-free-icons`, rendered through the `Icon` atom (`<Icon icon={Key01Icon} size={16} />`) per `frontend/AGENTS.md`.

Abhi1992002 (approved):
tested it on the preview and also reviewed the code. It looks good to merge.

PR #14020, opened 2026-08-13
fix(platform/llm): allow Ollama without credentials

### LINKED ISSUE #8952: ollama llm requires credentials
ollama llm blocks require credentials but that shouldn’t be required

Iitian001:
### Why / What / How

Fixes #8952.

**Why:** Ollama runs locally and does not require an API key, but the LLM credential schema and graph-level credential aggregation assume that every provider has a user-managed credential. Treating the shared credential field as globally optional would weaken validation for OpenAI, Anthropic, and other authenticated providers, so the no-auth case has to be scoped specifically to Ollama.

**What:** Make Ollama a credential-free discriminator variant while preserving the existing required-credential contract for providers that actually require authentication.

**How:**
- Ollama advertises no static authentication method.
- `AICredentialsField` maps Ollama model discriminators to zero supported credential types but keeps the shared credentials field required at the schema level.
- Graph credential aggregation skips discriminator variants with zero supported credential types, avoiding an empty credential schema/`cred_types[0]` path.
- Pre-execution credential validation recognizes no-auth discriminator variants before credential lookup, so Ollama graphs are not rejected before they are queued.
- The executor recognizes the same no-auth Ollama discriminator and injects the existing internal Ollama compatibility credential only at runtime, so block signatures and Pydantic validation remain unchanged without requiring a saved user credential.
- Authenticated providers continue through the normal credential lookup/preflight path and remain required.
- The built-in Ollama compatibility credential is no longer exposed in the user's credential list.
- Ollama's zero-cost configuration is keyed by model rather than by the compatibility credential.
- The frontend renders no credential action when a discriminator supports no credential type, avoiding a dead “Add credentials” button.

### Backward compatibility

Existing Ollama graphs that still contain the old built-in credential reference remain compatible: preflight recognizes the Ollama model as credential-free, and execution replaces any stale stored credential metadata with the internal no-auth runtime adapter before user credential lookup. New Ollama graphs do not need to expose or select that compatibility credential.

### Regression coverage

- [x] Ollama discriminates to zero supported credential types.
- [x] The shared LLM credential field remains required, so authenticated providers are not globally relaxed.
- [x] Preflight accepts an Ollama node with no saved credential.
- [x] Preflight still rejects an authenticated provider such as OpenAI when credentials are missing.
- [x] The no-auth runtime adapter is selected for Ollama.
- [x] OpenAI still discriminates to `api_key` and does not receive the no-auth adapter.
- [x] Credential-free providers render no credential control in the frontend.
- [x] Frontend lint passes.
- [x] Frontend generated API setup and credential UI regression tests pass.
- [x] Backend upstream lint gate passes.
- [x] Pyright passes for Python 3.11, 3.12, and 3.13.
- [x] Focused backend no-auth and preflight regression tests pass.
- [x] Generated block documentation is in sync.

### Configuration

- [x] `.env.default` is already compatible; no environment variable changes are required.
- [x] `docker-compose.yml` is already compatible; no service changes are required.
- [x] No new external secrets or infrastructure are introduced.


autogpt_platform/backend/backend/blocks/llm.py
```diff
@@ -531,7 +538,7 @@ async def llm_call(
         )
 
     async def run(
-        self, input_data: Input, *, credentials: APIKeyCredentials, **kwargs
+        self, input_data: Input, *, credentials: APIKeyCredentials | None, **kwargs
```
kcze: 🤖 🔴 **Blocker**: `credentials` stays keyword-only with **no default**, so the PR's headline path — an Ollama node with no credentials — never reaches the new `| None` handling.

The executor's no-credentials branch (`executor/manager.py:293-299`) sets `input_data[field_name] = None` and `continue`s **without** writing `extra_exec_kwargs["credentials"]`; `_base.py:924` then calls `self.run(...)` without the kwarg → `TypeError: run() missing 1 required keyword-only argument: 'credentials'`. The in-tree working pattern is `credentials: OAuth2Credentials | None = None` (`blocks/mcp/block.py:210`).

Same missing default at :959, :1051, :1278, :1426 and `orchestrator.py:1984`. Also note `blocks/ai_condition.py:151` and `blocks/stagehand/blocks.py:132-133,224,307` consume the same `AICredentialsField()` but were not updated at all — their signatures are still non-optional and they dereference `.api_key.get_secret_value()` directly, so those nodes now pass schema validation and then crash. _(flagged by: Claude + Codex)_

autogpt_platform/backend/backend/blocks/llm.py
```diff
@@ -96,11 +96,16 @@
...
+        description="API key for the LLM provider. Not required for Ollama.",
         discriminator="model",
         discriminator_mapping={
             model.value: model.metadata.provider for model in LLMModel
         },
+        discriminator_type_mapping={
+            model.value: [] if model.metadata.provider == ProviderName.OLLAMA else ["api_key"]
+            for model in LLMModel
+        },
+        default=None,
```
kcze: 🤖 🔴 **Blocker**: `default=None` is on the shared field for **every** model, not just Ollama — only `discriminator_type_mapping` is per-model, and it governs credential *type*, not requiredness. This is the "relaxation isn't narrowed to Ollama" hole.

`get_required_fields()` filters on `field_info.is_required()` (`_base.py:330-335`), so `credentials` drops out for all LLM blocks; `graph.py:559-566` then computes `creds_required_by_schema=False` and the aggregated credentials input is marked not-required for every provider. An agent with an OpenAI node can be set up and launched with no key attached.

It also silently disables pre-flight validation of credentials that *are* configured: `executor/utils.py:497` computes `field_is_optional = ... or field_name not in required_fields`, and lines 510-513 then deliberately discard `node.input_default` for optional fields. Delete your OpenAI key and instead of failing fast with `CRED_ERR_REQUIRED`, the graph starts, upstream nodes run and charge, and it dies mid-run at `manager.py:326-337`. Gate the optionality on the provider rather than defaulting the shared field. _(flagged by: Claude + Codex)_

autogpt_platform/backend/backend/blocks/llm.py
```diff
@@ -96,11 +96,16 @@
 
 def AICredentialsField() -> AICredentials:
     return CredentialsField(
-        description="API key for the LLM provider.",
+        description="API key for the LLM provider. Not required for Ollama.",
         discriminator="model",
         discriminator_mapping={
             model.value: model.metadata.provider for model in LLMModel
         },
+        discriminator_type_mapping={
```
kcze, line 104: 🤖 🔴 **Blocker**: mapping Ollama models to `[]` makes `CredentialsFieldInfo.discriminate()` return empty `supported_types` (`data/model.py:749-760`), and nothing downstream tolerates that.

`CredentialsFieldInfo.combine()` keys groups on `frozenset(field.supported_types)`, so the empty entry survives as its own group; `GraphModel.credentials_input_schema` then does `{"const": cred_types[0]}` when `len(cred_types) <= 1` (`data/graph.py:491-493`) → **IndexError for any graph containing an Ollama LLM node**. Either exclude credential-free fields from the aggregation or handle a zero-length type list at that branch.

Secondary effect: with no supported types the frontend still renders the control — `getActionButtonText()` returns "Add credentials" while `resolveActionTarget()` returns `null` (`CredentialsInput/helpers.ts:279-292`), so the button does nothing when clicked. _(flagged by: Codex)_

autogpt_platform/backend/backend/blocks/llm.py
```diff
@@ -290,6 +295,8 @@ async def _llm_call(
             - completion_tokens: The number of tokens used in the completion.
     """
     provider = llm_model.metadata.provider
+    if provider != "ollama" and not credentials:
```
kcze: 🤖 🟠 **Should Fix**: this guard sits inside `_llm_call`, which runs within the block's `for retry_count in range(input_data.retry)` loop under a broad `except Exception` (:718-749) that doesn't classify it as a user error. A missing OpenAI key is therefore retried N times, each emitting a full `logger.exception` traceback (Sentry noise), before surfacing as `RuntimeError: Error calling LLM: Credentials are required for provider openai`. The check belongs before the retry loop.

The diff also adds no tests. Nothing asserts (a) an Ollama node actually runs with `credentials=None` — one such test would have caught the missing `= None` default above — nor (b) that a non-Ollama provider without credentials is still rejected. The safety half of this change is entirely unverified. _(flagged by: Claude)_

autogpt_platform/backend/backend/blocks/llm.py
```diff
@@ -96,11 +96,16 @@
...
 def AICredentialsField() -> AICredentials:
     return CredentialsField(
-        description="API key for the LLM provider.",
+        description="API key for the LLM provider. Not required for Ollama.",
         discriminator="model",
         discriminator_mapping={
             model.value: model.metadata.provider for model in LLMModel
         },
+        discriminator_type_mapping={
+            model.value: [] if model.metadata.provider == ProviderName.OLLAMA else ["api_key"]
```
kcze: 🤖 🟡 **Nice to Have**: this comprehension isn't formatted — `poetry run black --check backend/blocks/llm.py` reports `1 file would be reformatted` on this branch, so the backend CI format job fails. _(flagged by: Claude)_

autogpt_platform/backend/backend/blocks/llm.py
```diff
@@ -96,11 +96,16 @@
 
 def AICredentialsField() -> AICredentials:
     return CredentialsField(
-        description="API key for the LLM provider.",
+        description="API key for the LLM provider. Not required for Ollama.",
```
kcze, line 99: 🤖 🟡 **Nice to Have**: two knock-on effects worth confirming against the stated motivation.

`data/block_cost_config.py:358-372` defines the explicit zero-cost `BlockCost` for Ollama models with a filter on `credentials: {id, provider, type}`, and `_is_cost_filter_match` (`executor/utils.py:287-294`) does subset matching — with credentials absent the entry never matches, so the builder's cost label falls back to unknown, which is exactly what that entry's comment says it exists to prevent.

Also, `integrations/credentials_store.py:50,362` already injects a built-in "Use Credits for Ollama" credential for every user, so the premise that users "must provide a dummy credential" is only partly accurate — and existing Ollama graphs still reference that ID. Worth stating in the PR description how those graphs migrate. _(flagged by: Claude)_

autogpt_platform/backend/backend/blocks/llm.py
```diff
@@ -1186,7 +1199,7 @@ class AIConversationBlock(AIBlockBase):
     class Input(BlockSchemaInput):
         prompt: str = SchemaField(
             description="The prompt to send to the language model.",
-            placeholder="Enter your prompt here...",
+            placeholder="Enter the prompt here...",
```
kcze: 🤖 🔵 **Nit**: unrelated copy change ("Enter your prompt here..." → "Enter the prompt here...") in `AIConversationBlock`, outside the PR's stated scope. _(flagged by: Claude)_

